PR #251, opened 2015-06-28
Fix ecma-property construction

ruben-ayrapetyan:
Perform operations that could invoke GC before linking property into an object's property list to avoid start of GC in the middle of property construction.


zherczeg:
Other engines have "do GC before each malloc" modes. We should add something similar to Jerry to find such errors.

zherczeg:
lgtm

PR #253, opened 2015-06-28
Fix ecma_op_get_value_object_base

ruben-ayrapetyan:
Fix ecma_op_get_value_object_base (GetValue, 8.7.1) for case HasPrimitiveBase (ref) is true


zherczeg:
lgtm

PR #268, opened 2015-06-29
Fix construction of Arguments object: add handler for 'caller' property that throws TypeError upon read or write access

ruben-ayrapetyan:


galpeter:
lgtm

egavrin:
`make push`

PR #270, opened 2015-06-29
Fix syntax check for case 'eval' or 'arguments' identifiers are used for name of function in strict mode

ruben-ayrapetyan:


galpeter:
lgtm

PR #271, opened 2015-06-29
Fix syntax check for duplicate parameter names in a strict mode function

ruben-ayrapetyan:


galpeter:
With this patch what happens if there is a code like this?
```js
function f(x,x) { return x; }
print(f(1,2));
```

If my memory serves right there is a testcase like this in the test262 and the result should be the 2nd argument.

ruben-ayrapetyan:
@galpeter, the patch should not change behaviour of your test case.

Please, look at #273 pull request, that fixes behaviour of the test case.

galpeter:
@ruben-ayrapetyan, okay, thanks for the info. 
lgtm

egavrin:
@ruben-ayrapetyan please rebase it to master and I'll push it.

PR #272, opened 2015-06-29
Fix propagation of strict mode

ruben-ayrapetyan:


tests/jerry/func-decl.js
```diff
@@ -23,3 +23,27 @@ assert ((function() {
     }
     return f();
 })() === 'bar');
+
+function check_syntax_error (s) {
```
galpeter: Is this function used at all in the test?

tests/jerry/func-decl.js
```diff
@@ -23,3 +23,27 @@ assert ((function() {
...
+  try {
+    eval (s);
+    assert (false);
+  }
+  catch (e) {
+    assert (e instanceof SyntaxError);
+  }
+}
+
```
galpeter: Would it be possible to create a test case which fails (as expected) with this patch and passes (incorrectly) without it?

tests/jerry/func-decl.js
ruben-ayrapetyan: No, it is incorrect merge. The function is from another patch.
That patch was already merged, so I'll rebase this pull request.

tests/jerry/func-decl.js
ruben-ayrapetyan: Actually, there is already a test case in the patch that satisfies to described conditions - the function below (lines 37-49).
```
+function test_strict_mode_propagation_in_func_expr_and_getters_setters () {
+  var p = function () {
+    'use strict';
+
+    return true;
+  }
+
+  var o = { get prop () { 'use strict'; return true; }, set prop (v) { 'use strict'; } };
+
+  function test () {
+    eval = 1;
+  }
+}
```

ruben-ayrapetyan:
@galpeter, pull request is updated:
 - rebased on current master;
 - improved test - now it doesn't actually change value of `eval`.

tests/jerry/func-decl.js
```diff
@@ -38,3 +38,18 @@ check_syntax_error ("'use strict'; function arguments () {}");
 check_syntax_error ("'use strict'; var l = function arguments () {}");
 
 check_syntax_error ("function f__strict_mode_duplicate_parameters (p, p) { 'use strict'; }");
+
+function test_strict_mode_propagation_in_func_expr_and_getters_setters () {
```
galpeter: So are this test fails with this PR? If so then shouldn't this be inside one of the tests/jerry/fail/* directory to test that it fails correctly? Or did I misunderstood something?

ruben-ayrapetyan: Seems that I have misunderstood what test case do you ask for.
The test in the PR fails without the patch, and passes with the patch.
Issue, solved by the patch, is incorrect propagation of strict mode to outer scopes.

Yes, it is possible to construct test for `fail` folder. Could you, please, describe why do we need it?

galpeter: I was thinking about that what if future modifications/improvements a case occurs when this test passes and also some other cases passes when in should return an error for correct behavior.

Just like some parser tests when we create tests to accept correct input, but there are also tests to see if it does not parse some weird cases. BTW: maybe this does not need to be in the `fail` folder, but instead instead in some kind of try-catch.

ruben-ayrapetyan:
@galpeter, it is good to add regression tests.

In the case, we check that "strict mode" isn't propagated to outer scope from "function expressions" and "getters" / "setters".

Test case, added in the pull request, checks this by trying to perform `eval = value` assignment in outer scope. The assignment is not permitted in "strict mode", but is ok without it.

We could also formulate the condition "strict mode is not propagated" with some another test case, but wouldn't the test case be redundant?

Are there any additional conditions that need to be checked? Could you, please, clarify them?

tests/jerry/func-decl.js
egavrin: @galpeter are you for something like this test:
```cpp
        try {
            // bla-bla-bla
        } catch (e) {
            return e instanceof SyntaxError;
        }
```

galpeter: @egavrin, yes I was thinking about something like that (aside from the fact that it may not throw syntax error but something else)

galpeter:
@ruben-ayrapetyan, you could be right, maybe it does only just give redundancy. Currently I don't see any other conditions to check, so we should go forward with this PR in the current state. If we find some additional condition then we will create a test for it.
So the patch looks good to me.

egavrin:
`make push`

Issue #114, opened 2015-05-28
Assertion 'simple_value == ECMA_SIMPLE_VALUE_UNDEFINED || simple_value == ECMA_SIMPLE_VALUE_NULL || simple_value == ECMA_SIMPLE_VALUE_FALSE || simple_value == ECMA_SIMPLE_VALUE_TRUE' failed in ecma_make_simple_completion_value


renatahodovan:
###### Jerry version:
    Checked revision: aaeec7f
    Build: debug.linux
###### OS: 
    Ubuntu 14.04, x86_64
###### Test case:
```javascript
v_1 = 1;
v_1[1] = 1;
```
###### Backtrace:
```javascript
ICE: Assertion 'simple_value == ECMA_SIMPLE_VALUE_UNDEFINED || simple_value == ECMA_SIMPLE_VALUE_NULL || simple_value == ECMA_SIMPLE_VALUE_FALSE || simple_value == ECMA_SIMPLE_VALUE_TRUE' failed at jerryscript/jerry-core/ecma/base/ecma-helpers-value.cpp(ecma_make_simple_completion_value):597.
Error: ERR_FAILED_INTERNAL_ASSERTION
(gdb) bt
# 0  ecma_make_simple_completion_value (simple_value=ECMA_SIMPLE_VALUE_EMPTY)
    at jerryscript/jerry-core/ecma/base/ecma-helpers-value.cpp:594
# 1  ecma_reject_put (is_throw=false) at jerryscript/jerry-core/ecma/operations/ecma-get-put-value.cpp:192
# 2  0x000000000047cd89 in ecma_op_put_value_object_base (ref=..., value=73)
    at jerryscript/jerry-core/ecma/operations/ecma-get-put-value.cpp:279
# 3  0x0000000000454062 in opfunc_prop_setter (opdata=..., int_data=0x7fffffffd590) at jerryscript/jerry-core/vm/opcodes.cpp:1171
# 4  0x00000000004bfa60 in run_int_loop (int_data=0x7fffffffd590) at jerryscript/jerry-core/vm/vm.cpp:440
# 5  0x00000000004b678d in run_int_from_pos (start_pos=0, this_binding_value=15, lex_env_p=0x7e9770 <_ZL13mem_heap_area.4015+48>, is_strict=false, 
    is_eval_code=false) at jerryscript/jerry-core/vm/vm.cpp:513
[...]

PR #288, opened 2015-07-01
Fix generation of empty completion value in ecma_reject_put.

egavrin:
Related issue: #114

JerryScript-DCO-1.0-Signed-off-by: Evgeny Gavrin e.gavrin@samsung.com

### DIFF
--- a/jerry-core/ecma/operations/ecma-get-put-value.cpp
+++ b/jerry-core/ecma/operations/ecma-get-put-value.cpp
@@ -226,7 +226,7 @@ ecma_reject_put (bool is_throw) /**< Throw flag */
   }
   else
   {
-    return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_EMPTY);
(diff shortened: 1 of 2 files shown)


ruben-ayrapetyan:
Looks good to me

galpeter:
What was the problem with the first solution (https://github.com/Samsung/jerryscript/commit/ae321fc3ad901736775688041ec3e7b2c873ccec)? Where you changed the `ecma_make_simple_completion_value` assert?

For me it seems natural that if there is a call to the `ecma_make_simple_completion_value` then I would be able to create an empty value as the name `ECMA_SIMPLE_VALUE_EMPTY` clearly states that it is a simple value.

egavrin:
@galpeter technically it's assumed that we can only create `ECMA_SIMPLE_VALUE_EMPTY` from constant value, not from `ecma_simple_value_t`. So, it's the way to check that this assumption is respected.

galpeter:
Looks good to me.

PR #292, opened 2015-07-01
Add helpers for checking if a character falls into an ECMA-specified character category

ruben-ayrapetyan:


egavrin:
Good to me.

zherczeg:
lgtm
However, implementing toLower/toUpper might affect these routines, so it would be good to think about this at this stage.

ruben-ayrapetyan:
Pull request is updated:
 - characters from ECMA-262 v5, Table 4 were added;
 - `ECMA_CHAR_NULL` was renamed to `LIT_CHAR_NULL.`

PR #294, opened 2015-07-02
Fix jmp_buf size for arm-hf target

seanshpark:
Related issue : https://github.com/Samsung/jerryscript/issues/231

### DIFF
--- a/jerry-libc/include/setjmp.h
+++ b/jerry-libc/include/setjmp.h
@@ -47,14 +47,14 @@
  *   - return address (to jump to upon longjmp)
  *   - x87 control word
  *
- * ARMv7 (8 * 4 + 16 * 4 bytes):
- *   - r4 - r11
+ * ARMv7 (10 * 4 + 16 * 4 bytes):
+ *   - r4 - r11, sp, lr
  *   - s16 - s31 (if hardfp enabled)
  *
  * See also:
  *          setjmp, longjmp
  */
-typedef uint64_t jmp_buf[12];
+typedef uint64_t jmp_buf[14];
 
 extern EXTERN_C int setjmp (jmp_buf env);
 extern EXTERN_C void longjmp (jmp_buf env, int val);


egavrin:
Good to me.

ruben-ayrapetyan:
@seanshpark, the `jerry-libc/include/setjmp.h` should not be used in nuttx build, as `jerry-libc` is not used for nuttx, because nuttx has its own libc.

As nuttx' libc doesn't have implementation of `setjmp`, seems that there is no `setjmp.h` header in the nuttx's libc include folder. Maybe, the header with the definition should be added to the `nuttx/nuttx/include/` folder.

seanshpark:
@ruben-ayrapetyan , ok, I'll copy setjmp.h to nuttx include and check how it works.
I think this can fix #296 . can you please check in your side?

ruben-ayrapetyan:
> can you please check in your side?

@seanshpark, I would gladly do this, but before this I should finish some activities. I'll do this as soon as possible.

seanshpark:
@ruben-ayrapetyan , thank you. take your time :)

ruben-ayrapetyan:
> I think this can fix #296 . can you please check in your side?

@seanshpark, I've checked. The patch fixes armv7-hf target.
Without the changes `print ("hello")` fails with `'al != null_list' assertion failed`.

Could you, please, update register list in the description of contents of `jmp_buf`?
Currently, the `sp` and `lr` registers are not listed.

```
diff --git a/jerry-libc/include/setjmp.h b/jerry-libc/include/setjmp.h
index 35ae5ae..521b63d 100644
--- a/jerry-libc/include/setjmp.h
+++ b/jerry-libc/include/setjmp.h
@@ -47,14 +47,14 @@
  *   - return address (to jump to upon longjmp)
  *   - x87 control word
  *
- * ARMv7 (8 * 4 + 16 * 4 bytes):
- *   - r4 - r11
+ * ARMv7 (10 * 4 + 16 * 4 bytes):
+ *   - r4 - r11, sp, lr
  *   - s16 - s31 (if hardfp enabled)
```

seanshpark:
Done, maybe change title? it's not for nuttx.

ruben-ayrapetyan:
@seanshpark, thank you!

ruben-ayrapetyan:
The patch looks good to me.

galpeter:
lgtm

egavrin:
`make push`

PR #297, opened 2015-07-02
Implement Date.parse

LaszloLango:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
@@ -15,8 +15,11 @@
  */
 
 #include "ecma-alloc.h"
+#include "ecma-builtin-helpers.h"
+#include "ecma-conversion.h"
 #include "ecma-globals.h"
 #include "ecma-helpers.h"
+#include "ecma-try-catch-macro.h"
 
 #ifndef CONFIG_ECMA_COMPACT_PROFILE_DISABLE_DATE_BUILTIN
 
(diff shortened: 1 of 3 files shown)


tests/jerry/date-parse.js
```diff
@@ -0,0 +1,58 @@
...
+var wrongFormats = ["", "2", "20", "201", "2015-", "2015-01-", "2015-01-01-",
+                    "qwerty", "2015-01-01T", "2015-01-01T1:1", "2015-01-01T01",
+                    "2015-01-01T01", "2015-01-01T01:01F", "T2015", "2015-01-01Z",
+                    "2015-01-01+01:00", "2015-01-01T00:00+01", "2015-01-01T00:00+1",
+                    "2015-01-01T00:00-01", "2015-01-01T00:00.000", "2015-01-01T00:00:",
+                    "2015-01-01T00:", "2015-01-01T00:00:00.1", "2015-01-01T00:00:00.01",
+                    "2015-01-01T00:00+01:00Z", "2015/01/01", "2015-01-32", "2015--1",
+                    "2015-13", "2015-01--1", "-215", "-215-01-01", "2015-01-00",
+                    "2015-01-01T25:00", "2015-01-01T00:60", "2015-01-01T-1:00",
+                    "2015-01-01T00:-1"];
```
galpeter: Could we have this as one entry per line? Also maybe some other types like undefined, numbers, objects?

LaszloLango: OK

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,258 @@
...
                          ecma_value_t arg) /**< string */
 {
-  ECMA_BUILTIN_CP_UNIMPLEMENTED (this_arg, arg);
+  ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
+  ecma_number_t *date_num_p = ecma_alloc_number ();
+  *date_num_p = ecma_number_make_nan ();
+  ecma_number_t time = ECMA_NUMBER_ZERO;
+
+  /* Date Time String fromat (ECMA-262 v5, 15.9.1.15) */
+  ecma_string_t *date_str_p;
```
galpeter: This is variable only used inside the ECMA_TRY_CATCH block, we should move it inside.

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,258 @@
...
+    }
+    else if (date_str_size != (lit_utf8_size_t) (date_char_p - date_start_p)
+             && (*date_char_p == '+' || *date_char_p == '-'))
+    {
+      JERRY_ASSERT ((date_char_p - date_start_p) >= 0);
+      uint32_t num_of_visited_chars = (uint32_t) (date_char_p - date_start_p);
+
+      if ((date_str_size - num_of_visited_chars) < 6)
+      {
+        goto end; /* NaN */
```
ruben-ayrapetyan: Could you, please, remove usage of `goto`? Maybe, through extraction of a subroutine.

LaszloLango: Could you give me an example how the you mean this exactly? Not sure how can a helper eliminate this.

ruben-ayrapetyan: One of the ways:
- create helper with operations from lines 102-285. The helper would return `ecma_number_t` (Instead of `goto` it would contain something like `return nan`);
- call this helper and return ecma-value containing the returned number, as it is currently performed with `ecma_make_date_time` at line 286.

LaszloLango: That won't be the same behavior.

ruben-ayrapetyan: Provide more details, please. This should be fixed.

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,258 @@
...
+  {
+    goto end; /* NaN */
+  }
+
+  /* 2. read month if any */
+  if (date_str_size != (lit_utf8_size_t) (date_char_p - date_start_p)
+      && *date_char_p == '-')
+  {
+    /* eat up '-' */
+    date_char_p += sizeof (lit_utf8_byte_t);
```
ruben-ayrapetyan: Btw, is it correct way to iterate string? Can't the string contain a character, encoded with a multi-byte sequence?

LaszloLango: Ok

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
LaszloLango: @ruben-ayrapetyan, Ok. I'll eliminate all goto from my code, but make this thing clear. We have to avoid using of `goto` everywhere? I'm asking that because there several places in parser.cpp where `goto` statments are used, so this why I think that we can use them here too. If we eliminate `goto` statments, then we should eliminate it in global. What do you think?

ruben-ayrapetyan: @LaszloLango, yes, we should avoid using of `goto` everywhere.

Unfortunately, there too many style issues in `parser.cpp` to use it as a reference.
The style issues in parser, as in other engine parts, definitely, should be fixed.
Style issues in parser were fixed step by step, in the context of recent development of feature support, and, probably, style fixing would be continued in context of parser optimization. I think, if after the changes there still would be any styles issues, then the issues should be fixed with a separate style fix patch.

LaszloLango: @ruben-ayrapetyan, could you create an issue to eliminate `goto` from Jerry, so we don't forget this later? Maybe also static checks should be added.

ruben-ayrapetyan: @LaszloLango, it is great idea. I think all style issues should be documented in the way. Maybe, in one issue with check-list. I'll raise the issue.

LaszloLango:
@galpeter, @ruben-ayrapetyan I've updated the PR. Please check.

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,258 @@
...
+
+  if (!ecma_number_is_nan (year)
+      && year >= 0)
+  {
+    ecma_number_t month = ECMA_NUMBER_ONE;
+    ecma_number_t day = ECMA_NUMBER_ONE;
+    ecma_number_t time = ECMA_NUMBER_ZERO;
+
+    /* 2. read month if any */
+    if (!lit_utf8_iterator_is_eos (&iter)
```
galpeter: Do we really need to check the `eos` in the `if` cases? Isn't the `peek_next` enough?

LaszloLango: First line of peek_next is `JERRY_ASSERT (!lit_utf8_iterator_is_eos (iter_p));`, so the answer is yes. :)

galpeter: Well, if the  `peek_next` throws an assert if the string is over, then that's not a good thing.

galpeter:
lgtm

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,258 @@
  */
 
 /**
+ * Helper function to try to parse a part of a date string
+ *
+ * @return NaN if cannot read from string, ToNumber() otherwise
+ */
+static ecma_number_t
+ecma_date_parse_date_chars (lit_utf8_iterator_t *iter, /**< iterator of the utf8 string */
+                            uint32_t num_of_chars) /**< number of characters to read and convert */
```
zherczeg: This is kind of a maximum for me (max_length) ?

zherczeg: Sorry, forget this comment. So hard to read code without early returns.

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,258 @@
...
+    }
+
+    copy_size += lit_get_unicode_char_size_by_utf8_first_byte (*iter->buf_p);
+    lit_utf8_iterator_incr (iter);
+  }
+
+  if (copy_size > 0)
+  {
+    ecma_string_t *str_p = ecma_new_ecma_string_from_utf8 (str_start_p, copy_size);
+    ret_value = ecma_string_to_number (str_p);
```
zherczeg: What happens if the prefix is number? E.g. parsing 200- as year? You can read 4 characters, but only the first 3 are decimals. What about hex/octal numbers? '0x40' as year? Or exponential stuff, '20e5' as year?

zherczeg: '0015' is octal or not, '2e-3' etc.

LaszloLango: Those are invalid formats, the result will be NaN.

LaszloLango: Update: You are right, this won't work good on some cases. I'll do a fix for it.

LaszloLango:
I've updated the PR

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
LaszloLango: fixed

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,259 @@
...
+      break;
+    }
+
+    copy_size += lit_get_unicode_char_size_by_utf8_first_byte (*(iter->buf_p + iter->buf_pos.offset));
+    lit_utf8_iterator_incr (iter);
+  }
+
+  if (copy_size > 0)
+  {
+    ecma_string_t *str_p = ecma_new_ecma_string_from_utf8 (str_start_p, copy_size);
```
ruben-ayrapetyan: Why not to use `ecma_utf8_string_to_number` here?

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,259 @@
...
+  ecma_number_t ret_value = ecma_number_make_nan ();
+  lit_utf8_size_t copy_size = 0;
+  const lit_utf8_byte_t *str_start_p = iter->buf_p + iter->buf_pos.offset;
+
+  while (num_of_chars--)
+  {
+    if (lit_utf8_iterator_is_eos (iter)
+        || !lit_char_is_unicode_digit (lit_utf8_iterator_peek_next (iter)))
+    {
+      copy_size = 0;
```
ruben-ayrapetyan: For me, early return would be better here.

I think that we should avoid early returns in functions, where they reduce readability / modifiability (and, maybe, quality) and use them in functions, where they improve quality / readability / modifiability.

For example, currently, early returns should not be used in functions, working with completion values.

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
LaszloLango: I didn't know is there such a function. :) I'll fix it.

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -38,19 +41,259 @@
...
 /**
+ * Helper function to try to parse a part of a date string
+ *
+ * @return NaN if cannot read from string, ToNumber() otherwise
+ */
+static ecma_number_t
+ecma_date_parse_date_chars (lit_utf8_iterator_t *iter, /**< iterator of the utf8 string */
+                            uint32_t num_of_chars) /**< number of characters to read and convert */
+{
+  ecma_number_t ret_value = ecma_number_make_nan ();
```
ruben-ayrapetyan: Also, I would ask you to rename the variable from `ret_value` to something like `ret_num`. `ret_value` is, currently, mostly used for completion values, so can be confusing here.

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
ruben-ayrapetyan: Ok.

LaszloLango:
@ruben-ayrapetyan, I've updated the PR, please check.

ruben-ayrapetyan:
Looks good to me

PR #298, opened 2015-07-03
Fix Number.prototype.toFixed if argument is outside int32 range

dbatyai:


egavrin:
Good to me

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
```diff
@@ -251,6 +248,9 @@ ecma_builtin_number_prototype_object_to_fixed (ecma_value_t this_arg, /**< this
         int32_t num_digits = 0;
         int32_t exponent = 1;
 
+        /* 1. */
+        int32_t frac_digits = ecma_number_to_int32 (arg_num);
+
```
zherczeg: number_to_int32 returns 0 for -0.9? I don't know, just asking.

dbatyai: It does. I'll add a test case for this.

galpeter:
lgtm

PR #300, opened 2015-07-03
Optimize encode/decode URI for valid UTF-8 input.

zherczeg:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
@@ -25,6 +25,7 @@
 #include "ecma-try-catch-macro.h"
 #include "jrt.h"
 #include "lit-magic-strings.h"
+#include "lit-strings.h"
 #include "vm.h"
 #include "jrt-libc-includes.h"
 
@@ -511,7 +512,12 @@ static uint8_t unescaped_uri_component_set[16] =
   0xfe, 0xff, 0xff, 0x87, 0xfe, 0xff, 0xff, 0x47
 };
 
-#define ECMA_BUILTIN_HEX_TO_BYTE_ERROR 0x100
+/*
+ * Format is a percent sign followed by two hex digits.
+ */
+#define URI_ENCODED_BYTE_SIZE (3)
+
+#define ECMA_BUILTIN_HEX_TO_BYTE_ERROR (0x100)
 
 /**
  * Helper function to decode a hexadecimal byte from a string.
@@ -598,7 +604,11 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
 
   while (input_char_p < input_end_p)
   {
-    /* Input validation. */
+    /*
+     * We expect that the input is a valid UTF-8 sequence,
+     * so characters >= 0x80 can be let through.
+     */
+
     if (*input_char_p != '%')
     {
       output_size++;
@@ -613,9 +623,9 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
       break;
     }
 
-    input_char_p += 3;
+    input_char_p += URI_ENCODED_BYTE_SIZE;
 
-    if (decoded_byte <= 0x7f)
+    if (decoded_byte <= LIT_UTF8_1_BYTE_CODE_POINT_MAX)
     {
       /*
        * We don't decode those bytes, which are part of reserved_uri_bitset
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -631,7 +636,7 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
         output_size++;
       }
     }
-    else if (decoded_byte < 0xc0 || decoded_byte >= 0xf8)
+    else if (decoded_byte < LIT_UTF8_2_BYTE_MARKER || decoded_byte >= 0xf8)
```
egavrin: Missing `LIT_UTF8_4_BYTE_MASK`

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -743,38 +748,34 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
       else
       {
         uint32_t count;
-        uint32_t character;
 
         /* The validator already checked this before. */
-        JERRY_ASSERT (decoded_byte >= 0xc0 && decoded_byte < 0xf8);
+        JERRY_ASSERT (decoded_byte >= LIT_UTF8_2_BYTE_MARKER && decoded_byte < 0xf8);
```
egavrin: Missing `LIT_UTF8_4_BYTE_MASK`

egavrin:
BTW Why did you rolled back the change that moves defines to *.h file?

galpeter:
@egavrin, he did not rolled it back, just the master changed with the commit https://github.com/Samsung/jerryscript/commit/ae3eea8ae81ec840f95eb3cd9fca7ba6b7d70a7c which contains the movement of defines into the header, just like this PR and thus there is no diff for it.

egavrin:
@galpeter ah, I see - missed that.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
zherczeg: Technically it is the same value as LIT_UTF8_4_BYTE_MASK, but this has a very different meaning. It would be LIT_UTF8_5_BYTE_MARKER, but there is no such thing. I would rather not change it.

egavrin: I see.
Could you explain why not to add `LIT_UTF8_5_BYTE_MARKER` to keep common style?

sand1k: Maybe, we can add LIT_UTF8_5_BYTE_MARKER to be consistent and use constant definitions throughout the code?

egavrin: I think we may express through naming, that it shouldn't be used: `LIT_UTF8_5_BYTE_MARKER_DO_NOT_USE`

zherczeg: But we do use it. What about LIT_UTF8_FIRST_BYTE_MAX?

sand1k: Yes, good name.

egavrin: @sand1k explain why?

ruben-ayrapetyan: Maybe, it is better to add `LIT_UTF8_5_BYTE_MARKER` as a specification-defined value, and also `#define LIT_UTF8_FIRST_BYTE_MAX LIT_UTF8_5_BYTE_MARKER` with comment, saying that it is not a specification-defined value, but maximum possible value in the whole engine.

egavrin: :+1:

egavrin:
After fixing, `make push`

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -878,16 +877,21 @@ ecma_builtin_global_object_encode_uri_helper (ecma_value_t uri, /**< uri argumen
...
-    /* Input validation. */
-    lit_code_point_t character = lit_utf8_iterator_read_next (&iter);
+    /*
+     * We expect that the input is a valid UTF-8 sequence,
+     * so we only need to reject stray surrogate pairs.
+     */
 
-    if (character <= 0x7f)
+    /* Input validation. */
+    if (*input_char_p <= LIT_UTF8_1_BYTE_CODE_POINT_MAX)
```
ruben-ayrapetyan: Isn't the utf-8 encoding / decoding logic implemented in `lit`? If so, maybe, we can use the `lit`'s helpers in this code.

egavrin: @sand1k please check.

zherczeg: Exactly how? We only need to take care of a few cases in a very special way.

zherczeg:
Patch is updated. Please review.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -647,23 +652,23 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
...
+        character = decoded_byte & LIT_UTF8_LAST_4_BITS_MASK;
       }
       else
       {
         count = 3;
-        min = 0x1000;
-        character = decoded_byte & 0x07;
+        min = LIT_UTF8_4_BYTE_CODE_POINT_MIN;
+        character = decoded_byte & LIT_UTF8_LAST_3_BITS_MASK;
       }
```
sand1k: Lines 655-672 are similar to  lines 48-65 of lit-strings.cpp (`lit_is_utf8_string_valid` function), possibly we could extract them into function like
`lit_utf8_size_t lit_utf8_analyze_first_byte (lit_utf8_byte_t first_byte, lit_code_point_t *min, lit_code_point_t *character)`

sand1k: And also, possibly, we can change code, starting from line 755:
```
        if (decoded_byte < LIT_UTF8_2_BYTE_MASK)
        {
          count = 1;
        }
        else if (decoded_byte < LIT_UTF8_3_BYTE_MASK)
        {
          count = 2;
        }
        else
        {
          count = 3;
        }
```
with call to `lit_get_unicode_char_size_by_utf8_first_byte`.

sand1k:
As for the rest, LGTM.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
zherczeg: I can do that, but I don't see any advantage. We would need to add another switch for getting the min and character values (except if we store them in a static array). Effectively duplicating the code size.

zherczeg:
I choose a different approach. I convert everything to a byte sequence, and use the UTF8 validator to validate the result. I added a new patch for that, so if you don't like it, it is easy to revert.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -615,7 +620,7 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
 
     input_char_p += 3;
```
sand1k: Could you, please, define this constant:
`#define URI_ENCODED_BYTE_SIZE (3)`
and replace all occurences of `+= 3`.

sand1k:
Please, introduce `URI_ENCODED_BYTE_SIZE` constant and `make push`.

zherczeg:
Landed in 06b4490ea161cc1cbc268e6c6bdb77c2154ee91e

Issue #122, opened 2015-05-28
Assertion 'ret_value == ecma_make_empty_completion_value ()' failed in opfunc_native_call


renatahodovan:
###### Jerry version:

```
Checked revision: aaeec7f
Build: debug.linux
```
###### OS:

```
Ubuntu 14.04, x86_64
```
###### Test case:

``` javascript
print(this, []);
```
###### Backtrace:

``` javascript
ICE: Assertion 'ret_value == ecma_make_empty_completion_value ()' failed at jerryscript/jerry-core/vm/opcodes-native-call.cpp(opfunc_native_call):77.
Error: ERR_FAILED_INTERNAL_ASSERTION
[Inferior 1 (process 13969) exited with code 0170]
(gdb) bt
#0  opfunc_native_call (opdata=..., int_data=0x7fffffffd590) at jerryscript/jerry-core/vm/opcodes-native-call.cpp:75
#1  0x00000000004bfa60 in run_int_loop (int_data=0x7fffffffd590) at jerryscript/jerry-core/vm/vm.cpp:440
#2  0x00000000004b678d in run_int_from_pos (start_pos=0, this_binding_value=15, lex_env_p=0x7e9770 <_ZL13mem_heap_area.4015+48>, is_strict=false, 
    is_eval_code=false) at jerryscript/jerry-core/vm/vm.cpp:513
#3  0x00000000004bf82e in run_int () at jerryscript/jerry-core/vm/vm.cpp:381
#4  0x00000000004bd4d0 in jerry_run () at jerryscript/jerry-core/jerry.cpp:1243
#5  0x00000000004bbaff in main (argc=2, argv=0x7fffffffd938) at jerryscript/main-linux.cpp:248
```

PR #303, opened 2015-07-03
Set global object's prototype

galpeter:
The global object should have the same prototype
as a simple object.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.inc.h
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.inc.h
@@ -301,7 +301,7 @@ BUILTIN (ECMA_BUILTIN_ID_COMPACT_PROFILE_ERROR,
 /* The Global object (15.1) */
 BUILTIN (ECMA_BUILTIN_ID_GLOBAL,
          ECMA_OBJECT_TYPE_GENERAL,
-         ECMA_BUILTIN_ID__COUNT /* no prototype */,
+         ECMA_BUILTIN_ID_OBJECT_PROTOTYPE, /* Implementation-dependent */
          true,
          true,
          global)
--- a/tests/jerry/global.js
+++ b/tests/jerry/global.js
@@ -13,13 +13,5 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-try
-{
-  print({toString: function() { throw new TypeError("foo"); }}, []);
-  assert (false);
-}
-catch (e)
-{
-  assert (e instanceof TypeError);
-  assert (e.message === "foo");
-}
+
+assert (this.hasOwnProperty !== undefined);


galpeter:
Related test262 test: https://github.com/tc39/test262/blob/es5-tests/test/suite/ch15/15.2/15.2.3/15.2.3.14/15.2.3.14-6-6.js

jerry-core/ecma/builtin-objects/ecma-builtins.inc.h
```diff
@@ -301,7 +301,7 @@ BUILTIN (ECMA_BUILTIN_ID_COMPACT_PROFILE_ERROR,
 /* The Global object (15.1) */
 BUILTIN (ECMA_BUILTIN_ID_GLOBAL,
          ECMA_OBJECT_TYPE_GENERAL,
-         ECMA_BUILTIN_ID__COUNT /* no prototype */,
+         ECMA_BUILTIN_ID_OBJECT_PROTOTYPE,
```
ruben-ayrapetyan: According to ECMA-262 v5, the value is implementation-dependent.
Could you, please, add comment about this?

Are there some test cases that depend on the `this.hasOwnProperty` value? Why do we need the change?

tests/jerry/regression-test-issue-122.js
```diff
@@ -15,10 +15,11 @@
 
 try
 {
-  print(this, []);
+  print({toString: function() { throw new TypeError("foo"); }}, []);
```
ruben-ayrapetyan: Seems that the change is not related to the patch.

jerry-core/ecma/builtin-objects/ecma-builtins.inc.h
galpeter: Ok, I'll add a comment.

Please see the https://github.com/tc39/test262/blob/es5-tests/test/suite/ch15/15.2/15.2.3/15.2.3.14/15.2.3.14-6-6.js testcase's 17 line, where it calls the 'hasOwnProperty' on the global object.

ruben-ayrapetyan: Ok. Thank you.

tests/jerry/regression-test-issue-122.js
galpeter: It is related, a with this PR the global object can be printed out and won't cause an Error. Due to the fact that the prototype is now set, and thus the toString method will be found when converting the object to string.

ruben-ayrapetyan: I see.

Maybe, now we could use the initial version of regression test case from #122:
`print(this, []);`

ruben-ayrapetyan:
Looks good to me, after adding comment.

tests/jerry/regression-test-issue-122.js
galpeter: You mean replace the whole `regression-test-issue-122.js` file with the simple print? But then we will not have a test case which creates an error for the `print` call.

ruben-ayrapetyan: I see. Yes, it would be incorrect to replace with simple print.

The change looks good to me.

egavrin:
`make push`

LaszloLango:
+1 lgtm

PR #207, opened 2015-06-18
Implement printing in the specified radix for Number.prototype.toString().

dbatyai:
Depends on #100.

JerryScript-DCO-1.0-Signed-off-by: Dániel Bátyai dbatyai.u-szeged@partner.samsung.com

PR #306, opened 2015-07-06
Implement printing in the specified radix for Number.prototype.toString().

dbatyai:
I had to re-open this, because I can't update the other one.
Previous discussion can be found here: #207

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
@@ -24,7 +24,9 @@
 #include "ecma-objects.h"
 #include "ecma-string-object.h"
 #include "ecma-try-catch-macro.h"
+#include "fdlibm-math.h"
 #include "jrt.h"
+#include "jrt-libc-includes.h"
 
 #ifndef CONFIG_ECMA_COMPACT_PROFILE_DISABLE_NUMBER_BUILTIN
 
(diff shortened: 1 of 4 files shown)


dbatyai:
I've updated this to calculate an approximate buffer size. It can be slightly larger than necessary, but that should be negligible. I've added tests for two digit numbers as well.

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
```diff
@@ -118,14 +120,220 @@ ecma_builtin_number_prototype_object_to_string (ecma_value_t this_arg, /**< this
...
+          this_arg_number /= (ecma_number_t) radix;
+        }
+        else
+        {
+          this_arg_number *= (ecma_number_t) radix;
+        }
+      }
+
+      uint64_t whole = (uint64_t) this_arg_number;
+      double fraction = this_arg_number - (double) whole;
```
ruben-ayrapetyan: Does usage of `double` instead of `ecma_number_t` improve precision in fp32 mode in the case?

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
```diff
@@ -118,14 +120,220 @@ ecma_builtin_number_prototype_object_to_string (ecma_value_t this_arg, /**< this
...
+      for (int i = 0; i < buff_index / 2; i++)
+      {
+        lit_utf8_byte_t swap = buff[i];
+        buff[i] = buff[buff_index - i - 1];
+        buff[buff_index - i - 1] = swap;
+      }
+
+      bool should_round = false;
+      /* Calculate digits for fractional part. */
+      for (int iter_count = 0; iter_count < 53 && (fraction != 0 || is_scale_negative); iter_count++)
```
ruben-ayrapetyan: Is it correct that `53` here is connected with width of mantissa?

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
dbatyai: This was unintentional. But since you asked, I don't think this improves precision. The original value is already constrained to a float, and because of the continued multiplication the required precision to represent the result would only decrease.

ruben-ayrapetyan: Ok, I see. The question was about additional precision lose, that may be can occur in intermediate operations.

I think we should use `ecma_number_t` in the case.

ruben-ayrapetyan: Maybe, we should move the following defines from `ecma-helpers-number.cpp` to `ecma-globals.h`, near definition of `ecma_number_t` and use the definitions in the function:
```
#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
JERRY_STATIC_ASSERT (sizeof (ecma_number_t) == sizeof (uint32_t));

/**
 * Width of sign field
 *
 * See also:
 *          IEEE-754 2008, 3.6, Table 3.5
 */
#define ECMA_NUMBER_SIGN_WIDTH       (1)

/**
 * Width of biased exponent field
 *
 * See also:
 *          IEEE-754 2008, 3.6, Table 3.5
 */
#define ECMA_NUMBER_BIASED_EXP_WIDTH (8)

/**
 * Width of fraction field
 *
 * See also:
 *          IEEE-754 2008, 3.6, Table 3.5
 */
#define ECMA_NUMBER_FRACTION_WIDTH   (23)
#elif CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
JERRY_STATIC_ASSERT (sizeof (ecma_number_t) == sizeof (uint64_t));

/**
 * Width of sign field
 *
 * See also:
 *          IEEE-754 2008, 3.6, Table 3.5
 */
#define ECMA_NUMBER_SIGN_WIDTH       (1)

/**
 * Width of biased exponent field
 *
 * See also:
 *          IEEE-754 2008, 3.6, Table 3.5
 */
#define ECMA_NUMBER_BIASED_EXP_WIDTH (11)

/**
 * Width of fraction field
 *
 * See also:
 *          IEEE-754 2008, 3.6, Table 3.5
 */
#define ECMA_NUMBER_FRACTION_WIDTH   (52)
#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */
```

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
dbatyai: Yes

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
dbatyai: Done

ruben-ayrapetyan: Great! Thank you.

ruben-ayrapetyan:
Looks good to me

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.cpp
```diff
@@ -118,14 +120,220 @@ ecma_builtin_number_prototype_object_to_string (ecma_value_t this_arg, /**< this
...
+           * one extra digit that we round off just to make sure.
+           */
+          should_round = true;
+          break;
+        }
+      }
+
+      if (should_round)
+      {
+        /* Round off last digit. */
```
LaszloLango: typo: of

dbatyai: No, it's not :)

LaszloLango: ok, right :)

LaszloLango:
+1 lgtm

egavrin:
`make push`

PR #319, opened 2015-07-06
Empty values in array creation should be not set

galpeter:
When creating arrays it is possible to set 'empty' values
which are not undefined, but simply unused values.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -79,6 +79,7 @@ typedef enum
   ECMA_SIMPLE_VALUE_NULL, /**< null value */
   ECMA_SIMPLE_VALUE_FALSE, /**< boolean false */
   ECMA_SIMPLE_VALUE_TRUE, /**< boolean true */
+  ECMA_SIMPLE_VALUE_UNUSED, /**< implemenetation defined value to distinguish not initialized values */
   ECMA_SIMPLE_VALUE_ARRAY_REDIRECT, /**< implementation defined value for an array's elements that exist,
                                          but are stored directly in the array's property list
                                          (used for array elements with non-default attribute values) */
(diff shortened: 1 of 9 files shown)


galpeter:
One related test262 testcase: https://github.com/tc39/test262/blob/es5-tests/test/suite/ch15/15.2/15.2.3/15.2.3.7/15.2.3.7-6-a-155.js

zherczeg:
+1 lgtm

ruben-ayrapetyan:
I think, this should be fixed by modification of byte-code / opcode handlers.

ruben-ayrapetyan:
Is there any issue opened for this?

galpeter:
@ruben-ayrapetyan , there is no issue opened for this. I'll create one. 

I've got one question regarding the byte-code/opcode way: in the `ecma_op_create_array_object` method how would the 'unset' arguments appear? (Currently) In that function we need to know which indexes are set, and which are not.

galpeter:
Created issue: #356

ruben-ayrapetyan:
@galpeter, thank you!

ECMA-262 v5, 11.1.4 defines that array object should be created empty, and then filled, step by step. So, we don't need to pass 'unset' arguments to `ecma_op_create_array_object`.

galpeter:
So that means that we the `ecma_op_create_array_object` will be called with empty argument list and then the elements will be added one by one (via define prop or something like that). If so then we can close this PR.

egavrin:
@galpeter @ruben-ayrapetyan let's close this PR is everyone is OK to continue with #356.

ruben-ayrapetyan:
@galpeter, @egavrin, I agree that the pull request can be closed, as the issue is already described in #356 with reference to the pull request.

egavrin:
Ok.

PR #335, opened 2015-07-08
Array.prototype.sort() should sort in place.

dbatyai:
Array.prototype.sort() should sort in place instead of creating a new Array with the sorted elements.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.cpp
+++ b/jerry-core/ecma/base/ecma-helpers-string.cpp
@@ -993,6 +993,39 @@ ecma_string_to_number (const ecma_string_t *str_p) /**< ecma-string */
 } /* ecma_string_to_number */
 
 /**
+ * Check if string is array index.
+ *
+ * @return true - if string is valid array index
+ *         false - otherwise
+ */
+bool
+ecma_string_get_array_index (const ecma_string_t *str_p, /**< ecma-string */
+                             uint32_t *out_index_p) /**< out: index */
+{
+  bool is_array_index = true;
+  if (str_p->container == ECMA_STRING_CONTAINER_UINT32_IN_DESC)
+  {
+    *out_index_p = str_p->u.uint32_number;
+  }
+  else
+  {
+    ecma_number_t num = ecma_string_to_number (str_p);
+    *out_index_p = ecma_number_to_uint32 (num);
+
+    ecma_string_t *to_uint32_to_string_p = ecma_new_ecma_string_from_uint32 (*out_index_p);
+
+    is_array_index = ecma_compare_ecma_strings (str_p,
+                                                to_uint32_to_string_p);
+
+    ecma_deref_ecma_string (to_uint32_to_string_p);
+  }
+
+  is_array_index = is_array_index && (*out_index_p != ECMA_MAX_VALUE_OF_VALID_ARRAY_INDEX);
+
+  return is_array_index;
+} /* ecma_string_is_array_index */
+
+/**
  * Convert ecma-string's contents to a utf-8 string and put it to the buffer.
  *
  * @return number of bytes, actually copied to the buffer - if string's content was copied successfully;
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
@@ -1449,36 +1449,75 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
   uint32_t len = ecma_number_to_uint32 (len_number);
 
   uint32_t defined_prop_count = 0;
-  /* Count number of defined properties. */
+  /* Count number of array index properties. */
   for (ecma_property_t *property_p = ecma_get_property_list (obj_p);
        property_p != NULL;
        property_p = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p))
   {
-    defined_prop_count++;
+    ecma_string_t *property_name_p;
+
+    if (property_p->type == ECMA_PROPERTY_NAMEDDATA)
+    {
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t,
+                                                   property_p->u.named_data_property.name_p);
+    }
+    else if (property_p->type == ECMA_PROPERTY_NAMEDACCESSOR)
+    {
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t,
+                                                   property_p->u.named_accessor_property.name_p);
+    }
+    else
+    {
+      continue;
+    }
+
+    uint32_t index;
+    if (ecma_string_get_array_index (property_name_p, &index) && index < len)
+    {
+      defined_prop_count++;
+    }
   }
 
   MEM_DEFINE_LOCAL_ARRAY (values_buffer, defined_prop_count, ecma_value_t);
   uint32_t copied_num = 0;
 
   /* Copy unsorted array into a native c array. */
-  for (uint32_t index = 0; index < len && ecma_is_completion_value_empty (ret_value); index++)
+  for (ecma_property_t *property_p = ecma_get_property_list (obj_p);
+       property_p != NULL && ecma_is_completion_value_empty (ret_value);
+       property_p = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p))
   {
-    ecma_string_t *index_string_p = ecma_new_ecma_string_from_uint32 (index);
-    if (ecma_op_object_get_property (obj_p, index_string_p) != NULL)
+    ecma_string_t *property_name_p;
+
+    if (property_p->type == ECMA_PROPERTY_NAMEDDATA)
     {
-      ECMA_TRY_CATCH (index_value, ecma_op_object_get (obj_p, index_string_p), ret_value);
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t,
+                                                   property_p->u.named_data_property.name_p);
+    }
+    else if (property_p->type == ECMA_PROPERTY_NAMEDACCESSOR)
+    {
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t,
+                                                   property_p->u.named_accessor_property.name_p);
+    }
+    else
+    {
+      continue;
+    }
+
+    uint32_t index;
+    if (ecma_string_get_array_index (property_name_p, &index) && index < len)
+    {
+      ECMA_TRY_CATCH (index_value, ecma_op_object_get (obj_p, property_name_p), ret_value);
 
       values_buffer[copied_num++] = ecma_copy_value (index_value, true);
 
       ECMA_FINALIZE (index_value);
     }
-    ecma_deref_ecma_string (index_string_p);
   }
 
-  JERRY_ASSERT (copied_num <= defined_prop_count);
+  JERRY_ASSERT (copied_num == defined_prop_count || !ecma_is_completion_value_empty (ret_value));
 
   /* Sorting. */
-  if (len > 1 && ecma_is_completion_value_empty (ret_value))
+  if (copied_num > 1 && ecma_is_completion_value_empty (ret_value))
   {
     ECMA_TRY_CATCH (sort_value,
                     ecma_builtin_array_prototype_object_array_heap_sort_helper (values_buffer,
@@ -1489,8 +1528,7 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
   }
 
   /* Put sorted values to the front of the array. */
-  uint32_t index;
-  for (index = 0; index < copied_num && ecma_is_completion_value_empty (ret_value); index++)
+  for (uint32_t index = 0; index < copied_num && ecma_is_completion_value_empty (ret_value); index++)
   {
     ecma_string_t *index_string_p = ecma_new_ecma_string_from_uint32 (index);
     ECMA_TRY_CATCH (put_value,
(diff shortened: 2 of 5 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1477,15 +1477,25 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
...
-     * 16 bits long, with an array of that size, we would run out of memory way before this happens.
-     */
-    JERRY_ASSERT ((ecma_length_t) len == len);
-    /* Copy the sorted array into a new array. */
-    ret_value = ecma_op_create_array_object (values_buffer, (ecma_length_t) len, false);
+    ecma_string_t *index_string_p = ecma_new_ecma_string_from_uint32 (index);
+
+    if (ecma_is_value_undefined (values_buffer[index]))
+    {
+      ECMA_TRY_CATCH (del_value, ecma_op_object_delete (obj_p, index_string_p, true), ret_value);
```
egavrin: What is the reason to have empty TRY/FINALIZE block?

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1477,15 +1477,25 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
...
+      ECMA_FINALIZE (del_value);
+    }
+    else
+    {
+      ECMA_TRY_CATCH (put_value,
```
egavrin: Likewise

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
dbatyai: In my opinion this look cleaner than using if/else to check/free.

ruben-ayrapetyan: Also, this is current common style. Maybe, this could be encoded in a more convenient way. I think, in the case it is better to change all the places at once.

egavrin: I see

egavrin:
No other questions. Good to me.

tests/jerry/array-prototype-sort.js
```diff
@@ -43,9 +43,20 @@ function f(arg1, arg2) {
...
-var sorted = array.sort(f);
+array.sort(f);
 for (i = 0; i < array.length; i++) {
-	assert(sorted[sorted.length - i - 1] === i);
+	assert(array[array.length - i - 1] === i);
+}
+
+// Sorting sparse array
+var array = [1,,2,,3,,4,undefined,5];
+var expected = [1,2,3,4,5,,,,];
```
ruben-ayrapetyan: Seems that `expected.length` is 8, and `array.length` is 9. Maybe, it is incorrect.

ruben-ayrapetyan: Could you, please, check whether conditions in the following loop should be `true`?

```
for (i = 0; i < array.length; i++)
{
  assert (expected.hasOwnProperty (i) == array.hasOwnProperty (i));
}
```

Currently, some of them are not `true`.

dbatyai: `expected` has 5 numbers, and then 4 empty places after those, that should be 9. Nevertheless, there's an assert for this in line `57` below.

ruben-ayrapetyan: The following test case print `8`:
```
var expected = [1,2,3,4,5,,,,];

print (expected.length);
```

By the way, assertion at line 57 is incorrect, as it performs assignment, instead of comparison.

dbatyai: The issue with `hasOwnProperty` is related with #319, in particular, the problem is that we currently can't distinguish between "empty" values that are unset, and explicitly defined `undefined` values, currently both are `undefined`. If we do it by the standard, then empty values should be deleted, but currently that also deletes the set `undefined` value.

dbatyai: My mistake, let me re-check.

ruben-ayrapetyan: To distinguish between undefined properties and properties with undefined values, we can use `ecma_op_object_get_property (...) != NULL`.

dbatyai: Yes, initially. But we copy values into a local array, so that we can sort them. When we do this both will become undefined, since we don't have a way to represent them differently.

ruben-ayrapetyan: I think, this is because values are copied to simple array of values without additional information.

It seems that the way also can lead to out-of-memory in case of sparse arrays. For example:
```
a = new Array ();
a[100000] = 1;
a[1000000] = 2;
a.sort ();
```

Is it correct?

dbatyai:
@ruben-ayrapetyan , I've updated the PR. We no longer copy undefined properties.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1448,44 +1448,65 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
...
+
+  MEM_DEFINE_LOCAL_ARRAY (values_buffer, defined_prop_count, ecma_value_t);
   uint32_t copied_num = 0;
 
   /* Copy unsorted array into a native c array. */
   for (uint32_t index = 0; index < len && ecma_is_completion_value_empty (ret_value); index++)
   {
     ecma_string_t *index_string_p = ecma_new_ecma_string_from_uint32 (index);
-    ECMA_TRY_CATCH (index_value, ecma_op_object_get (obj_p, index_string_p), ret_value);
+    if (ecma_op_object_get_property (obj_p, index_string_p) != NULL)
```
ruben-ayrapetyan: The test would be executed very long, because of iteration over all values in `[0.. len)`:


```
var a = new Array ();

a[100000] = 1;
a[1000000] = 2;

a.sort ();
```

There are also another places with the issue. For example, `ecma_op_array_object_define_own_property`.
Could you, please, raise an issue about this, with references to known places with the issue, and add `FIXME` to the code part, referencing the issue?

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1448,44 +1448,65 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
...
+    ecma_string_t *index_string_p = ecma_new_ecma_string_from_uint32 (index);
+    ECMA_TRY_CATCH (put_value,
+                    ecma_op_object_put (obj_p, index_string_p, values_buffer[index], true),
+                    ret_value);
+    ECMA_FINALIZE (put_value);
+    ecma_deref_ecma_string (index_string_p);
+  }
+
+  /* Undefined properties should be in the back of the array. */
+  for (;index < len && ecma_is_completion_value_empty (ret_value); index++)
```
ruben-ayrapetyan: Please, add `FIXME` with reference to the issue for the code part too.

ruben-ayrapetyan:
After raising the issue and adding `FIXME`s, the changes would look good to me.

ruben-ayrapetyan:
@dbatyai, could you, please, rebase to master?

dbatyai:
@ruben-ayrapetyan, rebased.

ruben-ayrapetyan:
@dbatyai, thank you!

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -962,6 +962,9 @@ ecma_builtin_array_prototype_object_index_of (ecma_value_t this_arg, /**< this a
     {
       JERRY_ASSERT (from_idx < len);
 
+        /* FIXME: Iterating over the whole lenght is unnecessary in case of sparse arrays,
+         *        and could take a long time if len is large. We should only iterate
+         *        over defined array indexes. */
```
zherczeg: I think there is a TODO () or similar macro somewhere.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1117,6 +1120,9 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
      * for an underflow instead. This is safe, because from_idx will always start in [0, len - 1],
      * and len is in [0, UINT_MAX], so from_idx >= len means we've had an underflow, and should stop.
      */
+    /* FIXME: Iterating over the whole lenght is unnecessary in case of sparse arrays,
```
ruben-ayrapetyan: Typo: "lenght" -> "length".

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1117,6 +1120,9 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
+     *        and could take a long time if len is large. We should only iterate
+     *        over defined array indexes. */
```
ruben-ayrapetyan: @dbatyai, could you, please, add reference to the raise issue?

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1448,44 +1454,71 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
...
+    defined_prop_count++;
+  }
+
+  MEM_DEFINE_LOCAL_ARRAY (values_buffer, defined_prop_count, ecma_value_t);
   uint32_t copied_num = 0;
 
   /* Copy unsorted array into a native c array. */
+  /* FIXME: Iterating over [0,len) is unnecessary in case of sparse arrays,
+   *        and could take a long time if len is large. We should only iterate
+   *        over defined array indexes. */
```
ruben-ayrapetyan: @dbatyai, could you, please, add reference to the issue in the comment too?

zherczeg:
I agree that the real index based sorting should be replaced by a virtual index based sorting. What is the sorting algorithm? Quick-sort?

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
ruben-ayrapetyan: Could you, please, also move the changes, adding `FIXME` to separate commit in the pull request?

ruben-ayrapetyan:
After fixing comments and moving them to separate commit, looks good to me.

dbatyai:
@zherczeg, current algorithm is heapsort.

dbatyai:
@ruben-ayrapetyan, please re-check
After talking with @zherczeg we decided that we should fix #362 in this patch, instead of adding FIXMEs.
After some investigation it appears that `sort` is actually the only candidate for #362, because other functions are required to also check the prototype chain, so iterating over their defined properties is not enough.

ruben-ayrapetyan:
@dbatyai, I'll check the changes today.

I think that fix of #362 in several other parts of engine (including, `ecma_op_array_object_define_own_property`) can significantly improve performance of sparse arrays handling. However, maybe, `sort` can be optimized even more, by skipping prototype chain.

dbatyai:
@ruben-ayrapetyan, @zherczeg, I've updated the PR.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -1448,44 +1448,128 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
...
+    ecma_property_t *next_prop = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p);
+
+    uint32_t index;
+    if (ecma_string_get_array_index (property_name_p, &index) && index < len && index >= copied_num)
+    {
+      ECMA_TRY_CATCH (del_value, ecma_op_object_delete (obj_p, property_name_p, true), ret_value);
+      ECMA_FINALIZE (del_value);
+    }
+
+    property_p = next_prop;
```
ruben-ayrapetyan: The code infitely loops for the following example:
```
Array.prototype.sort();
```

Reason is that `property_p` is not updated for internal properties.

It's better to put `property_p = next_prop` into `for`'s increment section.

jerry-core/ecma/base/ecma-helpers-string.cpp
```diff
@@ -993,6 +993,36 @@ ecma_string_to_number (const ecma_string_t *str_p) /**< ecma-string */
 } /* ecma_string_to_number */
 
 /**
+ * Check if string is array index.
+ */
```
ruben-ayrapetyan: `@return` sections is missing.

jerry-core/ecma/base/ecma-helpers-string.cpp
```diff
@@ -993,6 +993,36 @@ ecma_string_to_number (const ecma_string_t *str_p) /**< ecma-string */
...
+bool
+ecma_string_get_array_index (const ecma_string_t *str_p, /**< ecma-string */
+                             uint32_t *index) /**< index */
```
ruben-ayrapetyan: Please, add `out: ` mark to the comment, and rename the parameter to `out_index_p`.

ruben-ayrapetyan:
After fixing, the changes would look good to me.

ruben-ayrapetyan:
Looks good to me

zherczeg:
LGTM

egavrin:
I'll merge.

egavrin:
Merged a2c6663d43321f17a1ff10f04a59526a52629ecb

PR #336, opened 2015-07-08
Implement Date.prototype.toString

LaszloLango:


jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
  */
 
 /**
+ * Insert leading zeros to a string of a number if needed.
+ */
+static void
+date_insert_leading_zeros (ecma_string_t **str_p, /**< input/output string */
+                      ecma_number_t num, /**< input number */
```
egavrin: Broken formatting

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
...
+
+} /* insert_leading_zeros */
+
+/**
+ * Insert a number to the start of a string with a specific separator character and
+ * fix length.
+ */
+static void
+date_insert_num_with_sep (ecma_string_t **str_p, /**< input/output string */
+                     ecma_number_t num, /**< input number */
```
egavrin: Broken formatting

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -49,7 +102,62 @@
...
+      ecma_number_t minutes = ecma_date_min_from_time (*prim_value_num_p);
+      date_insert_num_with_sep (&output_str_p, minutes, LIT_MAGIC_STRING_COLON_CHAR, 10);
+
+      ecma_number_t hours = ecma_date_hour_from_time (*prim_value_num_p);
+      date_insert_num_with_sep (&output_str_p, hours, LIT_MAGIC_STRING_COLON_CHAR, 10);
+
+      ecma_number_t day = ecma_date_date_from_time (*prim_value_num_p);
+      date_insert_num_with_sep (&output_str_p, day, LIT_MAGIC_STRING_TIME_SEP_U, 10);
+
+      ecma_number_t month = ecma_date_month_from_time (*prim_value_num_p) + 1;
```
egavrin: Why we have `+1` here?

LaszloLango: `ecma_date_month_from_time` returns a number from 0 to 11, but we have to print the month from 1 to 12. (ISO 8601, http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.15)

egavrin: Please add comment for that

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
ruben-ayrapetyan: Also, prefix should be `ecma_date_`.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
  */
 
 /**
+ * Insert leading zeros to a string of a number if needed.
+ */
+static void
+ecma_date_insert_leading_zeros (ecma_string_t **str_p, /**< input/output string */
```
egavrin: @ruben-ayrapetyan  why `ecma_date_` should it be `ecma_builtin_date_`?

LaszloLango: I think `ecma_builtin_date_` will be too long for a prefix and we used `ecma_date_` for helpers in `ecma-builtin-date.cpp`

egavrin: I'm asking because neighbour function is named with `ecma_builtin_date_` : https://github.com/Samsung/jerryscript/pull/336/files#diff-4c3f6b953c0f51c11d2a079e9dcc10f8R103

LaszloLango: @egavrin, I know, but those two functions are not builtins, only helpers.

ruben-ayrapetyan: @egavrin, I think that you are right. Following current naming style, the prefix should be even `ecma_builtin_date_prototype_helper_`. But this prefix is too long. So, probably, naming scheme for built-in helpers should be somehow updated. I'm not sure how exactly. But at least, `ecma_date_`, of course, should be added.

egavrin: Let's leave as is ^_^

egavrin:
Good to em

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
...
+    div /= 10;
+
+    ecma_string_t *zero_str_p = ecma_new_ecma_string_from_number (ECMA_NUMBER_ZERO);
+    ecma_string_t *concat_p = ecma_concat_ecma_strings (zero_str_p, *str_p);
+    ecma_deref_ecma_string (zero_str_p);
+    ecma_deref_ecma_string (*str_p);
+    *str_p = concat_p;
+  }
+
+} /* insert_leading_zeros */
```
ruben-ayrapetyan: Incorrect comment

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -49,7 +102,67 @@
...
+
+    ecma_object_t *obj_p = ecma_get_object_from_value (obj_this);
+    ecma_property_t *prim_value_prop_p;
+    prim_value_prop_p = ecma_get_internal_property (obj_p, ECMA_INTERNAL_PROPERTY_PRIMITIVE_NUMBER_VALUE);
+    ecma_number_t *prim_value_num_p = ECMA_GET_NON_NULL_POINTER (ecma_number_t,
+                                                                 prim_value_prop_p->u.internal_property.value);
+
+    if (ecma_number_is_nan (*prim_value_num_p))
+    {
+      const lit_utf8_byte_t *inv_date_utf8_p = (const lit_utf8_byte_t *) "Invalid Date";
```
ruben-ayrapetyan: We should create a magic string for the value.

kkristof:
lgtm

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
...
+                                ecma_number_t num, /**< input number */
+                                int32_t div) /**< "length" {10, 100, 1000} */
+{
+  JERRY_ASSERT (div == 10 || div == 100 || div == 1000);
+
+  while (num < div && div > 1)
+  {
+    div /= 10;
+
+    ecma_string_t *zero_str_p = ecma_new_ecma_string_from_number (ECMA_NUMBER_ZERO);
```
ruben-ayrapetyan: ecma_new_ecma_string_from_uint32 would be better in the case, as no additional pools chunk would be allocated for floating point value.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
...
+    ecma_deref_ecma_string (zero_str_p);
+    ecma_deref_ecma_string (*str_p);
+    *str_p = concat_p;
+  }
+
+} /* ecma_date_insert_leading_zeros */
+
+/**
+ * Insert a number to the start of a string with a specific separator character and
+ * fix length.
```
ruben-ayrapetyan: The comment doesn't say anything about leading zeroes.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
  */
 
 /**
+ * Insert leading zeros to a string of a number if needed.
+ */
+static void
+ecma_date_insert_leading_zeros (ecma_string_t **str_p, /**< input/output string */
+                                ecma_number_t num, /**< input number */
+                                int32_t div) /**< "length" {10, 100, 1000} */
```
ruben-ayrapetyan: Maybe, `leading_zeroes_num` would be better. In the case value of the argument should actually be number of leading zeroes to insert.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -38,6 +42,55 @@
...
+
+/**
+ * Insert a number to the start of a string with a specific separator character and
+ * fix length.
+ */
+static void
+ecma_date_insert_num_with_sep (ecma_string_t **str_p, /**< input/output string */
+                               ecma_number_t num, /**< input number */
+                               lit_magic_string_id_t magic_str_id, /**< separator character id */
+                               int32_t div) /**< "length" {10, 100, 1000} */
```
ruben-ayrapetyan: Maybe, `leading_zeroes_num` is better.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
LaszloLango: It does. "fix length" means this.

ruben-ayrapetyan: "Fix length" doesn't imply "add leading zeroes".

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -42,6 +43,53 @@
...
+ */
+static void
+ecma_date_insert_leading_zeros (ecma_string_t **str_p, /**< input/output string */
+                                ecma_number_t num, /**< input number */
+                                uint32_t length) /**< length of string of number */
+{
+  JERRY_ASSERT (length >= 1);
+
+  /* If the length is bigger than the number of digits in num, then insert leding zeros. */
+  for (uint32_t i = length - 1; i > 0 && num < pow (10,i); i--)
```
ruben-ayrapetyan: Why do we need to perform `pow` on every iteration? It is unnecessarily resource consuming.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -42,6 +43,53 @@
...
+  ecma_string_t *concat_p = ecma_concat_ecma_strings (magic_string_p, *str_p);
+  ecma_deref_ecma_string (magic_string_p);
+  ecma_deref_ecma_string (*str_p);
+  *str_p = concat_p;
+
+  ecma_string_t *num_str_p = ecma_new_ecma_string_from_number (num);
+  concat_p = ecma_concat_ecma_strings (num_str_p, *str_p);
+  ecma_deref_ecma_string (num_str_p);
+  ecma_deref_ecma_string (*str_p);
+  *str_p = concat_p;
```
ruben-ayrapetyan: Why not to use `return` for this?

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
LaszloLango: I could argue with that, but I added more more a more descriptive comment for this function. :)

ruben-ayrapetyan: Current comment is good.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
LaszloLango: What do you suggest instead?

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
LaszloLango: In this  case we don't need a temporary variable on the call site (like using `ecma_concat_ecma_strings`). I think it is easier to use and more readable. Should I change this to `return`?

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
ruben-ayrapetyan:
```cpp
 uint32_t first_index = length - 1u;
 ecma_number_t power_i = (ecma_number_t) pow (10, first_index);
 for (uint32_t i = first_index; i > 0 && num < power_i; i--, power_i /= 10)
 {
   ecma_string_t *zero_str_p = ecma_new_ecma_string_from_uint32 (0);
   ecma_string_t *concat_p = ecma_concat_ecma_strings (zero_str_p, *str_p);
   ecma_deref_ecma_string (zero_str_p);
   ecma_deref_ecma_string (*str_p);
   *str_p = concat_p;
 }
```

LaszloLango: Thanks, I'll do it. Is it good for you if I add this change to #360?

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
ruben-ayrapetyan: Actually, we can perform the call without temporary variable on the call site:
```cpp
output_str_p = ecma_date_insert_num_with_sep (output_str_p,
                                              seconds,
                                              LIT_MAGIC_STRING_DOT_CHAR,
                                              2);
```
Difference in the case is that change of `output_str_p` value becomes evident, so readability is improved.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
ruben-ayrapetyan: It is better to open separate pull requests for unrelated changes.
If the change involves code, updated in the #360, it would be ok, if you would place the change into a separate commit of the pull request.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
LaszloLango: @ruben-ayrapetyan, I see. So we can dereference the string in the function and return the you new pointer. Do you mean like this:

```C++
static emca_string_t *str_p
ecma_date_insert_num_with_sep (ecma_string_t *str_p, /**< input string */
                               ecma_number_t num, /**< input number */
                               lit_magic_string_id_t magic_str_id, /**< separator character id */
                               uint32_t length) /**< length of string of number */
{
  ecma_string_t *magic_string_p = ecma_get_magic_string (magic_str_id);

  ecma_string_t *concat_p = ecma_concat_ecma_strings (magic_string_p, str_p);
  ecma_deref_ecma_string (magic_string_p);
  ecma_deref_ecma_string (str_p);
  str_p = concat_p;

  ecma_string_t *num_str_p = ecma_new_ecma_string_from_number (num);
  concat_p = ecma_concat_ecma_strings (num_str_p, str_p);
  ecma_deref_ecma_string (num_str_p);
  ecma_deref_ecma_string (str_p);
  return concat_p;
}
```
May we deref the parameter in this function? Isn't it dangerous?

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
LaszloLango: It is also using these two helper functions.

ruben-ayrapetyan: Ok.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
ruben-ayrapetyan: @LaszloLango, could you, please, clarify what dereferencing do you mean?
How is it different from dereferencing in `ecma_date_insert_num_with_sep (ecma_string_t **str_p,` version?

LaszloLango: @ruben-ayrapetyan, what makes me unsure is that if we want to use it like this:

```C++
output_str_p = ecma_date_insert_num_with_sep (output_str_p,
                                              seconds,
                                              LIT_MAGIC_STRING_DOT_CHAR,
                                              2);
```
then `ecma_date_insert_num_with_sep` must deref `output_str_p` because it returns with a new string pointer and it will cause a memory leak. If we use this function wrong like this:

```C++
ecma_date_insert_num_with_sep (output_str_p,
                                              seconds,
                                              LIT_MAGIC_STRING_DOT_CHAR,
                                              2);
```

then `output_str_p` will be freed and cannot be used after the call. Maybe I misunderstood you. Is this you are talking about?

ruben-ayrapetyan: @LaszloLango, I see. I think that the approach that you've described is good for now, until the hidden dereference issue wouldn't be solved somehow (maybe, after switching to C++ in the whole engine, it would be solved in a simple way).
The code is ok for me.

PR #337, opened 2015-07-08
Fix use strict detection

sand1k:
The patch adds the following features:
- Recognize Use Strict directive in the middle of directive prologue.
- Do not recognize Use Strict if it contains escape sequences
(see 14.1 Directive Prologues and the Use Strict Directive)

Fixes the following tests form Test262:
  ch14/14.1/14.1-14-s.js
  ch14/14.1/14.1-4-s.js
  ch14/14.1/14.1-5-s.js
  ch14/14.1/14.1-8-s.js

### DIFF
--- a/jerry-core/parser/js/lexer.cpp
+++ b/jerry-core/parser/js/lexer.cpp
@@ -1762,6 +1762,42 @@ lexer_are_tokens_with_same_identifier (token id1, /**< identifier token (TOK_NAM
 } /* lexer_are_tokens_with_same_identifier */
 
 /**
+ * Checks that TOK_STRING doesn't contain EscapeSequence or LineContinuation
+ *
+ * @return true, if token's string in source buffer doesn't contain backslash
+ *         false, otherwise
+ */
+bool
+lexer_is_no_escape_sequences_in_token_string (token tok) /**< token of type TOK_STRING */
+{
+  JERRY_ASSERT (tok.type == TOK_STRING);
+
+  lit_utf8_iterator_t iter = src_iter;
+  lit_utf8_iterator_seek (&iter, tok.loc);
+
+  JERRY_ASSERT (!lit_utf8_iterator_is_eos (&iter));
+  ecma_char_t c = lit_utf8_iterator_read_next (&iter);
+  JERRY_ASSERT (c == LIT_CHAR_SINGLE_QUOTE
+                || c == LIT_CHAR_DOUBLE_QUOTE);
+
+  const ecma_char_t end_char = c;
+
+  do
+  {
+    JERRY_ASSERT (!lit_utf8_iterator_is_eos (&iter));
+    c = lit_utf8_iterator_read_next (&iter);
+
+    if (c == LIT_CHAR_BACKSLASH)
+    {
+      return false;
+    }
+  }
+  while (c != end_char);
+
+  return true;
+} /* lexer_is_no_escape_sequences_in_token_string */
+
+/**
  * Initialize lexer to start parsing of a new source
  */
 void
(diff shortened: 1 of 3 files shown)


ruben-ayrapetyan:
Looks good to me

jerry-core/parser/js/lexer.cpp
```diff
@@ -1762,6 +1762,42 @@ lexer_are_tokens_with_same_identifier (token id1, /**< identifier token (TOK_NAM
...
+  {
+    JERRY_ASSERT (!lit_utf8_iterator_is_eos (&iter));
+    c = lit_utf8_iterator_read_next (&iter);
+
+    if (c == LIT_CHAR_BACKSLASH)
+    {
+      return false;
+    }
+  }
+  while (c != end_char);
```
LaszloLango: What happens on wrong formatted strings? Will it stuck in infinite loop? ex: `eof` is found before `end_char`

jerry-core/parser/js/parser.cpp
```diff
@@ -2881,11 +2881,23 @@ preparse_scope (bool is_global)
...
-    is_use_strict = true;
+    if (lit_literal_equal_type_cstr (lit_get_literal_by_cp (token_data_as_lit_cp ()), "use strict")
+        && lexer_is_no_escape_sequences_in_token_string (tok))
+    {
+      scopes_tree_set_strict_mode (STACK_TOP (scopes), true);
+      is_use_strict = true;
+      break;
+    }
+    skip_newlines ();
+    if (token_is (TOK_SEMICOLON))
```
LaszloLango: Please insert an empty line before this if statement.

jerry-core/parser/js/parser.cpp
```diff
@@ -2881,11 +2881,23 @@ preparse_scope (bool is_global)
...
-                                                            "use strict"))
+  /*
+   * Check Directive Prologue for Use Strict directive (see ECMA-262 5.1 section 14.1)
+   */
+  while (token_is (TOK_STRING))
   {
-    scopes_tree_set_strict_mode (STACK_TOP (scopes), true);
-    is_use_strict = true;
+    if (lit_literal_equal_type_cstr (lit_get_literal_by_cp (token_data_as_lit_cp ()), "use strict")
+        && lexer_is_no_escape_sequences_in_token_string (tok))
```
LaszloLango: I think it could be faster if you swap the part of the conditions. The last check seems much simpler for me and if it fails, the strings shouldn't be compared.

jerry-core/parser/js/lexer.cpp
sand1k: The function expects a token of type `TOK_STRING`, this means that this token was previously parsed with `lexer_parse_string`, which does all the checks. Hence the string should be well-formed, otherwise parse error would have been raised.

LaszloLango:
+1 lgtm

jerry-core/parser/js/parser.cpp
sand1k: I think that escape sequences in strings from directive prologue are not common.
The more common thing is string which is not equal to 'use strict'. In this case original version filters out such strings in the first check.

jerry-core/parser/js/parser.cpp
sand1k: Fixed.

egavrin:
`make push`

sand1k:
Pushed to master.
Commits: 
d5fd0b09b158f0ffc6741c702c7076aebb289340
17cdc35d6dfad66f70abc38f31dd0603e1327084

PR #331, opened 2015-07-07
Introduce 'USE_COMPILER_DEFAULT_LIBC' cmake option for switching from jerry-libc to a libc, provided by compiler

ruben-ayrapetyan:

PR #343, opened 2015-07-09
Fix EXTERNAL_LIBC_INTERFACE check with USE_COMPILER_DEFAULT_LIBC

seanshpark:
Related with #331


egavrin:
Cross build fails are not related to the PR.

egavrin:
Good to me

galpeter:
lgtm

ruben-ayrapetyan:
@seanshpark, could you, please, describe the issue that is fixed by the changes?

seanshpark:
Yes, 
What I understood is that, when 'USE_COMPILER_DEFAULT_LIBC' is 'YES' then 'EXTERNAL_LIBC_INTERFACE' should not be defined or if defined, value should be "UNDEFINED". if not, the both is defined by the caller and  I think should be treated as conflict.
With IoT.js, building for arm-linux with `USE_COMPILER_DEFAULT_LIBC` but without `EXTERNAL_LIBC_INTERFACE`, I get
```
CMake Error at CMakeLists.txt:98 (message):
  EXTERNAL_LIBC_INTERFACE='' should not be set in case compiler's default
  libc is used (USE_COMPILER_DEFAULT_LIBC=YES)
``` 
(line number 98 may not be the same)
So to solve this result is this PR

ruben-ayrapetyan:
@seanshpark, thank you for detailed explanation of the issue.

ruben-ayrapetyan:
Looks good to me

egavrin:
`make push`

Issue #255, opened 2015-06-29
Assertion 'mem_pools == NULL' failed in mem_pools_finalize


renatahodovan:
###### Jerry version:
    Checked revision: 06dffde
    Build: debug.linux
###### OS: 
    Ubuntu 14.10, x86_64
###### Test case:
```javascript
v_0 = /(?!(?!l{666,}))/;
v_0.exec("llllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllllll");
```
###### Backtrace:
```javascript
ICE: Assertion 'mem_pools == NULL' failed at jerryscript/jerry-core/mem/mem-poolman.cpp(mem_pools_finalize):89.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x00000000004852f6 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
32    SYSCALL_2
(gdb) bt
# 0  0x00000000004852f6 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
# 1  0x00000000004003c2 in syscall_2 (syscall_no=62, arg1=28622, arg2=6)
    at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:96
# 2  0x0000000000400563 in abort () at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:175
# 3  0x000000000043a9b7 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.cpp:65
# 4  0x000000000043aa06 in jerry_assert_fail (assertion=0x4a54b1 "mem_pools == NULL", 
    file=0x4a5470 "jerryscript/jerry-core/mem/mem-poolman.cpp", 
    function=0x485850 <_ZZ18mem_pools_finalizevE8__func__.lto_priv.1071> "mem_pools_finalize", line=89)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:92
# 5  0x0000000000481ce8 in mem_pools_finalize () at jerryscript/jerry-core/mem/mem-poolman.cpp:89
# 6  0x0000000000481a5d in mem_finalize (is_show_mem_stats=false) at jerryscript/jerry-core/mem/mem-allocator.cpp:58
[...]

Issue #139, opened 2015-06-01
Fix interfaces for processing values and completion values


ruben-ayrapetyan:
Hello.

Currently, interfaces like `ecma_get_string_from_completion_value` and `ecma_get_string_from_value` have the same prototypes, and so, can be confused.

We need to update their prototypes to check for the situation at compile-time.

LaszloLango:
After #888, this issue is irrelevant.

PR #349, opened 2015-07-09
Assertion fix in RegExp.

LaszloLango:
Assertion 'mem_pools == NULL' failed in mem_pools_finalize

### DIFF
--- a/jerry-core/ecma/operations/ecma-regexp-object.cpp
+++ b/jerry-core/ecma/operations/ecma-regexp-object.cpp
@@ -644,7 +644,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
         re_ctx_p->saved_p[RE_GLOBAL_END_IDX] = str_p;
         *res_p = str_p;
         re_ctx_p->recursion_depth--;
-        return ret_value = ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_TRUE); /* match */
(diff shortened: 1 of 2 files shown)


LaszloLango:
Related issue #255

egavrin:
Good to me

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -1003,12 +1003,13 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
           ecma_completion_value_t match_value = re_match_regexp (re_ctx_p, bc_p, str_p, &sub_str_p);
           if (!ecma_is_value_true (match_value))
```
galpeter: Not related to this PR:
The `ecma_is_value_true` call is not correct for the `match_value` as the function expects `ecma_value_t`, but the variable is an `ecma_completion_value_t`.

zherczeg: It is sad that the compiler didn't catch this. We should make a fix for this.

zherczeg:
+1 lgtm

egavrin:
`make push`

jerry-core/ecma/operations/ecma-regexp-object.cpp
ruben-ayrapetyan: @zherczeg, please, see #139.

PR #352, opened 2015-07-09
Implement String.prototype.charCodeAt()

lvidacs:
JerryScript-DCO-1.0-Signed-off-by: Laszlo Vidacs lvidacs.u-szeged@partner.samsung.com

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
@@ -169,7 +169,59 @@ static ecma_completion_value_t
 ecma_builtin_string_prototype_object_char_code_at (ecma_value_t this_arg, /**< this argument */
                                                    ecma_value_t arg) /**< routine's argument */
 {
-  ECMA_BUILTIN_CP_UNIMPLEMENTED (this_arg, arg);
+  ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
+
+  /* 1 */
+  ECMA_TRY_CATCH (check_coercible_val,
+                  ecma_op_check_object_coercible (this_arg),
+                  ret_value);
+
+  /* 2 */
+  ECMA_TRY_CATCH (to_string_val,
+                  ecma_op_to_string (this_arg),
+                  ret_value);
+
+  /* 3 */
+  ECMA_OP_TO_NUMBER_TRY_CATCH (index_num,
+                               arg,
+                               ret_value);
+
+  /* 4 */
+  ecma_string_t *original_string_p = ecma_get_string_from_value (to_string_val);
+  const ecma_length_t len = ecma_string_get_length (original_string_p);
+
+  ecma_number_t *ret_num_p = ecma_alloc_number ();
+
+  /* 5 */
+  // When index_num is NaN, then the first two comparisons are false
+  if (index_num < 0 || index_num >= len || (ecma_number_is_nan (index_num) && !len))
+  {
+    *ret_num_p = ecma_number_make_nan ();
+  }
+  else
+  {
+    /* 6 */
+    /*
+     * String length is currently uit32_t, but index_num may be bigger,
+     * ToInteger performs floor, while ToUInt32 performs modulo 2^32,
+     * hence after the check 0 <= index_num < len we assume to_uint32 can be used.
+     * We assume to_uint32 (NaN) is 0.
+     */
+    JERRY_ASSERT (ecma_number_is_nan (index_num) || ecma_number_to_uint32 (index_num) == ecma_number_trunc (index_num));
+
+    ecma_char_t new_ecma_char = ecma_string_get_char_at_pos (original_string_p, ecma_number_to_uint32 (index_num));
+    *ret_num_p = ecma_uint32_to_number (new_ecma_char);
+  }
+
+  ecma_value_t new_value = ecma_make_number_value (ret_num_p);
+  ret_value = ecma_make_normal_completion_value (new_value);
+
+  ECMA_OP_TO_NUMBER_FINALIZE (index_num);
+
+  ECMA_FINALIZE (to_string_val);
+  ECMA_FINALIZE (check_coercible_val);
+
+  return ret_value;
 } /* ecma_builtin_string_prototype_object_char_code_at */
 
 /**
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -168,7 +168,52 @@ static ecma_completion_value_t
...
+  /* 3 */
+  ECMA_OP_TO_NUMBER_TRY_CATCH (index_num,
+                               arg,
+                               ret_value);
+
+  /* 4 */
+  ecma_string_t *original_string_p = ecma_get_string_from_value (to_string_val);
+  const ecma_length_t len = ecma_string_get_length (original_string_p);
+
+  ecma_number_t * ret_num = ecma_alloc_number ();
```
egavrin: `*ret_num_p`

egavrin:
Good to me

lvidacs:
Thanks, the patch is updated.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -168,7 +168,52 @@ static ecma_completion_value_t
...
+  /* 3 */
+  ECMA_OP_TO_NUMBER_TRY_CATCH (index_num,
+                               arg,
+                               ret_value);
+
+  /* 4 */
+  ecma_string_t *original_string_p = ecma_get_string_from_value (to_string_val);
+  const ecma_length_t len = ecma_string_get_length (original_string_p);
+
+  ecma_number_t * ret_num_p = ecma_alloc_number ();
```
galpeter: No need for the space after the `*`

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -168,7 +168,52 @@ static ecma_completion_value_t
...
+  }
+  else
+  {
+    /* 6 */
+    ecma_char_t new_ecma_char = ecma_string_get_char_at_pos (original_string_p, ecma_number_to_uint32 (index_num));
+    *ret_num_p = ecma_uint32_to_number (new_ecma_char);
+  }
+
+  ecma_value_t new_value = ecma_make_number_value (ret_num_p);
+  ret_value = ecma_make_normal_completion_value (ecma_copy_value (new_value, true));
```
galpeter: Do we really need to copy the value?

tests/jerry/string-prototype-charcodeat.js
```diff
@@ -0,0 +1,106 @@
...
+try {
+  String.prototype.charCodeAt.call(null);
+  assert(false);
+} catch(e) {
+  assert(e instanceof TypeError);
+}
+
+// check coercible - undefined
+try {
+  assert(true.charCodeAt());
```
galpeter: This testcase will not invoke the `charCodeAt` at all. I think you wanted to write: `String.prototype.charCodeAt.call(true)`

zherczeg:
LGTM

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -169,7 +169,50 @@ static ecma_completion_value_t
...
+
+  /* 5 */
+  if (index_num < 0 || index_num >= len || !len)
+  {
+    *ret_num_p = ecma_number_make_nan ();
+  }
+  else
+  {
+    /* 6 */
+    ecma_char_t new_ecma_char = ecma_string_get_char_at_pos (original_string_p, ecma_number_to_uint32 (index_num));
```
ruben-ayrapetyan: While actual length of strings, currently, can't exceed `uint32_t` range, value of `index_num` can be bigger. `ToInteger` is not equivalent to `ToUInt32`, as `ToUInt32` performs "modulo 2^32" operation on its argument, and `ToInteger` performs `floor`.

The code would work correctly, as `index_num` doesn't exceed `len`, as checked above.

Could you, please, add an assertion that `ecma_number_to_uint32 (index_num) == ecma_number_trunc (index_num)` and add comment, describing why `ToInteger` and `ToUInt32` are equivalent in the case?

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -169,7 +169,50 @@ static ecma_completion_value_t
...
+                               ret_value);
+
+  /* 4 */
+  ecma_string_t *original_string_p = ecma_get_string_from_value (to_string_val);
+  const ecma_length_t len = ecma_string_get_length (original_string_p);
+
+  ecma_number_t *ret_num_p = ecma_alloc_number ();
+
+  /* 5 */
+  if (index_num < 0 || index_num >= len || !len)
```
ruben-ayrapetyan: `!len` part is not necessary, as if `len == 0`, `index_num < 0 || index_num >= len` becomes `index_num < 0 || index_num >= 0` that is always true.

tests/jerry/string-prototype-charcodeat.js
```diff
@@ -0,0 +1,91 @@
...
+
+assert(Object.getOwnPropertyDescriptor(String.prototype.charCodeAt, 'length').writable === false);
+
+assert(String.prototype.charCodeAt.length === 1);
+
+// check empty string
+assert(isNaN(String.prototype.charCodeAt.call(new String())));
+
+// check NaN
+assert("hello world!".charCodeAt(0) === 104);
```
ruben-ayrapetyan: Seems that comment "check NaN" is incorrect.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
lvidacs: in case of NaN both comparisons are false, and we have to return NaN iff len == 0

lvidacs:
@ruben-ayrapetyan thanks for the comments, the patch is updated

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
ruben-ayrapetyan: I see. Could you, please, change the condition part to explicit check for `NaN` value using `ecma_number_is_nan`?

ruben-ayrapetyan:
@lvidacs, thank you for update.

ruben-ayrapetyan:
@lvidacs, after fixing https://github.com/Samsung/jerryscript/pull/352#discussion_r34996436 the changes would look good to me

ruben-ayrapetyan:
Looks good to me

egavrin:
`make push`

Issue #333, opened 2015-07-08
Stress testing of the engine


egavrin:
Do you know any fuzzers or stress testing suites that can useful for us?

- jsfunfuzz - https://www.squarefree.com/2007/08/02/introducing-jsfunfuzz/
- afl-fuzz for javascript - https://github.com/tunz/afl-fuzz-js

akiss77:
The University of Szeged has a proprietary fuzz testing framework with no imminent plans for open sourcing but is happy to share any issues it discovers in Jerry. (Actually, several previously reported issues were discovered with its help.)

renatahodovan:
The "problem" with jsfunfuzz is that it generates the test cases on the fly which means that the whole generator script needs to be loaded by the engine every time. Hence jerry would need to load ~5500 LOC to generate a single test case but unfortunately it runs of the registers quickly.

I've also tried the recently popular afl fuzzer against jerry however it constantly crashes on the longjmp bug (#341) and it's unusable this way.

ruben-ayrapetyan:
@renatahodovan, maybe we can generate tests by running jsfunfuzz separately, and then use the generated tests for checking.

<br/>
> I've also tried the recently popular afl fuzzer against jerry however it constantly crashes on the longjmp bug (#341) and it's unusable this way.

Could you, please, check #390?

ruben-ayrapetyan:
#341 is fixed.

egavrin:
@renatahodovan could you please share how to run afl-fuzz for javascript?

renatahodovan:
@egavrin sure, however I'm also still working on how to exploit all of its features.

A customized version of afl-fuzz is available from here: https://github.com/tunz/afl-fuzz-js
You can run it two different ways: in the first case you can use your own jerry build, choose a set of initial test cases that will be mutated and a predefined dictionary to guide the mutation and go ahead. From the manual, starting a fuzz session looks something like this:

```javascript
/path/to/afl-fuzz -i [input directory] -o [output directory] -x [dictionary directory] -m 8G ./jerry @@
```
I've ran this for days but it hasn't found much (expect the longjump crashes).

However, there is second coverage driven option (what is said much more efficient). The problem is that to use this option the target must be instrumented with the afl-fuzzer's own wrapped gcc/g++. Unfortunately, replacing jerry's default compiler is not seems too easy. I've achieved to build everything with afl-g++, but it still fails on linking. This is what I'm experimenting with right now :)

akiss77:
[...]

PR #366, opened 2015-07-10
Style fixes for RegExp engine.

LaszloLango:


jerry-core/ecma/operations/ecma-regexp-object.h
```diff
@@ -29,26 +29,30 @@
...
-  uint32_t match_limit;
-  uint32_t recursion_depth;
-  uint32_t num_of_captures;
-  uint32_t num_of_non_captures;
-  uint32_t *num_of_iterations;
-  uint8_t flags;
+  const lit_utf8_byte_t **saved_p; /**< saved result string pointers */
+  const lit_utf8_byte_t *input_start_p; /**< start of input patter string */
+  const lit_utf8_byte_t *input_end_p; /**< end of input patter string */
+  uint32_t match_limit; /**< maching limit counter */
```
ruben-ayrapetyan: Typos: "matching", "pattern"

jerry-core/ecma/operations/ecma-regexp-object.h
```diff
@@ -29,26 +29,30 @@
...
-  const lit_utf8_byte_t **saved_p;
-  const lit_utf8_byte_t *input_start_p;
-  const lit_utf8_byte_t *input_end_p;
-  uint32_t match_limit;
-  uint32_t recursion_depth;
-  uint32_t num_of_captures;
-  uint32_t num_of_non_captures;
-  uint32_t *num_of_iterations;
-  uint8_t flags;
+  const lit_utf8_byte_t **saved_p; /**< saved result string pointers */
```
ruben-ayrapetyan: Could you, please, clarify what is the saved result string pointers? Is it already described somewhere?

jerry-core/ecma/operations/ecma-regexp-object.h
```diff
@@ -29,26 +29,30 @@
  * @{
  */
 
-#define RE_EXECUTE_RECURSION_LIMIT  1000  /* Limit of RegExp executor recursion depth */
-#define RE_EXECUTE_MATCH_LIMIT      10000 /* Limit of RegExp execetur matching steps */
+/**
+ * Limit of RegExp executor recursion depth
+ */
+#define RE_EXECUTE_RECURSION_LIMIT  1000
```
ruben-ayrapetyan: Could you, please, describe here, in the pull request notes - what recursion is meant here and where it occurs? Maybe, with some examples.

jerry-core/parser/regexp/re-compiler.cpp
```diff
@@ -25,26 +25,25 @@
...
-
-void
-regexp_dump_bytecode (re_bytecode_ctx_t *bc_ctx);
 
 /**
- * FIXME:
- *       Add missing 're' prefixes to the component's external and internal interfaces
+ * Get length of bytecode
  */
+#define BYTECODE_LEN(bc_ctx_p) ((uint32_t) (bc_ctx_p->current_p - bc_ctx_p->block_start_p))
```
ruben-ayrapetyan: It would be better to add function for this.

jerry-core/parser/regexp/re-compiler.cpp
```diff
@@ -118,60 +117,60 @@ bytecode_list_insert (re_bytecode_ctx_t *bc_ctx_p, /**< RegExp bytecode context
...
+                  uint32_t offset, /**< distance from the start of the container */
+                  re_opcode_t opcode) /**< input opcode */
 {
-  bytecode_list_insert (bc_ctx_p, offset, (re_bytecode_t*) &opcode, sizeof (re_bytecode_t));
-} /* insert_opcode */
+  re_bytecode_list_insert (bc_ctx_p, offset, (re_bytecode_t*) &opcode, sizeof (re_bytecode_t));
+} /* re_insert_opcode */
 
 /**
  * Insert a parameter of a RegExp opcode
```
ruben-ayrapetyan: Are parameters always uint32?

jerry-core/parser/regexp/re-compiler.cpp
```diff
@@ -287,8 +286,8 @@ get_start_opcode_type (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context
  * Get the type of a group end
  */
 static re_opcode_t
-get_end_opcode_type (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context */
-                     bool is_capturable) /**< is capturabel group */
+re_get_end_opcode_type (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context */
+                        bool is_capturable) /**< is capturabel group */
```
ruben-ayrapetyan: Typo: "capturable"

jerry-core/parser/regexp/re-compiler.cpp
```diff
@@ -330,40 +329,40 @@ insert_into_group (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context */
...
-} /* insert_into_group */
+} /* re_insert_into_group */
 
 /**
  * Enclose the given bytecode to a group and inster jump value
  */
 static void
-insert_into_group_with_jump (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context */
+re_insert_into_group_with_jump (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context */
                              uint32_t group_start_offset, /**< offset of group start */
```
ruben-ayrapetyan: Wrong indentation.

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -32,8 +32,13 @@
...
 #define RE_ADVANCE(str_p, advance) do { str_p += advance; } while (0)
 
+/**
+  * Get next input character
+  *
+  * @return ecma_char_t
+  */
 static ecma_char_t
-get_ecma_char (lit_utf8_byte_t **char_p)
+get_ecma_char (lit_utf8_byte_t **char_p) /**< pointer of input string */
```
ruben-ayrapetyan: Component prefix is missing.

jerry-core/parser/regexp/re-parser.h
```diff
@@ -60,21 +73,27 @@ typedef uint8_t token_type_t;
 #define RE_CONTROL_CHAR_FF   0x000c /* \f */
```
ruben-ayrapetyan: By the way, the values are now defined as `LIT_CHAR_*`.

jerry-core/ecma/operations/ecma-regexp-object.h
LaszloLango: It is used in `re_match_regexp` in `ecma-regexp-object.cpp`. It is used to avoid too deep stack calls and bail out with a range error. It also makes sure that an infinite loop never occurs in the executor, so the engine never stuck.

jerry-core/parser/regexp/re-compiler.cpp
LaszloLango: Yes. In the current implementation we use fix length for parameters. This should be optimized later.

jerry-core/parser/regexp/re-parser.h
LaszloLango: I plan to do the unicode related fixes in a separate PR.

jerry-core/ecma/operations/ecma-regexp-object.h
```diff
@@ -29,26 +29,30 @@
...
-  uint32_t *num_of_iterations;
-  uint8_t flags;
+  const lit_utf8_byte_t **saved_p; /**< saved result string pointers */
+  const lit_utf8_byte_t *input_start_p; /**< start of input patter string */
+  const lit_utf8_byte_t *input_end_p; /**< end of input patter string */
+  uint32_t match_limit; /**< maching limit counter */
+  uint32_t recursion_depth; /**< recursion depth counter */
+  uint32_t num_of_captures; /**< number of capture groups */
+  uint32_t num_of_non_captures; /**< number of non-capture groups */
+  uint32_t *num_of_iterations; /**< number of iterations */
```
kkristof: Please add `_p`

jerry-core/parser/regexp/re-compiler.cpp
LaszloLango: Ok

jerry-core/ecma/operations/ecma-regexp-object.h
LaszloLango: It is also described in `ecma-regexp-object.cpp`.

```
/**
 * RegExp results are stored in an array of string pointers. If N is the number
 * of groups then the length of the array is 2*N, because every group has a start
 * and end. We have to handle those pointers.
 *
 *   [0] RE global start
 *   [1] RE global end
 *   [2] 1st group start
 *   [3] 1st group end
 *   ...
 *   [n]   n/2 th group start
 *   [n+1] n/2 th group end
 */
```
It is basically the State of RegExp. I added reference to `ECMA 262 v5, 15.10.2.1, State`.

LaszloLango:
@ruben-ayrapetyan, I've updated the PR, please check.

jerry-core/parser/regexp/re-compiler.h
```diff
@@ -57,8 +57,14 @@
 #define RE_OP_CHAR_CLASS                                    26
 #define RE_OP_INV_CHAR_CLASS                                27
```
LaszloLango: @ruben-ayrapetyan, it looks like other parts of the engine are using enums instead of defines. Should I change definitions of RegExp opcodes and tokens to enum?

ruben-ayrapetyan: @LaszloLango, I think that it is better to use `enum`. If we would switch to C++, it would be even better to use `enum class`.

LaszloLango: OK, I'll updated the patch.

jerry-core/ecma/operations/ecma-regexp-object.h
ruben-ayrapetyan: @LaszloLango, thank you for explanation.

Could you, please, also provide a few examples of regular expressions with different recursion limits, and of an expression that would lead to infinite loop without the recursion limits?

Does the limit somehow restrict support of specification-defined semantics?

jerry-core/parser/regexp/re-compiler.cpp
ruben-ayrapetyan: So, in the future there would be several interfaces like `re_insert_u8`, `re_insert_u16`, `re_insert_u32`?

jerry-core/parser/regexp/re-parser.h
ruben-ayrapetyan: Ok.

jerry-core/parser/regexp/re-compiler.h
ruben-ayrapetyan: Ok. Thank you.

jerry-core/ecma/operations/ecma-regexp-object.h
LaszloLango: @ruben-ayrapetyan, a simple example is using groups. Each group usage increase recursion limit by one.
Maximum recursion of `/a/` is one, `/(a)/` is two, `/((a))/` is three. Iteration of groups also increase the recursion depth. Maximum recursion of `/(a){100}/` is 101. So it can restrict the engine, but not in specification-defined semantics. You cannot run too big or complex RegExp. In the current implementation the engine should not lead to infinite loop for any expressions. This is only a sanity check now.

ruben-ayrapetyan: @LaszloLango, I see. Thank you for detailed explanation.

While presence of the limitation looks good to me, I think way of limiting can be improved.
Currently, if I understand correctly, the following cases can occur:
 - there is enough heap and stack space, but limit is reached, and so RangeError is thrown;
 - there is not enough heap space, limit is not reached - out-of-memory is reported;
 - there is not enough stack space, limit is not reached - stack overflow occurs.

For me, it seems better, if after a while, we could avoid stack overflows in the engine.

Also, if I understand correctly, specification doesn't permit throwing RangeError in case regular expression can't be handled due to implementation-defined limitations. So, engine should try to handle a regular expression without throwing RangeError, until out-of-memory occurs.

What do you think about this? If you think it is correct, could you, please, consider this in context of future regular expressions implementation updates, and raise an issue about this?

jerry-core/parser/regexp/re-compiler.cpp
ruben-ayrapetyan: Or would there be some other, non-integer, types?

jerry-core/ecma/operations/ecma-regexp-object.h
LaszloLango: @ruben-ayrapetyan, yes, you understand correctly. If the memory management is stable, then we can eliminate these checks and execute until out-of-memory occurs. I will raise an issue for this.

ruben-ayrapetyan: Ok. Thank you.

By the way, heap memory management is currently stable, isn't it?

jerry-core/parser/regexp/re-compiler.cpp
LaszloLango: Not exactly, but maybe that could be a solution too. What I'd like to do is to store everything based on it's value. For example if we mask the highest bit for every byte then we can store 0-127 in one byte instead of 4. Any good idea for compressing RegExp bytecode is welcome.

jerry-core/ecma/operations/ecma-regexp-object.h
LaszloLango: I cannot judge that it stable or not. There is an open issue for stress testing #333, but if you say that it's stable, then I believe you. :)

LaszloLango:
@ruben-ayrapetyan, I've updated the PR, please check.

jerry-core/ecma/operations/ecma-regexp-object.h
ruben-ayrapetyan: I think it is stable enough to catch out-of-memory :)

ruben-ayrapetyan:
Looks good to me

kkristof:
Looks good to me.

jerry-core/parser/regexp/re-compiler.h
```diff
@@ -22,77 +22,83 @@
...
-#define RE_OP_NON_CAPTURE_GROUP_START                       6
-#define RE_OP_NON_CAPTURE_GREEDY_ZERO_GROUP_START           7
-#define RE_OP_NON_CAPTURE_NON_GREEDY_ZERO_GROUP_START       8
-#define RE_OP_NON_CAPTURE_GREEDY_GROUP_END                  9
-#define RE_OP_NON_CAPTURE_NON_GREEDY_GROUP_END              10
+typedef enum
+{
+  RE_OP_EOF,
+  /* Group opcode order is important, because RE_IS_CAPTURE_GROUP is based on it.
+   * Change it carfully. Capture opcodes should be at first.
```
galpeter: Typo: `carfully` -> `carefully`

jerry-core/parser/regexp/re-compiler.cpp
```diff
@@ -538,39 +541,42 @@ parse_alternative (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context */
...
+        re_insert_into_group_with_jump (re_ctx_p, new_atom_start_offset, idx, false);
         break;
       }
       case RE_TOK_START_CHAR_CLASS:
       case RE_TOK_START_INV_CHAR_CLASS:
       {
         JERRY_DDLOG ("Compile a character class\n");
-        append_opcode (bc_ctx_p,
+        re_append_opcode (bc_ctx_p,
                        re_ctx_p->current_token.type == RE_TOK_START_CHAR_CLASS
```
galpeter: Incorrect alignment

LaszloLango:
@galpeter, updeted, please check

galpeter:
lgtm

PR #369, opened 2015-07-10
Fix preparsing of variable declaration lists, which are not divided by a semicolon

sand1k:
This patch fixes the following tests from test262:
- ch13/13.2/13.2-1-s.js
- ch13/13.2/13.2-3-s.js

### DIFF
--- a/jerry-core/parser/js/parser.cpp
+++ b/jerry-core/parser/js/parser.cpp
@@ -3018,6 +3018,10 @@ preparse_scope (bool is_global)
           }
           else if (token_is (TOK_KEYWORD))
           {
+            if (is_keyword (KW_VAR))
+            {
+              is_in_var_declaration_list = false;
+            }
             break;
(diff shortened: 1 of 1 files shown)


egavrin:
Good to me

ruben-ayrapetyan:
Looks good to me

zherczeg:
The patch looks good to me.

However, I don't see any test case in the patch and the comment only mention two test262 files. Please at least add an example here as a comment, so I can see what is fixed.

ruben-ayrapetyan:
@zherczeg, the changes fix handling of two consecutive VariableDeclarationList statements, that are not divided by a semicolon.

egavrin:
@zherczeg example:
```cpp
var foo = function () {
      this.caller = 12;
}
var obj = new foo();
```

zherczeg:
Ok, land it.

Issue #130, opened 2015-05-29
Assertion '!ecma_number_is_nan (num)' failed in ecma_number_is_zero


renatahodovan:
###### Jerry version:
    Checked revision: 77df022
    Build: debug.linux
###### OS: 
    Ubuntu 14.04, x86_64
###### Test case:
```javascript
v_1 = Math.floor('\u3d52');
v_1 = Math.floor([]);
```
###### Backtrace:
```javascript
ICE: Assertion '!ecma_number_is_nan (num)' failed at jerryscript/jerry-core/ecma/base/ecma-helpers-number.cpp(ecma_number_is_zero):370.
Error: ERR_FAILED_INTERNAL_ASSERTION
[Inferior 1 (process 19644) exited with code 0170]
(gdb) bt
# 0  ecma_number_is_zero (num=nan(0x8000000000001)) at jerryscript/jerry-core/ecma/base/ecma-helpers-number.cpp:370
# 1  0x000000000046e626 in ecma_op_same_value (x=565, y=129) at jerryscript/jerry-core/ecma/operations/ecma-conversion.cpp:123
# 2  0x000000000046d0bb in ecma_op_general_object_define_own_property (obj_p=0x7eb758 <_ZL13mem_heap_area.4015+24>, property_name_p=0x7fffffffd330, 
    property_desc_p=0x7fffffffd060, is_throw=false) at jerryscript/jerry-core/ecma/operations/ecma-objects-general.cpp:669
# 3  0x000000000047cf59 in ecma_op_object_define_own_property (obj_p=0x7eb758 <_ZL13mem_heap_area.4015+24>, property_name_p=0x7fffffffd330, 
    property_desc_p=0x7fffffffd060, is_throw=false) at jerryscript/jerry-core/ecma/operations/ecma-objects.cpp:443
# 4  0x000000000046b909 in ecma_op_general_object_put (obj_p=0x7eb758 <_ZL13mem_heap_area.4015+24>, property_name_p=0x7fffffffd330, value=565, 
    is_throw=false) at jerryscript/jerry-core/ecma/operations/ecma-objects-general.cpp:277
# 5  0x000000000047cc66 in ecma_op_object_put (obj_p=0x7eb758 <_ZL13mem_heap_area.4015+24>, property_name_p=0x7fffffffd330, value=565, is_throw=false)
    at jerryscript/jerry-core/ecma/operations/ecma-objects.cpp:273
[...]

PR #370, opened 2015-07-10
Fix NaN handling in ecma_op_same_value

ruben-ayrapetyan:
Related issue: #130

### DIFF
--- a/jerry-core/ecma/operations/ecma-conversion.cpp
+++ b/jerry-core/ecma/operations/ecma-conversion.cpp
@@ -102,22 +102,28 @@ ecma_op_same_value (ecma_value_t x, /**< ecma-value */
   {
     return false;
   }
-
-  if (is_x_undefined
-      || is_x_null)
+  else if (is_x_undefined || is_x_null)
   {
     return true;
   }
-
-  if (is_x_number)
+  else if (is_x_number)
   {
     ecma_number_t *x_num_p = ecma_get_number_from_value (x);
     ecma_number_t *y_num_p = ecma_get_number_from_value (y);
 
-    if (ecma_number_is_nan (*x_num_p)
-        && ecma_number_is_nan (*y_num_p))
+    bool is_x_nan = ecma_number_is_nan (*x_num_p);
+    bool is_y_nan = ecma_number_is_nan (*y_num_p);
+
+    if (is_x_nan || is_y_nan)
     {
-      return true;
+      /*
+       * If both are NaN
+       *   return true;
+       * else
+       *   // one of the numbers is NaN, and another - is not
+       *   return false;
+       */
+      return (is_x_nan && is_y_nan);
     }
     else if (ecma_number_is_zero (*x_num_p)
              && ecma_number_is_zero (*y_num_p)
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-conversion.cpp
```diff
@@ -102,22 +102,30 @@ ecma_op_same_value (ecma_value_t x, /**< ecma-value */
   {
     return false;
   }
-
-  if (is_x_undefined
-      || is_x_null)
+  else if (is_x_undefined
```
egavrin: Why multiline?

```cpp
else if (is_x_undefined || is_x_null)
```

jerry-core/ecma/operations/ecma-conversion.cpp
```diff
@@ -102,22 +102,30 @@ ecma_op_same_value (ecma_value_t x, /**< ecma-value */
...
   {
     ecma_number_t *x_num_p = ecma_get_number_from_value (x);
     ecma_number_t *y_num_p = ecma_get_number_from_value (y);
 
-    if (ecma_number_is_nan (*x_num_p)
-        && ecma_number_is_nan (*y_num_p))
+    bool is_x_nan = ecma_number_is_nan (*x_num_p);
+    bool is_y_nan = ecma_number_is_nan (*y_num_p);
+
+    if (is_x_nan
```
egavrin: likewise

egavrin:
Except style, good to me

jerry-core/ecma/operations/ecma-conversion.cpp
ruben-ayrapetyan: I think that in the case, in multiline mode, as equal `is_x_` parts are closer, it is simpler to compare the condition parts, and so, simpler to read the code.

egavrin: In that case it's gimmick.

ruben-ayrapetyan: Why?

egavrin: Condition is too simple to move it on other line. The only reason to use multiline condition - it doesn't fit into 120.

ruben-ayrapetyan: Why shouldn't we make multiline conditions in other cases, if it improves readability?

egavrin: Because they don't - it just confusing readers.

zherczeg:
lgtm. I don't mind the condition splitting style as long as it is not a rule.

jerry-core/ecma/operations/ecma-conversion.cpp
ruben-ayrapetyan: // The question was discussed verbally.
// Agreed on combining very simple, similar conditions to one line, and splitting other on several lines.

ruben-ayrapetyan: I'll update pull request.

ruben-ayrapetyan:
@egavrin, pull request was updated. Please, check.

egavrin:
`make push`

PR #371, opened 2015-07-10
Support unicode whitespaces in string-to-number conversion

sand1k:
This patch fixes the following tests from test262:
- ch09/9.3/9.3.1/S9.3.1_A2.js
- ch09/9.3/9.3.1/S9.3.1_A3_T1.js
- ch09/9.3/9.3.1/S9.3.1_A3_T2.js

### DIFF
--- a/jerry-core/lit/lit-unicode-ranges.inc.h
+++ b/jerry-core/lit/lit-unicode-ranges.inc.h
@@ -21,6 +21,9 @@
  *          http://www.unicode.org/Public/3.0-Update/UnicodeData-3.0.0.html
  *
  * The range lists were generated using tools/print-unicode-ranges.sh script from UnicodeData-3.0.0.txt
+ *
+ * NOTE:
+ *   Some ranges in "Separator, Space" category were added manually, see the according definitions.
  */
 
 /**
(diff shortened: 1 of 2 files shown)


egavrin:
Good to me

zherczeg:
LGTM

jerry-core/lit/lit-unicode-ranges.inc.h
```diff
@@ -2436,12 +2436,16 @@ LIT_UNICODE_RANGE_ZS (0x00A0, 0x00A0) /* NO-BREAK SPACE */
...
 
+LIT_UNICODE_RANGE_ZS (0x180E, 0x180E) /* MONGOLIAN VOWEL SEPARATOR */
+
 LIT_UNICODE_RANGE_ZS (0x2000, 0x200B) /*     EN QUAD
                                               <--->
                                          ZERO WIDTH SPACE */
 
 LIT_UNICODE_RANGE_ZS (0x202F, 0x202F) /* NARROW NO-BREAK SPACE */
 
+LIT_UNICODE_RANGE_ZS (0x205F, 0x205F) /* MEDIUM MATHEMATICAL SPACE */
```
ruben-ayrapetyan: The comment in the file header says that `The range lists were generated using tools/print-unicode-ranges.sh script from UnicodeData-3.0.0.txt`, but this change makes the statement incorrect. Could you, please, update the header, and also, add special comment for manually added ranges, describing the ranges origin? Maybe, the manually added ranges should also be grouped at start of range block (near `ifdef`).

ruben-ayrapetyan:
After fixing https://github.com/Samsung/jerryscript/pull/371#discussion-diff-34455958, the changes would look good to me.

ruben-ayrapetyan:
Looks good to me

PR #375, opened 2015-07-10
Update RegExp.prototype.exec so that it returns 'null' instead of 'undefined' in case argument doesn't match corresponding regular expression

ruben-ayrapetyan:
ECMA-262 v5, 15.10.6.2 specifies that `null` is returned "if `string` didn't match".


egavrin:
Good to me

LaszloLango:
+1 lgtm, thanks for the fix.

PR #392, opened 2015-07-14
Support raising ReferenceError from parser; add temporary workaround for invalid left-hand-side expressions

ruben-ayrapetyan:


jerry-core/parser/js/parser.cpp
```diff
@@ -3173,8 +3170,8 @@ parser_parse_program (const jerry_api_char_t *source_p, /**< source code buffer
   scopes_tree_set_strict_mode (STACK_TOP (scopes), is_strict);
   lexer_set_strict_mode (scopes_tree_strict_mode (STACK_TOP (scopes)));
 
-  jmp_buf *syntax_error_label_p = syntax_get_syntax_error_longjmp_label ();
-  int r = setjmp (*syntax_error_label_p);
+  jmp_buf *jsp_early_error_error_label_p = jsp_early_error_get_early_error_longjmp_label ();
```
sand1k: Double "error".

jerry-core/parser/js/jsp-early-error.cpp
```diff
@@ -54,19 +59,36 @@ STATIC_STACK (U8, uint8_t)
...
+jsp_early_error_get_early_error_longjmp_label (void)
 {
-  return &jsp_syntax_error_label;
-} /* syntax_get_syntax_error_longjmp_label */
+  return &jsp_early_error_label;
+} /* jsp_early_error_get_early_error_longjmp_label */
 
 /**
  * Raise SyntaxError, i.e. perform longjmp to SyntaxError longjmp label
  */
```
sand1k: Please, update the comment.

sand1k:
After fixing precommit failure and two notes, the patch would look good to me.

ruben-ayrapetyan:
@sand1k, thank you for your notes. Pull request is updated.

zherczeg:
Am I see correctly that this is just introducing a new syntax error type? LGTM in this case.

ruben-ayrapetyan:
@zherczeg, the patch introduces way to raise ReferenceError from parser. As, sometimes, ReferenceError is an "early error" (for example, `1 = 1`-like expressions), it should be thrown before execution begins.

Also, second commit of the patch adds two checks that can raise ReferenceError to handle invalid left-hand-side in an assignment expression.

zherczeg:
Ok, land it.

PR #394, opened 2015-07-14
Remove isalpha, isdigit, isxdigit, isspace

ruben-ayrapetyan:


ruben-ayrapetyan:
@zherczeg, could you, please, clarify how `isalpha`, `isdigit` can be replaced in `ecma-builtin-json.cpp`, using `lit_char_*` helpers?

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -125,7 +126,8 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
   ecma_length_t end = str_size;
   for (ecma_length_t i = 0; i < end; i++)
   {
-    if (!(isspace (utf8_string_buff[i])))
+    if (!lit_char_is_white_space (utf8_string_buff[i])
+        && !lit_char_is_line_terminator (utf8_string_buff[i]))
```
zherczeg: Is this correct? If I understand correctly, we just pass a byte, rather than an uint16_t character. I think we should rewrite this to use the UTF iterators, since there are line terminators and white spaces outside of the [0-127] byte range.

ruben-ayrapetyan: @zherczeg, yes, you are right. The code doesn't fully support unicode yet.

Purpose of the pull request is only to remove `isspace`-like functions and switch to `lit_char_*` helpers. May be, it is too early to perform the change.

What do you think about this?

zherczeg: I think we should open an issue and land the patch then.

ruben-ayrapetyan: @zherczeg, I've opened the issue: #405.

ruben-ayrapetyan: @zherczeg, would it be ok, if I move the `isspace`-like functions to `ecma-builtins-json.cpp` (currently it is the only module that uses the functions), and raise an issue for replacing them with proper `lit_char_*` helpers?

zherczeg: OK for me.

ruben-ayrapetyan: Issue #424.

ruben-ayrapetyan:
Pull request is updated. Please, check.

egavrin:
Ok to me.

zherczeg:
LGTM

Issue #108, opened 2015-05-27
Flag to increase verbosity for `make precommit` and `make prerequisites`


egavrin:
Current scripts silently perform its actions, but Jenkins needs on-screen logs to store them.

PR #408, opened 2015-07-15
Makefile refactor

akiss77:
A loosely coupled, intentionally non-squashed set of patches improving the Makefile. Key property of the patches is that they don't change any current default behaviour (but extend possibilities, improve readability of recipes, or fix latent bugs).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -257,8 +257,8 @@ $(BUILD_ALL)_stm32f4: $(BUILD_DIRS_STM32F4)
 	$(Q) ($(MAKE) -C $(BUILD_DIR)/stm32f4 $(JERRY_STM32F4_TARGETS) VERBOSE=1 2>&1 | tee $(OUT_DIR)/$@/make.log $(QLOG) ; ( exit $${PIPESTATUS[0]} ) ) || \
           (echo "Build failed. See $(OUT_DIR)/$@/make.log for details."; exit 1;)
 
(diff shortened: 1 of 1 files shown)


egavrin:
Related issue: #108

egavrin:
Good to me

akiss77:
Ah, but I got it wrong. `tee` is a tricky beast and swallows the exit code of the command earlier in the pipe. (Which would cause always-successful builds from make's perspective.) Will update the PR with fix ASAP.

akiss77:
FYI: updated and rebased

egavrin:
@akiss77 is possible to split logs into two streams `stdout` and `stderr` and put them into separate logs? The idea is to print all errors on screen after in case of build failure - not to analyze logs by eye.

akiss77:
@egavrin can/will experiment. Could we have that as a follow-up, however? This patch set keeps everything working as was before by default. The log splitting approach would cause a change in the default behaviour.

egavrin:
@akiss77 of course :)

Good to me.

ruben-ayrapetyan:
Looks good to me

akiss77:
Seen two LGTMs, thanks, I'll push.

akiss77:
Merged at 6038173e76265ab30ae7b5b70a4d20633e1de9b7

PR #418, opened 2015-07-16
Implement Date.prototype.toISOString()

rtakacs:
Moved toString() logic into the toISOString() method extended with a Z postfix as the standard said.
Since the toString() is "implementation-dependent" the toString() calls the toISOString() method.


egavrin:
Will this change unify smth later?

tests/jerry/date-tostring.js
```diff
@@ -32,6 +32,10 @@ assert (new Date (NaN).toDateString () == "Invalid Date");
 assert (new Date ("2015-02-13").toDateString () == "2015-02-13");
 assert (new Date ("2015-07-08T11:29:05.023").toDateString () == "2015-07-08");
 
+assert (new Date (NaN).toISOString () == "Invalid Date");
+assert (new Date ("2015-07-16").toISOString () == "2015-07-16T00:00:00.000Z");
+assert (new Date ("2015-07-16T11:29:05.023").toISOString () == "2015-07-16T11:29:05.023Z");
```
LaszloLango: Please move the new test cases to the end of the file and add test for TypeError too.

jerry-core/lit/lit-magic-strings.inc.h
```diff
@@ -219,6 +219,7 @@ LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_G_CHAR, "g")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_I_CHAR, "i")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_M_CHAR, "m")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_TIME_SEP_U, "T")
+LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_TIME_ZONE_U, "Z")
```
LaszloLango: "Z" means UTC time. I think the name should contains it. Maybe `LIT_MAGIC_STRING_UTC_TIME_ZONE_U`?

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -130,7 +130,12 @@ ecma_builtin_date_prototype_to_string (ecma_value_t this_arg) /**< this argument
     else
     {
       ecma_number_t milliseconds = ecma_date_ms_from_time (*prim_value_num_p);
-      ecma_string_t *output_str_p = ecma_new_ecma_string_from_number (milliseconds);
+      ecma_string_t *msec_str_p = ecma_new_ecma_string_from_number (milliseconds);
+      ecma_string_t *time_zone_str_p = ecma_get_magic_string (LIT_MAGIC_STRING_TIME_ZONE_U);
+
+      ecma_string_t *output_str_p = ecma_concat_ecma_strings (msec_str_p, time_zone_str_p);
+      ecma_deref_ecma_string (msec_str_p);
+      ecma_deref_ecma_string (time_zone_str_p);
```
LaszloLango: Isn't this enough?
```C++
ecma_string_t *output_str_p = ecma_get_magic_string (LIT_MAGIC_STRING__EMPTY);
ecma_date_insert_num_with_sep (&output_str_p, milliseconds, LIT_MAGIC_STRING_UTC_TIME_ZONE_U, 3);
```

rtakacs: You are right, it is simpler.

jerry-core/lit/lit-magic-strings.inc.h
rtakacs: Renamed.

tests/jerry/date-tostring.js
rtakacs: Done.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -92,16 +92,16 @@ ecma_date_insert_num_with_sep (ecma_string_t **str_p, /**< input/output string *
 } /* ecma_date_insert_num_with_sep */
 
 /**
- * The Date.prototype object's 'toString' routine
+ * The Date.prototype object's 'toISOString' routine
  *
  * See also:
- *          ECMA-262 v5, 15.9.5.2
+ *          ECMA-262 v5, 15.9.5.43
```
kkristof: Could we keep the order of the functions as the ECMA standard declare them?

szledan: All builtin function names are already declared in `ecma-builtin-date-prototype.inc.h`. So we don't need to change the position of these functions.

rtakacs: Yes, we could. Added forward declaration for *to_iso_string.

kkristof: Why?

rtakacs: Ahh sorry, the declaration is already in the header.

kkristof:
Looks good to me.

LaszloLango:
+1 lgtm

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -1207,7 +1148,66 @@ ecma_builtin_date_prototype_to_utc_string (ecma_value_t this_arg) /**< this argu
...
+    if (ecma_number_is_nan (*prim_value_num_p))
+    {
+      ecma_string_t *magic_str_p = ecma_get_magic_string (LIT_MAGIC_STRING_INVALID_DATE_UL);
+      ret_value = ecma_make_normal_completion_value (ecma_make_string_value (magic_str_p));
+    }
+    else
+    {
+      ecma_number_t milliseconds = ecma_date_ms_from_time (*prim_value_num_p);
+      ecma_string_t *output_str_p = ecma_get_magic_string (LIT_MAGIC_STRING__EMPTY);
+      ecma_date_insert_num_with_sep (&output_str_p, milliseconds, LIT_MAGIC_STRING_UTC_TIME_ZONE_U, 3);
```
ruben-ayrapetyan: > 15.9.5.2 Date.prototype.toString ( )

> This function returns a String value. The contents of the String are implementation-dependent, but are intended to represent the Date in the current time zone in a convenient, human-readable form.


> 15.9.5.43 Date.prototype.toISOString ( )

> This function returns a String value represent the instance in time represented by this Date object. The format of the String is the Date Time string format defined in 15.9.1.15. All fields are present in the String. The time zone is always UTC, denoted by the suffix Z. If the time value of this object is not a finite Number a RangeError exception is thrown.

As output of the `toString` and the `toISOString` are different only in timezone part (toISOString outputs in UTC timezone and toString should output in "current" timezone), maybe we should extract the code to a helper with arguments `is_current_timezone`, and add `TODO` comment about implementing output in "current" timezone mode.

ruben-ayrapetyan: So, the helper could be called from both `toString` and `toISOString` implementations.

rtakacs: You are right, I missed that the time zone is different from toString and toISOString. The helper is a good solution. The toUTCString could use it as well.

ruben-ayrapetyan:
Looks good to me

zherczeg:
LGTM

egavrin:
Merged 63083b3b51157b0f02d1170fc0296845f2bcf79f

Issue #410, opened 2015-07-15
Assertion 'day_within_year >= 0 && day_within_year < 365 + in_leap_year' failed in ecma_date_month_from_time


renatahodovan:
###### Jerry version:
    Checked revision: abc2b55
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
Date(Number.POSITIVE_INFINITY).toString();
```

###### Backtrace:
```javascript
ICE: Assertion 'day_within_year >= 0 && day_within_year < 365 + in_leap_year' failed at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp(ecma_date_month_from_time):269.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x00000000004a0a9e in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
32    SYSCALL_2
(gdb) bt
# 0  0x00000000004a0a9e in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
# 1  0x00000000004003c2 in syscall_2 (syscall_no=62, arg1=31760, arg2=6) at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:96
# 2  0x0000000000400563 in abort () at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:175
# 3  0x0000000000440580 in jerry_fatal(jerry_fatal_code_t) (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:65
# 4  0x00000000004405cf in jerry_assert_fail(char const*, char const*, char const*, unsigned int) (
    assertion=0x4bae78 "day_within_year >= 0 && day_within_year < 365 + in_leap_year", 
    file=0x4bae18 "jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp", 
    function=0x4a3e50 <_ZZ25ecma_date_month_from_timedE8__func__.lto_priv.809> "ecma_date_month_from_time", line=269)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:92
# 5  0x0000000000474bf7 in ecma_date_month_from_time (time=inf)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp:269
[...]

PR #421, opened 2015-07-16
Fix Date(Number.POSITIVE_INFINITY) assertion fail in the Date object

szledan:
JerryScript-DCO-1.0-Signed-off-by: Szilard Ledan szledan.u-szeged@partner.samsung.com

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
@@ -556,6 +556,11 @@ ecma_builtin_date_dispatch_construct (const ecma_value_t *arguments_list_p, /**<
 
   if (ecma_is_completion_value_empty (ret_value))
   {
+    if (!ecma_number_is_nan (*prim_value_num_p) && ecma_number_is_infinity (*prim_value_num_p))
+    {
+      *prim_value_num_p = ecma_number_make_nan ();
+    }
+
     ecma_property_t *class_prop_p = ecma_create_internal_property (obj_p,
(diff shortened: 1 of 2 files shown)


szledan:
Fix for #410

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -556,6 +556,11 @@ ecma_builtin_date_dispatch_construct (const ecma_value_t *arguments_list_p, /**<
 
   if (ecma_is_completion_value_empty (ret_value))
   {
+    if (!ecma_number_is_nan (*prim_value_num_p) && ecma_number_is_infinity (*prim_value_num_p))
```
ruben-ayrapetyan: If a number is `Infinity`, it cannot be `NaN`.

galpeter: But this tests that if it is **not** NaN and is infinity.

Also note that in the first line of the `ecma_number_is_infinity` there is an assert:
```
JERRY_ASSERT (!ecma_number_is_nan (num));
```

ruben-ayrapetyan: @galpeter, I see. Thank you for explanation.

ruben-ayrapetyan:
Looks good to me

kkristof:
Looks good to me.

galpeter:
Landed: https://github.com/Samsung/jerryscript/commit/bbfddea032fc622cef6279b308f5d77e135dd19b

Issue #355, opened 2015-07-10
unittests/unit-test-date-helpers fail for 32bit 


seanshpark:
When I do pre-commit test I get
```
make[1]: Entering directory `/.../harmony/jerryscript'
Running ./build/bin/unittests/unit-test-api... OK
Running ./build/bin/unittests/unit-test-date-helpers... FAILED
Unit tests run failed. See ./build/bin/unittests/unit_tests_run.log for details.
make[1]: *** [unittests_run] Error 1
```
`./build/bin/unittests/unit_tests_run.log` says
```
ICE: Assertion 'ecma_date_day (MS_PER_DAY) == 1' failed at /.../harmony/jerryscript/tests/unit/test-date-helpers.cpp(main):32.
Error: ERR_FAILED_INTERNAL_ASSERTION
```

is this a problem for only me or a bug? or duplicate issue?

egavrin:
Could you please share your environment?

On dbf3708255217f360a72c02fadb185a76e5fbf67 :
```bash
$ make unittests_run 
Running ./build/bin/unittests/unit-test-api... OK
Running ./build/bin/unittests/unit-test-date-helpers... OK
Running ./build/bin/unittests/unit-test-heap... OK
Running ./build/bin/unittests/unit-test-literal-storage... OK
Running ./build/bin/unittests/unit-test-longjmp... OK
Running ./build/bin/unittests/unit-test-number-to-integer... OK
Running ./build/bin/unittests/unit-test-number-to-string... OK
Running ./build/bin/unittests/unit-test-parser... OK
Running ./build/bin/unittests/unit-test-pool... OK
Running ./build/bin/unittests/unit-test-poolman... OK
Running ./build/bin/unittests/unit-test-recordset... OK
Running ./build/bin/unittests/unit-test-strings... OK
Running ./build/bin/unittests/unit-test-string-to-number... OK
```

```bash
$ make unittests_run TOOLCHAIN="./build/configs/toolchain_linux_armv7l-el.cmake"
Running ./build/bin/unittests/unit-test-api... OK
Running ./build/bin/unittests/unit-test-date-helpers... OK
Running ./build/bin/unittests/unit-test-heap... OK
Running ./build/bin/unittests/unit-test-literal-storage... OK
Running ./build/bin/unittests/unit-test-longjmp... OK
Running ./build/bin/unittests/unit-test-number-to-integer... OK
Running ./build/bin/unittests/unit-test-number-to-string... OK
[...]

PR #425, opened 2015-07-16
Fix endian detection on 32bit

galpeter:
When using the -std=c99 the gcc does not defines the `i386` macro
just the `__i386` and `__i386__`.

Fixes issue #355

### DIFF
--- a/third-party/fdlibm/include/fdlibm.h
+++ b/third-party/fdlibm/include/fdlibm.h
@@ -13,7 +13,8 @@
 /* Sometimes it's necessary to define __LITTLE_ENDIAN explicitly
    but these catch some common cases. */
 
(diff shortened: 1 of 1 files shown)


seanshpark:
I got pre-commit success on 32bit Ubuntu. Thank you. #355 is ok.
The patch is related with endian detection from cpu architecture but I don't think it will work when change to big endian in arm with `-mbig-endian` option.
For now, I think it's ok to go and post an issue about this so later when we face this we can fix it.

zherczeg:
LGTM

egavrin:
Good to me

galpeter:
merged: https://github.com/Samsung/jerryscript/commit/021fc626526a141a4deae4f2e60669546209cf2c

PR #427, opened 2015-07-17
Fix leak in Array.prototype.indexOf() 

dbatyai:
The leak occurs when `fromIndex` can't coerce to primitive value.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
@@ -1881,12 +1881,11 @@ ecma_builtin_array_prototype_object_index_of (ecma_value_t this_arg, /**< this a
   /* 3. */
   uint32_t len = ecma_number_to_uint32 (len_number);
 
-  ecma_number_t *num_p = ecma_alloc_number ();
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -951,6 +950,9 @@ ecma_builtin_array_prototype_object_index_of (ecma_value_t this_arg, /**< this a
     /* 5. */
     ECMA_OP_TO_NUMBER_TRY_CATCH (arg_from_idx, arg2, ret_value);
 
+    ecma_number_t *num_p = ecma_alloc_number ();
+    *num_p = ecma_int32_to_number (-1);
```
ruben-ayrapetyan: The lines could be moved to `if` block at line 991, and here we can define variable of `ecma_number_t`, not pointer.
In the case, call to `ecma_dealloc_number` at line 995 could be removed.

ruben-ayrapetyan:
Looks good to me

zherczeg:
+1 LGTM

egavrin:
Merged 27161d31df551bbf3dc69b119a26e41e011a8f94

Issue #428, opened 2015-07-17
Segmentation fault in ecma_builtin_global_object_parse_int


renatahodovan:
###### Jerry version:
    Checked revision: 6038173
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
parseInt('');
```

###### Backtrace:
```javascript
Program received signal SIGSEGV, Segmentation fault.
0x000000000048dba2 in ecma_builtin_global_object_parse_int (this_arg=4, string=118, radix=4)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp:121
121   utf8_string_buff[str_size] = LIT_BYTE_NULL;
(gdb) bt
# 0  0x000000000048dba2 in ecma_builtin_global_object_parse_int (this_arg=4, string=118, radix=4)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp:121
# 1  0x000000000048d9a2 in ecma_builtin_global_dispatch_routine (builtin_routine_id=48, this_arg_value=4, 
    arguments_list=0x812b50 <_ZL13mem_heap_area.lto_priv.1058+262032>, arguments_number=1)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:232
# 2  0x00000000004961ff in ecma_builtin_dispatch_routine (builtin_object_id=ECMA_BUILTIN_ID_GLOBAL, builtin_routine_id=48, this_arg_value=4, 
    arguments_list=0x812b50 <_ZL13mem_heap_area.lto_priv.1058+262032>, arguments_number=1)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.inc.h:312
# 3  0x0000000000495601 in ecma_builtin_dispatch_call (obj_p=0x7d2c70 <_ZL13mem_heap_area.lto_priv.1058+176>, this_arg_value=4, 
    arguments_list_p=0x812b50 <_ZL13mem_heap_area.lto_priv.1058+262032>, arguments_list_len=1)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.cpp:414
# 4  0x000000000044c8cf in ecma_op_function_call (func_obj_p=0x7d2c70 <_ZL13mem_heap_area.lto_priv.1058+176>, this_arg_value=4, 
[...]

PR #430, opened 2015-07-17
Fix parseInt when passing empty string

galpeter:
Corrected the base buffer size for the string.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
@@ -113,147 +113,158 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
   ecma_string_t *number_str_p = ecma_get_string_from_value (string_var);
   lit_utf8_size_t str_size = ecma_string_get_size (number_str_p);
 
-  MEM_DEFINE_LOCAL_ARRAY (utf8_string_buff, str_size, lit_utf8_byte_t);
-
-  ssize_t bytes_copied = ecma_string_to_utf8_string (number_str_p,
-                                                     utf8_string_buff,
-                                                     (ssize_t) str_size);
-  JERRY_ASSERT (bytes_copied >= 0);
-  utf8_string_buff[str_size] = LIT_BYTE_NULL;
-
-  /* 2. Remove leading whitespace. */
-  ecma_length_t start = str_size;
-  ecma_length_t end = str_size;
-  for (ecma_length_t i = 0; i < end; i++)
+  if (str_size > 0)
   {
-    if (!lit_char_is_white_space (utf8_string_buff[i])
-        && !lit_char_is_line_terminator (utf8_string_buff[i]))
+    MEM_DEFINE_LOCAL_ARRAY (utf8_string_buff, str_size, lit_utf8_byte_t);
+
+    ssize_t bytes_copied = ecma_string_to_utf8_string (number_str_p,
+                                                       utf8_string_buff,
+                                                       (ssize_t) str_size);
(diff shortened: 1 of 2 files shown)


galpeter:
Related issue: #428

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -112,7 +112,7 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
   ecma_string_t *number_str_p = ecma_get_string_from_value (string_var);
   lit_utf8_size_t str_size = ecma_string_get_size (number_str_p);
 
-  MEM_DEFINE_LOCAL_ARRAY (utf8_string_buff, str_size, lit_utf8_byte_t);
+  MEM_DEFINE_LOCAL_ARRAY (utf8_string_buff, str_size + 1, lit_utf8_byte_t);
```
ruben-ayrapetyan: @galpeter, as string is empty in the test case from #428, maybe we don't need to process it. So, instead of `+1` we could put the code into `if` block with fastpath for empty string.

galpeter: @ruben-ayrapetyan, yeah that's why I was thinking initially, thus I have this: https://github.com/galpeter/jerryscript/commit/f3081a005fd91be09b151f725a23185489c05b3c  would that be ok?

ruben-ayrapetyan: @galpeter, yes, I think this would be ok.

galpeter:
@ruben-ayrapetyan, I've updated the PR, and now it uses `if` check.

ruben-ayrapetyan:
Looks good to me

dbatyai:
LGTM

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -113,147 +113,157 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
...
+        }
       }
-      else
+
+      /* 11. Check if characters are in [0, Radix - 1]. We also convert them to number values in the process. */
+      for (lit_utf8_size_t i = start; i < end; i++)
       {
-        /* Not a valid number char, set value to radix so it fails to pass as a valid character. */
-        utf8_string_buff[i] = (lit_utf8_byte_t) rad;
+        if ((utf8_string_buff[i]) >= 'a' && utf8_string_buff[i] <= 'z')
```
zherczeg: Literals has constants (defines) for these characters. Please use them.

zherczeg:
Please fix the above, and LGTM after that.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
galpeter: @zherczeg, I don't think it is wise to replace the with the equivalent `LIT_` define, as those are `ecma_char_t` types (`unit16_t`), but here we use `lit_utf8_byte_t` (`unit_8_t`). Should I really replace the constants?

galpeter: @zherczeg, I've got a follow up fix, which rewrites the parseInt to use the lit_iterators to support unicode characters and there the constants are fixed. So should I also include the constants in this PR?

galpeter:
@zherczeg, I've replaced the character constants with the `LIT_..` macros.

galpeter:
Landed: https://github.com/Samsung/jerryscript/commit/ee8d65063fd14cb5d15f11841666dd8be5807a0f

Issue #440, opened 2015-07-21
Regression after #397 has been landed


kkristof:
The following test can reproduce the problem:
```
(function () {
  var d = {};
  var a = [arguments];
})();
```
This now throws `ReferenceError`.
It also makes 29 test262 tests to fail.

ruben-ayrapetyan:
@kkristof, thank you for raising the issue.

In my environment, ReferenceError is also raised at 005eb044a93b5ecaf7ac385c8fa9c866e3554b38 (just before #397 - abc2b55297eff13538fca2440d127ba79cdcc8b3 - was merged).

So, maybe, the issue is not connected with the #397 patch.
Could you, please, check this in your environment?

ruben-ayrapetyan:
@kkristof, with the following patch test finishes successfully in my environment.

Could you, please, also check the following patch for this and other 29 failing tests in your environment, and list tests that are fixed with the patch?

Thank you in advance,

```
diff --git a/jerry-core/parser/js/parser.cpp b/jerry-core/parser/js/parser.cpp
index 082d8d6..6f60d73 100644
--- a/jerry-core/parser/js/parser.cpp
+++ b/jerry-core/parser/js/parser.cpp
@@ -2949,9 +2949,18 @@ preparse_scope (bool is_global)
     }
   }
 
-  while (!token_is (end_tt))
+  size_t nesting_level = 0;
+  while (nesting_level > 0 || !token_is (end_tt))
   {
-    if (token_is (TOK_NAME))
+    if (token_is (TOK_OPEN_BRACE))
+    {
+      nesting_level++;
+    }
+    else if (token_is (TOK_CLOSE_BRACE))
+    {
+      nesting_level--;
+    }
+    else if (token_is (TOK_NAME))
     {
       if (lit_literal_equal_type_cstr (lit_get_literal_by_cp (token_data_as_lit_cp ()), "arguments"))
       {
```

ruben-ayrapetyan:
Failure caused by missing check for current nesting level during search for identifier tokens.

kkristof:
Sorry, the minimized example was wrong, The original test case was the following:
```
(function () {
    function getFunc() {
        return 10;
    }
    Object.defineProperty(arguments, "0", {
        get: getFunc,
        enumerable: true,
        configurable: false
    });
    try {
        Object.defineProperty(arguments, "0", {
            configurable: true
        });
    } catch (e) {
        return e instanceof TypeError;
    }
    return false;
[...]

PR #442, opened 2015-07-21
Fix preparse_scope - consider nesting level during search for identifier tokens

ruben-ayrapetyan:
Related issue: #440


kkristof:
Looks good to me.

egavrin:
Good to me.

Issue #424, opened 2015-07-16
Remove usage of isalpha and isdigit from JSON built-in's implementation


ruben-ayrapetyan:
Hello.

Currently, implementation of JSON built-in uses isalpha and isdigit functions.
Instead, helpers from lit_char should be used.

After the replacement is performed, `isalpha` and `isdigit` functions should be removed.

akiss77:
There is a name clash with `isalpha` if compiler-default library is used for the build (on x86_64 Ubuntu 14.04.2 LTS with gcc 4.8.4). Is it worth reporting/fixing the issue (only 4 lines of changes) or should I wait for the replacement?

zherczeg:
Fix: PR #484

akiss77:
Fix landed, closing issue

PR #444, opened 2015-07-21
Add prefix to JSON builtin helper functions

akiss77:
`isalpha` in ecma-builtin-json.cpp can cause a name clash if the
build uses compiler-default libc. Thus, adding `ecma_builtin_json_`
prefix to all helper functions there.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
@@ -41,21 +41,22 @@
 
 /*
  * FIXME:
- *       Replace usage of isalpha and isdigit functions in the module with lit_char helpers and remove the functions.
+ *       Replace usage of ecma_builtin_json_isalpha and ecma_builtin_json_isdigit functions in the module with lit_char
+ *       helpers and remove the functions.
  *
  *       Related issue: #424
  */
 
 /** Checks for an alphabetic character. */
 static int
-isalpha (int c)
+ecma_builtin_json_isalpha (int c)
 {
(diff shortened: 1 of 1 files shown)


ruben-ayrapetyan:
Looks good to me

akiss77:
Related to  #424 (but does not fix it, keeps the issue around)

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -41,21 +41,22 @@
...
+ *       Replace usage of ecma_builtin_json_isalpha and ecma_builtin_json_isdigit functions in the module with lit_char
+ *       helpers and remove the functions.
  *
  *       Related issue: #424
  */
 
 /** Checks for an alphabetic character. */
 static int
-isalpha (int c)
+ecma_builtin_json_isalpha (int c)
```
zherczeg: There are a bunch of similar functions in lit-char-helpers.cpp such as lit_char_is_decimal_digit. Why don't we move everything to there?

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -41,21 +41,22 @@
...
 {
   return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z');
```
zherczeg: And replace these with LIT_ constants.

akiss77:
Offline communication concluded that this temporary rewrite wont be landed, only the lit_char helper-based final version once it's ready. Closing PR.

PR #448, opened 2015-07-22
Regexp unicode dev

LaszloLango:
JerryScript-DCO-1.0-Signed-off-by: Szilard Ledan szledan.u-szeged@partner.samsung.com

### DIFF
--- a/jerry-core/ecma/operations/ecma-regexp-object.cpp
+++ b/jerry-core/ecma/operations/ecma-regexp-object.cpp
@@ -75,17 +75,16 @@ re_parse_regexp_flags (ecma_string_t *flags_str_p, /**< Input string with flags 
 {
   ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
 
-  FIXME ("Unicode: properly process non-ascii characters.");
   lit_utf8_size_t flags_str_size = ecma_string_get_size (flags_str_p);
   MEM_DEFINE_LOCAL_ARRAY (flags_start_p, flags_str_size, lit_utf8_byte_t);
 
   ecma_string_to_utf8_string (flags_str_p, flags_start_p, (ssize_t) flags_str_size);
-  lit_utf8_byte_t *flags_char_p = flags_start_p;
+  lit_utf8_iterator_t iter = lit_utf8_iterator_create (flags_start_p, flags_str_size);
 
-  while (flags_char_p < flags_start_p + flags_str_size
+  while (!lit_utf8_iterator_is_eos (&iter)
          && ecma_is_completion_value_empty (ret_value))
   {
-    switch (*flags_char_p)
+    switch (lit_utf8_iterator_read_next (&iter))
     {
       case 'g':
       {
@@ -120,7 +119,6 @@ re_parse_regexp_flags (ecma_string_t *flags_str_p, /**< Input string with flags 
         break;
       }
     }
-    flags_char_p++;
   }
 
   MEM_FINALIZE_LOCAL_ARRAY (flags_start_p);
@@ -232,48 +230,6 @@ ecma_op_create_regexp_object (ecma_string_t *pattern_p, /**< input pattern */
 } /* ecma_op_create_regexp_object */
 
 /**
- * Backtrack a unicode character
- */
-static const lit_utf8_byte_t *
-re_utf8_backtrack (const lit_utf8_byte_t *str_p)
-{
-  /* FIXME: change to string iterator with unicode support, when it would be implemented */
-  return --str_p;
-} /* re_utf8_backtrack */
-
-/**
- * Helper to get an input character and increase string pointer.
- */
-static ecma_char_t
-re_get_input_char (const lit_utf8_byte_t **char_p)
-{
-  /* FIXME: change to string iterator with unicode support, when it would be implemented */
-  const lit_utf8_byte_t ch = **char_p;
-  (*char_p)++;
-  return ch;
-} /* re_get_input_char */
-
-/**
- * Helper to get current input character, won't increase string pointer.
- */
-static ecma_char_t
-re_lookup_input_char (const lit_utf8_byte_t *str_p)
-{
-  /* FIXME: change to string iterator with unicode support, when it would be implemented */
-  return *str_p;
-} /* re_lookup_input_char */
-
-/**
- * Helper to get previous input character, won't decrease string pointer.
- */
-static ecma_char_t
-re_lookup_prev_char (const lit_utf8_byte_t *str_p)
-{
-  /* FIXME: change to string iterator with unicode support, when it would be implemented */
-  return *(--str_p);
-} /* re_lookup_prev_char */
-
-/**
  * Recursive function for RegExp matching. Tests for a regular expression
  * match and returns a MatchResult value.
  *
@@ -286,8 +242,8 @@ re_lookup_prev_char (const lit_utf8_byte_t *str_p)
 static ecma_completion_value_t
 re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
                  re_bytecode_t *bc_p, /**< pointer to the current RegExp bytecode */
-                 const lit_utf8_byte_t *str_p, /**< pointer to the current input character */
-                 const lit_utf8_byte_t **res_p) /**< pointer to the matching substring */
+                 lit_utf8_iterator_t iter, /**< input string iterator */
+                 lit_utf8_iterator_t *out_iter_p) /**< Output: matching substring iterator */
 {
   ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
   re_opcode_t op;
@@ -313,37 +269,52 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
       case RE_OP_MATCH:
       {
         JERRY_DDLOG ("Execute RE_OP_MATCH: match\n");
-        *res_p = str_p;
+        *out_iter_p = iter;
         re_ctx_p->recursion_depth--;
         ret_value = ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_TRUE);
         return ret_value; /* match */
       }
       case RE_OP_CHAR:
       {
-        uint32_t ch1 = re_get_value (&bc_p);
-        uint32_t ch2 = re_get_input_char (&str_p);
+        if (lit_utf8_iterator_is_eos (&iter))
+        {
+          re_ctx_p->recursion_depth--;
+          return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_FALSE); /* fail */
+        }
+
+        ecma_char_t ch1 = (ecma_char_t) re_get_value (&bc_p);
+        ecma_char_t ch2 = lit_utf8_iterator_read_next (&iter);
         JERRY_DDLOG ("Character matching %d to %d: ", ch1, ch2);
 
-        if (ch2 == '\0' || ch1 != ch2)
+        if (ch1 != ch2)
         {
           JERRY_DDLOG ("fail\n");
           re_ctx_p->recursion_depth--;
           return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_FALSE); /* fail */
         }
+
         JERRY_DDLOG ("match\n");
+
         break; /* tail merge */
       }
       case RE_OP_PERIOD:
       {
-        uint32_t ch1 = re_get_input_char (&str_p);
-        JERRY_DDLOG ("Period matching '.' to %d: ", ch1);
+        if (lit_utf8_iterator_is_eos (&iter))
+        {
+          re_ctx_p->recursion_depth--;
+          return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_FALSE); /* fail */
+        }
 
-        if (ch1 == '\n' || ch1 == '\0')
+        ecma_char_t ch = lit_utf8_iterator_read_next (&iter);
+        JERRY_DDLOG ("Period matching '.' to %d: ", (uint32_t) ch);
+
+        if (lit_char_is_line_terminator (ch))
         {
           JERRY_DDLOG ("fail\n");
           re_ctx_p->recursion_depth--;
           return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_FALSE); /* fail */
         }
+
         JERRY_DDLOG ("match\n");
         break; /* tail merge */
       }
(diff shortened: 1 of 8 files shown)


jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -365,18 +365,20 @@ re_parse_char_class (re_parser_ctx_t *parser_ctx_p, /**< number of classes */
...
-        append_char_class (re_ctx_p, 0x003AUL, 0xFFFFUL);
+        append_char_class (re_ctx_p, 0x003AUL, LIT_UTF16_CODE_UNIT_MAX);
         ch = RE_CHAR_UNDEF;
       }
       else if (ch == 's')
       {
+        /* See ECMA-262 v5, 15.10.2.12 */
         append_char_class (re_ctx_p, 0x0009UL, 0x000DUL);
         append_char_class (re_ctx_p, 0x0020UL, 0x0020UL);
         append_char_class (re_ctx_p, 0x00A0UL, 0x00A0UL);
```
ruben-ayrapetyan: @LaszloLango, could you, please, describe how 15.10.2.12 defines / lists this ranges?

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -19,31 +19,30 @@
...
+    else
+    {
+      ch = 0;
+      break;
+    }
+  }
+
   return ch;
-} /* re_get_ecma_char */
+}
```
ruben-ayrapetyan: Comments for the function are missing.

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -286,8 +242,8 @@ re_lookup_prev_char (const lit_utf8_byte_t *str_p)
 static ecma_completion_value_t
 re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
                  re_bytecode_t *bc_p, /**< pointer to the current RegExp bytecode */
-                 const lit_utf8_byte_t *str_p, /**< pointer to the current input character */
-                 const lit_utf8_byte_t **res_p) /**< pointer to the matching substring */
+                 lit_utf8_iterator_t iter, /**< pointer to the current input character */
```
ruben-ayrapetyan: After the change the argument is not a pointer.

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -286,8 +242,8 @@ re_lookup_prev_char (const lit_utf8_byte_t *str_p)
...
+                 lit_utf8_iterator_t *res_iter_p) /**< pointer to the matching substring */
```
ruben-ayrapetyan: This is not pointer to a string too.

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -573,34 +545,33 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
...
           break; /* capture is 'undefined', always matches! */
         }
 
-        sub_str_p = re_ctx_p->saved_p[backref_idx];
+        lit_utf8_iterator_t sub_iter = re_ctx_p->saved_p[backref_idx];
 
-        while (sub_str_p < re_ctx_p->saved_p[backref_idx + 1])
+        while (sub_iter.buf_pos.offset < re_ctx_p->saved_p[backref_idx + 1].buf_pos.offset)
         {
           uint32_t ch1, ch2;
```
ruben-ayrapetyan: Could we use `ecma_char_t` for the variables?

jerry-core/parser/regexp/re-parser.cpp
LaszloLango: @ruben-ayrapetyan,
"The production CharacterClassEscape :: s evaluates by returning the set of characters containing the characters that are on the right-hand side of the WhiteSpace (7.2) or LineTerminator (7.3) productions."

LaszloLango:
@ruben-ayrapetyan, I've updated the PR. please check.

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -313,15 +269,21 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
...
       {
+        if (lit_utf8_iterator_is_eos (&iter))
+        {
+          re_ctx_p->recursion_depth--;
+          return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_FALSE); /* fail */
+        }
+
         uint32_t ch1 = re_get_value (&bc_p);
-        uint32_t ch2 = re_get_input_char (&str_p);
+        uint32_t ch2 = (uint32_t) lit_utf8_iterator_read_next (&iter);
```
zherczeg: thse should be ecma_char_t

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -314,69 +359,87 @@ re_parse_char_class (re_parser_ctx_t *parser_ctx_p, /**< number of classes */
...
+        if (lit_utf8_iterator_is_eos (&(parser_ctx_p->iter)))
         {
-          ret_value = ecma_raise_syntax_error ("invalid regexp control escape");
+          ret_value = ecma_raise_syntax_error ("invalid character class");
           return ret_value;
         }
+
+        ch = lit_utf8_iterator_read_next (&(parser_ctx_p->iter));
+
+        if ((ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z'))
```
zherczeg: We should use the LIT_ constants here

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -563,102 +637,123 @@ re_parse_next_token (re_parser_ctx_t *parser_ctx_p, /**< RegExp parser context *
...
+
         out_token_p->type = RE_TOK_CHAR;
-        /* FIXME: get unicode char from digits */
-        /* result.value = ...; */
+        out_token_p->value = codeunit;
       }
       else if (ch1 == 'd')
       {
-        advance = 2;
+        advance = 0;
```
zherczeg: Do we still use advance for anything?

zherczeg: Can't we set the default value to 0? We could get rid of these sets.

LaszloLango: In the line 625 you can see that advance is set to 2. If set it to 0 there, then there will be a lot more sets to 2 in other cases.

LaszloLango: I think I can simplify this a bit. Let me check.

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -403,11 +467,12 @@ re_parse_char_class (re_parser_ctx_t *parser_ctx_p, /**< number of classes */
         append_char_class (re_ctx_p, 0x2030UL, 0x205EUL);
         append_char_class (re_ctx_p, 0x2060UL, 0x2FFFUL);
         append_char_class (re_ctx_p, 0x3001UL, 0xFEFEUL);
```
ruben-ayrapetyan: @LaszloLango, if I understand correctly, these ranges uniquely correspond to some range identifier (`'S'` in the case). Maybe, we shouldn't put the ranges to context, but instead put only unique range identifier, and then use the identifier during range checking. What do you think about this?

LaszloLango: @ruben-ayrapetyan, you are right. I can change some of these that exists in `lit-char-helpers.h`. What about the missing ones? Where should I define them? In the `lit-char-helpers.h` or re-`parser.cpp`?

ruben-ayrapetyan: @LaszloLango, I mean, we should not put actual range values to `re_ctx_p`, instead we could just put identifier of range, like `'S'`, and in `re_match_regexp` we then could use the `lit_char_is_white_space` and `lit_char_is_space_separator`-like interfaces.

I think, it is better to define missing interfaces in lit-char-helpers.h. Could you, please, clarify list of missing interfaces / ranges?

LaszloLango:
@zherczeg, @ruben-ayrapetyan, I've updated the PR. Please check.

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -271,6 +277,39 @@ re_count_num_of_groups (re_parser_ctx_t *parser_ctx_p) /**< RegExp parser contex
...
 /**
+ * Helper function for \x and \u. Read hexa or decimal numbers and convert to
+ * unicode codeunit.
+ *
+ * @return completion value
+ *         Returned value must be freed with ecma_free_completion_value
+ */
+static ecma_completion_value_t
+re_get_codeunit_from_string (lit_utf8_iterator_t *iter, /**< input string iterator*/
+                             ecma_char_t *codeunit, /**< Output: read codeunit */
```
ruben-ayrapetyan: Please, add `_p` suffixes to pointer variables, and `out_` prefixes to output arguments.

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -271,6 +277,39 @@ re_count_num_of_groups (re_parser_ctx_t *parser_ctx_p) /**< RegExp parser contex
...
+                             bool is_hexa) /**< true: read hexa, false: read unicode */
+{
+  ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
+  *codeunit = LIT_UNICODE_CODE_POINT_NULL;
+  uint32_t base = is_hexa ? 16u : 10u;
+
+  for (int i = 0; i < (is_hexa ? 2 : 4); i++)
```
ruben-ayrapetyan: Maybe, `lit_read_code_point_from_hex` could be used for the conversion.

jerry-core/parser/regexp/re-parser.cpp
LaszloLango: @ruben-ayrapetyan, I see. For a RegExp like `/\s/.exec(" ")` you can do that, but you cannot not when it is used in a character class like `/[\s]/.exec(" ")`. So we cannot eliminate this part easily.

ruben-ayrapetyan: @LaszloLango, I see.

Maybe, this can be optimized in future. Could you, please, raise issue about this?

By the way, why do we need to use callback for `append_char_class`?

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -304,129 +349,152 @@ re_parse_char_class (re_parser_ctx_t *parser_ctx_p, /**< number of classes */
...
+          ret_value = ecma_raise_syntax_error ("invalid character class");
           return ret_value;
         }
+
+        ch = lit_utf8_iterator_read_next (&(parser_ctx_p->iter));
+
+        if ((ch >= LIT_CHAR_ASCII_UPPERCASE_LETTERS_BEGIN && ch <= LIT_CHAR_ASCII_UPPERCASE_LETTERS_END)
+            || (ch >= LIT_CHAR_ASCII_LOWERCASE_LETTERS_BEGIN && ch <= LIT_CHAR_ASCII_LOWERCASE_LETTERS_END))
+        {
+          ch = (ch % 32);
```
ruben-ayrapetyan: Could you, please, clarify why `32`?

LaszloLango: http://www.ecma-international.org/ecma-262/5.1/#sec-15.10.2.10, point 3

zherczeg:
LGTM

jerry-core/parser/regexp/re-parser.cpp
ruben-ayrapetyan: Thank you. Please, add comment with reference to the specification paragraph.

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -286,8 +242,8 @@ re_lookup_prev_char (const lit_utf8_byte_t *str_p)
 static ecma_completion_value_t
 re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
                  re_bytecode_t *bc_p, /**< pointer to the current RegExp bytecode */
-                 const lit_utf8_byte_t *str_p, /**< pointer to the current input character */
-                 const lit_utf8_byte_t **res_p) /**< pointer to the matching substring */
+                 lit_utf8_iterator_t iter, /**< input string iterator */
+                 lit_utf8_iterator_t *res_iter_p) /**< matching substring iterator */
```
ruben-ayrapetyan: Is it output parameter? If so, please, update naming and comment accordingly.

jerry-core/parser/regexp/re-parser.cpp
```diff
@@ -271,6 +277,53 @@ re_count_num_of_groups (re_parser_ctx_t *parser_ctx_p) /**< RegExp parser contex
...
+
+    lit_utf8_iterator_advance (iter_p, 2);
+    *out_codeunit_p = (ecma_char_t) code_point;
+  }
+  else
+  {
+    for (int i = 0; i < 4; i++)
+    {
+      if (lit_utf8_iterator_is_eos (iter_p)
+          || (!is_hexa && !lit_char_is_decimal_digit (re_lookup (*iter_p, 0))))
```
ruben-ayrapetyan: Couldn't we use `lit_read_code_point_from_hex` here too?

jerry-core/ecma/operations/ecma-regexp-object.cpp
LaszloLango: sure

jerry-core/parser/regexp/re-parser.cpp
ruben-ayrapetyan: Seems that `is_hexa` is not changed in the loop and so is always `false`.

jerry-core/parser/regexp/re-parser.cpp
LaszloLango: ok

jerry-core/parser/regexp/re-parser.cpp
ruben-ayrapetyan: Shouldn't hex characters be checked for `\u....`?

LaszloLango: Sorry, I left `is_hexa` there. After `\u` only four decimal characters are acceptable.

jerry-core/parser/regexp/re-parser.cpp
LaszloLango: Raised issue: #468
Currently we need it because the same reason. We handle `/\s/` always as `/[\s]/`.

LaszloLango:
@ruben-ayrapetyan, I've updated the patch and raised #468 and #469 issues.

ruben-ayrapetyan:
Looks good to me

Issue #414, opened 2015-07-15
Assertion 'ecma_number_is_negative (ret_num)' failed in ecma_builtin_math_object_max


renatahodovan:
###### Jerry version:
    Checked revision: abc2b55
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
Math.max(Number.POSITIVE_INFINITY, 234);
```

###### Backtrace:
```javascript
ICE: Assertion 'ecma_number_is_negative (ret_num)' failed at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-math.cpp(ecma_builtin_math_object_max):367.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x00000000004ab0d7 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
32    SYSCALL_2
(gdb) bt
# 0  0x00000000004ab0d7 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
# 1  0x00000000004003c2 in syscall_2 (syscall_no=62, arg1=1176, arg2=6) at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:96
# 2  0x0000000000400563 in abort () at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:175
# 3  0x0000000000444144 in jerry_fatal(jerry_fatal_code_t) (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.cpp:65
# 4  0x0000000000444193 in jerry_assert_fail(char const*, char const*, char const*, unsigned int) (assertion=0x4b9850 "ecma_number_is_negative (ret_num)", 
    file=0x4b9730 "jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-math.cpp", 
    function=0x4b2dd0 <_ZZL28ecma_builtin_math_object_maxjPKjjE8__func__.lto_priv.388> "ecma_builtin_math_object_max", line=367) at jerryscript/jerry-core/jrt/jrt-fatals.cpp:92
[...]

PR #452, opened 2015-07-22
Improve Infinity handling in Math.min/max methods

galpeter:
In Math.max case: if we already found an infinity value update the result
only when the previous value was a negative infinity.

In Math.min case: if we already found an infinity value update the result
only when the previous value was a positive infinity.


galpeter:
Fixes issue: #414

dbatyai:
lgtm

egavrin:
Good to me

galpeter:
Landed: https://github.com/Samsung/jerryscript/commit/b5de03a808cb01eec91fb4273e10da3989a746e6

PR #454, opened 2015-07-22
Build system changes to allow compiler toolchain override

akiss77:
E.g., works with Clang as
```
make clean && make debug.linux CC=clang-3.5 CXX=clang++-3.5 AR=/usr/bin/ar RANLIB=/usr/bin/ranlib \
  EXTRA_CFLAGS="-Wno-unknown-warning-option -Wno-sign-conversion -Wno-sign-compare -Wno-nested-anon-types -Wno-c99-extensions -Wno-float-conversion -Wno-null-conversion -Wno-invalid-noreturn" \
  LTO=OFF
```

(Works on x86_64/linux at least.)

### DIFF
--- a/build/configs/toolchain_linux_x86_64.cmake
+++ b/build/configs/toolchain_linux_x86_64.cmake
@@ -15,9 +15,15 @@
 set(CMAKE_SYSTEM_NAME Linux)
 set(CMAKE_SYSTEM_PROCESSOR x86_64)
 
-find_program(CMAKE_C_COMPILER NAMES x86_64-linux-gnu-gcc x86_64-unknown-linux-gnu-gcc)
-find_program(CMAKE_CXX_COMPILER NAMES x86_64-linux-gnu-g++ x86_64-unknown-linux-gnu-g++)
+if(NOT DEFINED CMAKE_C_COMPILER)
+  find_program(CMAKE_C_COMPILER NAMES x86_64-linux-gnu-gcc x86_64-unknown-linux-gnu-gcc)
+endif()
+if(NOT DEFINED CMAKE_CXX_COMPILER)
+  find_program(CMAKE_CXX_COMPILER NAMES x86_64-linux-gnu-g++ x86_64-unknown-linux-gnu-g++)
+endif()
 # FIXME: This could break cross compilation, when the strip is not for the target architecture
 find_program(CMAKE_STRIP NAMES x86_64-linux-gnu-strip x86_64-unknown-linux-gnu-strip strip)
 
-set(FLAGS_COMMON_ARCH -ffixed-rbp)
+if(CMAKE_COMPILER_IS_GNUCC AND CMAKE_COMPILER_IS_GNUCXX)
+  set(FLAGS_COMMON_ARCH -ffixed-rbp)
+endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,7 +17,7 @@ project (Jerry CXX C ASM)
 
 # Require g++ of version >= 4.7.0
  if(NOT CMAKE_COMPILER_IS_GNUCXX)
-  message(FATAL_ERROR "g++ compiler is required")
+  message(WARNING "g++ compiler is required")
  else()
   execute_process(COMMAND ${CMAKE_CXX_COMPILER} -dumpversion
                   OUTPUT_VARIABLE GNU_CXX_VERSION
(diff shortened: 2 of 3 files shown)


akiss77:
Also enables the compilation with afl-fuzzer's wrapped gcc/g++ ( #333 ) but link fails because generated code depends on functions available in the compiler-default libc only.

Blocked by #424

build/configs/toolchain_linux_x86_64.cmake
```diff
@@ -15,9 +15,15 @@
 set(CMAKE_SYSTEM_NAME Linux)
 set(CMAKE_SYSTEM_PROCESSOR x86_64)
 
-find_program(CMAKE_C_COMPILER NAMES x86_64-linux-gnu-gcc x86_64-unknown-linux-gnu-gcc)
-find_program(CMAKE_CXX_COMPILER NAMES x86_64-linux-gnu-g++ x86_64-unknown-linux-gnu-g++)
+if(NOT DEFINED CMAKE_C_COMPILER)
+  find_program(CMAKE_C_COMPILER NAMES x86_64-linux-gnu-gcc x86_64-unknown-linux-gnu-gcc)
```
ruben-ayrapetyan: @akiss77, maybe, we should introduce some set of `toolchain_*.cmake` files for different compilers. The files would set compiler executable names and also apply compiler-specific values of build options (for example `FLAGS_LTO=-flto` for gcc / g++) / warnings configuration etc.
Also, maybe we should extract the architecture-specific part to sub-configuration files for each supported compiler-architecture pair and put compiler-architecture specific flags / options to the files.

In the case, we could remove gcc / g++ specific `if` blocks from CMakeLists.txt, and also would support clang build for arhitectures other than x86_64.

What do you think about this?

galpeter:
FYI: missing sign-off dco.

build/configs/toolchain_linux_x86_64.cmake
akiss77: @ruben-ayrapetyan, having "worked-for-us" defaults in the build system is a good thing (actually, that's what we already have in the cmake listfile) but not letting developers deviate from the defaults (i.e., "set compiler executable names") is not, IMHO. There will always be platforms and/or tool chains what we don't have access to (or setups which slightly differ from our development environment, see #489 ). I'd definitely vote for a solution where developers have the maximum freedom to choose tools and settings for the build if they want to adapt to their environment or simply experiment.

ruben-ayrapetyan: @akiss77, I agree to you that we should provide maximum freedom to choose build tools and settings.

I think, we should make the choice as simple, as possible.
To make choice simple, we should provide a simple way of switching to another tool (from predefined set or some new / unknown).

In current version of CMakeLists.txt there are some dependencies on gcc / g++.
I think that the dependencies should be moved to configuration files, external to CMakeLists.txt and also add another necessary configurations (currently, for clang).
In the way, we could save logic and structure of CMakeLists.txt.

In the case, any developer would have simple way to add support of another compiler, without changing logic of CMakeLists.txt.

If there would be some features that would not be provided up to the moment, the CMakeLists.txt could be generalized for the cases.

What do you think about this?

akiss77: Would the above-described approach allow overriding the defaults from command line as usually expected to work with make, e.g., `make CC=xxx CFLAGS=xxx` or only by defining configuration files?

egavrin:
@akiss77 which commonly accepted flags from the list below do you like to support and what behavior do you expect?
- `AS`
- `CC`
- `CXX`
- `CFLAGS`
- `CXXFLAGS`
- ` LDFLAGS`
- `ASFLAGS`

akiss77:
`CC`, `CFLAGS`, `CXX`, `CXXFLAGS`, and `LDFLAGS` would certainly be good, while the definition of `AR` and `RANLIB` seems also be needed sometimes (although quite some documentation discourages their definition saying that CMake should figure that out by itself). `AS` and `ASFLAGS` are also nice to have (for the sake of completeness), however, I haven't seen a case where they would have been needed yet.

As for the behaviour: use the specified compilers and archive tools, as well as use the specified compiler/linker options instead of the defaults. (Same for the assembler if implemented.)

egavrin:
@akiss77 great! We see this the same way. Please update this PR with support of `CC`, `CFLAGS`, `CXX`, `CXXFLAGS`, and `LDFLAGS` - I agree that other flags are not needed yet. Provided options should override defaults.

tilmannOSG:
Can you please elaborate why we need to be able to specify the path to the compiler binary (and the other build tools) via environment variables?

With CMake-based projects you usually just set the path with -DCMAKE_C_COMPILER=... when you configure the project.

egavrin:
@akiss77 do you have any plans to update this PR?

tilmannOSG:
@egavrin I had an offline discussion about this with Akos a while ago. The conclusion was that this can be closed.

egavrin:
Closed after offline discussion.

PR #455, opened 2015-07-23
Implement Date.prototype.toJSON()

rtakacs:


jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -1097,7 +1098,7 @@ ecma_builtin_date_prototype_to_utc_string (ecma_value_t this_arg) /**< this argu
  *         Returned value must be freed with ecma_free_completion_value.
  */
 static ecma_completion_value_t
-ecma_builtin_date_prototype_to_iso_string (ecma_value_t this_arg) /**< this argument */
+ecma_builtin_date_prototype_to_iso_string (ecma_value_t this_arg __attr_unused___) /**< this argument */
```
dbatyai: Why do we need to add unused attribute here?

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -1113,9 +1114,69 @@ ecma_builtin_date_prototype_to_iso_string (ecma_value_t this_arg) /**< this argu
...
+    /* 6. */
+    else
+    {
+      ecma_object_t *toISO_obj_p = ecma_get_object_from_value (toISO);
+
+      ECMA_TRY_CATCH (func_ret_val,
+                      ecma_op_function_call (toISO_obj_p, this_arg, NULL, 0),
+                      ret_value);
+
+      ret_value = ecma_copy_value (func_ret_val, true);
```
dbatyai: We could simply do `ret_value = ecma_op_function_call (toISO_obj_p, this_arg, NULL, 0)` instead of `TRY_CATCH` here.

dbatyai: Also, it's possible that a custom `toISOString` of an object returns a non-string primitive type. Since this function should  return a string, we should probably handle this case as well.

tests/jerry/date-tostring.js
```diff
@@ -88,3 +88,17 @@ catch (e)
   assert (e instanceof TypeError);
   assert (e.message === "Incompatible type");
 }
+
+assert (new Date (NaN).toJSON () == null);
+assert (new Date ("2015-07-16").toJSON () == "2015-07-16T00:00:00.000Z");
+assert (new Date ("2015-07-16T11:29:05.023").toJSON () == "2015-07-16T11:29:05.023Z");
+
```
galpeter: We should also add a test to see if the toJSON returns a string which can be processed by the Date constructor.

rtakacs: Added a plus test case.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
rtakacs: 1. Fixed.
2. Those cases are already handled when the toISOString returns with non string values. e.g.:
```
d = new Date ();
d.toISOString () = function () { return null; }
d.toJOSN ();
```
The output of the code above is null.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
rtakacs: Ooops. indeed, sorry.

galpeter:
looks good to me

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -1113,9 +1114,62 @@ ecma_builtin_date_prototype_to_iso_string (ecma_value_t this_arg) /**< this argu
...
+      ret_value = ecma_op_function_call (toISO_obj_p, this_arg, NULL, 0);
+    }
+
+    ECMA_FINALIZE (toISO);
+
+    ecma_deref_ecma_string (to_iso_str_p);
+  }
+
+  ECMA_FINALIZE (tv)
+  ECMA_FINALIZE (obj)
```
ruben-ayrapetyan: Please, add semicolons at end of the lines.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
```diff
@@ -1113,9 +1114,62 @@ ecma_builtin_date_prototype_to_iso_string (ecma_value_t this_arg) /**< this argu
...
+    }
+  }
+
+  if (ecma_is_completion_value_empty (ret_value))
+  {
+    ecma_string_t *to_iso_str_p = ecma_get_magic_string (LIT_MAGIC_STRING_TO_ISO_STRING_UL);
+    ecma_object_t *value_obj_p = ecma_get_object_from_value (obj);
+
+    /* 4. */
+    ECMA_TRY_CATCH (toISO,
```
ruben-ayrapetyan: Currently, variable names are usually named in lowercase. Could you, please, update the name?

rtakacs: Fixed.

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.cpp
rtakacs: Added.

egavrin:
Good to me

dbatyai:
lgtm

ruben-ayrapetyan:
Looks good to me

Issue #348, opened 2015-07-09
Assertion 'ecma_date_year_from_time (time) == ym' failed in ecma_date_make_day


renatahodovan:
###### Jerry version:
    Checked revision: 146ac15
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
new Date(20000000, 0, 1);
```

###### Backtrace:
```javascript
ICE: Assertion 'ecma_date_year_from_time (time) == ym' failed at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp(ecma_date_make_day):634.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x0000000000497a28 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
32    SYSCALL_2
(gdb) bt
# 0  0x0000000000497a28 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
# 1  0x00000000004003c2 in syscall_2 (syscall_no=62, arg1=32710, arg2=6) at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:96
# 2  0x0000000000400563 in abort () at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:175
# 3  0x0000000000447ae3 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.cpp:65
# 4  0x0000000000447b32 in jerry_assert_fail (assertion=0x4a4f10 "ecma_date_year_from_time (time) == ym", 
    file=0x4a4dd8 "jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp", 
    function=0x49f3e0 <_ZZ18ecma_date_make_daydddE8__func__.lto_priv.313> "ecma_date_make_day", line=634)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:92
# 5  0x0000000000423bc9 in ecma_date_make_day (year=20000000, month=0, date=1)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp:634
# 6  0x000000000042c287 in ecma_date_construct_helper(unsigned int const*, unsigned int) [clone .lto_priv.320] (
[...]

PR #459, opened 2015-07-23
Fix assert in the Date prototype

kkristof:
Fix for issue #348

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
@@ -726,6 +726,7 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
                     ecma_number_t month, /**< month value */
                     ecma_number_t date) /**< date value */
 {
+  /* 1. */
   if (ecma_number_is_nan (year)
       || ecma_number_is_nan (month)
       || ecma_number_is_nan (date)
@@ -736,25 +737,66 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
     return ecma_number_make_nan ();
   }
 
+  /* 2., 3., 4. */
   ecma_number_t y = ecma_number_trunc (year);
   ecma_number_t m = ecma_number_trunc (month);
   ecma_number_t dt = ecma_number_trunc (date);
+  /* 5. */
   ecma_number_t ym = y + (ecma_number_t) floor (m / 12);
+  /* 6. */
   ecma_number_t mn = (ecma_number_t) fmod (m, 12);
   mn = (mn < 0) ? 12 + mn : mn;
+
+  /* 7. */
   ecma_number_t time = ecma_date_time_from_year (ym);
+  int32_t step = 182;
+  int32_t delta = step;
 
-  JERRY_ASSERT (ecma_date_year_from_time (time) == ym);
-
-  while (ecma_date_month_from_time (time) < mn)
+  /**
+   * The algorithm below searches the following date: ym-mn-1
+   * To find this time it starts from the beggining of the year (ym)
+   * then increase it by ECMA_DATE_MS_PER_DAY until it reaches the
+   * right date.
+   */
+  if (ecma_date_year_from_time (time) == ym)
   {
-    time += ECMA_DATE_MS_PER_DAY;
+    /**
+     * Binary search to get the closest day in the previous month.
+     * It has to use integer days so sometimes the found time
+     * needs some more increasing which is done day by day below.
+     */
+    while (delta)
+    {
+      if (ecma_date_month_from_time (time + step * ECMA_DATE_MS_PER_DAY) < mn)
+      {
+        step += delta;
+      }
+      else
+      {
+        step -= delta;
+      }
+      delta = delta / 2;
+    }
+
+    if (ecma_date_month_from_time (time + step) < mn)
+    {
+      time += step * ECMA_DATE_MS_PER_DAY;
+    }
+
+    /* Get the month's first day */
+    while (ecma_date_month_from_time (time) < mn)
+    {
+      time += ECMA_DATE_MS_PER_DAY;
+    }
+
+    if (ecma_date_month_from_time (time) == mn && ecma_date_date_from_time (time) == 1)
+    {
+      /* 8. */
+      return ecma_date_day (time) + dt - ((ecma_number_t) 1.0);
+    }
   }
 
-  JERRY_ASSERT (ecma_date_month_from_time (time) == mn);
-  JERRY_ASSERT (ecma_date_date_from_time (time) == 1);
-
-  return ecma_date_day (time) + dt - ((ecma_number_t) 1.0);
+  return ecma_number_make_nan ();
 } /* ecma_date_make_day */
 
 /**
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.h
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.h
@@ -56,13 +56,13 @@ extern ecma_completion_value_t ecma_builtin_helper_def_prop (ecma_object_t *obj_
  *          ECMA-262 v5, 15.9.1.1, 15.9.1.10
  */
 /* Hours in a day. */
-#define ECMA_DATE_HOURS_PER_DAY         24
+#define ECMA_DATE_HOURS_PER_DAY         ((ecma_number_t) 24)
 /* Minutes in an hour. */
-#define ECMA_DATE_MINUTES_PER_HOUR      60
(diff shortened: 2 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
```diff
@@ -825,6 +825,33 @@ ecma_date_timezone_offset (ecma_number_t time) /**< time value */
...
+ * See also:
+ *          ECMA-262 v5, 15.9.1.1
+ *
+ * Used by:
+ *         - The ecma_date_make_day Date helper routine.
+ *
+ * @return  clamped year to 1970-285,616 and 1970+285,616 range
+ */
+ecma_number_t
+ecma_date_clamp_year (ecma_number_t year)
```
ruben-ayrapetyan: @kkristof, could you, please, add comment to the argument?

kkristof:
Hi, I have updated the branch.

ruben-ayrapetyan:
@kkristof, thank you for the update.

ruben-ayrapetyan:
@kkristof, there is a test case that seems to be not processed according to ECMA specification:
```
d = new Date(8.64e+15);
print (d);
d = new Date(8.64e+15 + 1);
print (d);
```

Output:
```
275760-09-13T00:00:00.000
275760-09-13T00:00:00.001
```

Correct output:
```
275760-09-13T00:00:00.000
Invalid Date
```

ruben-ayrapetyan:
Sorry, I've unintentionally pressed "Closed and comment" instead of "Comment", so pull request is now closed and reopened.

ruben-ayrapetyan:
I think that may be we should add `TimeClip` call somewhere for the test case from https://github.com/Samsung/jerryscript/pull/459#issuecomment-124176046 to be processed correctly.

May be this would also help to solve the assertion, described in the pull request.

ruben-ayrapetyan:
> May be this would also help to solve the assertion, described in the pull request.

No, seems that the issue is not related to the pull request.

ruben-ayrapetyan:
Could you, please, describe in comments why current implementation of `ecma_date_make_day` correctly performs 15.9.1.12, step 7?

kkristof:
Hmm, it seems the implementation of the ecma_date_make_day doesn't match with the standard. Thank you for pointing this out. I'm going to fix it.

kkristof:
Hi, I have updated the branch.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
```diff
@@ -741,17 +741,20 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
...
-  JERRY_ASSERT (ecma_date_year_from_time (time) == ym);
-
-  while (ecma_date_month_from_time (time) < mn)
+  if (ecma_date_year_from_time (time) == ym)
   {
-    time += ECMA_DATE_MS_PER_DAY;
-  }
+    while (ecma_date_month_from_time (time) < mn)
+    {
+      time += ECMA_DATE_MS_PER_DAY;
```
ruben-ayrapetyan: Could we remove the the loop by calculating necessary difference and adding it at once, so improving performance?

If not, we could:
- add 28 days (minimal number of days in a month) at first step, as initial day number is 1, so reducing number of iterations to at most 3;
- or use binary search for calculating `time` value.

ruben-ayrapetyan:
@kkristof, thank you for update!

Could you, please, add comment with explanation of algorithm used in `ecma_date_make_day`, i.e. description of why the performed steps are equivalent to behaviour, specified in 15.9.1.12 step 7 (https://github.com/Samsung/jerryscript/pull/459#issuecomment-124178575)?

kkristof:
Hi, I have updated the branch.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
```diff
@@ -733,25 +734,58 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
...
+  /* 6. */
   ecma_number_t mn = (ecma_number_t) fmod (m, 12);
   mn = (mn < 0) ? 12 + mn : mn;
-  ecma_number_t time = ecma_date_time_from_year (ym);
 
-  JERRY_ASSERT (ecma_date_year_from_time (time) == ym);
+  /* 7. We need to find the time where: YearFromTime(time) == ym
+   *    and MonthFromTime(time) == mn and DateFromTime(time) == 1.
+   *    If this is not possible return NaN.
+   */
```
ruben-ayrapetyan: @kkristof, the comment is requirement for algorithm below (I think description of the requirement could be replaced with `/* 7. */` comment in the case).

However, there is no explanation of why the algorithm below does really find value, described in the comment / specification, i.e. there is no description of why the steps performed below are equivalent to behaviour, specified in 15.9.1.12 step 7.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
```diff
@@ -733,25 +734,58 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
...
+      }
+      delta = floor (delta / 2);
+    }
+
+    if (ecma_date_month_from_time (time + step) < mn)
+    {
+      time += step * ECMA_DATE_MS_PER_DAY;
+    }
+
+    while (ecma_date_month_from_time (time) < mn)
```
egavrin: Shouldn't this loop be removed?

kkristof: No, because the binary search has to use integer days so the approximation is given by the binary search needs some fine tuning here.

kkristof:
Hi, I have rebased and updated the branch with comments.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
```diff
@@ -733,25 +734,66 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
...
+  ecma_number_t step = floor ((ecma_number_t) (365) / 2);
+  ecma_number_t delta = step;
+
+  /**
+   * The algorithm below searches the following date: ym-mn-1
+   * To find this time it starts from the beggining of the year (ym)
+   * then increase it by ECMA_DATE_MS_PER_DAY until it reaches the
+   * right date.
+   */
+  if (ecma_date_year_from_time (time) == ym)
```
egavrin: @galpeter could you please revise this block for performance. I suppose that algorithm can be significantly improved.

kkristof: Hi, I have made a quick test with every date in a year.
The binary search always takes 9 iterations and in the worst case it needs 4 iterations in the day by day phase while in the best case it skips it. The whole test took 0.458s.
The old algorithm is far worse because it needs 335 (december) iterations at the worst case but 0 in the best case (january). The test took 0.574s.
We could use lookup table for the months to get the days but I think we won't get any measurable performance gain in normal usage with this function.

galpeter: @egavrin , I think it's fine for now, it's faster than the original one. If we want to further improve it then creating a lookup table would be a good idea imho, but that should be a separate PR as this is for fixing the assert. Maybe we can raise an issue for it.

galpeter:
lgtm

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
ruben-ayrapetyan: For example, `floor ((ecma_number_t) (365) / 2)` constant could be calculated more efficiently. Also, maybe there are some other places in the patch that can be improved. By the way, why do we need floating point for [[Date]] values?

kkristof: The `Date` needs 52 significant bit to represent the date as the standard describes it. Probably it would be better to use `int64_t` especially on those target where the `ecma_numer_t` is single precision. I think this refactoring should be in different PR with other optimizations as well, but I'm not sure how much performance gain we would get through optimizing the `Date` object.

ruben-ayrapetyan: @kkristof, I agree with you that switching to `uint64_t` should not be performed in this PR.

kkristof:
Hi, I have rebased and updated the branch.

egavrin:
Ok, lets push it. But Date should be optimized a bit: according to `perf` ~20% of test262 execution time is Date-related staff. Please, raise issue for performance optimization of Date.

kkristof:
I have raised #560 for this. I haven't thought that it can cause this much impact.

egavrin:
`make push`

PR #461, opened 2015-07-23
Support local time output

rtakacs:


kkristof:
Could you add test to the PR?

rtakacs:
We cannot test the local time output, just now, while the TZA and DST are the same for everyone. If they are implemented, we should remove the these test cases.
Also added a test case for the Date object that constructed with an UTC time with a 'Z' suffix.

kkristof:
Looks good to me.

zherczeg:
LGTM

egavrin:
Good to me. Rebase and `make push`

PR #462, opened 2015-07-23
Fix push() and unshift() in case result length is larger than UINT_MAX

dbatyai:


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -57,21 +57,29 @@
...
 {
   ecma_completion_value_t ret_value;
-  ecma_string_t *magic_string_length_p = ecma_get_magic_string (LIT_MAGIC_STRING_LENGTH);
 
-  ecma_number_t *len_p = ecma_alloc_number ();
-  *len_p = ecma_uint32_to_number (length);
+  if (ecma_object_get_class_name (object) == LIT_MAGIC_STRING_ARRAY_UL
+      && unlikely (length > (uint32_t)(-1))) /* Check if length greater than UINT_MAX */
+  {
+    ret_value = ecma_make_throw_obj_completion_value (ecma_new_standard_error (ECMA_ERROR_RANGE));
```
ruben-ayrapetyan: @dbatyai, the error should be raised in `ecma_op_array_object_define_own_property`. Is there any case, when it isn't raised?

dbatyai:
@ruben-ayrapetyan, I've updated the patch. The error does get raised correctly in `ecma_op_array_object_define_own_property` when it should.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp
```diff
@@ -644,61 +643,38 @@ ecma_builtin_array_prototype_object_push (ecma_value_t this_arg, /**< this argum
...
-       index < arguments_number;
+       index < arguments_number && ecma_is_completion_value_empty (ret_value);
        index++, n++)
   {
     /* 5.a */
     ecma_value_t e_value = argument_list_p[index];
 
     /* 5.b */
-    ecma_string_t *n_str_p = ecma_new_ecma_string_from_uint32 (n);
+    ecma_string_t *n_str_p = ecma_new_ecma_string_from_number (n);
```
ruben-ayrapetyan: It is better to use `ecma_new_ecma_string_from_uint32` instead of `ecma_new_ecma_string_from_number`, as it reduces memory usage.

dbatyai: We can't use `ecma_new_ecma_string_from_uint32` here, because the index can go above 2^32.

ruben-ayrapetyan: I see. Thanks for explanation.

ruben-ayrapetyan:
Looks good to me

galpeter:
Looks good to me.

galpeter:
Landed: https://github.com/Samsung/jerryscript/commit/fec5933a3db4b2bd7657152d14a33e7439c77a87

PR #467, opened 2015-07-24
Make parseInt unicode tolerant

### LINKED ISSUE #405: Unicode support in implementation of parseInt
Hello.

Implementation of `parseInt` (ecma-builtin-global.cpp) should be updated to use UTF-8 iterators for processing characters buffer.

Comment of @zherczeg in #394:
> Is this correct? If I understand correctly, we just pass a byte, rather than an uint16_t character. I think we should rewrite this to use the UTF iterators, since there are line terminators and white spaces outside of the [0-127] byte range.

galpeter:
Fixes #405 

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
@@ -121,17 +121,25 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
                                                        utf8_string_buff,
                                                        (ssize_t) str_size);
     JERRY_ASSERT (bytes_copied >= 0);
-    utf8_string_buff[str_size] = LIT_BYTE_NULL;
+    lit_utf8_iterator_t iter = lit_utf8_iterator_create (utf8_string_buff, str_size);
 
     /* 2. Remove leading whitespace. */
-    ecma_length_t start = str_size;
-    ecma_length_t end = str_size;
-    for (ecma_length_t i = 0; i < end; i++)
+    lit_utf8_iterator_seek_eos (&iter);
+
+    lit_utf8_iterator_pos_t start = lit_utf8_iterator_get_pos (&iter);
+    lit_utf8_iterator_pos_t end = lit_utf8_iterator_get_pos (&iter);
+
+    lit_utf8_iterator_seek_bos (&iter);
+
+    while (!lit_utf8_iterator_is_eos (&iter))
     {
-      if (!lit_char_is_white_space (utf8_string_buff[i])
-          && !lit_char_is_line_terminator (utf8_string_buff[i]))
+      ecma_char_t current_char = lit_utf8_iterator_read_next (&iter);
+
+      if (!lit_char_is_white_space (current_char)
+          && !lit_char_is_line_terminator (current_char))
       {
-        start = i;
+        lit_utf8_iterator_read_prev (&iter);
+        start = lit_utf8_iterator_get_pos (&iter);
         break;
       }
     }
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -185,47 +195,55 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
...
       }
 
       /* 11. Check if characters are in [0, Radix - 1]. We also convert them to number values in the process. */
-      for (lit_utf8_size_t i = start; i < end; i++)
+      lit_utf8_iterator_seek (&iter, start);
+      while (!lit_utf8_iterator_is_eos (&iter))
       {
-        if ((utf8_string_buff[i]) >= LIT_CHAR_LOWERCASE_A && utf8_string_buff[i] <= LIT_CHAR_LOWERCASE_Z)
+        ecma_char_t current_char = lit_utf8_iterator_read_next (&iter);
+        ecma_number_t current_number;
```
dbatyai: I think we should use an integer type here instead of an `ecma_number_t`.

zherczeg:
LGTM

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -140,15 +148,17 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
...
       sign = -1;
     }
 
     /* 5. */
-    if (utf8_string_buff[start] == LIT_CHAR_MINUS || utf8_string_buff[start] == LIT_CHAR_PLUS)
+    if (current == LIT_CHAR_MINUS || current == LIT_CHAR_PLUS)
     {
-      start++;
+      start = lit_utf8_iterator_get_pos (&iter);
+      current = lit_utf8_iterator_read_next (&iter);
```
dbatyai: Shouldn't we check for `eos` before doing this?

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
galpeter: good point, will fix it in the update.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
galpeter: Yeah, good catch!

galpeter:
PR updated.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -185,47 +198,55 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg __attr_unused___, /*
...
       if (strip_prefix)
       {
-        if (end - start >= 2
-            && utf8_string_buff[start] == LIT_CHAR_0
-            && (utf8_string_buff[start + 1] == LIT_CHAR_LOWERCASE_X
-                || utf8_string_buff[start + 1] == LIT_CHAR_UPPERCASE_X))
+        if (end.offset - start.offset >= 2 && current == LIT_CHAR_0)
         {
-          start += 2;
+          ecma_char_t next = lit_utf8_iterator_peek_next (&iter);
```
dbatyai: There probably should be an `eos` check here as well.

galpeter: In this case it is not needed, as the offset check in the `if` statement above already checked that there are at least two characters, so we are not at the end of the string yet.

dbatyai: Ah, okay.

dbatyai:
lgtm

sand1k:
LGTM.

sand1k:
LGTM.

egavrin:
`make push`

dbatyai:
Landed: 3fbe54393f4900ea03b452326d1e96dbd9bf2d3a

Issue #439, opened 2015-07-21
Segmentation fault at `jerry_api_string_to_char_buffer()` for empty string.


ILyoan:
`jerry_api_string_to_char_buffer()` was designed to return negation of required bytes for buffer when given buffer is not sufficient.

This function can take `NULL` for buffer to calculate required bytes to allocate buffer for a string. But if the string is actually empty string ('') then this will result segmentation fault.
```
  ssize_t ret_val = ecma_string_to_utf8_string (string_p, (lit_utf8_byte_t *) buffer_p, buffer_size);
  if (ret_val >= 0)
  {
    buffer_p[ret_val] = 0;
    ret_val++;
  }
  else
  {
    ret_val--;
  }
``` 
here, `ecma_string_to_utf8_string` returns `0`, next if statement hold, and then seg fault.

sand1k:
@ILyoan, thanks for reporting, (NULL, 0) case should be processed gracefully.

Maybe, we should also change this function, so it doesn't write '\0' at the end of buffer, because unicode strings may contain 0 in the middle? 
@ILyoan, @ruben-ayrapetyan what do you think?

ruben-ayrapetyan:
@sand1k, I think removing `\0` from end of buffer is good solution, as it would also make API more accurate and clear in relation to possible presence of '\0' inside of character set.

ILyoan:
@sand1k I don't mind removing `\0` at the end. Actually I didn't know that this API was adding `\0` before encounter this segmentation fault.

ILyoan:
@sand1k Can we close this?

egavrin:
@ILyoan closing.

PR #471, opened 2015-07-24
Change jerry_api_string_to_char_buffer so it doesn't add '\0' to the end of output buffer.

sand1k:
Fixes issue #439.

### DIFF
--- a/jerry-core/jerry.cpp
+++ b/jerry-core/jerry.cpp
@@ -274,10 +274,13 @@ jerry_api_convert_api_value_to_ecma_value (ecma_value_t *out_value_p, /**< out: 
  */
 
 /**
- * Copy string characters to specified buffer, append zero character at end of the buffer.
+ * Copy string characters to specified buffer.
+ *
+ * Note:
+ *   '\0' could occur in characters.
  *
  * @return number of bytes, actually copied to the buffer - if string's content was copied successfully;
- *         otherwise (in case size of buffer is insuficcient) - negative number, which is calculated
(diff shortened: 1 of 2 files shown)


jerry-core/jerry.cpp
```diff
@@ -274,10 +274,13 @@ jerry_api_convert_api_value_to_ecma_value (ecma_value_t *out_value_p, /**< out:
...
 /**
- * Copy string characters to specified buffer, append zero character at end of the buffer.
+ * Copy string characters to specified buffer.
+ *
+ * Note:
+ *   '\0' could occur in characters.
  *
  * @return number of bytes, actually copied to the buffer - if string's content was copied successfully;
- *         otherwise (in case size of buffer is insuficcient) - negative number, which is calculated
+ *         otherwise (in case size of buffer is insufficient) - negative number, which is calculated
```
egavrin: ^_^

egavrin:
@seanshpark will this change affect IoTjs?

seanshpark:
@egavrin , I think the change is ok with iotjs. but with latest jerryscript master (not this change) fails to run iotjs tests so cannot check with tests. have to find the revision of the failure.

zherczeg:
From technical point of view, LGTM.

sand1k:
@seanshpark, did you find the reason of failure?
@egavrin, could I land the patch?

egavrin:
Good to me. I'll land

seanshpark:
@sand1k , `@seanshpark, did you find the reason of failure?`: yes, ILyoan has found and fixed in iotjs.

PR #474, opened 2015-07-27
Implement String.prototype.replace function.

zherczeg:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
@@ -20,6 +20,7 @@
 #include "ecma-builtins.h"
 #include "ecma-conversion.h"
 #include "ecma-exceptions.h"
+#include "ecma-function-object.h"
 #include "ecma-gc.h"
 #include "ecma-globals.h"
 #include "ecma-helpers.h"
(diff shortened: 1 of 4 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+ * @return the constructed string
+ */
+static ecma_string_t *
+ecma_builtin_string_prototype_object_replace_append_substr (ecma_string_t *base_string_p, /**< base string */
+                                                            ecma_string_t *appended_string_p, /**< appended string */
+                                                            ecma_length_t start, /**< start position */
+                                                            ecma_length_t end, /**< end position */
+                                                            bool free_base_string) /**< free base string or not */
+{
+  ecma_string_t *appended_string;
```
ruben-ayrapetyan: The variable name is very similar to argument's name `appended_string_p`. Maybe, `ret_string_p` would be better.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+       * we can trust in the returned value. */
+
+      JERRY_ASSERT (ecma_is_value_number (index_value));
+      JERRY_ASSERT (ecma_is_value_string (result_string_value));
+
+      /* We use the length of the result string to determine the
+       * match end. This works regardless the global flag is set. */
+      ecma_string_t *result_string_p = ecma_get_string_from_value (result_string_value);
+
+      context_p->match_start = (ecma_length_t) (*ecma_get_number_from_value (index_value));
```
ruben-ayrapetyan: Is it correct that the value always fits `ecma_length_t` (would it be correct if, for example, `ecma_length_t` would be `uint16_t`)?

zherczeg: Should be since it is a valid string index, and I expect the maximum strings length is <= ecma_length_t (since get_string_length returns with ecma_length_t).

ruben-ayrapetyan: Maybe, we should add check like `(ecma_length_t) ecma_number_to_uint32 (index_num) == index_num`.

What do you think about this?

zherczeg: An Assert? Or use ecma_number_to_uint32 instead of ecma_get_number_from_value? I am not sure I follow you.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+                    ret_value);
+
+    JERRY_ASSERT (ecma_is_value_number (length_value));
+
+    ecma_length_t length = (ecma_length_t) (*ecma_get_number_from_value (length_value));
+
+    JERRY_ASSERT (length >= 1);
+
+    MEM_DEFINE_LOCAL_ARRAY (arguments_list,
+                            length + 2,
```
ruben-ayrapetyan: In a very special case, the value would overflow. Seems that in the case out-of-memory should be indicated.
I'm not sure, currently, how to catch the case in general, without adding corresponding check in all similar cases that could occur in future. Maybe, we should add the check to `MEM_DEFINE_LOCAL_ARRAY`.

What do you think about this?

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+      else if (action == LIT_CHAR_SINGLE_QUOTE)
+      {
+        ecma_string_t *input_string_p = ecma_get_string_from_value (context_p->input_string);
+        result_string_p = ecma_builtin_string_prototype_object_replace_append_substr (result_string_p,
+                                                                                      input_string_p,
+                                                                                      context_p->match_end,
+                                                                                      context_p->input_length,
+                                                                                      true);
+      }
+      else if (action != LIT_CHAR_NULL)
```
ruben-ayrapetyan: Could we put lines 1127-1194 to `if` block at lines 1117-1125?

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
zherczeg: I am not sure I follow you here as well. You mean length is maximum length - 1, and length + 2 overflows? This problem is more general, e.g. if the allocated length is less than maximum, but the sizeof allocated type is not 1, it can also be an overflow. I mean if we allocate max_length - 1 ecma_value-s, the (max_length-1)*sizeof(ecma_value) also overflows. So this is definitely needs to be solved in the macro if we want it.

Am I understand correctly that you plan to change ecma_length to uint16?

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+
+    ECMA_FINALIZE (match_value);
+  }
+
+  ecma_deref_ecma_string (result_string_p);
+  return ret_value;
+} /* ecma_builtin_string_prototype_object_replace_loop */
+
+/**
+ * Generic helper function to analiyse the search value
```
ruben-ayrapetyan: Typo: "analyse"

ruben-ayrapetyan: Could you, please, add more details on analysis, performed by the function?

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -756,11 +1384,90 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+
+      ECMA_TRY_CATCH (put_value,
+                      ecma_op_object_put (regexp_obj_p,
+                                          last_index_string_p,
+                                          ecma_make_number_value (zero_number_p),
+                                          true),
+                      ret_value);
+
+      ecma_dealloc_number (zero_number_p);
+      ECMA_FINALIZE (put_value);
```
ruben-ayrapetyan: Maybe, the `dealloc` should be performed after the `ECMA_FINALIZE`.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+    context_p->replace_iterator = lit_utf8_iterator_create (replace_start_p, replace_size);
+
+    ret_value = ecma_builtin_string_prototype_object_replace_loop (context_p);
+
+    MEM_FINALIZE_LOCAL_ARRAY (replace_start_p);
+    ECMA_FINALIZE (to_string_replace_val);
+  }
+
+  return ret_value;
+}
```
ruben-ayrapetyan: Comment is missing.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
ruben-ayrapetyan: @zherczeg, I think adding an assert would be ok for the case. If `index_num` can be represented in `ecma_length_t`, then simple C-style conversion would work correctly. Maybe, later we should add interface for the conversion that would perform the check internally.

P.S. Sorry for deleting my first comment to the code block (about "would be the conversion correct if `ecma_length_t` would be, for example, `uint16_t`"). I've deleted the comment before I've seen your answer https://github.com/Samsung/jerryscript/pull/474#discussion_r35731210, and added new comment https://github.com/Samsung/jerryscript/pull/474#discussion_r35731390.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
ruben-ayrapetyan: @zherczeg, I see. You are right. The check, if we need it, should be performed in macro.
Thank you for your explanation.

No, I have no plans to change `ecma_length_t` to `uint16_t`. The `uint16_t` is an example to consider what would occur if this or similar change would be performed in future.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+  ecma_length_t match_start; /**< starting position of the match */
+  ecma_length_t match_end; /**< end position of the match */
+
+  /* Replace value callable part. */
+  ecma_object_t *replace_function_p;
+
+  /* Replace value string part. */
+  ecma_string_t *replace_string_p; /**< replace string */
+  lit_utf8_iterator_t replace_iterator; /**< replace string iterator */
+} replace_search_context;
```
ruben-ayrapetyan: Could you, please, rename the structure to `ecma_replace_search_ctx_t`?

`ecma_builtin_replace_search_ctx_t` would be even better from my view point. If it is ok for you, could you, please, perform the change?

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,7 +744,634 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
 
   return ret_value;
 } /* ecma_builtin_string_prototype_object_match */
-#endif /* !CONFIG_ECMA_COMPACT_PROFILE_DISABLE_REGEXP_BUILTIN */
+
+typedef struct
```
ruben-ayrapetyan: Could you, please, add description comment for the structure?

ruben-ayrapetyan:
Could you, please, add note to `ecma_builtin_string_prototype_object_replace` comment that would briefly describe scheme of replacement (sequence of actions) with references to helper function?

zherczeg:
I think I did the requested changes.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
```diff
@@ -743,11 +744,672 @@ ecma_builtin_string_prototype_object_match (ecma_value_t this_arg, /**< this arg
...
+        else if (action == LIT_CHAR_SINGLE_QUOTE)
+        {
+          ecma_string_t *input_string_p = ecma_get_string_from_value (context_p->input_string);
+          result_string_p = ecma_builtin_string_prototype_object_replace_append_substr (result_string_p,
+                                                                                        input_string_p,
+                                                                                        context_p->match_end,
+                                                                                        context_p->input_length,
+                                                                                        true);
+        }
+        else if (action != LIT_CHAR_NULL)
```
ruben-ayrapetyan: The condition is always true.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
ruben-ayrapetyan: `JERRY_ASSERT ((ecma_length_t) ecma_number_to_uint32 (*length_number_p) == length);` is missing.

ruben-ayrapetyan:
@zherczeg, thank you for the update!

After updating https://github.com/Samsung/jerryscript/pull/474#discussion_r35745483, https://github.com/Samsung/jerryscript/pull/474#discussion_r35746541, the changes would look good to me.

zherczeg:
Thank you for the review. I added two new tests.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.cpp
zherczeg: I realized I added it to a wrong place.

galpeter:
lgtm

ruben-ayrapetyan:
Looks good to me

egavrin:
Good to me

zherczeg:
Landed 048e20925e9b4e494575c86e075e715d0e15543f

PR #476, opened 2015-07-27
Add opcode description table containing an opcode's name, name and possible type of the opcode's arguments

ruben-ayrapetyan:


ruben-ayrapetyan:
Merged as part of #482.

PR #478, opened 2015-07-27
Optimize lit_char_is_unicode_letter.

sand1k:
Add fast path for ASCII characters which are not letters.

### DIFF
--- a/jerry-core/lit/lit-char-helpers.cpp
+++ b/jerry-core/lit/lit-char-helpers.cpp
@@ -15,6 +15,8 @@
 
 #include "lit-char-helpers.h"
 
+#include "lit-strings.h"
+
 /**
  * Check if specified character is one of the Format-Control characters
  *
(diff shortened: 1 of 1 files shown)


jerry-core/lit/lit-char-helpers.cpp
```diff
@@ -112,6 +114,10 @@ lit_char_is_unicode_letter (ecma_char_t c) /**< code unit */
   {
     return true;
   }
+  else if (c <= LIT_UTF8_1_BYTE_CODE_POINT_MAX)
+  {
+    return false;
```
ruben-ayrapetyan: Could you, please, add a comment like `/* Other one-byte code points are not corresponding to letters */`?

ruben-ayrapetyan:
Looks good to me

zherczeg:
Nice idea. LGTM

jerry-core/lit/lit-char-helpers.cpp
ruben-ayrapetyan: The comment wasn't added.

PR #481, opened 2015-07-27
Handle Unicode characters in parseFloat()

dbatyai:


jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -290,147 +290,225 @@ ecma_builtin_global_object_parse_float (ecma_value_t this_arg __attr_unused___,
...
+    const lit_utf8_byte_t *infinity_utf8_str_p = lit_get_magic_string_utf8 (LIT_MAGIC_STRING_INFINITY_UL);
+    lit_utf8_iterator_t infinity_iter = lit_utf8_iterator_create (infinity_utf8_str_p,
+                                                                  sizeof (*infinity_utf8_str_p));
+
+    JERRY_ASSERT (!lit_utf8_iterator_is_eos (&infinity_iter));
+
+    /* Check if string is equal to "Infinity". */
+    while (!lit_utf8_iterator_is_eos (&iter)
+           && (lit_utf8_iterator_read_next (&iter) == lit_utf8_iterator_read_next (&infinity_iter)))
+    {
```
sand1k: I think its better to use `lit_compare_utf8_string_and_magic_string` to compare with "Infinity", the code would be more compact:
```
if (lit_compare_utf8_string_and_magic_string (iter.buf_p + start.offset,
                                              iter.buf_size - start.offset,
                                              LIT_MAGIC_STRING_INFINITY_UL))
{
  ...
}
```

dbatyai: We can't use `lit_compare_utf8_string_and_magic_string` here, because we have check if a prefix of the string contains "Infinity". For example, "Infinityfoo" should also be valid.

tests/jerry/global-parsefloat.js
```diff
@@ -38,12 +38,26 @@ assert(parseFloat("Infinityfoo") === Infinity);
...
 assert(isNaN(parseFloat("")));
 assert(isNaN(parseFloat(".")));
+assert(isNaN(parseFloat("+")));
+assert(isNaN(parseFloat("-")));
+assert(isNaN(parseFloat("e")));
+assert(isNaN(parseFloat("a")));
+assert(isNaN(parseFloat("e+")));
+assert(isNaN(parseFloat("+e-")));
+assert(isNaN(parseFloat(".e")));
+assert(isNaN(parseFloat(".a")));
```
galpeter: Maybe we should also add test-cases like these:
```js
parseFloat("1.1")
parseFloat("1..1")
parseFloat("1.1.2")
```

dbatyai: Added.

galpeter:
Thanks for the tests,
lgtm

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -290,147 +290,225 @@ ecma_builtin_global_object_parse_float (ecma_value_t this_arg __attr_unused___,
...
-      /* Check digits of whole part. */
-      for (lit_utf8_size_t i = current; i < str_size; i++, current++)
+    if (current == LIT_CHAR_MINUS || current == LIT_CHAR_PLUS)
+    {
+      /* Set starting position to be after the sign character. */
+      start = lit_utf8_iterator_get_pos (&iter);
+    }
+    else
+    {
+      lit_utf8_iterator_read_prev (&iter);
```
sand1k: Could you, please, replace `lit_utf8_iterato_read_prev` with `lit_utf8_iterator_decr` when you don't use return value?

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
sand1k: OK, I see.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
dbatyai: Fixed.

sand1k:
LGTM.

PR #482, opened 2015-07-27
Split opcode and instruction entities

ruben-ayrapetyan:
The pull request currently includes commit ba5a906 from #476.


egavrin:
Good to me

zherczeg:
Looks good for me as well.

I would probably simply use instr instead of vm_instr, since it is obvious that the instruction is a virtual machine instruction, but whatever naming is fine for me.

PR #483, opened 2015-07-28
Implement String.prototype.lastIndexOf()

lvidacs:
JerryScript-DCO-1.0-Signed-off-by: Laszlo Vidacs lvidacs.u-szeged@partner.samsung.com

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp
@@ -16,6 +16,7 @@
 
 #include "ecma-builtin-helpers.h"
 
+#include "ecma-alloc.h"
 #include "ecma-array-object.h"
 #include "ecma-builtins.h"
 #include "ecma-conversion.h"
@@ -470,22 +471,33 @@ ecma_builtin_helper_array_concat_value (ecma_object_t *obj_p, /**< array */
  * This function clamps the given index to the [0, length] range.
  * If the index is negative, 0 value is used.
  * If the index is greater than the length of the string, the normalized index will be the length of the string.
+ * NaN is mapped to zero or length depending on the nan_to_zero parameter.
  *
  * See also:
  *          ECMA-262 v5, 15.5.4.15
  *
  * Used by:
  *         - The String.prototype.substring routine.
+ *         - The String.prototype.indexOf routine.
+ *         - The ecma_builtin_helper_string_prototype_object_index_of helper routine.
  *
  * @return uint32_t - the normalized value of the index
  */
 uint32_t
 ecma_builtin_helper_string_index_normalize (ecma_number_t index, /**< index */
-                                            uint32_t length) /**< string's length */
+                                            uint32_t length, /**< string's length */
+                                            bool nan_to_zero) /**< whether NaN is mapped to zero (t) or length (f) */
 {
   uint32_t norm_index = 0;
 
-  if (!ecma_number_is_nan (index) && !ecma_number_is_negative (index))
+  if (ecma_number_is_nan (index))
+  {
+    if (!nan_to_zero)
+    {
+      norm_index = length;
+    }
+  }
+  else if (!ecma_number_is_negative (index))
   {
     if (ecma_number_is_infinity (index))
     {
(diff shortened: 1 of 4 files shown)


lvidacs:
The helper function is prepared to handle indexOf() as well, it will be adjusted in a separate step.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp
```diff
@@ -506,6 +515,163 @@ ecma_builtin_helper_string_index_normalize (ecma_number_t index, /**< index */
...
+
+      /* iterate original string and try to match at each position */
+      bool found = false;
+      bool searching = true;
+
+      while (!found && searching)
+      {
+        /* match as long as possible */
+        ecma_length_t match_len = 0;
+        lit_utf8_iterator_pos_t stored_original_pos = lit_utf8_iterator_get_pos (&original_it);
```
zherczeg: Can't we save the iterator itself? Seeking is not a cheap operation.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.h
```diff
@@ -34,7 +34,11 @@ extern ecma_completion_value_t ecma_builtin_helper_array_concat_value (ecma_obje
                                                                        uint32_t *length,
                                                                        ecma_value_t);
 extern uint32_t ecma_builtin_helper_array_index_normalize (ecma_number_t index, uint32_t length);
-extern uint32_t ecma_builtin_helper_string_index_normalize (ecma_number_t index, uint32_t length);
+extern uint32_t ecma_builtin_helper_string_index_normalize (ecma_number_t index, uint32_t length, bool nan_to_zero);
+extern ecma_completion_value_t ecma_builtin_helper_string_prototype_object_index_of (ecma_value_t this_arg,
+                                                                                            ecma_value_t arg1,
+                                                                                            ecma_value_t arg2,
+                                                                                            bool firstIndex);
```
galpeter: Incorrect indentation

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp
```diff
@@ -481,11 +482,19 @@ ecma_builtin_helper_array_concat_value (ecma_object_t *obj_p, /**< array */
  */
 uint32_t
 ecma_builtin_helper_string_index_normalize (ecma_number_t index, /**< index */
-                                            uint32_t length) /**< string's length */
+                                            uint32_t length, /**< string's length */
+                                            bool nan_to_zero) /**< whether NaN is mapped to zero (t) or length (f) */
```
galpeter: This argument modifies the behavior a bit so we need to update the function comment above.

lvidacs:
@galpeter, thanks for the comments, the patch is rebased and updated.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp
zherczeg: This is the first time this appeared as problem? I think charCodeAt also depends on this.

lvidacs: charCodeAt needs similar checks, but is a bit simpler and does not use the helper function.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp
```diff
@@ -505,6 +517,163 @@ ecma_builtin_helper_string_index_normalize (ecma_number_t index, /**< index */
...
+               lit_utf8_iterator_read_next (&original_it) == lit_utf8_iterator_read_next (&search_it))
+        {
+          match_len++;
+        }
+
+        /* check for match */
+        if (match_len == search_len)
+        {
+          *ret_num_p = ecma_uint32_to_number (index);
+          found = true;
```
galpeter: Do we need the `found` at all? Can't we just break out?

galpeter:
lgtm

zherczeg:
looks good, but please open an issue that where the followup work is described

lvidacs:
Issue #515 is opened for follow-up work.

egavrin:
`make push`

kkristof:
Landed in https://github.com/Samsung/jerryscript/commit/554305d88c497ff04ef4602ffa93f5b821c7f4af

PR #484, opened 2015-07-28
Rename isalpha, isdigit

zherczeg:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
@@ -39,27 +39,6 @@
 #define BUILTIN_UNDERSCORED_ID json
 #include "ecma-builtin-internal-routines-template.inc.h"
 
-/*
- * FIXME:
- *       Replace usage of isalpha and isdigit functions in the module with lit_char helpers and remove the functions.
- *
- *       Related issue: #424
- */
-
-/** Checks for an alphabetic character. */
-static int
-isalpha (int c)
-{
-  return (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z');
-}
-
-/** Checks for a digit (0 through 9).  */
-static int
-isdigit (int c)
-{
-  return c >= '0' && c <= '9';
-}
-
 /** \addtogroup ecma ECMA
  * @{
  *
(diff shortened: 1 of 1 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -126,9 +105,12 @@ ecma_builtin_json_check_id (lit_utf8_byte_t *string_p, /**< start position */
   {
     string_p++;
     id_p++;
-    if (*id_p == '\0')
+    if (*id_p == LIT_CHAR_NULL)
     {
-      return !isalpha (*string_p) && !isdigit (*string_p) && *string_p != '$' && *string_p != '_';
+      return (!lit_char_is_ascii_alpha (*string_p)
```
ruben-ayrapetyan: @zherczeg, could you, please, clarify ECMA-262 paragraph that defines the conditions?

zherczeg: No paragraph, just convenience. I think it would work without this condition, since two constants like truefalse is not allowed.

ruben-ayrapetyan: @zherczeg, I'm trying to figure out what part / paragraph of ECMA-262 corresponds to the following conditions block:
```
(!lit_char_is_ascii_alpha (*string_p)
 && !lit_char_is_decimal_digit (*string_p)
 && *string_p != LIT_CHAR_DOLLAR_SIGN
 && *string_p != LIT_CHAR_UNDERSCORE)
```

Could you, please, clarify what is the paragraph / part?

zherczeg: I said above no paragraph, just convenience :)

According to the spec, the lexer part of JSON accepts things like truetrue  and parses them as two token (which is not how identifiers usually handled). But two JSONValue-s cannot follow each other, so this will eventually aborts. But I think it is nicer to abort after the first ID.

jerry-core/lit/lit-char-helpers.cpp
```diff
@@ -288,6 +288,18 @@ lit_char_is_hex_digit (ecma_char_t c) /**< code unit */
 } /* lit_char_is_hex_digit */
 
 /**
+ * Check if specified character is an ascii alpha character
+ *
+ * @return true / false
+ */
+bool
+lit_char_is_ascii_alpha (ecma_char_t c) /**< code unit */
```
ruben-ayrapetyan: @zherczeg, seems that this function is unused now. Could you, please, remove it?

ruben-ayrapetyan:
After `lit_char_is_ascii_alpha` is removed (https://github.com/Samsung/jerryscript/pull/484#discussion-diff-35640406), the changes would look good to me.

zherczeg:
Function removed.

ruben-ayrapetyan:
Looks good to me

kkristof:
Looks good to me.

egavrin:
`make push`

zherczeg:
Landed 41aef56cbaec0a91ad9089028676c6b7bca6a53c

PR #486, opened 2015-07-28
Implement IgnoreCase in RegExp engine

szledan:

### DIFF
--- a/jerry-core/ecma/operations/ecma-regexp-object.cpp
+++ b/jerry-core/ecma/operations/ecma-regexp-object.cpp
@@ -286,7 +286,39 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
         ecma_char_t ch2 = lit_utf8_iterator_read_next (&iter);
         JERRY_DDLOG ("Character matching %d to %d: ", ch1, ch2);
 
-        if (ch1 != ch2)
+        if (re_ctx_p->flags & RE_FLAG_IGNORE_CASE)
+        {
+          ecma_char_t ch1_buffer[LIT_MAXIMUM_OTHER_CASE_LENGTH];
+          ecma_char_t ch2_buffer[LIT_MAXIMUM_OTHER_CASE_LENGTH];
+          lit_utf8_size_t ch1_length = lit_char_to_lower_case (ch1,
+                                                               ch1_buffer,
+                                                               LIT_MAXIMUM_OTHER_CASE_LENGTH);
+
+          lit_utf8_size_t ch2_length = lit_char_to_lower_case (ch2,
+                                                               ch2_buffer,
+                                                               LIT_MAXIMUM_OTHER_CASE_LENGTH);
+
+          JERRY_ASSERT (ch1_length >= 1 && ch1_length <= LIT_MAXIMUM_OTHER_CASE_LENGTH);
+          JERRY_ASSERT (ch2_length >= 1 && ch2_length <= LIT_MAXIMUM_OTHER_CASE_LENGTH);
+
+          if (ch1_length != ch2_length)
+          {
+            JERRY_DDLOG ("fail\n");
+            re_ctx_p->recursion_depth--;
+            return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_FALSE); /* fail */
+          }
+
+          for (lit_utf8_size_t i = 0; i < ch1_length; i++)
+          {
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -286,7 +286,39 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
...
+        {
+          ecma_char_t ch1_buffer[LIT_MAXIMUM_OTHER_CASE_LENGTH];
+          ecma_char_t ch2_buffer[LIT_MAXIMUM_OTHER_CASE_LENGTH];
+          lit_utf8_size_t ch1_length = lit_char_to_lower_case (ch1,
+                                                               ch1_buffer,
+                                                               LIT_MAXIMUM_OTHER_CASE_LENGTH);
+
+          lit_utf8_size_t ch2_length = lit_char_to_lower_case (ch2,
+                                                               ch2_buffer,
+                                                               LIT_MAXIMUM_OTHER_CASE_LENGTH);
```
zherczeg: The standard explicitly says (in 15.10.2.8 Atom) that you should use the uppercase function. Also, if the length is > 1, the original character should be used (also there is a rule for characters > 127). I think during the pattern construction you should uppercase the pattern characters, so only one call for uppercase could be done here.

Personally I would introduce a regexp_uppercase or some similar helper function. It expects an ecma_char_t and returns with an ecma_char_t. That is all.

egavrin:
Good to me

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -230,6 +230,43 @@ ecma_op_create_regexp_object (ecma_string_t *pattern_p, /**< input pattern */
...
+static ecma_char_t __attr_always_inline___
+re_canonicalize (ecma_char_t ch, /**< character */
+                 bool is_ignorecase) /**< IgnoreCase flag */
+{
+  ecma_char_t ret_value = ch;
+
+  if (is_ignorecase)
+  {
+    /* 2. */
+    MEM_DEFINE_LOCAL_ARRAY (u, LIT_MAXIMUM_OTHER_CASE_LENGTH, ecma_char_t);
```
zherczeg: Can't we just use a simple stack array?
ecma_char_t u[LIT_MAXIMUM_OTHER_CASE_LENGTH];

I don't think we need a dynamic alocation here.

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -230,6 +230,43 @@ ecma_op_create_regexp_object (ecma_string_t *pattern_p, /**< input pattern */
...
+ *
+ * @return ecma_char_t canonicalized character
+ */
+static ecma_char_t __attr_always_inline___
+re_canonicalize (ecma_char_t ch, /**< character */
+                 bool is_ignorecase) /**< IgnoreCase flag */
+{
+  ecma_char_t ret_value = ch;
+
+  if (is_ignorecase)
```
zherczeg: I would add a fast case for ASCII characters (ch <= 127). Could improve the speed a lot.

jerry-core/ecma/operations/ecma-regexp-object.cpp
```diff
@@ -282,8 +319,9 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
           return ecma_make_simple_completion_value (ECMA_SIMPLE_VALUE_FALSE); /* fail */
         }
 
-        ecma_char_t ch1 = (ecma_char_t) re_get_value (&bc_p);
-        ecma_char_t ch2 = lit_utf8_iterator_read_next (&iter);
+        bool is_ignorecase = re_ctx_p->flags & RE_FLAG_IGNORE_CASE;
+        ecma_char_t ch1 = re_canonicalize ((ecma_char_t) re_get_value (&bc_p), is_ignorecase);
+        ecma_char_t ch2 = re_canonicalize (lit_utf8_iterator_read_next (&iter), is_ignorecase);
```
zherczeg: Should affect the compile, not just the match. The compiled regex should only have uppercase characters.

Casesless backreferences are supported correctly?

dbatyai:
@zherczeg, I've updated the patch according to your comments.

zherczeg:
LGTM

Issue #406, opened 2015-07-15
Assertion 'bytes_copied > 0 || !string_len' failed in ecma_builtin_json_quote


renatahodovan:
###### Jerry version:
    Checked revision: abc2b55
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
JSON.stringify('\u2028')
```

###### Backtrace:
```javascript
ICE: Assertion 'bytes_copied > 0 || !string_len' failed at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp(ecma_builtin_json_quote):1137.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x00000000004a0a9e in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
32    SYSCALL_2
(gdb) bt
# 0  0x00000000004a0a9e in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
# 1  0x00000000004003c2 in syscall_2 (syscall_no=62, arg1=27033, arg2=6) at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:96
# 2  0x0000000000400563 in abort () at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:175
# 3  0x0000000000440580 in jerry_fatal(jerry_fatal_code_t) (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:65
# 4  0x00000000004405cf in jerry_assert_fail(char const*, char const*, char const*, unsigned int) (assertion=0x4bb680 "bytes_copied > 0 || !string_len", 
    file=0x4bb108 "jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp", 
    function=0x4a4010 <_ZZL23ecma_builtin_json_quoteP13ecma_string_tE8__func__.lto_priv.798> "ecma_builtin_json_quote", line=1137)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:92
# 5  0x0000000000478a8e in ecma_builtin_json_quote(ecma_string_t*) [clone .lto_priv.796] (string_p=0x7d2cd0 <_ZL13mem_heap_area.lto_priv.1061+272>)
[...]

PR #491, opened 2015-07-29
Fix assertion 'bytes_copied > 0 || !string_len' in JSON.stringify()

rtakacs:
Modified some places to support unicode characters.
Also renamed some variables to be more clearer.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
@@ -917,22 +917,22 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
 {
   ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
 
-  stringify_context_t context_p;
+  stringify_context_t context;
 
   /* 1. */
-  context_p.occurence_stack.block_start_p = NULL;
-  context_p.occurence_stack.block_end_p = NULL;
-  context_p.occurence_stack.current_p = NULL;
+  context.occurence_stack.block_start_p = NULL;
+  context.occurence_stack.block_end_p = NULL;
+  context.occurence_stack.current_p = NULL;
 
   /* 2. */
-  context_p.indent_str_p = ecma_get_magic_string (LIT_MAGIC_STRING__EMPTY);
+  context.indent_str_p = ecma_get_magic_string (LIT_MAGIC_STRING__EMPTY);
 
   /* 3. */
-  context_p.property_list.block_start_p = NULL;
-  context_p.property_list.block_end_p = NULL;
-  context_p.property_list.current_p = NULL;
+  context.property_list.block_start_p = NULL;
+  context.property_list.block_end_p = NULL;
+  context.property_list.current_p = NULL;
 
-  context_p.replacer_function_p = NULL;
+  context.replacer_function_p = NULL;
 
   /* 4. */
   if (ecma_is_value_object (arg2))
@@ -942,7 +942,7 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
     /* 4.a */
     if (ecma_op_is_callable (arg2))
     {
-      context_p.replacer_function_p = obj_p;
+      context.replacer_function_p = obj_p;
     }
     /* 4.b */
     else if (ecma_object_get_class_name (obj_p) == LIT_MAGIC_STRING_ARRAY_UL)
@@ -1013,9 +1013,9 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
         {
           ecma_string_t *item_str_p = ecma_get_string_from_value (item);
 
-          if (!list_has_ecma_string_element (&context_p.property_list, item_str_p))
+          if (!list_has_ecma_string_element (&context.property_list, item_str_p))
           {
-            list_append (&context_p.property_list, item_str_p);
+            list_append (&context.property_list, item_str_p);
           }
           else
           {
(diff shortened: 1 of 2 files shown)


rtakacs:
Solution for #406

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -1117,11 +1117,11 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
     {
       ecma_length_t string_len = ecma_string_get_length (space_str_p);
```
zherczeg: Are you sure this is length not size?

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -1198,23 +1198,24 @@ ecma_builtin_json_quote (ecma_string_t *string_p) /**< string that should be quo
   ecma_string_t *product_str_p = ecma_copy_or_ref_ecma_string (quote_str_p);
   ecma_string_t *tmp_str_p;
 
-  ecma_length_t string_len = ecma_string_get_length (string_p);
+  ecma_length_t string_size = ecma_string_get_size (string_p);
```
zherczeg: I see you changed length to size here. I think you should do this above as well.

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -1130,13 +1130,13 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
 
       for (uint32_t i = 0; i < 10; i++)
       {
-        space_buff[i] = zt_string_buff[i];
+        space_buff[i] = string_buff[i];
       }
```
zherczeg: What happens if the string length is < 10 byte?

rtakacs: There is a condition above
```
if (num_of_chars < 10)
{
  context_p.gap_str_p = ecma_copy_or_ref_ecma_string (space_str_p);
}
else
{
  ...
}
```
If the number of characters are bigger than 10, the first 10 characters will be used.
If the number of characters are less than 10, all the characters will be used.

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
rtakacs: You are right. Fixed.

zherczeg:
LGTM

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -1115,28 +1115,27 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
...
-      {
-        space_buff[i] = zt_string_buff[i];
-      }
+      uint32_t space_buff_size = iter.buf_pos.offset;
+      /* Buffer for the first 10 characters. */
+      MEM_DEFINE_LOCAL_ARRAY (space_buff, space_buff_size, lit_utf8_byte_t);
+      memcpy (space_buff, string_buff, space_buff_size);
 
-      context_p.gap_str_p = ecma_new_ecma_string_from_utf8 ((lit_utf8_byte_t *) space_buff, 10);
+      context_p.gap_str_p = ecma_new_ecma_string_from_utf8 ((lit_utf8_byte_t *) space_buff, space_buff_size);
```
sand1k: Do we really need space_buff?
As I see, we can just call `ecma_new_ecma_string_from_utf8 (string_buff, space_buff_size)`.

rtakacs: hm... You are right, we don't need space_buff.
Updating.

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -1115,28 +1115,21 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
...
-      {
-        space_buff[i] = zt_string_buff[i];
-      }
+      context_p.gap_str_p = ecma_new_ecma_string_from_utf8 (string_buff, iter.buf_pos.offset);
 
-      context_p.gap_str_p = ecma_new_ecma_string_from_utf8 ((lit_utf8_byte_t *) space_buff, 10);
-
-      MEM_FINALIZE_LOCAL_ARRAY (space_buff);
-      MEM_FINALIZE_LOCAL_ARRAY (zt_string_buff);
+      MEM_FINALIZE_LOCAL_ARRAY (string_buff);
```
sand1k: To extract a substirng, it is better to use:
`context_p.gap_str_p = ecma_string_substr (space_str_p, 0, 10)`.

Sorry, noticed just now, that this fragment could be simplified so much.

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -1115,28 +1115,21 @@ ecma_builtin_json_stringify (ecma_value_t this_arg __attr_unused___, /**< 'this'
...
-      /* Buffer for the first 10 characters. */
-      MEM_DEFINE_LOCAL_ARRAY (space_buff, 10, lit_utf8_byte_t);
+      lit_utf8_iterator_t iter = lit_utf8_iterator_create (string_buff, string_size);
+      lit_utf8_iterator_advance (&iter, 10);
 
-      for (uint32_t i = 0; i < 10; i++)
-      {
-        space_buff[i] = zt_string_buff[i];
-      }
+      context_p.gap_str_p = ecma_new_ecma_string_from_utf8 (string_buff, iter.buf_pos.offset);
```
sand1k: And one more issue, `context_p` is not a pointer, could you please rename it to `context`?

rtakacs: Indeed. Of course!

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
rtakacs: nice! FIxed :)

sand1k:
LGTM.

PR #492, opened 2015-07-29
Fix RegExp constructor when called with undefined arguments

dbatyai:
The RegExp constructor does not work correctly in some cases when one of it's arguments is undefined.


egavrin:
Good to me

zherczeg:
LGTM

PR #499, opened 2015-07-30
Remove 'print' instrinsic, add 'print' external handler in main-linux.cpp instead

ruben-ayrapetyan:
// Landing patch for removing 'native call' opcode.


egavrin:
Great! Good to me

galpeter:
there are two commits on this PR.

main-linux.cpp
```diff
@@ -122,17 +122,102 @@ assert_handler (const jerry_api_object_t *function_obj_p __attr_unused___, /** <
...
+        break;
+      }
+      case JERRY_API_DATA_TYPE_OBJECT:
+      {
+        printf ("[Object]");
+        break;
+      }
+      default:
+      {
+        printf ("'<<<argument type is unsupported in current <print> implementation>>>'");
```
galpeter: Would it make sense to print out what type is not supported here?

galpeter: What would the `print(void 0)` print out with this PR?

ruben-ayrapetyan:
Sorry, pull request was closed, because it was based on another branch that is now deleted.

main-linux.cpp
ruben-ayrapetyan: @galpeter, `undefined` is printed with the test.

ruben-ayrapetyan: > Would it make sense to print out what type is not supported here?

I could add this, if you think that it would be better.

ruben-ayrapetyan: I've changed it to the following:
`+        printf ("'<<<argument type (%d) is unsupported in current <print> implementation>>>'", (int) args_p[i].type);`

Is it ok for you?

galpeter: Seems better.
BTW I'm getting the following abort:
```
Assertion 'Code is unreachable' failed at vfprintf (jerry-libc/jerry-libc-printf.c:644
```

with the ` print(11) ` js code

ruben-ayrapetyan: Thank you for the test case.
I also just realized, while debugging another test case, that our `vfprintf` doesn't support printing of floating point values.
How do you think, should we add the support, or leave print of numbers unsupported for now?

galpeter: Well at least I would really like to be able to print out integers. Maybe we can convert everything to int and print out that way for now, and add a FIXME entry for it. Or is it possible to somehow do a toString call on the arguments? Maybe for that the print should be a built-in method (just for easier implementation)?

ruben-ayrapetyan: @galpeter, thank you for the ideas.

I think, implementing as built-in would be good.

@egavrin, what do you think about this?

zherczeg:
Shall we reopen it then?

ruben-ayrapetyan:
@zherczeg, I don't see a way to reopen this PR. I'll open new PR today, and cross-reference the PRs.

ruben-ayrapetyan:
The PR is reopened as #508

PR #500, opened 2015-07-30
Remove intrinsics (native_call opcode)

ruben-ayrapetyan:


egavrin:
Nearly perfect patch! Good to me.

PS should be merged after #499

galpeter:
If I only take into account the intrinsic removal on this branch then it's ok. But before this we need make the `print` function.

egavrin:
`make push`

ruben-ayrapetyan:
Merged (f7781bb7bcfeee5913cf772d3b577e9798040b7f).

PR #503, opened 2015-07-30
Optimize block size calculation in functions working with linked list

sand1k:


egavrin:
Good to me

zherczeg:
What is the exact optimization here?

sand1k:
@zherczeg, this optimisation removes dependency of `linked_list_block_size` on list element's size, so it depends only on constants.

zherczeg:
Ok. LGTM

PR #508, opened 2015-07-31
Remove 'print' instrinsic, add 'print' implementation-defined built-in routine to the Global object

ruben-ayrapetyan:
```
print (undefined);
print (null);
print (true);
print (false);
print (123.23213213);
print (-Infinity);
print (Infinity);
print (NaN);
print (0);
print (-0);
print('str\u0013\u0010 str')
print('str\u0101 str')
print ({});
```

outputs:

```
undefined
null
true
false
123.23213213
-Infinity
Infinity
NaN
0
0
str str
str\u0101 str
[object Object]
```

See also: #499


jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -48,6 +48,96 @@
  */
 
 /**
+ * The implementation-defined Global object's 'print' routine
+ *
+ * The routine converts all of its arguments to strings and outputs them using 'printf'.
+ *
+ * Code points, with except of <NUL> character, that are representable with one utf8-byte are outputted as is, using '%c' format argument,
+ * and other code points are outputted as '\uhhll', where hh and ll are values of code point's high and low bytes, correspondingly.
```
galpeter: These two lines are too long, that's why the precommit fails.

galpeter:
looks good to me

ruben-ayrapetyan:
@galpeter, thank you!

I've updated PR to fix length of the lines.

egavrin:
Good to me

PR #532, opened 2015-08-05
Fix ecma_get_property_descriptor_from_property for accessor properties with [[Get]] / [[Set]] set to undefined value

kkristof:
The standard doesn't prohibit the overwriting of the accessor's member if it is configurable and in the test262 there are 50 tests which fail with assert because of this.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers.cpp
+++ b/jerry-core/ecma/base/ecma-helpers.cpp
@@ -1293,24 +1293,17 @@ ecma_get_property_descriptor_from_property (ecma_property_t *prop_p) /**< proper
   else if (prop_p->type == ECMA_PROPERTY_NAMEDACCESSOR)
   {
     prop_desc.get_p = ecma_get_named_accessor_property_getter (prop_p);
+    prop_desc.is_get_defined = true;
     if (prop_desc.get_p != NULL)
     {
-      prop_desc.is_get_defined = true;
       ecma_ref_object (prop_desc.get_p);
     }
-    else
-    {
-      prop_desc.is_get_defined = false;
-    }
+
     prop_desc.set_p = ecma_get_named_accessor_property_setter (prop_p);
+    prop_desc.is_set_defined = true;
     if (prop_desc.set_p != NULL)
     {
-      prop_desc.is_set_defined = true;
       ecma_ref_object (prop_desc.set_p);
-    }
-    else
-    {
-      prop_desc.is_set_defined = false;
     }
   }
 
--- a/tests/jerry/object-defineproperty.js
+++ b/tests/jerry/object-defineproperty.js
@@ -38,3 +38,20 @@ try {
     assert (desc.value === 2010);
     assert (typeof (desc.get) === 'undefined');
 }
+
+obj = {};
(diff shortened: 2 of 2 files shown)


jerry-core/ecma/operations/ecma-conversion.cpp
```diff
@@ -489,9 +489,6 @@ ecma_op_from_property_descriptor (const ecma_property_descriptor_t* src_prop_des
   else
   {
     // 4.
-    JERRY_ASSERT (src_prop_desc_p->is_get_defined
-                  || src_prop_desc_p->is_set_defined);
```
ruben-ayrapetyan: @kkristof, the assertion is specification-defined.

http://www.ecma-international.org/ecma-262/5.1/#sec-8.10.4, item 4:
`IsAccessorDescriptor(Desc) must be true`

It seems to me that actual issue is in `ecma_get_property_descriptor_from_property` function: instead of putting `undefined` value to [[Get]] and [[Set]] fields, the function sets the fields as undefined.

kkristof: @ruben-ayrapetyan I tried to set `undefined` value for the [[Get]] / [[Set]] field in the `ecma_get_property_descriptor_from_property` but I got another assert when I tried to create object from the `undefined` value (the `ecma_property_descriptor_t`'s  `get_p` and `set_p` are `ecma_object_t*`). Could you give any hint how this should be done?

kkristof: Is it a valid solution to set the `get_p`/`set_p` to `NULL` and the `is_get_defined`/`is_set_defined` to `true`? IMHO I don't like this kind of hacks.

ruben-ayrapetyan: Yes, it's valid solution.

This is not a hack, but the only right way to represent 'undefined' value of defined [[Get]] in ecma_property_descriptor_t, according to current design of the structure.

If you think that this should be implemented in another way, please describe the new design here or in a new issue, for discussion.

kkristof:
Hi, I have updated the branch.

jerry-core/ecma/operations/ecma-conversion.cpp
kkristof: Okay than. I don't have better idea.

ruben-ayrapetyan:
@kkristof, the changes look good to me.

Please, update commit message and title of the pull request.

kkristof:
Hi, I have already changed the commit message and the title of the PR. It is good now or should I change it?

ruben-ayrapetyan:
@kkristof, I missed the change, sorry.

I mean commit message like "Fix `ecma_get_property_descriptor_from_property` for accessor properties with [[Get]] / [[Set]] set to `undefined` value", as assertion in `ecma_op_from_property_descriptor` is not changed.

kkristof:
I have updated the message and the title.

ruben-ayrapetyan:
@kkristof, thank you!

Looks good to me.

LaszloLango:
LGTM

PR #554, opened 2015-08-10
Fix RegExp literal parsing

dbatyai:


jerry-core/parser/js/lexer.cpp
```diff
@@ -1294,16 +1290,16 @@ lexer_parse_token (void)
...
+      && !(prev_non_lf_token.type == TOK_NAME
+           || prev_non_lf_token.type == TOK_NULL
+           || prev_non_lf_token.type == TOK_BOOL
+           || prev_non_lf_token.type == TOK_CLOSE_BRACE
+           || prev_non_lf_token.type == TOK_CLOSE_SQUARE
+           || prev_non_lf_token.type == TOK_CLOSE_PAREN
+           || prev_non_lf_token.type == TOK_SMALL_INT
+           || prev_non_lf_token.type == TOK_NUMBER
+           || prev_non_lf_token.type == TOK_STRING
+           || prev_non_lf_token.type == TOK_REGEXP))
```
sand1k: `TOK_CLOSE_PAREN` and `TOK_CLOSE_BRACE` could precede regular expression, for example:
```
if (true) /abc/.exec("abc");
if (true) {} /abc/.exec("abc");
```
This case is harder to support, possibly, we could implement it in another patch.

jerry-core/parser/js/lexer.cpp
```diff
@@ -1482,6 +1478,7 @@ lexer_next_token (void)
   if (!is_empty (saved_token))
   {
     sent_token = saved_token;
+    prev_non_lf_token = sent_token;
```
sand1k: `prev_non_lf_token` should be tracked more carefully. It seems, that it should be changed in `lexer_seek` and `lexer_save_token` functions.
For example, the following code:
```
{for (;false;/abc/.exec("abc")) (5)}
```
fails with syntax error due to `prev_non_lf_token` equals to `TOK_CLOSE_BRACE` after call to `lexer_seek` (parser.cpp:1920).

dbatyai:
@sand1k, I've updated the patch, it should work correctly now. Please check.

jerry-core/parser/js/lexer.cpp
dbatyai: I agree.

sand1k: Another case:
```
var a
/abc/.exec("abc");
```
Could you, please, create an issue for the mentioned tests?

sand1k:
LGTM.

zherczeg:
LGTM

jerry-core/parser/js/lexer.cpp
dbatyai: Issue can be found here: #563

PR #571, opened 2015-08-13
Replace ecma_create_object with high-level Object and Array construction interfaces in JSON built-in implementation

ruben-ayrapetyan:


egavrin:
Good to me

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
```diff
@@ -597,34 +598,15 @@ ecma_builtin_json_parse_value (ecma_json_token_t *token_p) /**< token argument *
...
-          *length_num_p = ecma_uint32_to_number (length);
-
-          ecma_property_t *length_prop_p = ecma_create_named_data_property (array_p,
-                                                                            length_magic_string_p,
-                                                                            true,
-                                                                            false,
-                                                                            false);
-          ecma_set_named_data_property_value (length_prop_p, ecma_make_number_value (length_num_p));
-          ecma_deref_ecma_string (length_magic_string_p);
-
```
zherczeg: Who sets the length?

zherczeg:
Is this a performance optimization or what is the purpose of the change?

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
ruben-ayrapetyan: `ecma_op_create_array_object` and `ecma_op_array_object_define_own_property`.

zherczeg: Tricky. There was no reference to ecma_op_array_object_define_own_property in the patch, so this is a hidden feature.

ruben-ayrapetyan:
> Is this a performance optimization or what is the purpose of the change?

Purpose of the change is replacement direct usage of `ecma_create_object` with high-level object constructors where possible. It is a landing patch for optimization of object's prototype initialization and storage.

zherczeg:
Fair enough, LGTM

jerry-core/ecma/builtin-objects/ecma-builtin-json.cpp
ruben-ayrapetyan: If I understand correctly, ArrayLiteral in JSON should be parsed exactly as ArrayLiteral in source code (11.1.4, Array Initialiser). So, I've just moved implementation nearer to specification-defined algorithm that uses `[[DefineOwnProperty]]`.

PR #586, opened 2015-08-18
Fix 'control reaches end of non-void function' compile error for 'noreturn'

seanshpark:

### DIFF
--- a/jerry-core/jrt/jrt-fatals.cpp
+++ b/jerry-core/jrt/jrt-fatals.cpp
@@ -68,6 +68,11 @@ jerry_fatal (jerry_fatal_code_t code) /**< status code */
   {
     exit (code);
   }
+
+  /* to make compiler happy for some RTOS: 'control reaches end of non-void function' */
+  while (true)
+  {
+  }
 } /* jerry_fatal */
 
 /**


egavrin:
`__builtin_unreachable();` didn't work?

seanshpark:
sorry but cannot find `__builtin_unreachable()`. 
you mean to add `__builtin_unreachable` function with `while(true) {}` ?

zherczeg:
What I don't understand is: this function has actually returns with void

seanshpark:
@zherczeg , can you describe more about `this function` ?

zherczeg:
It is called when there is a fatal error, and the engine should stop. The function is void.
Why the compiler says it is a non-void function???

seanshpark:
I think not the compiler(mips-sde-elf-gcc), but the function `abort()` or `exit()` in the RTOS I'm working on may return. I do not have the source code thus cannot confirm, and cause of same reason, cannot fix that platform.
I don't like this approach, fixing JerryScript, but for now I can't think of better one.

seanshpark:
Accept or Reject?

ruben-ayrapetyan:
Looks good to me

zherczeg:
lgtm

seanshpark:
Thank you :)

Issue #585, opened 2015-08-18
Assertion 'instrs_equal (instrs_p, instrs, 5)' failed with 32bit Linux


seanshpark:
I get below error with `make precommit` on Ubuntu Linux 32bit
```
LITERALS:
0x8083514 [  8] a : STRING
0x808351c [ 36]  : EMPTY RECORD
  0:                 meta   12    6  255     // [no 'arguments'] [no 'eval']
  1:         reg_var_decl  128  130          // var tmp128 .. tmp130;
  2:             var_decl    0               // var a;
  3:           assignment  130    1    1     // tmp130 = 1: SMALLINT;
  4:           assignment    0    6  130     // a = tmp130 : TYPEOF(tmp130);
  5:                  ret                    // ret;
ICE: Assertion 'instrs_equal (instrs_p, instrs, 5)' failed at /.../harmony/jerryscript/tests/unit/test-parser.cpp(main):74.
Error: ERR_FAILED_INTERNAL_ASSERTION
```
in `./build/bin/unittests/unit_tests_run.log` file.

console output is
```
Running ./build/bin/unittests/unit-test-number-to-string... OK
Running ./build/bin/unittests/unit-test-parser... FAILED
Unit tests run failed. See ./build/bin/unittests/unit_tests_run.log for details.
make[1]: *** [unittests_run] Error 1
make[1]: Leaving directory `/.../harmony/jerryscript'
make: *** [precommit] Error 1
```

64bit is OK.

PR #588, opened 2015-08-18
Remove comparison of unused instruction operands fields (the fields could be uninitialized)

ruben-ayrapetyan:
Related issue: #585


seanshpark:
Good, #585 is fixed with this.

ruben-ayrapetyan:
@seanshpark, thank you!

egavrin:
good to me

zherczeg:
LGTM. What is the effect on the binary size?

ruben-ayrapetyan:
@zherczeg, the change is only in a unit test, so it doesn't influence binary size of the engine's core.

PR #592, opened 2015-08-19
Fix perf scripts

egavrin:
Current scripts doesn't allow to pass engine with flags as arguments - this is the fix.

Now it's possible to do the following:
`./tools/perf.sh 5 "./jr-master --parse-only"`


ruben-ayrapetyan:
Looks good to me

zherczeg:
LGTM

Issue #531, opened 2015-08-05
Function.prototype.bind: incorrect behaviour of ecma_gc_mark()


bzsolt:
Related discussion: [#513 (comment)](https://github.com/Samsung/jerryscript/pull/513#issuecomment-127973747)

bzsolt:
Fixed by #599

PR #599, opened 2015-08-26
Function.prototype.bind: incorrect behaviour of ecma_gc_mark()

bzsolt:
Related issue: #531

### DIFF
--- a/jerry-core/ecma/base/ecma-gc.cpp
+++ b/jerry-core/ecma/base/ecma-gc.cpp
@@ -351,13 +351,47 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
             case ECMA_INTERNAL_PROPERTY_NON_INSTANTIATED_BUILT_IN_MASK_0_31: /* an integer (bit-mask) */
             case ECMA_INTERNAL_PROPERTY_NON_INSTANTIATED_BUILT_IN_MASK_32_63: /* an integer (bit-mask) */
             case ECMA_INTERNAL_PROPERTY_REGEXP_BYTECODE:
-            case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_TARGET_FUNCTION:
-            case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_BOUND_THIS:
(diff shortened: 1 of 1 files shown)


jerry-core/ecma/base/ecma-gc.cpp
```diff
@@ -352,9 +352,43 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
             case ECMA_INTERNAL_PROPERTY_NON_INSTANTIATED_BUILT_IN_MASK_32_63: /* an integer (bit-mask) */
             case ECMA_INTERNAL_PROPERTY_REGEXP_BYTECODE:
             case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_TARGET_FUNCTION:
```
ruben-ayrapetyan: Object in a [[TargetFunction]] property also need to be marked as visited.

jerry-core/ecma/base/ecma-gc.cpp
```diff
@@ -352,9 +352,43 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
...
+            {
+              break;
+            }
+
             case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_BOUND_THIS:
```
ruben-ayrapetyan: Please, add comments near cases, related to bound functions, about what can be contained in the property.
For example, `/* an ecma-value */` for `case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_BOUND_THIS:`.

jerry-core/ecma/base/ecma-gc.cpp
bzsolt: You're right, I've fixed it. Thanks.

ruben-ayrapetyan:
Looks good to me

zherczeg:
LGTM. Is it possible to add a test?

bzsolt:
I think it isn't possible, because we can't externally control the behaviour of the GC.

egavrin:
@bzsolt I think we

bzsolt:
@egavrin Can we merge this PR?

ruben-ayrapetyan:
See also: #622

PR #602, opened 2015-08-27
Add support for memory measure with Valgrind

rtakacs:
Added malloclike and freelike Valgrind macros to the custom memory allocator.

### DIFF
--- a/jerry-core/mem/mem-heap.cpp
+++ b/jerry-core/mem/mem-heap.cpp
@@ -40,12 +40,14 @@
 #ifdef JERRY_VALGRIND
 # include "memcheck.h"
 
-# define VALGRIND_NOACCESS_STRUCT(s)    (void)VALGRIND_MAKE_MEM_NOACCESS((s), sizeof (*(s)))
-# define VALGRIND_UNDEFINED_STRUCT(s)   (void)VALGRIND_MAKE_MEM_UNDEFINED((s), sizeof (*(s)))
-# define VALGRIND_DEFINED_STRUCT(s)     (void)VALGRIND_MAKE_MEM_DEFINED((s), sizeof (*(s)))
-# define VALGRIND_NOACCESS_SPACE(p, s)  (void)VALGRIND_MAKE_MEM_NOACCESS((p), (s))
-# define VALGRIND_UNDEFINED_SPACE(p, s) (void)VALGRIND_MAKE_MEM_UNDEFINED((p), (s))
-# define VALGRIND_DEFINED_SPACE(p, s)   (void)VALGRIND_MAKE_MEM_DEFINED((p), (s))
+# define VALGRIND_NOACCESS_STRUCT(s)     (void)VALGRIND_MAKE_MEM_NOACCESS((s), sizeof (*(s)))
+# define VALGRIND_UNDEFINED_STRUCT(s)    (void)VALGRIND_MAKE_MEM_UNDEFINED((s), sizeof (*(s)))
+# define VALGRIND_DEFINED_STRUCT(s)      (void)VALGRIND_MAKE_MEM_DEFINED((s), sizeof (*(s)))
+# define VALGRIND_NOACCESS_SPACE(p, s)   (void)VALGRIND_MAKE_MEM_NOACCESS((p), (s))
+# define VALGRIND_UNDEFINED_SPACE(p, s)  (void)VALGRIND_MAKE_MEM_UNDEFINED((p), (s))
+# define VALGRIND_DEFINED_SPACE(p, s)    (void)VALGRIND_MAKE_MEM_DEFINED((p), (s))
+# define VALGRIND_MALLOCLIKE_SPACE(p, s) (void)VALGRIND_MALLOCLIKE_BLOCK((p), (s), 0, 0)
+# define VALGRIND_FREELIKE_SPACE(p)            VALGRIND_FREELIKE_BLOCK((p), 0)
 #else /* JERRY_VALGRIND */
 # define VALGRIND_NOACCESS_STRUCT(s)
 # define VALGRIND_UNDEFINED_STRUCT(s)
@@ -53,6 +55,8 @@
 # define VALGRIND_NOACCESS_SPACE(p, s)
 # define VALGRIND_UNDEFINED_SPACE(p, s)
 # define VALGRIND_DEFINED_SPACE(p, s)
+# define VALGRIND_MALLOCLIKE_SPACE(p, s)
+# define VALGRIND_FREELIKE_SPACE(p)
 #endif /* JERRY_VALGRIND */
 
 /**
@@ -691,6 +695,7 @@ mem_heap_alloc_block_try_give_memory_back (size_t size_in_bytes, /**< size of re
 
   if (likely (data_space_p != NULL))
   {
+    VALGRIND_MALLOCLIKE_SPACE (data_space_p, size_in_bytes);
     return data_space_p;
   }
 
@@ -704,6 +709,7 @@ mem_heap_alloc_block_try_give_memory_back (size_t size_in_bytes, /**< size of re
 
     if (data_space_p != NULL)
     {
+      VALGRIND_MALLOCLIKE_SPACE (data_space_p, size_in_bytes);
       return data_space_p;
     }
   }
@@ -769,6 +775,7 @@ void
 mem_heap_free_block (void *ptr) /**< pointer to beginning of data space of the block */
 {
   uint8_t *uint8_ptr = (uint8_t*) ptr;
+  VALGRIND_FREELIKE_SPACE (ptr);
 
   /* checking that uint8_ptr points to the heap */
   JERRY_ASSERT (uint8_ptr >= mem_heap.heap_start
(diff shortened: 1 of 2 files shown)


ruben-ayrapetyan:
Is it ok to mark malloc-like a block inside another malloc-like marked block?
Pools are allocated using heap allocator, so the whole pool area would be already marked using the valgrind hint, so it would be marked again upon pool chunk allocation.

zherczeg:
The logic here is different. We mark only those bytes, which are available in the user space (not from allocator perspective). Valgrind did not complain, and the results were similar what the mem stat feature provides.

ruben-ayrapetyan:
Could you, please, provide an example on how to start `valgrind` to see the measured values?

jerry-core/mem/mem-heap.cpp
```diff
@@ -40,12 +40,14 @@
...
 # include "memcheck.h"
 
-# define VALGRIND_NOACCESS_STRUCT(s)    (void)VALGRIND_MAKE_MEM_NOACCESS((s), sizeof (*(s)))
-# define VALGRIND_UNDEFINED_STRUCT(s)   (void)VALGRIND_MAKE_MEM_UNDEFINED((s), sizeof (*(s)))
-# define VALGRIND_DEFINED_STRUCT(s)     (void)VALGRIND_MAKE_MEM_DEFINED((s), sizeof (*(s)))
-# define VALGRIND_NOACCESS_SPACE(p, s)  (void)VALGRIND_MAKE_MEM_NOACCESS((p), (s))
-# define VALGRIND_UNDEFINED_SPACE(p, s) (void)VALGRIND_MAKE_MEM_UNDEFINED((p), (s))
-# define VALGRIND_DEFINED_SPACE(p, s)   (void)VALGRIND_MAKE_MEM_DEFINED((p), (s))
+# define VALGRIND_NOACCESS_STRUCT(s)     (void)VALGRIND_MAKE_MEM_NOACCESS((s), sizeof (*(s)))
+# define VALGRIND_UNDEFINED_STRUCT(s)    (void)VALGRIND_MAKE_MEM_UNDEFINED((s), sizeof (*(s)))
```
ruben-ayrapetyan: Why do we need the `(void)` parts? With the parts, the code doesn't compile in my configuration.

Command line I've used: `make VERBOSE=1 VALGRIND=ON debug.linux -j`

ruben-ayrapetyan:
If malloc-like requests are not removed from pool, I get the following valgrind messages for `tests/jerry/array.js`:
```
...
==7044== 2502 errors in context 147 of 151:
==7044== Invalid read of size 1
==7044==    at 0x4CAD75: ecma_find_named_property(ecma_object_t*, ecma_string_t*) (ecma-helpers.cpp:608)
==7044==    by 0x49FE31: ecma_op_general_object_get_own_property(ecma_object_t*, ecma_string_t*) (ecma-objects-general.cpp:202)
==7044==    by 0x4A7B12: ecma_op_object_get_own_property_longpath(ecma_object_t*, ecma_string_t*) [clone .30508.2910] (ecma-objects.cpp:120)
==7044==    by 0x4A7D9E: ecma_op_object_get_own_property(ecma_object_t*, ecma_string_t*) (ecma-objects.cpp:177)
==7044==    by 0x49FEC3: ecma_op_general_object_get_property(ecma_object_t*, ecma_string_t*) (ecma-objects-general.cpp:224)
==7044==    by 0x4A7E49: ecma_op_object_get_property(ecma_object_t*, ecma_string_t*) (ecma-objects.cpp:218)
==7044==    by 0x4972AE: ecma_op_has_binding(ecma_object_t*, ecma_string_t*) (ecma-lex-env.cpp:131)
==7044==    by 0x496F1F: ecma_op_resolve_reference_base(ecma_object_t*, ecma_string_t*) (ecma-reference.cpp:49)
==7044==    by 0x51641A: set_variable_value(vm_frame_ctx_t*, unsigned short, unsigned char, unsigned int) (opcodes-helpers-variables.cpp:166)
==7044==    by 0x5260D7: opfunc_assignment(vm_instr_t, vm_frame_ctx_t*) (opcodes.cpp:101)
==7044==    by 0x52D926: vm_loop(vm_frame_ctx_t*, vm_run_scope_t*) (vm.cpp:484)
==7044==    by 0x52DD8B: vm_run_from_pos(vm_instr_t const*, unsigned short, unsigned int, ecma_object_t*, bool, bool) (vm.cpp:574)
==7044==  Address 0x86aa61 is 9 bytes after a recently re-allocated block of size 8 alloc'd
==7044==    at 0x4C5E8D: ecma_alloc_number() (mem-poolman.cpp:526)
==7044==    by 0x4C44F9: ecma_copy_value(unsigned int, bool) (ecma-helpers-value.cpp:375)
==7044==    by 0x4C51C2: ecma_append_to_values_collection(ecma_collection_header_t*, unsigned int, bool) (ecma-helpers-values-collection.cpp:185)
==7044==    by 0x479DD7: vm_fill_varg_list(vm_frame_ctx_t*, unsigned int, ecma_collection_header_t*) (opcodes-varg.cpp:57)
==7044==    by 0x51AD95: opfunc_array_decl(vm_instr_t, vm_frame_ctx_t*) (opcodes.cpp:954)
==7044==    by 0x52D926: vm_loop(vm_frame_ctx_t*, vm_run_scope_t*) (vm.cpp:484)
==7044==    by 0x52DD8B: vm_run_from_pos(vm_instr_t const*, unsigned short, unsigned int, ecma_object_t*, bool, bool) (vm.cpp:574)
==7044==    by 0x52D6F2: vm_run_global() [clone .3196] (vm.cpp:413)
==7044==    by 0x5294E3: jerry_run.3798 (jerry.cpp:1445)
==7044==    by 0x5278B9: main (main-linux.cpp:295)
...
```

jerry-core/mem/mem-heap.cpp
zherczeg: Well, void was in the original code. I don't mind getting rid of it.

zherczeg:
Yes, you are right. You know the allocator better than us. Could you help us fixing this patch?

jerry-core/mem/mem-heap.cpp
ruben-ayrapetyan: Thank you for pointing this out. I've missed that.

Anyway, without removing the `(void)` the code isn't compiled successfully with the changes.

ruben-ayrapetyan:
@zherczeg, could you, please, describe valgrind's mechanism you want to use for measurements and measurements you want to get?

zherczeg:
We want to track user (non-allocator) memory allocations. I.e. capturing the results of malloc like requests (malloc property, malloc double, malloc local buffer, etc.). We would like to capture the free operations as well, regardless they are performed by GC or manual free (local buffers). We are not interested how the allocator works (pools, etc.), that is a black box from our perspective.

ruben-ayrapetyan:
@zherczeg, I see. Could you, please, clarify how the measured value should differ form `allocated bytes` value, measured in mem-stats mode? Or should the values be the same?

zherczeg:
Yes, it would be good it they would be the same. On simple tests, they are.

ruben-ayrapetyan:
@zherczeg, we could add malloc-like / free-like valgrind requests to heap allocator, but not to pool allocator, as pools are on heap. So, it doesn't seem that the way makes possible separate profiling of pool allocation requests.

May be, it is better to implement own valgrind client requests for specific allocation types.
What do you think about this?

By the way, why do we need valgrind profiling, if it provides the same, as mem-stats mode?

zherczeg:
Could you give a short overview about heaps and pools? As far as I know Jerry has a 256 KByte array for storing data. Is this the heap? Is this divided to pools?

Our Freya tool ( https://github.com/szeged/Freya , http://webkit.sed.hu/node/29 ) provides more detailed information about the memory management of the program that mem-stats. Unfortunately Jerry does not support malloc / free at this moment, so we need to provide allocation information to the tool using these client requests. Or adding malloc / free support to Jerry if you prefer that.

ruben-ayrapetyan:
@zherczeg, the array is the heap.

Heap is divided into fixed-sized heap chunks.

Heap can allocate blocks of arbitrary sizes.
Upon a heap allocation request, several contiguous heap chunks with total size greater or equal to size in allocation request, are combined into allocated "block" and returned as response to the allocation request. Upon free request, all chunks of freed block are marked free.
<br>
Pool manager allocates chunks of fixed size (the size is compile-time constant).

Upon an allocation request, pool manager takes a free pool chunk, if there is any, change its state to "allocated" and returns as response to request. Upon free request, state of pool chunk is changed to "free". In the cases, no requests to heap allocator are performed by pool allocator.

If there is no any free chunk upon pool allocation request, pool manager requests a block from heap allocator and then, internally, divides it into free pool chunks.
<br>
So, any allocated pool chunk passes two allocation stages: allocation of a chunks pool on heap, and allocation of a particular chunk from the pool.

zherczeg:
This sounds good. The pool allocation is memory manager stuff, so it should not be reported to valgrind. How can we detect this?

ruben-ayrapetyan:
As far as I understand, Freya would associate the memory allocation profile with call stacks, is it correct?

We can just add malloc-like requests to heap allocator, without adding them to pools allocator.

But, in the case, we would get incorrect profile information for places that call pool allocator:
 - most of them would seem to not allocate anything;
 - and several (actually, random) places would seem to allocate large blocks (allocated for pools).

Seems that one of ways to solve the issue is to support custom valgrind requests for handling various allocation types properly. Could this be performed?

zherczeg:
I opened a new pull request:
https://github.com/Samsung/jerryscript/pull/611

PR #607, opened 2015-08-30
Enable build on Mac OS X with clang

lemmaa:
* Add build target 'darwin'
* Modify compiler options to work with clang.
* Support 'jerry-libc' on Mac OS X

For MPU target build, install `gcc-arm-none-eabi` tool chain using `Homebrew` as following.
````
brew tap PX4/homebrew-px4
brew update
brew install gcc-arm-none-eabi
````
https://pixhawk.org/dev/toolchain_installation_mac


lemmaa:
Can't pass the precommit test on build machine, but I can't reproduce it.
So I close this PR, and create another PR https://github.com/Samsung/jerryscript/pull/609 with new branch.

PR #611, opened 2015-08-31
Add support for reporting mallocs / frees to valgrind through client requests.

zherczeg:


zherczeg:
It seems this patch is working, no more issues with memcheck / Freya.

jerry-core/mem/mem-poolman.cpp
```diff
@@ -430,12 +436,16 @@ mem_pools_alloc_longpath (void)
   JERRY_ASSERT (MEM_POOL_SIZE <= mem_heap_get_chunked_block_data_size ());
   JERRY_ASSERT (MEM_POOL_CHUNKS_NUMBER >= 1);
 
+  MEM_HEAP_VALGRIND_MEMPOOL_REQUEST_START ();
   mem_pool_chunk_t *pool_start_p = (mem_pool_chunk_t*) mem_heap_alloc_chunked_block (MEM_HEAP_ALLOC_LONG_TERM);
```
ruben-ayrapetyan: The allocation could invoke GC that can free heap block. The free wouldn't be accounted, as the request appears to be performed from pool allocator. For valgrind this appears as memory leak:

 LEAK SUMMARY:
    definitely lost: 256 bytes in 1 blocks
 ...

Example code:
// make VALGRIND=ON debug.linux
//  a26c45421988d138a01408ecfefb80816e3bcf81 with merged 8f44fa70253a9270ff4b2cd58be12b91004087b4
// valgrind ./build/bin/debug.linux/jerry test.js
// allocates RegExp object that stores byte-code on heap and is freed by GC during a call to `f`
```js
new RegExp ("foo", undefined);

function f()
{
  function g()
  {
  }

  return g;
}

for (i = 0; i < 100; i++)
{
  f ();
}
```

zherczeg:
Any suggestion to fix it? Is there a maximum nesting level (I only saw two as maximum)? If two is the maximum, we could track the types.

I am happy that we are getting closer to do this :) Today I saw some nice mem info about access-binary-trees.js.

zherczeg:
I had an idea: moved the pool manager request to a local variable. This fixes the issue you reported.

ruben-ayrapetyan:
@zherczeg, I think that current version of changes is correct.

Could we remove the `valgrind_mempool_request` global variable?
Maybe, by introducing special heap interfaces for pool allocator.

Also, as far as I understand, `MALLOCLIKE`- and `FREELIKE`-requests can change valgrind's validity and addressable bits. If it is so, what are the new values?
Wouldn't this override value that are currently set using `VALGRIND_NOACCESS_SPACE` / `VALGRIND_UNDEFINED_SPACE` / `VALGRIND_DEFINED_SPACE` macro?

Also, what would we do, if for example, there would be necessary to separately account memory, allocated in recordset-based storages, as we currently trying to do for pool allocator?

zherczeg:
I would not want to introduce special interfaces (nor extra arguments for existing functions) for mempool just for debuging like features. It would increase code size for nothing (duplicate function bodies). Static global variables are used in Jerry so I don't really see any problem with that.

Client requests are processed by a tool and I don't know what memcheck does. But if you feel this can be a problem, I could introduce a separate define for enabling freay support. Which one you prefer: JERRY_FREYA or JERRY_VALGRIND_FREYA?

What are these recordset based storages? We can probably adopt this feature later them if it is needed.

ruben-ayrapetyan:
> I would not want to introduce special interfaces (nor extra arguments for existing functions) for mempool just for debuging like features. It would increase code size for nothing (duplicate function bodies).

The function could be just inlined in `!JERRY_VALGRIND` mode, so there should be no issues with binary size, isn't it? 

> Static global variables are used in Jerry so I don't really see any problem with that.

In this case, the new global variable introduces a non-trivial context (in debug mode, not introducing it in release). The context should be supported during any future changes of allocator. It seems that introducing explicit interface for the pool allocator's requests would make design more reliable.

> Client requests are processed by a tool and I don't know what memcheck does. But if you feel this can be a problem, I could introduce a separate define for enabling freay support. Which one you prefer: JERRY_FREYA or JERRY_VALGRIND_FREYA?

I would prefer adding `JERRY_VALGRIND_FREYA` with `VALGRIND_FREYA_ACCOUNT_ALLOC` / `VALGRIND_FREYA_ACCOUNT_FREE`-like macro in the mode.

Also, if Freya correctly accounts allocation of block in memory area, which is already marked as allocated block (like pool on heap), we could just use the macro without any additional global variables, or interfaces.

If Freya would additionally support two additional client request types like "pass block ownership to another allocator" (with "allocator id" and "block address" pointers) / "restore ownership of block" (with "block address" pointers), we could also provide per-allocator statistics.

I see that currently Freya uses two valgrind-defined client requests, in `fr_handle_client_request` (fr_main.c). Maybe, we could just add the two new client requests to "freya.h" header like memcheck adds its own headers, and handle them in the `fr_handle_client_request` function.

Maybe, there are also another, more suitable, designs based on introducing new Freya client requests. What do you think about this?

> What are these recordset based storages? We can probably adopt this feature later them if it is needed.

The recordset-based storages are allocators of non-continous memory areas. This could be considered as one another allocator (based on heap, in the same way like the pool allocator is). Maybe, separate accounting of related memory allocations would be necessary.

zherczeg:
Am I understand correctly that he allocator would look like this:
```
#ifdef JERRY_VALGRIND_FREYA
// Duplicate mem_heap_free_block in the cpp file.
void mem_heap_mempool_free_block(void *p);
#else
void mem_heap_mempool_free_block(void *p) { mem_heap_free_block(p); }
#endif
```

> Maybe, there are also another, more suitable, designs based on introducing new Freya client requests.
> What do you think about this?

I really don't want to go this way. Freya is not designed for allocator efficiency measurement. It is designed for user memory analysis. For example, the tool can reveal that scopes_tree_add_op_meta allocates too much memory, but it doesn't matter which allocator handles this (and how). From the optimization point of view, this information is enough to direct our focus on this function, and invent something to improve this.

From user program perspective we allocate (reallocate) and free memory, and that is all. And the current client requests satisfy this. If we wouldn't have a custom allocator, we could even forget about this. Freya also groups allocations according to a configuration file (grouping is NOT controlled by the executed program), but its purpose ends here.

Anyway, if start addresses are different, Freya can track the allocations independently (it doesn't matter they overlap or not). So if the pool allocator allocates 1024bytes, and (1024/8)-1 8 byte long blocks are returned from pool allocator (except the first 8 byte block of the range), it will report 1024+1024-8 byte allocation. But from user point of view, this is a false data, since the actual memory allocation is 1024-8 bytes.

egavrin:
How can we ensure that results reported by Freya is correct?

zherczeg:
Easy. Printf. I did it to check that the byte-code is really that big. See:
https://github.com/Samsung/iotjs/issues/212

egavrin:
I'll check, thank you!

Maybe we should `invent` mechanism to double check our results, like to compare result from `mem_stats`, Freya and RSS. Otherwise, we may miss really misleading issue.

ruben-ayrapetyan:
> Am I understand correctly that he allocator would look like this:

> #ifdef JERRY_VALGRIND_FREYA
> // Duplicate mem_heap_free_block in the cpp file.
> void mem_heap_mempool_free_block(void *p);
> #else
> void mem_heap_mempool_free_block(void *p) { mem_heap_free_block(p); }
> #endif

@zherczeg, no, I didn't mean duplication of the function. I've missed the case about turning off `FREELIKE_SPACE` request.

So, if the issue couldn't be solved through special `no-track` interfaces without code duplication for now, let's leave the version with global variable for now.

Could you, please, change naming from `valgrind_mempool_request`-like to `valgrind_freya_no_track_next_request`, `VALGRIND_FREYA_IS_TRACK_REQUEST`, etc.?
In the case, later we could use the same interface for other heap-based second level allocators, if it would be necessary.

Could you, please, also change the client request so that memcheck / freya modes would be mutually exclusive (through defines like `JERRY_VALGRIND_MEMCHECK` and `JERRY_VALGRIND_FREYA`, so that `NOACCESS_SPACE` client requests etc. would not interfere with Freya requests), or maybe make certain that `MALLOCLIKE_SPACE` and `FREELIKE_SPACE` in memcheck mode set correct `NOACCESS` / `UNDEFINED` / `DEFINED` states for memory areas and so `NOACCESS_SPACE`-like requests could be removed for data spaces?

Also, maybe it would be better to track allocation not in `mem_heap_alloc_block_try_give_memory_back`, but in `mem_heap_alloc_block` and `mem_heap_alloc_chunked_block` so the `mem_heap_alloc_block_try_give_memory_back` function would not occur in call stack. What do you think about this?

egavrin:
> Easy. Printf. I did it to check that the byte-code is really that big.

@zherczeg I mean, systematically, any change in allocator may broke Freya support.

zherczeg:
I think I fixed everything, patch is up for review.

egavrin:
Good to me

galpeter:
@zherczeg what should we do if we enable VALGIND_FREYA but disable VALGRIND? Should we write out a message that it won't be used?

egavrin:
@zherczeg @galpeter for me `VALGIND_FREYA` should override `VALGRIND` flag.

zherczeg:
Well, these two flags cannot be combined. So you need two binaries if you want use valgrind and freya as well. Currently valgrind memcheck is confused when freya is enabled. I can add a compile time error if both are enabled.

egavrin:
@zherczeg compile time error is OK for me.

egavrin:
@zherczeg do you plan to land it soon?

zherczeg:
Patch updated with the check. If you accept it, I land it.

egavrin:
Great! 'make push'

zherczeg:
Thx.
Landed in c81c73012918f5153727405900d78311a4fba34a

PR #616, opened 2015-09-03
Change internal encoding of strings to CESU-8

dbatyai:
Use CESU-8 encoding internally for strings instead of UTF-8. This simplifies handling of strings, since in this case we don't need special handling for surrogate pairs.
This is still work in progress.

Current status (measured on x86 linux):

                               Benchmark |         RSS<br>(+ is better) |        Perf<br>(+ is better) |
                               --------- |                          --- |                         ---- |
                              3d-cube.js | 156->152 (2.564) | 0.429->0.417 (2.797) | 
                  access-binary-trees.js | 96->92 (4.167) | 0.382->0.369 (3.403) | 
                      access-fannkuch.js | 44->48 (-9.091) | 1.111->1.053 (5.221) | 
                         access-nbody.js | 68->68 (0.000) | 0.617->0.565 (8.428) | 
             bitops-3bit-bits-in-byte.js | 44->44 (0.000) | 0.407->0.39 (4.177) | 
                  bitops-bits-in-byte.js | 40->36 (10.000) | 0.519->0.509 (1.927) | 
                   bitops-bitwise-and.js | 32->36 (-12.500) | 0.438->0.433 (1.142) | 
                controlflow-recursive.js | 292->288 (1.370) | 0.384->0.384 (0.000) | 
                    date-format-xparb.js | 108->104 (3.704) | 0.422->0.26 (38.389) | 
                          math-cordic.js | 52->48 (7.692) | 0.491->0.463 (5.703) | 
                    math-partial-sums.js | 44->40 (9.091) | 0.294->0.285 (3.061) | 
                   math-spectral-norm.js | 52->56 (-7.692) | 0.392->0.379 (3.316) | 
                         string-fasta.js | 64->60 (6.250) | 3.79->2.172 (42.691) | 
                         Geometric mean: |       RSS reduction: 1.4238% |            Speed up: 10.505% |

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
@@ -48,27 +48,22 @@
  * @return NaN if cannot read from string, ToNumber() otherwise
  */
 static ecma_number_t
-ecma_date_parse_date_chars (lit_utf8_iterator_t *iter, /**< iterator of the utf8 string */
+ecma_date_parse_date_chars (lit_utf8_byte_t **str_p, /**< pointer to the cesu8 string */
+                            const lit_utf8_byte_t *str_end_p, /**< pointer to the end of the string */
                             uint32_t num_of_chars) /**< number of characters to read and convert */
 {
   JERRY_ASSERT (num_of_chars > 0);
-
-  lit_utf8_size_t copy_size = 0;
-  const lit_utf8_byte_t *str_start_p = iter->buf_p + iter->buf_pos.offset;
+  const lit_utf8_byte_t *str_start_p = *str_p;
 
   while (num_of_chars--)
   {
-    if (lit_utf8_iterator_is_eos (iter)
-        || !lit_char_is_decimal_digit (lit_utf8_iterator_peek_next (iter)))
+    if (*str_p >= str_end_p || !lit_char_is_decimal_digit (lit_utf8_read_next (str_p)))
     {
       return ecma_number_make_nan ();
     }
-
-    copy_size += lit_get_unicode_char_size_by_utf8_first_byte (*(iter->buf_p + iter->buf_pos.offset));
-    lit_utf8_iterator_incr (iter);
   }
 
-  return ecma_utf8_string_to_number (str_start_p, copy_size);
+  return ecma_utf8_string_to_number (str_start_p, (lit_utf8_size_t) (*str_p - str_start_p));
 } /* ecma_date_parse_date_chars */
 
 /**
@@ -211,10 +206,11 @@ ecma_builtin_date_parse (ecma_value_t this_arg __attr_unused___, /**< this argum
   ssize_t sz = ecma_string_to_utf8_string (date_str_p, date_start_p, (ssize_t) date_str_size);
   JERRY_ASSERT (sz >= 0);
 
-  lit_utf8_iterator_t iter = lit_utf8_iterator_create (date_start_p, date_str_size);
+  lit_utf8_byte_t *date_str_curr_p = date_start_p;
+  const lit_utf8_byte_t *date_str_end_p = date_start_p + date_str_size;
 
   /* 1. read year */
-  ecma_number_t year = ecma_date_parse_date_chars (&iter, 4);
+  ecma_number_t year = ecma_date_parse_date_chars (&date_str_curr_p, date_str_end_p, 4);
 
   if (!ecma_number_is_nan (year)
       && year >= 0)
@@ -224,12 +220,12 @@ ecma_builtin_date_parse (ecma_value_t this_arg __attr_unused___, /**< this argum
     ecma_number_t time = ECMA_NUMBER_ZERO;
 
     /* 2. read month if any */
-    if (!lit_utf8_iterator_is_eos (&iter)
-        && lit_utf8_iterator_peek_next (&iter) == '-')
+    if (date_str_curr_p < date_str_end_p
+        && *date_str_curr_p == '-')
     {
       /* eat up '-' */
-      lit_utf8_iterator_incr (&iter);
-      month = ecma_date_parse_date_chars (&iter, 2);
+      date_str_curr_p++;
+      month = ecma_date_parse_date_chars (&date_str_curr_p, date_str_end_p, 2);
 
       if (month > 12 || month < 1)
       {
@@ -238,12 +234,12 @@ ecma_builtin_date_parse (ecma_value_t this_arg __attr_unused___, /**< this argum
     }
 
     /* 3. read day if any */
-    if (!lit_utf8_iterator_is_eos (&iter)
-        && lit_utf8_iterator_peek_next (&iter) == '-')
+    if (date_str_curr_p < date_str_end_p
+        && *date_str_curr_p == '-')
     {
       /* eat up '-' */
-      lit_utf8_iterator_incr (&iter);
-      day = ecma_date_parse_date_chars (&iter, 2);
+      date_str_curr_p++;
+      day = ecma_date_parse_date_chars (&date_str_curr_p, date_str_end_p, 2);
 
       if (day < 1 || day > 31)
       {
@@ -252,24 +248,24 @@ ecma_builtin_date_parse (ecma_value_t this_arg __attr_unused___, /**< this argum
     }
 
     /* 4. read time if any */
-    if (!lit_utf8_iterator_is_eos (&iter)
-        && lit_utf8_iterator_peek_next (&iter) == 'T')
+    if (date_str_curr_p < date_str_end_p
+        && *date_str_curr_p == 'T')
     {
+      /* eat up 'T' */
+      date_str_curr_p++;
+
       ecma_number_t hours = ECMA_NUMBER_ZERO;
       ecma_number_t minutes = ECMA_NUMBER_ZERO;
       ecma_number_t seconds = ECMA_NUMBER_ZERO;
       ecma_number_t milliseconds = ECMA_NUMBER_ZERO;
 
-      ecma_length_t num_of_visited_chars = lit_utf8_iterator_get_index (&iter);
-      ecma_length_t date_str_len = lit_utf8_string_length (iter.buf_p, iter.buf_size) - 1;
+      ecma_length_t remaining_length = lit_utf8_string_length (date_str_curr_p,
+                                                               (lit_utf8_size_t) (date_str_end_p - date_str_curr_p));
 
-      if ((date_str_len - num_of_visited_chars) >= 5)
+      if (remaining_length >= 5)
       {
-        /* eat up 'T' */
-        lit_utf8_iterator_incr (&iter);
-
         /* 4.1 read hours and minutes */
-        hours = ecma_date_parse_date_chars (&iter, 2);
+        hours = ecma_date_parse_date_chars (&date_str_curr_p, date_str_end_p, 2);
 
         if (hours < 0 || hours > 24)
         {
@@ -281,9 +277,9 @@ ecma_builtin_date_parse (ecma_value_t this_arg __attr_unused___, /**< this argum
         }
 
         /* eat up ':' */
-        lit_utf8_iterator_incr (&iter);
+        date_str_curr_p++;
 
-        minutes = ecma_date_parse_date_chars (&iter, 2);
+        minutes = ecma_date_parse_date_chars (&date_str_curr_p, date_str_end_p, 2);
 
         if (minutes < 0 || minutes > 59)
         {
@@ -291,11 +287,12 @@ ecma_builtin_date_parse (ecma_value_t this_arg __attr_unused___, /**< this argum
         }
 
         /* 4.2 read seconds if any */
-        if (!lit_utf8_iterator_is_eos (&iter) && lit_utf8_iterator_peek_next (&iter) == ':')
+        if (date_str_curr_p < date_str_end_p
+            && *date_str_curr_p == ':')
         {
           /* eat up ':' */
-          lit_utf8_iterator_incr (&iter);
-          seconds = ecma_date_parse_date_chars (&iter, 2);
+          date_str_curr_p++;
+          seconds = ecma_date_parse_date_chars (&date_str_curr_p, date_str_end_p, 2);
 
           if (seconds < 0 || seconds > 59)
           {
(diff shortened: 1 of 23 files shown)


egavrin:
fyi, measurements with RaspPi2

                               Benchmark |         RSS<br>(+ is better) |        Perf<br>(+ is better) |
                               --------- |                          --- |                         ---- |
                              3d-cube.js |          132->   132 (0.000) |       3.7904-> 3.756 (0.908) | 
                  access-binary-trees.js |           88->    88 (0.000) |       2.8832->2.8456 (1.304) | 
                      access-fannkuch.js |           40->    40 (0.000) |      10.1056->9.8144 (2.882) | 
                         access-nbody.js |           64->    64 (0.000) |       4.7656->4.6792 (1.813) | 
             bitops-3bit-bits-in-byte.js |           36->    36 (0.000) |       3.3376->3.2632 (2.229) | 
                  bitops-bits-in-byte.js |           36->    36 (0.000) |        4.468->4.4256 (0.949) | 
                   bitops-bitwise-and.js |           32->    32 (0.000) |       4.3424->4.2184 (2.856) | 
                controlflow-recursive.js |          220->   220 (0.000) |       3.3696->3.2576 (3.324) | 
                    date-format-xparb.js |          100->   100 (0.000) |      3.0168->1.8536 (38.557) | 
                          math-cordic.js |           48->    48 (0.000) |       4.9792->4.8968 (1.655) | 
                    math-partial-sums.js |           40->    40 (0.000) |       2.7184-> 2.644 (2.737) | 
                   math-spectral-norm.js |           52->    52 (0.000) |       3.3016-> 3.256 (1.381) | 
                         string-fasta.js |           52->    52 (0.000) |     27.4368->19.016 (30.692) | 
                         Geometric mean: |            RSS reduction: 0% |            Speed up: 7.9482% |

jerry-core/lit/lit-globals.h
```diff
@@ -94,6 +94,16 @@ typedef ecma_char_t *ecma_char_ptr_t;
 #define LIT_UTF8_MAX_BYTES_IN_CODE_POINT (4)
```
ruben-ayrapetyan: Could you, please, clarify, do we need the define after switching to CESU-8?

Is it correct that CESU-8 uses UTF-8 encoding scheme, but only for 1-byte and 2-byte code points?
Shouldn't we rename all `UTF8` / `utf8` definitions to `CESU8` / `cesu8`?

dbatyai: Sorry for the delay.
You are correct, CESU uses UTF-8 encoding, the only difference is with non-BMP code points. In that case we first separate the code point into a surrogate pair, and encode the surrogates separately in UTF-8, which means that these characters will be encoded in 6 bytes instead of 4. This may seem like a waste, but in this case we can directly decode the string into code units, without the need for special handling of non-BMP code points, which are very rare anyway.

We'll have to keep some of the UTF-8 support though, so I don't think that we should mass rename every definition.

dbatyai:
Hi, I've finished refactoring the built-ins to use CESU, so I think there won't be any major updates to this patch anymore, only some tweaks. @ruben-ayrapetyan, @egavrin, @zherczeg, please review.

Latest measurements with RasPi2:

Benchmark |         RSS<br>(+ is better) |        Perf<br>(+ is better) |
                               --------- |                          --- |                         ---- |
                              3d-cube.js | 132 -> 128 (3.0303) | 3.7 -> 3.696 (0.1081) | 
                  access-binary-trees.js | 84 -> 88 (-4.762) | 2.846 -> 2.828 (0.6325) | 
                      access-fannkuch.js | 48 -> 44 (8.3333) | 9.874 -> 9.874 (0) | 
                         access-nbody.js | 64 -> 64 (0) | 4.674 -> 4.632 (0.8986) | 
             bitops-3bit-bits-in-byte.js | 36 -> 36 (0) | 3.088 -> 3.152 (-2.073) | 
                  bitops-bits-in-byte.js | 36 -> 36 (0) | 4.224 -> 4.308 (-1.989) | 
                   bitops-bitwise-and.js | 28 -> 28 (0) | 4.06 -> 4.084 (-0.591) | 
                controlflow-recursive.js | 216 -> 212 (1.8519) | 3.138 -> 3.12 (0.5736) | 
                    date-format-xparb.js | 100 -> 100 (0) | 3.042 -> 1.798 (40.8941) | 
                          math-cordic.js | 44 -> 44 (0) | 4.166 -> 4.172 (-0.144) | 
                    math-partial-sums.js | 36 -> 36 (0) | 2.508 -> 2.55 (-1.675) | 
                   math-spectral-norm.js | 48 -> 52 (-8.333) | 3.178 -> 3.178 (0) | 
                         string-fasta.js | 56 -> 52 (7.1429) | 26.888 -> 9.534 (64.5418) | 
                         Geometric mean: |       RSS reduction: 0.6443% |           Speed up: 11.0395% |

jerry-core/ecma/builtin-objects/ecma-builtin-date.cpp
```diff
@@ -324,34 +322,35 @@ ecma_builtin_date_parse (ecma_value_t this_arg __attr_unused___, /**< this argum
...
+        remaining_length = lit_utf8_string_length (date_str_curr_p,
+                                                   (lit_utf8_size_t) (date_str_end_p - date_str_curr_p)) - 1;
 
-        if ((date_str_len - num_of_visited_chars) == 5)
+        if (remaining_length == 5)
         {
           bool is_negative = false;
 
-          if (lit_utf8_iterator_peek_next (&iter) == '-')
+          if (lit_cesu8_peek_next (&date_str_curr_p) == '-')
```
zherczeg: Can't we simply use *date_str_curr_p ?

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -827,15 +814,14 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
     }
 
     lit_code_point_t decoded_byte;
+    input_char_p++;
 
-    if (!lit_read_code_point_from_hex (input_char_p + 1, 2, &decoded_byte))
+    if (!lit_read_code_point_from_hex (&input_char_p, 2, &decoded_byte))
```
zherczeg: why & ?

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -902,6 +892,54 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
...
         }
       }
+      else if ((decoded_byte & LIT_UTF8_4_BYTE_MASK) == LIT_UTF8_4_BYTE_MARKER)
+      {
+        lit_utf8_byte_t octets[LIT_UTF8_MAX_BYTES_IN_CODE_POINT];
+        octets[0] = (lit_utf8_byte_t) decoded_byte;
+        bool is_valid = true;
+
+        for (int i = 1; i < LIT_UTF8_MAX_BYTES_IN_CODE_POINT && is_valid; i++)
+        {
```
zherczeg: you added && is_valid here, but there is a break after each is_valid = false. I think it is unnecessary

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -902,6 +892,54 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
...
+          if (input_char_p >= input_end_p || *input_char_p != '%')
```
zherczeg: *input_char_p != '%' ????

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -902,6 +892,54 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
...
+          {
+            is_valid = false;
+            break;
+          }
+          else
+          {
+            lit_code_point_t cp;
+
+            input_char_p++;
+            lit_read_code_point_from_hex (&input_char_p, 2, &cp);
```
zherczeg: I don't think you need & for passing pointers

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -909,57 +947,64 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
...
-      output_type_p = output_start_p + output_size;
+      JERRY_ASSERT (output_start_p + output_size == output_char_p);
 
-      while (!lit_utf8_iterator_is_eos (&characters))
-      {
-        bool original_byte = output_type_p[characters.buf_pos.offset] == URI_DECODE_ORIGINAL_BYTE;
+      bool valid_cesu8 = lit_is_cesu8_string_valid (output_start_p, output_size);
 
-        ecma_char_t character = lit_utf8_iterator_read_next (&characters);
+      if (valid_cesu8)
```
zherczeg: This complicated post processing is not needed here. The input is valid cesu8,, and invalid parts can only be constructed from %xx parts. Check the 3 octet long %xx sequences, whether they are surrogate pair or not. That is all.

jerry-core/lit/lit-char-helpers.cpp
```diff
@@ -360,6 +361,7 @@ lit_read_code_point_from_hex (const lit_utf8_byte_t *buf_p, /**< buffer with cha
 
     buf_p++;
   }
+  *str_p = buf_p;
```
zherczeg: Why?

jerry-core/lit/lit-strings.cpp
```diff
@@ -153,6 +153,77 @@ lit_is_utf8_string_valid (const lit_utf8_byte_t *utf8_buf_p, /**< utf-8 string *
...
+      {
+        /* invalid continuation byte */
+        return false;
+      }
+      code_point <<= LIT_UTF8_BITS_IN_EXTRA_BYTES;
+      code_point |= (c & LIT_UTF8_LAST_6_BITS_MASK);
+    }
+
+    if (code_point < min_code_point
+        || code_point > LIT_UNICODE_CODE_POINT_MAX)
```
zherczeg: code point cannot be > LIT_UNICODE_CODE_POINT_MAX

jerry-core/lit/lit-strings.cpp
```diff
@@ -598,6 +556,183 @@ lit_read_code_point_from_utf8 (const lit_utf8_byte_t *buf_p, /**< buffer with ch
...
+    ret = ((lit_code_point_t) (c & LIT_UTF8_LAST_5_BITS_MASK));
+  }
+  else
+  {
+    JERRY_ASSERT ((c & LIT_UTF8_3_BYTE_MASK) == LIT_UTF8_3_BYTE_MARKER);
+    bytes_count = 3;
+    ret = ((lit_code_point_t) (c & LIT_UTF8_LAST_4_BITS_MASK));
+  }
+
+  for (uint32_t i = 1; i < bytes_count; ++i)
```
zherczeg: I don't think we need a for here.
Just while (bytes_count-- > 0) and increase buf

zherczeg:
I do like the direction. I am happy that so much code can be removed. Please check all & usage, I think most of them are unnecessary. And please avoid for loops when simpler loops are available.

dbatyai:
@zherczeg, thanks for your comments, I've updated the patch.

jerry-core/lit/lit-strings.cpp
dbatyai: We need to keep `bytes_count` because we `return` with it.

jerry-core/lit/lit-strings.cpp
```diff
@@ -598,6 +555,158 @@ lit_read_code_point_from_utf8 (const lit_utf8_byte_t *buf_p, /**< buffer with ch
...
+} /* lit_cesu8_incr */
+
+/**
+ * Decrease character pointer by one code unit.
+ */
+void
+lit_cesu8_decr (lit_utf8_byte_t **buf_p) /**< in-out:buffer with characters */
+{
+  JERRY_ASSERT (*buf_p);
+  lit_utf8_byte_t *current_p = *buf_p;
```
egavrin: Add extra space.

jerry-core/lit/lit-strings.cpp
```diff
@@ -598,6 +555,158 @@ lit_read_code_point_from_utf8 (const lit_utf8_byte_t *buf_p, /**< buffer with ch
...
+  do
+  {
+    current_p--;
+  }
+  while ((*(current_p) & LIT_UTF8_EXTRA_BYTE_MASK) == LIT_UTF8_EXTRA_BYTE_MARKER);
+  *buf_p = current_p;
```
egavrin: Likewise, after loop

egavrin:
Is it needed to add `cesu8` instead of `utf8`? As far as I understand CESU-8 is a variant of UTF-8, so I think we can leave `utf8` and mention CESU-8 only in comments.
In some places there are mix of `utf8` and `cesu8`:
```cpp
lit_utf8_size_t read_size = lit_read_code_unit_from_cesu8 (current_p, &current_char);
```

Regardless style, this PR is OK for me.

jerry-core/lit/lit-strings.cpp
```diff
@@ -545,6 +479,29 @@ lit_utf8_string_length (const lit_utf8_byte_t *utf8_buf_p, /**< utf-8 string */
 } /* lit_utf8_string_length */
```
ruben-ayrapetyan: Do we need this function?

jerry-core/lit/lit-strings.cpp
```diff
@@ -828,31 +989,34 @@ lit_compare_utf8_strings (const lit_utf8_byte_t *string1_p, /**< utf-8 string */
...
  */
-bool lit_compare_utf8_strings_relational (const lit_utf8_byte_t *string1_p, /**< utf-8 string */
+bool lit_compare_cesu8_strings_relational (const lit_utf8_byte_t *string1_p, /**< utf-8 string */
                                           lit_utf8_size_t string1_size, /**< string size */
                                           const lit_utf8_byte_t *string2_p, /**< utf-8 string */
                                           lit_utf8_size_t string2_size) /**< string size */
 {
-  lit_utf8_iterator_t iter1 = lit_utf8_iterator_create (string1_p, string1_size);
-  lit_utf8_iterator_t iter2 = lit_utf8_iterator_create (string2_p, string2_size);
+  lit_utf8_byte_t *string1_pos = (lit_utf8_byte_t *) string1_p;
```
ruben-ayrapetyan: Wouldn't it be useful to introduce CESU iterators?

zherczeg:
LGTM

jerry-core/lit/lit-strings.cpp
dbatyai: We don't, thanks for noting. I'll remove it.

jerry-core/lit/lit-strings.cpp
dbatyai: In some cases it could be useful, but it would also over complicate others. In lots of cases we know   the string is only ASCII, or we expect it to be ASCII. If we don't use iterators, then we can easily take advantage of these situations.

dbatyai:
@egavrin, I don't think mixing the notation would cause any confusion, since, as you said, CESU-8 is a variant of UTF-8. On the other hand, keeping everything as `utf8` would be misleading in my opinion, since in that case some functions would not behave as implied by their names, while others would. Of course, lots of comments could help with this, but still, I think naming functions by their behavior is much clearer.

egavrin:
@dbatyai we can simply write in documentation that we're implementing "UTF-8 in CESU-8", but we can leave as is.

dbatyai:
@egavrin, I updated the patch to keep `utf8`.

egavrin:
Good to me. `make push`

PR #625, opened 2015-09-09
Fix MEM_STATS build configuration.

lemmaa:

### DIFF
--- a/jerry-core/jerry.cpp
+++ b/jerry-core/jerry.cpp
@@ -1330,7 +1330,7 @@ jerry_init (jerry_flag_t flags) /**< combination of Jerry flags */
 #endif /* !JERRY_ENABLE_LOG */
   }
 
-  if (flags & (JERRY_FLAG_MEM_STATS))
+  if (flags & (JERRY_FLAG_MEM_STATS | JERRY_FLAG_MEM_STATS_PER_OPCODE | JERRY_FLAG_MEM_STATS_SEPARATE))
   {
 #ifndef MEM_STATS
     flags &= ~(JERRY_FLAG_MEM_STATS
@@ -1339,13 +1339,6 @@ jerry_init (jerry_flag_t flags) /**< combination of Jerry flags */
 
     JERRY_WARNING_MSG ("Ignoring memory statistics option because of '!MEM_STATS' build configuration.\n");
 #endif /* !MEM_STATS */
-  }
-  else if (flags & (JERRY_FLAG_MEM_STATS_PER_OPCODE | JERRY_FLAG_MEM_STATS_SEPARATE))
-  {
-    flags &= ~(JERRY_FLAG_MEM_STATS_PER_OPCODE | JERRY_FLAG_MEM_STATS_SEPARATE);
-
-    JERRY_WARNING_MSG (
-      "Ignoring detailed memory statistics options because memory statistics dump mode is not enabled.\n");
   }
 
   jerry_flags = flags;
--- a/jerry-core/jerry.cpp
+++ b/jerry-core/jerry.cpp
@@ -1338,6 +1338,11 @@ jerry_init (jerry_flag_t flags) /**< combination of Jerry flags */
                | JERRY_FLAG_MEM_STATS_SEPARATE);
 
     JERRY_WARNING_MSG ("Ignoring memory statistics option because of '!MEM_STATS' build configuration.\n");
+#else /* !MEM_STATS */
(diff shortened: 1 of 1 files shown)


jerry-core/jerry.cpp
```diff
@@ -1340,13 +1340,6 @@ jerry_init (jerry_flag_t flags) /**< combination of Jerry flags */
     JERRY_WARNING_MSG ("Ignoring memory statistics option because of '!MEM_STATS' build configuration.\n");
 #endif /* !MEM_STATS */
   }
-  else if (flags & (JERRY_FLAG_MEM_STATS_PER_OPCODE | JERRY_FLAG_MEM_STATS_SEPARATE))
-  {
-    flags &= ~(JERRY_FLAG_MEM_STATS_PER_OPCODE | JERRY_FLAG_MEM_STATS_SEPARATE);
-
-    JERRY_WARNING_MSG (
-      "Ignoring detailed memory statistics options because memory statistics dump mode is not enabled.\n");
```
ruben-ayrapetyan: @lemmaa, the `else if` block checks that `--mem-stats-separate` and `--mem-stats-per-opcode` options are not passed if `--mem-stats` is not passed, i.e.  "separate" and "per-opcode" memory statistics mode are turned off without `--mem-stats` option.

Currently, behaviour in case `JERRY_FLAG_MEM_STATS_SEPARATE` flag is set, and `JERRY_FLAG_MEM_STATS` is not set - dump of statistics only for parser phase, and not for execution phase (the first of the dumps is performed in `jerry_parse`, and the second should be performed in in `mem_finalize`). So, in the case, only parser phase statistics are dumped, and not separate statistics for parser and execution phases.

 `JERRY_FLAGS_MEM_STATS` could be set in the `if` block, so implicitly enabling `--mem-stats` mode, if `JERRY_FLAG_MEM_STATS_SEPARATE` or `JERRY_FLAG_MEM_STATS_PER_OPCODE` are set.

lemmaa:
@ruben-ayrapetyan I modified the PR. Tried to remove dependencies between options, following your guide.

ruben-ayrapetyan:
@lemmaa, the change looks good to me.

There is a minor style comment. `#endif`s in other code places are usually followed by condition, under which code block, located just above the `#endif`, executes. For example:
- https://github.com/Samsung/jerryscript/blob/cf2bc459bb59728d4e4dc691f0e042c519e5bcba/jerry-core/ecma/base/ecma-helpers-conversion.cpp#L773
- https://github.com/Samsung/jerryscript/blob/311cc65b33a150c1eed055f4a96670922d5478ca/jerry-core/jrt/jrt.h#L43
- https://github.com/Samsung/jerryscript/blob/fd9ff8e3bdd243d386d02dbf26a3707994519c75/jerry-core/ecma/base/ecma-lcache.cpp#L69

Could you, please, change the condition to `/* MEM_STATS */`?

lemmaa:
@ruben-ayrapetyan I see. But I have confusion. From the following `#else` clause, which one is right?

case 1:
 ````
#ifndef MEM_STATS
...
#else /* MEM_STATS */
...
endif /* MEM_STATS */
````
or

case 2:
````
#ifndef MEM_STATS
...
#else /* !MEM_STATS */
...
endif /* MEM_STATS */
````

sand1k:
@lemmaa, the case 2 is correct since the comment of #else or #endif corresponds to the condition of the previous block (block which is closed by #else or #endif).

lemmaa:
@sand1k In case of `#ifdef`, it's not confusing. But in case of `#ifndef`, it looks confusing for me. :)
Anyway, then following are incorrect,right?

https://github.com/Samsung/jerryscript/blob/master/jerry-core/ecma/base/ecma-lcache.cpp#L224
https://github.com/Samsung/jerryscript/blob/master/jerry-core/ecma/operations/ecma-lex-env.cpp#L54

sand1k:
@lemmaa, yes, those are incorrect. It looked confusing for me too, the main thing is to catch the idea :)

lemmaa:
@ruben-ayrapetyan I updated PR.

ruben-ayrapetyan:
@lemmaa, I think that, as both options seem to be correct and readable, in this case the best option would be the one that is consistent with other places of the code.

Unfortunately, at present, there is inconsistency in the comments after `#else` that, in my opinion, is a style issue that should be corrected after decision is made about which of the options should be used (maybe, the decision should be recorded in style guide as was proposed in https://github.com/Samsung/jerryscript/pull/623#discussion_r39261691).

ruben-ayrapetyan:
Looks good to me

lemmaa:
@galpeter @LaszloLango @zherczeg Would you please review this?

zherczeg:
LGTM

PR #631, opened 2015-09-16
Fix tr arguments in the Makefile

galpeter:
On Linux the original tr call did nothing and thus the Makefile
tried to load the toolchain_Linux_x86_64.cmake file, which doesn't exists.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,7 +78,7 @@
    QLOG := >/dev/null
   endif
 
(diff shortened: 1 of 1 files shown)


dbatyai:
lgtm

lemmaa:
@galpeter I can't experience your situation. In my case, original also works well. Anyway, It also good.
LGTM

egavrin:
good to me

PR #632, opened 2015-09-16
Support `gcc` in Mac OS X build

lemmaa:
- if `gcc` exist, use `gcc` for build, otherwise use `cc(=clang)` for build.


egavrin:
Good to me

galpeter:
lgtm

PR #638, opened 2015-09-21
Fix benchmark scripts to run on Mac OS X

lemmaa:
- perf: bypass the regex interpreting error of 'sed'.
- rss-measure: use alternative way to measure approx. RSS.

### DIFF
--- a/tools/run-perf-test.sh
+++ b/tools/run-perf-test.sh
@@ -107,6 +107,7 @@ function run-suite()
   done
 }
 
+date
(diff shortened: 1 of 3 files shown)


egavrin:
Good to me

tools/run-perf-test.sh
```diff
@@ -107,6 +107,7 @@ function run-suite()
   done
 }
 
+date
```
galpeter: What is the reason for the `date` call here? (and also at the end of the file)

lemmaa: @galpeter Just mark the benchmark time and duration. :)

galpeter:
lgtm

PR #651, opened 2015-10-02
Add foreach function for inquiry properties into obj

idkiller:
Signed-off-by: pius.lee <pius.lee@samsung.com>

This new function makes api user can search enumerable properties in the javascript object.

### DIFF
--- a/jerry-core/jerry-api.h
+++ b/jerry-core/jerry-api.h
@@ -140,6 +140,13 @@ typedef bool (*jerry_external_handler_t) (const jerry_api_object_t *function_obj
  */
 typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
 
+/**
+ * function type applied for each fields in objects
+ */
+typedef bool (*jerry_object_field_foreach_t) (const jerry_api_string_t *field_name_p,
+                                              const jerry_api_value_t *field_value_p,
+                                              void *user_data_p);
+
 extern EXTERN_C ssize_t
 jerry_api_string_to_char_buffer (const jerry_api_string_t *, jerry_api_char_t *, ssize_t);
 extern EXTERN_C
@@ -199,6 +206,9 @@ bool jerry_api_set_object_field_value (jerry_api_object_t *, const jerry_api_cha
 extern EXTERN_C
 bool jerry_api_set_object_field_value_sz (jerry_api_object_t *, const jerry_api_char_t *, jerry_api_size_t,
                                           const jerry_api_value_t *);
+
+extern EXTERN_C
+bool jerry_api_foreach_object_field (jerry_api_object_t *, jerry_object_field_foreach_t, void *);
 
 extern EXTERN_C
 bool jerry_api_get_object_native_handle (jerry_api_object_t *, uintptr_t *);
--- a/jerry-core/jerry.cpp
+++ b/jerry-core/jerry.cpp
@@ -26,6 +26,7 @@
 #include "ecma-init-finalize.h"
 #include "ecma-objects.h"
 #include "ecma-objects-general.h"
+#include "ecma-try-catch-macro.h"
 #include "lit-magic-strings.h"
 #include "parser.h"
 #include "serializer.h"
@@ -897,6 +898,63 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
                                               lit_zt_utf8_string_size (field_name_p),
                                               field_value_p);
 }
+
+/**
+ * Applies the given function to the every fields in the objects
+ *
+ * @return true, if object fields traversal was performed successfully, i.e.:
+ *                - no unhandled exceptions were thrown in object fields traversal;
+ *                - object fields traversal was stopped on callback that returned false;
+ *         false - otherwise,
+ *                 if getter of field threw a exception or unhandled exceptions were thrown during traversal;
+ */
+bool
+jerry_api_foreach_object_field (jerry_api_object_t *object_p, /**< object */
+                               jerry_object_field_foreach_t foreach_p, /**< foreach function */
+                               void *user_data_p) /**< user data for foreach function */
+{
+  jerry_assert_api_available ();
+
+  ecma_collection_iterator_t names_iter;
+  ecma_collection_header_t *names_p = ecma_op_object_get_property_names (object_p, false, true, true);
+  ecma_collection_iterator_init (&names_iter, names_p);
+
+  ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
+
+  bool continuous = true;
+
+  while (ecma_is_completion_value_empty (ret_value)
+         && continuous
+         && ecma_collection_iterator_next (&names_iter))
+  {
+    ecma_string_t *property_name_p = ecma_get_string_from_value (*names_iter.current_value_p);
+
+    ECMA_TRY_CATCH (property_value, ecma_op_object_get (object_p, property_name_p), ret_value);
+
+    jerry_api_value_t field_value;
+    jerry_api_convert_ecma_value_to_api_value (&field_value, property_value);
+
+    continuous = foreach_p (property_name_p, &field_value, user_data_p);
+
+    jerry_api_release_value (&field_value);
+
+    ECMA_FINALIZE (property_value);
+  }
+
+  ecma_free_values_collection (names_p, true);
+  if (ecma_is_completion_value_empty (ret_value))
+  {
+    return true;
+  }
+  else
+  {
+    JERRY_ASSERT (ecma_is_completion_value_throw (ret_value));
+
+    ecma_free_completion_value (ret_value);
+
+    return false;
+  }
+} /* jerry_api_foreach_object_field */
 
 /**
  * Get value of field in the specified object
--- a/tests/unit/test-api.cpp
+++ b/tests/unit/test-api.cpp
@@ -15,6 +15,7 @@
 
 #include "jerry.h"
 #include "jerry-api.h"
+#include "config.h"
 
 #include "test-common.h"
 
@@ -54,6 +55,9 @@ const char *test_source = (
                            "function throw_reference_error() { "
                            " throw new ReferenceError ();"
                            "} "
+                           "p = {'alpha':32, 'bravo':false, 'charlie':{}, 'delta':123.45, 'echo':'foobar'};"
+                           "np = {}; Object.defineProperty (np, 'foxtrot', { "
+                           "get: function() { throw 'error'; }, enumerable: true }) "
                            );
 
 bool test_api_is_free_callback_was_called = false;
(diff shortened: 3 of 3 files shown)


idkiller:
hello, I have plan to use jerryscript in my local project. (maybe it will be open soon)
I want to get names and values of properties in object from javascript.
but I can not found function for getting names in object properties.
This function made to get unknown properties and there names in C/C++ scope.

jerry-core/jerry.cpp
```diff
@@ -899,6 +899,64 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+                               jerry_object_field_foreach_t foreach_p, /**< foreach function */
+                               void *user_data) /**< user data for foreach function */
+{
+  jerry_assert_api_available ();
+
+  for (ecma_property_t *property_p = ecma_get_property_list (object_p);
+       property_p != NULL;
+       property_p = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p))
+  {
+    if (!ecma_is_property_enumerable (property_p)) continue;
```
galpeter: If someone calls this (the api_foreach) function to a built-in object, then the `ecma_is_property_enumerable` can assert on internal properties. Please move this after the if-elseif-else block below.

Additionally format it like this:
```js
if (!ecma....(property_))
{
  continue;
}
```

jerry-core/jerry.cpp
```diff
@@ -899,6 +899,64 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t,
+                                                   property_p->u.named_accessor_property.name_p);
+    }
+    else
+    {
+      continue;
+    }
+
+    jerry_api_value_t field_value;
+    ecma_completion_value_t get_completion = ecma_op_object_get (object_p,
```
galpeter: @ruben-ayrapetyan, could you please confirm that in this case ECMA_TRY_CATCH and ECMA_FINALIZE should be used?

jerry-core/jerry-api.h
```diff
@@ -140,6 +140,13 @@ typedef bool (*jerry_external_handler_t) (const jerry_api_object_t *function_obj
  */
 typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
 
+/**
+ * function type applied for each fields in objects
+ */
+typedef bool (*jerry_object_field_foreach_t) (const jerry_api_string_t * field_name_p,
+                                              const jerry_api_value_t * field_value_p,
```
galpeter: Minor formatting:
```js
const jerry_api_string_t *field_name_p,
const jerry_api_string_t *field_value_p,
```

That is the `*` should not have a space before the variable name.

tests/unit/test-api.cpp
```diff
@@ -222,6 +223,48 @@ const jerry_api_char_ptr_t magic_string_items[] =
...
+                     const jerry_api_value_t *value, void *user_data)
+{
+  char str_buf_p[128];
+  ssize_t sz = jerry_api_string_to_char_buffer (name, (jerry_api_char_t *)str_buf_p, 128);
+  str_buf_p[sz] = '\0';
+  
+  if (!strncmp (str_buf_p, "alpha", (size_t)sz))
+  {
+    JERRY_ASSERT (value->type == JERRY_API_DATA_TYPE_FLOAT64);
+    JERRY_ASSERT (value->v_float64 == 32);
```
galpeter: These two lines are only valid if the 
`CONFIG_ECMA_NUMBER_TYPE` define equals to `CONFIG_ECMA_NUMBER_FLOAT64` it is also possible to have `CONFIG_ECMA_NUMBER_FLOAT32` values. Please see https://github.com/Samsung/jerryscript/blob/master/jerry-core/jerry.cpp#L160 .

tests/unit/test-api.cpp
```diff
@@ -222,6 +223,48 @@ const jerry_api_char_ptr_t magic_string_items[] =
...
+  else if (!strncmp (str_buf_p, "delta", (size_t)sz))
+  {
+    JERRY_ASSERT (value->type == JERRY_API_DATA_TYPE_FLOAT64);
+    JERRY_ASSERT (value->v_float64 == 123.45);
+  }
+  else if (!strncmp (str_buf_p, "echo", (size_t)sz))
+  {
+    JERRY_ASSERT (value->type == JERRY_API_DATA_TYPE_STRING);
+    ssize_t echo_sz = jerry_api_string_to_char_buffer (
+        value->v_string, (jerry_api_char_t *)str_buf_p, 128);
```
galpeter: AFAIK: we don't need to wrap this line, it fits into one 120 long line

tests/unit/test-api.cpp
```diff
@@ -222,6 +223,48 @@ const jerry_api_char_ptr_t magic_string_items[] =
...
+    JERRY_ASSERT (value->v_bool == false);
+  }
+  else if (!strncmp (str_buf_p, "charlie", (size_t)sz))
+  {
+    JERRY_ASSERT (value->type == JERRY_API_DATA_TYPE_OBJECT);
+  }
+  else if (!strncmp (str_buf_p, "delta", (size_t)sz))
+  {
+    JERRY_ASSERT (value->type == JERRY_API_DATA_TYPE_FLOAT64);
+    JERRY_ASSERT (value->v_float64 == 123.45);
```
galpeter: Same CONFIG_ECMA_NUMBER_TYPE problem as above

jerry-core/jerry.cpp
ruben-ayrapetyan: @idkiller, @galpeter, yes, its better to use the `ECMA_TRY_CATCH`-style here, like in https://github.com/Samsung/jerryscript/blob/master/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.cpp#L1070

jerry-core/jerry.cpp
```diff
@@ -899,6 +899,64 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
 }
 
 /**
+ * Applies the given function to the every fields in the objects
+ *
+ */
+void
+jerry_api_foreach_object_field(jerry_api_object_t *object_p, /**< object */
+                               jerry_object_field_foreach_t foreach_p, /**< foreach function */
+                               void *user_data) /**< user data for foreach function */
```
egavrin: What kind of data do you plan to pass through this pointer?

jerry-core/jerry.cpp
idkiller: ok! thanks I will patch it.

jerry-core/jerry.cpp
idkiller: Thanks for your advice. I'll patch it.

jerry-core/jerry-api.h
idkiller: ok sure I'll fix it!

tests/unit/test-api.cpp
idkiller: missing point! thank you!

tests/unit/test-api.cpp
idkiller: I don't understand it. you means this projects use 120 column in one line.

tests/unit/test-api.cpp
idkiller: I miss it!

jerry-core/jerry.cpp
idkiller: if someone need keep there context in foreach callback, the argument passed to callback function is necessary.
for example
```cpp
Foobar foo;
auto foreach = [] (const jerry_api_string_t *name, const jerry_api_value_t *value, void *user_data)
{
    Foobar *foo = reinterpret_cast<Foobar*>(user_data);
    foo->somefunction(name, value);
}
jerry_api_foreach_object_field(obj, foreach, &foo);
                                                            
```

if callback function pointer can be replaced to std::function, user data does not need. (std::function can carry other data) but I think this project avoid c++ STL by design.

egavrin:
@idkiller 
```
Signed-off-by message is incorrect. The following line should be at the end of the 50aa8e60faf4256c8cf024997075234d9fd494d4 commit's message: 'JerryScript-DCO-1.0-Signed-off-by: pius.lee pius.lee@samsung.com'. 
```

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,58 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+
+    jerry_api_value_t field_value;
+    ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
+
+    ECMA_TRY_CATCH (property_value, ecma_op_object_get (object_p, property_name_p),
+                    ret_value);
+    jerry_api_convert_ecma_value_to_api_value (&field_value, property_value);
+    ECMA_FINALIZE (property_value);
+
+    if (!foreach_p (property_name_p, &field_value, user_data))
```
galpeter: If the ecma_op_object_get call fails, then the field_value would contain incorrect data. In that case should we really call the foreach_p callback? Consider the following user object:
```js
var obj = {}
Object.defineProperty(obj, "value", { get: function() { throw "error"; }} );
```
So when accessing the `obj.value` we'll get an error, how should we handle such cases?

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,58 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+    {
+      jerry_api_release_value (&field_value);
+      break;
+    }
+    jerry_api_release_value (&field_value);
+  }
+}
```
galpeter: Missing `/* jerry_api_foreach_object_field */ ` comment from the end of the method.

jerry-core/jerry.cpp
idkiller: I think it should be return jerry_completion_code_t for normal or exceptional case.
I'll update it with jerry_completion_code_t.

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,65 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+    }
+    if (!ecma_is_property_enumerable (property_p))
+    {
+      continue;
+    }
+
+    jerry_api_value_t field_value;
+    ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
+
+    ECMA_TRY_CATCH (property_value, ecma_op_object_get (object_p, property_name_p),
```
galpeter: The ECMA_TRY_CATCH should be used the following way in this case:
```c
ECMA_TRY_CATCH(...)

jerry_api_convert...

if(!foreach_p(..))
{
  // note you should not break out from here, because the finalize would not be executed in that case.
   ...
}
jerry_api_release_value(...);

ECMA_FINALIZE(..)
```

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,57 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
 }
 
 /**
+ * Applies the given function to the every fields in the objects
+ *
+ */
+bool
+jerry_api_foreach_object_field (jerry_api_object_t *object_p, /**< object */
```
ruben-ayrapetyan: @idkiller, could you, please, add `@return` comment part?

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,57 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+                               jerry_object_field_foreach_t foreach_p, /**< foreach function */
+                               void *user_data) /**< user data for foreach function */
```
ruben-ayrapetyan: As this is a variable of pointer type, its better to add `_p` suffix for it - `user_data_p`.

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,57 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+    {
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t, property_p->u.named_data_property.name_p);
+    }
+    else if (property_p->type == ECMA_PROPERTY_NAMEDACCESSOR)
+    {
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t, property_p->u.named_accessor_property.name_p);
+    }
+    else
+    {
+      continue;
```
ruben-ayrapetyan: Could you, please, add assertion that `property_p->type == ECMA_PROPERTY_INTERNAL`?

jerry-core/jerry-api.h
```diff
@@ -140,6 +140,13 @@ typedef bool (*jerry_external_handler_t) (const jerry_api_object_t *function_obj
  */
 typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
 
+/**
+ * function type applied for each fields in objects
+ */
+typedef bool (*jerry_object_field_foreach_t) (const jerry_api_string_t *field_name_p,
+                                              const jerry_api_value_t *field_value_p,
+                                              void *user_data);
```
ruben-ayrapetyan: `user_data_p`

ruben-ayrapetyan:
I've not got other comments.

egavrin:
Good to me

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,62 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+
+    ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();
+
+    ECMA_TRY_CATCH (property_value, ecma_op_object_get (object_p, property_name_p), ret_value);
+    jerry_api_value_t field_value;
+    jerry_api_convert_ecma_value_to_api_value (&field_value, property_value);
+    continuous = foreach_p (property_name_p, &field_value, user_data_p);
+    jerry_api_release_value (&field_value);
+    ECMA_FINALIZE (property_value);
+    if (ecma_is_completion_value_throw (ret_value))
```
galpeter: I would suggest using the `!ecma_is_completion_value_empty (ret_value)` condition instead. Also please add an empty line before the if.

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,62 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+    else if (property_p->type == ECMA_PROPERTY_NAMEDACCESSOR)
+    {
+      property_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t, property_p->u.named_accessor_property.name_p);
+    }
+    else
+    {
+      JERRY_ASSERT (property_p->type == ECMA_PROPERTY_INTERNAL);
+      continue;
+    }
+    if (!ecma_is_property_enumerable (property_p))
```
galpeter: Please add an empty line before the if

galpeter:
I don't have any other comments

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,66 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+                               jerry_object_field_foreach_t foreach_p, /**< foreach function */
+                               void *user_data_p) /**< user data for foreach function */
+{
+  jerry_assert_api_available ();
+
+  bool continuous = true;
+  for (ecma_property_t *property_p = ecma_get_property_list (object_p);
+       property_p != NULL && continuous;
+       property_p = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p))
+  {
```
ruben-ayrapetyan: @idkiller, could you, please, update the property enumeration code like in commit 09b4d2071812b3212be80cadd016e407c1fc699b, as the #694 patch was landed?

jerry-core/jerry.cpp
```diff
@@ -899,6 +900,51 @@ bool jerry_api_get_object_field_value (jerry_api_object_t *object_p,
...
+    ECMA_TRY_CATCH (property_value, ecma_op_object_get (object_p, property_name_p), ret_value);
+    jerry_api_value_t field_value;
+    jerry_api_convert_ecma_value_to_api_value (&field_value, property_value);
+    continuous = foreach_p (property_name_p, &field_value, user_data_p);
+    jerry_api_release_value (&field_value);
+    ECMA_FINALIZE (property_value);
+
+    if (!ecma_is_completion_value_empty (ret_value))
+    {
+      ecma_free_completion_value (ret_value);
```
ruben-ayrapetyan: In the case, exceptions would be just ignored.

ruben-ayrapetyan: I see, this is just what is necessary in the case - ignore exception and return `false` value.

ruben-ayrapetyan: Maybe, its better to implement exception handling as the following:
```cpp
ecma_collection_iterator_t names_iter;
ecma_collection_header_t *names_p = ecma_op_object_get_property_names (object_p, false, true, true);
ecma_collection_iterator_init (&names_iter, names_p);

ecma_completion_value_t ret_value = ecma_make_empty_completion_value ();

bool continuous = true;

while (ecma_is_completion_value_empty (ret_value)
       && continuous
       && ecma_collection_iterator_next (&names_iter))
{
  ecma_string_t *property_name_p = ecma_get_string_from_value (*names_iter.current_value_p);

  ECMA_TRY_CATCH (property_value, ecma_op_object_get (object_p, property_name_p), ret_value);

  jerry_api_value_t field_value;
  jerry_api_convert_ecma_value_to_api_value (&field_value, property_value);

  continuous = foreach_p (property_name_p, &field_value, user_data_p);

  jerry_api_release_value (&field_value);

  ECMA_FINALIZE (property_value);
}

ecma_free_values_collection (names_p, true);

if (ecma_is_completion_value_empty (ret_value))
{
  return true;
}
else
{
  JERRY_ASSERT (ecma_is_completion_value_throw (ret_value));

  ecma_free_completion_value (ret_value);

  return false;
}
```

So, it would be more similar to how its implemented in other engine parts.

What do you think about this?

idkiller: It's good! I will copy it to my patch.

galpeter:
lgtm

jerry-core/jerry.cpp
```diff
@@ -1067,6 +1125,7 @@ jerry_api_set_object_native_handle (jerry_api_object_t *object_p, /**< object to
                                                            ECMA_INTERNAL_PROPERTY_FREE_CALLBACK);
     if (prop_p != NULL)
     {
+      ecma_free_external_pointer_in_property (prop_p);
       ecma_delete_property (object_p, prop_p);
```
ruben-ayrapetyan: The free operation is also performed in `ecma_free_property` that is called from `ecma_delete_property`, so it is not necessary to call `ecma_free_external_pointer_in_property` here.

ruben-ayrapetyan:
After fix of https://github.com/Samsung/jerryscript/pull/651#discussion_r44264948, the changes would look good to me

jerry-core/jerry.cpp
idkiller: 1128 line is code for jerry_api_set_object_native_handle, would I fix it in other patch?

ruben-ayrapetyan:
Looks good to me

ruben-ayrapetyan:
@egavrin, I've already reviewed the pull request - https://github.com/Samsung/jerryscript/pull/651#issuecomment-155386419.

egavrin:
Merged ece246ef9e003f6365fa4828e945bba4d04d4168

Issue #563, opened 2015-08-12
Correctly parse RegExp literals in places where they should be allowed.


dbatyai:
Currently we fail to parse RegExp literals in some places, where they should be allowed.
For example
```
if (true) /abc/.exec("abc");

if (true) {} /abc/.exec("abc");

var a
/abc/.exec("abc");
```
should all be valid.

wateret:
Closing as #660 landed.

Issue #655, opened 2015-10-02
if statement followed by a regex literal throws parse error


renatahodovan:
###### Jerry version:
    Checked revision: 98fc409
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
if (true) {}
/a/;
```
Output:
```
/a/;
^
ERROR: Ln 2, Col 1: Unknown token /
```

wateret:
Closing as #660 landed.

PR #660, opened 2015-10-08
Fix parser bug between regex and div operator

### LINKED ISSUE #652: Divisions are not parsed correctly
###### Jerry version:
    Checked revision: 98fc409
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
this / 10
```
Output:
```
this / 10
     ^
ERROR: Ln 1, Col 6: Unterminated RegExp literal
```

wateret:
For Lexer alone, it is hard to find out whether a token that starts with `/` is a division operator or a regex literal. Parser should give its context information to Lexer that it is expecting which. 

In my implementation, Lexer expects division operator by default.
When Parser sees token `/` or `/=` in `parse_primary_expression()`, the token is re-lexed as a regex literal. Seems work good.

This will resolve #652 and #655.

### DIFF
--- a/jerry-core/parser/js/lexer.cpp
+++ b/jerry-core/parser/js/lexer.cpp
@@ -1262,10 +1262,15 @@ lexer_parse_comment (void)
  *      Currently, lexer token doesn't fully correspond to Token, defined in ECMA-262, v5, 7.5.
  *      For example, there is no new-line token type in the token definition of ECMA-262 v5.
  *
+ * Note:
+ *      For Lexer alone, it is hard to find out a token is whether a regexp or a division.
+ *      Parser must set maybe_regexp to true if a regexp is expected.
+ *      Otherwise, a division is expected.
+ *
  * @return constructed token
  */
 static token
-lexer_parse_token (void)
+lexer_parse_token (bool maybe_regexp) /**< read '/' as regexp? */
 {
   ecma_char_t c = LA (0);
 
@@ -1335,21 +1340,11 @@ lexer_parse_token (void)
     }
     else
     {
-      return lexer_parse_token ();
+      return lexer_parse_token (maybe_regexp);
     }
   }
 
-  if (c == LIT_CHAR_SLASH
-      && !(prev_non_lf_token.type == TOK_NAME
-           || prev_non_lf_token.type == TOK_NULL
-           || prev_non_lf_token.type == TOK_BOOL
-           || prev_non_lf_token.type == TOK_CLOSE_BRACE
-           || prev_non_lf_token.type == TOK_CLOSE_SQUARE
-           || prev_non_lf_token.type == TOK_CLOSE_PAREN
-           || prev_non_lf_token.type == TOK_SMALL_INT
-           || prev_non_lf_token.type == TOK_NUMBER
-           || prev_non_lf_token.type == TOK_STRING
-           || prev_non_lf_token.type == TOK_REGEXP))
+  if (c == LIT_CHAR_SLASH && maybe_regexp)
   {
     return lexer_parse_regexp ();
   }
@@ -1517,7 +1512,7 @@ lexer_parse_token (void)
 } /* lexer_parse_token */
 
 token
-lexer_next_token (void)
+lexer_next_token (bool maybe_regexp) /**< read '/' as regexp? */
 {
   lit_utf8_iterator_pos_t src_pos = lit_utf8_iterator_get_pos (&src_iter);
   if (src_pos.offset == 0 && !src_pos.is_non_bmp_middle)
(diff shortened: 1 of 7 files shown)


galpeter:
Few points:
* When modifying method signatures, please update also the relevant method comments. For example: the comment on `lexer_parse_token` was not updated. (Also the parameter comment is missing from the method)
* You mentioned the two issues (#652 and #655), those are good test cases, we should add them into our `tests/jerry/` directory
* Partially related: when I was trying out the PR I've found out that the parsing of the following js code is incorrect:
```js
1 / 2
/=a/;
```
The returned error message is: `ERROR: Ln 2, Col 5: Unknown token ;`. However this error also occurs without this PR

wateret:
@galpeter 
Thank you for your feedback. I made some changes:
* Updated comments as commented.
* Added test cases of #652 and #655. Added additional tricky test cases in `tests/jerry/regexp-literal.js`
* I'm not 100% sure about this, but it seems like no semicolon is inserted after first line(`1 / 2`) in your code. (Major JS engines like `v8` does not insert a semicolon.)
 * If so, the next token should be parsed as `/=`. This PR doesn't affect it.
 * If not, this PR will successfully parse the next token as `/=a/`.
 * Whichever, I think this issue is more about *automatic semicolon insertion*.

sand1k:
This resolves #563 too.

jerry-core/parser/js/parser.cpp
```diff
@@ -830,6 +830,14 @@ parse_primary_expression (void)
 
   switch (tok.type)
   {
+    case TOK_DIV:
+    case TOK_DIV_EQ:
+    {
+      // must be a regexp literal so re-lex token as a regexp literal
+      lexer_seek (tok.loc);
+      skip_token (true);
+      /* FALLTHRU */
```
sand1k: I would prefer to add a function:
```
static void
rescan_regexp_token ()
{
  lexer_seek (tok.loc);
  tok = lexer_next_token (true);
}
```
and call it here.  This would allow removing default argument from `skip_token` and make the code more explicit.

wateret:
@sand1k 
Thanks, I totally agree with you.
Updated as commented and added a test case for #563.

jerry-core/parser/js/lexer.cpp
```diff
@@ -1218,10 +1218,15 @@ lexer_parse_comment (void)
...
+ * Note:
+ *      For Lexer alone, it is hard to find out a token is whether a regexp or a division.
+ *      Parser must set maybe_regexp to true if a regexp is expected.
+ *      Otherwise, a division is expected.
+ *
  * @return constructed token
  */
 static token
-lexer_parse_token (void)
+lexer_parse_token (bool maybe_regexp = false) /**< read '/' as regexp? */
```
sand1k: Default argument value is not required here. Every call site of `lexer_parse_token` passes an argument.

jerry-core/parser/js/lexer.h
```diff
@@ -171,7 +171,7 @@ typedef struct
 
 void lexer_init (const jerry_api_char_t *, size_t, bool);
 
-token lexer_next_token (void);
+token lexer_next_token (bool = false);
```
sand1k: Could you please remove default argument value and use explicit argument value at every call site. We've agreed to use C++ features as little as possible in Jerry.

jerry-core/parser/js/parser.cpp
```diff
@@ -99,6 +99,13 @@ skip_token (void)
 }
 
 static void
+rescan_regexp_token (void)
```
sand1k: Please add comment with function description.

jerry-core/parser/js/parser.cpp
```diff
@@ -99,6 +99,13 @@ skip_token (void)
...
+{
+  lexer_seek (tok.loc);
+  tok = lexer_next_token (true);
+}
```
sand1k: /* rescan_regexp_token */

sand1k:
@wateret, please, fix minor issues.
As for the rest, LGTM.

jerry-core/parser/js/parser.cpp
```diff
@@ -252,7 +262,7 @@ jsp_find_next_token_before_the_locus (token_type token_to_find, /**< token to se
         if (lit_utf8_iterator_pos_cmp (tok.loc, end_loc) >= 0)
         {
           lexer_seek (end_loc);
-          tok = lexer_next_token ();
+          tok =  lexer_next_token (false);
```
sand1k: Extra space here and below.

wateret:
@sand1k Sorry for the extra spaces. I fixed them.
I think I need one more review to merge this?

egavrin:
@galpeter could you please check this PR?

galpeter:
lgtm

egavrin:
@wateret `make push`

PR #663, opened 2015-10-13
Rename gcc-ar, gcc-ranlib only when LTO is enabled in CMakefile.txt

seanshpark:
Fix is for xtensa-lx106-elf- tools. It does not support LTO yet.


ruben-ayrapetyan:
Looks good to me

egavrin:
`make push`

Issue #654, opened 2015-10-02
Object literals in case selectors are not supported


renatahodovan:
###### Jerry version:
    Checked revision: 98fc409
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
switch (true) {
    case {"foo": "bar"}:
        break;
}
```
Output:
```
        break;
             ^
ERROR: Ln 3, Col 14: No corresponding statement for the break
```

According to ecma5, the expression above should be parsed (in spite of that the given clause value will never be equal to any switch condition).

wateret:
@renatahodovan 
I think we can close this. Check out #666.

egavrin:
Works on master (f17f785b6ec0c71c2ef37e4e80a98aff1eff0963)

Issue #653, opened 2015-10-02
Conditional operator is not supported in case clauses


renatahodovan:
###### Jerry version:
    Checked revision: 98fc409
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
var a = "foo";
switch(a) {
    case true ? "foo" : "bar":
        print("OK");
        break;
}
```
Output:
```
    case true ? "foo" : "bar":
                             ^
ERROR: Ln 3, Col 30: Expected either ';' or newline token

```

According to ecma5, any expression can stand in a case clause after the 'case' keyword but ternary statements are rejected in jerry.

wateret:
@renatahodovan 
I think we can close this. Check out #666.

egavrin:
Works on master (f17f785b6ec0c71c2ef37e4e80a98aff1eff0963)

PR #666, opened 2015-10-13
Fix: parser bug in parse_switch_statement

wateret:
In *switch statement*, when the *expression* that follows `case` contains one or more `:` token 
It turns out issue #653 and #654 have the same problem.

In `parse_switch_statement`, bytecode is generated with two-pass. In second pass, the *expression* is not parsed with `parse_expression` but it just simply waits for a `:` token show. However `parse_expression` would generate unwanted bytecode, so I chose another approach - storing all  starting locations of case/default bodies.

% While working on this patch, I found another bug about *switch statement*. I'll post it as a separated issue.


egavrin:
Good to me

galpeter:
looks good to me

egavrin:
`make push`

Issue #667, opened 2015-10-13
switch statement jumps to incorrect case body


wateret:
###### Jerry version:
    Checked revision: 08c618e
    Build: debug.linux
###### OS: 
    Ubuntu 14.04, x86_64
###### Test case:
```javascript
switch (1) {
    case 0:
        assert(false);
        break;
}
```
Output:
```
Script assertion failed
```

The output shows that it executes `case 0:` body.

From my observation, this happens when there is no `default` clause. Looks like wrong bytecode is generated.

wateret:
Closing as #670 landed.

PR #670, opened 2015-10-16
Fix: switch statement jumps to incorrect case body

wateret:
The test case in the issue #667,
```javascript
switch (1) {
    case 0:
        assert(false);
        break;
}
```
This code fails because if there is no matching `case` and there is no `default` in switch statement, it does not jump out of the switch statement, and go on to the next opcode.

bytecode for this was:
```
  0:                 meta   12    6  255     // [no 'arguments'] [no 'eval'] 
  1:         reg_var_decl  128  132    0     // var tmp128 .. tmp132;
  2:           assignment  130    1    1     // tmp130 = 1: SMALLINT;
  3:           assignment  131    1    0     // tmp131 = 0: SMALLINT;
  4:     equal_value_type  132  130  131     // tmp132 = tmp130 === tmp131;
  5:     is_true_jmp_down  132    0    1     // if (tmp132) goto 6;
( jmp_down to line 10 is missing here )
  6:               call_n  130    0    1     // 
  7:           assignment  130    0    3     // tmp130 = false: SIMPLE;
  8:                 meta    2  130  255     // tmp130 = assert (tmp130);
  9:             jmp_down    0    1          // goto 10;
 10:                  ret                    // ret;
```

So I changed code to insert `jmp_down` even if there is no `default`.

### DIFF
--- a/jerry-core/parser/js/parser.cpp
+++ b/jerry-core/parser/js/parser.cpp
@@ -2462,10 +2462,7 @@ parse_switch_statement (void)
   }
   current_token_must_be (TOK_CLOSE_BRACE);
 
-  if (was_default)
-  {
-    dump_default_clause_check_for_rewrite ();
-  }
+  dump_default_clause_check_for_rewrite ();
 
   lexer_seek (start_loc);
   next_token_must_be (TOK_OPEN_BRACE);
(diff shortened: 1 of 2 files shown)


egavrin:
Good to me

galpeter:
lgtm

egavrin:
`make push`

Issue #339, opened 2015-07-08
Assertion '!ecma_number_is_nan (res)' failed in lexer_parse_number


renatahodovan:
###### Jerry version:
    Checked revision: 146ac15
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
y = 4e;
```

###### Backtrace:
```javascript
ICE: Assertion '!ecma_number_is_nan (res)' failed at jerryscript/jerry-core/parser/js/lexer.cpp(lexer_parse_number):943.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x0000000000497a28 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
32	  SYSCALL_2
(gdb) bt
# 0  0x0000000000497a28 in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
# 1  0x00000000004003c2 in syscall_2 (syscall_no=62, arg1=32682, arg2=6) at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:96
# 2  0x0000000000400563 in abort () at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:175
# 3  0x0000000000447ae3 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.cpp:65
# 4  0x0000000000447b32 in jerry_assert_fail (assertion=0x4a4011 "!ecma_number_is_nan (res)", file=0x4a35e0 "jerryscript/jerry-core/parser/js/lexer.cpp", 
    function=0x49fc10 <_ZZL18lexer_parse_numbervE8__func__.lto_priv.177> "lexer_parse_number", line=943) at jerryscript/jerry-core/jrt/jrt-fatals.cpp:92
# 5  0x000000000041ba82 in lexer_parse_number() [clone .lto_priv.171] () at jerryscript/jerry-core/parser/js/lexer.cpp:943
# 6  0x0000000000411ae0 in lexer_parse_token () at jerryscript/jerry-core/parser/js/lexer.cpp:1241
# 7  0x0000000000412916 in lexer_next_token() () at jerryscript/jerry-core/parser/js/lexer.cpp:1483
[...]

PR #673, opened 2015-10-19
Fix lexing illegal exponential, Fix for issue #339

ILyoan:


ruben-ayrapetyan:
@ILyoan, `lexer_parse_number` currently parses numbers in two phases:
 - determine type of number literal, at the same time checking for syntax correctness;
 - convert characters to number value.

`ecma_utf8_string_to_number` itself checks for correctness of number literal, so the code works correctly.
On the other hand, the function doesn't report what specifically is wrong in an incorrect number literal, so in the case we can't output message that would be more detailed than "invalid number literal".

Because of this, it seems better to check for correctness of exponent in the first phase, like the following:
```cpp
is_exp = true;
consume_char ();

if (LA (0) == LIT_CHAR_MINUS
    || LA (0) == LIT_CHAR_PLUS)
{
  consume_char ();
}

if (!lit_char_is_decimal_digit (LA (0)))
{
  PARSE_ERROR (JSP_EARLY_ERROR_SYNTAX,
               "Exponent mark in a numeric literal should be followed by a signed integer",
               lit_utf8_iterator_get_pos (&src_iter));
}
```

ILyoan:
@ruben-ayrapetyan Ok, so any syntax checks should go into the first phase. It make sense. I'll update this PR.

ILyoan:
@ruben-ayrapetyan Fixed it and rebased on top of master

egavrin:
Fix commit message, please ^_^

```
Fix lexing of illegal exponential.

Related issue: #339

JerryScript-DCO-1.0-Signed-off-by: Ilyong Cho ily.cho@samsung.com
```

ILyoan:
@egavrin Thanks, updated commit message.

ruben-ayrapetyan:
Looks good to me

galpeter:
lgtm

egavrin:
`make push`

Issue #669, opened 2015-10-14
Unknown token error in the return statement


rtakacs:
<b>Jerry revision:</b>
<pre>08c618e</pre>

<b>Test case:</b>
```javascript
function f(){return}
```

It returns with <b>'ERROR: Ln 1, Col 20: Unknown token }'</b> however it is a valid JavaScript code.

<b>Expected behaviour:</b>
The functions should be parsed correctly that does nothing.

wateret:
And I found more examples that have similar problem:
```javascript
switch (1) {
  case 1:
    break}
```

```javascript
while (true) {continue}
```

PR #674, opened 2015-10-20
Fix: parse error when no expression follows after 'return', 'break', …

wateret:
When no expression follows after keyword 'return', 'break', 'continue' and `}` follows immediately after these keywords, throws an error.

Fix for #669.

### DIFF
--- a/jerry-core/parser/js/parser.cpp
+++ b/jerry-core/parser/js/parser.cpp
@@ -2831,6 +2831,11 @@ parse_statement (jsp_label_t *outermost_stmt_label_p) /**< outermost (first) lab
       }
     }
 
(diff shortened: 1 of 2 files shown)


jerry-core/parser/js/parser.cpp
```diff
@@ -2837,6 +2842,10 @@ parse_statement (jsp_label_t *outermost_stmt_label_p) /**< outermost (first) lab
     else
     {
       dump_ret ();
+      if (token_is (TOK_CLOSE_BRACE))
```
egavrin: Please, add extra empty line before condition.

egavrin:
Good to me

galpeter:
lgtm

egavrin:
`make push`

Issue #614, opened 2015-09-03
Assertion '!ecma_is_value_empty (reg_value)' failed in get_variable_value


renatahodovan:
###### Jerry version:
    Checked revision: a26c454
    Build: debug.linux
###### OS: 
    Ubuntu 15.04, x86_64
###### Test case:
```javascript
JSON.stringify & (Date = 1)
```

###### Backtrace:
```javascript

ICE: Assertion '!ecma_is_value_empty (reg_value)' failed at jerryscript/jerry-core/vm/opcodes-helpers-variables.cpp(get_variable_value):87.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x00000000004b445f in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
32    SYSCALL_2
(gdb) bt
# 0  0x00000000004b445f in syscall_2_asm () at jerryscript/jerry-libc/target/linux/jerry-asm.S:32
# 1  0x00000000004003c2 in syscall_2 (syscall_no=62, arg1=4162, arg2=6) at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:96
# 2  0x0000000000400563 in abort () at jerryscript/jerry-libc/target/linux/jerry-libc-target.c:175
# 3  0x000000000043de52 in jerry_fatal(jerry_fatal_code_t) (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:65
# 4  0x000000000043dea1 in jerry_assert_fail(char const*, char const*, char const*, unsigned int) (assertion=0x4d99b0 "!ecma_is_value_empty (reg_value)", 
    file=0x4d9918 "jerryscript/jerry-core/vm/opcodes-helpers-variables.cpp", 
    function=0x4b5960 <_ZZ18get_variable_valueP14vm_frame_ctx_thbE8__func__.lto_priv.1158> "get_variable_value", line=87)
    at jerryscript/jerry-core/jrt/jrt-fatals.cpp:97
# 5  0x00000000004aa574 in get_variable_value (frame_ctx_p=0x7fffffffd550, var_idx=131 '\203', do_eval_or_arguments_check=false)
    at jerryscript/jerry-core/vm/opcodes-helpers-variables.cpp:87
[...]

PR #679, opened 2015-10-21
Fix prop_getters stack manipulation for assignment expression. Fix for  #614

ILyoan:
Current implementation of Jerry checks whether the last dumped opcode is 'prop_getter' before treating rhs of assignment expression. If it is, it assumes lhs was temporary register operand which is not always true. 

`this.a; b = 1;` is a counter example. Here after parsing lsh of `b = 1`, the last opcode is still 'prop_getter' (result of `this.a`) while in actually the lhs is not temporary register operand and the last opcode is not standing for the lhs. Issue #614 reported it.

This PR is to fix it.

### DIFF
--- a/jerry-core/parser/js/opcodes-dumper.cpp
+++ b/jerry-core/parser/js/opcodes-dumper.cpp
@@ -628,25 +628,23 @@ dump_prop_setter_or_triple_address_res (vm_op_t opcode,
                                         jsp_operand_t res,
                                         jsp_operand_t op)
 {
-  const op_meta last = STACK_TOP (prop_getters);
-  if (last.op.op_idx == VM_OP_PROP_GETTER)
+  if (res.is_register_operand ())
   {
+    /*
+     * Left-hand-side must be a member expression and corresponding prop_getter
+     * op is on top of the stack.
+     */
+    const op_meta last = STACK_TOP (prop_getters);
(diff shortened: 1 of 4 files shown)


egavrin:
Please, change commit message to the following one:
```
Fix prop_getters stack manipulation for assignment expression. 

Related issue: #614

JerryScript-DCO-1.0-Signed-off-by: Ilyong Cho ily.cho@samsung.com
```

egavrin:
Good to me

ILyoan:
@egavrin Updated commit message.

ruben-ayrapetyan:
Looks good to me

jerry-core/parser/js/opcodes-dumper.cpp
```diff
@@ -1824,37 +1822,59 @@ rewrite_jump_to_end (void)
 }
 
 void
-start_dumping_assignment_expression (void)
+start_dumping_assignment_expression (jsp_operand_t lhs, __attribute__((unused)) locus loc)
```
galpeter: For the unused attr we have the following construct:
```c
method(int a, int b __attr_unused__)
```

ILyoan:
@galpeter Updated PR. also rebased on top of master.

galpeter:
lgtm

PR #684, opened 2015-10-22
Simplify [[DefineOwnProperty]] usage, use the helper function where p…

bzsolt:
…ossible.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects-arguments.cpp
+++ b/jerry-core/ecma/operations/ecma-objects-arguments.cpp
@@ -21,6 +21,7 @@
  */
 
 #include "ecma-alloc.h"
+#include "ecma-builtin-helpers.h"
 #include "ecma-builtins.h"
 #include "ecma-function-object.h"
 #include "ecma-gc.h"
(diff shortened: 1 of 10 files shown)


jerry-core/ecma/operations/ecma-objects-arguments.cpp
```diff
@@ -186,14 +167,11 @@ ecma_op_create_arguments_object (ecma_object_t *func_obj_p, /**< callee function
         {
           ecma_string_t *indx_string_p = ecma_new_ecma_string_from_uint32 ((uint32_t) indx);
 
-          prop_desc = ecma_make_empty_property_descriptor ();
```
galpeter: Why did we moved this out from here?

bzsolt: Because we should use a ```property_descriptor``` out of ```if (formal_params_p != NULL)``` block, in [L232](https://github.com/Samsung/jerryscript/pull/684/files?diff=unified#diff-c21c3dee8026ba5fc844a73046372a83R232).

egavrin:
Good to me

jerry-core/ecma/operations/ecma-objects-arguments.cpp
galpeter: I see. ok

galpeter:
lgtm

egavrin:
`make push`

galpeter:
Landed: https://github.com/Samsung/jerryscript/commit/4aaa0ace1749ab490cdf6c0885719d28f01b8112

Issue #681, opened 2015-10-22
Jerry API for retrieving heap stats.


ILyoan:
For profiling and tracing memory usage during running a program. It would be useful to have API for getting heap memory stats.

Interesting numbers are total heap size, allocated bytes, and peak bytes.

```
typedef struct jerry_api_heap_status_t 
{
  size_t heap_size;
  size_t allocated_bytes;
  size_t peak_allocated_bytes;
} jerry_api_heap_stats_t;

void jerry_api_heap_stats (jerry_api_heap_stats_t*);
```

egavrin:
Important numbers are:
```   Peak allocated chunks count = 1386
   Peak allocated = 33469 bytes
   Peak waste = 55235 bytes
```

Currently available:
```
Heap stats:
   Heap size = 258048 bytes
   Chunk size = 64 bytes
   Allocated chunks count = 0
   Allocated = 0 bytes
   Waste = 0 bytes
   Peak allocated chunks count = 1386
   Peak allocated = 33469 bytes
   Peak waste = 55235 bytes

Pools stats:
  Chunk size: 8
   Pools: 0
   Allocated chunks: 0
   Free chunks: 0
   Peak pools: 380
   Peak allocated chunks: 3039
```

egavrin:
Additionally, it'd be nice to have API for resetting statistics: `void jerry_api_mem_stats_reset ();` - since we don't need to know the allocated peek after parsing, for example.

ILyoan:
@egavrin Thanks for your opinion, I agree with you. I'll prepare PR for this issue.

zherczeg:
No plans to implement this at the moment.

PR #685, opened 2015-10-23
Adding Jerry api for triggering gc.

ILyoan:
Introducing new Jerry API for triggering GC.
```
void jerry_api_gc ()
```

### DIFF
--- a/jerry-core/jerry.cpp
+++ b/jerry-core/jerry.cpp
@@ -1318,6 +1318,17 @@ jerry_api_eval (const jerry_api_char_t *source_p, /**< source code */
 } /* jerry_api_eval */
 
 /**
+ * Perform GC
+ */
+void
+jerry_api_gc (void)
+{
+  jerry_assert_api_available ();
+
+  ecma_gc_run ();
+} /* jerry_api_gc */
+
+/**
  * Jerry engine initialization
  */
 void
--- a/tests/unit/test-api.cpp
+++ b/tests/unit/test-api.cpp
@@ -556,6 +556,9 @@ main (void)
   // cleanup.
   jerry_api_release_object (global_obj_p);
 
+  // TEST: run gc.
+  jerry_api_gc ();
+
   jerry_cleanup ();
 
   JERRY_ASSERT (test_api_is_free_callback_was_called);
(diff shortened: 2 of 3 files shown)


tests/unit/test-api.cpp
```diff
@@ -556,6 +556,9 @@ main (void)
   // cleanup.
   jerry_api_release_object (global_obj_p);
 
+  // TEST: run gc.
+  jerry_api_gc ();
```
egavrin: How can we ensure that GC actually performed sweeping?

jerry-core/jerry.cpp
```diff
@@ -1318,6 +1318,17 @@ jerry_api_eval (const jerry_api_char_t *source_p, /**< source code */
 } /* jerry_api_eval */
 
 /**
+ * Perform GC
+ */
+void
```
egavrin: I'd prefer to have amount of sweeped data as return value of `jerry-api-gc()`

ruben-ayrapetyan: Shouldn't we enable the API only in `MEM_STATS` mode in the case (as, in general configuration, gc has no exact information on exact amount of bytes freed)?

zherczeg: Usually calling GC is a black box, depending on the implementation it might be a deferred call, priority settings change, etc. So keeping it void is perfect for me (otherwise we just restrict ourselves).

egavrin: @zherczeg Agree.

tests/unit/test-api.cpp
ILyoan: @egavrin I just want to make sure that calling this api does cause any faults.

ILyoan:
@zherczeg Agree too.
As at least we can check how many memory space were reclaim by calling GC with separate API like `jerry_api_mem_stats()` discussed in #681.

egavrin:
Good to me.
@galpeter what do you think?

galpeter:
lgtm

egavrin:
Merged 1c21e16c7b735ca37e5806ba043d5527fceaf750

Issue #692, opened 2015-10-29
15.1.3 (URI Handling) chapter test262 failures


ruben-ayrapetyan:
Hello.

The following test262 tests related to 15.1.3 chapter are failing on current master (906558b8e94aa9118971188af9880eb47ff9b791):
- ch15/15.1/15.1.3/15.1.3.1/S15.1.3.1_A2.5_T1
- ch15/15.1/15.1.3/15.1.3.2/S15.1.3.2_A2.5_T1

egavrin:
Works on master (f17f785b6ec0c71c2ef37e4e80a98aff1eff0963)

PR #696, opened 2015-11-01
Fix ecma_builtin_global_object_decode_uri_helper

ruben-ayrapetyan:
@zherczeg, @dbatyai, please, check if the changes actually reflect semantics of decodeURI routine defined by ECMA-262 specification. Could you, please, also suggest some formal way to check current implementation on accordance to specification (except running test262 or some other test suite)?

However, the following test262 tests are fixed with the changes:
 - ch15/15.1/15.1.3/15.1.3.1/S15.1.3.1_A2.5_T1
 - ch15/15.1/15.1.3/15.1.3.2/S15.1.3.2_A2.5_T1

Related issue: #692

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
@@ -857,7 +857,12 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
 
       lit_code_point_t decoded_byte;
 
-      lit_read_code_point_from_hex (input_char_p + 1, 2, &decoded_byte);
+      if (!lit_read_code_point_from_hex (input_char_p + 1, 2, &decoded_byte))
+      {
+        ret_value = ecma_make_throw_obj_completion_value (ecma_new_standard_error (ECMA_ERROR_URI));
+        break;
+      }
+
       input_char_p += URI_ENCODED_BYTE_SIZE;
 
       if (decoded_byte <= LIT_UTF8_1_BYTE_CODE_POINT_MAX)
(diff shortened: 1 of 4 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -935,8 +940,8 @@ ecma_builtin_global_object_decode_uri_helper (ecma_value_t uri __attr_unused___,
         if ((bytes_count == 2 && cp <= LIT_UTF8_1_BYTE_CODE_POINT_MAX)
             || (bytes_count == 3 && cp <= LIT_UTF8_2_BYTE_CODE_POINT_MAX)
             || (bytes_count == 4 && cp <= LIT_UTF8_3_BYTE_CODE_POINT_MAX)
-            || lit_is_code_unit_high_surrogate ((ecma_char_t) cp)
-            || lit_is_code_unit_low_surrogate ((ecma_char_t) cp)
+            || (cp <= LIT_UTF16_CODE_UNIT_MAX && lit_is_code_unit_high_surrogate ((ecma_char_t) cp))
+            || (cp <= LIT_UTF16_CODE_UNIT_MAX && lit_is_code_unit_low_surrogate ((ecma_char_t) cp))
```
ruben-ayrapetyan: @zherczeg, @dbatyai, maybe instead of checking range of `cp` here, its better to change argument type of `lit_is_code_unit_high_surrogate` / `lit_is_code_unit_low_surrogate` to `lit_code_point_t`. What do you think about this?

zherczeg: Do you mean we should overwrite the arguments in the header? I like this idea.
LGTM otherwise

ruben-ayrapetyan: @zherczeg, could you, please, describe meaning of the conditions block so that it becomes more clear why do we check code point as a code unit here? Why do we check whether the code point is high / low surrogate?

dbatyai: The condition is there to check if the decoded code point is valid UTF-8 or not. Now that I think about it, we could replace the whole condition and simply use `!lit_is_utf8_string_valid (octets, bytes_count)` instead.

ruben-ayrapetyan: @dbatyai, @zherczeg, thank you for checking. I've updated pull request - the check replaced with call to `lit_is_utf8_string_valid`.

zherczeg:
LGTM

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
ruben-ayrapetyan: Seems that checking for equality to high / low surrogate is anyway required. I've added the checks.

egavrin:
Good to me

Issue #704, opened 2015-11-05
REPL support for Jerry?


wateret:
Most of Javascript Engines support REPL and I also think REPL is useful in many ways.
If binary size or something matters then we can give a build option for it. Is there some other issues?

PR #712, opened 2015-11-10
Support REPL for Linux

wateret:
REPL for LInux. It runs REPL when no JS files given from command line.
For #704.

### DIFF
--- a/main-unix.cpp
+++ b/main-unix.cpp
@@ -219,6 +219,8 @@ main (int argc,
   bool is_dump_snapshot_mode_for_global_or_eval = false;
   const char *dump_snapshot_file_name_p = NULL;
 
+  bool is_repl_mode = false;
+
 #ifdef JERRY_ENABLE_LOG
   const char *log_file_name = NULL;
 #endif /* JERRY_ENABLE_LOG */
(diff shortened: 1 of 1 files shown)


seanshpark:
+1
How about printing name and version when executed?
```
JerryScript version x.y.z
jerry>
```
BTW, do we have version string?

egavrin:
@seanshpark we have commit hash and branch, no versio yet.

```cpp
/**
 * Jerry engine build date
 */
const char *jerry_build_date = JERRY_BUILD_DATE;

/**
 * Jerry engine build commit hash
 */
const char *jerry_commit_hash = JERRY_COMMIT_HASH;

/**
 * Jerry engine build branch name
 */
const char *jerry_branch_name = JERRY_BRANCH_NAME;

```

galpeter:
Do I see correctly that we can specify the `is_dump_snapshot_mode` and repl mode at the same time? Is that what we want?

wateret:
@galpeter `is_repl_mode` is set to true when `files_counter == 0 && exec_snapshots_count == 0`. So basically Jerry turns to REPL mode when no files given. Plus, when options like `--exec-snapshot` or `--dump-snapshot` are given, we'll see an error just like current master.
```
$ ./jerry --exec-snapshot
Error: wrong format of the arguments
```

galpeter:
the PR seems ok to me, but we'll need to adapt this for the darwin main file also. IMHO: we'll need to come up a way to minimize differences (btw: do we need the darwin main at all?)

egavrin:
> btw: do we need the darwin main at all?
We don't.

wateret:
@egavrin, @galpeter Rebased to current master. Now that #819 is merged, can I push this PR?

galpeter:
@wateret yeah, its good to push.

LaszloLango:
Landed (0e88e819eb0f2c6e3c63e9b73246133b21a84059)

PR #719, opened 2015-11-12
Pass -Wno-error to linker in case of LTO builds

### LINKED ISSUE #718: make release.linux USE_COMPILER_DEFAULT_LIBC=YES fails
Lately, I've been trying to build Jerry on a platform which is not supported by jerry-libc (aarch64-linux), thus I had to fall back to the default glibc available there. Debug build worked fine, however, release build failed. Then it turned out that the error is also reproducible on an x86-64 box as well:

```
/usr/bin/x86_64-linux-gnu-g++    -ffixed-rbp    -ffixed-rbp -fno-builtin -flto -fno-fat-lto-objects -fno-stack-protector -g -gdwarf-4 -Wall -Wextra -pedantic -Wformat-nonliteral -Winit-self -Wno-stack-protector -Wconversion -Wsign-conversion -Wformat-security -Wmissing-declarations -Wno-attributes -Wfatal-errors -Werror -Wlogical-op -std=c++11 -fno-exceptions -fno-rtti -fno-implicit-templates -fno-implicit-inline-templates -Os -Wl,-z,noexecstack -flto  -static CMakeFiles/release.linux.dir/main-linux.cpp.o  -o release.linux  jerry-core/librelease.jerry-core.a third-party/fdlibm/librelease.jerry-fdlibm.third_party.lib.a /usr/lib/gcc/x86_64-linux-gnu/4.9/libgcc.a /usr/lib/gcc/x86_64-linux-gnu/4.9/libgcc_eh.a 
In function ‘__fread_alias’,
    inlined from ‘read_snapshot’ at /home/akiss/devel/jerry/jerryscript/main-linux.cpp:153:69,
    inlined from ‘main’ at /home/akiss/devel/jerry/jerryscript/main-linux.cpp:390:70:
/usr/include/x86_64-linux-gnu/bits/stdio2.h:290:66: error: call to ‘__fread_chk_warn’ declared with attribute warning: fread called with bigger size * nmemb than length of destination buffer [-Werror]
  return __fread_chk (__ptr, __bos0 (__ptr), __size, __n, __stream);
                                                                  ^
compilation terminated due to -Wfatal-errors.
lto-wrapper: /usr/bin/x86_64-linux-gnu-g++ returned 1 exit status
/usr/bin/ld: lto-wrapper failed
collect2: error: ld returned 1 exit status
make[4]: *** [release.linux] Error 1
```

It seems that a long-open gcc bug causes spurious warnings when LTO is enabled: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61886 And the -Werror flag turns these eventually into errors, and thus the build process fails. Unfortunately, there is no way to selectively disable the warning with pragmas or -Wno-error=xxx flags.

akiss77:
When linking a release-built command line shell on Linux against
default glibc with LTO enabled, a long-open gcc bug causes spurious
warning to be emitted around a call to fread. The -Werror flag
turns this into an error and the build process fails.
Unfortunately, there is no way to selectively disable the warning
with pragmas or -Wno-error=xxx flags.

Thus, this patch passes -Wno-error to the linker in case of LTO
builds. (But only to the linker. The compilation still applies the
warnings-are-errors policy.)

fixes #718

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -240,15 +240,31 @@ project (Jerry CXX C ASM)
   set(COMPILE_FLAGS_JERRY "${COMPILE_FLAGS_JERRY} -g -gdwarf-4")
 
  # Warnings
-  set(COMPILE_FLAGS_JERRY "${COMPILE_FLAGS_JERRY} -Wall -Wextra -pedantic")
-  set(COMPILE_FLAGS_JERRY "${COMPILE_FLAGS_JERRY} -Wformat-nonliteral -Winit-self -Wno-stack-protector")
-  set(COMPILE_FLAGS_JERRY "${COMPILE_FLAGS_JERRY} -Wconversion -Wsign-conversion -Wformat-security")
-  set(COMPILE_FLAGS_JERRY "${COMPILE_FLAGS_JERRY} -Wmissing-declarations -Wno-attributes")
-  set(COMPILE_FLAGS_JERRY "${COMPILE_FLAGS_JERRY} -Wfatal-errors")
+  macro(append variable value)
+    set(${variable} "${${variable}} ${value}")
+  endmacro()
+
+  macro(add_jerry_compile_flags)
+    foreach(_flag ${ARGV})
+      append(COMPILE_FLAGS_JERRY ${_flag})
+    endforeach()
+  endmacro()
+
+  macro(add_jerry_compile_warnings)
+    foreach(_warning ${ARGV})
+      add_jerry_compile_flags(-W${_warning})
+      if(CMAKE_COMPILER_IS_GNUCC OR CMAKE_COMPILER_IS_GNUCXX)
+        add_jerry_compile_flags(-Werror=${_warning})
+      endif()
+    endforeach()
+  endmacro()
+
+  add_jerry_compile_warnings(all extra format-nonliteral init-self conversion sign-conversion format-security missing-declarations)
+  add_jerry_compile_flags(-pedantic -Wno-stack-protector -Wno-attributes -Wfatal-errors)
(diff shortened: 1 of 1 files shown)


egavrin:
Do you have link for this issue?

akiss77:
Please, see at issue #718 , as mentioned above.

akiss77:
I've pushed a patch on top of the original to show the approach suggested by @ruben-ayrapetyan in #718 . The patches are intentionally not squashed for now so that we can compare and discuss the different approaches.

And yes, this also solves the problem. The warning is printed during release link but it causes no error.

akiss77:
I've improved the PR with macros to make the handling of warnings more compact and easier to maintain. Some of the macros could be utilized elsewhere in the CMakeLists.txt file as well but the changes are kept intentionally local for now.

CMakeLists.txt
```diff
@@ -240,15 +240,37 @@ project (Jerry CXX C ASM)
...
+  endmacro()
+
+  add_jerry_compile_warnings(all extra)
+  add_jerry_compile_flags(-pedantic)
+  add_jerry_compile_warnings(format-nonliteral init-self)
+  add_jerry_compile_flags(-Wno-stack-protector)
+  add_jerry_compile_warnings(conversion sign-conversion format-security)
+  add_jerry_compile_warnings(missing-declarations)
+  add_jerry_compile_flags(-Wno-attributes)
+  add_jerry_compile_flags(-Wfatal-errors)
```
ruben-ayrapetyan: Could you, please, merge the above lists into two - for warnings and for other flags?

ruben-ayrapetyan:
Looks good to me

LaszloLango:
LGTM

PR #729, opened 2015-11-20
Add option to enable all-in-one build mode

bzsolt:


jerry-core/CMakeLists.txt
```diff
@@ -121,8 +121,11 @@ project (JerryCore CXX C ASM)
    set(INCLUDE_THIRD_PARTY_VALGRIND ${CMAKE_SOURCE_DIR}/third-party/valgrind)
 
 # Sources
+ # All-in-one build
+ if("${ENABLE_ALL_IN_ONE}" STREQUAL "ON")
+   set(SOURCE_CORE jerry-all-in.cpp)
```
ruben-ayrapetyan: This implementation requires to manually update contents of `jerry-all-in.cpp` upon adding / removing / renaming a module.

It would be better, if the file would be created automatically during build and if it would be placed in cmake's build directory instead of repository root.

egavrin: +1

ruben-ayrapetyan:
Looks good to me

LaszloLango:
LGTM

Issue #725, opened 2015-11-19
Incorrect SyntaxError instead of ReferenceError


egavrin:
```js
switch(foo){} {}
```
Evaluation of this code should produce ReferenceError not SyntaxError:
```
switch(foo){} {}
               ^
ERROR: Ln 1, Col 16: Unexpected token
```

LaszloLango:
Works on current master (91a0514).

PR #730, opened 2015-11-23
Fix parse error of parse_switch_statement

wateret:
Related issue: #725

* Do not eat the last token of switch statement. (so it does not emit a SyntaxError)
* Add dump assignment bytecode if literal (so it emits a ReferenceError)


LaszloLango:
LGTM

egavrin:
Great! Good to me.

PR #759, opened 2015-12-03
Disable LTO by default in Mac OS X build.

lemmaa:
Because, debugger is not working properly with LTO build in Mac OS X
```
Reading symbols from build/bin/debug.darwin/jerry...
warning: `/tmp/lto.o': can't open to read symbols: No such file or directory.
(no debugging symbols found)...done.
(gdb) _
```
And enable ALL_IN_ONE as default, to minimize the performance degradation by disable LTO.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,6 +70,8 @@ project (Jerry CXX C ASM)
 
   option(STRIP_RELEASE_BINARY "Strip symbols from release binaries" ON)
  elseif("${PLATFORM}" STREQUAL "DARWIN")
+  option(ENABLE_LTO            "Enable LTO build" OFF)
+  option(ENABLE_ALL_IN_ONE     "Enable ALL_IN_ONE build" ON)
   set(PLATFORM_EXT "DARWIN")
   set(EXTERNAL_BUILD FALSE)
 
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,8 @@
 #   Parallel build of several selected targets started manually is not supported.
 #
 
+export TARGET_NATIVE_SYSTEMS = $(shell uname -s | tr '[:upper:]' '[:lower:]')
+
 # Options
  # Valgrind
   VALGRIND ?= OFF
@@ -64,7 +66,11 @@
   endif
 
  # LTO
-  LTO ?= ON
+  ifeq ($(TARGET_NATIVE_SYSTEMS),darwin)
+   LTO ?= OFF
+  else
+   LTO ?= ON
+  endif
 
   ifneq ($(LTO),ON)
    LTO := OFF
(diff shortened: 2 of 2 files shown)


CMakeLists.txt
```diff
@@ -60,9 +60,13 @@ project (Jerry CXX C ASM)
 # Defining options
  option(ENABLE_VALGRIND       "Enable valgrind helpers in memory allocators" OFF)
  option(ENABLE_VALGRIND_FREYA "Enable valgrind-freya helpers in memory allocators" OFF)
- option(ENABLE_LTO            "Enable LTO build" ON)
  option(ENABLE_LOG            "Enable LOG build" OFF)
  option(ENABLE_ALL_IN_ONE     "Enable ALL_IN_ONE build" OFF)
```
egavrin: I think we may enable ALL_IN_ONE by default, not to loose performance.

lemmaa:
@egavrin , I have updated the PR. Enabled ALL_IN_ONE by default a you commented.

egavrin:
LGTM

lemmaa:
@galpeter Would you please review this PR?

lemmaa:
@zherczeg @galpeter @LaszloLango please...

LaszloLango:
I think it would be better to check Clang compiler instead of darwin.

lemmaa:
@LaszloLango It's Apple Clang compiler. I checked this issue with 3 times of Apple Xcode update.

lemmaa:
@LaszloLango You mean that check `CMAKE_<LANG>_COMPILER_ID`? It doesn't work properly on Mac OS X currently.

LaszloLango:
@lemmaa, isn't this a compiler related issue? I don't know how the Mac OS X works, so maybe I'm not the right person to review this patch. :) Btw, the changes looks good, but I'm not sure this is the right fix.  @galpeter, can you take a look at this?

CMakeLists.txt
```diff
@@ -60,9 +60,14 @@ project (Jerry CXX C ASM)
 # Defining options
  option(ENABLE_VALGRIND       "Enable valgrind helpers in memory allocators" OFF)
  option(ENABLE_VALGRIND_FREYA "Enable valgrind-freya helpers in memory allocators" OFF)
- option(ENABLE_LTO            "Enable LTO build" ON)
  option(ENABLE_LOG            "Enable LOG build" OFF)
- option(ENABLE_ALL_IN_ONE     "Enable ALL_IN_ONE build" OFF)
+ if("${PLATFORM}" STREQUAL "DARWIN")
+  option(ENABLE_LTO            "Enable LTO build" OFF)
+  option(ENABLE_ALL_IN_ONE     "Enable ALL_IN_ONE build" ON)
```
galpeter: Would it be possbile to have this darwin case moved into the other elseif (... Darwin) which is down a few lines from here and have here only the default option? If that's not possible then the PR is ok for me. However we should investigate that is this only Mac related or clang related issue (but that's another topic for now imho)

lemmaa:
@galpeter Good point. I merged darwin case together and rebased the PR.

galpeter:
lgtm

Issue #762, opened 2015-12-03
Build fails with enabled Valgrind support


bzsolt:
Build fails with enabled Valgrind support for {debug,release}.mcu_stm32f* targets.
Seemingly unused params and variables triggered these errors:
```
jerryscript/third-party/valgrind/valgrind.h:3714:29: error: unused parameter 'format' [-Werror=unused-parameter]
    VALGRIND_PRINTF(const char *format, ...)
```
```
jerryscript/jerry-core/mem/mem-heap.cpp:337:62: error: statement has no effect [-Werror=unused-value]
    VALGRIND_NOACCESS_SPACE (mem_heap.area, MEM_HEAP_AREA_SIZE);
```
```
jerryscript/jerry-core/mem/mem-poolman.cpp:312:13: error: unused variable 'vbits' [-Werror=unused-variable]
    uint8_t vbits[MEM_POOL_CHUNK_SIZE];
```

ruben-ayrapetyan:
Valgrind-enabled build is not supported for `mcu` targets. We need to add some "Configuration is unsupported"-like message, after which build process should exit with fail status code.

PR #790, opened 2015-12-21
Test the build options

bzsolt:
Now the precommit checks only the {release,debug} native targets and unittests
with all combinations. Because testing on every target needs a lots of efforts,
added a new target called "precommit_heavy".
This target checks all targets with every combinations.


bzsolt:
The precommit fail related to #762.

LaszloLango:
@bzsolt, please rebase to the current master

CMakeLists.txt
```diff
@@ -113,9 +113,9 @@ project (Jerry C ASM)
   set(MCU_SCRIPT_GENERATED_HEADER ${CMAKE_BINARY_DIR}/generated.h)
 
 # Should we use external libc?
- if(DEFINED USE_COMPILER_DEFAULT_LIBC AND USE_COMPILER_DEFAULT_LIBC STREQUAL "YES")
+ if(DEFINED COMPILER_DEFAULT_LIBC AND COMPILER_DEFAULT_LIBC STREQUAL "ON")
```
LaszloLango: Why did you rename this? Are you sure you changed all reference of `USE_COMPILER_DEFAULT_LIBC`?

zherczeg:
LGTM

zherczeg:
Landed in b8b587f76c41fcd3b46550642a64d0fad34f5610.

zherczeg:
Sorry wrong PR :)

Makefile
```diff
@@ -332,6 +349,8 @@ precommit: prerequisites
 	$(Q)+$(MAKE) --no-print-directory clean
 	$(Q) echo "Running checks..."
 	$(Q)+$(MAKE) --no-print-directory check-signed-off check-vera check-cppcheck
+	$(Q) echo "...testing build-options..."
+	$(Q)+$(MAKE) --no-print-directory test-buildoptions
```
akiss77: I'm concerned about performance here. I quite often just run `make` (== `make precommit`) to see that everything is all right (checks, builds, tests). However, with this addition, this will perform all the builds with various options too. And this will call out to a shell script, which sequentially walks through targets and options, and then calls back to make again, thus loosing all parallelization potential. (Not to mention that ninja builds are way faster than non-ninja builds, but even if I run `make precommit NINJA=1`, because of the make->shell->make calls, this setting will be lost and loose ninja as well.) Is there no way to get everything done within the Makefile?

CMakeLists.txt
bzsolt: Yes, I've changed all occurences of ```USE_COMPILER_DEFAULT_LIBC```.
The ```YES/NO``` -> ```ON/OFF``` modification just simplify the testing-mechanism, because only this option use ```YES/NO```, the others use ```ON/OFF```. 
In my opinion, the ```ON/OFF``` is more obvious without the ```USE_``` prefix.

LaszloLango: OK, thanks for explanation.

Makefile
bzsolt: You're right, I will check the way to move this logic to the Makefile.

bzsolt:
I've updated the PR, @akiss77 could you check it please?

Makefile
```diff
@@ -311,6 +324,19 @@ check-vera:
 check-cppcheck:
 	$(Q) $(call SHLOG,./tools/check-cppcheck.sh,$(OUT_DIR)/cppcheck.log,Cppcheck)
 
+.PHONY: test-buildoptions
+test-buildoptions:
+	$(Q) for target in $(JERRY_BUILD_OPTIONS_TEST_TARGETS); do \
+           if [[ $$target =~ "mcu" ]]; then \
+             current_options=$$BUILD_OPTIONS_TEST_MCU; \
```
akiss77: We are in the Makefile, so we can use `$(BUILD_OPTIONS_TEST_MCU)`, we don't have to rely on the make variable being exported to a shell variable.

Makefile
```diff
@@ -311,6 +324,19 @@ check-vera:
...
+           else \
+             current_options=$$BUILD_OPTIONS_TEST_NATIVE; \
```
akiss77: ditto

Makefile
akiss77: It should be discussed and decided whether we want this to be included in the precommit target or have a separate target for the "heavy" checks (as was the case in a previous version, if I recall correctly). What's the impact of this addition on the precommit time? (Ofc, it can also be a valid answer that we don't care and want this tested, whatever it takes.)

Makefile
```diff
@@ -311,6 +324,19 @@ check-vera:
...
+           fi; \
+           for option in $$current_options; do \
+             $(MAKE) $$option=ON $$target || exit 1; \
+           done \
+         done
```
akiss77: What still bothers me is that these for loops are still sequential. Cannot we declare a make target for each of the `$(MAKE) $$option=ON $$target` commands and have all those targets be a dependecy of test-buildoptions? Then make would take care of parallelization, if it's available on the system. (Ex: test-js.*.* targets)

Makefile
bzsolt: With this version the runtime is more than twice before.
But now when the option-testing running in parallel and the build-option step included into the precommit process, the runtime only increased to ~5:55 minutes from ~3:25 on my local machine. So basically option-testing takes ~2:30 minutes.
What do you think, is this slow-down acceptable or not?

Makefile
```diff
@@ -196,7 +216,9 @@ define WRITE_TOOLCHAIN_CONFIG
 endef
 
 .PHONY: $(BUILD_DIR)/$(NATIVE_SYSTEM)/toolchain.config
-$(BUILD_DIR)/$(NATIVE_SYSTEM)/toolchain.config:
+.PHONY: $(BUILD_DIR)/$(NATIVE_SYSTEM)/unittests/toolchain.config
```
bzsolt: The cause of this (and related) modifications is that the parallel native builds can interfere with the unittests build.

akiss77:
LGTM (please rebase)

LaszloLango:
LGTM
@zherczeg, please check the changes.

zherczeg:
LGTM

PR #804, opened 2016-01-14
targets: merge js2c.py into targets/tools folder

seanshpark:
* also add .PHONY to target makefiles


targets/tools/js2c.py
```diff
@@ -45,7 +45,7 @@ def removeWhitespaces(code):
     return re.sub('\n+', '\n', re.sub('\n +', '\n', code))
 
 
-LICENSE = '''/* Copyright 2015 Samsung Electronics Co., Ltd.
+LICENSE = '''/* Copyright 2016 Samsung Electronics Co., Ltd.
```
egavrin: I think it should be `2015-2016` too.

egavrin:
LGTM

galpeter:
lgtm

seanshpark:
thanks for quick review :)

PR #805, opened 2016-01-17
targets: add nuttx interpreter build

seanshpark:

### DIFF
--- a/targets/nuttx-stm32f4/Makefile.nuttx
+++ b/targets/nuttx-stm32f4/Makefile.nuttx
@@ -1,0 +1,69 @@
+# Copyright 2016 Samsung Electronics Co., Ltd.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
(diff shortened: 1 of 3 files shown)


seanshpark:
any one interested to review? :)

zherczeg:
In general argument, function comments are missing from main-nuttx.cpp. I think there was a discussion to reduce the number of main-**.cpp files. Do we really need a separated file this platform?

tinysense:
@zherczeg , well, this PR was before merging main-*.cpp and I wasn't thinking of  adding snapshot for arm/nuttx.

seanshpark:
@zherczeg , well,I wasn't thinking of adding snapshot to arm/nuttx and when I made this code, linux and darwin was in separate files. And I think for embed target ports, I prefer more simple codes as an example and entry.

zherczeg:
Is it still a valid PR or you plan some changes?

seanshpark:
Hope to merge as is but if you think better to use main-unix.cpp, I need to check how it works.

LaszloLango:
The main reason of merging `main-linux.cpp` and `main-darwin.cpp` was that they were almost identical. I think it is OK to place in a separate file, because `main-nuttx.cpp` is different.

zherczeg:
Ok I was just asking.

zherczeg:
LGTM, though it would be good to reduce code duplications on these main-*.cpp files in a followup patches.

LaszloLango:
Landed (ef1252e29e5777ed8010e7a02a0c4c1c62530201)

Issue #716, opened 2015-11-11
Change arguments of `assert` to explicit boolean values in internal test suite


ruben-ayrapetyan:
Some tests in Jerryscript's internal test suite that are currently not included into precommit testing contain code that provides non-boolean values to `assert` function.

Because current implementation of `assert` accepts only explicit boolean true values, the tests that pass non-boolean values, are currently failing during full run of internal test suite as described in #711.

Maybe, the tests should be updated like in https://github.com/Samsung/jerryscript/pull/498/files#diff-c7d78eb307700cc443dc3324384a4033 and included into precommit testing list.

LaszloLango:
#814 is merged so we can close this issue.

Issue #711, opened 2015-11-10
jerry-test-suite produces strange result


chunseoklee:
```sh ./tools/runners/run-test-suite-jerry.sh  ./build/bin/debug.linux/jerry```
There are 15 failed test for test above.
But, some tests(15.5.2.1-001.js,15.5.4.7.00x.js ) seems to work well in my own test : make separate test.js file.

**Comment by seanshpark** [comment]
```assert(true);
```
is ok.
```
  assert(1);
```
fails with `Script assertion failed`

galpeter:
The `assert(1)` fails because the assert's implementation currently only checks that the argument is a true boolean or not. (The implementation is in the main-linux.cpp file). I can create a PR which adds a case that if the argument is an integer 1 then it accept just like the boolean true.

akiss77:
Shouldn't any non-0 integer count as true, not only integer 1?

galpeter:
@akiss77, sure we can do that also, and maybe it's even better.

galpeter:
@akiss77 , should 0.01 be also a valid true? :)

akiss77:
```
>>> Boolean(0)
false
>>> Boolean(1)
true
>>> Boolean(2)
true
>>> Boolean(0.0)
false
>>> Boolean(1.0)
true
>>> Boolean(0.01)
true
>>> Boolean(NaN) 
false
```

LaszloLango:
IMHO not only integers, but every number should be true except zero.

akiss77:
What's more:
```
>>> Boolean("")
false
>>> Boolean("non-empty string")
true
>>> Boolean({})
true
```
I'd try to cast everything to boolean. Whatever yields true, should be accepted.

galpeter:
hmmm then the best way would be to expose an API call which can correctly convert the values to bool.

lemmaa:
@LaszloLango @akiss77 @galpeter, I'm currently working on it. I'm trying to define new API `jerry_apt_value_to_char_buffer()` to get character representation of `jerry_api_value_t` object. I conclude that it is best way to solve this issue after short investigation. I think it can be used to REPL implementation of @wateret. If you agree, I will keep going.

````
ssize_t
jerry_api_value_to_char_buffer (const jerry_api_value_t *value_p, /**< value descriptor */
                                 jerry_api_char_t *buffer_p, /**< output characters buffer */
                                 ssize_t buffer_size) /**< size of output buffer */

````

[...]

PR #810, opened 2016-01-19
Add ToBoolean api call

galpeter:
Implement a way to convert the api values to boolean values
according to the ECMA standard. Also use this method for the
assert implementation.

### DIFF
--- a/jerry-core/jerry.cpp
+++ b/jerry-core/jerry.cpp
@@ -2105,3 +2105,23 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
   return JERRY_COMPLETION_CODE_INVALID_SNAPSHOT_VERSION;
 #endif /* !JERRY_ENABLE_SNAPSHOT */
 } /* jerry_exec_snapshot */
+
+
+/**
+ * Call the ToBoolean ecma builtin operation on the api value.
+ *
+ * @return true or false depending on the ToBoolean operaion.
+ */
+bool jerry_api_value_to_boolean (const jerry_api_value_t *in_value_p) /** < input value */
+{
+  jerry_assert_api_available ();
+
+  ecma_value_t in_value;
+  jerry_api_convert_api_value_to_ecma_value (&in_value, in_value_p);
+
+  ecma_completion_value_t bool_completion_value = ecma_op_to_boolean (in_value);
+
+  ecma_free_value (in_value, true);
+
+  return ecma_is_completion_value_normal_true (bool_completion_value);
+} /* jerry_api_value_to_boolean */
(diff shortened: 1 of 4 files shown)


main-darwin.cpp
```diff
@@ -123,9 +123,7 @@ assert_handler (const jerry_api_object_t *function_obj_p __attr_unused___, /** <
                 const jerry_api_value_t args_p[], /** < function arguments */
                 const jerry_api_length_t args_cnt) /** < number of function arguments */
 {
-  if (args_cnt == 1
-      && args_p[0].type == JERRY_API_DATA_TYPE_BOOLEAN
-      && args_p[0].v_bool == true)
+  if (args_cnt == 1 &&  && jerry_api_value_to_boolean (&args_p[0]))
```
ruben-ayrapetyan: The conversion hides the some kind of issues. For example, `assert ("abc".replace("abc", "$x$$5$0$00$" === "$x$5$0$00$"));` should be `assert ("abc".replace("abc", "$x$$5$0$00$") === "$x$5$0$00$");`, but assertion passes, and so, possible issue becomes hidden.

Please, see also https://github.com/Samsung/jerryscript/issues/711#issuecomment-155385250 and issue #716.

jerry-core/jerry.cpp
```diff
@@ -2105,3 +2105,23 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
   return JERRY_COMPLETION_CODE_INVALID_SNAPSHOT_VERSION;
 #endif /* !JERRY_ENABLE_SNAPSHOT */
 } /* jerry_exec_snapshot */
+
+
```
ruben-ayrapetyan: One newline seems to be redundant.

main-darwin.cpp
LaszloLango: @ruben-ayrapetyan, @galpeter, please check: #814

zherczeg:
Is this still relevant?

LaszloLango:
After fixing the tests this patch is not a blocker of enabling all of them. Besides that I prefer this change and I think this is how an assert function supposed to behave. @ruben-ayrapetyan what do you think? Should we close or update this PR?

zherczeg:
I would close it, since a stricter requirement strengthens testing.

akiss77:
@zherczeg @LaszloLango Even if it wont be used in the assert implementation, shouldn't the `jerry_api_value_to_boolean` function be kept?

zherczeg:
Yes, the function can be kept. Please update the patch.

LaszloLango:
In this case I suggest to add all of the converting functions in one patch like it was mentioned in #711

galpeter:
@LaszloLango, @zherczeg: okay I'll update the PR with a few conversion calls.

zherczeg:
Any plans with this PR?

LaszloLango:
This PR is out of date and inactive for a long time, so we have to close it.

PR #815, opened 2016-01-22
Print messages when error happens

rtakacs:
Print error messages to be more informative.

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -1704,7 +1704,8 @@ jerry_reg_err_callback (jerry_error_callback_t callback) /**< pointer to callbac
  */
 bool
 jerry_parse (const jerry_api_char_t *source_p, /**< script source */
-             size_t source_size) /**< script source size */
+             size_t source_size, /**< script source size */
+             jerry_api_object_t **error_obj_p) /**< [out] error object */
 {
   jerry_assert_api_available ();
 
@@ -1717,11 +1718,12 @@ jerry_parse (const jerry_api_char_t *source_p, /**< script source */
 
   parse_status = parser_parse_script (source_p,
                                       source_size,
-                                      &bytecode_data_p);
+                                      &bytecode_data_p,
+                                      error_obj_p);
 
   if (parse_status != JSP_STATUS_OK)
   {
-    JERRY_ASSERT (parse_status == JSP_STATUS_SYNTAX_ERROR || parse_status == JSP_STATUS_REFERENCE_ERROR);
+    JERRY_ASSERT (parse_status == JSP_STATUS_SYNTAX_ERROR);
 
     return false;
   }
@@ -1747,11 +1749,11 @@ jerry_parse (const jerry_api_char_t *source_p, /**< script source */
  * @return completion status
  */
 jerry_completion_code_t
-jerry_run (void)
+jerry_run (jerry_api_object_t **error_obj_p)
 {
   jerry_assert_api_available ();
 
-  return vm_run_global ();
+  return vm_run_global (error_obj_p);
 } /* jerry_run */
 
 /**
@@ -1767,8 +1769,9 @@ jerry_run_simple (const jerry_api_char_t *script_source, /**< script source */
   jerry_init (flags);
 
   jerry_completion_code_t ret_code = JERRY_COMPLETION_CODE_OK;
+  jerry_api_object_t *error_obj_p = NULL;
 
-  if (!jerry_parse (script_source, script_source_size))
+  if (!jerry_parse (script_source, script_source_size, &error_obj_p))
   {
     /* unhandled SyntaxError */
     ret_code = JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION;
@@ -1777,7 +1780,7 @@ jerry_run_simple (const jerry_api_char_t *script_source, /**< script source */
   {
     if ((flags & JERRY_FLAG_PARSE_ONLY) == 0)
     {
-      ret_code = jerry_run ();
+      ret_code = jerry_run (&error_obj_p);
     }
   }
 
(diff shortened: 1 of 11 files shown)


jerry-core/ecma/operations/ecma-exceptions.cpp
```diff
@@ -29,6 +31,76 @@
  * @{
  */
 
+void
+ecma_print_error_messages (ecma_object_t *obj_p)
```
ruben-ayrapetyan: Please, add comments.

ruben-ayrapetyan: The function is better to be named like `ecma_print_exception_message`

ruben-ayrapetyan:
@rtakacs, could you, please, also implement pass of SyntaxError's / ReferenceError's messages from parser to corresponding exception object, and turn off print of the message during parser operation?

See also:
 #724, #726

rtakacs:
@ruben-ayrapetyan: of course :)

LaszloLango:
@rtakacs, please update to the current master.

rtakacs:
@LaszloLango: Before I update this PR I would like to add error object logic to the parser as well.

rtakacs:
I've uploaded a WIP of my work. Any suggestions are welcome. It works for file and for REPL as well.

Some output: 

```
> a = 1;
> b ====
SyntaxError: Primary expression expected. [Line: 2, Column: 6]
```
```
> foo
ReferenceError: foo is not defined.
```

LaszloLango:
@rtakacs, I think the string concatenation should be on the client side. The core should return with an error object like:
```C++
struct error_obj {
  uint8_t type;
  ecma_string_t *msg;
  uint16_t line;
  uint16_t column;
};
```

Similar to `parser_error_location` used in `parser_parse_source`

jerry-core/jerry.c
```diff
@@ -2055,10 +2058,17 @@ jerry_parse_and_save_snapshot (const jerry_api_char_t *source_p, /**< script sou
...
 
   if (is_for_global)
   {
-    parse_status = parser_parse_script (source_p, source_size, &bytecode_data_p);
+    parse_status = parser_parse_script (source_p, source_size, &bytecode_data_p, &error_obj_p);
+
+    if (parse_status != JSP_STATUS_OK)
+    {
+      JERRY_ASSERT (error_obj_p != NULL);
+      ecma_deref_object (error_obj_p);
```
zherczeg: Deref object?

jerry-core/parser/js/js-parser.c
```diff
@@ -2254,19 +2254,22 @@ parser_set_show_instrs (int show_instrs) /**< flag indicating whether to dump by
...
-
 /**
  * Parse EcamScript source code
  */
 jsp_status_t
 parser_parse_script (const jerry_api_char_t *source_p, /**< source code */
                      size_t size, /**< size of the source code */
-                     ecma_compiled_code_t **bytecode_data_p) /**< result */
+                     ecma_compiled_code_t **bytecode_data_p, /**< [out] JS bytecode */
+                     jerry_api_object_t **error_obj_p __attr_unused___) /**< [out] error object */
```
zherczeg: Why unused?

tests/unit/test-api.c
```diff
@@ -332,13 +332,14 @@ main (void)
   jerry_api_object_t *global_obj_p, *obj_p;
   jerry_api_object_t *external_func_p, *external_construct_p;
   jerry_api_object_t *throw_test_handler_p;
+  jerry_api_object_t *err_obj_p = NULL;
   jerry_api_value_t res, args[2];
   char buffer[32];
 
-  is_ok = jerry_parse ((jerry_api_char_t *) test_source, strlen (test_source));
+  is_ok = jerry_parse ((jerry_api_char_t *) test_source, strlen (test_source), &err_obj_p);
   JERRY_ASSERT (is_ok);
```
zherczeg: is_ok && err_obj_p == NULL

jerry-core/jerry.c
LaszloLango: If an unhandled exception occurs, then the error object parameter will be non-null and someone must to free it.

tests/unit/test-api.c
```diff
@@ -332,13 +332,14 @@ main (void)
...
 
-  is_ok = (jerry_run () == JERRY_COMPLETION_CODE_OK);
+  is_ok = (jerry_run (&err_obj_p) == JERRY_COMPLETION_CODE_OK);
   JERRY_ASSERT (is_ok);
```
zherczeg: Ditto. Please check that the err_obj_p is NULL.

jerry-core/parser/js/js-parser.c
LaszloLango: Compiler complains about it is set but not used.

LaszloLango:
@zherczeg, updated

jerry-core/parser/js/js-parser.c
zherczeg: Where? It seems to me a bug if we create an object and don't use (free) it.

LaszloLango: This is an output parameter. What do you suggest? If we free it here, then we are not able to print it in `main-unix`

tests/unit/test-api.c
```diff
@@ -332,14 +332,15 @@ main (void)
...
   jerry_api_object_t *external_func_p, *external_construct_p;
   jerry_api_object_t *throw_test_handler_p;
+  jerry_api_object_t *err_obj_p = NULL;
   jerry_api_value_t res, args[2];
   char buffer[32];
 
-  is_ok = jerry_parse ((jerry_api_char_t *) test_source, strlen (test_source));
-  JERRY_ASSERT (is_ok);
+  is_ok = jerry_parse ((jerry_api_char_t *) test_source, strlen (test_source), &err_obj_p);
+  JERRY_ASSERT (is_ok && err_obj_p == NULL);
```
akiss77: Note: It's better practice to split and'ed asserts into two. In release, two asserts cost the same as one (i.e., nothing). However, if it comes to debugging, it's easier if the assert points exactly at source of failure. I.e., with the `&&` syntax, there will be a need for an extra round of execution to find out whether the `is_ok` or the `err_obj_p == NULL` part failed. With two asserts, that's not an issue.

LaszloLango:
@zherczeg, updated

jerry-core/jerry.c
```diff
@@ -2361,7 +2372,14 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
 
-    ret_code = vm_run_global ();
+    ecma_object_t *error_obj_p = NULL;
+
+    ret_code = vm_run_global (&error_obj_p);
+
+    if (ret_code == JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION)
+    {
+      ecma_deref_object (error_obj_p);
+    }
```
zherczeg: There should be an ASSERT here that error_obj_p is NULL or not-NULL depending on the ret_code

tests/unit/test-api.c
zherczeg: I don't mind.

zherczeg:
LGTM after the issues above fixed.

PR #825, opened 2016-01-27
Fix Vera++ rule for switch statements.

LaszloLango:
Don't warn on continue statements without comment in a switch-case.

### DIFF
--- a/tools/vera++/scripts/rules/jerry_switch_case.tcl
+++ b/tools/vera++/scripts/rules/jerry_switch_case.tcl
@@ -127,7 +127,9 @@ foreach fileName [getSourceFileNames] {
                                 set state "wait-for-break"
                                 continue
                             }
-                        } elseif {$next_token_type == "break" || $next_token_type == "return"} {
+                        } elseif {$next_token_type == "break"
+                                  || $next_token_type == "continue"
+                                  || $next_token_type == "return"} {
                             set state "wait-for-semicolon"
                             continue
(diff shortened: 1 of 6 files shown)


zherczeg:
LGTM! Those fallthrus are so annoying.

tools/vera++/scripts/rules/jerry_switch_case.tcl
```diff
@@ -163,7 +165,10 @@ foreach fileName [getSourceFileNames] {
                             set state "fallthru"
                             set line_num [lindex $next_token 1]
                             continue
-                        } elseif {$next_token_type == "break" || $next_token_type == "return" || $next_token_type == "goto"} {
+                        } elseif {$next_token_type == "break"
+                                  || $next_token_type == "continue"
+                                  || $next_token_type == "return"
+                                  || $next_token_type == "goto"} {
```
ruben-ayrapetyan: Do we use `goto` anywhere?

ruben-ayrapetyan: I don't think that `goto` is really good idea.

LaszloLango: There is `goto` in several places in the current code. This patch is about only the vera++ check. Please open a separate issue for eliminating `goto` and suggest your solution. I also don't like `goto` too much, but it can be handy in some cases. :)

PR #828, opened 2016-01-29
Remove jerry_ref_unused_variables template function

rtakacs:
Removed this function since it was unused and had a C++ feature.

### DIFF
--- a/jerry-core/jrt/jrt.h
+++ b/jerry-core/jrt/jrt.h
@@ -126,7 +126,7 @@ extern void __noreturn jerry_unimplemented (const char *, const char *, const ch
 /**
  * Mark for unreachable points and unimplemented cases
  */
-template<typename... values> extern void jerry_ref_unused_variables (const values & ... unused);
+extern void jerry_ref_unused_variables (void *, ...);
 
 #ifndef JERRY_NDEBUG
 #define JERRY_UNREACHABLE() \
--- a/jerry-core/ecma/builtin-objects/ecma-builtins-internal.h
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins-internal.h
@@ -93,20 +93,4 @@ extern void \
 ecma_builtin_ ## lowercase_name ## _sort_property_names (void);
 #include "ecma-builtins.inc.h"
 
-
-#ifndef CONFIG_ECMA_COMPACT_PROFILE
-# define ECMA_BUILTIN_CP_UNIMPLEMENTED(...) \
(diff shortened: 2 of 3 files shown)


jerry-core/jrt/jrt.h
```diff
@@ -108,26 +108,14 @@ extern void __noreturn jerry_unimplemented (const char *, const char *, const ch
...
+#define JERRY_DDDLOG(...)
 #endif /* !JERRY_ENABLE_LOG */
 
 #define JERRY_ERROR_MSG(...) jerry_port_errormsg (__VA_ARGS__)
 #define JERRY_WARNING_MSG(...) JERRY_ERROR_MSG (__VA_ARGS__)
 
-/**
- * Mark for unreachable points and unimplemented cases
- */
-template<typename... values> extern void jerry_ref_unused_variables (const values & ... unused);
```
ruben-ayrapetyan: Couldn't we implement the same using macro with `__VA_ARGS__` and `(void) var_name;` statements?

Seems that its better to use `__attr_unused__` as least as possible, because the attribute could hide actual issues.

rtakacs: Hm, I think I understand you. In this case the easiest way is just remove the template expression from the jerry_ref_unused_variables.

Note:
  * Removed ECMA_BUILTIN_CP_UNIMPLEMENTED macros since every built-in are implemented.
  * When I ran a precommit I got the following messages:

[jerry-core/parser/regexp/re-compiler.cpp:53]: (style) A pointer can not be negative so it is either pointless or an error to check if it is not.
[jerry-core/parser/regexp/re-compiler.cpp:62]: (style) A pointer can not be negative so it is either pointless or an error to check if it is not.

The static cppchecker emits these errors. Do you have any ideas how to eliminate these?

ruben-ayrapetyan: @rtakacs, I'm unsure why the warning is issued. It looks like a bug in the static checker.

Maybe, it could be solved by replacing
```JERRY_ASSERT (bc_ctx_p->block_end_p - bc_ctx_p->block_start_p >= 0);```
with
```JERRY_ASSERT (bc_ctx_p->block_end_p >= bc_ctx_p->block_start_p);```

rtakacs: Thanks Ruben, your advice worked ;)

ruben-ayrapetyan:
Looks good to me

LaszloLango:
LGTM

PR #833, opened 2016-02-02
Use concrete types instead of 'auto'.

rtakacs:


LaszloLango:
We should use consistence parameter order in ECMA_GET_POINTER and ECMA_SET_POINTER macros. In ECMA_GET_POINTER the first parameter defines the type of the pointer.

zherczeg:
I agree. Btw I don't see where the macro is changed to accept 3 arguments.

LaszloLango:
@zherczeg, jerry-core/ecma/base/ecma-helpers.h:53

rtakacs:
You are right. I've updated the PR.

zherczeg:
I was thinking whether we need the types for setting compressed pointers (the target is always an unsigned int variable), but perhaps because of the getter macros this approach is more consistent. I am happy with either way. Consider this as an LGTM from my side.

jerry-core/mem/mem-allocator.h
```diff
@@ -96,10 +96,10 @@ typedef void (*mem_try_give_memory_back_callback_t) (mem_try_give_memory_back_se
  * Set value of compressed pointer so that it will correspond
  * to specified non_compressed_pointer
  */
-#define MEM_CP_SET_POINTER(cp_value, non_compressed_pointer) \
+#define MEM_CP_SET_POINTER(pointer_type, non_compressed_pointer, cp_value) \
   do \
   { \
-    auto __temp_pointer = non_compressed_pointer; \
+    pointer_type *__temp_pointer = non_compressed_pointer; \
```
ruben-ayrapetyan: Upon being introduced, this check was intented to catch rvalues, not to check type.
It was supposed that lvalue could be evaluated multiple times without side effects. However, the assumption is not correct. So, maybe we should introduce another type of check here, or reimplement the macro without double evaluation, like the following:
```
#define MEM_CP_SET_POINTER(non_compressed_pointer, cp_value)
  do
  {
    void *ptr_value = (void *) non_compressed_pointer;

    if (unlikely ((ptr_value) == NULL))
    {
      (cp_value) = MEM_CP_NULL;
    }
    else
    {
      MEM_CP_SET_NON_NULL_POINTER (cp_value, ptr_value);
    }
  } while (false);
```

Anyway, I don't mind of introducing type check in the macro; previously it was implemented like the following:
```
#define ECMA_SET_POINTER(field, non_compressed_pointer) \
  do \
  { \
    void *__temp_pointer = non_compressed_pointer; \
    non_compressed_pointer = __temp_pointer; \
  } while (0); \
  \
  (field) = (unlikely ((non_compressed_pointer) == NULL) ? ECMA_NULL_POINTER \
                                                         : (mem_compress_pointer (non_compressed_pointer) \
                                                            & ((1u << ECMA_POINTER_FIELD_WIDTH) - 1)))
```

rtakacs:
Thanks your suggestions, I have updated the PR.

zherczeg:
LGTM

PR #834, opened 2016-02-02
Revive mem_stats.sh

akiss77:
The tool became bitrotten. Bringing it up-to-date to handle the
output of jerry and rss-measure.sh properly.


LaszloLango:
I think the combination of  `mem_stats.sh` and `run-perf.sh`would be a great tool. The current method used in `run-perf.sh` for memory measurement is quite inaccurate. Apart from that, this path looks good to me.

LaszloLango:
Landed (9c124e263436968f169de9df2b45ccb2473ddcc1)

PR #836, opened 2016-02-03
remove the c++ syntax, struct::xxx

jiangzidong:
jerry now has the c++ syntax, struct_name:xxx and enum_name:xxx
such as `sizeof (ecma_collection_chunk_t::data)` and ` mem_block_length_type_t::GENERAL`
those are not standard c syntax.

I convert them to `JERRY_SIZE_OF_STRUCT_MEMBER (ecma_collection_chunk_t, data)` and `GENERAL`

JERRY_SIZE_OF_STRUCT_MEMBER  is a macro defined in `jrt.h`

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-values-collection.cpp
+++ b/jerry-core/ecma/base/ecma-helpers-values-collection.cpp
@@ -38,7 +38,7 @@ ecma_new_values_collection (const ecma_value_t values_buffer[], /**< ecma-values
 {
   JERRY_ASSERT (values_buffer != NULL || values_number == 0);
 
-  const size_t values_in_chunk = sizeof (ecma_collection_chunk_t::data) / sizeof (ecma_value_t);
+  const size_t values_in_chunk = JERRY_SIZE_OF_STRUCT_MEMBER (ecma_collection_chunk_t, data) / sizeof (ecma_value_t);
 
   ecma_collection_header_t *header_p = ecma_alloc_collection_header ();
 
@@ -86,7 +86,7 @@ ecma_free_values_collection (ecma_collection_header_t *header_p, /**< collection
 {
   JERRY_ASSERT (header_p != NULL);
 
-  const size_t values_in_chunk = sizeof (ecma_collection_chunk_t::data) / sizeof (ecma_value_t);
+  const size_t values_in_chunk = JERRY_SIZE_OF_STRUCT_MEMBER (ecma_collection_chunk_t, data) / sizeof (ecma_value_t);
 
   ecma_collection_chunk_t *chunk_p = ECMA_GET_POINTER (ecma_collection_chunk_t,
                                                        header_p->first_chunk_cp);
@@ -128,7 +128,7 @@ ecma_append_to_values_collection (ecma_collection_header_t *header_p, /**< colle
                                   bool do_ref_if_object) /**< if the value is object value,
                                                               increase reference counter of the object */
 {
-  const size_t values_in_chunk = sizeof (ecma_collection_chunk_t::data) / sizeof (ecma_value_t);
+  const size_t values_in_chunk = JERRY_SIZE_OF_STRUCT_MEMBER (ecma_collection_chunk_t, data) / sizeof (ecma_value_t);
 
   size_t values_number = header_p->unit_number;
   size_t pos_of_new_value_in_chunk = values_number % values_in_chunk;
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/base/ecma-helpers-values-collection.cpp
```diff
@@ -38,7 +38,7 @@ ecma_new_values_collection (const ecma_value_t values_buffer[], /**< ecma-values
 {
   JERRY_ASSERT (values_buffer != NULL || values_number == 0);
 
-  const size_t values_in_chunk = sizeof (ecma_collection_chunk_t::data) / sizeof (ecma_value_t);
+  const size_t values_in_chunk = sizeof (((ecma_collection_chunk_t *)0)->data) / sizeof (ecma_value_t);
```
rtakacs: Please do not use a hack for getting the size. Instead, you should use macros.

```
 lit_utf8_byte_t data[ sizeof (uint64_t) - sizeof (mem_cpointer_t) ];
```

For example:
```
#define ECMA_COLLECTION_DATA_COUNT (sizeof (uint64_t) - sizeof (mem_cpointer_t))
#define ECMA_COLLECTION_DATA_SIZE (sizeof (lit_utf8_byte_t) * ECMA_COLLECTION_DATA_COUNT)

lit_utf8_byte_t data[ ECMA_COLLECTION_DATA_COUNT ];
```
```
const size_t values_in_chunk = ECMA_COLLECTION_DATA_SIZE;
```

As I wrote (https://github.com/Samsung/jerryscript/pull/806#issuecomment-179133239) I'm already eliminating the C++ features. So the current workflow not the best because it results work duplication. The example that I show is my result for this problem.

rtakacs: Could we talk on the jerryscript gitter about this topic?

jiangzidong: @rtakacs Should I close this PR and let you continue C-Convertion work? Of course we can talk on gitter. What about your 13:00 (GMT+1), which is my 20:00(GMT+8) today?

rtakacs: @jiangzidong It is not necessary. You spent time to do this like me, but you were faster to upload. Your solution is also good (the null pointer surprised me, I have never seen similar before). In my comment I just want to symbolize that either work is unnecessary when we do this together (at the same time) without any conversation.

The time you recommended is good for me.

jiangzidong: @rtakacs I don't think it's a hack. it is a well accepted method to get size of c struct member. (see http://stackoverflow.com/questions/3553296/c-sizeof-single-struct-member and http://stackoverflow.com/questions/3864583/sizeof-a-struct-member)

your solution directly calculated the size of `ecma_collection_chunk_t::data`

akiss77: @jiangzidong It's not that nice to repeat the sizeof-null-deref logic all over the code. It'd be better to have it in a macro, as suggested in the linked stackoverflow post, e.g.,
```c
#define member_size(type, member) sizeof(((type *)0)->member)
```
(If we choose to use this approach. There are some comments that it is either an undefined behaviour in old C, or needs C99 compiler at least.)

jiangzidong: @akiss77 @rtakacs I find that there already has the marco in jrt.h
line 191 `#define JERRY_SIZE_OF_STRUCT_MEMBER(struct_name, member_name) sizeof (((struct_name *)NULL)->member_name)`

I will fix my patch soon

jerry-core/mem/mem-heap.cpp
```diff
@@ -349,7 +349,7 @@ mem_heap_init (void)
   for (size_t i = 0; i < MEM_HEAP_CHUNKS_NUM; i++)
   {
 #ifndef JERRY_NDEBUG
-    JERRY_ASSERT (mem_heap_length_types[i] == mem_block_length_type_t::GENERAL);
+    JERRY_ASSERT (mem_heap_length_types[i] == GENERAL);
```
ruben-ayrapetyan: @jiangzidong, please rename `GENERAL` / `ONE_CHUNKED` to `MEM_BLOCK_LENGTH_TYPE_GENERAL` / `MEM_BLOCK_LENGTH_TYPE_ONE_CHUNKED`.

jiangzidong: @ruben-ayrapetyan updated, thanks

jerry-core/mem/mem-heap.cpp
```diff
@@ -94,13 +94,13 @@ void mem_heap_valgrind_freya_mempool_request (void)
  */
 typedef enum : uint8_t
 {
-  GENERAL     = 0, /**< general (may be multi-chunk) block
+  MEM_BLOCK_LENGTH_TYPE_GENERAL     = 0, /**< general (may be multi-chunk) block
                     *
                     *   Note:
```
ruben-ayrapetyan: Please, fix indentation in the comment.

ruben-ayrapetyan:
Looks good to me

jerry-core/mem/mem-heap.cpp
jiangzidong: @ruben-ayrapetyan fixed

zherczeg:
LGTM

LaszloLango:
Landed (9ab24b3241a2b81f05029929610578824bbf756c)

PR #838, opened 2016-02-04
Fix the correct number of command line arguments check.

robertsipka:


sand1k:
I think we need to add a test for this.

robertsipka:
Do you have any suggestions where I could do this kind of test? It might be part of the unit tests or what do you mean exactly?

LaszloLango:
Do we really need this limit? I prefer to remove this condition.

akiss77:
@LaszloLango The fixed-length arrays `file_names` and `exec_snapshot_file_names` rely on `JERRY_MAX_COMMAND_LINE_ARGS`. Actually, these arrays are the reason for the existence of this macro. Since we cannot rely on dynamic memory allocation, there has to be a fixed limit, for now at least.

LaszloLango:
@akiss77, I see. Thanks for the explanation. The patch looks good to me.

LaszloLango:
Landed (2221c00ad97f856dd464233ab6d99678f61cb466)

PR #841, opened 2016-02-04
Eliminate code duplication in memory statistics printing

akiss77:

### DIFF
--- a/jerry-core/mem/mem-allocator.cpp
+++ b/jerry-core/mem/mem-allocator.cpp
@@ -51,29 +51,14 @@ mem_finalize (bool is_show_mem_stats) /**< show heap memory stats
 {
   mem_pools_finalize ();
 
+#ifdef MEM_STATS
   if (is_show_mem_stats)
   {
-    mem_heap_print (false, false, true);
-
-#ifdef MEM_STATS
-    mem_pools_stats_t stats;
-    mem_pools_get_stats (&stats);
-
-    printf ("Pools stats:\n");
-    printf (" Chunk size: %zu\n"
-            "  Pools: %zu\n"
-            "  Allocated chunks: %zu\n"
-            "  Free chunks: %zu\n"
-            "  Peak pools: %zu\n"
-            "  Peak allocated chunks: %zu\n\n",
-            MEM_POOL_CHUNK_SIZE,
-            stats.pools_count,
-            stats.allocated_chunks,
-            stats.free_chunks,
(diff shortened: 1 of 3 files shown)


jerry-core/mem/mem-allocator.cpp
```diff
@@ -53,26 +53,7 @@ mem_finalize (bool is_show_mem_stats) /**< show heap memory stats
...
-            "  Peak pools: %zu\n"
-            "  Peak allocated chunks: %zu\n\n",
-            MEM_POOL_CHUNK_SIZE,
-            stats.pools_count,
-            stats.allocated_chunks,
-            stats.free_chunks,
-            stats.peak_pools_count,
-            stats.peak_allocated_chunks);
-#endif /* MEM_STATS */
+    mem_stats_print ();
```
ruben-ayrapetyan: @akiss77, for me it seems better to put `mem_stats_print` under `#ifdef MEM_STATS`, as without the define the call is no-op.

akiss77: @ruben-ayrapetyan I've been considering that as well. The reason why I proposed the above is that it remained semantically equivalent with the original version, which also left the call to `mem_heap_print` unguarded. Even in `mem_heap_print`, there are quite some code lines which are always compiled, and only a final portion is ifdef'd by `MEM_STATS`. If we choose to ifdef the call here, then maybe we should consider putting much more code under ifdef guards,

ruben-ayrapetyan: @akiss77, I see. Thank you for your explanation.
The part of `mem_heap_print` that is not currently guarded by `MEM_STATS`, is not executed during the call, because the corresponding arguments are set to `false`, and only `dump_stats` argument is set to `true`.
So, maybe, its better to extract part of `mem_heap_print` that outputs statistics to another routine like `mem_heap_print_stats`, guard it with `MEM_STATS`, and call it from `mem_heap_print` and `mem_stats_print`. In the case, `mem_stats_print` could be guarded by `#ifdef` around the function, not inside.
What do you think about this?

akiss77:
@ruben-ayrapetyan I've updated the patch. I guess I got it according to your feedback. The only change I made is that I used the name `mem_heap_stats_print` for the new function as it seemed to align better with the existing names.

zherczeg:
LGTM

LaszloLango:
Landed (196e8196fc67d296dc6d350810ee71a1a2de7dbe)

PR #845, opened 2016-02-07
Refactoring jerry-libc to squeeze out some extra bytes from the code (+fixes)

akiss77:
'cause every byte matters :)

### DIFF
--- a/jerry-libc/jerry-libc-printf.c
+++ b/jerry-libc/jerry-libc-printf.c
@@ -1,4 +1,5 @@
-/* Copyright 2014-2015 Samsung Electronics Co., Ltd.
+/* Copyright 2014-2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -210,20 +211,10 @@ libc_printf_write_d_i (FILE *stream, /**< stream pointer */
   switch (length)
   {
     case LIBC_PRINTF_ARG_LENGTH_TYPE_NONE:
+    case LIBC_PRINTF_ARG_LENGTH_TYPE_HH: /* char is promoted to int */
+    case LIBC_PRINTF_ARG_LENGTH_TYPE_H: /* short int is promoted to int */
     {
       value = (uintmax_t) va_arg (*args_list_p, int);
-      break;
-    }
-
-    case LIBC_PRINTF_ARG_LENGTH_TYPE_HH:
-    {
-      value = (uintmax_t) va_arg (*args_list_p, int); /* char is promoted to int */
-      break;
-    }
-
-    case LIBC_PRINTF_ARG_LENGTH_TYPE_H:
-    {
-      value = (uintmax_t) va_arg (*args_list_p, int); /* short int is promoted to int */
       break;
     }
 
@@ -324,20 +315,10 @@ libc_printf_write_u_o_x_X (FILE *stream, /**< stream pointer */
   switch (length)
   {
     case LIBC_PRINTF_ARG_LENGTH_TYPE_NONE:
+    case LIBC_PRINTF_ARG_LENGTH_TYPE_HH: /* char is promoted to int */
+    case LIBC_PRINTF_ARG_LENGTH_TYPE_H: /* short int is promoted to int */
     {
       value = (uintmax_t) va_arg (*args_list_p, unsigned int);
-      break;
-    }
-
-    case LIBC_PRINTF_ARG_LENGTH_TYPE_HH:
-    {
-      value = (uintmax_t) va_arg (*args_list_p, unsigned int); /* char is promoted to int */
-      break;
-    }
-
-    case LIBC_PRINTF_ARG_LENGTH_TYPE_H:
-    {
-      value = (uintmax_t) va_arg (*args_list_p, unsigned int); /* short int is promoted to int */
       break;
     }
 
@@ -398,7 +379,7 @@ libc_printf_write_u_o_x_X (FILE *stream, /**< stream pointer */
     }
   }
 
-  uint32_t radix = 10;
+  uint32_t radix;
   const char *alphabet;
 
   switch (specifier)
--- a/jerry-libc/jerry-libc.c
+++ b/jerry-libc/jerry-libc.c
@@ -1,4 +1,5 @@
-/* Copyright 2014-2015 Samsung Electronics Co., Ltd.
+/* Copyright 2014-2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


akiss77:
The result and effect of the three patches on code size as applied on top of master and incrementally on each other. There is almost always some bytes or some houndred bytes gain, except for 3 hardly explicable code size increases.

executable | master | mem&str | gain | printf | gain | rnd | gain
-----------|--------|---------|------|--------|------|-----|------
debug.linux-cp/jerry | 684307 | 684180 | 127 | 683853 | 454 | 683945 | 362
debug.linux-cp_minimal/jerry | 470308 | 470181 | 127 | 469854 | 454 | 469882 | 426
debug.linux/jerry | 877392 | 877265 | 127 | 876938 | 454 | 877030 | 362
debug.linux-mem_stats/jerry | 882099 | 881972 | 127 | 882493 | -394 | 882585 | -486
debug.linux-mem_stress_test/jerry | 877412 | 877285 | 127 | 876958 | 454 | 877050 | 362
debug.mcu_stm32f4-cp/jerry | 855572 | 855372 | 200 | 855364 | 208 | 855452 | 120
debug.mcu_stm32f4-cp_minimal/jerry | 504588 | 504436 | 152 | 504328 | 260 | 504408 | 180
release.linux-cp/jerry | 218809 | 218747 | 62 | 218681 | 128 | 218681 | 128
release.linux-cp_minimal/jerry | 135597 | 135519 | 78 | 135477 | 120 | 135477 | 120
release.linux/jerry | 289829 | 289962 | -133 | 289049 | 780 | 289049 | 780
release.linux-mem_stats/jerry | 291124 | 290382 | 742 | 290316 | 808 | 290316 | 808
release.linux-mem_stress_test/jerry | 290240 | 289714 | 526 | 289875 | 365 | 289875 | 365
release.mcu_stm32f3-cp/jerry | 142656 | 142640 | 16 | 142632 | 24 | 142632 | 24
release.mcu_stm32f3-cp_minimal/jerry | 81864 | 81840 | 24 | 81824 | 40 | 81824 | 40
release.mcu_stm32f4-cp/jerry | 142656 | 142640 | 16 | 142632 | 24 | 142632 | 24
release.mcu_stm32f4-cp_minimal/jerry | 81864 | 81840 | 24 | 81824 | 40 | 81824 | 40

jerry-libc/jerry-libc.c
```diff
@@ -225,12 +202,12 @@ strncmp (const char *s1, const char *s2, size_t n)
 char * __attr_used___ // FIXME
```
LaszloLango: Is this FIXME comment valid after your patch?

LaszloLango:
LGTM

jerry-libc/jerry-libc-printf.c
```diff
@@ -210,23 +211,13 @@ libc_printf_write_d_i (FILE *stream, /**< stream pointer */
...
-      value = (uintmax_t)va_arg (*args_list_p, int); /* char is promoted to int */
-      break;
-    }
-
-    case LIBC_PRINTF_ARG_LENGTH_TYPE_H:
-    {
-      value = (uintmax_t)va_arg (*args_list_p, int); /* short int is promoted to int */
-      break;
-    }
-
```
zherczeg: I think the compielr do this automatically.

jerry-libc/jerry-libc.c
```diff
@@ -62,9 +63,9 @@ memset (void *s,  /**< area to set values in */
         size_t n) /**< area size */
 {
   uint8_t *area_p = (uint8_t *) s;
-  for (size_t index = 0; index < n; index++)
+  while (n--)
   {
-    area_p[ index ] = (uint8_t) c;
+    *area_p++ = (uint8_t) c;
   }
```
zherczeg: I think the compiler optimize this as well.

akiss77:
@LaszloLango I have to be honest, I don't know why some of the functions are tagged with the used attribute while others are not. I did not dare to touch it...

The description of the attribute is: "This attribute, attached to a function, means that code must be emitted for the function even if it appears that the function is not referenced."

jerry-libc/jerry-libc.c
```diff
@@ -263,7 +240,8 @@ strlen (const char *s)
 int
 rand (void)
 {
-  uint32_t intermediate = libc_random_gen_state[0] ^ (libc_random_gen_state[0] << 11);
+  uint32_t intermediate = libc_random_gen_state[0];
+  intermediate ^= intermediate << 11;
```
zherczeg: This is harder to me to read. And I think the compiler can optimize this as well.

zherczeg:
As for me I am not conviced it is worth to refactor the code this much. Perhaps in a few complicated cases.

jerry-libc/jerry-libc-printf.c
akiss77: Me too. Still, as the table above shows, doing the merge manually does have an impact on the code size.

Generally, I think it is good practice not to write overly verbose, or suboptimal, or clone-full code and rely on the compiler to optimize it away. Since sometimes it fails. (Up to a certain point of course. Extreme brevity and incomprehensible optimizations can have a critical maintenance cost in the long run, but that's another story. And this is not the case here, IMHO.)

jerry-libc/jerry-libc.c
akiss77: Same as above. Why introduce variables if you can do well without them? This is C, so `*p++` is not that much unheard of.

jerry-libc/jerry-libc.c
akiss77: The body contains two more `^=`s... And as for the compiler optimization, same as above.

akiss77:
@zherczeg We are aiming at a super light libc, aren't we?
(Btw, the pr fixes 2 bugs. And I wouldn't like fixing the one in strncmp by keeping the current style. And if that changes, then the other funcs should be changed as well.)

akiss77:
This PR has been pending for a while. What shall happen to it? And should it be rejected, what should happen to the issues it fixes?

LaszloLango:
@akiss77, I'm still thinking of this change is needed, but @zherczeg had doubts. It depends on him.

jerry-libc/jerry-libc.c
```diff
@@ -284,5 +260,8 @@ rand (void)
 void
 srand (unsigned int seed) /**< new seed */
 {
+  libc_random_gen_state[0] =
+  libc_random_gen_state[1] =
+  libc_random_gen_state[2] =
```
zherczeg: Isn't this increase code size?

zherczeg:
I think in general the patch is ok, but there are a few changes I don't particularly like. But I have no strong objection.

jerry-libc/jerry-libc.c
akiss77: Most probably it is increasing the code size with some bytes. However, I cannot think of a smaller solution for fixing `srand`. Since that function _must_ reinitialize the whole state of the pseudo-RNG to ensure that always the same sequence is generated from the same seed.

akiss77:
@zherczeg, are the changes those you mentioned before, or something else? I could try and come up with "nicer" approaches.

jerry-libc/jerry-libc.c
zherczeg: Ok, keep this.

jerry-libc/jerry-libc.c
zherczeg: I think this is the only thing needs to be changed, and after that LGTM.

PR #862, opened 2016-02-09
Use simple types instead of reference types.

robertsipka:
The C99 standard does not support reference types.


LaszloLango:
LGTM, but commit message should be "Use value types instead of reference types"

robertsipka:
Thanks, I've fixed it.

LaszloLango:
Landed (cc6fced17ad47f0b52dab3e7d483902ec56cb0ad)

PR #863, opened 2016-02-09
Use 'static inline' for function definitions that are in header files.

robertsipka:
It's required by the C99 standard.


zherczeg:
These functions should be moved to C files. They don't need to be in the header.

robertsipka:
@zherczeg : Thanks, I modified based on your comment.

zherczeg:
LGTM

robertsipka:
thanks, rebased to master.

PR #865, opened 2016-02-10
Fix build with -O3

LaszloLango:
Related issue: #657, #853 

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
@@ -1064,6 +1064,7 @@ ecma_builtin_global_object_encode_uri_helper (ecma_value_t uri, /**< uri argumen
   lit_code_point_t cp;
   ecma_char_t ch;
   lit_utf8_byte_t octets[LIT_UTF8_MAX_BYTES_IN_CODE_POINT];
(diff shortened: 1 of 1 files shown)


rtakacs:
Tested with gcc 5.2 and there are no errors.

zherczeg:
I think we need to change the other side to clear the buffer in advance (with an uit32_t zero). The problem is, if normal buffer is passed, you may overwrite the next values in the buffer.

LaszloLango:
@rtakacs, @zherczeg I've updated the PR. Please check.

jerry-core/ecma/builtin-objects/ecma-builtin-global.cpp
```diff
@@ -1064,6 +1064,7 @@ ecma_builtin_global_object_encode_uri_helper (ecma_value_t uri, /**< uri argumen
   lit_code_point_t cp;
   ecma_char_t ch;
   lit_utf8_byte_t octets[LIT_UTF8_MAX_BYTES_IN_CODE_POINT];
+  memset (octets, LIT_BYTE_NULL , LIT_UTF8_MAX_BYTES_IN_CODE_POINT);
```
rtakacs: Plus space after LIT_BYTE_NULL.

rtakacs:
Checked, there are no build errors.

LaszloLango:
@rtakacs, thanks. @zherczeg, please check.

zherczeg:
I like this much better. LGTM

PR #866, opened 2016-02-10
Abort the build-process when the build-configuration is unsupported

bzsolt:
The MCU targets doesn't support Valgrind.
Related issue: #762

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,6 +58,15 @@ export TARGET_NATIVE_SYSTEMS = $(shell uname -s | tr '[:upper:]' '[:lower:]')
    VALGRIND_FREYA := OFF
   endif
 
+ # Indicate when Valgrind or Valgrind_Freya option is enabled.
+  VALGRIND_IS_ENABLED ?= OFF
+
+  ifneq ($(VALGRIND)$(VALGRIND_FREYA), OFFOFF)
+   VALGRIND_IS_ENABLED := ON
+  else
+   VALGRIND_IS_ENABLED := OFF
+  endif
+
  # Static checkers
   STATIC_CHECK ?= OFF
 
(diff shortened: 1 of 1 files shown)


LaszloLango:
@bzsolt, #790 depends on this PR?

bzsolt:
@LaszloLango, this change independent from #790 
That PR needs another modifications.

LaszloLango:
@bzsolt, I see. LGTM

akiss77:
Hm, this might have beed a tad too quick. Would it not have been better to implement this check on CMakeLists.txt side? The more checks we have there the better IMHO, since this Makefile is just a wrapper around the cmake-generated build system. There are some use cases where this Makefile is not called at all. Those cases cannot utilize these sanity checks.

LaszloLango:
I think the Makefile should be an interface for builds and calling directly the CMake is a bit hack. What use cases use the CMakeList.txt directly? I don't mind if you move this to CMakeLists.txt side, but the merged PR is still good for me. Maybe we should rework the whole build system. :)

akiss77:
Our own makefiles at `targets/esp8266/Makefile.esp8266` and `targets/mbedk64f/Makefile.mbedk64f` come to my mind first. More to come with the nuttx target PR soon to be landed.

The build system rework is exactly what I'm working on and this patch will cause trouble rebasing to.

LaszloLango:
@akiss77, I see. Sorry for the inconvenience. BTW, nuttx target PR (#805) is landed. :) I accept your arguments. This change would be better on the CMakeList.txt side. Could you do this in part of your rework?

bzsolt:
@akiss77 @LaszloLango  Sorry for the trouble, if necessary, I could move this to the CMake side too.

PR #870, opened 2016-02-10
Rename docs/Doxygen to docs/Doxyfile

akiss77:
This is a really minuscule change, but it is more idiomatic to name
the config file of doxygen as Doxyfile.


LaszloLango:
If it is a config file then `Doxyfile.config` or `Doxyconfig` would be a better name. What do you think?

akiss77:
I object, your honour! `Doxyfile` is the regular name of a doxygen config file. AFAIK the author took the naming scheme from `Makefile`s.

LaszloLango:
@akiss77, Ok. :) LGTM

PR #872, opened 2016-02-10
Remove cppcheck and vera++ from prerequisites

### LINKED ISSUE #844: Reduce prerequisite dependencies?
In our development docs we state that Ubuntu is the supported dev platform, and there are already packages that must be installed via the distro's package manager, i.e., with apt-get. Still, vera++ and cppcheck tools are downloaded and installed with tools/prerequisites.sh into a local subdir, although they are present in Ubuntu's package repositories. Is there any reason not to list them simply in the dev docs as dependencies and remove their setup logic from prerequisites.sh?

(Happy to propose a PR if we reach consensus.)

akiss77:
Rely on platform-provided versions. Thus, no need to download and
build them, neither to wrap them with shell scripts. CMake and
precommit updated to call the new tools.

PS: On my Ubuntu 14.04.3, cppcheck has version 1.61, while prereq
version was 1.69. The older version reports and fails on a strange
style issue in ecma/builtin-objects/ecma-builtin-helpers.cpp, for
which the only solution found was to suppress cppcheck errors with
`variableScope` id.

resolves #844


tools/cppcheck/suppressions-list
```diff
@@ -1,3 +1,4 @@
 operatorEqVarError
 noConstructor
 duplicateExpression
+variableScope
```
LaszloLango: Are you sure suppressing all of the variable scope expressions is a good idea?

LaszloLango:
Please update the copyright in the touched files.

LaszloLango:
Please update `docs/DEVELOPMENT.md` too in this PR.

tools/cppcheck/suppressions-list
akiss77: That's the ID cppcheck reported for the error when running with --xml. I couldn't find any detailed description of the error, nor how many checks this suppression would turn off.

In function `ecma_builtin_helper_object_to_string` the following is reported:
```
[jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp:87]: (style) The scope of the variable 'buffer_size' can be reduced.
```
But I couldn't find any way to get rid of the error (moved the declaration earlier, moved a var declaration between the declaration of `buffer_size` az the local array allocation, removed the const specifier, etc.). Anyway, they would only have been hacks. Working around false positives of a static checker cannot be the motivation for refactoring the code so I dropped that idea and suppressed the error ID.

Not sure about alternatives.

akiss77:
Updated according to the review, and rebased to current master.

tools/cppcheck/suppressions-list
LaszloLango: I checked and you are right. This is definitely a false positive. I actually like the variable scope check, so I'd like to keep it. If there is only one false positive then we can suppress only that with a `//cppcheck-suppress variableScope` comment in the previous line (jerry-core/ecma/builtin-objects/ecma-builtin-helpers.cpp:86). Could you try it?

akiss77:
Restricted the error suppression to the file with the false positive. Updated & rebased.

LaszloLango:
LGTM

Issue #691, opened 2015-10-29
Date built-in (test262 failures)


ruben-ayrapetyan:
Hello.

The following test262 tests related to `Date` built-in are failing on current master (906558b8e94aa9118971188af9880eb47ff9b791):
- ch15/15.9/15.9.3/S15.9.3.1_A5_T1
- ch15/15.9/15.9.3/S15.9.3.1_A5_T2
- ch15/15.9/15.9.3/S15.9.3.1_A5_T3
- ch15/15.9/15.9.3/S15.9.3.1_A5_T4
- ch15/15.9/15.9.3/S15.9.3.1_A5_T5
- ch15/15.9/15.9.3/S15.9.3.1_A5_T6

galpeter:
This could be a related to https://github.com/Samsung/jerryscript/issues/213

dbatyai:
The problem is that the tests were created for a specific timezone (namely US/Pacific) and fail when they run in anything other. Also, as far as I know there's no functionality in jerry-libc for getting system timezone, so in our case these always run with GMT time.

ruben-ayrapetyan:
@dbatyai, thank you for checking.

Issue #213, opened 2015-06-19
Missing standard time functions in jerry-libc


szledan:
For Date object implementations we need few time functions and variables, example: `localtime_r`, `gmtime_r`, `daylight` for [1] functions.

These are system dependent features.
Could we somehow use the standard `<time.h>`? 

[1] http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.7

rtakacs:
The following test262 conformance tests fail because we don't support TZA:
```
15.9.3/S15.9.3.1_A5_T1.js
15.9.3/S15.9.3.1_A5_T2.js
15.9.3/S15.9.3.1_A5_T3.js
15.9.3/S15.9.3.1_A5_T4.js
15.9.3/S15.9.3.1_A5_T5.js
15.9.3/S15.9.3.1_A5_T6.js
```

ruben-ayrapetyan:
@rtakacs, could you, please, provide more details on how should Date built-in's implementation be changed to fix the tests?

PR #873, opened 2016-02-11
Fix system call related date builtin functions

LaszloLango:
Related issues: #213, #691
 * Fixed 'ecma_date_local_tza' and 'ecma_date_daylight_saving_ta' date builtin helper functions
 * Added syscall of gettimeofday function to get the current system time and timezone.
 * Fixed related regression test files.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
@@ -26,6 +26,8 @@
 
 #ifndef CONFIG_ECMA_COMPACT_PROFILE_DISABLE_DATE_BUILTIN
 
+#include <time.h>
+
 /** \addtogroup ecma ECMA
  * @{
  *
@@ -445,13 +447,14 @@ ecma_date_week_day (ecma_number_t time) /**< time value */
 ecma_number_t __attr_always_inline___
 ecma_date_local_tza ()
 {
-  /*
-   * FIXME:
-   *        Get the real system time. ex: localtime_r, gmtime_r, daylight on Linux
-   *        Introduce system macros at first.
-   */
-  TODO ("Implement time functions in jerry-libc.");
-  return ECMA_NUMBER_ZERO;
+  struct timezone tz;
+
+  if (gettimeofday (NULL, &tz) != 0)
+  {
+    return ecma_raise_type_error ("gettimeofday failed");
+  }
+
+  return tz.tz_minuteswest * -ECMA_DATE_MS_PER_MINUTE;
 } /* ecma_date_local_tza */
 
 /**
@@ -470,13 +473,14 @@ ecma_date_daylight_saving_ta (ecma_number_t time) /**< time value */
     return time; /* time is NaN */
   }
 
-  /*
-   * FIXME:
-   *        Get the real system time. ex: localtime_r, gmtime_r, daylight on Linux
-   *        Introduce system macros at first.
-   */
-  TODO ("Implement time functions in jerry-libc.");
-  return ECMA_NUMBER_ZERO;
+  struct timezone tz;
+
+  if (gettimeofday (NULL, &tz) != 0)
+  {
+    return ecma_raise_type_error ("gettimeofday failed");
+  }
+
+  return tz.tz_dsttime;
 } /* ecma_date_daylight_saving_ta */
 
 /**
@@ -1062,7 +1066,7 @@ ecma_date_value_to_string (ecma_number_t datetime_number) /**< datetime */
   /*
    * Character length of the result string.
    */
-  const uint32_t result_string_length = 33;
+  const uint32_t result_string_length = 34;
 
   lit_utf8_byte_t character_buffer[result_string_length];
   lit_utf8_byte_t *dest_p = character_buffer;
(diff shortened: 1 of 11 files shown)


jerry-libc/target/darwin/jerry-libc-target.c
```diff
@@ -384,3 +386,12 @@ fwrite (const void *ptr, /**< data to write */
   return bytes_written / size;
 } /* fwrite */
 
+/**
+ * This function can get the time as well as a timezone.
+ */
```
zherczeg: return value description?

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.cpp
```diff
@@ -445,13 +447,14 @@ ecma_date_week_day (ecma_number_t time) /**< time value */
...
-  TODO ("Implement time functions in jerry-libc.");
-  return ECMA_NUMBER_ZERO;
+  struct timezone tz;
+
+  if (gettimeofday (NULL, &tz) != 0)
+  {
+    return ecma_raise_type_error ("gettimeofday failed");
+  }
+
+  return tz.tz_minuteswest * -60000;
```
zherczeg: 60000? I would define a constant for it and add a description.

LaszloLango: The timezone is in minutes and we have to convert to milliseconds. I'll give a name for it.

LaszloLango:
@zherczeg, I've updated the PR.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
```diff
@@ -24,6 +24,8 @@
 #include "fdlibm-math.h"
 #include "lit-char-helpers.h"
 
+#include <time.h>
+
 #ifndef CONFIG_ECMA_COMPACT_PROFILE_DISABLE_DATE_BUILTIN
```
zherczeg: I think the <time.h> include should be inside the ifndef. It is not needed otherwise.

LaszloLango:
@zherczeg, I've updated the PR.

jerry-libc/include/time.h
```diff
@@ -0,0 +1,46 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRY_LIBC_TIME_H
+#define JERRY_LIBC_TIME_H
+
+#ifdef __cplusplus
+# define EXTERN_C "C"
```
zherczeg: Is not this will be a symbol clash? I think there is a { } based solution for that.

LaszloLango: This worked fine for me. Could you be more specific? BTW, this is used in every jerry-libc header.

akiss77: I agree with @zherczeg here. It would be better to use the more traditional
```c
#ifdef __cplusplus
extern "C" {
#endif
// header main content here
#ifdef __cplusplus
}
#endif
```
approach. The issue is that the `#define EXTERN_C "C"` way is used all over the project. I'd love to see that change.

LaszloLango: Ok, I'll change it everywhere and add it in a separate commit, but part of this PR.

LaszloLango:
@akiss77, @zherczeg, I've updated the PR.

zherczeg:
LGTM

akiss77:
It's much better practice to have the `#ifdef __cplusplus` at the very beginning of the headers, not only where function declarations start. In jerry.h I've even spotted that those guards are used twice. So, I'd really prefer to have the order

* copyright
* `#ifndef HEADERNAME_H` `#define HEADERNAME_H`
* includes
* `#ifdef _cplusplus` `#define extern "C" {`
* all header content (including typedes and all, not only function declarations)
* `#ifdef _cplusplus` `}`
* `#endif // HEADERNAME_H`

Something like that. Opinion?

LaszloLango:
@akiss77, the suggested style is convenient for me. The biggest problem here is the vera++ checker. We have to modify the rule and currently not sure how.

@zherczeg, @akiss77, do you mind if I raise an issue for this and do it in a separate PR?

akiss77:
@LaszloLango OK with me

jerry-core/jerry-port.h
```diff
@@ -19,29 +20,26 @@
...
+  int jerry_port_putchar (int c);
 
 /**
  * @}
  */
 
 #endif /* !JERRY_API_H */
+#ifdef __cplusplus
+} /*  */
+#endif /* !__cplusplus */
```
akiss77: The order seems to be mixed up here. JERRY_PORT_H starts, extern C starts, JERRY_PORT_H ends, extern C ends.
And, lo, wrong closing comment! this is not JERRY_API_H...

jerry-core/jerry.h
```diff
@@ -117,4 +122,7 @@ extern EXTERN_C void jerry_pop_ctx (void);
  * @}
  */
 
+#ifdef __cplusplus
+} /*  */
```
akiss77: Don't we want to add extern "C" to the comment here? (And all other similar places as wel.)

LaszloLango: Currently this empty comments needs for vera++. :(

LaszloLango:
A reverted the last commit, so this PR only contains the code related to the Date builtin, because the change of the EXTERN_C macro is not trivial at the moment. I raised an issue for that: #900 and pushed a new branch to my fork: https://github.com/LaszloLango/jerryscript/tree/refactor-extern-c-define

@zherczeg, @akiss77 please check this PR again. Is this ready for merge?

akiss77:
LGTM (FWIW)

PR #678, opened 2015-10-20
Use heap allocated buffer instead of char collection in ecma_strings

dbatyai:
Remove char collection from ecma_strings, and use a heap buffer instead, where we also store string size and length.

Results (measured on rpi2):

                               Benchmark |         RSS<br>(+ is better) |        Perf<br>(+ is better) |
                               --------- |                          --- |                         ---- |
                              3d-cube.js | 136 -> 136  (0) | 3.728 -> 3.71  (0.4828) | 
                          3d-raytrace.js | 304 -> 264  (13.1579) | 5.918 -> 5.554  (6.1507) | 
                  access-binary-trees.js | 88 -> 92  (-4.545) | 2.72 -> 2.72  (0) |
                      access-fannkuch.js | 52 -> 52  (0) | 10.03 -> 9.94  (0.8973) | 
                         access-nbody.js | 68 -> 68  (0) | 4.65 -> 4.63  (0.4301) | 
             bitops-3bit-bits-in-byte.js | 40 -> 40  (0) | 3.18 -> 3.13  (1.5723) | 
                  bitops-bits-in-byte.js | 40 -> 40  (0) | 4.35 -> 4.36  (-0.23) | 
                   bitops-bitwise-and.js | 36 -> 36  (0) | 4.26 -> 4.26  (0) | 
                controlflow-recursive.js | 224 -> 224  (0) | 3.23 -> 3.17  (1.8576) | 
                           crypto-aes.js | 156 -> 152  (2.5641) | 5.96 -> 5.71  (4.1946) | 
                           crypto-md5.js | 216 -> 212  (1.8519) | 33.57 -> 10.53  (68.6327) | 
                          crypto-sha1.js | 156 -> 152  (2.5641) | 15.1 -> 5.71  (62.1854) | 
                    date-format-xparb.js | 104 -> 96  (7.6923) | 2.03 -> 1.82  (10.3448) | 
                          math-cordic.js | 48 -> 48  (0) | 4.25 -> 4.23  (0.4706) | 
                    math-partial-sums.js | 40 -> 40  (0) | 2.55 -> 2.5  (1.9608) | 
                   math-spectral-norm.js | 52 -> 52  (0) | 3.22 -> 3.18  (1.2422) | 
                        string-base64.js | 192 -> 164  (14.5833) | 235.22 -> 38.6  (83.5898) | 
                         string-fasta.js | 60 -> 60  (0) | 5.54 -> 4.72  (14.8014) | 
[...]

PR #874, opened 2016-02-11
New Allocator and improved String handling.

dbatyai:
This PR is the continuation of the work in #678. It was suggested there to use the recordset implementation to reduce wasted memory. I investigated this approach, but the results were rather dissapointing, as execution became almost two times slower on each test.
So, in order to reduce wasted memory while also gaining the performance benefit of using heap buffers, I experimented with a new allocator that can allocate blocks that are a multiple of 8 bytes.
Here are the results:

                      Benchmark |      RSS<br>(+ is better) |               Perf<br>(+ is better)
                     3d-cube.js |   84k ->   88k :  -4.762% |   2.643s ->   2.657s :  -0.505%  (+-2.810%)
         access-binary-trees.js |   80k ->   80k :  +0.000% |   1.410s ->   1.373s :  +2.601%  (+-2.346%)
             access-fannkuch.js |   36k ->   36k :  +0.000% |   7.973s ->   7.793s :  +2.257%  (+-1.150%)
                access-nbody.js |   40k ->   40k :  +0.000% |   3.080s ->   3.067s :  +0.433%  (+-2.842%)
    bitops-3bit-bits-in-byte.js |   32k ->   28k : +12.500% |   1.727s ->   1.687s :  +2.317%  (+-2.678%)
         bitops-bits-in-byte.js |   32k ->   28k : +12.500% |   2.657s ->   2.457s :  +7.528%  (+-2.618%)
          bitops-bitwise-and.js |   28k ->   24k : +14.286% |   3.680s ->   3.467s :  +5.797%  (+-2.695%)
          bitops-nsieve-bits.js |  152k ->  152k :  +0.000% |  30.107s ->  29.920s :  +0.620%  (+-2.970%)
       controlflow-recursive.js |  176k ->  176k :  +0.000% |   0.890s ->   0.897s :  -0.749%  (+-3.717%)
                  crypto-aes.js |  120k ->  108k : +10.000% |   4.707s ->   3.680s : +21.813%  (+-1.099%)
                  crypto-md5.js |  180k ->  168k :  +6.667% |  21.473s ->   8.527s : +60.292%  (+-0.166%)
                 crypto-sha1.js |  124k ->  124k :  +0.000% |   9.733s ->   4.497s : +53.801%  (+-1.646%)
           date-format-tofte.js |   60k ->   64k :  -6.667% |   2.240s ->   2.273s :  -1.488%  (+-1.477%)
           date-format-xparb.js |   60k ->   56k :  +6.667% |   1.083s ->   0.893s : +17.538%  (+-3.958%)
                 math-cordic.js |   32k ->   36k : -12.500% |   3.000s ->   2.913s :  +2.889%  (+-1.103%)
           math-partial-sums.js |   32k ->   28k : +12.500% |   1.660s ->   1.643s :  +1.004%  (+-1.993%)
          math-spectral-norm.js |   36k ->   36k :  +0.000% |   1.520s ->   1.460s :  +3.947%  (+-0.166%)
               string-base64.js |  164k ->  156k :  +4.878% | 149.297s ->  31.220s : +79.089%  (+-0.718%)
                string-fasta.js |   44k ->   36k : +18.182% |   4.547s ->   3.570s : +21.481%  (+-3.595%)
                Geometric mean: |     RSS reduction: 4.224% |               Speed up: 19.896%  (+-0.632%)

Memory usage increased slightly in most cases by a few hundred bytes, because of the removal of recordset, and the rework of the literal storage, but the total memory available also increased by around the same amount, because we no longer have to store bitmaps for memory regions, so this basically evens out.
On the other hand, strings that were previously stored in char_collections now use around 15% less memory in buffers, while also providing significant performance improvement.
It's worth noting, that we also require less intermediate buffers for manipulating strings.
When using --mem-stats, peak memory usage may seem to increase  by a larger amount, than stated before, this is the result of garbage collection triggering differently with the new allocation scheme and the increased heap size.

It's also worth noting that the new allocator could also support separated heap regions, which I plan to investigate further, since it could be very useful in some cases. Also, the maximum memory size is now less constrained, although we are still limited to 512kB because of compressed pointers.


zherczeg:
Wow, nice improvement!

LaszloLango:
@dbatyai, the following regression test fails: `./tests/jerry/array-prototype-sort.js`

dbatyai:
@LaszloLango, Thanks, the issue is fixed.

LaszloLango:
@dbatyai, could you rebase the patch to  the current master?

dbatyai:
@LaszloLango, done.

jerry-core/config.h
```diff
@@ -42,15 +42,15 @@
...
+# define CONFIG_MEM_HEAP_AREA_SIZE (512 * 1024)
+#elif CONFIG_MEM_HEAP_AREA_SIZE > (512 * 1024)
+# error "Currently, maximum 512 kilobytes heap size is supported"
 #endif /* !CONFIG_MEM_HEAP_AREA_SIZE */
 
 /**
  * Desired limit of heap usage
  */
-#define CONFIG_MEM_HEAP_DESIRED_LIMIT (CONFIG_MEM_HEAP_AREA_SIZE / 32)
+#define CONFIG_MEM_HEAP_DESIRED_LIMIT (JERRY_MIN(CONFIG_MEM_HEAP_AREA_SIZE / 32, 8192))
```
LaszloLango: '8192'?

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -44,6 +44,15 @@
...
                      ECMA_STRING_MAX_CONCATENATION_LENGTH);
 
+/**
+ * String header
+ */
+typedef struct ecma_string_heap_header_t
+{
+  uint16_t size;
+  uint16_t length;
+} ecma_string_heap_header_t;
```
LaszloLango: Common style is
```C++
typedef struct ecma_string_heap_header_t
{
  uint16_t size;
  uint16_t length;
};
```

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -44,6 +44,15 @@
 JERRY_STATIC_ASSERT ((uint32_t) ((int32_t) ECMA_STRING_MAX_CONCATENATION_LENGTH) ==
                      ECMA_STRING_MAX_CONCATENATION_LENGTH);
 
+/**
+ * String header
+ */
+typedef struct ecma_string_heap_header_t
+{
+  uint16_t size;
```
LaszloLango: missing comment

jerry-core/ecma/base/ecma-helpers-string.c
akiss77: @LaszloLango That does not seem valid C to me. `typedef struct { ... } xxx_t;` could work however.

jerry-core/lit/lit-literal-storage.h
```diff
@@ -17,16 +17,71 @@
...
+  LIT_RECORD_TYPE_MAGIC_STR = 2, /**< Magic string record that holds a magic string id. */
+  LIT_RECORD_TYPE_MAGIC_STR_EX = 3, /**< External magic string record that holds an extrernal magic string id. */
+  LIT_RECORD_TYPE_NUMBER = 4 /**< Number record that holds a numeric value. */
+} lit_record_type_t;
+
+typedef struct lit_record_t
+{
+  uint16_t next;
+  uint8_t type;
+} lit_record_t;
```
LaszloLango: ditto

jerry-core/lit/lit-literal-storage.h
```diff
@@ -17,16 +17,71 @@
...
+
+typedef struct lit_charset_record_t
+{
+  uint16_t next;
+  uint8_t type;
+  uint8_t hash;
+  uint16_t size;
+  uint16_t length;
+  char data;
+} lit_charset_record_t;
```
LaszloLango: ditto

jerry-core/lit/lit-literal-storage.h
```diff
@@ -17,16 +17,71 @@
...
+
+typedef struct lit_number_record_t
+{
+  uint16_t next;
+  uint8_t type;
+  ecma_number_t number;
+} lit_number_record_t;
```
LaszloLango: ditto

jerry-core/lit/lit-literal-storage.h
```diff
@@ -17,16 +17,71 @@
...
+
+typedef struct lit_magic_record_t
+{
+  uint16_t next;
+  uint8_t type;
+  uint32_t magic_id;
+} lit_magic_record_t;
```
LaszloLango: ditto

jerry-core/lit/lit-literal.c
```diff
@@ -365,6 +374,30 @@ lit_literal_equal_num (lit_literal_t lit,     /**< literal to check */
...
+
+/**
+ * Check if literal contains the string equal to the buffer
+ *
+ * @return true if equal
+ *         false otherwise
+ */
+
+bool
+lit_literal_equal_charset (lit_literal_t lit,
```
LaszloLango: missing comments

jerry-core/mem/mem-heap.c
```diff
@@ -90,99 +90,53 @@ void mem_heap_valgrind_freya_mempool_request (void)
...
-
-/**
- * Size of heap data area
- */
-#define MEM_HEAP_AREA_SIZE (MEM_HEAP_CHUNKS_NUM * MEM_HEAP_CHUNK_SIZE)
+typedef struct mem_heap_free_t
+{
+  uint32_t next_offset; /* Offset of next region in list */
+  uint32_t size; /* Size of region */
+} mem_heap_free_t;
```
LaszloLango: };

jerry-core/mem/mem-heap.c
```diff
@@ -570,219 +404,139 @@ mem_heap_alloc_block_try_give_memory_back (size_t size_in_bytes, /**< size of re
...
- *
- * @return pointer to allocated memory block - if allocation is successful,
- *         NULL - if requested region size is zero.
+ * Note: block will only be aligned to 4 bytes.
  */
-void *
-mem_heap_alloc_block (size_t size_in_bytes,             /**< size of region to allocate in bytes */
-                      mem_heap_alloc_term_t alloc_term) /**< expected allocation term */
+void * __attr_always_inline___
+mem_heap_alloc_block_store_size (size_t size)
```
LaszloLango: missing comment

jerry-core/mem/mem-heap.c
```diff
@@ -825,13 +579,13 @@ mem_heap_compress_pointer (const void *pointer_p) /**< pointer to compress */
  *
  * @return unpacked heap pointer
  */
-void *
+void*  __attr_pure___ __attribute__((hot))
```
LaszloLango: missing space: `void *`

jerry-core/mem/mem-heap.c
```diff
@@ -1065,58 +705,45 @@ mem_heap_stat_alloc (size_t first_chunk_index, /**< first chunk of the allocated
...
+  mem_heap_stats.allocated_bytes -= aligned_size;
   mem_heap_stats.waste_bytes -= waste_bytes;
 } /* mem_heap_stat_free */
 
-/**
- * Print heap statistics
- */
-void
-mem_heap_stats_print (void)
+static void
```
LaszloLango: missing comment

jerry-core/mem/mem-poolman.c
```diff
@@ -34,116 +34,45 @@
...
-
-/**
- * The condition is assumed when using pointer arithmetics on (mem_pool_chunk_t *) pointer type
- */
-JERRY_STATIC_ASSERT (sizeof (mem_pool_chunk_t) == MEM_POOL_CHUNK_SIZE);
+  struct mem_pools_chunk_t *next_p;
+#ifndef MEM_HEAP_PTR_64
+  uint32_t dummy;
+#endif
+} mem_pools_chunk_t;
```
LaszloLango: };

tests/unit/test-poolman.c
```diff
@@ -47,7 +47,7 @@ main (int __attr_unused___ argc,
 
     for (size_t j = 0; j < subiters; j++)
     {
-      ptrs[j] = mem_pools_alloc ();
+      ptrs[j] = (uint8_t *) mem_pools_alloc ();
       // JERRY_ASSERT (ptrs[j] != NULL);
```
LaszloLango: Please also remove this unused assert.

LaszloLango:
@dbatyai, thanks for the update. Apart from several minor style issues the patch look good to me. :)

dbatyai:
@LaszloLango, I fixed the issues you mentioned.

LaszloLango:
@dbatyai, great. Thank you! @zherczeg, could you also review this patch?

LaszloLango:
@dbatyai, could you check the following test262 test fails:

 *  ch12/12.6/12.6.4/S12.6.4_A4.1 in non-strict mode
 *  ch12/12.6/12.6.4/S12.6.4_A4 in non-strict mode
 *  ch15/15.1/15.1.3/15.1.3.1/S15.1.3.1_A2.5_T1 in non-strict mode
 *  ch15/15.1/15.1.3/15.1.3.2/S15.1.3.2_A2.5_T1 in non-strict mode
 *  ch15/15.5/15.5.5/S15.5.5.1_A2 in non-strict mode

dbatyai:
@LaszloLango, i checked the failures. It appears they were caused by an over-indexed array in `ecma_op_object_get_property_names` which was unnoticed until now. I fixed it and updated the patch.

LaszloLango:
@dbatyai, thanks for the fix. I did a double-check on your patch and the test262 tests run fine in debug Jerry, so If @zherczeg has no objection then we can land this patch soon.

jerry-core/config.h
```diff
@@ -42,15 +42,20 @@
...
-# error "Currently, maximum 256 kilobytes heap size is supported"
+# define CONFIG_MEM_HEAP_AREA_SIZE (512 * 1024)
+#elif CONFIG_MEM_HEAP_AREA_SIZE > (512 * 1024)
+# error "Currently, maximum 512 kilobytes heap size is supported"
 #endif /* !CONFIG_MEM_HEAP_AREA_SIZE */
 
 /**
+ * Max heap usage limit
+ */
+#define CONFIG_MEM_HEAP_MAX_LIMIT 8192
```
zherczeg: What is the purpose of this macro?

jerry-core/config.h
```diff
@@ -80,7 +85,7 @@
  * Also the option affects size of ECMA Object Model's data types.
  * In any case size of any of the types should not exceed CONFIG_MEM_POOL_CHUNK_SIZE.
  */
-#define CONFIG_ECMA_REFERENCE_COUNTER_WIDTH (10)
+#define CONFIG_ECMA_REFERENCE_COUNTER_WIDTH (9)
```
zherczeg: I wouldn't decrease this, since even 1024 is not too big. Could we keep the original somehow?

jerry-core/lit/lit-cpointer.h
```diff
@@ -0,0 +1,38 @@
...
+ *
+ * Note:
+ *      the pointer can represent addresses aligned by lit_DYN_STORAGE_LENGTH_UNIT,
+ *      while mem_cpointer_t can only represent addresses aligned by MEM_ALIGNMENT.
+ */
+typedef uint16_t lit_cpointer_t;
+
+extern lit_cpointer_t lit_cpointer_compress (lit_record_t *);
+extern lit_record_t *lit_cpointer_decompress (lit_cpointer_t);
+extern lit_cpointer_t lit_cpointer_null_cp ();
```
zherczeg: Do we need these in the future?

zherczeg: Hm, lets keep them for now, but I think we should get rid of them eventually.

zherczeg:
Very nice patch

jerry-core/config.h
dbatyai: It defines a maximum value for the heap limit. Heap limit is, as the name suggest, a limit of memory usage, and whenever the size of allocated memory since the previous limit would hit the next one, we would run garbage collection. The limit was previously set to `HEAP_AREA_SIZE / 32 == 8192`. Because I increased the size of the heap, garbage collection became less frequent, so I added a maximum value. This way it wouldn't affect environments where the heap size is different.

jerry-core/config.h
dbatyai: With the increased heap size, we now need to use full 16 bit compressed pointers, while previously 15 bits were enough to address the entire heap. The current object model requires it to fit into 64 bits, unfortunately this meant that I needed to cut those bits from somewhere else. We could keep it the way it was, but we then have to keep the heap at 256kB as well.

zherczeg: Hm, I would keep the 256K for now.

zherczeg:
LGTM after the ref count is restored to its old value.

jerry-core/config.h
dbatyai: Okay, in that case I can actually increase ref count bits to 12.

dbatyai:
Patch is updated.

Issue #118, opened 2015-05-28
 Assertion 'new_bit_field_value < (1ull << width)' failed in jrt_set_bit_field_value


renatahodovan:
###### Jerry version:
    Checked revision: aaeec7f
    Build: debug.linux
###### OS: 
    Ubuntu 14.04, x86_64
###### Test case:
```javascript
function f_1() {
    f_1();
}
f_1();
```
###### Backtrace:
```javascript
ICE: Assertion 'new_bit_field_value < (1ull << width)' failed at jerryscript/jerry-core/jrt/jrt-bit-fields.cpp(jrt_set_bit_field_value):55.
Error: ERR_FAILED_INTERNAL_ASSERTION
(gdb) bt
# 0  0x0000000000455cf1 in jrt_set_bit_field_value (container=4611692615966851101, new_bit_field_value=1024, lsb=16, width=10)
    at jerryscript/jerry-core/jrt/jrt-bit-fields.cpp:55
# 1  0x0000000000485117 in ecma_gc_set_object_refs(ecma_object_t*, unsigned int) [clone .24708] (object_p=0x7e9758 <_ZL13mem_heap_area.4015+24>, 
    refs=1024) at jerryscript/jerry-core/ecma/base/ecma-gc.cpp:98
# 2  0x000000000047947c in ecma_ref_object (object_p=0x7e9758 <_ZL13mem_heap_area.4015+24>)
    at jerryscript/jerry-core/ecma/base/ecma-gc.cpp:193
# 3  0x00000000004a69ec in ecma_builtin_get (builtin_id=ECMA_BUILTIN_ID_GLOBAL)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.cpp:80
# 4  0x00000000004682c8 in ecma_op_function_call (func_obj_p=0x7e97b8 <_ZL13mem_heap_area.4015+120>, this_arg_value=4, arguments_list_p=0x0, 
    arguments_list_len=0) at jerryscript/jerry-core/ecma/operations/ecma-function-object.cpp:559
# 5  0x000000000045ad16 in opfunc_call_n (opdata=..., int_data=0x7ffffff12450) at jerryscript/jerry-core/vm/opcodes.cpp:694
# 6  0x00000000004bfa81 in run_int_loop (int_data=0x7ffffff12450) at jerryscript/jerry-core/vm/vm.cpp:440
[...]

PR #875, opened 2016-02-11
Check reference count of objects

LaszloLango:
Throw an error if the object reference count reached the limit.
Related issue: #118


zherczeg:
Byte code should also throw the same error, please update the throw there as well.

LaszloLango:
@zherczeg, I've updated th PR.

zherczeg:
LGTM

PR #877, opened 2016-02-12
Improve 'tools/run-perf-test.sh'

LaszloLango:
Always write results on stdout and save markdown file optionally.
Use "Peak allocated" if binaries were built with MEM_STATS.


LaszloLango:
@zherczeg, @akiss77, could you review this?

tools/mem-stats-measure.sh
```diff
@@ -0,0 +1,23 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+JERRY=$1
+TEST=$2
+
```
akiss77: Don't we need here something like this below?

```bash
# Check if the specified build supports memory statistics options
function is_mem_stats_build
{
  [ -x "$1" ] || fail_msg "Engine '$1' is not executable"

  tmpfile=`mktemp`
  "$1" --mem-stats $tmpfile 2>&1 | grep -- "Ignoring memory statistics option because of '!MEM_STATS' build configuration." 2>&1 > /dev/null
  code=$?
  rm $tmpfile

  return $code
}

is_mem_stats_build "$JERRY" && fail_msg "Engine specified should be built with memory statistics support"
```
Yes, it smells like copy-pasting...

akiss77:
Other than the above, it's OK with me.

tools/mem-stats-measure.sh
LaszloLango: I put this check in `run-perf.sh`, so it is not necessary to check that here too until this script is only used from there. We can place a double check here, but in that case the `is_mem_stats_build` function will be in three different files (`run-perf.sh, mem-stats.sh, mem-stats-measure.sh`).

akiss77: Ahem, if this script is executed only from `run-perf.sh` then why do we need it separated? Why not just write this single line there?

LaszloLango: @akiss77, because of the current implementation of 'run-compare()' helper function.

akiss77: Fine by me then.

zherczeg:
LGTM

PR #885, opened 2016-02-15
Warning fixes.

robertsipka:
ISO C99 doesn’t support unnamed structs/unions.
Comparison of distinct pointer types lacks a cast.
Dereferencing type-punned pointer will break strict-aliasing rules.

There is still a warning with gcc-4.8.4. (it's supported with gcc 4.9.2.):
type of bit-field ‘ext’ is a GCC extension
In C99, gcc issues a warning with -pedantic but it is permitted to have an implementation defined type for the bit-field.


jerry-core/vm/vm-stack.c
```diff
@@ -75,7 +75,8 @@ vm_stack_context_abort (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
         ecma_collection_chunk_t *chunk_p = MEM_CP_GET_NON_NULL_POINTER (ecma_collection_chunk_t,
                                                                         current);
 
-        ecma_free_value (*(ecma_value_t *) chunk_p->data, true);
+        lit_utf8_byte_t *data_ptr = chunk_p->data;
```
akiss77: Is the introduction of this temporary variable really needed? Cannot the expression be kept in one line as before? (2 more similar constructs below.)

robertsipka:
In this case accessing data at 'chunk_p->data' through a pointer to a different type leads the following error: 'Dereferencing type-punned pointer will break strict-aliasing rules'.
The crucial difference is that I store the real data in a variable of the correct type, and only after having allocated that variable do I treat the variable as an ecma_value_t  (which is allowed).
Maybe it's not the best solution, but it is an exception to 'type punning'.

LaszloLango:
@robertsipka, please add `-Werror=pedantic` to the build to avoid these warnings in the future.

akiss77:
@robertsipka does chaining type casts not help? Something like `* (ecma_value_t *) ((lit_utf8_byte_t *) chunk_p->data)`?

robertsipka:
Sadly, no. This solution would lead to the same error.

akiss77:
@robertsipka Too bad. But thanks for trying. Ok with me then.

robertsipka:
@LaszloLango: Yes, it would be correct way. The remained problem is with the error 'type of bit-field ‘base_cp’ is a GCC extension', as I mentioned in the first comment. 
My possible solutions:
1, Use __extension__ before every 'defined' type of bit-field ‘ext’. 
2, Use #pragma GCC diagnostic ignored "-pedantic" before it, which is ugly, I think.

robertsipka:
@LaszloLango: I've modified the patch, and rebased with master.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -294,8 +294,8 @@ typedef enum
  */
 typedef struct
 {
-  mem_cpointer_t getter_p : ECMA_POINTER_FIELD_WIDTH; /**< pointer to getter object */
-  mem_cpointer_t setter_p : ECMA_POINTER_FIELD_WIDTH; /**< pointer to setter object */
+  __extension__ mem_cpointer_t getter_p : ECMA_POINTER_FIELD_WIDTH; /**< pointer to getter object */
+  __extension__ mem_cpointer_t setter_p : ECMA_POINTER_FIELD_WIDTH; /**< pointer to setter object */
```
zherczeg: Is this working on other compilers not just GCC? On the long run perhaps an unsigned int will also do.

jerry-core/jerry-api.h
```diff
@@ -118,12 +118,10 @@ typedef struct jerry_api_value_t
 
     uint32_t v_uint32; /**< number converted 32-bit unsigned integer */
 
-    union
-    {
-      jerry_api_string_t *v_string; /**< pointer to a JS string */
-      jerry_api_object_t *v_object; /**< pointer to a JS object */
-    };
```
zherczeg: Uh, union in union :) Nice catch.

jerry-core/vm/vm-stack.c
zherczeg: Good point.

zherczeg: I think this is an alignment problem, perhaps a double casting could help:

    ecma_free_value (*(ecma_value_t *) (void*) chunk_p->data, true);

jerry-core/ecma/base/ecma-globals.h
akiss77: The following _could_ work somewhere in jrt.h where the attributes are defined.
```c
#ifndef __GNUC__
#define __extension__
#endif
```
But do note that right now we have no compiler specific ifdefs in the core code base.

LaszloLango:
@robertsipka, thanks. Good to me, but let's wait the review of @zherczeg

jerry-core/vm/vm-stack.c
robertsipka: The chaining type casts doesn't help. The following warning: 'dereferencing type-punned pointer will break strict-aliasing rules [-Wstrict-aliasing]' is still persist in this cases.

jerry-core/ecma/base/ecma-globals.h
robertsipka: It looks good to me, thanks, I'll define this __extension__.

robertsipka:
Thanks for the comments, I've updated the patch.

PR #886, opened 2016-02-16
Add support for travis-ci integration

akiss77:
Fully fledged checks, builds, and tests on Linux, non-voting native builds and unit tests on OS X.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,29 @@
+language: c
+
+os: linux
+dist: trusty
+sudo: required
+
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tools/apt-get-install-deps.sh; fi
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then tools/brew-install-deps.sh; fi
+
+install: make prerequisites
+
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -0,0 +1,32 @@
...
+  include:
+    - os: osx
+      env: TARGET="build.darwin test-js-precommit"
+    - os: osx
+      env: TARGET=test-unit
+  allow_failures:
+    - os: osx
+
+#branches:
+#  only: master
```
LaszloLango: If we don't need this, then remove it please (or uncomment if we do).

LaszloLango:
LGTM

.travis.yml
akiss77: That's something to be decided. That's why I've left it there, to show that we have an option. But I can delete those lines ofc.

akiss77:
Updated (removed the comments)

akiss77:
Ready for landing. Does it need a squash before the merge?

LaszloLango:
@akiss77, sure, you may land this. And yes, squash it please. Our policy is to squash all the commits into one and one commit will belongs to one PR on the master. There are a few exceptions when the patch is too huge. We can change this policy, but not now :) You could raise an issue if you think this should be changed.

PR #889, opened 2016-02-16
Eliminating doxygen warnings by fixing the documentation

akiss77:
I've worked my way through the warnings emitted by doxygen, fixing one group of related issues at a time. The documentation is still far from being perfect -- or, from being consistent -- but now the comments are syntactically correct from doxygen's point of view, at least.


LaszloLango:
LGTM, please squash before merge

PR #891, opened 2016-02-17
Get the alpha order of lexer keywords right

akiss77:
Two keywords violate the ordering. Should someone depend on the
alpha order in the future (because the `keyword_length_*` arrays
seem to be ordered) then this issue will be a nightmare to debug.
Better get this right now.


zherczeg:
LGTM.

The list is scanned lineraly, so ordering is not relevant, but we might use a binary search later, so this is a good catch.

PR #896, opened 2016-02-17
Fix style issues and improve vera++ rules.

LaszloLango:

### DIFF
--- a/jerry-core/ecma/base/ecma-alloc.c
+++ b/jerry-core/ecma/base/ecma-alloc.c
@@ -69,7 +69,7 @@ JERRY_STATIC_ASSERT (sizeof (ecma_getter_setter_pointers_t) <= sizeof (uint64_t)
 #define DEALLOC(ecma_type) void \
   ecma_dealloc_ ## ecma_type (ecma_ ## ecma_type ## _t *p ## ecma_type) \
 { \
-  mem_pools_free ((uint8_t*) p ## ecma_type); \
+  mem_pools_free ((uint8_t *) p ## ecma_type); \
 }
 
 /**
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -117,7 +117,7 @@ ecma_gc_set_object_refs (ecma_object_t *object_p, /**< object */
 /**
  * Get next object in list of objects with same generation.
  */
-static ecma_object_t*
+static ecma_object_t *
 ecma_gc_get_object_next (ecma_object_t *object_p) /**< object */
 {
   JERRY_ASSERT (object_p != NULL);
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -592,54 +592,54 @@ typedef double ecma_number_t;
  *
  * See also: ECMA_262 v5, 15.7.3.2
  */
-# define ECMA_NUMBER_MAX_VALUE ((ecma_number_t)1.7976931348623157e+308)
+# define ECMA_NUMBER_MAX_VALUE ((ecma_number_t) 1.7976931348623157e+308)
 /**
  * Number.MIN_VALUE
  *
  * See also: ECMA_262 v5, 15.7.3.3
  */
-# define ECMA_NUMBER_MIN_VALUE ((ecma_number_t)5e-324)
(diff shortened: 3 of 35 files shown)


tools/vera++/scripts/rules/jerry_dereference_operator_always_on_right.tcl
```diff
@@ -18,7 +18,9 @@
 foreach f [getSourceFileNames] {
     set lineNumber 1
     foreach line [getAllLines $f] {
-        if {[regexp {\w\*\s\w+} $line]} {
+        if {[regexp {\w\*\s\w+} $line]
+             || [regexp {\w\*\)} $line]
+             || [regexp {\w\*$} $line]} {
```
akiss77: I got badly misled by the name of the rule. The first/old pattern matches something like `x* abc` (or, more realistically, the tail of `void* p`). The new ones match `void*)` and `void*`-at-end-of-line. If I'm not mistaken, none of these are pointer dereferences but pointer declarations.

As for the original pattern: why does `\s` have no `+` multiplier? And why does the second `\w` have one? Would changing these cause too many false positives?

If we are already touching this file, it would be good to get it right.

tools/vera++/scripts/rules/jerry_funcname_space_parentheses.tcl
```diff
@@ -29,6 +29,10 @@ proc check_part_of_the_file {file line_num col_start col_end} {
     if {[regexp {([[:alnum:]][\s]{2,}\()|([[:alnum:]]\()} $line]} {
         report $file $line_num "there should be exactly one space before left parentheses"
     }
+
+    if {[regexp {\)\w} $line]} {
+        report $file $line_num "there should be exactly one space after right parentheses"
+    }
```
akiss77: This addition seems to be an abuse of this rule, which was created -- I guess -- to make sure that function declarations and call sites always have a space between the function name and the parameter list. But as the changes in the other files reveal, we'd like to use it for type cast styling. That does not look right to me.

tools/vera++/scripts/rules/jerry_dereference_operator_always_on_right.tcl
LaszloLango: The first check is for `type* var_name`
The second is for type casting for a pointer type `(type*)`
The third one is for return values of functions. They are usually in a separate line.
The plus for `\w` is not necessary, we can remove if it bothers you. `\s+` may be a good idea.
About the name, my opinion is that the `*` symbol is the pointer dereference operator, but the expressions are really pointer declerations. I'd like to refer the `*` in the name of the rule not the expressions, but we can change the name of rule if it is misleading. Please suggest a proper rule name and message.

tools/vera++/scripts/rules/jerry_funcname_space_parentheses.tcl
LaszloLango: Should I move this to a new file?

tools/vera++/scripts/rules/jerry_dereference_operator_always_on_right.tcl
akiss77: I've made a terrible mistake. I've taken a look at the C99 standard :). Now I know that `*` is a punctuator, which is known as an operator, if (and only if) used in a context where it specifies an operation (6.4.6 p2). However, the first and the third checks are for declarators (6.7.5), while the second check is for the type-name part of a cast operator (6.5.4), which leads back again to declarations (6.7.6 p2).

Could jerry_pointer_declarator_space.tcl work as a filename, and "there should be a space between the referenced type and the pointer declarator" as a message work?

tools/vera++/scripts/rules/jerry_funcname_space_parentheses.tcl
akiss77: I'd suggest that.

tools/vera++/scripts/rules/jerry_dereference_operator_always_on_right.tcl
LaszloLango: @akiss77, it convenient for me. I'll update the file name and the message.

LaszloLango:
@akiss77, I've updated the PR.

akiss77:
LGTM (FWIW)

zherczeg:
LGTM

Issue #900, opened 2016-02-18
Remove the EXTERN_C marco


LaszloLango:
Related discussion: https://github.com/Samsung/jerryscript/pull/873#issuecomment-185697225

Use the block style instead of define EXTERN_C marco. It's much better practice to have the #ifdef __cplusplus at the very beginning of the headers. Prefer to have the order

```C++

/**
  * Copyright ...
  */

#ifndef HEADERNAME_H
#define HEADERNAME_H

#include ...

#ifdef _cplusplus
extern "C" {
#endif /* !__cplusplus */
all header content (including typedes and all, not only function declarations)
#ifdef _cplusplus
}
#endif /* !__cplusplus */
#endif // HEADERNAME_H
```

PR #901, opened 2016-02-19
Remove EXTERN_C macros and use block based solution

LaszloLango:
Related issue: #900


jerry-core/jerry-api.h
```diff
@@ -271,119 +270,77 @@ jerry_api_create_boolean_value (bool value);
...
+jerry_api_value_t
 jerry_api_create_object_value (jerry_api_object_t *value);
 
 /**
  * Creates a JERRY_API_DATA_TYPE_STRING type jerry_api_value_t from the
  * given jerry_api_string_t *parameter and returns with it.
  */
-extern EXTERN_C jerry_api_value_t
-jerry_api_create_string_value (jerry_api_string_t *value);
+jerry_api_value_t jerry_api_create_string_value (jerry_api_string_t *value);
```
akiss77: Why did the style change here? Above this point, return type was in a separate line, from this on, it's on the same line as the rest of the function declaration. It would be good to follow one style everywhere.

LaszloLango: I agree with that we should use one style, but which one?

jerry-libc/include/time.h
```diff
@@ -18,29 +18,31 @@
...
+   * Timezone structure
+   */
+  struct timezone
+  {
+    int tz_minuteswest;     /**< minutes west of Greenwich */
+    int tz_dsttime;         /**< type of DST correction */
+  };
 
-extern EXTERN_C int gettimeofday (void *tp, void *tzp);
+  int gettimeofday (void *tp, void *tzp);
```
akiss77: Why is everything indented here? All the other headers are non-indented.

jerry-libc/include/string.h
```diff
@@ -19,18 +19,19 @@
...
-extern EXTERN_C int    strncmp (const char *s1, const char *s2, size_t n);
-extern EXTERN_C char*  strncpy (char *dest, const char *src, size_t n);
+int    memcmp (const void *s1, const void *s2, size_t n);
+void*  memcpy (void *dest, const void *src, size_t n);
+void*  memset (void *s, int c, size_t n);
+int    strcmp (const char *s1, const char *s2);
+size_t strlen (const char *s);
+void*  memmove (void *dest, const void *src, size_t n);
+int    strncmp (const char *s1, const char *s2, size_t n);
+char*  strncpy (char *dest, const char *src, size_t n);
```
akiss77: If we are alread touching style issues: I'd drop the alignment spaces here. Nowhere else do we use them..

jerry-libc/include/time.h
LaszloLango: accident, I'll fix

jerry-libc/include/string.h
LaszloLango: sure, good idea

jerry-core/jerry-api.h
akiss77: Well, the public headers mostly use the same-line style, while internal headers mix it, sometimes even inside the same header. However, C sources use the return-type-on-separate-line style everywhere. So, I'm not sure... Happy to hear the opinion of others.

LaszloLango:
I've updated the PR

tools/vera++/scripts/rules/jerry_braces_on_separate_line.tcl
```diff
@@ -101,7 +101,8 @@ foreach file_name [getSourceFileNames] {
                     set expect_newline false
                 }
             } elseif {$tok_type in {leftbrace rightbrace}} {
-                if {[lindex $prev_tok 1] == $line_num} {
+                if {[lindex $prev_tok 1] == $line_num
+                    && ![regexp {^extern "C"} [getLine $file_name $line_num]]} {
```
akiss77: Is this still needed?

LaszloLango: No

akiss77:
If tools/vera++/scripts/rules/jerry_braces_on_separate_line.tcl is fixed, LGTM (FWIW)

tools/vera++/scripts/rules/jerry_braces_on_separate_line.tcl
LaszloLango: fixed

LaszloLango:
@zherczeg, please check

jerry-core/jerry-api.h
zherczeg: It is easier to read when the return type is in the same line, so I prefer that.

zherczeg:
LGTM

PR #903, opened 2016-02-19
Move cppcheck logic from Makefile and CMakeLists.txt to tools/check-cppcheck.sh

akiss77:
The legacy approach executed cppcheck for every build target, which
resulted in a huge number of re-checks of the sources if more than
one targets were built. The main reason behind that was to get the
right macro-guarded code paths analyzed. However, cppcheck can
analyze every configuration of the sources in one go.

(The patch also contains some aesthetic changes around the way
vera++ is called and how errors are reported.)


akiss77:
Runtime of checks on Travis drop from ~350 secs to ~260 secs (25% improvement, or 1.34x speedup)

LaszloLango:
Nice improvement. LGTM

zherczeg:
LGTM

PR #904, opened 2016-02-19
Rearrange fields of ecma_property_t to be naturally aligned.

zherczeg:
Rearrange fields of ecma_property_t to be naturally aligned. Packed attribute and __extension__ keywords are removed. The standard approach reduced the binary size by 2K.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -83,16 +83,6 @@ typedef enum
 } ecma_simple_value_t;
 
 /**
- * Type of ecma-property
- */
-typedef enum
-{
-  ECMA_PROPERTY_NAMEDDATA, /**< named data property */
-  ECMA_PROPERTY_NAMEDACCESSOR, /**< named accessor property */
-  ECMA_PROPERTY_INTERNAL /**< internal property */
-} ecma_property_type_t;
-
-/**
  * Description of an ecma value
  *
  * Bit-field structure: type (2) | error (1) | value (ECMA_POINTER_FIELD_WIDTH)
@@ -213,84 +203,78 @@ typedef enum
 } ecma_property_configurable_value_t;
 
 /**
- * Width of internal property type field's width
+ * Property's flag list.
  */
-#define ECMA_PROPERTY_INTERNAL_PROPERTY_TYPE_WIDTH   (5)
+typedef enum
+{
+  ECMA_PROPERTY_FLAG_NAMEDDATA = 1u << 0, /**< property is named data */
+  ECMA_PROPERTY_FLAG_NAMEDACCESSOR = 1u << 1, /**< property is named accessor */
+  ECMA_PROPERTY_FLAG_INTERNAL = 1u << 2, /**< property is internal property */
+  ECMA_PROPERTY_FLAG_CONFIGURABLE = 1u << 3, /**< property is configurable */
+  ECMA_PROPERTY_FLAG_ENUMERABLE = 1u << 4, /**< property is enumerable */
+  ECMA_PROPERTY_FLAG_WRITABLE = 1u << 5, /**< property is writable */
+  ECMA_PROPERTY_FLAG_LCACHED = 1u << 6, /**< property is lcached */
+} ecma_property_flags_t;
 
 /**
  * Pair of pointers - to property's getter and setter
  */
 typedef struct
 {
-  __extension__ mem_cpointer_t getter_p : ECMA_POINTER_FIELD_WIDTH; /**< pointer to getter object */
-  __extension__ mem_cpointer_t setter_p : ECMA_POINTER_FIELD_WIDTH; /**< pointer to setter object */
+  mem_cpointer_t getter_p; /**< pointer to getter object */
+  mem_cpointer_t setter_p; /**< pointer to setter object */
 } ecma_getter_setter_pointers_t;
 
 /**
  * Description of ecma-property
  */
-typedef struct __attr_packed___ ecma_property_t
+typedef struct ecma_property_t
 {
-  /** Property's type (ecma_property_type_t) */
-  unsigned int type : 2;
-
   /** Compressed pointer to next property */
-  __extension__ mem_cpointer_t next_property_p : ECMA_POINTER_FIELD_WIDTH;
+  mem_cpointer_t next_property_p;
 
-  /** Property's details (depending on Type) */
+  /** Property's flags (ecma_property_flags_t) */
+  uint8_t flags;
+
+  /** Property's header part (depending on Type) */
   union
   {
-    /** Description of named data property */
-    struct __attr_packed___ ecma_named_data_property_t
+    /** Named data property value upper bits */
+    uint8_t named_data_property_value_high;
+    /** Internal property type */
+    uint8_t internal_property_type;
+  } h;
+
+  /** Property's value part (depending on Type) */
+  union
+  {
+    /** Description of named data property (second part) */
+    struct
     {
-      /** Value */
-      __extension__ ecma_value_t value : ECMA_VALUE_SIZE;
-
       /** Compressed pointer to property's name (pointer to String) */
-      __extension__ mem_cpointer_t name_p : ECMA_POINTER_FIELD_WIDTH;
+      mem_cpointer_t name_p;
 
-      /** Flag indicating whether the property is registered in LCache */
-      unsigned int is_lcached : 1;
-
-      /** Attribute 'Writable' (ecma_property_writable_value_t) */
-      unsigned int writable : 1;
-
-      /** Attribute 'Enumerable' (ecma_property_enumerable_value_t) */
-      unsigned int enumerable : 1;
-
-      /** Attribute 'Configurable' (ecma_property_configurable_value_t) */
-      unsigned int configurable : 1;
+      /** Lower 16 bits of value */
+      uint16_t value_low;
     } named_data_property;
 
-    /** Description of named accessor property */
-    struct __attr_packed___ ecma_named_accessor_property_t
+    /** Description of named accessor property (second part) */
+    struct
     {
       /** Compressed pointer to property's name (pointer to String) */
-      __extension__ mem_cpointer_t name_p : ECMA_POINTER_FIELD_WIDTH;
-
-      /** Attribute 'Enumerable' (ecma_property_enumerable_value_t) */
-      unsigned int enumerable : 1;
-
-      /** Attribute 'Configurable' (ecma_property_configurable_value_t) */
-      unsigned int configurable : 1;
-
-      /** Flag indicating whether the property is registered in LCache */
-      unsigned int is_lcached : 1;
+      mem_cpointer_t name_p;
 
       /** Compressed pointer to pair of pointers - to property's getter and setter */
-      __extension__ mem_cpointer_t getter_setter_pair_cp : ECMA_POINTER_FIELD_WIDTH;
+      mem_cpointer_t getter_setter_pair_cp;
     } named_accessor_property;
 
-    /** Description of internal property */
-    struct __attr_packed___ ecma_internal_property_t
+    /** Description of internal property (second part) */
+    struct
     {
-      /** Internal property's type */
-      unsigned int type : ECMA_PROPERTY_INTERNAL_PROPERTY_TYPE_WIDTH;
-
       /** Value (may be a compressed pointer) */
       uint32_t value;
     } internal_property;
-  } u;
+  } v;
 } ecma_property_t;
 
 /**
--- a/jerry-core/ecma/base/ecma-helpers.c
+++ b/jerry-core/ecma/base/ecma-helpers.c
@@ -438,21 +438,16 @@ ecma_create_internal_property (ecma_object_t *object_p, /**< the object */
 
   ecma_property_t *new_property_p = ecma_alloc_property ();
 
-  new_property_p->type = ECMA_PROPERTY_INTERNAL;
+  new_property_p->flags = ECMA_PROPERTY_FLAG_INTERNAL;
 
   ecma_property_t *list_head_p = ecma_get_property_list (object_p);
   ECMA_SET_POINTER (new_property_p->next_property_p, list_head_p);
   ecma_set_property_list (object_p, new_property_p);
 
-  JERRY_STATIC_ASSERT (
-    ECMA_INTERNAL_PROPERTY__COUNT <= (1ull << ECMA_PROPERTY_INTERNAL_PROPERTY_TYPE_WIDTH),
-    ECMA_PROPERTY_INTERNAL_PROPERTY_TYPE_WIDTH_bits_must_be_able_to_represent_ECMA_INTERNAL_PROPERTY__COUNT_values
-  );
-
   JERRY_ASSERT (property_id < ECMA_INTERNAL_PROPERTY__COUNT);
 
-  new_property_p->u.internal_property.type = property_id & ((1ull << ECMA_PROPERTY_INTERNAL_PROPERTY_TYPE_WIDTH) - 1);
-  new_property_p->u.internal_property.value = ECMA_NULL_POINTER;
+  new_property_p->h.internal_property_type = (uint8_t) property_id;
+  new_property_p->v.internal_property.value = ECMA_NULL_POINTER;
 
   return new_property_p;
 } /* ecma_create_internal_property */
@@ -476,12 +471,10 @@ ecma_find_internal_property (ecma_object_t *object_p, /**< object descriptor */
        property_p != NULL;
        property_p = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p))
   {
-    if (property_p->type == ECMA_PROPERTY_INTERNAL)
+    if ((property_p->flags & ECMA_PROPERTY_FLAG_INTERNAL)
+        && property_p->h.internal_property_type == property_id)
     {
-      if (property_p->u.internal_property.type == property_id)
-      {
-        return property_p;
-      }
+      return property_p;
     }
   }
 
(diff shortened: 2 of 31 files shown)


zherczeg:
Benchmark |                 Perf(+ is better)
                   3d-cube.js |   2.620s ->   2.660s :  -1.527%  (+-1.198%) : [-]
       access-binary-trees.js |   1.405s ->   1.370s :  +2.491%  (+-0.626%) : [+]
           access-fannkuch.js |   8.003s ->   7.952s :  +0.645%  (+-0.906%) : [≈]
              access-nbody.js |   3.103s ->   3.062s :  +1.342%  (+-0.824%) : [+]
  bitops-3bit-bits-in-byte.js |   1.735s ->   1.745s :  -0.576%  (+-0.737%) : [≈]
       bitops-bits-in-byte.js |   2.663s ->   2.663s :  +0.000%  (+-0.814%) : [≈]
        bitops-bitwise-and.js |   3.647s ->   3.575s :  +1.965%  (+-0.441%) : [+]
     controlflow-recursive.js |   0.890s ->   0.892s :  -0.187%  (+-0.755%) : [≈]
                crypto-aes.js |   4.623s ->   4.612s :  +0.252%  (+-0.324%) : [≈]
                crypto-md5.js |  21.475s ->  20.520s :  +4.447%  (+-0.574%) : [+]
               crypto-sha1.js |   9.740s ->   9.293s :  +4.586%  (+-0.466%) : [+]
         date-format-tofte.js |   2.222s ->   2.152s :  +3.151%  (+-0.768%) : [+]
         date-format-xparb.js |   1.057s ->   1.052s :  +0.473%  (+-1.417%) : [≈]
               math-cordic.js |   3.017s ->   3.005s :  +0.387%  (+-0.535%) : [≈]
         math-partial-sums.js |   1.662s ->   1.657s :  +0.301%  (+-0.651%) : [≈]
        math-spectral-norm.js |   1.483s ->   1.507s :  -1.573%  (+-0.817%) : [-]
              string-fasta.js |   4.555s ->   4.485s :  +1.537%  (+-0.802%) : [+]
              Geometric mean: |                 Speed up: 1.058% (+-0.192%) : [+]

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -522,16 +519,22 @@ ecma_create_named_data_property (ecma_object_t *obj_p, /**< object */
...
+  {
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_CONFIGURABLE);
+  }
+  if (is_enumerable)
+  {
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_ENUMERABLE);
+  }
+  if (is_writable)
+  {
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_WRITABLE);
```
akiss77: Wouldn't `prop_p->flags |= (uint8_t) ECMA_PROPERTY_FLAG_xxx;` work here and above? (It's a bit more compact.)

Actually, using a local variable to compose the flags and then copy that into `prop_p->flags` would be even nicer.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -568,18 +571,20 @@ ecma_create_named_accessor_property (ecma_object_t *obj_p, /**< object */
...
+  ECMA_SET_NON_NULL_POINTER (prop_p->v.named_accessor_property.name_p, name_p);
 
-  prop_p->u.named_accessor_property.is_lcached = false;
+  if (is_configurable)
+  {
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_CONFIGURABLE);
+  }
+  if (is_enumerable)
+  {
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_ENUMERABLE);
```
akiss77: Same here.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -861,28 +866,19 @@ void
...
+  if (prop_p->flags & ECMA_PROPERTY_FLAG_NAMEDDATA)
+  {
+    ecma_free_named_data_property (object_p, prop_p);
+  }
+  else if (prop_p->flags & ECMA_PROPERTY_FLAG_NAMEDACCESSOR)
+  {
+    ecma_free_named_accessor_property (object_p, prop_p);
+  }
+  else
+  {
```
akiss77: I'd add an assert here.
`JERRY_ASSERT (prop_p->flags & ECMA_PROPERTY_FLAG_INTERNAL);`

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1096,9 +1094,16 @@ ecma_set_property_writable_attr (ecma_property_t *prop_p, /**< property */
...
+  JERRY_ASSERT (prop_p->flags & ECMA_PROPERTY_FLAG_NAMEDDATA);
 
-  prop_p->u.named_data_property.writable = is_writable ? ECMA_PROPERTY_WRITABLE : ECMA_PROPERTY_NOT_WRITABLE;
+  if (is_writable)
+  {
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_WRITABLE);
+  }
+  else
+  {
+    prop_p->flags = (uint8_t) (prop_p->flags & ~ECMA_PROPERTY_FLAG_WRITABLE);
```
akiss77: Again, `|=` and `&=` could be used.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1130,17 +1128,15 @@ ecma_set_property_enumerable_attr (ecma_property_t *prop_p, /**< property */
...
-                                                ECMA_PROPERTY_ENUMERABLE : ECMA_PROPERTY_NOT_ENUMERABLE);
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_ENUMERABLE);
   }
   else
   {
-    JERRY_ASSERT (prop_p->type == ECMA_PROPERTY_NAMEDACCESSOR);
-
-    prop_p->u.named_accessor_property.enumerable = (is_enumerable ?
-                                                    ECMA_PROPERTY_ENUMERABLE : ECMA_PROPERTY_NOT_ENUMERABLE);
+    prop_p->flags = (uint8_t) (prop_p->flags & ~ECMA_PROPERTY_FLAG_ENUMERABLE);
```
akiss77: ditto

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1173,17 +1162,15 @@ ecma_set_property_configurable_attr (ecma_property_t *prop_p, /**< property */
...
-                                                  ECMA_PROPERTY_CONFIGURABLE : ECMA_PROPERTY_NOT_CONFIGURABLE);
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_CONFIGURABLE);
   }
   else
   {
-    JERRY_ASSERT (prop_p->type == ECMA_PROPERTY_NAMEDACCESSOR);
-
-    prop_p->u.named_accessor_property.configurable = (is_configurable ?
-                                                      ECMA_PROPERTY_CONFIGURABLE : ECMA_PROPERTY_NOT_CONFIGURABLE);
+    prop_p->flags = (uint8_t) (prop_p->flags & ~ECMA_PROPERTY_FLAG_CONFIGURABLE);
```
akiss77: ditto

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1215,16 +1194,15 @@ void
...
-  if (prop_p->type == ECMA_PROPERTY_NAMEDDATA)
+  if (is_lcached)
   {
-    prop_p->u.named_data_property.is_lcached = (is_lcached != 0);
+    prop_p->flags = (uint8_t) (prop_p->flags | ECMA_PROPERTY_FLAG_LCACHED);
   }
   else
   {
-    prop_p->u.named_accessor_property.is_lcached = (is_lcached != 0);
+    prop_p->flags = (uint8_t) (prop_p->flags & ~ECMA_PROPERTY_FLAG_LCACHED);
```
akiss77: ditto

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -1815,7 +1815,8 @@ ecma_builtin_string_prototype_object_split (ecma_value_t this_arg, /**< this arg
               ecma_string_t *magic_index_str_p = ecma_get_magic_string (LIT_MAGIC_STRING_INDEX);
               ecma_property_t *index_prop_p = ecma_get_named_property (match_array_obj_p, magic_index_str_p);
 
-              ecma_number_t *index_num_p = ecma_get_number_from_value (index_prop_p->u.named_data_property.value);
+              ecma_value_t index_value = ecma_get_named_data_property_value (index_prop_p);
```
akiss77: Is the `index_value` local var needed? Couldn't we just inline `ecma_get_named_data_property_value (index_prop_p)` into the line below?

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -472,12 +471,10 @@ ecma_find_internal_property (ecma_object_t *object_p, /**< object descriptor */
        property_p != NULL;
        property_p = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p))
   {
-    if (property_p->type == ECMA_PROPERTY_INTERNAL)
+    if (property_p->flags & ECMA_PROPERTY_FLAG_INTERNAL
```
akiss77: I'd suggest using some more parentheses to make operator precedence explicit. (Just like in other parts of the patch.)

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -214,84 +204,78 @@ typedef enum
...
-      unsigned int writable : 1;
-
-      /** Attribute 'Enumerable' (ecma_property_enumerable_value_t) */
-      unsigned int enumerable : 1;
+      mem_cpointer_t name_p;
 
-      /** Attribute 'Configurable' (ecma_property_configurable_value_t) */
-      unsigned int configurable : 1;
+      /** Lower 16 bits of value */
+      uint16_t value_low;
```
akiss77: With this layout, we are expecting that the value of a named data property fits into 8+16 bits. Right now, this is true, but it would be good to check somehow that `ECMA_VALUE_SIZE` (which gets calculated from at least 2 components from different header files) is less than 24. (A static assert could work, but according to its documentation, they are not to be placed in headers.)

jerry-core/ecma/base/ecma-helpers.c
zherczeg: No :( c99 int conversion issue. I cannot say how much I hate it but only ++ and -- is accepted for types smaller than it. Everything else will be a warning. I even introduced macros for them in the parser.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
zherczeg: Line length (>120 char) issue.

jerry-core/ecma/base/ecma-helpers.c
akiss77: Ahh, too bad. I was not aware of that limitation. OK then.

jerry-core/ecma/base/ecma-helpers.c
zherczeg: I was thinking about it, but the free_internal property checks it. From the backtrace it is visible where it is called.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
akiss77: The two expressions can still be merged into one, only the lines have to be broken at the right places in a nicely-formatted way.

zherczeg: You suggested using local variables above, and here you don't like it :) Anyway, I would prefer that myself, but I couldn't do it. Please suggest me a form which is valid according to Jerry formatting guidelines.

jerry-core/ecma/base/ecma-helpers.c
akiss77: True. But still, IMHO it is good coding practice which helps maintainibility to state that "normally, I'd have written `else if` here but I'm sure that its predicate will always hold therefore I just go with an `else`, however, I still assert on that predicate". And what's more, it does not cost anything in release builds.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
akiss77: I'm sure you realize the differences between the two cases. With the flags above, you are accessing a field via a pointer multiple times. There, it makes sense to use a local variable to compose the needed value and then do the indirection only once. Here, however, a local variable is introduced just to get a line within length limits...

Something like below may work with stlye checkers (however, this no-indent style looks ugly to me :/ ). There is something similar in rcs/rcs-chunked-list.c already.
```
              ecma_value_t index_value =
              ecma_get_named_data_property_value (
              ecma_get_number_from_value (
              index_prop_p->u.named_data_property.value));
```

Eventually, I'm (un)happy either way.

zherczeg: This is too hard to read.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -214,84 +204,78 @@ typedef enum
...
+    struct
     {
-      /** Internal property's type */
-      unsigned int type : ECMA_PROPERTY_INTERNAL_PROPERTY_TYPE_WIDTH;
-
       /** Value (may be a compressed pointer) */
       uint32_t value;
     } internal_property;
-  } u;
+  } v;
```
LaszloLango: I'd prefer a real name for this. `value`?

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -214,84 +204,78 @@ typedef enum
...
-    /** Description of named data property */
-    struct __attr_packed___ ecma_named_data_property_t
-    {
-      /** Value */
-      __extension__ ecma_value_t value : ECMA_VALUE_SIZE;
+    /** Named data property value upper bits */
+    uint8_t named_data_property_value_high;
+    /** Internal property type */
+    uint8_t internal_property_type;
+  } h;
```
LaszloLango: ditto, `header` ?

zherczeg: The problem is the names will be too long, and in some cases it is easy to reach the 120 char.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1107,19 +1112,12 @@ ecma_set_property_writable_attr (ecma_property_t *prop_p, /**< property */
...
-    return (prop_p->u.named_data_property.enumerable == ECMA_PROPERTY_ENUMERABLE);
-  }
-  else
-  {
-    JERRY_ASSERT (prop_p->type == ECMA_PROPERTY_NAMEDACCESSOR);
+  JERRY_ASSERT (prop_p->flags & (ECMA_PROPERTY_FLAG_NAMEDDATA | ECMA_PROPERTY_FLAG_NAMEDACCESSOR));
 
-    return (prop_p->u.named_accessor_property.enumerable == ECMA_PROPERTY_ENUMERABLE);
-  }
+  return (prop_p->flags & ECMA_PROPERTY_FLAG_ENUMERABLE) != 0;
```
LaszloLango: ` != 0;` looks unnecessary for me

LaszloLango: Sorry, my bad. It is good.

jerry-core/ecma/base/ecma-globals.h
LaszloLango: Ok, then leave as is.

zherczeg:
On Sunspider benchmark suite the currently used MaxRSS measuring script reported a large memory growth, and I decided to analyse the reason. First I created a new valgrind tool called heimdall which reports all page ranges that are accessed by an application. After that I picked bitops-bitwise-and.js which memory consumption is growth from 24K to 32K.

Here are the results:

Original:

    Page size: 4096 bytes
      Modified pages: 0x3e000 - 0x3f000 [size: 4 Kb]
      Modified pages: 0x13e000 - 0x13f000 [size: 4 Kb]
      Modified pages: 0x17e000 - 0x180000 [size: 8 Kb]
      Modified pages: 0x27f000 - 0x280000 [size: 4 Kb]
      Modified pages: 0x7dbcb000 - 0x7dbcd000 [size: 8 Kb]
    Total modified pages: 28 Kb

It turned out that the MaxRSS measuring script underestimated the real memory consumption. This is a known weakness of the script, since its precision depends on its sampling frequency.

After the patch:

    Page size: 4096 bytes
      Modified pages: 0x3d000 - 0x3e000 [size: 4 Kb]
      Modified pages: 0x13d000 - 0x13f000 [size: 8 Kb]
      Modified pages: 0x17d000 - 0x17f000 [size: 8 Kb]
      Modified pages: 0x27e000 - 0x27f000 [size: 4 Kb]
      Modified pages: 0x7d96b000 - 0x7d96d000 [size: 8 Kb]
    Total modified pages: 32 Kb

Because of the binary size reduction the compiler decreased the base address of the data section which overlaps two pages now.

However, the interesting part comes here. We change the page size to 1K (down from 4K) and measured these tests again:

Original:

    Page size: 1024 bytes
      Modified pages: 0x3e000 - 0x3e400 [size: 1 Kb]
      Modified pages: 0x13e000 - 0x13e800 [size: 2 Kb]
      Modified pages: 0x17e000 - 0x17f400 [size: 5 Kb]
      Modified pages: 0x27f000 - 0x27f400 [size: 1 Kb]
      Modified pages: 0x7dbb3c00 - 0x7dbb4800 [size: 3 Kb]
    Total modified pages: 12 Kb

After the patch:

    Page size: 1024 bytes
      Modified pages: 0x3dc00 - 0x3e000 [size: 1 Kb]
      Modified pages: 0x13dc00 - 0x13e400 [size: 2 Kb]
      Modified pages: 0x17dc00 - 0x17f000 [size: 5 Kb]
      Modified pages: 0x27ec00 - 0x27f000 [size: 1 Kb]
      Modified pages: 0x7dbddc00 - 0x7dbde800 [size: 3 Kb]
    Total modified pages: 12 Kb

We can notice that the memory consumption decreased to 12K, and that is not changed after the patch. The ranges are the same, although the base address with the patch is a bit lower than the original. So the page alignment increase the 12K memory consumption to 28-32K on Linux. A huge waste! Fortunately this does not affect the microcontrollers, our real targets, so I propose to land this patch.

LaszloLango:
LGTM

zherczeg:
Landed in b8b587f76c41fcd3b46550642a64d0fad34f5610.

PR #907, opened 2016-02-20
Drop unimplemented run context stubs

akiss77:
Run context stubs have been lingering in the code base for almost a
year but they have never been more than unimplemented ideas.
Moreover, they are guarded by a macro that's nowhere defined, thus
the code is never compiled or tested.

Should contexts be implemented in the future, they may or may not
be implemented the way they are sketched now. Thus, it's safe to
remove this dead code from the sources.

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -1751,6 +1751,7 @@ jerry_run (void)
 
   return vm_run_global ();
 } /* jerry_run */
+
 /**
  * Simple jerry runner
  *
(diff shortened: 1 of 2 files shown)


LaszloLango:
LGTM. If we decide that we want such a feature, then we can bring it back. I think in the current phase of Jerry (optimizations) we shouldn't leave unimplemented functions in the master.

zherczeg:
LGTM

PR #909, opened 2016-02-21
Improve the usability of test runner scripts

akiss77:
* First of all, remove the counter-intuitive "OUT_DIR" second
  argument of run-test-suite.sh. This, way, the invocation of the
  script becomes easier to remember:
  `tools/runners/run-test-suite.sh <engine> <testsuite>`
  However, this also means that all output files (lists of
  executed, passed, and failed tests) are generated in the current
  working directory.

* Align the behaviour of run-unittests.sh with the above, i.e.,
  don't try to guess where to put output files but write them in
  the CWD.

* Adapt Makefile to the change in the use of the test runner
  scripts: create and change to "check" directories before invoking
  test runner scripts.

Extras:
* tools/runners/run-test-suite.sh collected fail tests from
  directories twice. This does no harm but is inefficient, thus
  removing.
* tools/runners/run-test-suite.sh was too permissive on the
  contents of test suite list files. Better to accept those lines
  only which really contain paths to JS test files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -105,6 +105,15 @@ export NATIVE_MODS := $(MCU_MODS) mem_stats mem_stress_test
  # Compiler to use for external build
  EXTERNAL_C_COMPILER ?= arm-none-eabi-gcc
 
+# Directories
+export ROOT_DIR := $(shell pwd)
+export BUILD_DIR_PREFIX := $(ROOT_DIR)/build/obj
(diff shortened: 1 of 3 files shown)


zherczeg:
I like simplifications, LGTM

LaszloLango:
LGTM

PR #912, opened 2016-02-22
Fix 'jerry-test-suite/13/13-011.js' fail in compact profile

LaszloLango:
Arguments object is not supported in compact profile, so remove
the test from 'compact-profile-list'. Do not create arguments
object on function calls in compact profile. Remove unnecessary
arguments check.
Related issue: #879


zherczeg:
LGTM

akiss77:
LGTM too (FWIW)

PR #924, opened 2016-02-29
Remove unused `EXTERNAL_` variables from Makefile

akiss77:
They have been kept when the build system was refactored but it
turns out that they are not used at all (they are not passed on to
cmake).


LaszloLango:
LGTM

zherczeg:
LGTM

PR #925, opened 2016-03-01
Remove default handling of build options from Makefile

akiss77:
The handling of option defaults is already done in CMakeLists, it
is no good practice to duplicate the logic in the Makefile.


LaszloLango:
LGTM

zherczeg:
LGTM

PR #928, opened 2016-03-02
Add `STRIP_RELEASE_BINARY` option to Makefile

akiss77:
It is already supported by CMakeLists, might be useful to allow it
to be passed through Makefile.


zherczeg:
Isn't this the default?

akiss77:
On Linux, Darwin, and OpenWRT targets, this is ON by default, on MCU targets, the default is OFF. It'd be good to have a way of overriding it.

E.g., I'd like to have symbols in my release linux binaries to be able to get at least somewhat meaningful objdumps/disassemblies -- this would need turning that option OFF.

(According to my understanding, it has no effect on the "important" parts of the binary, e.g., on text and data sizes.)

LaszloLango:
LGTM

zherczeg:
LGTM

Issue #923, opened 2016-02-29
Date-related test failures on arm/linux


akiss77:
On current master (5d3aa98), arm/linux target fails on 5 date-related tests. Details below.

I've been building jerry on Ubuntu 14.04.3 LTS x86_64 GNU/Linux 3.13.0 with cross toolchain arm-linux-gnueabihf-gcc 4.8.2, and testing it with qemu user space emulation as follows:
```
TOOLCHAIN=build/configs/toolchain_linux_armv7l-hf.cmake TIMEOUT=300 make NINJA=1 test-js.debug.linux.j
```

5 tests in the suite fail (all seem to be date-related):
```
[32/259] /home/akiss/devel/jerry/jerryscript/build/bin/debug.linux/jerry  /home/akiss/devel/jerry/jerryscript/tests/jerry/date-annexb.js: FAIL (1)
Script assertion failed
[33/259] /home/akiss/devel/jerry/jerryscript/build/bin/debug.linux/jerry  /home/akiss/devel/jerry/jerryscript/tests/jerry/date-construct.js: FAIL (1)
Script assertion failed
[34/259] /home/akiss/devel/jerry/jerryscript/build/bin/debug.linux/jerry  /home/akiss/devel/jerry/jerryscript/tests/jerry/date-getters.js: FAIL (1)
Script assertion failed
[36/259] /home/akiss/devel/jerry/jerryscript/build/bin/debug.linux/jerry  /home/akiss/devel/jerry/jerryscript/tests/jerry/date-setters.js: FAIL (1)
Script assertion failed
[37/259] /home/akiss/devel/jerry/jerryscript/build/bin/debug.linux/jerry  /home/akiss/devel/jerry/jerryscript/tests/jerry/date-tostring.js: FAIL (1)
Script assertion failed
```

To reproduce the above failures, the installation of some packages (cross-gcc and qemu) may be needed:
```
sudo apt-get install gcc-arm-linux-gnueabihf libc6-dev-armhf-cross qemu-user-static
```

The failures are blocking #898

PR #929, opened 2016-03-02
Disable date object related system calls by default.

LaszloLango:
`gettimeofday` system call cause problems on multiple platforms, so better to disable by default.

Related issue: #923

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
@@ -26,16 +26,10 @@
 
 #ifndef CONFIG_ECMA_COMPACT_PROFILE_DISABLE_DATE_BUILTIN
 
+#ifdef JERRY_ENABLE_DATE_SYS_CALLS
 #include <sys/time.h>
 
-/**
- * Timezone structure
- */
-struct timezone
-{
-  int tz_minuteswest;     /**< minutes west of Greenwich */
-  int tz_dsttime;         /**< type of DST correction */
-};
+#endif /* JERRY_ENABLE_DATE_SYS_CALLS */
 
 /** \addtogroup ecma ECMA
  * @{
@@ -456,15 +450,21 @@ ecma_date_week_day (ecma_number_t time) /**< time value */
 ecma_number_t __attr_always_inline___
 ecma_date_local_tza ()
 {
+#ifdef JERRY_ENABLE_DATE_SYS_CALLS
   struct timeval tv;
   struct timezone tz;
 
+  tz.tz_minuteswest = 0; /* gettimeofday may not fill tz, so zero-initializing */
+
   if (gettimeofday (&tv, &tz) != 0)
   {
-    return ecma_raise_type_error ("gettimeofday failed");
+    return ecma_number_make_nan ();
   }
 
   return tz.tz_minuteswest * -ECMA_DATE_MS_PER_MINUTE;
+#else /* !JERRY_ENABLE_DATE_SYS_CALLS */
+  return ECMA_NUMBER_ZERO;
+#endif /* JERRY_ENABLE_DATE_SYS_CALLS */
 } /* ecma_date_local_tza */
 
 /**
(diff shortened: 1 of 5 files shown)


akiss77:
Issue analysis:

In jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c, we rely on the second argument of `gettimeofday(&tv, &tz)` to determine the time zone. Unfortunately, Linux manuals state that "The use of the timezone structure is obsolete; the tz argument should normally be specified as NULL." Our bad luck is that qemu takes this seriously, and its `gettimeofday` syscall implementation simply avoids the second argument: it does not fill anything there, even if the pointer is non-NULL. 

(Note: Real Linux kernels still provide valid tz info, that's why we haven't faced this problem until now. Moreover, our main target is a real device and not qemu emulation, of course. However, qemu is our only chance to get ARM tests work on Travis. So we should deal with the issue, IMHO.)

We could work around the trouble. At both places where we use the second argument, we use `struct timezone tz;`. If we would pre-initialize `tz` with zeros (`= { 0, 0 };`), then however we run jerry, we wouldn't find garbage in `tz.tz_minuteswest` and thus would be on the safe side.

Conclusions:

I wouldn't guard out date syscalls but would go with the initializations.

Notes:

- Changing `ecma_raise_type_error ("gettimeofday failed")` to `ecma_number_make_nan ()` seems an unrelated change to me. Why's that?
- I'd definitely move `struct timezone` to sys/time.h since it's surely not the responsibility of ecma-builtin-helpers-date.c. (I haven't noticed its introduction in the last buildfix.)
- The above is kinda hipothetical, it hasn't beed tried out on jerry but on an example test app only.

LaszloLango:
> Changing ecma_raise_type_error ("gettimeofday failed") to ecma_number_make_nan () seems an unrelated change to me. Why's that?

The return value is `ecma_number_t` and `ecma_raise_type_error` returns `ecma_value_t`, so this is a bug and should be fixed.

LaszloLango:
@akiss77, updated the PR. There are still build issues with external compilers. (Ex.: https://launchpad.net/gcc-arm-embedded/5.0/5-2015-q4-major/+download/gcc-arm-none-eabi-5_2-2015q4-20151219-linux.tar.bz2) So I still prefer to disable by default.

jerry-core/config.h
```diff
@@ -180,4 +180,12 @@
  */
 #define CONFIG_PARSER_ENABLE_PARSE_TIME_BYTE_CODE_OPTIMIZER
 
+/**
+  * Enable timezone and daylightsaving system calls for date object
+  *
+  * Note:
+  *      Experimental. Works on Ubuntu 14.04.
+  */
+// #define ECMA_ENABLE_DATE_SYS_CALLS
```
akiss77: Is there a way to enable the calls without editing config.h? If not, we wont be testing it "ever" -- neither with local precommit checks, nor on Travis. Which makes all gettimeofday-related code paths effectively dead. External tool chains are not the primary targets (strictly IMHO). If their users need tweaks, they shoud be the one to do edits to config.h.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
```diff
@@ -456,15 +450,19 @@ ecma_date_week_day (ecma_number_t time) /**< time value */
 ecma_number_t __attr_always_inline___
 ecma_date_local_tza ()
 {
-  struct timeval tv;
-  struct timezone tz;
+#ifdef ECMA_ENABLE_DATE_SYS_CALLS
+  struct timeval tv = { 0, 0 }; /* gettimeofday may not fill tv, so zero-initializing */
```
akiss77: Actually, _this_ comment is not true. Only `tz` is obsolete.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
```diff
@@ -483,15 +481,19 @@ ecma_date_daylight_saving_ta (ecma_number_t time) /**< time value */
     return time; /* time is NaN */
   }
 
-  struct timeval tv;
-  struct timezone tz;
+#ifdef ECMA_ENABLE_DATE_SYS_CALLS
+  struct timeval tv = { 0, 0 }; /* gettimeofday may not fill tv, so zero-initializing */
```
akiss77: ditto

jerry-core/config.h
LaszloLango: It looked like to me that is the common way. (see: CONFIG_VM_RUN_GC_AFTER_EACH_OPCODE) But I agree with you. A build option would be good.

LaszloLango:
@akiss77, updated the patch. You can enable sys calls with `DATE_SYS_CALLS=ON make...` build option.

akiss77:
The guards still make me sad, but LGTM (FWIW).

zherczeg:
LGTM

PR #935, opened 2016-03-07
Build fix for mbedk64f target

robertsipka:

### DIFF
--- a/jerry-core/jerry-port.c
+++ b/jerry-core/jerry-port.c
@@ -50,5 +50,5 @@ int jerry_port_errormsg (const char *format, /**< format string */
  */
 int jerry_port_putchar (int c) /**< character to put */
 {
-  return putchar (c);
+  return putchar ((unsigned char) c);
 } /* jerry_port_putchar */


LaszloLango:
LGTM

jerry-core/jerry-port.c
```diff
@@ -50,5 +50,5 @@ int jerry_port_errormsg (const char *format, /**< format string */
  */
 int jerry_port_putchar (int c) /**< character to put */
 {
-  return putchar (c);
+  return putchar ((unsigned char) c);
```
zherczeg: Why this is a problem?

http://www.cplusplus.com/reference/cstdio/putchar/

putchar should accept int argument.

robertsipka:
@zherczeg : A conversion error (conversion to 'unsigned char' from 'int' may alter its value) occurs when I tried to build to mbedk64f target with enabled USE_COMPILER_DEFAULT_LIBC. 
It's the default value of this option (hard-coded in the Makefile.mbedk64f), and the same happens when I compile another board with the following options:
USE_COMPILER_DEFAULT_LIBC=YES make release.mcu_stm32f3-cp
As I assume this is a compiler-specific error, and I dared to casting int to unsigned char because the value is internally converted to an unsigned char anyway.

zherczeg:
Btw could we make jerry_port_putchar to accept unsigned char? Is it frequently used?

robertsipka:
No, it's very rarely used, but we couldn't get away with the cast, for example the following line :
jerry_port_putchar (ecma_char)
 will be reported the same error, because the type of the variable is an ecma_char_t which is a typedef to uint16_t.
(In case of this kind of build configuration)

zherczeg:
LGTM

robertsipka:
Rebased with master.

PR #936, opened 2016-03-07
Change pool-based allocation to heap-based allocation in jerry.c

akiss77:
There was an abuse of memory pools in jerry.c: `compiled_code_map_entry_t`
is not an ECMA data model object but was still alocated/freed via
mem pool functions. Changed the appropriate calls to heap-related
functions.

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -1851,7 +1851,8 @@ snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled
 
   snapshot_last_compiled_code_offset = snapshot_buffer_write_offset;
 
-  compiled_code_map_entry_t *new_entry = (compiled_code_map_entry_t *) mem_pools_alloc ();
+  compiled_code_map_entry_t *new_entry;
+  new_entry = (compiled_code_map_entry_t *) mem_heap_alloc_block (sizeof (compiled_code_map_entry_t));
 
   if (new_entry == NULL)
   {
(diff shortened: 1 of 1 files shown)


jerry-core/jerry.c
```diff
@@ -1848,7 +1848,8 @@ snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled
 
   snapshot_last_compiled_code_offset = snapshot_buffer_write_offset;
 
-  compiled_code_map_entry_t *new_entry = (compiled_code_map_entry_t *) mem_pools_alloc ();
+  compiled_code_map_entry_t *new_entry =
+  (compiled_code_map_entry_t *) mem_heap_alloc_block (sizeof (compiled_code_map_entry_t));
```
zherczeg: I think this is an incorrect style. A temporary variable could help.

akiss77: No. Seriously. I - will - not - add - extra - statements, - temporaries, - assignments - just - because - of - a - style - checker.

Style checkers should promote readability. Nicer style. But as soon as they force you to write _different_ (not differently styled!) code than you think it's optimal, then there is something terribly wrong.

(And the other issue is: the temp should be of type `compiled_code_map_entry_t *`, I guess. But then, there is no way to get that line under the 120 chars limit. Even a one-character temp would make a 121 chars line. So, that line should be broken somewhere too, which causes the same issue as we have now.)

zherczeg: But the proposed solution looks ugly. Introduce a two parameter macro for allocating, or improve the style checker.

LaszloLango:
```C++
compiled_code_map_entry_t *new_entry;
new_entry = (compiled_code_map_entry_t *) mem_heap_alloc_block (sizeof (compiled_code_map_entry_t));
```

akiss77:
Updated.
@LaszloLango thanks for the idea.

zherczeg:
LGTM

LaszloLango:
LGTM

PR #938, opened 2016-03-07
Warning fixes.

robertsipka:


jerry-core/mem/mem-poolman.c
```diff
@@ -34,16 +34,21 @@
...
-typedef struct
+struct mem_pool_chunk
 {
-  struct mem_pools_chunk_t *next_p; /* pointer to next pool chunk */
+  mem_pools_chunk_t *next_p; /* pointer to next pool chunk */
 #ifndef MEM_HEAP_PTR_64
   uint32_t dummy; /* dummy member for alignment */
 #endif
-} mem_pools_chunk_t;
+} mem_pool_chunk;
```
akiss77: This rewrite doesn't seem to be equivalent to the original version. This code defines a `struct mem_pool_chunk` -- and then also defines a variable named `mem_pool_chunk` of type `struct mem_pool_chunk`. This global variable is certainly not what we want.

BTW, why do we need this rewrite?

robertsipka: Yes, you are right. I changed this with another solution.
This modification fix the following warnings:
ezeket fixálja:
150:22: assignment from incompatible pointer type : mem_free_chunk_p = chunk_p->next_p;
173:27: assignment from incompatible pointer type : chunk_to_free_p->next_p = mem_free_chunk_p;
190:39: initialization from incompatible pointer type : mem_pools_chunk_t *const next_p = mem_free_chunk_p->next_p;

jerry-core/parser/regexp/re-compiler.c
```diff
@@ -472,7 +472,7 @@ re_find_bytecode_in_cache (ecma_string_t *pattern_str_p, /**< pattern string */
           && ecma_compare_ecma_strings (cached_pattern_str_p, pattern_str_p))
       {
         JERRY_DDLOG ("RegExp is found in cache\n");
-        return re_cache[*idx];
+        return (re_compiled_code_t *) re_cache[*idx];
```
zherczeg: I would change re_cache to store re_compiled_code_t pointers. This is what is there anyway.

robertsipka:
Thanks for the comments. I've updated the patch.
@zherczeg: At the first solution I didn't choose the safest way. I considered that and eliminated these warnings with adding the missing const modifiers, which I think is much closer to the original purpose. Please take a look at it.

jerry-core/mem/mem-poolman.c
```diff
@@ -37,9 +37,9 @@
 /**
  * Node for free chunk list
  */
-typedef struct
+typedef struct mem_pools_chunk
```
akiss77: Thanks for fixing this. However, there are some more recursive `struct`s in the code base, which solve the the issue inconsitently, sometimes even incorrectly. Could you please open an issue or a follow-up PR to have a uniform solution for all recursive `struct`s out there?

LaszloLango:
LGTM

jerry-core/mem/mem-poolman.c
zherczeg: Agreed. I would prefer a naming like mem_pools_chunk_forward_t in general

zherczeg:
LGTM

jerry-core/parser/regexp/re-compiler.c
```diff
@@ -388,7 +388,7 @@ re_parse_alternative (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context
 
         ECMA_TRY_CATCH (empty,
                         re_parse_char_class (re_ctx_p->parser_ctx_p,
-                                             re_append_char_class,
+                                             (re_char_class_callback) re_append_char_class,
```
franc0is: In my opinion, it is bad practice to cast function pointers. Instead, change `re_append_char_class` to match `re_char_class_callback`'s type and cast the uint32_t to ecma_char_t inside the function.

LaszloLango: @franc0is, good point. @robertsipka, please change `typedef void (*re_char_class_callback) (void *re_ctx_p, uint32_t start, uint32_t end);` to `typedef void (*re_char_class_callback) (void *re_ctx_p, ecma_char_t start, ecma_char_t end);` and remove the cast.

robertsipka: Yes, thanks for the comments. I've updated the patch.
@LaszloLango : This modification will affect some helper functions, and more code refactoring needed. I think we could open a follow-up PR to do this kind of conversions.

LaszloLango: OK

robertsipka:
Rebased with master.

Issue #897, opened 2016-02-17
Maintain and improve error messages


LaszloLango:
We must check the current error messages.
- [x] Add missing messages
- [ ] Use magic string for the same errors
- [x] Provide position information (line:column)
- [x] Add build option to use errors without messages (this can save memory)

zherczeg:
No need to add magic strings, the compiler should handle the same string merging.

Closing due to inactivity. Please reopen if needed.

PR #941, opened 2016-03-07
Don't use messages for errors by default

LaszloLango:
Related issue: #897

### DIFF
--- a/jerry-core/ecma/operations/ecma-exceptions.h
+++ b/jerry-core/ecma/operations/ecma-exceptions.h
@@ -28,6 +28,12 @@
  * @{
  */
 
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+#define ECMA_ERR_MSG(msg) msg
+#else /* !JERRY_ENABLE_ERROR_MESSAGES */
+#define ECMA_ERR_MSG(msg) NULL
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
+
 /**
  * Native errors.
  *
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ project (Jerry C ASM)
  option(ENABLE_LTO            "Enable LTO build" ON)
  option(ENABLE_LOG            "Enable LOG build" OFF)
  option(ENABLE_ALL_IN_ONE     "Enable ALL_IN_ONE build" OFF)
+ option(ENABLE_ERROR_MESSAGES "Enable error messages for builtin error objects" OFF)
 
  if("${PLATFORM}" STREQUAL "LINUX")
   set(PLATFORM_EXT "LINUX")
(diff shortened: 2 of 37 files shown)


LaszloLango:
File size of `release.mcu_stm32f3` target
 *  before: 188956 byte
 * after: 187320 byte

Result of `size` command
before:
```
   text	   data	    bss	    dec	    hex	filename
 188884	     72	  22300	 211256	  33938	jerry
```
after:
```
   text	   data	    bss	    dec	    hex	filename
 187248	     72	  22300	 209620	  332d4	jerry
```

jerry-core/ecma/operations/ecma-exceptions.h
```diff
@@ -28,6 +28,12 @@
  * @{
  */
 
+#ifndef JERRY_NDEBUG
+#define ERROR_MSG(msg) msg
+#else /* JERRY_NDEBUG */
+#define ERROR_MSG(msg) NULL
```
akiss77: I'd suggest prefixing this macro, e.g., with `ECMA_`.

jerry-core/ecma/operations/ecma-exceptions.h
```diff
@@ -28,6 +28,12 @@
  * @{
  */
 
+#ifndef JERRY_NDEBUG
```
akiss77: Is it really the release build of the full profile where we want to disable error messages? I'm wondering whether the idea would be more fit for the compact profile, both debug and release builds. (The full profile should provide full ES5.1 functionality, which also means error messages. IMO)

LaszloLango: In compact profile there are no error builtins, so in CP this is not relevant.

LaszloLango: This patch does not take effect on conformance. The error messages are always implementation defined, so they could be empty. There is a `message` property for the error object, but it contains empty string. I think this not a problem when you flash on mcu. In development you could use debug build for checking whats wrong. Further more we currently use really less error messages. After this PR we are able to provide (more descriptive) messages for errors without any negative effect on binary size in release builds.

akiss77: I have doubts then. Actually, I'd argue against this change. This could potentially change the way a script is executed depending only on the build mode of the engine. A debug build should make easier to debug the engine but not add extra features to it (and vice versa, release builds may/should be more optimized thus harder to debug but not lack engine functionality or perceivable features). For me, this goes against these concepts. But it would be good to hear the opinion of others.

zherczeg: It adds extra features like opcode dump.

LaszloLango:
@zherczeg, @akiss77 the patch is updated. I introduced a new build option (`ERROR_MESSAGES`). With  the current settings both release and debug build will use empty messages by default. Please review.

akiss77:
LGTM (FWIW)

zherczeg:
LGTM

Issue #942, opened 2016-03-08
Conversion error to ‘ssize_t’ from ‘jerry_api_size_t’ (in main-unix.c, on arm/linux)


akiss77:
When cross-building jerry for arm/linux with 
`TOOLCHAIN=build/configs/toolchain_linux_armv7l-hf.cmake make NINJA=1 build.linux`, 
the following error is reported:

```
../../../main-unix.c: In function ‘main’:
../../../main-unix.c:563:7: error: conversion to ‘ssize_t’ from ‘jerry_api_size_t’ may change the sign of the result [-Werror=sign-conversion]
       ssize_t sz = jerry_api_string_to_char_buffer (err_str_p, err_str_buf, err_str_size);
```

akiss77:
Introduced by a187e6d6

akiss77:
Solved by #943

PR #943, opened 2016-03-08
Fix size types

akiss77:
Related issue #942

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -204,11 +204,11 @@ ecma_new_ecma_string_from_uint32 (uint32_t uint32_number) /**< UInt32-represente
   string_desc_p->container = ECMA_STRING_CONTAINER_UINT32_IN_DESC;
 
   lit_utf8_byte_t byte_buf[ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32];
-  ssize_t bytes_copied = ecma_uint32_to_utf8_string (uint32_number,
-                                                     byte_buf,
-                                                     ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32);
+  lit_utf8_size_t bytes_copied = ecma_uint32_to_utf8_string (uint32_number,
+                                                             byte_buf,
+                                                             ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32);
 
-  string_desc_p->hash = lit_utf8_string_calc_hash (byte_buf, (lit_utf8_size_t) bytes_copied);
+  string_desc_p->hash = lit_utf8_string_calc_hash (byte_buf, bytes_copied);
 
   string_desc_p->u.common_field = 0;
   string_desc_p->u.uint32_number = uint32_number;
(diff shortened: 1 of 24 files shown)


akiss77:
Note: this PR does not only solve the related issue but does a lot more. The reason is that the issue pointed at a problem (incorrect or illogical use of various `size_t` types) that could have been solved two ways: either with a quick hack (no), or by going through (hopefully) all ecma-to-utf8 string conversion routines (and some more) and applying and propagating all required changes (that's the proper way to go, IMHO).

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -609,29 +609,22 @@ ecma_string_get_array_index (const ecma_string_t *str_p, /**< ecma-string */
 
 /**
  * Convert ecma-string's contents to a cesu-8 string and put it to the buffer.
+ * It is the caller's responsibility to make sure that the string fits in the buffer.
  *
- * @return number of bytes, actually copied to the buffer - if string's content was copied successfully;
- *         otherwise (in case size of buffer is insufficient) - negative number, which is calculated
- *         as negation of buffer size, that is required to hold the string's content.
+ * @return number of bytes, actually copied to the buffer.
```
LaszloLango: What will happen if the buffer is too small?

LaszloLango:
This might take effect on performance too. @bzsolt, could you measure it on RPi2?

bzsolt:
RPi 2 results
master: e926aa7

                     3d-cube.js |   2.538s ->   2.582s :  -1.734%  (+-1.268%) : [-]
         access-binary-trees.js |   1.366s ->   1.356s :  +0.732%  (+-1.163%) : [≈]
             access-fannkuch.js |   7.654s ->   7.616s :  +0.496%  (+-0.852%) : [≈]
                access-nbody.js |   2.938s ->   2.930s :  +0.272%  (+-1.038%) : [≈]
    bitops-3bit-bits-in-byte.js |   1.666s ->   1.682s :  -0.960%  (+-1.517%) : [≈]
         bitops-bits-in-byte.js |   2.404s ->   2.456s :  -2.163%  (+-0.671%) : [-]
          bitops-bitwise-and.js |   3.274s ->   3.248s :  +0.794%  (+-1.174%) : [≈]
          bitops-nsieve-bits.js |  20.134s ->  20.110s :  +0.119%  (+-0.829%) : [≈]
       controlflow-recursive.js |   0.920s ->   0.930s :  -1.087%  (+-0.000%) : [-]
                  crypto-aes.js |   3.544s ->   3.568s :  -0.677%  (+-0.924%) : [≈]
                  crypto-md5.js |   7.130s ->   7.126s :  +0.056%  (+-1.033%) : [≈]
                 crypto-sha1.js |   3.888s ->   3.902s :  -0.360%  (+-1.288%) : [≈]
           date-format-tofte.js |   2.184s ->   2.182s :  +0.092%  (+-1.154%) : [≈]
           date-format-xparb.js |   0.880s ->   0.864s :  +1.818%  (+-1.282%) : [+]
                 math-cordic.js |   2.862s ->   2.902s :  -1.398%  (+-1.122%) : [-]
           math-partial-sums.js |   1.586s ->   1.588s :  -0.126%  (+-1.590%) : [≈]
          math-spectral-norm.js |   1.428s ->   1.442s :  -0.980%  (+-0.916%) : [-]
               string-base64.js |  31.038s ->  31.048s :  -0.032%  (+-1.075%) : [≈]
                string-fasta.js |   3.444s ->   3.360s :  +2.439%  (+-1.068%) : [+]
                Geometric mean: |                Speed up: -0.136% (+-0.253%) : [≈]

Binary sizes (bytes): 187,160 -> 187,164

LaszloLango:
@bzsolt, thanks. @akiss77, LGTM. Please squash before merge.

jerry-core/ecma/base/ecma-helpers-string.c
akiss77: Well, we fail with an assert (in debug). That's aligned with other similar functions (e.g., `ecma_uint32_to_utf8_string`).

Actually, at almost all call sites, we rely on the buffer being preallocated to the exact size of the string. Perhaps the only counter example is `ecma_compare_ecma_strings_relational`, where a buffer of fixed size is available and there was a logic to decide whether to used that fixed buffer or allocate a new one, and that logic really relied on a negative return value from `ecma_string_to_utf8_string`, but it was to rewrite to use `ecma_string_get_size ` instead.

akiss77:
Honestly, I've expected both size and performance gain. Two dead functons have been eliminated, and `ecma_string_to_utf8_string` (which seems to be a quite often called function) does not call `ecma_string_get_size` anymore in release. I expected these to show somewhere. :(

Anyway, the compiler error is fixed at least, as well as abuse of the `size_t` types.

zherczeg:
I am happy you spent time on checking this. Thank you.
LGTM

PR #944, opened 2016-03-08
Add acquire_value API to simplify caller code

franc0is:

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -635,7 +635,40 @@ jerry_api_release_object (jerry_api_object_t *object_p) /**< pointer acquired th
 } /* jerry_api_release_object */
 
 /**
+ * Acquire specified Jerry API value.
+ *
+ * Note:
+ *      For values of string and object types this acquires the underlying data,
+ *      for all other types it is a no-op.
+ *
+ * Warning:
+ *         Acquired pointer should be released with jerry_api_release_value
+ *
+ * @return pointer that may be used outside of the engine
+ */
+jerry_api_value_t *
+jerry_api_acquire_value (jerry_api_value_t *value_p) /**< API value */
+{
+  jerry_assert_api_available ();
+
+  if (value_p->type == JERRY_API_DATA_TYPE_STRING)
+  {
+    jerry_api_acquire_string (value_p->u.v_string);
+  }
+  else if (value_p->type == JERRY_API_DATA_TYPE_OBJECT)
+  {
+    jerry_api_acquire_object (value_p->u.v_object);
+  }
+
+  return value_p;
+} /* jerry_api_acquire_value */
+
+/**
  * Release specified Jerry API value
+ *
+ * Note:
+ *      For values of string and object types this releases the underlying data,
+ *      for all other types it is a no-op.
  */
 void
 jerry_api_release_value (jerry_api_value_t *value_p) /**< API value */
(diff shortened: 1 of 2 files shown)


jerry-core/jerry.c
```diff
@@ -635,6 +635,24 @@ jerry_api_release_object (jerry_api_object_t *object_p) /**< pointer acquired th
 } /* jerry_api_release_object */
 
 /**
+ * Acquire specified Jerry API value
+ */
```
LaszloLango: Missing `@return ...` comment

LaszloLango:
Besides that missing comment the patch looks good to me.

jerry-core/jerry.c
```diff
@@ -635,6 +635,24 @@ jerry_api_release_object (jerry_api_object_t *object_p) /**< pointer acquired th
...
+jerry_api_value_t *
+jerry_api_acquire_value (jerry_api_value_t *value_p) /**< API value */
+{
+  jerry_assert_api_available ();
+
+  if (value_p->type == JERRY_API_DATA_TYPE_STRING)
+  {
+    jerry_api_acquire_string (value_p->u.v_string);
```
LaszloLango: missing `return` statement

jerry-core/jerry.c
```diff
@@ -635,6 +635,24 @@ jerry_api_release_object (jerry_api_object_t *object_p) /**< pointer acquired th
...
+  }
+  else if (value_p->type == JERRY_API_DATA_TYPE_OBJECT)
+  {
+    jerry_api_acquire_object (value_p->u.v_object);
```
LaszloLango: ditto

jerry-core/jerry.c
```diff
@@ -635,6 +635,24 @@ jerry_api_release_object (jerry_api_object_t *object_p) /**< pointer acquired th
...
+  }
```
akiss77: The type of a jerry API value (given with the `jerry_api_data_type_t` enum) can be different from string or object. The function does not deal with them (there is no final `else` branch) nor does it document what the expected outcome on such a value is.

franc0is: @akiss77 In my opinion, this is an implementation detail. Perhaps in the future, new data types have acquire/release semantics as well, in which case `jerry_api_acquire_value` and `jerry_api_release_value` should do the right thing. Note that `jerry_api_release_value` does not have an else clause either.

jerry-core/jerry.c
franc0is: Duh. Sorry for the silly mistake. Added the return statement at the bottom of the function.

jerry-core/jerry.c
franc0is: Added.

franc0is:
Will squash the commits after approval.

jerry-core/jerry.c
akiss77: I humbly disagree. Internal-only functions can be vague about implementation details (even though it's not ideal but that's how it works in practice) but this is the _public_ API of jerry. It _may_ be a no-op for everything beyond strings and objects but it _must_ document this. (My comment above also mentioned "nor does it document".)

And true enough, `jerry_api_release_value` has a deficiency in its documentation. This is no justification for making the same mistake in a new function. Better to get the new function right _and_ fix the existing doc as well.

franc0is: Fair enough! I will add documentation to both.

LaszloLango:
@franc0is, thank you for the updates. LGTM

zherczeg:
LGTM.

I would like to (probably will) rename these acquire / release stuff to ref / deref.

akiss77:
(Once rebased, we can/will land it.)

franc0is:
rebased

PR #945, opened 2016-03-09
Make target provide jerry-port.c

franc0is:
This change-set pushes ports forward a bit and adds a Pebble port (happy to remove the latter if it is inappropriate).


zherczeg:
Please check CI, and I recommend to run make precommit before uploading a patch.

Signed-off-by message is incorrect. The following line should be at the end of the 3413e96 commit's message: 'JerryScript-DCO-1.0-Signed-off-by: François Baldassari francois@pebble.com'.

akiss77:
As you also mention in your comment, this patch does several things. Since some of them are easier to accept than others, I'd suggest splitting it up to several PRs.

* I'm all in for the use of `jerry_port_errormsg` in jrt-fatals. However, there are macros in jrt.h to use it (`JERRY_ERROR_MSG`).
* The use of `jerry_port_putchar` in ecma-builtin-global.c is incorrect. E.g., current implementation prints the char literals backslash-u-zero-zero-zero-zero for a NUL char, while the suggested change here would print the NUL char itself. Similar incompatible change happens to the two-byte code points. It can be OK to use the port function to output a char to the console instead of `printf` but then it should be done correctly.
* About extending the port API, I haven't come to a consensus with myself yet. I have a feeling that it starts duplicating the libc API with slightly different names, which is somewhat disturbing.

franc0is:
@zherczeg I am planning on squashing that commit after I address other comments. Thanks for taking a look!

franc0is:
@akiss77 Thank you for the thoughtful comments. I'll try to address them as best I can:

> As you also mention in your comment, this patch does several things. Since some of them are easier to accept than others, I'd suggest splitting it up to several PRs.

OK - I think I can split it into 2 to start. One for the port, one for some of the CMakelist changes

> I'm all in for the use of jerry_port_errormsg in jrt-fatals. However, there are macros in jrt.h to use it (JERRY_ERROR_MSG).

Ah, I did not see this. Will change.

> The use of jerry_port_putchar in ecma-builtin-global.c is incorrect. E.g., current implementation prints the char literals backslash-u-zero-zero-zero-zero for a NUL char, while the suggested change here would print the NUL char itself.

As best I can tell, those two are equivalent. `\u0000` is the unicode representation of the NUL character.

> Similar incompatible change happens to the two-byte code points. It can be OK to use the port function to output a char to the console instead of printf but then it should be done correctly.

You are right. I need to encode those bytes before I output them with putchar. Thanks for catching this, I will fix it.

> About extending the port API, I haven't come to a consensus with myself yet. I have a feeling that it starts duplicating the libc API with slightly different names, which is somewhat disturbing.

The reason I introduced `jerry_port_abort` rather than simply rely on libc functionality is that on my platform I want to do something specific to jerryscript (i.e. clean-up the JS context and show a dialog on the device) whereas libc's abort is necessarily more broad.

franc0is:
Split #955 and #956 out of this PR

targets/pebble/source/jerry_port.c
```diff
@@ -0,0 +1,56 @@
...
+/**
+ * Provide output character to console implementation for the engine.
+ */
+int jerry_port_putchar (int c)
+{
+  dbgserial_putchar((uint8_t)c);
+  return c;
+}
+
+void wtf(void);
```
LaszloLango: Please use a better name

targets/pebble/source/jerry_port.c
```diff
@@ -0,0 +1,56 @@
...
+
+void jerry_port_abort () {
```
LaszloLango: Brace should be placed on a separate line

build/configs/toolchain_external.cmake
```diff
@@ -17,4 +17,4 @@ include(CMakeForceCompiler)
 set(CMAKE_SYSTEM_NAME EXTERNAL)
 set(CMAKE_SYSTEM_PROCESSOR "${EXTERNAL_CMAKE_SYSTEM_PROCESSOR}")
 
-CMAKE_FORCE_C_COMPILER(${EXTERNAL_CMAKE_C_COMPILER} GNU)
+CMAKE_FORCE_C_COMPILER(${EXTERNAL_CMAKE_C_COMPILER} ${EXTERNAL_CMAKE_C_COMPILER_FAMILY})
```
LaszloLango: Looks duplicated change of #955

targets/pebble/source/jerry_port.c
franc0is: This is the name of the abort API internal to the Pebble firmware, I cannot change it easily. Only option is to remove the Pebble port from this PR.

build/configs/toolchain_external.cmake
franc0is: Good catch

targets/pebble/source/jerry_port.c
franc0is: Fixed.

franc0is:
I removed the pebble port (it is of little use to those without access to the pebble firmware anyways), so this PR is quite simple now.

franc0is:
Ping! I think this does not overalp with the proposal in #964 and should be considered individually.

LaszloLango:
LGTM

zherczeg:
LGTM

franc0is:
rebased.

PR #946, opened 2016-03-09
Allow the port to allocate/deallocate the heap

franc0is:
This builds upon #945, which should be reviewed first.

See https://github.com/Samsung/jerryscript/commit/10a1bb53dc3f87f47f003a88c64d8deba40b9b1f for the description & changes specific to this PR.


jerry-core/jerry-port.h
```diff
@@ -35,6 +41,18 @@ int jerry_port_logmsg (FILE *stream, const char *format, ...);
 int jerry_port_errormsg (const char *format, ...);
 int jerry_port_putchar (int c);
 
+void jerry_port_abort (void);
+
+
+/**
+ * Target port functions for memory management
+ */
+mem_heap_t *jerry_port_init_heap (void);
```
zherczeg: I think we should add a size argument to this function. At least the implementation can check the size.

jerry-core/jerry.c
```diff
@@ -2444,15 +2444,3 @@ jerry_api_get_string_size (const jerry_api_string_t *str_p) /**< input string */
...
- *
- * @return number of characters in the string
- */
-jerry_api_length_t
-jerry_api_get_string_length (const jerry_api_string_t *str_p) /**< input string */
-{
-  jerry_assert_api_available ();
-
-  return ecma_string_get_length ((ecma_string_t *) str_p);
-} /* jerry_api_get_string_length */
```
zherczeg: Why do you remove this function?

jerry-core/mem/mem-heap-internal.h
```diff
@@ -0,0 +1,81 @@
...
+
+  /**
+   * Current limit of heap usage, that is upon being reached, causes call of "try give memory back" callbacks
+   */
+  size_t limit;
+
+  /**
+   * Heap area
+   */
+  uint8_t area[MEM_HEAP_AREA_SIZE] __attribute__ ((aligned (MEM_ALIGNMENT)));
```
zherczeg: This isn't that simple. The allocator should ensure the alignment. I would only put this member into a dynamic struct.

jerry-core/mem/mem-heap.c
```diff
@@ -243,46 +186,48 @@ void *mem_heap_alloc_block_internal (const size_t size)
   const size_t required_size = ((size + MEM_ALIGNMENT - 1) / MEM_ALIGNMENT) * MEM_ALIGNMENT;
   mem_heap_free_t *data_space_p = NULL;
 
-  VALGRIND_DEFINED_SPACE (&mem_heap.first, sizeof (mem_heap_free_t));
+  mem_heap_t *mem_heap = jerry_port_get_heap ();
```
zherczeg: I am worried about the performance cost of this change. Especially because the base address is required for all heap pointer compression and decompression. So I think we should keep the base address and other properties in global variables.

Perhaps the following API would have a smaller overhead:

    struct { uint8_t area[MEM_HEAP_AREA_SIZE] /* forced alignment by posix memalign or something */ } heap_area;

    // Jerry calls this only once. Returned heap is stored in a global variable
    heap_area* jerry_port_alloc_heap(size_t size, size_t alignment); 
    // Jerry calls this only once. Free heap.
    void jerry_port_free_heap(heap_area* area); 

This will also have a performance overhead, we should measure it first before we make a decision.

jerry-core/mem/mem-heap.c
```diff
@@ -555,8 +502,9 @@ mem_heap_compress_pointer (const void *pointer_p) /**< pointer to compress */
 {
   JERRY_ASSERT (pointer_p != NULL);
 
+  mem_heap_t *mem_heap = jerry_port_get_heap ();
   uintptr_t int_ptr = (uintptr_t) pointer_p;
-  const uintptr_t heap_start = (uintptr_t) &mem_heap;
+  const uintptr_t heap_start = (uintptr_t) mem_heap;
```
zherczeg: This is my main concern.

jerry-core/jerry.c
franc0is: Sorry, this part is stale. I had implemented this function myself, and realized that someone else had too.

jerry-core/mem/mem-heap.c
franc0is: You make a good point, but hopefully we can come up with a different solution than adding more statics.

My goal over time is to remove as many statics as possible as I would like to run more than one instance of the engine at a time.

In this particular case, I think we can achieve similar performance by forcing `jerry_port_get_heap()` to be inlined. Would you be ok with that @zherczeg ?

jerry-core/jerry-port.h
franc0is: ok!

jerry-core/mem/mem-heap.c
zherczeg: Technically, inlining should help (of course we need to prove it, I don't trust the compiler that much as in the past).

My main concern is actually not technical. I don't really understand how this function provides multiple heaps:

    mem_heap_t *jerry_port_get_heap(void)

There is no argument which tells which heap should be used. This patch adds some kind of context management API, but I feel it focuses only on one use case. Since we are in the middle of redesigning the API, we should carefully design this part as well.

What about an OpenGL like jerry_port_set_context()? This would update the appropriate global variables. Or maybe we could add an opaque context pointer, which would be nothing on embedded systems, but would be a pointer on others.

    #if EMBEDDED
    #define CONTEXT
    #else
    #define CONTEXT context *ctx,
    #endif

    retval some_jerry_function(CONTEXT
                                                arg1, ...

I know this would be quite a major change, but provides the most freedom of using Jerry API. I would lijke to hear the others opinion as well, @LaszloLango @akiss77 ?

zherczeg:
The handling of globals are reworked at the moment by patches such as #1034, #1219. This patch should be adapted by adding a new context model for this particular use case.

zherczeg:
The infrastructure for adding new heap models are available, but I think this PR is too old, and new ones should be opened for adding new heap models. Feel free to reopen it if necessary.

Issue #927, opened 2016-03-01
Add eviction mechanism to RegExp cache


LaszloLango:
Implement an eviction mechanism for RegExp cache, if the cache is full throw out an element and add the new bytecode into the cache. May be a FIFO algorithm would cause the less performance regression.

akiss77:
Solved by #947 and 6f536c79423ecca4c45897d6be555e08739ed71b

PR #947, opened 2016-03-09
Few improvements for RegExp

LaszloLango:
Added eviction mechanism to RegExp cache and small
refactoring. Fixed a bug when logging is enabled.

Related issue: #927

### DIFF
--- a/jerry-core/parser/regexp/re-compiler.c
+++ b/jerry-core/parser/regexp/re-compiler.c
@@ -445,23 +445,23 @@ re_parse_alternative (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context 
 } /* re_parse_alternative */
 
 static const re_compiled_code_t *re_cache[RE_CACHE_SIZE];
+static uint8_t re_cache_idx = RE_CACHE_SIZE;
 
 /**
  * Search for the given pattern in the RegExp cache
  *
- * @return compiled bytecode - if found
- *         NULL              - otherwise
+ * @return index of bytecode in cache - if found
+ *         RE_CACHE_SIZE              - otherwise
  */
-const re_compiled_code_t *
+static uint8_t
 re_find_bytecode_in_cache (ecma_string_t *pattern_str_p, /**< pattern string */
-                           uint16_t flags, /**< flags */
-                           uint32_t *idx) /**< [out] index */
+                           uint16_t flags) /**< flags */
 {
-  uint32_t free_idx = RE_CACHE_SIZE;
+  uint8_t free_idx = RE_CACHE_SIZE;
 
-  for (*idx = 0u; *idx < RE_CACHE_SIZE; (*idx)++)
+  for (uint8_t idx = 0u; idx < RE_CACHE_SIZE; idx++)
   {
-    const re_compiled_code_t *cached_bytecode_p = re_cache[*idx];
+    const re_compiled_code_t *cached_bytecode_p = re_cache[idx];
 
     if (cached_bytecode_p != NULL)
     {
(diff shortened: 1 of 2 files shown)


jerry-core/parser/regexp/re-compiler.c
```diff
@@ -533,90 +545,93 @@ re_compile_bytecode (const re_compiled_code_t **out_bytecode_p, /**< [out] point
...
-      JERRY_ASSERT (bc_ctx.block_start_p != NULL);
-      *out_bytecode_p = (re_compiled_code_t *) bc_ctx.block_start_p;
+    /* The RegExp bytecode contains at least a RE_OP_SAVE_AT_START opdoce, so it cannot be NULL. */
+    JERRY_ASSERT (bc_ctx.block_start_p != NULL);
+    *out_bytecode_p = (re_compiled_code_t *) bc_ctx.block_start_p;
 
-      if (cache_idx < RE_CACHE_SIZE)
+    if (cache_idx >= RE_CACHE_SIZE)
+    {
+      if (re_cache_idx < 1u)
```
akiss77: Actually, in these two `if`s, we are checking whether `cache_idx == RE_CACHE_SIZE` and `re_cache_idx == 0`, aren't we? (I mean, there is no way for `cache_idx` to be striclty larger that `RE_CACHE_SIZE`, and `re_cache_idx` being unsigned, it cannot have negative value.) I'd vote for the exact equality tests.

jerry-core/parser/regexp/re-compiler.c
```diff
@@ -452,12 +453,12 @@ static const re_compiled_code_t *re_cache[RE_CACHE_SIZE];
  * @return compiled bytecode - if found
  *         NULL              - otherwise
  */
-const re_compiled_code_t *
+static const re_compiled_code_t *
 re_find_bytecode_in_cache (ecma_string_t *pattern_str_p, /**< pattern string */
                            uint16_t flags, /**< flags */
-                           uint32_t *idx) /**< [out] index */
+                           uint16_t *idx) /**< [out] index */
```
akiss77: I've reviewed this function and its call site in `re_compile_bytecode` and I'm wondering whether the complexity of its signature is really needed. Wouldn't 
`static uin16_t re_find_bytecode_in_cache (ecma_string_t *pattern_str_p, uint16_t flags)`
suffice? 

With the following semantics: if the returned value equals `RE_CACHE_SIZE`, then the pattern was not found in the cache and also there is no free slot in the cache; if the returned value is a valid index and the corresponding slot in the cache is `NULL`, then the pattern was not found but index points to a free slot; finally, if the index is valid and the slot is non-NULL, then that's the bytecode we looked for.

This would eliminate the need for the out parameter, i.e., a pointer to a short int, which is quite often dereferenced in a loop in the body of the function.

jerry-core/parser/regexp/re-compiler.c
```diff
@@ -445,23 +445,23 @@ re_parse_alternative (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context
...
 /**
  * Search for the given pattern in the RegExp cache
  *
- * @return compiled bytecode - if found
- *         NULL              - otherwise
+ * @return index of bytecode in cache - if found
+ *         RE_CACHE_SIZE              - otherwise
  */
-const re_compiled_code_t *
+static uint16_t
```
zherczeg: uint32_t

zherczeg: I see now that re_cache_idx is uint16_t. If we want to force a small number we could use uint8_t. Anyway I am even happy with uint32_t as well.
Otherwise LGTM.

LaszloLango:
@zherczeg, updated. May I land this?

zherczeg:
LGTM

PR #948, opened 2016-03-09
Storing byte code size in the byte code header.

zherczeg:
This reduces the memory consumption, because the new allocator uses less memory if
the size as available when a block is freed. Snapshot generation is also simplified.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -741,16 +741,14 @@ typedef uintptr_t ecma_external_pointer_t;
   */
 typedef struct
 {
-  uint16_t status_flags;            /**< various status flags */
+  uint16_t size;                    /**< real size >> MEM_ALIGNMENT_LOG */
+  uint16_t refs;                    /**< reference counter for the byte code */
+  uint16_t status_flags;            /**< various status flags:
+                                      *    CBC_CODE_FLAGS_FUNCTION flag tells whether
+                                      *    the byte code is function or regular expression.
+                                      *    If function, the other flags must be CBC_CODE_FLAGS...
+                                      *    If regexp, the other flags must be RE_FLAG... */
 } ecma_compiled_code_t;
-
-/**
- * Shift value for byte code reference counting.
- * The last 10 bit of the first uint16_t value
- * of compact byte code or regexp byte code
- * is reserved for reference counting.
- */
-#define ECMA_BYTECODE_REF_SHIFT 6
 
 /**
  * @}
--- a/jerry-core/ecma/base/ecma-helpers.c
+++ b/jerry-core/ecma/base/ecma-helpers.c
@@ -1316,12 +1316,12 @@ void
 ecma_bytecode_ref (ecma_compiled_code_t *bytecode_p) /**< byte code pointer */
 {
   /* Abort program if maximum reference number is reached. */
-  if ((bytecode_p->status_flags >> ECMA_BYTECODE_REF_SHIFT) >= 0x3ff)
+  if (bytecode_p->refs >= UINT16_MAX)
   {
     jerry_fatal (ERR_REF_COUNT_LIMIT);
   }
 
-  bytecode_p->status_flags = (uint16_t) (bytecode_p->status_flags + (1u << ECMA_BYTECODE_REF_SHIFT));
+  bytecode_p->refs++;
 } /* ecma_bytecode_ref */
 
 /**
@@ -1331,11 +1331,11 @@ ecma_bytecode_ref (ecma_compiled_code_t *bytecode_p) /**< byte code pointer */
 void
 ecma_bytecode_deref (ecma_compiled_code_t *bytecode_p) /**< byte code pointer */
 {
-  JERRY_ASSERT ((bytecode_p->status_flags >> ECMA_BYTECODE_REF_SHIFT) > 0);
+  JERRY_ASSERT (bytecode_p->refs > 0);
 
-  bytecode_p->status_flags = (uint16_t) (bytecode_p->status_flags - (1u << ECMA_BYTECODE_REF_SHIFT));
+  bytecode_p->refs--;
 
-  if (bytecode_p->status_flags >= (1u << ECMA_BYTECODE_REF_SHIFT))
+  if (bytecode_p->refs > 0)
   {
     /* Non-zero reference counter. */
     return;
@@ -1388,7 +1388,8 @@ ecma_bytecode_deref (ecma_compiled_code_t *bytecode_p) /**< byte code pointer */
 #endif /* !CONFIG_ECMA_COMPACT_PROFILE_DISABLE_REGEXP_BUILTIN */
   }
 
-  mem_heap_free_block_size_stored (bytecode_p);
+  mem_heap_free_block (bytecode_p,
+                       ((size_t) bytecode_p->size) << MEM_ALIGNMENT_LOG);
 } /* ecma_bytecode_deref */
 
 /**
--- a/jerry-core/jerry-snapshot.h
+++ b/jerry-core/jerry-snapshot.h
@@ -24,27 +24,19 @@
  */
 typedef struct
 {
-  uint32_t last_compiled_code_offset; /**< offset of the last compiled code */
+  /* The size of this structure is recommended to be divisible by
+   * MEM_ALIGNMENT. Otherwise some bytes after the header are wasted. */
+  uint32_t version; /**< version number */
+  uint32_t lit_table_offset; /**< offset of the literal table */
   uint32_t lit_table_size; /**< size of literal table */
-  __extension__ uint32_t is_run_global : 1; /**< flag, indicating whether the snapshot
-                                             *   was dumped as 'Global scope'-mode code (true)
-                                             *   or as eval-mode code (false) */
+  uint32_t is_run_global; /**< flag, indicating whether the snapshot
+                            *   was dumped as 'Global scope'-mode code (true)
+                            *   or as eval-mode code (false) */
 } jerry_snapshot_header_t;
 
 /**
  * Jerry snapshot format version
  */
-#define JERRY_SNAPSHOT_VERSION (3u)
-
-#ifdef JERRY_ENABLE_SNAPSHOT_SAVE
-
-/* Snapshot support functions */
-
-extern bool snapshot_report_byte_code_compilation;
-
-extern void
-snapshot_add_compiled_code (ecma_compiled_code_t *, const uint8_t *, uint32_t);
-
-#endif /* JERRY_ENABLE_SNAPSHOT_SAVE */
+#define JERRY_SNAPSHOT_VERSION (4u)
 
 #endif /* !JERRY_SNAPSHOT_H */
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -1832,41 +1832,24 @@ jerry_register_external_magic_strings (const jerry_api_char_ptr_t *ex_str_items,
 #ifdef JERRY_ENABLE_SNAPSHOT_SAVE
 
 /*
- * Tells whether snapshot taking is in progress.
- */
-bool snapshot_report_byte_code_compilation = false;
-
-/*
- * Mapping snapshot to offset.
- */
-typedef struct
-{
-  mem_cpointer_t next_cp;
-  mem_cpointer_t compiled_code_cp;
-  uint16_t offset;
-} compiled_code_map_entry_t;
-
-/*
  * Variables required to take a snapshot.
  */
-static size_t snapshot_buffer_write_offset;
-static size_t snapshot_last_compiled_code_offset;
 static bool snapshot_error_occured;
+static size_t snapshot_buffer_write_offset;
 static uint8_t *snapshot_buffer_p;
 static size_t snapshot_buffer_size;
-static compiled_code_map_entry_t *snapshot_map_entries_p;
 
 /**
  * Snapshot callback for byte codes.
+ *
+ * @return start offset
  */
-void
-snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled code */
-                            const uint8_t *regexp_pattern, /**< regular expression pattern */
-                            uint32_t size) /**< compiled code or regular expression size */
+static uint16_t
+snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p) /**< compiled code */
 {
   if (snapshot_error_occured)
   {
-    return;
+    return 0;
   }
 
   JERRY_ASSERT ((snapshot_buffer_write_offset & (MEM_ALIGNMENT - 1)) == 0);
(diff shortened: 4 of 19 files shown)


jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1316,12 +1316,12 @@ void
 ecma_bytecode_ref (ecma_compiled_code_t *bytecode_p) /**< byte code pointer */
 {
   /* Abort program if maximum reference number is reached. */
-  if ((bytecode_p->status_flags >> ECMA_BYTECODE_REF_SHIFT) >= 0x3ff)
+  if (bytecode_p->refs >= 0xffff)
```
franc0is: use `UINT16_MAX`

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -741,18 +741,12 @@ typedef uintptr_t ecma_external_pointer_t;
   */
 typedef struct
 {
-  uint16_t status_flags;            /**< various status flags */
+  uint16_t size;                    /**< real size >> MEM_ALIGNMENT_LOG */
+  uint16_t refs;                    /**< reference counter for the byte code */
+  uint16_t status_flags;            /**< various status flags. */
```
franc0is: Are the flags used for anything after you broke out the references? If not, why keep the field at all?

jerry-core/jerry.c
```diff
@@ -1845,81 +1828,112 @@ snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled
...
+    return 0;
   }
 
-  snapshot_last_compiled_code_offset = snapshot_buffer_write_offset;
+  uint16_t start_offset = (uint16_t) (snapshot_buffer_write_offset >> MEM_ALIGNMENT_LOG);
+  ecma_compiled_code_t *copied_compiled_code_p;
 
-  compiled_code_map_entry_t *new_entry;
-  new_entry = (compiled_code_map_entry_t *) mem_heap_alloc_block (sizeof (compiled_code_map_entry_t));
+  copied_compiled_code_p = (ecma_compiled_code_t *) (snapshot_buffer_p + snapshot_buffer_write_offset);
```
franc0is: declare and define on a single line

jerry-core/jerry.c
```diff
@@ -1845,81 +1828,112 @@ snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled
...
-      return;
     }
 
-    data_p = regexp_pattern;
+    MEM_FINALIZE_LOCAL_ARRAY (buffer_p);
+
+    snapshot_buffer_write_offset = JERRY_ALIGNUP (snapshot_buffer_write_offset, MEM_ALIGNMENT);
+
+    /* Regexp character size is stored in refs. */
+    copied_compiled_code_p->refs = (uint16_t) pattern_size;
```
franc0is: This seems to rely on the re_compiled_code_t and ecma_compiled_code_t structs to stay stable relative to each other. Why not create a new re_compiled_code_t variable pointing at the same memory instead?

jerry-core/jerry.c
```diff
@@ -2079,64 +2047,56 @@ jerry_parse_and_save_snapshot (const jerry_api_char_t *source_p, /**< script sou
...
+    return 0;
+  }
 
-      jrt_write_to_buffer_by_offset (buffer_p,
-                                     buffer_size,
-                                     &header_offset,
-                                     &header,
-                                     sizeof (header));
+  jerry_snapshot_header_t header;
+  header.version = JERRY_SNAPSHOT_VERSION;
```
franc0is: Any reason why you wouldn't prefer struct initilization here?
```
jerry_snapshot_header_t header = {
  .version = JERRY_SNAPSHOT_VERSION,
  ...
};
```

It has the advantage that new fields added to the structure are zeroed out rather than left uninitialized.

jerry-core/jerry.c
```diff
@@ -2213,21 +2169,25 @@ snapshot_load_compiled_code (const uint8_t *snapshot_data_p, /**< snapshot data
     header_size = sizeof (cbc_uint8_arguments_t);
   }
 
-  if (copy_bytecode)
+  if (copy_bytecode
+      || (header_size + (literal_end * sizeof (uint16_t)) + 8 > code_size))
```
franc0is: can something more descriptive be used instead of `+ 8` ?

jerry-core/parser/js/byte-code.h
```diff
@@ -617,6 +617,8 @@
  */
 typedef struct
 {
+  uint16_t size;                    /**< real size >> MEM_ALIGNMENT_LOG */
```
franc0is: why not embed an `ecma_compiled_code_t` in these structs instead of copying the structure?

jerry-core/parser/js/js-parser-limits.h
```diff
@@ -54,7 +54,7 @@
 /* Maximum code size.
  * Limit: 16777215. Recommended: 65535, 16777215. */
 #ifndef PARSER_MAXIMUM_CODE_SIZE
-#define PARSER_MAXIMUM_CODE_SIZE 16777215
+#define PARSER_MAXIMUM_CODE_SIZE (65535 << (MEM_ALIGNMENT_LOG))
```
franc0is: Looks suspiciously like UINT16_MAX

jerry-core/ecma/base/ecma-globals.h
zherczeg: The flags contains whether the byte code is function or regexp, so it is still important. The other flag bits depends on this.

jerry-core/jerry.c
zherczeg: Would break the 120 character maximum line length. Other kind of line breaking is also not allowed so this is our only option currently.

jerry-core/jerry.c
zherczeg: I don't understand this.

jerry-core/jerry.c
zherczeg: Unfortunately you have to put { to the next line and I don't like that form. There cannot be any fields which are not initialized of this structure.

jerry-core/parser/js/byte-code.h
zherczeg: This could be done in a follow-up patch. This patch is big enough, and I am sure such a change will require several line changes.

jerry-core/parser/js/js-parser-limits.h
zherczeg: In this file numbers are explicit, so I prefer it this way.

LaszloLango:
@zherczeg, please rebase to the current master

zherczeg:
Raspberry Pi2 results:

                      Benchmark |          RSS(+ is better) |                   Perf(+ is better)
                     3d-cube.js |   80k ->   76k :  +5.000% |   2.530s ->   2.552s :  -0.870%  (+-0.892%) : [≈]
         access-binary-trees.js |   80k ->   76k :  +5.000% |   1.334s ->   1.348s :  -1.049%  (+-1.098%) : [≈]
             access-fannkuch.js |   40k ->   32k : +20.000% |   7.546s ->   7.596s :  -0.663%  (+-1.132%) : [≈]
                access-nbody.js |   40k ->   36k : +10.000% |   2.914s ->   2.920s :  -0.206%  (+-0.806%) : [≈]
    bitops-3bit-bits-in-byte.js |   28k ->   24k : +14.286% |   1.666s ->   1.684s :  -1.080%  (+-0.963%) : [-]
         bitops-bits-in-byte.js |   28k ->   24k : +14.286% |   2.420s ->   2.472s :  -2.149%  (+-0.932%) : [-]
          bitops-bitwise-and.js |   32k ->   24k : +25.000% |   3.290s ->   3.260s :  +0.912%  (+-1.318%) : [≈]
          bitops-nsieve-bits.js |  152k ->  148k :  +2.632% |  20.138s ->  20.132s :  +0.030%  (+-1.035%) : [≈]
       controlflow-recursive.js |  140k ->  136k :  +2.857% |   0.926s ->   0.932s :  -0.648%  (+-1.578%) : [≈]
                  crypto-aes.js |  108k ->  104k :  +3.704% |   3.544s ->   3.502s :  +1.185%  (+-0.575%) : [+]
                  crypto-md5.js |  176k ->  164k :  +6.818% |   7.128s ->   7.180s :  -0.730%  (+-1.304%) : [≈]
                 crypto-sha1.js |  124k ->  116k :  +6.452% |   3.884s ->   3.890s :  -0.154%  (+-0.712%) : [≈]
           date-format-tofte.js |   64k ->   60k :  +6.250% |   2.152s ->   2.168s :  -0.743%  (+-1.488%) : [≈]
           date-format-xparb.js |   56k ->   52k :  +7.143% |   0.856s ->   0.848s :  +0.935%  (+-2.399%) : [≈]
                 math-cordic.js |   36k ->   32k : +11.111% |   2.898s ->   2.880s :  +0.621%  (+-1.360%) : [≈]
           math-partial-sums.js |   32k ->   24k : +25.000% |   1.584s ->   1.574s :  +0.631%  (+-1.004%) : [≈]
          math-spectral-norm.js |   36k ->   32k : +11.111% |   1.430s ->   1.432s :  -0.140%  (+-0.644%) : [≈]
               string-base64.js |  156k ->  152k :  +2.564% |  30.992s ->  30.930s :  +0.200%  (+-0.876%) : [≈]
                string-fasta.js |   40k ->   36k : +10.000% |   3.504s ->   3.386s :  +3.368%  (+-1.091%) : [+]
                Geometric mean: |    RSS reduction: 10.232% |                Speed up: -0.022% (+-0.273%) : [≈]

## Binary sizes ##
master: 183,388
branch: 182,948

jerry-core/parser/js/byte-code.h
zherczeg: Anyway, this is a good idea.

LaszloLango:
I'd use the `'u'`suffix for number constants for unsigned variables and return types. LGTM besides that. Nice improvement.

jerry-core/jerry-snapshot.h
```diff
@@ -24,27 +24,17 @@
...
 typedef struct
 {
-  uint32_t last_compiled_code_offset; /**< offset of the last compiled code */
+  uint32_t version; /**< version number */
+  uint32_t lit_table_offset; /**< offset of the literal table */
   uint32_t lit_table_size; /**< size of literal table */
-  __extension__ uint32_t is_run_global : 1; /**< flag, indicating whether the snapshot
-                                             *   was dumped as 'Global scope'-mode code (true)
-                                             *   or as eval-mode code (false) */
+  uint8_t is_run_global; /**< flag, indicating whether the snapshot
```
akiss77: Why not simply use the type `bool` here? I don't see any alignment requirements or need for multiple flags that would necessitate an integer type with fixed bit-length.

jerry-core/jerry.c
```diff
@@ -2023,42 +2019,19 @@ jerry_parse_and_save_snapshot (const jerry_api_char_t *source_p, /**< script sou
...
-  snapshot_report_byte_code_compilation = true;
   jerry_api_object_t *error_obj_p = NULL;
 
   if (is_for_global)
   {
-    parse_status = parser_parse_script (source_p, source_size, &bytecode_data_p, &error_obj_p);
+    parse_status = parser_parse_script (source_p,
+                                        source_size,
+                                        &bytecode_data_p,
+                                        &error_obj_p);
```
akiss77: Did anything else change here other than whitespace? If no, why the change then? The original code still seemed to fit within the 120 chars-per-line limit.

jerry-core/jerry-snapshot.h
zherczeg: I don't like using bools in structures because its storage size is unclear to me.

jerry-core/jerry.c
zherczeg: Because parser_parse_eval below uses this style, and for me this looked ugly.

jerry-core/jrt/jrt.h
```diff
@@ -204,7 +204,7 @@ extern void __noreturn jerry_fatal (jerry_fatal_code_t);
  *
  * Returns minimum positive value, that divides @a alignment and is more than or equal to @a value
  */
-#define JERRY_ALIGNUP(value, alignment) ((alignment) * (((value) + (alignment) - 1) / (alignment)))
+#define JERRY_ALIGNUP(value, alignment) (((value) + ((alignment) - 1)) & ~((alignment) - 1))
```
akiss77: This works only if `alignment` is a power of two, while original version worked for any positive integer number. True enough, power-of-2 alignments are used in practice, but then please update the doc accordingly.

jerry-core/parser/js/common.h
```diff
@@ -52,11 +52,11 @@
...
 
-#define PARSER_MALLOC(size) mem_heap_alloc_block_store_size (size)
-#define PARSER_FREE(ptr) mem_heap_free_block_size_stored ((void *) ptr)
+#define PARSER_MALLOC(size) mem_heap_alloc_block (size)
+#define PARSER_FREE(ptr, size) mem_heap_free_block ((ptr), (size))
 
-#define PARSER_MALLOC_LOCAL(size) mem_heap_alloc_block_store_size (size)
-#define PARSER_FREE_LOCAL(ptr) mem_heap_free_block_size_stored (ptr)
+#define PARSER_MALLOC_LOCAL(size) mem_heap_alloc_block (size)
+#define PARSER_FREE_LOCAL(ptr, size) mem_heap_free_block ((ptr), (size))
```
akiss77: Do we still need to keep these macros?
- They seem to duplicate each other.
- Their only valid uses seem to be in js-parser-mem.c.
- This patch already replaces them at some points.

I'd vote for replacing them with calls to `parser__{malloc,free}[_local]` or `mem_heap_{alloc,free}_block`, depending on the use case.

jerry-core/jerry.c
akiss77: _This_ style takes too much vertical space from my screen for a single statement but _de gustibus non est disputandum_.

jerry-core/jerry-snapshot.h
akiss77: First I thought that exact size does not matter here since the struct is not used that often. Now I realize that it does matter, since this header becomes part of the snapshot buffer and will be written to disk, so we will need to know the exact offsets of fields and data, etc. That's really a point against `bool`.

However, there might still be some issues with `uint_8` and also with the way the snapshot buffer is constructed (in jerry.c). The size of a struct may differ from machine to machine, AFAIK it's up to the compiler to define alignments, padding, whatnot. Would this struct be packed, then its size would be 13 bytes, on a regular 32 bit machine it would most probably be 16 bytes in unpacked form, but actually there is no guarantee for that (and there may be exotic architectures where we might get a different result). Making the last field `uint32_t` makes things a bit more portable IMHO.

As for my just-realized concerns with jerry.c, I'll make a comment there.

jerry-core/jerry.c
```diff
@@ -2023,42 +2019,19 @@ jerry_parse_and_save_snapshot (const jerry_api_char_t *source_p, /**< script sou
   jsp_status_t parse_status;
   ecma_compiled_code_t *bytecode_data_p;
 
-  snapshot_buffer_write_offset = 0;
-  snapshot_last_compiled_code_offset = 0;
+  snapshot_buffer_write_offset = JERRY_ALIGNUP (sizeof (jerry_snapshot_header_t),
+                                                MEM_ALIGNMENT);
```
akiss77: As we construct the snapshot buffer, we seem to use alignments based on `MEM_ALIGNMENT` and `MEM_ALIGNMENT_LOG` (scattered all around the code, not just here). Do we realize that changing those defines invalidates all our snapshots? I mean, there is a `JERRY_SNAPSHOT_VERSION` macro which is incremented every time there is a breaking change to the snapshot format (just like this patch). Memory alignment seems only to be a memory configuration parameter, I'm not sure it should have any effect on the snapshot format.

jerry-core/jerry.c
```diff
@@ -1841,81 +1824,117 @@ snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled
...
-  snapshot_last_compiled_code_offset = snapshot_buffer_write_offset;
+  uint16_t start_offset = (uint16_t) (snapshot_buffer_write_offset >> MEM_ALIGNMENT_LOG);
+  ecma_compiled_code_t *copied_compiled_code_p;
 
-  compiled_code_map_entry_t *new_entry;
-  new_entry = (compiled_code_map_entry_t *) mem_heap_alloc_block (sizeof (compiled_code_map_entry_t));
+  copied_compiled_code_p = (ecma_compiled_code_t *) (snapshot_buffer_p + snapshot_buffer_write_offset);
 
-  if (new_entry == NULL)
+  if (!(compiled_code_p->status_flags & CBC_CODE_FLAGS_FUNCTION))
```
akiss77: The issue that @franc0is also mentioned is that within the body of this `if` we know that we are actually not dealing with `ecma_compiled_code_t` but `re_compiled_code_t`. It would be much comprehendible (and even safer) to access whatever fields we access as fields of `re_compiled_code_t` (by casting `compiled_code_p`).

Moreover, it should be noted at the struct declarations that the first 3 fields _must_ always be kept in sync.

jerry-core/ecma/base/ecma-globals.h
akiss77: The doc of this field should be amended with some short description of the valid flags (e.g., "various status flags (CBC_CODE_FLAGS_xxx)")

jerry-core/jerry.c
zherczeg: This simplifies things. There are much more complex things we should worry about.

jerry-core/jerry.c
zherczeg: No, saving a regex is not saving its byte code. We are only saving the ecma_compiled_code_t as a header and the pattern string. So we are not dealing with re_compiled_code_t.

zherczeg:
Issues fixed.

jerry-core/jerry.c
zherczeg: Basically all offsets in the snapshot are shifted with MEM_ALIGNMENT_LOG. We could support different alignments and other endian supports, but I would prefer that in a separate patch. I think we should evaluate the overhead first (whether it is worth it).

jerry-core/jerry.c
akiss77: At line 1847 we have `((re_compiled_code_t *) compiled_code_p)->pattern_cp`, which hints that `compiled_code_p` is a regex compiled code. However, it's true that `copied_compiled_code_p` is an `ecma_compiled_code_t`. Since it seems that `compiled_code_p` is only used once, leave it as is.

jerry-core/parser/js/common.c
```diff
@@ -165,7 +165,7 @@ util_free_literal (lexer_literal_t *literal_p) /**< literal */
   {
     if (!(literal_p->status_flags & LEXER_FLAG_SOURCE_PTR))
     {
-      PARSER_FREE ((uint8_t *) literal_p->u.char_p);
+      mem_heap_free_block_size_stored ((uint8_t *) literal_p->u.char_p);
```
akiss77: Better cast this to `void *`

jerry-core/parser/js/js-parser.c
```diff
@@ -105,7 +105,7 @@ parser_compute_indicies (parser_context_t *context_p, /**< context */
 
         if (!(literal_p->status_flags & LEXER_FLAG_SOURCE_PTR))
         {
-          PARSER_FREE ((uint8_t *) char_p);
+          mem_heap_free_block_size_stored ((uint8_t *) char_p);
```
akiss77: ditto

akiss77:
After the casts are fixed, LGTM (FWIW)

jerry-core/jerry.c
franc0is: It's fine if this is just not acceptable style for this project. I was only pointing it out because I recently had to fix a pretty nasty regression where a field was added to a struct that was initialized field by field as you do here and ended up uninitialized:

```
typedef struct {
  int a;
  int b;
  int c;
} foo_t;
...
foo_t bar;
// when this code was written, foo_t only had two fields, a and b.
bar.a = 1;
bar.b = 2;
...
// bar.c is uninitialized
```

zherczeg:
I hope I fixed everything.

LaszloLango:
LGTM

PR #950, opened 2016-03-09
Initialize ecma gc state on jerry init

franc0is:
While trying to run `jerry_init` and `jerry_cleanup` a few times in succession within the same application, I noticed that the memory management code would get very confused.

I narrowed it down to the fact that `ecma_gc_init` is not called anywhere (!!!) and thus that state is stale on the second invocation of the engine.


zherczeg:
LGTM

LaszloLango:
LGTM

akiss77:
(You got LGTMs here. Could you please rebase the PR to master so that it can be merged in neatly?)

franc0is:
Rebased.

PR #951, opened 2016-03-09
Check unit tests with vera++ & fix reported issues

akiss77:
Cppcheck has already been used for checking unit test sources
statically but vera++ style-checking of unit tests was left out,
most probably only by mistake. This patch adds unit tests to the
set of style-checked sources and also fixes the stlye issues that
were present in the test code.

(Extra: since the project is pure C now, not trying to collect
*.cpp files anymore.)


LaszloLango:
LGTM

zherczeg:
LGTM

PR #952, opened 2016-03-09
Unify the naming scheme of all header defines

akiss77:
There have been several renamings of header files but the
ifndef/define/endif preprocessor directives did not always follow
the file name changes (or perhaps never followed a unified naming
scheme). This patch gets all headers aligned with the style of the
majority.

### DIFF
--- a/jerry-core/ecma/base/ecma-alloc.h
+++ b/jerry-core/ecma/base/ecma-alloc.h
@@ -20,8 +20,8 @@
  * @{
  */
 
-#ifndef JERRY_ECMA_ALLOC_H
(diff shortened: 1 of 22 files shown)


LaszloLango:
LGTM

zherczeg:
LGTM. How did you find them? Could vera do this automatically?

akiss77:
I was familiarizing myself with the literals and spotted a mismatch in one of the headers. Then I used all the super powers of my midnight commander to sweep through all the headers of the code base :)
(I most definitely don't plan to write tcl code for vera... If that's necessary, I leave that for someone else.)

PR #953, opened 2016-03-09
Fix terminology of snapshot saving in literals

akiss77:
In case of literals, snapshot saving is incorrectly named dumping.
Elsewhere in the code, 'dump' functions output debug data (even
literals have a `lit_dump_literals` debug function). To help
distinction and to align terminologies, snapshot saving functions
of literals are also re-named 'save'.

### DIFF
--- a/jerry-core/jerry-snapshot.h
+++ b/jerry-core/jerry-snapshot.h
@@ -30,7 +30,7 @@ typedef struct
   uint32_t lit_table_offset; /**< offset of the literal table */
   uint32_t lit_table_size; /**< size of literal table */
   uint32_t is_run_global; /**< flag, indicating whether the snapshot
-                            *   was dumped as 'Global scope'-mode code (true)
+                            *   was saved as 'Global scope'-mode code (true)
                             *   or as eval-mode code (false) */
 } jerry_snapshot_header_t;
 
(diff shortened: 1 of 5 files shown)


zherczeg:
I am ok with this patch, but could it wait until #948 landed, and do a rebase after?

akiss77:
@zherczeg What's the time frame for that? If that takes longer, why not the other way around?

LaszloLango:
@akiss77, the term `dump_snapshot` is used in `main-unix.c` as well.

akiss77:
@LaszloLango You are right, dump is even used in the name of a command line switch. (Which is a pitty, since the public api calls it 'save', too.) I've missed that, focusing only on jerry-core. 

I'm not sure how to deal with that since I know at least about iot.js using the command line switch. (Ofc, we can decide to rename everything here, including the command line switch, and adapt iot-js to that when it gets rebased to latest jerry.) Opinions?

LaszloLango:
I think we should rename everywhere, but the http://samsung.github.io/jerryscript/benchmark/benchmark.html?show=sum site use snapshot somehow. It may use only the load. AFAIK @wateret maintains the benchmark.

akiss77:
Updated (changed terminology in main-unix.cpp as well).

akiss77:
Now that #948 got landed, what about this PR? (Rebased, resolved conflicts, pushed.)

LaszloLango:
LGTM

zherczeg:
LGTM

PR #954, opened 2016-03-10
Remove support for per-instruction memory statistics

akiss77:
The new CBC interpreter does not support it anymore.

### DIFF
--- a/jerry-core/jerry.h
+++ b/jerry-core/jerry.h
@@ -32,22 +32,20 @@ extern "C"
  * @{
  */
 
-#define JERRY_FLAG_EMPTY                  (0u)      /**< empty flag set */
-#define JERRY_FLAG_SHOW_OPCODES           (1u << 0) /**< dump byte-code to stdout after parse */
-#define JERRY_FLAG_MEM_STATS              (1u << 1) /**< dump memory statistics */
-#define JERRY_FLAG_MEM_STATS_PER_OPCODE   (1u << 2) /**< dump per-instruction memory statistics during execution
-                                                     *   (in the mode full GC is performed after execution of each
-                                                     *   opcode handler) */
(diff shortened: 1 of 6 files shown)


jerry-core/jerry.h
```diff
@@ -36,8 +36,7 @@ extern "C"
 #define JERRY_FLAG_SHOW_OPCODES           (1u << 0) /**< dump byte-code to stdout after parse */
 #define JERRY_FLAG_MEM_STATS              (1u << 1) /**< dump memory statistics */
 #define JERRY_FLAG_MEM_STATS_PER_OPCODE   (1u << 2) /**< dump per-instruction memory statistics during execution
-                                                     *   (in the mode full GC is performed after execution of each
-                                                     *   opcode handler) */
+                                                     *   FIXME: Remove. */
```
LaszloLango: Why only comment?

akiss77: I wasn't sure whether I should renumber all unrelated flags or leave an unused hole in the bits. But I'd go for the renumbering then.

LaszloLango: Could you also change the style of the flags to be an enum?

akiss77: Note that these are bit flags that can be combined (or'd). Do we use enums for such constructs? If we'd like to go for a change then bitfield structs seem a better fit to me. If.

LaszloLango: See `re_flags_t` and `ecma_property_flags_t`

zherczeg: I also prefer enum. see vm/vm.h
LGTM after that

akiss77:
Updated according to the reviews.

LaszloLango:
LGTM

zherczeg:
LGTM

PR #957, opened 2016-03-10
Add jerry_port_abort API

franc0is:
Calling `abort()` is making assumptions about what the target wants to happen on fatal errors. In fact, I would argue the stdlib definition of `abort` doesn't make much sense on some platforms.

### DIFF
--- a/jerry-core/jerry-port.c
+++ b/jerry-core/jerry-port.c
@@ -15,6 +15,7 @@
 
 #include "jerry-port.h"
 #include <stdarg.h>
+#include <stdlib.h>
 
 /**
  * Provide log message to filestream implementation for the engine.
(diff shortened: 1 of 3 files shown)


seanshpark:
@franc0is , I had similar problem with `abort()`. 
what do you think about moving system related codes like this code.
```
+void jerry_port_abort (void)
 +{
 +  abort ();
 +} /* jerry_port_abort */
```
outside of jerry and let users implement them as the system works?

franc0is:
In one of my other PRs, I moved the jerry-port code to the "target"
directory so it can be implemented by the port.

Additionally, you can always provide those symbols elsewhere in your system
once my PR lands (hopefully!) and the default port code is no longer
included for every build.

Does that work for you?

seanshpark:
@franc0is , sounds nice :)

franc0is:
Superseded by #1015

PR #961, opened 2016-03-11
Use code unit instead of code point

LaszloLango:

### DIFF
--- a/jerry-core/lit/lit-char-helpers.c
+++ b/jerry-core/lit/lit-char-helpers.c
@@ -1,4 +1,5 @@
-/* Copyright 2015 Samsung Electronics Co., Ltd.
+/* Copyright 2015-2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


jerry-core/lit/lit-char-helpers.c
```diff
@@ -289,32 +289,32 @@ lit_char_hex_to_int (ecma_char_t c) /**< code unit, corresponding to
  * @return true if decoding was successful, false otherwise
  */
 bool
-lit_read_code_point_from_hex (lit_utf8_byte_t *buf_p, /**< buffer with characters */
-                              lit_utf8_size_t number_of_characters, /**< number of characters to be read */
-                              lit_code_point_t *out_code_point_p) /**< [out] decoded result */
+lit_read_code_unit_from_hex (lit_utf8_byte_t *buf_p, /**< buffer with characters */
```
akiss77: The doc still refers to code points.

jerry-core/lit/lit-char-helpers.c
```diff
@@ -289,32 +289,32 @@ lit_char_hex_to_int (ecma_char_t c) /**< code unit, corresponding to
...
+                             lit_utf8_size_t number_of_characters, /**< number of characters to be read */
+                             ecma_char_ptr_t out_unit_point_p) /**< [out] decoded result */
 {
-  lit_code_point_t code_point = 0;
+  ecma_char_t code_unit = LIT_BYTE_NULL;
```
akiss77: This seems to be a bit of an abuse of `LIT_BYTE_NULL`. Its documentation in lit-strings.h states that it's an utf-8 string end marker. Too bad that in the same header, only `LIT_UNICODE_CODE_POINT_NULL` is defined but no `LIT_UTF16_CODE_UNIT_NULL`. What about adding it (and making use of it here)?

jerry-core/parser/regexp/re-parser.h
```diff
@@ -78,7 +78,7 @@ typedef enum
 /**
  * Undefined character (out of the range of the codeunit)
  */
-#define RE_CHAR_UNDEF 0xFFFFFFFF
+#define RE_CHAR_UNDEF 0xFFFF
```
akiss77: Is this correct? The code unit type seems to cover the whole 16 bits. This change would make the 16bit 0xFFFF a special invalid value. The original code had a 32 bit value that was indeed out of the range of the code unit type, as the doc states.

akiss77: Let me change my review. This can be correct. According to the Unicode standard (http://www.unicode.org/versions/Unicode5.2.0/ch16.pdf, section 16.7):

> Noncharacters: U+FFFE, U+FFFF, and Others

> U+FFFF and U+10FFFF. These two noncharacter code points have the attribute of being
associated with the largest code unit values  for particular Unicode encoding forms. In
UTF-16, U+FFFF is associated with the largest 16-bit code unit value, FFFF_16. U+10FFFF is associated with the largest legal UTF-32 32-bit code unit value, 10FFFF_16. This attribute
renders these two noncharacter code points useful for internal purposes as sentinels. For
example, they might be used to indicate the end of a list, to represent a value in an index
guaranteed to be higher than any valid character value, and so on.

So, we can use this as a marker. But why use it only here and why not move this definition also to lit-strings.h?

jerry-core/lit/lit-char-helpers.c
akiss77: With the help of the non-character markers (as in my previous comment), couldn't we just return a code unit from this function and compare it to U+FFFF to see whether we succeeded or not? (I'm really against out parameters given with pointers, so prefer to replace them wherever possible.)

jerry-core/parser/regexp/re-parser.h
LaszloLango: Because only the RegExp use it :) But I'm not against to move this define to `lit-strings.h` as `LIT_UNDEFINED_CODE_UNIT` or `LIT_UNDEF_CHAR`. But this value has already a define in lit-strings.h: `#define LIT_UTF16_CODE_UNIT_MAX (0xFFFF)`
If we remove `RE_CHAR_UNDEF` and use `LIT_UTF16_CODE_UNIT_MAX`, then the code of RegExp parser will be less readable.

jerry-core/lit/lit-char-helpers.c
LaszloLango: No. The helper must be able to read `"FFFF"` and does not have to deal with it a valid code unit or not. Currently it might work on the test suite but this change would be dangerous.

jerry-core/parser/regexp/re-parser.h
akiss77: We should certainly not use `LIT_UTF16_CODE_UNIT_MAX` in place of `RE_CHAR_UNDEF`, since -- as its name suggests -- it is a type domain limit. We might have a `#define LIT_CHAR_NONCHAR ((ecma_char_t) 0xFFFF)` -- perhaps in lit-char-helpers.h instead of the previously suggested lit-strings.h (now seems to fit there better).

jerry-core/lit/lit-char-helpers.c
akiss77: Sorry, for keeping on amending my own reviews, but I just had a closer look at lit-char-helpers.h and found `LIT_CHAR_NULL` therein. Seems to fit here perfectly.

jerry-core/lit/lit-char-helpers.c
akiss77: I just sweeped through the call sites of this function, and it seems that only three functions call it with `number_of_characters` being 4: `ecma_builtin_json_parse_string`, `re_parse_char_class`, and `re_parse_next_token`. (Where the param is 2, there can be no possible way of getting a valid value of 0xFFFF.) Also, in a valid json, there must not be a U+FFFF according to the unicode specs. Where I'm not completely sure is the regexes. :/ While googling around I've seen some artificial examples with regex character classes where the upper limit was `\uFFFF`. What I couldn't find out with certainty is whether they are valid examples.

LaszloLango:
@akiss77, @zherczeg, updated. Please check.

jerry-core/lit/lit-char-helpers.c
zherczeg: JS supports all characters between 0-0xffff. JS was not designed to be UTF compatible, every character is a two byte number.

LaszloLango:
@zherczeg, related discussion: https://github.com/Samsung/jerryscript/pull/938#discussion-diff-55458630
This is a follow up work after #938 and #910. The function pointer is wrong and we have to change it's parameters to `ecma_char_t` so we can avoid unnecessary type casts. It has no measurable benefit on sunspider.

zherczeg:
Ok. LGTM

PR #965, opened 2016-03-13
Fix SCCS IDs in fdlibm

akiss77:
Most of the legacy version control IDs in fdlibm are already simple
comments at the beginning of the source files. However, in some
files, there are leftover global variables and pragmas. (The global
variable in e_pow.c even increases the .data section in both debug
and release builds.) This patch turns all legacy IDs to comments of
unified style.


zherczeg:
LGTM

LaszloLango:
LGTM

PR #969, opened 2016-03-18
Add jerry_port_get_time API and use it for Date.now()

franc0is:


LaszloLango:
What is the motivation of this patch? If you want to move date related functions to jerry port, then you need to move `ecma_date_local_tza` and `ecma_date_daylight_saving_ta` too.

franc0is:
The system I am porting jerryscript to isnt POSIX and does not provide
gettimeofday, the port API seemed like the right place to hold such
platform dependent code.

You are correct on the second part and I do intend to move those those
other function as well (little by little).

I thought I'd start small so that we'd have this conversation before I went
too far down this road.

franc0is:
Superceded by #1018

PR #971, opened 2016-03-18
Optimize string character access for ascii strings

zherczeg:
Optimize string character access for ascii strings and refactor type store to allow two byte hashes in the future.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -1,4 +1,5 @@
 /* Copyright 2014-2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -686,8 +687,37 @@ typedef enum
   ECMA_STRING_CONTAINER_UINT32_IN_DESC, /**< actual data is UInt32-represeneted Number
                                              stored locally in the string's descriptor */
   ECMA_STRING_CONTAINER_MAGIC_STRING, /**< the ecma-string is equal to one of ECMA magic strings */
-  ECMA_STRING_CONTAINER_MAGIC_STRING_EX /**< the ecma-string is equal to one of external magic strings */
+  ECMA_STRING_CONTAINER_MAGIC_STRING_EX, /**< the ecma-string is equal to one of external magic strings */
+
+  ECMA_STRING_CONTAINER__MAX = ECMA_STRING_CONTAINER_MAGIC_STRING_EX /**< maximum value */
 } ecma_string_container_t;
+
+/**
+ * Mask for getting the container of a string.
+ */
+#define ECMA_STRING_CONTAINER_MASK 0x7u
+
+/**
+ * Value for increasing or decreasing the reference counter.
+ */
+#define ECMA_STRING_REF_ONE (1u << 3)
+
+/**
+ * Maximum value of the reference counter (8191).
+ */
+#define ECMA_STRING_MAX_REF (0x1fffu << 3)
+
+/**
+ * Set reference counter to zero (for refs_and_container member below).
+ */
+#define ECMA_STRING_SET_REF_TO_ONE(refs_and_container) \
+  ((uint16_t) (((refs_and_container) & ECMA_STRING_CONTAINER_MASK) | ECMA_STRING_REF_ONE))
+
+/**
+ * Returns with the container type of a string.
+ */
+#define ECMA_STRING_GET_CONTAINER(string_desc_p) \
+  ((ecma_string_container_t) ((string_desc_p)->refs_and_container & ECMA_STRING_CONTAINER_MASK))
 
 /**
  * ECMA string-value descriptor
@@ -695,10 +725,7 @@ typedef enum
 typedef struct ecma_string_t
 {
   /** Reference counter for the string */
-  unsigned int refs : CONFIG_ECMA_REFERENCE_COUNTER_WIDTH;
-
-  /** Where the string's data is placed (ecma_string_container_t) */
-  uint8_t container;
+  uint16_t refs_and_container;
 
   /** Hash of the string (calculated from two last characters of the string) */
   lit_string_hash_t hash;
@@ -712,10 +739,10 @@ typedef struct ecma_string_t
     mem_cpointer_t lit_cp;
 
     /** Compressed pointer to an ecma_collection_header_t */
-    __extension__ mem_cpointer_t collection_cp : ECMA_POINTER_FIELD_WIDTH;
+    mem_cpointer_t collection_cp;
 
     /** Compressed pointer to an ecma_number_t */
-    __extension__ mem_cpointer_t number_cp : ECMA_POINTER_FIELD_WIDTH;
+    mem_cpointer_t number_cp;
 
     /** UInt32-represented number placed locally in the descriptor */
     uint32_t uint32_number;
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -47,6 +47,24 @@ JERRY_STATIC_ASSERT ((int32_t) ECMA_STRING_MAX_CONCATENATION_LENGTH ==
                      ECMA_STRING_MAX_CONCATENATION_LENGTH_should_be_representable_with_int32_t);
 
 /**
+ * The ecma string ref counter should start after the container field.
+ */
+JERRY_STATIC_ASSERT (ECMA_STRING_CONTAINER_MASK + 1 == ECMA_STRING_REF_ONE,
+                     ecma_string_ref_counter_should_start_after_the_container_field);
+
+/**
+ * The ecma string container types must be lower than the container mask.
+ */
+JERRY_STATIC_ASSERT (ECMA_STRING_CONTAINER_MASK >= ECMA_STRING_CONTAINER__MAX,
+                     ecma_string_container_types_must_be_lower_than_the_container_mask);
+
+/**
+ * The ecma string ref and container fields should fill the 16 bit field.
+ */
+JERRY_STATIC_ASSERT ((ECMA_STRING_MAX_REF | ECMA_STRING_CONTAINER_MASK) == UINT16_MAX,
+                     ecma_string_ref_and_container_fields_should_fill_the_16_bit_field);
+
+/**
  * String header
  */
 typedef struct
@@ -92,8 +110,7 @@ ecma_init_ecma_string_from_lit_cp (ecma_string_t *string_p, /**< descriptor to i
 
   JERRY_ASSERT (LIT_RECORD_IS_CHARSET (lit));
 
-  string_p->refs = 1;
-  string_p->container = ECMA_STRING_CONTAINER_LIT_TABLE;
+  string_p->refs_and_container = ECMA_STRING_CONTAINER_LIT_TABLE | ECMA_STRING_REF_ONE;
   string_p->hash = lit_charset_literal_get_hash (lit);
 
   string_p->u.common_field = 0;
@@ -108,8 +125,7 @@ ecma_init_ecma_string_from_magic_string_id (ecma_string_t *string_p, /**< descri
                                             lit_magic_string_id_t magic_string_id) /**< identifier of
                                                                                          the magic string */
 {
-  string_p->refs = 1;
-  string_p->container = ECMA_STRING_CONTAINER_MAGIC_STRING;
+  string_p->refs_and_container = ECMA_STRING_CONTAINER_MAGIC_STRING | ECMA_STRING_REF_ONE;
   string_p->hash = lit_utf8_string_calc_hash (lit_get_magic_string_utf8 (magic_string_id),
                                               lit_get_magic_string_size (magic_string_id));
 
@@ -125,8 +141,7 @@ ecma_init_ecma_string_from_magic_string_ex_id (ecma_string_t *string_p, /**< des
                                                lit_magic_string_ex_id_t magic_string_ex_id) /**< identifier of
                                                                                            the external magic string */
 {
-  string_p->refs = 1;
-  string_p->container = ECMA_STRING_CONTAINER_MAGIC_STRING_EX;
+  string_p->refs_and_container = ECMA_STRING_CONTAINER_MAGIC_STRING_EX | ECMA_STRING_REF_ONE;
   string_p->hash = lit_utf8_string_calc_hash (lit_get_magic_string_ex_utf8 (magic_string_ex_id),
                                               lit_get_magic_string_ex_size (magic_string_ex_id));
 
@@ -161,8 +176,7 @@ ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 stri
   JERRY_ASSERT (string_size > 0);
 
   ecma_string_t *string_desc_p = ecma_alloc_string ();
-  string_desc_p->refs = 1;
-  string_desc_p->container = ECMA_STRING_CONTAINER_HEAP_CHUNKS;
+  string_desc_p->refs_and_container = ECMA_STRING_CONTAINER_HEAP_CHUNKS | ECMA_STRING_REF_ONE;
   string_desc_p->hash = lit_utf8_string_calc_hash (string_p, string_size);
 
   string_desc_p->u.common_field = 0;
@@ -200,8 +214,7 @@ ecma_string_t *
 ecma_new_ecma_string_from_uint32 (uint32_t uint32_number) /**< UInt32-represented ecma-number */
 {
   ecma_string_t *string_desc_p = ecma_alloc_string ();
-  string_desc_p->refs = 1;
-  string_desc_p->container = ECMA_STRING_CONTAINER_UINT32_IN_DESC;
+  string_desc_p->refs_and_container = ECMA_STRING_CONTAINER_UINT32_IN_DESC | ECMA_STRING_REF_ONE;
 
   lit_utf8_byte_t byte_buf[ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32];
   lit_utf8_size_t bytes_copied = ecma_uint32_to_utf8_string (uint32_number,
@@ -246,8 +259,7 @@ ecma_new_ecma_string_from_number (ecma_number_t num) /**< ecma-number */
   }
 
   ecma_string_t *string_desc_p = ecma_alloc_string ();
-  string_desc_p->refs = 1;
-  string_desc_p->container = ECMA_STRING_CONTAINER_HEAP_NUMBER;
+  string_desc_p->refs_and_container = ECMA_STRING_CONTAINER_HEAP_NUMBER | ECMA_STRING_REF_ONE;
   string_desc_p->hash = lit_utf8_string_calc_hash (str_buf, str_size);
 
   string_desc_p->u.common_field = 0;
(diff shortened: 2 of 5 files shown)


jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -682,6 +666,153 @@ ecma_string_to_utf8_string (const ecma_string_t *string_desc_p, /**< ecma-string
...
+static lit_utf8_size_t __attr_always_inline___
+ecma_string_get_heap_number_size (mem_cpointer_t number_cp) /**< Compressed pointer to an ecma_number_t */
+{
+  const ecma_number_t *num_p = ECMA_GET_NON_NULL_POINTER (ecma_number_t, number_cp);
+  lit_utf8_byte_t buffer[ECMA_MAX_CHARS_IN_STRINGIFIED_NUMBER];
+
+  return ecma_number_to_utf8_string (*num_p, buffer, sizeof (buffer));
+} /* ecma_string_get_heap_number_size */
+
+#define ECMA_STRING_IS_ASCII(char_p, size) ((size) == lit_utf8_string_length ((char_p), (size)))
```
LaszloLango: Missing comment

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -682,6 +666,153 @@ ecma_string_to_utf8_string (const ecma_string_t *string_desc_p, /**< ecma-string
...
+ * @return number of bytes in the buffer
+ */
+static lit_utf8_size_t __attr_always_inline___
+ecma_string_get_heap_number_size (mem_cpointer_t number_cp) /**< Compressed pointer to an ecma_number_t */
+{
+  const ecma_number_t *num_p = ECMA_GET_NON_NULL_POINTER (ecma_number_t, number_cp);
+  lit_utf8_byte_t buffer[ECMA_MAX_CHARS_IN_STRINGIFIED_NUMBER];
+
+  return ecma_number_to_utf8_string (*num_p, buffer, sizeof (buffer));
+} /* ecma_string_get_heap_number_size */
```
akiss77: Why did these functions and data get moved around? Are they affected by this change?

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -690,15 +690,31 @@ typedef enum
...
 
 /**
+ * ECMA string related constants
+ */
+typedef enum
+{
+  ECMA_STRING_CONTAINER_MASK = 0x7u, /**< container mask */
+  ECMA_STRING_REF_ONE = 1u << 3, /**< value for increasing or decreasing the reference counter */
+  ECMA_STRING_MAX_REF = 0x1fffu << 3, /**< maximum value of the reference counter (8191) */
+} ecma_string_constants_t;
```
akiss77: These are hard-wired constants here. Shouldn't we have some assert to ensure that the highest value in `ecma_string_container_t` fits in `ECMA_STRING_CONTAINER_MASK`? The use of `<< 3` is also related.

Moreover, IMHO this is an abuse of enums. These are just constants and the type `ecma_string_constants_t` is not used anywhere in the code. We do use enums with non-sequential values in the code already (i.e., for bit-fields), but always when the type itself is really used somewhere. I haven't seen enums used just for grouping values, as proposed here. (And I've vote against it.)

jerry-core/ecma/base/ecma-helpers-string.c
zherczeg: I used them in the new function. Either I need to create a definition or move them. I chose the second.

jerry-core/ecma/base/ecma-globals.h
zherczeg: You didn't see them? I didn't like it either, but this is the preferred way in Jerry. Just search for "typedef enum" and you can find a lot of similar ones. E.g jerry_flag_t, jerry_fatal_code_t, vm_oc_get_types, ecma_property_flags_t, re_flags_t. The reason was something like you cannot misuse them. The compiler gives better error messages for them than for defines.

I will add a static assert.

zherczeg:
```
                                        3d-cube.js |   2.565s ->   2.605s :  -1.559%  (+-1.623%) : [≈]
                            access-binary-trees.js |   1.357s ->   1.345s :  +0.921%  (+-1.637%) : [≈]
                                access-fannkuch.js |   7.580s ->   7.723s :  -1.880%  (+-1.238%) : [-]
                                   access-nbody.js |   2.940s ->   2.940s :  +0.000%  (+-1.147%) : [≈]
                       bitops-3bit-bits-in-byte.js |   1.675s ->   1.712s :  -2.239%  (+-1.349%) : [-]
                            bitops-bits-in-byte.js |   2.455s ->   2.482s :  -1.120%  (+-0.914%) : [-]
                             bitops-bitwise-and.js |   3.285s ->   3.265s :  +0.609%  (+-2.171%) : [≈]
                             bitops-nsieve-bits.js |  20.165s ->  20.230s :  -0.322%  (+-1.766%) : [≈]
                          controlflow-recursive.js |   0.938s ->   0.943s :  -0.533%  (+-2.209%) : [≈]
                                     crypto-aes.js |   3.540s ->   3.502s :  +1.059%  (+-0.790%) : [+]
                                     crypto-md5.js |   7.180s ->   3.583s : +50.105%  (+-0.804%) : [+]
                                    crypto-sha1.js |   3.885s ->   2.208s : +43.179%  (+-0.761%) : [+]
                              date-format-tofte.js |   2.178s ->   2.130s :  +2.181%  (+-0.656%) : [+]
                              date-format-xparb.js |   0.860s ->   0.840s :  +2.326%  (+-0.000%) : [+]
                                    math-cordic.js |   2.880s ->   2.925s :  -1.562%  (+-1.326%) : [-]
                              math-partial-sums.js |   1.580s ->   1.613s :  -2.057%  (+-0.924%) : [-]
                             math-spectral-norm.js |   1.448s ->   1.485s :  -2.591%  (+-1.558%) : [-]
                                  string-base64.js |  30.968s ->   5.105s : +83.515%  (+-0.239%) : [+]
                                   string-fasta.js |   3.328s ->   3.270s :  +1.728%  (+-2.047%) : [≈]
                                   Geometric mean: |                  Speed up: 14.673% (+-0.330%) : [+]
```

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1171,7 +1245,19 @@ ecma_string_get_char_at_pos (const ecma_string_t *string_p, /**< ecma-string */
   ecma_length_t string_length = ecma_string_get_length (string_p);
   JERRY_ASSERT (index < string_length);
 
-  lit_utf8_size_t buffer_size = ecma_string_get_size (string_p);
+  lit_utf8_size_t buffer_size;
+  bool is_ascii;
+  const lit_utf8_byte_t *chars_p = ecma_string_raw_chars (string_p, &buffer_size, &is_ascii);
+
+  if (chars_p != NULL)
```
LaszloLango: I can't see how could be the chars_p NULL. Could you show me an example? :)

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1180,7 +1266,14 @@ ecma_string_get_char_at_pos (const ecma_string_t *string_p, /**< ecma-string */
   lit_utf8_size_t sz = ecma_string_to_utf8_string (string_p, utf8_str_p, buffer_size);
   JERRY_ASSERT (sz == buffer_size);
 
-  ch = lit_utf8_string_code_unit_at (utf8_str_p, buffer_size, index);
+  if (is_ascii)
```
LaszloLango: If I see this correctly then we can only get here, when chars_p is NULL and that means we have to deal with ECMA_STRING_CONTAINER_HEAP_NUMBER and ECMA_STRING_CONTAINER_UINT32_IN_DESC. This means that this will be always true. I'd change this to a JERRY_ASSERT

jerry-core/ecma/base/ecma-helpers-string.c
LaszloLango: I found out. (ECMA_STRING_CONTAINER_UINT32_IN_DESC and ECMA_STRING_CONTAINER_HEAP_NUMBER has no raw string data)

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -690,15 +690,31 @@ typedef enum
...
+
+/**
+ * Set reference counter to zero (for refs_and_container member below).
+ */
+#define ECMA_STRING_SET_REF_TO_ONE(refs_and_container) \
+  ((uint16_t) (((refs_and_container) & ECMA_STRING_CONTAINER_MASK) | ECMA_STRING_REF_ONE))
+
+#define ECMA_STRING_GET_CONTAINER(string_desc_p) \
```
LaszloLango: Missing comment

jerry-core/ecma/base/ecma-globals.h
akiss77: I did search for the enums. There are exactly 49 enum typedefs in jerry-core. Wherever they occur, they are used for grouping a set of values we're choosing one from, or a set of flags we are bit-or-ing together. They are almost always also used as types in data structures, local variables, function arguments, or return values. In some rare cases only their values (labels) are used, because we chose not to use their real type in a struct but went for a type with an exactly known size (like `uint8_t`). (Note: this is going against the type safety you also mentioned above.) I haven't seen any use where a bit mask, a starting counter, and a counter limit would be grouped together in an enum.

(Case-by-case analysis: `jerry_flag_t`, `jerry_fatal_code_t` are used as types in the code. `ecma_property_flags_t` is not used as a type, since the struct that countains it forces the flags into a `unit8_t`. `vm_oc_get_types` is used in a similar way, it is used as a set of values, only the real type used in the code is different by some reason. `re_flags_t` is not used as a type either, but it could be -- there are already some code paths in js-lexer.c which cast around 16 and 32 bit integers when dealing with these flags because of no apparent reason.)

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -682,6 +666,153 @@ ecma_string_to_utf8_string (const ecma_string_t *string_desc_p, /**< ecma-string
...
+
+/**
+ * Return the raw byte array of the string, if it is available.
+ *
+ * @return byte array start - if the byte array of a string is available
+ *         NULL - otherwise
+ */
+const lit_utf8_byte_t *
+ecma_string_raw_chars (const ecma_string_t *string_p, /* ecma-string */
+                       lit_utf8_size_t *size_p, /* size of the ecma string */
```
LaszloLango: Minor style issue:
`/* [out] size of the ecma string */`

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -682,6 +666,153 @@ ecma_string_to_utf8_string (const ecma_string_t *string_desc_p, /**< ecma-string
...
+                       bool *is_ascii_p) /* true, if ascii string (size == length) */
```
LaszloLango: Minor style issue:
`/* [out] true, if ascii string (size == length) */`

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -682,6 +666,153 @@ ecma_string_to_utf8_string (const ecma_string_t *string_desc_p, /**< ecma-string
...
+      result_p = lit_get_magic_string_ex_utf8 (string_p->u.magic_string_ex_id);
+
+      /* All extended magic strings must be ascii strings. */
+      JERRY_ASSERT (ECMA_STRING_IS_ASCII (result_p, size));
+      break;
+    }
+  }
+
+  *size_p = size;
+  *is_ascii_p = (length == size);
```
LaszloLango: I think `is_ascii_p` could explicitly set up for numbers and magic strings, but that may not have measurable benefit.

jerry-core/ecma/base/ecma-helpers-string.c
akiss77: Is the new function used before the original location of these? Would writing the new function just below them have been an option?

zherczeg: Yes, but the function belongs to that group I placed it. I like when function implementation follows the declaration order in the header, but if you insist, I can change it.

jerry-core/ecma/base/ecma-helpers-string.c
zherczeg: I don't think we gain too much. I was thinking about that myself but choose this cleaner solution at the end.

jerry-core/ecma/base/ecma-globals.h
zherczeg: I can change it to defines if you want that.

jerry-core/ecma/base/ecma-helpers-string.c
zherczeg: I wanted to prepare for future uses, but it is true, we can reimplement this.

jerry-core/ecma/base/ecma-helpers-string.c
akiss77: Wasn't insisting just wondering. But it's clear now, thanks for the explanation. It's good as is.

jerry-core/ecma/base/ecma-globals.h
LaszloLango: In this particular case I think defines would be better.

akiss77: +1 to `#define`s

zherczeg:
Please review the patch again.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -47,6 +47,18 @@ JERRY_STATIC_ASSERT ((int32_t) ECMA_STRING_MAX_CONCATENATION_LENGTH ==
...
+ * The ecma string ref counter should start after the container field.
+ */
+JERRY_STATIC_ASSERT (ECMA_STRING_CONTAINER_MASK + 1 == ECMA_STRING_REF_ONE,
+                     ecma_string_ref_counter_should_start_after_the_container_field);
+
+/**
+ * The ecma string ref and container fields should fill the 16 bit field.
+ */
+JERRY_STATIC_ASSERT ((ECMA_STRING_MAX_REF | ECMA_STRING_CONTAINER_MASK) == UINT16_MAX,
+                     ecma_string_ref_and_container_fields_should_fill_the_16_bit_field);
```
akiss77: I think one more static assert is missing: the one that checks that `ECMA_STRING_CONTAINER_MASK` is bigger than or equal to the largest possible value of `ecma_string_container_t`.

zherczeg: Shall I add a new enum for max testing? I wouldn't want to overdo this.

akiss77: A new enum is certainly not needed. But you can add a new label to the end of the existing `ecma_string_container_t` to get the count of labels automatically calculated. I've seen that used in some other enums as well (and called something like `xxx__COUNT`).

zherczeg: Unfortunately the side effect is that I need to extend switches with this new value (doesn't look too nice). Do you still want it?

akiss77: I still would like to see an assert and also an automatically calculated __COUNT label for the enum. But that does not necessarily require the extension of switches. 

I've looked for the switches with `ECMA_STRING_CONTAINER_` case labels and almost all of them already had default branches with some kind of an assert (unreachable, or checking for an explicitly given container type), which is what they should do IMHO. The only exception I've found is in function `ecma_deref_ecma_string`, which does not check for invalid values -- but it should (or already should have) IMHO!

Where have you found switches that needed extension?

zherczeg: Added a commit which adds the changes.
I don't really like the double JERRY_UNREACHABLE() sequences.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -510,10 +512,14 @@ ecma_deref_ecma_string (ecma_string_t *string_p) /**< ecma-string */
     case ECMA_STRING_CONTAINER_MAGIC_STRING_EX:
     {
       /* only the string descriptor itself should be freed */
+      break;
+    }
+    case ECMA_STRING_CONTAINER__COUNT:
```
akiss77: It's unnecessary to have this case explicitly refer to `ECMA_STRING_CONTAINER__COUNT`. A `default:` branch would suite here perfectly. (Actually, a default unreachable assert was missing from here, unrelated to the current change. I don't know why this was forgotten originally.)

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -569,6 +575,10 @@ ecma_string_to_number (const ecma_string_t *str_p) /**< ecma-string */
 
       return num;
     }
+    case ECMA_STRING_CONTAINER__COUNT:
+    {
+      JERRY_UNREACHABLE ();
+    }
```
akiss77: This is indeed a duplication and one of the asserts is not needed. The `JERRY_UNREACHABLE();` below, after the switch, acts exactly as if `default: JERRY_UNREACHABLE();` was written in the switch. So, either just drop the `case ECMA_STRING_CONTAINER__COUNT:` or drop the `JERRY_UNREACHABLE();` below (but in the second case, rename the case to `default:`. Thus, no mentioning of `ECMA_STRING_CONTAINER__COUNT` is needed in either case.)

jerry-core/ecma/base/ecma-helpers-string.c
zherczeg: I was thinking about this, and realized what is the wrong with adding a default approach. The "default: unreachable" suggests that "there are other valid enums, but they cannot appear here because of certain reasons". But that is not the case here! I consider this COUNT thing a style issue, and I don't want to reduce the readability of the code because of style. I have an idea though, I will update the code.

zherczeg:
Patch updated with a new approach.

LaszloLango:
LGTM, but please update the copyright headers in

- jerry-core/ecma/base/ecma-globals.h
- jerry-core/ecma/operations/ecma-string-object.c
- jerry-core/lit/lit-magic-strings.inc.h

akiss77:
@zherczeg With the latest revision, you loose an important feature of the `__COUNT` pattern: it's used in C projects to _automatically_ count the valid enums. This helps keeping the code safe, as however and whoever edits the enum (adds or removes labels), the count is always updated and its value is always correct. So, it's more then just a style pattern. The manually maintained `__MAX` approach looses these benefits. (Also, it goes against the style of the project, which uses the `__COUNT` approach all around, but not a `__MAX` terminology.)

Regarding the `default` branches: they are not only to signal that "there are other valid enums, but they cannot appear here because of certain reasons". They are also safeguards to take care of the case when the data format (i.e., the list of valid labels in an enum) has been extended but someone forgot to update the code that works with it. (Basically, whenever you have a switch, it's a very good practice to also have a default branch, or to have explicitly documented why it's not needed.) Even existing string container code uses switch with a default branch and unreachable assert -- even though it covers all valid cases (somewhere around string copies). Moreover, if you only have an unreachable assert macro in the default branch, it turns to a no-op in release code, so it has no effect on code size or running time there.

zherczeg:
"someone forgot to update the code that works with it" -> this is my biggest fear. Updating a max is easy, since it is around the typedef, and whoever works with it will notice it. However, default branches eliminate compiler warnings and will create hidden errors which the testing system might not capture (or a lot of debugging needed). Personally I like full (all values checked) switches without defaults. Anyway if you insist I will do your request even if I don't agree at all.

zherczeg:
I changed the __MAX to be part of the enum, and it works.

akiss77:
LGTM (FWIW)

zherczeg:
Copyright updated.

PR #973, opened 2016-03-19
Fix: code-related vera++ rules should not be enforced in comments

akiss77:

### DIFF
--- a/tools/vera++/scripts/rules/jerry_no_space_before_closing_parentheses.tcl
+++ b/tools/vera++/scripts/rules/jerry_no_space_before_closing_parentheses.tcl
@@ -1,6 +1,7 @@
 #!/usr/bin/tclsh
 
-# Copyright 2015 Samsung Electronics Co., Ltd.
+# Copyright 2015-2016 Samsung Electronics Co., Ltd.
+# Copyright 2016 University of Szeged
 #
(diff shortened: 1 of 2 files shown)


zherczeg:
Thank you for fixing this. Could you add a test as well which proves that non-comments checks are not broken?

akiss77:
AFAIK we have no automated testing facility for styles, do we? I've tried the following manually: before and after applying the patch, I've tweaked `main-mcu.c` (any other source file would have done it) and added the following lines:

```c
/* ( sigh *sigh* sigh ) */
char* p = NULL;
size_t s = sizeof (int );
```

Then I ran `make check-vera`. Without the patch, I got 4 errors (even within the comment), with the patch the remaining 2 errors come from code only, not from the comment. So, code is correctly checked still but comments are not checked anymore.

zherczeg:
I see you updated the patch, but it seems a squash was done, so I don't see what issue turned up. Anyway, similar to the fdlibm unittests it would be good at some point to have some vera++ tests since now we have a lot of changes there. I worry that we accidentally broke the whole thing and no issues will be reported anymore. That can be done in another patch.

LGTM

akiss77:
There was no issue, the patch has not been changed, just rebased to master, since it moved forward a bit since the pull request has been opened.

LaszloLango:
LGTM

PR #974, opened 2016-03-19
Fix regexp character classes with \uNNNN and \xNN code points

### LINKED ISSUE #962: Incorrect regexp character class handling with \uNNNN and \xNN code points
Test code:
```js
function re_test (pattern, string)
{
  var r = new RegExp (pattern);
  print ("'" + pattern + "' ~ '" + string + "' => '" + (new RegExp (pattern)).exec(string) + "'");
}

re_test ("[\\u0020]", "u");
re_test ("[\\u0020]", " ");
re_test ("[\\u0020]", "x");

re_test ("[\\x20]", "u");
re_test ("[\\x20]", " ");
re_test ("[\\x20]", "x");
```

Output:
```
'[\u0020]' ~ 'u' => 'u'
'[\u0020]' ~ ' ' => ' '
'[\u0020]' ~ 'x' => 'null'
'[\x20]' ~ 'u' => 'null'
'[\x20]' ~ ' ' => ' '
'[\x20]' ~ 'x' => 'x'
```

The first and last results are incorrect.

This could either be fixed in #961 or handled separately (in which case a regtest could be added to the test suite as well, based on the above).

akiss77:
Also adding a regression test covering the issue.

Fixes #962

### DIFF
--- a/jerry-core/parser/regexp/re-parser.c
+++ b/jerry-core/parser/regexp/re-parser.c
@@ -421,6 +421,7 @@ re_parse_char_class (re_parser_ctx_t *parser_ctx_p, /**< number of classes */
 
         parser_ctx_p->input_curr_p += 2;
         append_char_class (re_ctx_p, code_unit, code_unit);
+        ch = LIT_CHAR_UNDEF;
       }
       else if (ch == LIT_CHAR_LOWERCASE_U)
       {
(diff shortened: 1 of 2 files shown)


zherczeg:
What is the problem exactly?

akiss77:
It's described in #962, as mentioned in the first comment. (In short: if a regexp contains a character class, and that character class contains a hex-specified code point -- either \uNNNN or \xNN --, then the current implementation incorrectly adds the character 'u' or 'x' to the character class as well.)

zherczeg:
LGTM. I think in such cases it is better to describe the problem in the commit log as well.

LaszloLango:
LGTM

PR #976, opened 2016-03-22
Remove MEM_HEAP_PTR_64 macro

akiss77:
MEM_HEAP_PTR_64 is duplicating existing information: stdint.h, which
header is already used by the project, defines various _MAX macros
for upper limits of integer types. The comparison of UINTPTR_MAX and
UINT32_MAX can give the same info as encoded in MEM_HEAP_PTR_64.
The stdint.h-based approach has the benefit that jerry can support
any 64-bit architecture without the need for editing the build
system. (With the existing approach, CMakeLists has to know about
every 64-bit architecture to work properly.)

Thus, removing the extraneous macro from the code.

The patch also changes the mem_pools_chunk_t struct, as it turned
out that the struct does not have to be padded to MEM_POOL_CHUNK_SIZE.
(The padding also depended on MEM_HEAP_PTR_64.) It is enough if the
size of the struct is smaller than (or equal to) MEM_POOL_CHUNK_SIZE.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,9 +223,6 @@ project (Jerry C ASM)
 
 # Compiler / Linker flags
  set(COMPILE_FLAGS_JERRY "-fno-builtin")
- if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "x86_64")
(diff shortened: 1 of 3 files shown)


LaszloLango:
LGTM

dbatyai:
LGTM

PR #981, opened 2016-03-24
Added ST32F401 Nucelo to targets folder.

polaroi8d:
As the mbed is a cpp project from which several classes were used we need to have .cpp files.

JerryScript-DCO-1.0-Signed-off-by: Levente Orban orbanl@inf.u-szeged.hu

PR #982, opened 2016-03-24
Modified mbedk64f target 

polaroi8d:
As the mbed is a cpp project from which several classes were used we need to have .cpp files. 


polaroi8d:
There are some problems with the renamed files, because if we rewrite their names, didn't find specific libaries, like cstddef.

zherczeg:
LGTM. 
Just one more question: I am surprised that c -> cpp conversions do not cause any linking problems. Is it because of the extern C in api headers?

targets/mbedk64f/source/jerry_extapi.cpp
```diff
@@ -14,19 +14,15 @@
  */
 
 #include "jerry-core/jerry.h"
-
 #include <stdlib.h>
 #include <stdio.h>
-
```
LaszloLango: These two empty lines help to separate the include types, such as jerry API, system and target headers. Why did you remove them?

targets/mbedk64f/source/jerry_run.cpp
```diff
@@ -35,7 +36,7 @@ int js_entry (const char *source_p, const size_t source_size)
 
   js_register_functions ();
 
-  if (!jerry_parse (jerry_src, source_size))
+  if (!jerry_parse (jerry_src, source_size, &err_obj_p))
   {
     printf ("Error: jerry_parse failed\r\n");
     ret_code = JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION;
```
LaszloLango: 'err_obj_p' must be released here.

targets/mbedk64f/source/jerry_run.cpp
```diff
@@ -44,7 +45,7 @@ int js_entry (const char *source_p, const size_t source_size)
   {
     if ((flags & JERRY_FLAG_PARSE_ONLY) == 0)
     {
-      ret_code = jerry_run ();
+      ret_code = jerry_run (&err_obj_p);
```
LaszloLango: Please check `err_obj_p` and release if it is not `NULL`

polaroi8d:
@zherczeg I guess.

targets/mbedk64f/source/jerry_run.cpp
polaroi8d: Okey.

akiss77:
@polaroi8d I've taken a closer look at the 4 open mbed-related PRs (this one + #981, #986, #989). They are way too much clones of each other. Except for differences in style (whitespace, comments, order of includes), the `js` and `source` directories are copy-paste clones of each other (perhaps the only difference between them is the serial port baud rate set in `main.cpp`). The real differences seem to be "only" in readme, yotta, module, and make files in the target root directory.

I'd propose unifying the mbed-related targets to prevent duplications. I could imagine a structure like
```
targets/
    mbed/
        js/
        source/
        port/
        board/
            k64f/
            stm32f401nucleo/
            stm32f4/
            stm32f429i/
```
Since the k64f support is already in the code base, I propose to update this PR to lay the foundations of the new structure as well as update the code to latest jerry (please note that an external port implementation is needed as identified in #1032).

Do have some discussion with @knightburton to make sure that the other PRs are aligned.

polaroi8d:
@akiss77 there is a new PR #1019 where we already done, what you describe.

akiss77:
@polaroi8d Ahm, working my way up from the bottom of the list, I did not get that far. My bad. However, if these four PRs are superceded by #1019 then could you please close these with an appropriate comment so they don't show up anymore as ones waiting for review?

polaroi8d:
@akiss77 Of course.

polaroi8d:
In this and some other PRs what we did, they are way too much clones of each other, so this is the reason why I'm closed this pull request, because we created a new PR ( #1019 ) where merged the mbed releated targets into one soruce.

PR #984, opened 2016-03-25
Refactor object properties to become real fields.

zherczeg:
The positive side effect is that the maximum Jerry memory is increased to 512K.
Furthermore a slight (1.3%) performance improvement was measured on RPi2 with
SunSpider.

JerryScript-DCO-1.0-Signed-off-by: Zoltan Herczeg zherczeg.u-szeged@partner.samsung.com

### DIFF
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -88,108 +88,64 @@ static void ecma_gc_mark (ecma_object_t *object_p);
 static void ecma_gc_sweep (ecma_object_t *object_p);
 
 /**
- * Get GC reference counter of the object.
- */
-static uint32_t
-ecma_gc_get_object_refs (ecma_object_t *object_p) /**< object */
-{
-  JERRY_ASSERT (object_p != NULL);
-
-  return (uint32_t) JRT_EXTRACT_BIT_FIELD (uint64_t, object_p->container,
-                                           ECMA_OBJECT_GC_REFS_POS,
-                                           ECMA_OBJECT_GC_REFS_WIDTH);
-} /* ecma_gc_get_object_refs */
-
-/**
- * Set GC reference counter of the object.
- */
-static void
-ecma_gc_set_object_refs (ecma_object_t *object_p, /**< object */
-                         uint32_t refs) /**< new reference counter */
-{
-  JERRY_ASSERT (object_p != NULL);
-
-  object_p->container = JRT_SET_BIT_FIELD_VALUE (uint64_t, object_p->container,
-                                                 refs,
-                                                 ECMA_OBJECT_GC_REFS_POS,
-                                                 ECMA_OBJECT_GC_REFS_WIDTH);
-} /* ecma_gc_set_object_refs */
-
-/**
  * Get next object in list of objects with same generation.
  */
-static ecma_object_t *
+static inline ecma_object_t *
 ecma_gc_get_object_next (ecma_object_t *object_p) /**< object */
 {
   JERRY_ASSERT (object_p != NULL);
 
-  JERRY_ASSERT (sizeof (uintptr_t) * JERRY_BITSINBYTE >= ECMA_OBJECT_GC_NEXT_CP_WIDTH);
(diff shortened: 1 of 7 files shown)


zherczeg:
Review please.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -279,151 +279,95 @@ typedef struct ecma_property_t
...
+ * Description of ECMA-object or lexical environment
+ * (depending on is_lexical_environment).
  */
-#define ECMA_OBJECT_LEX_ENV_OUTER_REFERENCE_CP_POS (ECMA_OBJECT_LEX_ENV_TYPE_POS + \
-                                                    ECMA_OBJECT_LEX_ENV_TYPE_WIDTH)
-#define ECMA_OBJECT_LEX_ENV_OUTER_REFERENCE_CP_WIDTH (ECMA_POINTER_FIELD_WIDTH)
+typedef struct ecma_object_t
+{
+  /** object type, flags, and refs */
+  uint16_t type_flags_refs;
```
LaszloLango: Could you add a more descriptive comment for this? For example it is not trivial why 
`if (object_p->type_flags_refs < ECMA_OBJECT_MAX_REF)` will work. (I know it is correct at the moment :) ) I think we need to give some help here. How should be tpye/flags/refs used?

zherczeg:
Fixed.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -180,16 +137,20 @@ ecma_gc_set_object_visited (ecma_object_t *object_p, /**< object */
     is_visited = !is_visited;
   }
 
-  object_p->container = JRT_SET_BIT_FIELD_VALUE (uint64_t, object_p->container,
-                                                 is_visited,
-                                                 ECMA_OBJECT_GC_VISITED_POS,
-                                                 ECMA_OBJECT_GC_VISITED_WIDTH);
+  if (is_visited)
```
akiss77: Aren't the above 5 lines simply equal to `if (is_visited != ecma_gc_visited_flip_flag)`?

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -88,88 +88,45 @@ static void ecma_gc_mark (ecma_object_t *object_p);
...
-                                                 refs,
-                                                 ECMA_OBJECT_GC_REFS_POS,
-                                                 ECMA_OBJECT_GC_REFS_WIDTH);
-} /* ecma_gc_set_object_refs */
-
-/**
  * Get next object in list of objects with same generation.
  */
-static ecma_object_t *
+static inline ecma_object_t *
```
akiss77: Are these (`ecma_gc_{get,set}_object_next`) functions still needed? With the proposed changes, they are merely wrappers around `ECMA_{GET,SET}_POINTER`. (`ecma_gc_{get,set}_object_refs` have been dropped as well, not caring about the asserts. So, do we need these?)

zherczeg: To me this style follows Jerry logic better. But I can remove them if you wish.

akiss77: I'm only looking for consistency. Why have `ecma_gc_{get,set}_object_refs` been removed then?

I thought that all these getters/setters were used/useful to access multiple data packed together in a 64-bit container bitfield. As that container is now being separated into multiple fields that have their own names, accessing them through getters/setters is unnecessary. At least, that was my impression.

zherczeg: That is a different reason. Getting and setting is used for incrementing and decrementing reference counters, but that is an inefficient way to do that (lots of shifting, masking). After the refcounter updates are optimized, those functions are not used anymore, so they are removed.

akiss77: Thanks for the explanation. OK with me then.

akiss77:
LGTM (FWIW)

LaszloLango:
@zherczeg, thanks for the update. LGTM

zherczeg:
Landed in d674b92f266c4bd8892d823e041a8d8f1f77a961

PR #987, opened 2016-04-01
Buildfix for MCUs

bzsolt:
The MCU-targets depend on some platform-related resources.
When these files are missing the build cannot be performed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -229,7 +229,7 @@ $(BUILD_DIR)/$(NATIVE_SYSTEM)/unittests/toolchain.config:
 # $(1) - mcu system name
 define GEN_MCU_TOOLCHAIN_CONFIG_RULE
 .PHONY: $$(BUILD_DIR)/$(1)/toolchain.config
-$$(BUILD_DIR)/$(1)/toolchain.config:
+$$(BUILD_DIR)/$(1)/toolchain.config: prerequisites
 	$$(Q) TOOLCHAIN="build/configs/toolchain_mcu_$(1).cmake"; \
           $$(call WRITE_TOOLCHAIN_CONFIG,$$(BUILD_DIR)/$(1))
 endef


akiss77:
It's more of a theoretical question whether we want this or not. This `make prerequisites` step is an installation of dependencies to the machine used to build jerry. Just like `tools/apt-get-install-deps.sh` and the `brew` variant. We don't run those either at every build but require the developer to run them once. I feel the same way with the prerequsites. They should be installed once and _explicitly_. (I might even be tempted to drop it as a make rule and leave the script in tools only.) But that's just my two pence. I'm open to discussion.

LaszloLango:
LGTM

zherczeg:
LGTM

akiss77:
That was not much of a discussion :)

LaszloLango:
@akiss77, please raise an issue for https://github.com/Samsung/jerryscript/pull/987#issuecomment-204367138
Or do you object to land this PR? I know it is a quick fix, but I think we need it.

akiss77:
@LaszloLango @bzsolt What I could not get from the PR/commit message is what the bug is?

If someone has issued the `make prerequisites` or `tools/prerequisites.sh` command once (manually) then from that point on not a single `make` neads it really. Just like when I once set up my devel machine to have all the make, cmake, ninja, gcc, etc. tools, I don't have to run `apt-get install gcc` again.

So, what's the use case where you regularly need to install prereq's?

bzsolt:
Sorry for the late response. The main problem is, when the STM-related libraries are missing (because you're a newcomer or you removed them with e.g. ```make prerequisites_clean``` and forget to re-install them) the build-process will fail. This situation is a little bit quaint, because now ```precommit``` target requires ```prerequisite```, but a single MCU target isn't.
This fix isn't the most elegant, but fits to the current apparatus.
The ```DEVELOPMENT.md``` suggests that the buildsystem try to setup the prerequisites automatically.
I agree with you, that it should be in the ```apt-get-install-deps.sh``` and in similar variants.
But in this case there is a risk, these scripts are currently used in Travis CI to set-up the environment, and when the source site become unavailable, the CI is going to be red.

akiss77:
I was not clearly explaining my point: I would *not* add the installation of the STM libs to `tools/apt-get-install-deps.sh` nor to `tools/brew-install-deps.sh`. I'd just keep it in `tools/prerequisites.sh` (or with a slightly different name perhaps, but it's not important right now). Partially because I'm aware that we had issues with Travis. But I'd also remove all references to that script from `Makefile` -- and update the documentation to reflect that.

That being said, I wont block the PR, go on with it.

PR #989, opened 2016-04-04
Add stm32f4 board to the targets.

knightburton:
JerryScript-DCO-1.0-Signed-off-by: Imre Kiss kissi@inf.u-szeged.hu

### DIFF
--- a/targets/mbedstm32f4/js/blink.js
+++ b/targets/mbedstm32f4/js/blink.js
@@ -1,0 +1,9 @@
+var check = 1;
+
+function blink() {
+  var blk = (check > 8) ? 1 : 0;
+  led(1, blk);
+  check = (check >= 10) ? 1 : check+1;
+}
+
+print("blink js OK");
--- a/targets/mbedstm32f4/js/blink.js
+++ b/targets/mbedstm32f4/js/blink.js
@@ -1,9 +1,10 @@
 var check = 1;
 
-function blink() {
+function blink ()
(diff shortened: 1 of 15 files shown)


targets/mbedstm32f4/js/blink.js
```diff
@@ -0,0 +1,9 @@
+var check = 1;
+
+function blink() {
+  var blk = (check > 8) ? 1 : 0;
+  led(1, blk);
+  check = (check >= 10) ? 1 : check+1;
+}
```
zherczeg: Please follow Jerry coding style here.

targets/mbedstm32f4/source/main.cpp
```diff
@@ -0,0 +1,75 @@
...
+  int retcode;
+  int src;
+
+  DECLARE_JS_CODES;
+
+  /* run main.js */
+  retcode = js_entry (js_codes[0].source, js_codes[0].length);
+  if (retcode != 0)
+  {
+    pc.printf ("js_entry failed code(%d) [%s]\r\n", retcode, js_codes[0].name);
```
robertsipka: I think pc wasn't declared in this function scope.

knightburton:
In this and some other PRs what we did, they are way too much clones of each other, so this is the reason why I'm closed this pull request, because we created a new PR ( #1019 ) where merged the mbed releated targets into one source.

PR #990, opened 2016-04-06
Eliminate TODO and FIXME macros

akiss77:
Those macros are legacy and are not used consitently throughout the
code base. This patch eliminates their definitions and rewrites
their remaining occurrences (mostly in lit/) to comments.

All occurrences of the words "todo" and "fixme" have been checked
(case insensitively) and made sure that the comments used a
consistent style.

### DIFF
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -39,10 +39,7 @@
 #define JERRY_INTERNAL
 #include "jerry-internal.h"
 
-/**
- * TODO:
- *      Extract GC to a separate component
- */
+/* TODO: Extract GC to a separate component */
 
 /**
  * An object's GC color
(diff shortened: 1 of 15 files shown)


zherczeg:
Actually I would prefer to use them more extensively. Keep only one, and use it everywhere. Easier to find them than comments.

LaszloLango:
@zherczeg, I think the opposite. Those macros print the string to the build log. No one ever will open the build log after a successful build. :) You need to grep for FIXME anyway.

zherczeg:
But we have TODO-s as well. Lets go to comments, but I would prefer only one type (either TODO or FIXME).

akiss77:
Any preferences? What about using TODO everywhere? (All FIXMEs are TODOs but not all TODOs are FIXMEs...)

LaszloLango:
IMHO, "FIXME" comments are enough.

akiss77:
Updated. Only TODOs remained.

zherczeg:
LGTM

LaszloLango:
LGTM

PR #977, opened 2016-03-23
Specialize fdlibm to jerry

akiss77:
Extensive rewrite of fdlibm. Please see commit comments for detailed description of the changes.

This change set gains 3K in code size on release ARM linux and 3.5K on release STM32F3,4 (debug mode experiences even higher reductions, 6-7K of code). Also, it shrinks the r/w data section even smaller.

On x86_64 linux, the benchmark results don't change: maxrss oscillates with +-4K, while running times changes also stay within the +-0.01 sec limits.

Detailed size results:
```
2027cae (authored on 2016-03-17 11:08:53 +0100, committed on 2016-03-22 08:05:00 +0100) [master] Performance optimizations
94f8cf8 (authored on 2016-03-13 22:30:47 +0100, committed on 2016-03-23 01:02:42 +0100) Specialize fdlibm to jerry
453dde6 (authored on 2016-03-16 19:01:31 +0100, committed on 2016-03-23 01:04:04 +0100) Turn simple constants to preprocessor macros
9a8128d (authored on 2016-03-16 21:05:00 +0100, committed on 2016-03-23 01:04:10 +0100) Merge sin/cos/tan and all helper functinos in a single source
af927c3 (authored on 2016-03-17 10:42:00 +0100, committed on 2016-03-23 01:04:20 +0100) Re-style fdlibm to conform to jerry guidelines
a351f91 (authored on 2016-03-20 14:32:05 +0100, committed on 2016-03-23 01:05:02 +0100) [fdlibm-spec] Fix asin

binary                                | data (202) | data (94f-453) | data (9a8-af9-a35) | diff | text (202) | text (94f) | diff | text (453) | diff | text (9a8-af9) | diff | text (a35) | diff |
--------------------------------------|------------|----------------|--------------------|------|------------|------------|------|------------|------|----------------|------|------------|------|
release.linux-arm7lhf                 |         24 |             20 |                 20 |    4 |     180086 |     178354 | 1732 |     177146 | 2940 |         177122 | 2964 |     177114 | 2972 |
release.linux-arm7lhf-cp              |         20 |             16 |                 16 |    4 |     132214 |     130573 | 1641 |     129373 | 2841 |         129325 | 2889 |     129317 | 2897 |
release.linux-arm7lhf-cp_minimal      |          0 |              0 |                  0 |      |      80749 |      80749 |      |      80749 |      |          80749 |      |      80749 |      |
[...]

PR #993, opened 2016-04-07
Turn modified fdlibm into Jerry's own libm

akiss77:
The next step after #977 

Summary of changes (compiled and edited from patch commit messages):

* Rename modified fdlibm to jerry-libm
  * Move third-party/fdlibm to jerry-libm
  * Rename original fdlibm.h to jerry-libm-internal.h
    * And remove it from the public headers.
  * Rename jerry-libm's public header to math.h
    * This also makes jerry-core sources include `<math.h>`. Therefore,
      should anyone want to use a different libm implementation with
      jerry, it becomes possible. (The same way as we provide a minimal
      libc with standard headers, but should it be insufficient or
      conflicting for someone, it can be replaced.)
  * Drop `s_` prefix from jerry-libm sources
    * The original fdlibm implementation had various prefixes (e.g., `k_`
      for sources of kernel routines, and `w_` for wrapper routines), but
      after the specialization of fdlibm to jerry, only `s_` remained.
      Since it does not encode anything anymore, it can be dropped.
  * Stylistic edits to jerry-libm's CMakeLists
    * Align project name with other CMakeLists in the code base
  * Move Jerry-LibM under Apache License
    * Using the same approach as was used by linux-wireless when ath5k
      driver license needed clarification. Solution was proposed by SFLC.
      External mail for future reference: http://lwn.net/Articles/247806/

* Tests & checks
  * Remove FD from the name of libm unit test-related files
  * Make vera++ and cppcheck check jerry-libm

* Targets
  * Speculative update of targets to use jerry-libm


jerry-libm/CMakeLists.txt
```diff
@@ -0,0 +1,76 @@
...
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-error=unused-but-set-variable")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-error=unused-variable")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-error=conversion")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-sign-conversion")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-sign-compare")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-parentheses")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-maybe-uninitialized")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-unknown-pragmas")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-unused-but-set-variable")
+set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-unused-variable")
```
LaszloLango: Is this a short term solution? Do you have any plans to fix these warnings instead of turning them off?

akiss77: Well, it is already a long term solution, since this has been around since we adopted fdlibm. But yes, it would be good to get rid of these at some point in the future.

LaszloLango: Please add a `TODO` comment for this and raise an issue, so we won't forget.

zherczeg:
LGTM. It is sad that git cannot track the moving of small files.

LaszloLango:
LGTM besides https://github.com/Samsung/jerryscript/pull/993#discussion_r59020867

PR #995, opened 2016-04-08
Fix assertion on unhandled exceptions

LaszloLango:
Follow up fix after #815 ('a187e6d'). Fixed the following
assertion on unhandled exceptions:
  * ICE: Assertion 'ecma_get_value_type_field (value) == ECMA_TYPE_OBJECT'
    failed at ecma-helpers-value.c(ecma_get_object_from_value):375.


jerry-core/jerry.c
```diff
@@ -1805,7 +1816,14 @@ jerry_run_simple (const jerry_api_char_t *script_source, /**< script source */
   {
     if ((flags & JERRY_FLAG_PARSE_ONLY) == 0)
     {
-      ret_code = jerry_run (&error_obj_p);
+      jerry_api_value_t error_value;
+      ret_code = jerry_run (&error_value);
+
+      if (jerry_api_value_is_object (&error_value)
+          || jerry_api_value_is_string (&error_value))
```
akiss77: These checks seem to be superfluous here since `jerry_api_release_value` repeats them (and is a no-op if its argument is not an object or a string). AFAICR that was the point of introducing `jerry_api_release_value`.

jerry-core/jerry.c
```diff
@@ -1774,11 +1778,18 @@ jerry_parse (const jerry_api_char_t *source_p, /**< script source */
  * @return completion status
  */
 jerry_completion_code_t
-jerry_run (jerry_api_object_t **error_obj_p)
+jerry_run (jerry_api_value_t *error_value_p)
```
akiss77: Since we expect the caller to release the out param `error_value_p`, this should be mentioned in the doc comment. Elsewhere in this file we have notes like:
```c
  * Note:
  *      returned value should be freed with jerry_api_release_value
  *      just when the value becomes unnecessary.
```

BTW, now that we are touching this function, it would be a good time to add a doc comment to the parameter as well.

akiss77:
Other than my two comments above, the patch looks good to me (FWIW).

jerry-core/jerry.c
LaszloLango: good point

LaszloLango:
@akiss77, fixed.

akiss77:
@LaszloLango thanks. Still LGTM (FWIW)

zherczeg:
LGTM

PR #996, opened 2016-04-08
Remove `jerry_reg_err_callback` API function

akiss77:
Discussions on Termination Port API point in a different direction
than what's already in the API. Moreover, since the function has
been throwing unimplemented assert error since its introduction,
we should remove it from the code base.


LaszloLango:
LGTM

zherczeg:
LGTM

PR #997, opened 2016-04-09
Doc comment maintenance

akiss77:
Fixed doc comments issues:

* Fixed mistyped param doc comments (`/**<` is OK, `/** <` is not).

* Put special characters (e.g., pipe, backslash, etc.) in quotes, as they can
  confuse doxygen and it will print lots of various warnings.

* Added missing `@{`s, removed extra `@}`s.

Ensured same style for doc groups everywhere:

* Where `\addtogroup`, `@{`, and `@}` doxygen commands are used, the order to be
  followed is: license, `#ifndef` guards (in headers), includes, `\addtogroup`
  and `@{`, main code content, `@}`, `#endif` guards (in headers).

* Multiple `\addtogroup`s or multiple `@}`s should be in the same doc comment.

* First `\addtogroup` should be on the very first line of a doc comment, i.e.,
  `/** \addtogroup`.

### DIFF
--- a/jerry-core/parser/regexp/re-parser.h
+++ b/jerry-core/parser/regexp/re-parser.h
@@ -35,27 +35,27 @@
 typedef enum
 {
   RE_TOK_EOF,                        /**< EOF */
-  RE_TOK_BACKREFERENCE,              /**< \[0..9] */
+  RE_TOK_BACKREFERENCE,              /**< "\[0..9]" */
   RE_TOK_CHAR,                       /**< any character */
-  RE_TOK_ALTERNATIVE,                /**< | */
-  RE_TOK_ASSERT_START,               /**< ^ */
-  RE_TOK_ASSERT_END,                 /**< $ */
-  RE_TOK_PERIOD,                     /**< . */
-  RE_TOK_START_CAPTURE_GROUP,        /**< ( */
-  RE_TOK_START_NON_CAPTURE_GROUP,    /**< (?: */
-  RE_TOK_END_GROUP,                  /**< ')' */
-  RE_TOK_ASSERT_START_POS_LOOKAHEAD, /**< (?= */
-  RE_TOK_ASSERT_START_NEG_LOOKAHEAD, /**< (?! */
-  RE_TOK_ASSERT_WORD_BOUNDARY,       /**< \b */
-  RE_TOK_ASSERT_NOT_WORD_BOUNDARY,   /**< \B */
-  RE_TOK_DIGIT,                      /**< \d */
-  RE_TOK_NOT_DIGIT,                  /**< \D */
-  RE_TOK_WHITE,                      /**< \s */
-  RE_TOK_NOT_WHITE,                  /**< \S */
-  RE_TOK_WORD_CHAR,                  /**< \w */
-  RE_TOK_NOT_WORD_CHAR,              /**< \W */
-  RE_TOK_START_CHAR_CLASS,           /**< [ ] */
-  RE_TOK_START_INV_CHAR_CLASS,       /**< [^ ] */
+  RE_TOK_ALTERNATIVE,                /**< "|" */
+  RE_TOK_ASSERT_START,               /**< "^" */
+  RE_TOK_ASSERT_END,                 /**< "$" */
+  RE_TOK_PERIOD,                     /**< "." */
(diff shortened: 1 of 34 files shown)


LaszloLango:
There is a doc comment fix in my in-progress PR too. Could you add that small fix to this? It fits better here. https://github.com/Samsung/jerryscript/pull/985/files#diff-809091506ab5ef21bfcf6883f66a2ddbR748

akiss77:
@LaszloLango , Of course. Added, updated.

jerry-core/parser/regexp/re-parser.h
```diff
@@ -35,27 +35,27 @@
...
+  RE_TOK_ASSERT_WORD_BOUNDARY,       /**< "\b" */
+  RE_TOK_ASSERT_NOT_WORD_BOUNDARY,   /**< "\B" */
+  RE_TOK_DIGIT,                      /**< "\d" */
+  RE_TOK_NOT_DIGIT,                  /**< "\D" */
+  RE_TOK_WHITE,                      /**< "\s" */
+  RE_TOK_NOT_WHITE,                  /**< "\S" */
+  RE_TOK_WORD_CHAR,                  /**< "\w" */
+  RE_TOK_NOT_WORD_CHAR,              /**< "\W" */
+  RE_TOK_START_CHAR_CLASS,           /**< "[ ]" */
+  RE_TOK_START_INV_CHAR_CLASS,       /**< "[^ ]" */
```
zherczeg: Why do we need to put quotes around?

akiss77: Otherwise doxygen prints warnings like these:
```
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:53: warning: Found unknown command `\s'
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:56: warning: Found unknown command `\W'
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:50: warning: Found unknown command `\B'
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:52: warning: Found unknown command `\D'
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:55: warning: Found unknown command `\w'
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:54: warning: Found unknown command `\S'
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:51: warning: Found unknown command `\d'
/home/akiss/devel/jerry/jerryscript/jerry-core/parser/regexp/re-parser.h:44: warning: End of list marker found without any preceding list items
```

zherczeg: Is there a way to disable this in doxigen perhaps? We don't use any \ commands at all.

akiss77: I haven't found any such configuration option for doxygen. So, no, we cannot have `@` commands only. (BTW, we do have `\` commands, e.g., `\addtogroup` is everywhere in the sources.) (BTW2, we already had such an issue in re-bytecode.h. BTW3, it seems that the quoting hasn't been applied completely consequently there, my bad.)

LaszloLango:
LGTM

akiss77:
Updated to current master. Moreover, I've added an extra patch on top of the originally proposed one that adds consistency to the quoting of special characters in re-bytecode.h . (If this also gets accepted, then it can be squashed. If not, it will be easier to drop.)

LaszloLango:
Still LGTM

jerry-core/parser/regexp/re-parser.h
zherczeg: If I understand correctly the problem is the \, the others don't need quoting. Perhaps we could replace \ to \\ or BS.

akiss77: It's not the backslash only. E.g., 
```
re-parser.h:44: warning: End of list marker found without any preceding list items
```
There, it was the open parenthesis. And we have had trouble with `$` and `^` before, if I remember correctly.

(If your really don't like the quotes, we will have to revert them not only in this patch but wherever they have already been applied.)

zherczeg:
LGTM

PR #999, opened 2016-04-11
Fix syscall assertion when input file does not exist

dbatyai:
Errors should be handled by caller instead of exiting.

### DIFF
--- a/jerry-libc/target/darwin/jerry-libc-target.c
+++ b/jerry-libc/target/darwin/jerry-libc-target.c
@@ -33,15 +33,6 @@
 
 LIBC_UNREACHABLE_STUB_FOR (int raise (int sig_no __attr_unused___))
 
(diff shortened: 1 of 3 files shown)


akiss77:
It seems that you've removed `LIBC_EXIT_ON_ERROR` from the linux port only. I feel that we should keep the darwin port in sync as well.
Moreover, once the suggested changes are applied, will there be any use of `LIBC_EXIT_ON_ERROR`? (If not, that can go away too.)

dbatyai:
I've updated the darwin port, and removed `LIBC_EXIT_ON_ERROR` entirely.

zherczeg:
LGTM

akiss77:
LGTM (FWIW)

akiss77:
(Please, update to current master.)

PR #1000, opened 2016-04-11
Add help flag and usage to UNIX Jerry standalone.

dbatyai:

### DIFF
--- a/main-unix.c
+++ b/main-unix.c
@@ -188,6 +188,36 @@ assert_handler (const jerry_api_object_t *function_obj_p __attr_unused___, /** <
   }
 } /* assert_handler */
 
+static void
+print_usage (char *name)
+{
+  printf ("Usage: %s [OPTION]... [FILE]...\n"
+          "Try '%s --help' for more information.\n",
+          name, name);
+} /* print_usage */
+
+static void
+print_help (char *name)
+{
+  printf ("Usage: %s [OPTION]... [FILE]...\n"
+          "\n"
+          "Options:\n"
+          "  -h, --help\n"
+          "  -v, --version\n"
+          "  --mem-stats\n"
+          "  --mem-stats-separate\n"
+          "  --parse-only\n"
+          "  --show-opcodes\n"
+          "  --save-snapshot-for-global FILE\n"
+          "  --save-snapshot-for-eval FILE\n"
+          "  --exec-snapshot FILE\n"
+          "  --log-level [0-3]\n"
+          "  --log-file FILE\n"
+          "  --abort-on-fail\n"
+          "\n",
+          name);
+} /* print_help */
+
 int
 main (int argc,
       char **argv)
@@ -226,12 +256,18 @@ main (int argc,
 #endif /* JERRY_ENABLE_LOG */
   for (i = 1; i < argc; i++)
   {
-    if (!strcmp ("-v", argv[i]))
+    if (!strcmp ("-h", argv[i]) || !strcmp ("--help", argv[i]))
+    {
+      print_help (argv[0]);
+      return JERRY_STANDALONE_EXIT_CODE_OK;
+    }
+    else if (!strcmp ("-v", argv[i]) || !strcmp ("--version", argv[i]))
     {
       printf ("Build date: \t%s\n", jerry_build_date);
       printf ("Commit hash:\t%s\n", jerry_commit_hash);
       printf ("Branch name:\t%s\n", jerry_branch_name);
       printf ("\n");
+      return JERRY_STANDALONE_EXIT_CODE_OK;
     }
     else if (!strcmp ("--mem-stats", argv[i]))
     {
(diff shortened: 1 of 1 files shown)


main-unix.c
```diff
@@ -188,6 +188,27 @@ assert_handler (const jerry_api_object_t *function_obj_p __attr_unused___, /** <
   }
 } /* assert_handler */
 
+static void
+print_usage ()
+{
+  printf ("Usage: ./jerry [flags] [js-file ...]\n"
```
akiss77: Better to use `argv[0]` instead of `./jerry`, IMHO.

main-unix.c
```diff
@@ -199,6 +220,11 @@ main (int argc,
 
     return JERRY_STANDALONE_EXIT_CODE_FAIL;
   }
+  else if (argc < 2)
```
akiss77: This is no good! We couldn't go into repl mode then.

main-unix.c
```diff
@@ -188,6 +188,27 @@ assert_handler (const jerry_api_object_t *function_obj_p __attr_unused___, /** <
...
+  printf ("Usage: ./jerry [flags] [js-file ...]\n"
+          "\n"
+          "Flags:\n"
+          "  -h, --help\n"
+          "  -v, --version\n"
+          "  --mem-stats\n"
+          "  --mem-stats-separate\n"
+          "  --parse-only\n"
+          "  --show-opcodes\n"
+          "  --save-snapshot-for-global $output_file\n"
```
akiss77: This `$` notation of the required argument is a bit unorthodox to me.

main-unix.c
```diff
@@ -308,13 +342,19 @@ main (int argc,
...
+        print_usage ();
         return JERRY_STANDALONE_EXIT_CODE_FAIL;
       }
     }
     else if (!strcmp ("--abort-on-fail", argv[i]))
     {
       flags |= JERRY_FLAG_ABORT_ON_FAIL;
     }
+    else if (!strncmp ("-", argv[i], 1))
+    {
```
akiss77: I think we need an error message here to let the user know why we are exiting. E.g., `"Error: unrecognized option: %s\n", argv[i]` .

BTW, might be worth revisiting the other error messages of arg handling as well and improve them a little. E.g, `"Error: wrong format of the arguments\n"` doesn't say much at `--save-snapshot-for-global`. We could have something like:
```c
if (save_snapshot_file_name_p != NULL)
{
  JERRY_ERROR_MSG ("Error: snapshot file name already specified\n");
  return JERRY_STANDALONE_EXIT_CODE_FAIL
}
if (++i >= argc)
{
  JERRY_ERROR_MSG ("Error: missing argument of %s\n", argv[i-1]);
  return JERRY_STANDALONE_EXIT_CODE_FAIL
}
```

Similarly elsewhere.

main-unix.c
```diff
@@ -308,13 +342,19 @@ main (int argc,
...
+      print_usage ();
```
akiss77: It seems that you have missed `--log-level`. At least, `print_usage ();` is not inserted there after the error message.

Moreover, I'm a bit concerned about printing the whole long usage info whenever something is wrong. Well-established tools normally just print a short hint in such cases, e.g.,
```
$ grep --XXX
grep: unrecognized option '--XXX'
Usage: grep [OPTION]... PATTERN [FILE]...
Try 'grep --help' for more information.
```

(Help messages of such tools may also give ideas how to fine-tune the usage string to become more alike to what users are familiar with -- see my "unorthodoxy" comment above.)

dbatyai:
@akiss77, I've updated the patch, the format now falls more in line with other tools.

main-unix.c
```diff
@@ -257,14 +293,21 @@ main (int argc,
...
+      }
+
+      if (++i < argc)
       {
         save_snapshot_file_name_p = argv[i];
       }
       else
       {
-        JERRY_ERROR_MSG ("Error: wrong format of the arguments\n");
+        JERRY_ERROR_MSG ("Error: no file specified for %s\n", argv[i-1]);
```
akiss77: Oh, my bad. We are using spaces around operators. So, it's `i - 1`. (I know, I was the one who wrote `i-1` in the review. Sorry :( )

main-unix.c
```diff
@@ -188,6 +188,36 @@ assert_handler (const jerry_api_object_t *function_obj_p __attr_unused___, /** <
   }
 } /* assert_handler */
 
+static void
+print_usage (char **argv)
```
akiss77: It is smelling a little bit bad to have an array passed without knowing its size. Especially since we are only using its 0-th element. Why not just have `char *name` as an argument?

main-unix.c
```diff
@@ -188,6 +188,36 @@ assert_handler (const jerry_api_object_t *function_obj_p __attr_unused___, /** <
...
+{
+  printf ("Usage: %s [OPTION]... [FILE]...\n"
+          "Try '%s --help' for more information.\n",
+          argv[0], argv[0]);
+} /* print_usage */
+
+static void
+print_help (char **argv)
```
akiss77: ditto

main-unix.c
```diff
@@ -257,14 +293,21 @@ main (int argc,
...
-      if (save_snapshot_file_name_p == NULL
-          && ++i < argc)
+      if (save_snapshot_file_name_p != NULL)
+      {
+        JERRY_ERROR_MSG ("Error: snapshot file name already specified\n");
+        print_usage (argv);
+        return JERRY_STANDALONE_EXIT_CODE_FAIL;
+      }
+
+      if (++i < argc)
```
akiss77: One more request. Organized like this, we have check#1 -> bail out, check#2 -> do the thing, else -> bail out. I'd suggest (check -> bail out)* and if no checks failed -> do the thing (an then `else` is not needed).

Several occurrences below.

akiss77:
Even though I might seem picky (which I know I am), I think that the patch is improving in the right direction. And it *is* needed.

dbatyai:
It's okay to be picky :). Updated again.

akiss77:
Thanks for the rewrites. Now: LGTM (FWIW)

LaszloLango:
LGTM

akiss77:
(Please, rebase to current master)

dbatyai:
@akiss77, Done.

zherczeg:
LGTM

PR #1001, opened 2016-04-11
Refactor object property chain to use property pairs.

zherczeg:
The patch itself seems a step back, but the primary aim is opening future
optimization opportunities. The list of changes follows:

 - Property is changed to be an abstract type, which has type, flags,
   and a value. It does not have a name anymore and property pointers
   cannot be compressed.
 - Full (32 bit) ecma values can be property values. This allows
   using non-compressed pointers for ecma values in the future.
 - The property chain is not restricted to the same item anymore,
   it can contain hash maps, arrays in the future.

### DIFF
--- a/jerry-core/ecma/base/ecma-alloc.c
+++ b/jerry-core/ecma/base/ecma-alloc.c
@@ -21,8 +21,12 @@
 #include "jrt.h"
 #include "mem-poolman.h"
 
-JERRY_STATIC_ASSERT (sizeof (ecma_property_t) <= sizeof (uint64_t),
-                     size_of_ecma_property_t_must_be_less_than_or_equal_to_8_bytes);
+JERRY_STATIC_ASSERT (sizeof (ecma_property_value_t) == sizeof (ecma_value_t),
+                     size_of_ecma_property_value_t_must_be_equal_to_size_of_ecma_value_t);
+JERRY_STATIC_ASSERT (((sizeof (ecma_property_value_t) - 1) & sizeof (ecma_property_value_t)) == 0,
+                     size_of_ecma_property_value_t_must_be_power_of_2);
+JERRY_STATIC_ASSERT (sizeof (ecma_property_pair_t) == sizeof (uint64_t) * 2,
+                     size_of_ecma_property_pair_t_must_be_equal_to_16_bytes);
 
 JERRY_STATIC_ASSERT (sizeof (ecma_object_t) <= sizeof (uint64_t),
                      size_of_ecma_object_t_must_be_less_than_or_equal_to_8_bytes);
@@ -85,13 +89,32 @@ JERRY_STATIC_ASSERT (sizeof (ecma_getter_setter_pointers_t) <= sizeof (uint64_t)
   DEALLOC (ecma_type)
 
 DECLARE_ROUTINES_FOR (object)
-DECLARE_ROUTINES_FOR (property)
 DECLARE_ROUTINES_FOR (number)
 DECLARE_ROUTINES_FOR (collection_header)
 DECLARE_ROUTINES_FOR (collection_chunk)
 DECLARE_ROUTINES_FOR (string)
 DECLARE_ROUTINES_FOR (getter_setter_pointers)
 DECLARE_ROUTINES_FOR (external_pointer)
+
+/**
+ * Allocate memory for ecma-property pair
+ *
+ * @return pointer to allocated memory
+ */
+ecma_property_pair_t *
+ecma_alloc_property_pair (void)
+{
+  return mem_heap_alloc_block (sizeof (ecma_property_pair_t));
+} /* ecma_alloc_property_pair */
+
+/**
+ * Dealloc memory from an ecma-property
+ */
+extern void
+ecma_dealloc_property_pair (ecma_property_pair_t *property_pair_p) /**< property pair to be freed */
+{
+  mem_heap_free_block (property_pair_p, sizeof (ecma_property_pair_t));
+} /* ecma_dealloc_property_pair */
 
 /**
  * @}
--- a/jerry-core/ecma/base/ecma-alloc.h
+++ b/jerry-core/ecma/base/ecma-alloc.h
@@ -38,18 +38,6 @@ extern ecma_object_t *ecma_alloc_object (void);
 extern void ecma_dealloc_object (ecma_object_t *);
 
 /**
- * Allocate memory for ecma-property
- *
- * @return pointer to allocated memory
- */
-extern ecma_property_t *ecma_alloc_property (void);
-
-/**
- * Dealloc memory from an ecma-property
- */
-extern void ecma_dealloc_property (ecma_property_t *);
-
-/**
  * Allocate memory for ecma-number
  *
  * @return pointer to allocated memory
@@ -120,6 +108,18 @@ extern ecma_external_pointer_t *ecma_alloc_external_pointer (void);
 * Dealloc memory from external pointer
 */
 extern void ecma_dealloc_external_pointer (ecma_external_pointer_t *);
+
+/**
+ * Allocate memory for ecma-property pair
+ *
+ * @return pointer to allocated memory
+ */
+extern ecma_property_pair_t *ecma_alloc_property_pair (void);
+
+/**
+ * Dealloc memory from an ecma-property pair
+ */
+extern void ecma_dealloc_property_pair (ecma_property_pair_t *);
 
 /**
  * @}
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -200,6 +200,139 @@ ecma_gc_init (void)
 } /* ecma_gc_init */
 
 /**
+ * Mark referenced object from property
+ */
+static void
+ecma_gc_mark_property (ecma_property_t *property_p) /**< property */
+{
+  switch (ECMA_PROPERTY_GET_TYPE (property_p))
+  {
+    case ECMA_PROPERTY_TYPE_NAMEDDATA:
+    {
+      ecma_value_t value = ecma_get_named_data_property_value (property_p);
+
+      if (ecma_is_value_object (value))
+      {
+        ecma_object_t *value_obj_p = ecma_get_object_from_value (value);
+
+        ecma_gc_set_object_visited (value_obj_p, true);
+      }
+      break;
+    }
+    case ECMA_PROPERTY_TYPE_NAMEDACCESSOR:
+    {
+      ecma_object_t *getter_obj_p = ecma_get_named_accessor_property_getter (property_p);
+      ecma_object_t *setter_obj_p = ecma_get_named_accessor_property_setter (property_p);
+
+      if (getter_obj_p != NULL)
+      {
+        ecma_gc_set_object_visited (getter_obj_p, true);
+      }
+
+      if (setter_obj_p != NULL)
+      {
+        ecma_gc_set_object_visited (setter_obj_p, true);
+      }
+      break;
+    }
+    case ECMA_PROPERTY_TYPE_INTERNAL:
+    {
+      uint32_t property_value = ECMA_PROPERTY_VALUE_PTR (property_p)->value;
+
+      switch (ECMA_PROPERTY_GET_INTERNAL_PROPERTY_TYPE (property_p))
+      {
+        case ECMA_INTERNAL_PROPERTY_NUMBER_INDEXED_ARRAY_VALUES: /* a collection of ecma values */
+        case ECMA_INTERNAL_PROPERTY_STRING_INDEXED_ARRAY_VALUES: /* a collection of ecma values */
+        {
+          JERRY_UNIMPLEMENTED ("Indexed array storage is not implemented yet.");
+        }
+
+        case ECMA_INTERNAL_PROPERTY_PROTOTYPE: /* the property's value is located in ecma_object_t
+                                                * (see above in the routine) */
+        case ECMA_INTERNAL_PROPERTY_EXTENSIBLE: /* the property's value is located in ecma_object_t
+                                                 * (see above in the routine) */
+        case ECMA_INTERNAL_PROPERTY__COUNT: /* not a real internal property type,
+                                             * but number of the real internal property types */
+        {
+          JERRY_UNREACHABLE ();
+        }
+
+        case ECMA_INTERNAL_PROPERTY_PRIMITIVE_STRING_VALUE: /* compressed pointer to a ecma_string_t */
+        case ECMA_INTERNAL_PROPERTY_PRIMITIVE_NUMBER_VALUE: /* compressed pointer to a ecma_number_t */
+        case ECMA_INTERNAL_PROPERTY_PRIMITIVE_BOOLEAN_VALUE: /* a simple boolean value */
+        case ECMA_INTERNAL_PROPERTY_CLASS: /* an enum */
+        case ECMA_INTERNAL_PROPERTY_CODE_BYTECODE: /* compressed pointer to a bytecode array */
+        case ECMA_INTERNAL_PROPERTY_NATIVE_CODE: /* an external pointer */
+        case ECMA_INTERNAL_PROPERTY_NATIVE_HANDLE: /* an external pointer */
+        case ECMA_INTERNAL_PROPERTY_FREE_CALLBACK: /* an object's native free callback */
+        case ECMA_INTERNAL_PROPERTY_BUILT_IN_ID: /* an integer */
+        case ECMA_INTERNAL_PROPERTY_BUILT_IN_ROUTINE_DESC: /* an integer */
+        case ECMA_INTERNAL_PROPERTY_EXTENSION_ID: /* an integer */
+        case ECMA_INTERNAL_PROPERTY_NON_INSTANTIATED_BUILT_IN_MASK_0_31: /* an integer (bit-mask) */
+        case ECMA_INTERNAL_PROPERTY_NON_INSTANTIATED_BUILT_IN_MASK_32_63: /* an integer (bit-mask) */
+        case ECMA_INTERNAL_PROPERTY_REGEXP_BYTECODE:
+        {
+          break;
+        }
+
+        case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_BOUND_THIS: /* an ecma value */
+        {
+          if (ecma_is_value_object (property_value))
+          {
+            ecma_object_t *obj_p = ecma_get_object_from_value (property_value);
+
+            ecma_gc_set_object_visited (obj_p, true);
+          }
+
+          break;
+        }
+
+        case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_BOUND_ARGS: /* a collection of ecma values */
+        {
+          ecma_collection_header_t *bound_arg_list_p = ECMA_GET_NON_NULL_POINTER (ecma_collection_header_t,
+                                                                                  property_value);
+
+          ecma_collection_iterator_t bound_args_iterator;
+          ecma_collection_iterator_init (&bound_args_iterator, bound_arg_list_p);
+
+          for (ecma_length_t i = 0; i < bound_arg_list_p->unit_number; i++)
+          {
+            bool is_moved = ecma_collection_iterator_next (&bound_args_iterator);
+            JERRY_ASSERT (is_moved);
+
+            if (ecma_is_value_object (*bound_args_iterator.current_value_p))
+            {
+              ecma_object_t *obj_p = ecma_get_object_from_value (*bound_args_iterator.current_value_p);
+
+              ecma_gc_set_object_visited (obj_p, true);
+            }
+          }
+
+          break;
+        }
+
+        case ECMA_INTERNAL_PROPERTY_BOUND_FUNCTION_TARGET_FUNCTION: /* an object */
+        case ECMA_INTERNAL_PROPERTY_SCOPE: /* a lexical environment */
+        case ECMA_INTERNAL_PROPERTY_PARAMETERS_MAP: /* an object */
+        {
+          ecma_object_t *obj_p = ECMA_GET_NON_NULL_POINTER (ecma_object_t, property_value);
+
+          ecma_gc_set_object_visited (obj_p, true);
+
+          break;
+        }
+      }
+      break;
+    }
+    default:
+    {
+      JERRY_UNREACHABLE ();
+      break;
+    }
+  }
+} /* ecma_gc_mark_property */
+
+/**
  * Mark objects as visited starting from specified object as root
  */
 void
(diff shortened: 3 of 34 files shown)


zherczeg:
Unfortunately the patch is big, 142K.

LaszloLango:
#### RPi2 results

Base rev: 920a9ee

                                        3d-cube.js |   2.482s ->   2.408s :  +2.982%  (+-1.266%) : [+]
                            access-binary-trees.js |   1.254s ->   1.210s :  +3.509%  (+-0.868%) : [+]
                                access-fannkuch.js |   7.548s ->   7.144s :  +5.352%  (+-1.277%) : [+]
                                   access-nbody.js |   2.812s ->   2.672s :  +4.979%  (+-1.106%) : [+]
                       bitops-3bit-bits-in-byte.js |   1.616s ->   1.566s :  +3.094%  (+-0.972%) : [+]
                            bitops-bits-in-byte.js |   2.440s ->   2.316s :  +5.082%  (+-0.925%) : [+]
                             bitops-bitwise-and.js |   3.442s ->   3.090s : +10.227%  (+-0.645%) : [+]
                             bitops-nsieve-bits.js |  18.046s ->  13.356s : +25.989%  (+-0.875%) : [+]
                          controlflow-recursive.js |   0.930s ->   0.910s :  +2.151%  (+-0.000%) : [+]
                                     crypto-aes.js |   3.278s ->   3.138s :  +4.271%  (+-1.204%) : [+]
                                     crypto-md5.js |   3.252s ->   2.604s : +19.926%  (+-0.655%) : [+]
                                    crypto-sha1.js |   2.034s ->   1.754s : +13.766%  (+-1.249%) : [+]
                              date-format-tofte.js |   2.006s ->   1.942s :  +3.190%  (+-0.712%) : [+]
                              date-format-xparb.js |   0.790s ->   0.778s :  +1.519%  (+-1.166%) : [+]
                                    math-cordic.js |   2.844s ->   2.768s :  +2.672%  (+-0.504%) : [+]
                              math-partial-sums.js |   1.628s ->   1.474s :  +9.460%  (+-0.861%) : [+]
                             math-spectral-norm.js |   1.424s ->   1.374s :  +3.511%  (+-1.101%) : [+]
                                  string-base64.js |   4.914s ->   4.932s :  -0.366%  (+-1.394%) : [≈]
                                   string-fasta.js |   3.198s ->   3.054s :  +4.503%  (+-1.070%) : [+]
                                   Geometric mean: |                   Speed up: 6.871% (+-0.243%) : [+]

Binary size:
base: 179896 bytes
branch: 183096 bytes

jerry-core/ecma/base/ecma-alloc.c
```diff
@@ -94,6 +97,26 @@ DECLARE_ROUTINES_FOR (getter_setter_pointers)
 DECLARE_ROUTINES_FOR (external_pointer)
 
 /**
+ * Allocate memory for ecma-property pair
+ *
+ * @return pointer to allocated memory
+ */
+extern ecma_property_pair_t *
```
akiss77: I think we are not using `extern` in sources but in headers only.

jerry-core/ecma/base/ecma-alloc.c
```diff
@@ -94,6 +97,26 @@ DECLARE_ROUTINES_FOR (getter_setter_pointers)
 DECLARE_ROUTINES_FOR (external_pointer)
 
 /**
+ * Allocate memory for ecma-property pair
+ *
+ * @return pointer to allocated memory
```
LaszloLango: What if allocation fails? Bails out or return NULL?

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -199,6 +199,136 @@ ecma_gc_init (void)
...
+    case ECMA_PROPERTY_TYPE_INTERNAL:
+    {
+      uint32_t property_value = ECMA_OBJECT_PROPERTY_VALUE (property_p)->value;
+
+      switch (ECMA_OBJECT_GET_INTERNAL_PROPERTY_TYPE (property_p))
+      {
+        case ECMA_INTERNAL_PROPERTY_NUMBER_INDEXED_ARRAY_VALUES: /* a collection of ecma values */
+        case ECMA_INTERNAL_PROPERTY_STRING_INDEXED_ARRAY_VALUES: /* a collection of ecma values */
+        {
+          JERRY_UNIMPLEMENTED ("Indexed array storage is not implemented yet.");
```
LaszloLango: Does it mean that the patch need future work? Is there any detailed description for this (issue or TODO comment)?

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -199,6 +199,136 @@ ecma_gc_init (void)
...
+        }
+
+        case ECMA_INTERNAL_PROPERTY_PROTOTYPE: /* the property's value is located in ecma_object_t
+                                                * (see above in the routine) */
+        case ECMA_INTERNAL_PROPERTY_EXTENSIBLE: /* the property's value is located in ecma_object_t
+                                                 * (see above in the routine) */
+        case ECMA_INTERNAL_PROPERTY__COUNT: /* not a real internal property type,
```
LaszloLango: Default case do the same. I'd move these into the default case as an assert.

LaszloLango: Ignore my prev. comment. That default belongs to the outer switch. :) I missed that.

jerry-core/ecma/base/ecma-gc.c
zherczeg: That code is copied. I don't want to change that in this patch. I want to work on the internal properties in the next patch.

jerry-core/ecma/base/ecma-alloc.c
zherczeg: In Jerry a failed allocation stops the engine. Other allocator functions has this return value specification, I didin't want to change the custom.

zherczeg:
Patch updated.

jerry-core/ecma/base/ecma-alloc.c
LaszloLango: Could you extend the comment of this function with this note?

jerry-core/ecma/base/ecma-gc.c
LaszloLango: OK

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -227,56 +306,61 @@ typedef struct
...
-  } v;
-} ecma_property_t;
+/**
+ * Returns true if the property pointer is a property pair.
+ */
+#define ECMA_OBJECT_IS_PROPERTY_PAIR(property_header_p) \
+  (ECMA_OBJECT_GET_PROPERTY_TYPE (&((property_header_p)->first_type)) <= ECMA_PROPERTY_TYPE_PROPERTY_PAIR__MAX)
+
+/**
+ * Returns the internal property type
```
LaszloLango: Missing dot at end of statement. :)

LaszloLango:
LGTM

zherczeg:
Renamed the remaining ECMA_OBJECT_ prefixed macros to ECMA_PROPERTY_ macros

LaszloLango:
still LGTM

jerry-core/ecma/base/ecma-alloc.c
akiss77: I'd keep doc comment as is. This function simply wraps `mem_heap_alloc_block`. Should its functionality change, e.g., may we prepare for OOM situations in the (not too near) future, then this function will change its behaviour as well. And then the suggested note would become invalid.

jerry-core/ecma/base/ecma-alloc.h
```diff
@@ -38,16 +38,16 @@ extern ecma_object_t *ecma_alloc_object (void);
...
  */
-extern ecma_property_t *ecma_alloc_property (void);
+extern ecma_property_pair_t *ecma_alloc_property_pair (void);
 
 /**
- * Dealloc memory from an ecma-property
+ * Dealloc memory from an ecma-property pair
  */
-extern void ecma_dealloc_property (ecma_property_t *);
+extern void ecma_dealloc_property_pair (ecma_property_pair_t *);
```
akiss77: Please, move both these declarations to the end of the header file to match the order of functions in the C source.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -199,6 +199,136 @@ ecma_gc_init (void)
   ecma_gc_new_objects_since_last_gc = 0;
 } /* ecma_gc_init */
 
+static void
+ecma_gc_mark_property (ecma_property_t *property_p) /**< property */
```
akiss77: This function most definitely needs a doc comment.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -204,20 +204,99 @@ typedef enum
...
-  ECMA_PROPERTY_FLAG_CONFIGURABLE = 1u << 3, /**< property is configurable */
-  ECMA_PROPERTY_FLAG_ENUMERABLE = 1u << 4, /**< property is enumerable */
-  ECMA_PROPERTY_FLAG_WRITABLE = 1u << 5, /**< property is writable */
-  ECMA_PROPERTY_FLAG_LCACHED = 1u << 6, /**< property is lcached */
+  ECMA_PROPERTY_TYPE_DELETED, /**< deleted property */
+  ECMA_PROPERTY_TYPE_INTERNAL, /**< internal property */
+  ECMA_PROPERTY_TYPE_NAMEDDATA, /**< property is named data */
+  ECMA_PROPERTY_TYPE_NAMEDACCESSOR, /**< property is named accessor */
+
+  ECMA_PROPERTY_TYPE_HASH, /**< property hash (TODO: implement this) */
```
akiss77: Right now, this enum label (`ECMA_PROPERTY_TYPE_HASH`) is not used anywhere. (Except for the `__MAX` label.) Please, remove this from the PR and add it back in another PR once there is an implementation that makes use of it.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -204,20 +204,99 @@ typedef enum
...
+ *
+ * A property is a type_and_flags byte and an ecma_value_t value pair.
+ * This pair is represented by a single pointer in JerryScript. Although
+ * a packed struct would only consume sizeof(ecma_value_t)+1 memory
+ * bytes, accessing such structure is inefficient from the CPU viewpoint
+ * because the value is not naturally aligned. To improve performance,
+ * type bytes and values are packed together. It ecma_value_t is four
+ * bytes long, maximum four properties can be packed together:
+ *
+ *  [Type 1][Type 2][Type 3][Type 4][Value 1][Value 2][Value 3][Value 4]
```
akiss77: This explanation can be a bit misleading. How many ecma values are used in the above example? My guess is 5. If I'm right, then I'd suggest writing the above example as `[Type 1,Type 2,Type 3,Type 4][Value 1][Value 2][Value 3][Value 4]` so that the brackets always denote `ecma_value_t` boundaries.

(If I'm wrong, then we need an even better explanation...)

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -204,20 +204,99 @@ typedef enum
...
+} ecma_property_t;
+
+/**
+ * Property header for all items in a property list.
+ */
+typedef struct
+{
+  ecma_property_t first_type; /**< first property type byte */
+  ecma_property_t second_type; /**< second property type byte
+                                *   might be unused for certain property types */
```
akiss77: Having worked my way through most of the patch, I'm not sure that this comment is right. (Or, I might be misunderstanding something.) I couldn't find how the used/unused status of this second field depends on the type of the property. As I understood it, the nice feature of this structure is that multiple 1-byte-sized `type_n_flag`s are packed after each other. But any `type_and_flag` item can contain any valid types.

At least, `ecma_create_property` puts first property into the second slot of a pair, the next into the first slot, for a third one, it creates another pair and uses its second slot, the fourth prop goes again into the first slot, and so on. (If I get it right.) But there seems to be no dependecy on the type of the property itself.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -227,56 +306,61 @@ typedef struct
...
+/**
+ * Property pair.
+ */
+typedef struct
+{
+  ecma_property_header_t header; /**< header of the property */
+  ecma_property_value_t first_value; /**< first property value */
+  ecma_property_value_t second_value; /**< second property value */
+  mem_cpointer_t first_name_cp; /**< first property name */
+  mem_cpointer_t second_name_cp; /**< second property name */
```
akiss77: I wonder whether packing together 2 properties is the most efficient approach? Why not 4 or 6? 6 would keep alignments right as well.

I have a probable answer to that myself: having `third_`, `fourth_`, `fifth_`, `sixth_` etc. fields (1) would be ugly and (2) would blow up the code. As I've already seen quite some code duplications and special case handlings below to deal with first and second fields.

I also have a potential treatment for that. Why not use arrays both in `ecma_property_header_t` and here (in `ecma_property_pair_t`) with a constant size? It can be 2, but can also be configured to another number (e.g., to 6). It could also have the benefit to rewrite code dups to loops (e.g., in `ecma_delete_property`, `ecma_op_object_get_property_names`). I have the gut feeling that this approach would keep all the benefits of yours and result in cleaner and easier to maintain code. (I'd guess that the compiler would unroll loops of 2, so performance should not be hindered.)

What do you think about this idea?

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -227,56 +306,61 @@ typedef struct
...
+} ecma_property_value_t;
 
-  /** Property's flags (ecma_property_flags_t) */
-  uint8_t flags;
+/**
+ * Property pair.
+ */
+typedef struct
+{
+  ecma_property_header_t header; /**< header of the property */
```
akiss77: I'm not sure why this needs to be in a separate structure? Cannot we have the contents of `ecma_property_header_t` right here?

jerry-core/ecma/base/ecma-globals.h
zherczeg: That is correct. It cannot be that misleading if you understand it :) But I try to improve it a bit.

jerry-core/ecma/base/ecma-globals.h
zherczeg: This is a property header conveniently organized for exploiting its layout later. The only thing matters here is that it has 3 bit for type and a compressed pointer for the next item. That is all.

jerry-core/ecma/base/ecma-globals.h
zherczeg: Two values minimizes memory consumption. We can try out other layouts later (this is not the aim of the patch).

jerry-core/ecma/base/ecma-globals.h
zherczeg: The ecma_property_header_t will be used in other structures in future patches.

jerry-core/ecma/base/ecma-globals.h
akiss77: I'n not sure about your reply here. I was asking why the comment of `second_type` says that it might be unused for some property types when this is not visible from the other parts of the PR?

jerry-core/ecma/base/ecma-globals.h
akiss77: Then, please, give a very short heads-up what will make use of the same header. Right now, what we have in this PR is but a linked list of pairs of properties, both of which have `type`, `value`, and `name_cp` fields organized in a layout efficient way. What's coming up next?

jerry-core/ecma/base/ecma-globals.h
akiss77: But if the structure remains the way proposed in the PR then it will be hardly extensible (effectively meaning no chance for "try out other layouts later"). What's blocking moving to a 

```c
typedef struct
{
  ecma_property_t types[2]; /**< property type bytes */
} ecma_property_header_t;

typedef struct
{
  ecma_property_header_t header; /**< header of the property */
  ecma_property_value_t values[2]; /**< property values */
  mem_cpointer_t name_cp[2]; /**< property names */
} ecma_property_pair_t;
```

style? This has really no semantic difference to the PR's approach.

zherczeg: I am doing this right now. But we need to test it whether it is better.

jerry-core/ecma/base/ecma-globals.h
zherczeg: Hash array (what you requested to remove) and arrays (they don't need string representation) are planned now.

akiss77: That was a bit too short of a heads-up. Please, explain how a hash array and/or an array will re-use the type fields of `ecma_property_header_t`?

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -204,20 +204,97 @@ typedef enum
...
+
+/**
+ * Property header for all items in a property list.
+ */
+typedef struct
+{
+  ecma_property_t types[2]; /**< two property type slot. The first represent
+                             *   the type of this property (e.g. property pair) */
+  mem_cpointer_t next_property_cp; /**< next cpointer */
+} ecma_property_header_t;
```
akiss77: This looks much cleaner to me. One more question/request, however. Would it be possible to have the 2 as a non-hard-wired constant but as a macro? And the rest of the code is also full with for loops with `i < 2`, those could use that macro as well.

Oh, and I still miss the explanation to my now-hidded question: how do hashes and array reuse this header struct?

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -237,130 +370,24 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
...
+    ecma_property_header_t *prop_iter_p = ecma_get_property_list (object_p);
+
+    while (prop_iter_p != NULL)
     {
-      next_property_p = ECMA_GET_POINTER (ecma_property_t,
-                                          property_p->next_property_p);
+      JERRY_ASSERT (ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
 
-      if (property_p->flags & ECMA_PROPERTY_FLAG_NAMEDDATA)
+      if (prop_iter_p->types[0].type_and_flags != ECMA_PROPERTY_TYPE_DELETED)
```
akiss77: These two ifs are dups of each other. Could go in a loop and use the index var instead of hard-wired 0 and 1.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -421,15 +473,26 @@ ecma_find_internal_property (ecma_object_t *object_p, /**< object descriptor */
...
-       property_p = ECMA_GET_POINTER (ecma_property_t, property_p->next_property_p))
+  ecma_property_header_t *prop_iter_p = ecma_get_property_list (object_p);
+
+  while (prop_iter_p != NULL)
   {
-    if ((property_p->flags & ECMA_PROPERTY_FLAG_INTERNAL)
-        && property_p->h.internal_property_type == property_id)
+    JERRY_ASSERT (ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
+
+    if (ECMA_PROPERTY_GET_TYPE (&prop_iter_p->types[0]) == ECMA_PROPERTY_TYPE_INTERNAL
```
akiss77: These two ifs are also clones.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -842,34 +875,61 @@ ecma_free_property (ecma_object_t *object_p, /**< object the property belongs to
...
+
+        if (cur_prop_p->types[1 - i].type_and_flags != ECMA_PROPERTY_TYPE_DELETED)
+        {
+          /* The other property is still valid. */
+          return;
+        }
 
-      return;
+        JERRY_ASSERT (cur_prop_p->types[0].type_and_flags == ECMA_PROPERTY_TYPE_DELETED
+                      && cur_prop_p->types[1].type_and_flags == ECMA_PROPERTY_TYPE_DELETED);
```
akiss77: This assert checks types_and_flags twice. It's enough to assert on `cur_prop_p->types[i].type_and_flags == ECMA_PROPERTY_TYPE_DELETED` since we are in the loop.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -880,15 +940,22 @@ ecma_assert_object_contains_the_property (const ecma_object_t *object_p, /**< ec
...
+
+  while (prop_iter_p != NULL)
   {
-    if (prop_iter_p == prop_p)
+    JERRY_ASSERT (ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
+
+    ecma_property_pair_t *prop_pair_p = (ecma_property_pair_t *) prop_iter_p;
+
+    if ((prop_pair_p->header.types + 0) == prop_p
+        || (prop_pair_p->header.types + 1) == prop_p)
```
akiss77: Please, implement this check with a loop. (Also, we are in debug code here, so performance is not an issue.)

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -399,14 +426,41 @@ ecma_gc_sweep (ecma_object_t *object_p) /**< object to free */
...
+          if (name_p != NULL)
+          {
+            ecma_deref_ecma_string (name_p);
+          }
+        }
+      }
+
+      /* Both must be deleted. */
+      JERRY_ASSERT (prop_iter_p->types[0].type_and_flags == ECMA_PROPERTY_TYPE_DELETED
+                    && prop_iter_p->types[1].type_and_flags == ECMA_PROPERTY_TYPE_DELETED);
```
akiss77: This assert can be moved into the loop above chacking `prop_iter_p->types[i].type_and_flags == ECMA_PROPERTY_TYPE_DELETED`.

akiss77:
Looks much better to me. As my in-code comments may show, I'd suggest using as few hard-wired constants (especially 1s and 2s, i.e., last index and index limit) as possible. Some `COUNT` define for the limit and `COUNT - 1` for the last item would be safer.

jerry-core/ecma/base/ecma-globals.h
zherczeg: It cannot, because ecma_property_header_t must be 4 byte long. I have no plans to change it in the near future. You can work on extending this if you wish.

The point of the header is that everything in the property chain must start with it, and algorithms can ignore those property items which are unknown to them.

```
struct {
  ecma_property_header_t header;
  fields
} hash;
```

jerry-core/ecma/base/ecma-gc.c
zherczeg: According to the latest measurements the latest changes are already slow down Jerry. I worry that these loops are not that efficient as we would like, and I am reluctant to add more of them.

jerry-core/ecma/base/ecma-globals.h
akiss77: Let's skip the first part then and focus on the second. I've already guessed that multiple structs would share the same first item (header). But how would those "fields" in a hash look like? Or in an array? How would they reuse the two first type bytes of the header?

zherczeg: I don't know yet. Probably I will use it for flags. 2 bytes are 16 bits, and the type require 3 bits. So the extension can use 13 bits for anything.

jerry-core/ecma/base/ecma-helpers.c
zherczeg: I prefer to keep it. It is an assert, so it does not hurt release code.

jerry-core/ecma/base/ecma-helpers.c
zherczeg: I don't want to change performance critical paths. However, I am willing to add asserts to these places.

There is another thing. Pair means two. If we want to add more items to a property list, it will not be called pair anymore, and the code needs to be refactored anyway.

zherczeg:
I hope I fixed the issues.

jerry-core/ecma/base/ecma-gc.c
akiss77: I'm not sure why this has been omitted.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -204,20 +204,102 @@ typedef enum
...
+ */
+typedef struct
+{
+  uint8_t type_and_flags; /**< ecma_property_types_t (3 bit) and ecma_property_flags_t */
+} ecma_property_t;
+
+/**
+ * Number of items in a property pair.
+ */
+#define ECMA_PROPERTY_PAIR_ITEM_COUNT 2
```
akiss77: Thanks

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -842,34 +879,63 @@ ecma_free_property (ecma_object_t *object_p, /**< object the property belongs to
...
 
-      return;
+        if (cur_prop_p->types[1 - i].type_and_flags != ECMA_PROPERTY_TYPE_DELETED)
+        {
+          /* The other property is still valid. */
+          return;
+        }
+
+        JERRY_ASSERT (cur_prop_p->types[0].type_and_flags == ECMA_PROPERTY_TYPE_DELETED
+                      && cur_prop_p->types[1].type_and_flags == ECMA_PROPERTY_TYPE_DELETED);
```
akiss77: I strongly vote for changing this assert to `types[i]`

zherczeg:
Done

jerry-core/ecma/base/ecma-gc.c
akiss77: Left unanswered

akiss77:
Let's see
LGTM FWIW

PR #1003, opened 2016-04-12
Implement `raise ()` in libc

akiss77:
It is used for defining an asserting stub for the function `raise`
only, which is not used anywhere.

### DIFF
--- a/jerry-libc/jerry-libc-defs.h
+++ b/jerry-libc/jerry-libc-defs.h
@@ -54,19 +54,4 @@ libc_fatal (const char *msg,
    } while (0)
 #endif /* !LIBC_NDEBUG */
 
-/**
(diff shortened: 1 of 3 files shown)


akiss77:
Because of issues that occured on ARM, had to shift the original focus of the PR:

The core functionality (i.e., the equivalent of `kill (getpid (), sig);`)
was already there in the implementation of `abort ()`. Now, it got
factored out to `raise ()` so that others (most importantly,
`__aeabi_ldiv0`) can call and link to it as well.

Also, removed `LIBC_UNREACHABLE_STUB_FOR` macro, as it is not used anymore.

zherczeg:
LGTM

LaszloLango:
LGTM

PR #1005, opened 2016-04-13
Turn assert functionality standard-conforming in jerry-libc

akiss77:
Introducing the `assert.h` public header with the `assert(x)` macro
and already making use of it in jerry-libc sources.


jerry-libc/include/assert.h
```diff
@@ -0,0 +1,45 @@
...
+#include <stdio.h>
+#include <stdlib.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif /* !__cplusplus */
+
+#ifndef NDEBUG
+#define assert(x) \
```
LaszloLango: I don't see why did you change the name (`LIBC_ASSERT`)?

akiss77: Because `assert.h` and the `assert` macro (and its depencendcy on the `NDEBUG` macro) are part of the C standard, while `LIBC_ASSERT` and `LIBC_NDEBUG` are not. It is not cost declaring things the standard-conforming way but has the benefit that we can expose it in a public header and make it useful to libc users.

I have to admit that I have future plans with it. I came to realize the embedders might need a proper `assert`.

LaszloLango: I see. Thanks for the explanation. Could you extend the commit message a bit?

LaszloLango:
LGTM

akiss77:
Updated the commit message as requested.

LaszloLango:
@akiss77, thanks :)

jerry-libc/jerry-libc-printf.c
```diff
@@ -252,7 +253,7 @@ libc_printf_write_d_i (FILE *stream, /**< stream pointer */
 
     case LIBC_PRINTF_ARG_LENGTH_TYPE_HIGHL:
     {
-      LIBC_UNREACHABLE ();
+      assert (!"unsupported length field L");
```
zherczeg: Is this ! is a widely used form? It looks strange to me, although I understand its purpose.

zherczeg:
I think the patch is OK in general. Is there really no standard way to assert in unreachable code except that !"string" ?

akiss77:
@zherczeg Sure, there is. It's `assert (0);` The `assert (!"msg");` form is just a/the way of adding a more descriptive message to the assert. So that when an assert hits (which ofc should never happen...), then a human readable error message is printed insted of a 0.

zherczeg:
Since asserts have effect only in debug mode, it is ok for me. It is still strange there is no standard unreachable.
LGTM

PR #1006, opened 2016-04-14
Refactor ecma value to store pointers directly in ecma values rather than compressing them.

zherczeg:
Ecma value is 32 bit long, and only 3 bits are used for type and flags. The rest is enough for encoding 8 byte aligned pointers without compressing them.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -58,7 +58,8 @@ typedef enum
   ECMA_TYPE_SIMPLE, /**< simple value */
   ECMA_TYPE_NUMBER, /**< 64-bit integer */
   ECMA_TYPE_STRING, /**< pointer to description of a string */
-  ECMA_TYPE_OBJECT /**< pointer to description of an object */
+  ECMA_TYPE_OBJECT, /**< pointer to description of an object */
+  ECMA_TYPE___MAX = ECMA_TYPE_OBJECT /** highest value for ecma types */
 } ecma_type_t;
 
 /**
@@ -85,34 +86,33 @@ typedef enum
 /**
  * Description of an ecma value
  *
- * Bit-field structure: type (2) | error (1) | value (ECMA_POINTER_FIELD_WIDTH)
+ * Bit-field structure: type (2) | error (1) | value (29)
  */
 typedef uint32_t ecma_value_t;
 
-/**
- * Value type (ecma_type_t)
- */
-#define ECMA_VALUE_TYPE_POS (0)
-#define ECMA_VALUE_TYPE_WIDTH (2)
+#if UINTPTR_MAX <= UINT32_MAX
 
 /**
- * Value is error (boolean)
+ * MEM_ALIGNMENT_LOG aligned pointers can be stored directly in ecma_value_t
  */
-#define ECMA_VALUE_ERROR_POS (ECMA_VALUE_TYPE_POS + \
-                              ECMA_VALUE_TYPE_WIDTH)
-#define ECMA_VALUE_ERROR_WIDTH (1)
+#define ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY
+
+#endif /* UINTPTR_MAX <= UINT32_MAX */
 
 /**
- * Simple value (ecma_simple_value_t) or compressed pointer to value (depending on value_type)
+ * Mask for ecma types in ecma_type_t
  */
-#define ECMA_VALUE_VALUE_POS (ECMA_VALUE_ERROR_POS + \
-                              ECMA_VALUE_ERROR_WIDTH)
-#define ECMA_VALUE_VALUE_WIDTH (ECMA_POINTER_FIELD_WIDTH)
+#define ECMA_VALUE_TYPE_MASK 0x3u
 
 /**
- * Size of ecma value description, in bits
+ * Error flag in ecma_type_t
  */
-#define ECMA_VALUE_SIZE (ECMA_VALUE_VALUE_POS + ECMA_VALUE_VALUE_WIDTH)
+#define ECMA_VALUE_ERROR_FLAG 0x4u
+
+/**
+ * Shift for value part in ecma_type_t
+ */
+#define ECMA_VALUE_SHIFT 3
 
 /**
  * Internal properties' identifiers.
--- a/jerry-core/ecma/base/ecma-helpers-value.c
+++ b/jerry-core/ecma/base/ecma-helpers-value.c
@@ -22,9 +22,6 @@
 #include "jrt-bit-fields.h"
 #include "vm-defines.h"
 
-JERRY_STATIC_ASSERT (sizeof (ecma_value_t) * JERRY_BITSINBYTE >= ECMA_VALUE_SIZE,
-                     bits_in_ecma_value_t_must_be_greater_than_or_equal_to_ECMA_VALUE_SIZE);
-
 /** \addtogroup ecma ECMA
  * @{
  *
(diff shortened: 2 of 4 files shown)


zherczeg:
```
                     3d-cube.js |   2.408s ->   2.148s : +10.797%  (+-0.919%) : [+]
         access-binary-trees.js |   1.226s ->   1.130s :  +7.830%  (+-0.848%) : [+]
             access-fannkuch.js |   6.980s ->   6.280s : +10.029%  (+-1.087%) : [+]
                access-nbody.js |   2.628s ->   2.334s : +11.187%  (+-1.035%) : [+]
    bitops-3bit-bits-in-byte.js |   1.564s ->   1.370s : +12.404%  (+-0.632%) : [+]
         bitops-bits-in-byte.js |   2.314s ->   1.984s : +14.261%  (+-0.642%) : [+]
          bitops-bitwise-and.js |   3.136s ->   2.890s :  +7.844%  (+-0.541%) : [+]
          bitops-nsieve-bits.js |  13.238s ->  12.804s :  +3.278%  (+-1.217%) : [+]
       controlflow-recursive.js |   0.920s ->   0.780s : +15.217%  (+-0.000%) : [+]
                  crypto-aes.js |   3.088s ->   2.872s :  +6.995%  (+-0.742%) : [+]
                  crypto-md5.js |   2.602s ->   2.472s :  +4.996%  (+-1.309%) : [+]
                 crypto-sha1.js |   1.750s ->   1.614s :  +7.771%  (+-1.002%) : [+]
           date-format-tofte.js |   1.936s ->   1.812s :  +6.405%  (+-1.259%) : [+]
           date-format-xparb.js |   0.780s ->   0.710s :  +8.974%  (+-0.000%) : [+]
                 math-cordic.js |   2.710s ->   2.402s : +11.365%  (+-1.071%) : [+]
           math-partial-sums.js |   1.472s ->   1.314s : +10.734%  (+-0.948%) : [+]
          math-spectral-norm.js |   1.380s ->   1.170s : +15.217%  (+-0.000%) : [+]
               string-base64.js |   4.894s ->   4.714s :  +3.678%  (+-0.522%) : [+]
                string-fasta.js |   2.878s ->   2.668s :  +7.297%  (+-1.300%) : [+]
                Geometric mean: |                   Speed up: 9.344% (+-0.223%) : [+]

Binaries:
jerry_a09c473_rel_refactor_obj_prop: 183,748
jerry_41eb163_rel_pointer_encoding: 173,732
```

Overall: 10% speedup and 10K binary reduction.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -58,7 +58,8 @@ typedef enum
   ECMA_TYPE_SIMPLE, /**< simple value */
   ECMA_TYPE_NUMBER, /**< 64-bit integer */
   ECMA_TYPE_STRING, /**< pointer to description of a string */
-  ECMA_TYPE_OBJECT /**< pointer to description of an object */
+  ECMA_TYPE_OBJECT, /**< pointer to description of an object */
+  ECMA_TYPE___MAX = ECMA_TYPE_OBJECT /** highest value for ecma types */
```
LaszloLango: AFAIK, the common suffix is `__COUNT` instead of `__MAX`.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -33,60 +30,87 @@ JERRY_STATIC_ASSERT (sizeof (ecma_value_t) * JERRY_BITSINBYTE >= ECMA_VALUE_SIZE
...
+    return (void *) (uintptr_t) ((value) & ~ECMA_VALUE_FULL_MASK);
+  }
+  else
+  {
+    return ECMA_GET_NON_NULL_POINTER (ecma_number_t,
+                                      value >> ECMA_VALUE_SHIFT);
+  }
+} /* ecma_get_pointer_from_ecma_value */
+
+#define ECMA_SIMPLE_VALUE_TO_VALUE(simple_value) \
```
LaszloLango: Missing comment

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -33,60 +30,87 @@ JERRY_STATIC_ASSERT (sizeof (ecma_value_t) * JERRY_BITSINBYTE >= ECMA_VALUE_SIZE
...
- * @return ecma value with updated field
+ * @return true - if the value is equal to the given simple value,
+ *         false - otherwise.
  */
-static ecma_value_t __attr_pure___
-ecma_set_value_value_field (ecma_value_t value, /**< ecma value to set field in */
-                            uintptr_t value_field) /**< new field value */
+static inline bool __attr_pure___ __attr_always_inline___
+ecma_is_value_equal_to_simple_value (ecma_value_t value, /**< ecma value */
+                                     ecma_simple_value_t simple_value) /**< simple value */
```
LaszloLango: The name is misleading. This is type of simple_value, so I suggest "value_type" or "simple_value_type"

jerry-core/ecma/base/ecma-globals.h
zherczeg: I introduced this form some time ago and use it nowadays. It is useful for switches.

LaszloLango: There is 36 matches for `__COUNT` and 6 matches for `__MAX` in the current source base. I'd prefer to use only one of them, because they mean the same. But this could be in a followup patch. Leave as is, the code looks good to me. :)

LaszloLango:
LGTM, if @akiss77 has no objection, you can land.

jerry-core/ecma/base/ecma-globals.h
akiss77: Just a clarification comment: `__MAX` is the highest possible value, `__COUNT` is the number of possible values. If an enum is numbered from 0, as usual, then `__COUNT == __MAX + 1`. However, if enum labels have "arbitrarily" set values, then either `__MAX` or `__COUNT` may make no sense to use.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -33,60 +30,90 @@ JERRY_STATIC_ASSERT (sizeof (ecma_value_t) * JERRY_BITSINBYTE >= ECMA_VALUE_SIZE
...
-static uintptr_t __attr_pure___
-ecma_get_value_value_field (ecma_value_t value) /**< ecma value */
+static inline ecma_value_t __attr_pure___ __attr_always_inline___
+ecma_pointer_to_ecma_value (const void *ptr) /**< pointer */
 {
-  return (uintptr_t) JRT_EXTRACT_BIT_FIELD (ecma_value_t, value,
-                                            ECMA_VALUE_VALUE_POS,
-                                            ECMA_VALUE_VALUE_WIDTH);
-} /* ecma_get_value_value_field */
+  if (sizeof (void *) <= sizeof (ecma_value_t))
```
akiss77: This should rather be a compile-time decision. We've had similar before: #976 .
```
#if UINTPTR_MAX <= UINT32_MAX
```
That would be the equivalent, if I'm right. (A comment might be added pointing out that `ecma_value_t` is a typedef to `uint32_t`.)

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -33,60 +30,90 @@ JERRY_STATIC_ASSERT (sizeof (ecma_value_t) * JERRY_BITSINBYTE >= ECMA_VALUE_SIZE
...
+    uintptr_t uint_ptr = (uintptr_t) ptr;
+    JERRY_ASSERT ((uint_ptr & ECMA_VALUE_FULL_MASK) == 0);
+    return (ecma_value_t) uint_ptr;
+  }
+  else
+  {
+    mem_cpointer_t ptr_cp;
+    ECMA_SET_NON_NULL_POINTER (ptr_cp, ptr);
+
+    return ((ecma_value_t) ptr_cp) << ECMA_VALUE_SHIFT;
```
akiss77: We do use the true branch if a regular pointer fits into an `ecma_value_t`. But do we check whether a compressed pointer fits into the 29 bits we have for value (is there a static assert for that)?

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -33,60 +30,90 @@ JERRY_STATIC_ASSERT (sizeof (ecma_value_t) * JERRY_BITSINBYTE >= ECMA_VALUE_SIZE
...
-                           ecma_type_t type_field) /**< new field value */
+static inline void * __attr_pure___ __attr_always_inline___
+ecma_get_pointer_from_ecma_value (ecma_value_t value) /**< value */
 {
-  return JRT_SET_BIT_FIELD_VALUE (ecma_value_t, value,
-                                  type_field,
-                                  ECMA_VALUE_TYPE_POS,
-                                  ECMA_VALUE_TYPE_WIDTH);
-} /* ecma_set_value_type_field */
+  if (sizeof (void *) <= sizeof (ecma_value_t))
```
akiss77: This is also a runtime decision that could be a compile-time one.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -136,9 +160,8 @@ ecma_is_value_null (ecma_value_t value) /**< ecma value */
 inline bool __attr_pure___ __attr_always_inline___
 ecma_is_value_boolean (ecma_value_t value) /**< ecma value */
 {
-  return (ecma_get_value_type_field (value) == ECMA_TYPE_SIMPLE
-          && (ecma_get_value_value_field (value) == ECMA_SIMPLE_VALUE_TRUE
-              || ecma_get_value_value_field (value) == ECMA_SIMPLE_VALUE_FALSE));
+  return (ecma_is_value_equal_to_simple_value (value, ECMA_SIMPLE_VALUE_TRUE)
+          || ecma_is_value_equal_to_simple_value (value, ECMA_SIMPLE_VALUE_FALSE));
```
akiss77: If you'd stay with the now-proposed logic, this could simply be `ecma_is_value_true (value) || ecma_is_value_false (value)`. The original version made use of the fact the checking the "simple" property was done only once. If you'd like to do some similar optimization with the current approach, then it would be somethin like:
```c
const ecma_value_t v = value | ECMA_VALUE_ERROR_FLAG;
return  (v == (ECMA_SIMPLE_VALUE_TO_VALUE (ECMA_SIMPLE_VALUE_TRUE) | ECMA_VALUE_ERROR_FLAG))
  || (v == (ECMA_SIMPLE_VALUE_TO_VALUE (ECMA_SIMPLE_VALUE_FALSE) | ECMA_VALUE_ERROR_FLAG));
```
As this looks overly complicated and we expect the compiler to inline and optimize all these tiny functions away, I'd suggest the first approach.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: This IS a compile time decision. I don't understand this.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: There are some plans for a follow-up work here. But I try to simplify this a bit.

jerry-core/jerry.c
```diff
@@ -1766,7 +1766,7 @@ jerry_run (jerry_api_value_t *error_value_p) /**< [out] error value */
   ecma_value_t error_value;
   jerry_completion_code_t ret_code = vm_run_global (&error_value);
 
-  error_value &= ~(1u << ECMA_VALUE_ERROR_POS);
+  error_value &= ~ECMA_VALUE_ERROR_FLAG;
```
akiss77: Do we need to clear this flag at all? `ecma_free_value` does not seem to care about it at all.

jerry-core/jerry.c
```diff
@@ -2385,7 +2385,7 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
     if (ret_code == JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION)
     {
       JERRY_ASSERT (ecma_is_value_error (error_value));
-      error_value = error_value & ~(1u << ECMA_VALUE_ERROR_POS);
+      error_value = error_value & ~ECMA_VALUE_ERROR_FLAG;
```
akiss77: ditto

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: I'll add it.

jerry-core/jerry.c
zherczeg: Well, it is in the original code. I think we can remove this in a follow-up patch if unnecessary. It would be better for tracking issues.

zherczeg:
Patch updated.

jerry-core/ecma/base/ecma-helpers-value.c
akiss77: There is a slight difference between `if (0)` and `#if 0`. You wrote the former, I prefer the latter.

`if (0)` *can* be optiized away (and usually is, I admit) but still has to be valid code (e.g., it must reference existing structure fields, which may not be there because of preproc guards used elsewhere -- not applicable to this case), while `#if 0` certainly *is* dead code that the complier does not even see. That's why the latter is used normally.

Even in the current code base, we (almost) never decide on type sizes with C code, but (almost) always with preprocessor conditionals. I've worked my way through 250+ occurrences of `sizeof` in jerry-core, and the only exception I've found is `ecma-helpers-external-pointers.c`, which uses `if (sizeof (x) < sizeof (y))` style. Which it shouldn't either, IMO.

zherczeg: And we don't need special comments. Compilers are clever enough to evaluate constants inside ifs/loops. E.g. while(true) is optimized with unconditional branches, and not load true into a register, and branch if that register is true. Or do {} while(false) is eliminated.

jerry-core/ecma/base/ecma-helpers-value.c
akiss77: You can just go with `return ecma_is_value_true (value) || ecma_is_value_false (value);` as suggested as the first alternative.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: Here the code is valid in both cases. I worry that we might change the type of ecma_value_t and forget about this.

jerry-core/jerry.c
akiss77: OK with me. Just don't let it be forgotten.

jerry-core/ecma/base/ecma-helpers-value.c
akiss77: Well, then the ultimate proper approach would be (IMHO) to add `#define ECMA_VALUE_MAX UINT32_MAX` after the typedef of `ecma_value_t` (i.e., follow the libc style as it declares limits for its types) and then use `#if UINTPTR_MAX <= ECMA_VALUE_MAX` here.

If you are strongly against that, then the `if` should check the right `sizeof`s at least. The real question is whether an `uintptr_t` fits into an `ecma_value_t`. Yes, it takes a really strange architecture to make it differ from the size of `void *` but if we can do it properly and in a future proof way, why shouldn't we.

zherczeg: My problem with ECMA_VALUE_MAX is that it suggests that ecma value is a number. In reality it is a bit structure stored in a word sized value. I would like to store small integer values in ecma_value_t in the future, and would like to reserve MAX and MIN stuff for that purpose.

akiss77: Then go for the second alternative for now.

zherczeg: What about the following define

    #define ECMA_VALUE_CAN_HOLD_POINTERS

This define could be used for other purposes, like encoding internal pointer properties. They also don't need compresiion on 32 bit systems.

akiss77: Could you add a commit on top of the current one to show how it would like?

zherczeg: Added

akiss77:
Much nicer to my eyes.
LGTM FWIW

PR #1008, opened 2016-04-14
Merge darwin and linux jerry-libc targets into a single posix target

akiss77:
The two target implementations are very close clones of each other.
The only known differences are the following:

* The asm component of the linux target has .type and .size
  declarations for the functions defined therein, while the darwin
  target doesn't support those.

* Linux uses `__NR_xxx` mnemonics for syscall numbers, while darwin
  uses `SYS_xxx` format. A

* Darwin does not have `exit_group` syscall but `exit` only.

* The linux target has a commented out `jrt_set_mem_limits` fuction
  declaration at the end of the C source. (Based in its name, this
  function does not really belong here.)

Simple preprocessor macros can unify the first three differences.
While for the sake of legacy, we can keep the fourth commented-out
code in the code base; it might turn out to be useful elsewhere in
the future. Since it remains commented out it wont cause any
problems on any OSs.

So, this patch gets rid of a lot of duplication.


zherczeg:
LGTM. I trust that you tested it on mac.

akiss77:
@zherczeg, Me not but the CI did. I took care not to introduce regressions on that platform. (However, did not fix any either, so the darwin jobs are still red.)

jerry-libc/CMakeLists.txt
```diff
@@ -77,10 +77,10 @@ set(COMPILE_FLAGS_LIBC "${COMPILE_FLAGS_JERRY} ${C_FLAGS_JERRY}")
 
  # Platform-specific
   # Linux
-   file(GLOB SOURCE_LIBC_LINUX target/linux/*.c target/linux/*.S)
+   file(GLOB SOURCE_LIBC_LINUX target/posix/*.c target/posix/*.S)
 
   # Darwin
-   file(GLOB SOURCE_LIBC_DARWIN target/darwin/*.c target/darwin/*.S)
+   file(GLOB SOURCE_LIBC_DARWIN target/posix/*.c target/posix/*.S)
```
LaszloLango: Can't we use simply `SOURCE_LIBC_POSIX` insted of `SOURCE_LIBC_DARWIN` and `SOURCE_LIBC_DARWIN`?

jerry-libc/target/posix/jerry-libc-target.c
```diff
@@ -25,11 +25,24 @@
...
+#include <sys/syscall.h>
 #include <sys/types.h>
 #include <sys/stat.h>
 #include <sys/time.h>
 
+#if defined (__linux__)
+#define SYSCALL_NO(NAME) __NR_ ## NAME
+#endif /* defined(__linux__) */
+
+#if defined (__APPLE__) && defined (__MACH__)
```
LaszloLango: elseif

jerry-libc/target/posix/jerry-libc-target.c
```diff
@@ -25,11 +25,24 @@
...
+#define SYS_exit_group SYS_exit
+#define SYSCALL_NO(NAME) SYS_ ## NAME
+#endif /* defined(__APPLE__) && defined(__MACH__) */
+
+#ifndef SYSCALL_NO
```
LaszloLango: else

jerry-libc/CMakeLists.txt
akiss77: We have later an indirection:
```
add_library(${TARGET_NAME}.lib STATIC ${SOURCE_LIBC} ${SOURCE_LIBC_${PLATFORM_EXT}})
```

And we don`t have a posix platform but linux and darwin only.

LaszloLango: `add_library(${TARGET_NAME}.lib STATIC ${SOURCE_LIBC} ${SOURCE_LIBC_POSIX})` isn't enough?

akiss77: No, because we have non-posix targets as well, MCU_STM32F3 and MCU_STM32F4. They have their corresponding `SOURCE_LIBC_MCU_STM32F3` and `SOURCE_LIBC_MCU_STM32F4` variables and use the implementation in `target/mcu-stubs`. (See the next 5-6 lines.)

akiss77:
Updated

jerry-libc/CMakeLists.txt
LaszloLango: OK, thanks for checking.

LaszloLango:
LGTM

PR #1009, opened 2016-04-15
Fix layering violation in applications

akiss77:
Apps that use jerry should not (must not!) include and rely on its
internal headers. Typical issue is the use of "jrt/jrt.h". Fixing
`main-{unix,mcu}.c`, and speculatively the apps under the `targets`
directory as well.

### DIFF
--- a/main-mcu.c
+++ b/main-mcu.c
@@ -22,7 +22,6 @@
 #define JERRY_STANDALONE_EXIT_CODE_FAIL (1)
 
 #include JERRY_MCU_SCRIPT_HEADER
-#include "jerry-core/jerry-api.h"
 
 static const char generated_source[] = JERRY_MCU_SCRIPT;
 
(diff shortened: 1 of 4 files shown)


LaszloLango:
This makes JERRY_ERROR_MSG unused, but the definition is still there. I agree that apps should not include `jrt/jrt.h`. I suggest to move the definition of `JERRY_ERROR_MSG` into `jerry.h`.

akiss77:
@LaszloLango I'd be against that. This is an internal macro, I suggest not to expose it. However, alternatively, what do you think about using the function(s) exposed by the jerry-port.h header? Thus, not only jerry could use those functions but the embedder(s) as well. This time, it would be `jerry_port_errormsg()`.

LaszloLango:
@akiss77, sounds better. :)

akiss77:
Updated

LaszloLango:
@akiss77, thanks for the update. LGTM

zherczeg:
LGTM

PR #1010, opened 2016-04-15
Set the correct pointer value in 'ecma_copy_ecma_string' function.

robertsipka:

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -422,7 +422,7 @@ ecma_copy_ecma_string (ecma_string_t *string_desc_p) /**< string descriptor */
       ecma_string_heap_header_t *new_data_p = (ecma_string_heap_header_t *) mem_heap_alloc_block (data_size);
       memcpy (new_data_p, data_p, data_p->size + sizeof (ecma_string_heap_header_t));
 
-      ECMA_SET_NON_NULL_POINTER (new_str_p->u.collection_cp, data_p);
+      ECMA_SET_NON_NULL_POINTER (new_str_p->u.collection_cp, new_data_p);
 
       break;
     }


zherczeg:
Please add some regression test for this.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -422,7 +422,7 @@ ecma_copy_ecma_string (ecma_string_t *string_desc_p) /**< string descriptor */
       ecma_string_heap_header_t *new_data_p = (ecma_string_heap_header_t *) mem_heap_alloc_block (data_size);
       memcpy (new_data_p, data_p, data_p->size + sizeof (ecma_string_heap_header_t));
 
-      ECMA_SET_NON_NULL_POINTER (new_str_p->u.collection_cp, data_p);
+      ECMA_SET_NON_NULL_POINTER (new_str_p->u.collection_cp, new_data_p);
```
akiss77: Quick question: is this the only issue in this code block? I'm not familiar with this function but we have `*new_str_p = *string_desc_p;` above. That copies over `refs_and_container`, `hash`, and `u.collection_cp`. The latter will be overwritten (fixed in this patch), `hash` does not change. What about the reference count? Is that OK not to (re)set?

zherczeg: Just check the comment at the beginning:
@return pointer to copy of ecma-string with reference counter set to 1

zherczeg: So probably copying the refcounter is also a bug. I had a feelewng that this is an untested code path. That is why we need a reftest.

akiss77: I see the comment. But where is it set to 1 in this code path?

zherczeg: I agree this is probably another bug. Good catch.

akiss77: I'm not sure that it is possible to construct a regression test case that can catch the refcount issue... It seems to me that this will be "just" eating up memory but not cause any semantic error. I'd be glad to be proven wrong though.

(The pointer issue is different. That might be possible to catch.)

zherczeg: Here the problem is that there is a reference to the old string. A regression test would look like this:

    var a = "string";
    var b = [];
    for (i =0 i < 10000; i++) b[i] = a;
    gc();
    for (i =0 i < 9000; i++) b[i] = null;
    gc(); // could be a crash

zherczeg:
Is this issue still valid?

PR #1012, opened 2016-04-15
Update existing and add missing copyright & license notices

akiss77:
* LICENSE needed year update.

* The asm component of the posix target of jerry-libc had no
  copyright & license notice since it has been introduced in 2015.
  Traced back history and added missing header with correct years.

* Three tests in jerry/fail/1 also missed header. Added.


LaszloLango:
LGTM

zherczeg:
LGTM

Issue #964, opened 2016-03-12
Re-thinking the Jerry Port API (proposal)


akosthekiss:
# Motivation

Lately, there have been several PRs (#945 #946 #956 #957) that pointed out that the port API of jerry is not that easy to use by embedders that are not in the main code base and thus could benefit from improvements.

Taking a closer look at some parts of the code also revealed that even the use cases in the main code base could become cleaner after some rethinking-refactoring.

# Analysis

I've built jerry with `make LTO=OFF ALL_IN_ONE=ON` and ran `nm` on the resulting jerry-core library to see what symbols it defines and -- most importantly -- what external symbols it references. Below is the list of external symbols (categorization by me):

```
"termination"
                 U abort
                 U exit
"i/o"
                 U printf
                 U putchar
                 U stderr
                 U vfprintf
"math"
                 U acos
                 U asin
                 U atan
                 U atan2
                 U ceil
                 U cos
                 U exp
                 U fabs
                 U floor
                 U fmod
                 U log
                 U pow
                 U sin
                 U sqrt
                 U tan
"general libc"
                 U longjmp
                 U memcmp
                 U memcpy
                 U memmove
                 U memset
                 U rand
                 U setjmp
                 U strlen
                 U strncmp
```

IMO, the "math" and "general libc" categories cover very fundamental functionalities, which a port shouldn't be allowed to override. However, the "termination" and "i/o" categories can be port specific, and we should think about what we actually use them for and how we should let the ports give their specific implementation.

# Proposal

I propose not to replicate the low-level libc API in the port API but to make the port functions "semantic".

## Port API: Termination

```c
/**
 * Signal the port that jerry experienced a fatal failure from which it cannot
 * recover.
 *
 * @param code gives the cause of the error.
 *
 * Note: jerry expects the function not to return.
 *
 * Example: a libc-based port may implement this with exit() or abort().
 */
void jerry_port_fatal (jerry_fatal_code_t code);
```

It should be the embedder application and/or the port that decides whether exit or abort is to be called. This differs from the current approach, where jerry even provides an api function to switch between exit/abort mode of fatal termination. This would make that function unnecessary.

_Note:_ It is highly questionable whether a library should be able to terminate an application. However, as of now, we only have the concept of completion code around jerry_parse and jerry_run. Most of the other API functions have no way of signaling an error. So, let's keep the termination approach for now with this port function.

## Port API: I/O

```c
/**
 * Print a string to the console. The function should implement a printf-like
[...]

PR #1015, opened 2016-04-18
Introduce the Termination Port API

akiss77:
* Moved the error codes to jerry-port.h and declared port function
  `jerry_port_fatal`.

* Moved "exit or abort on fail" functionality to the newly added
  jerry-port-default-fatal.c.

* This implied that a default port-specific API had to be introduced:
  functions `jerry_port_default_set_abort_on_fail` and
  `jerry_port_default_is_abort_on_fail` declared in jerry-port-default.h
  control the fatal exit behaviour.

* For the sake of clarity, renamed jerry-port.c to
  jerry-port-default-io.c.

* Adapted CMakeLists to deal with port implementations consisting of
  more then one source file and exposing headers. This also required
  the renaming of `EXTERNAL_PORT_FILE` cmake option to
  `EXTERNAL_PORT_DIR`.

* Adapted main sources to use the default port header for the
  abort-on-fail functionality, as that is not part of the core jerry
  API anymore.

* Added default port implementation to the static source code checker
  tools.

### DIFF
--- a/jerry-core/jerry-port.h
+++ b/jerry-core/jerry-port.h
@@ -35,6 +35,42 @@ int jerry_port_logmsg (FILE *stream, const char *format, ...);
 int jerry_port_errormsg (const char *format, ...);
 int jerry_port_putchar (int c);
 
+/*
+ * Termination Port API
+ *
+ * Note:
+ *      It is questionable whether a library should be able to terminate an
+ *      application. However, as of now, we only have the concept of completion
+ *      code around jerry_parse and jerry_run. Most of the other API functions
+ *      have no way of signaling an error. So, we keep the termination approach
+ *      with this port function.
+ */
+
+/**
+ * Error codes
+ */
+typedef enum
+{
+  ERR_OUT_OF_MEMORY = 10,
+  ERR_SYSCALL = 11,
+  ERR_REF_COUNT_LIMIT = 12,
+  ERR_UNIMPLEMENTED_CASE = 118,
+  ERR_FAILED_INTERNAL_ASSERTION = 120
+} jerry_fatal_code_t;
+
+/**
+ * Signal the port that jerry experienced a fatal failure from which it cannot
+ * recover.
+ *
+ * @param code gives the cause of the error.
+ *
+ * Note:
+ *      Jerry expects the function not to return.
+ *
+ * Example: a libc-based port may implement this with exit() or abort(), or both.
+ */
+void jerry_port_fatal (jerry_fatal_code_t code);
+
 /**
  * @}
  */
--- a/targets/default/jerry-port-default.h
+++ b/targets/default/jerry-port-default.h
@@ -1,0 +1,42 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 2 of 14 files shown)


akiss77:
Related to #964 . Not a full implementation of the ideas suggested there, of course, but a first step towards it. In addition to introducing the Termination API (which seemed to be the one that everyone agreed on), the PR already exemplifies how a port implementation can have its own extension(s) to the core Jerry API.

akiss77:
Note: also related to #957

jerry-core/jerry-port.h
```diff
@@ -35,6 +35,40 @@ int jerry_port_logmsg (FILE *stream, const char *format, ...);
 int jerry_port_errormsg (const char *format, ...);
 int jerry_port_putchar (int c);
 
+/*
+ * Termination Port API
+ *
+ * Note: It is questionable whether a library should be able to terminate an
```
LaszloLango: Minor style issue. The common style for note comments are that using newline and indent to colon.

```C++
/*
 * Note:
 *      It is questionable whether...
```

jerry-core/jerry-port.h
```diff
@@ -35,6 +35,40 @@ int jerry_port_logmsg (FILE *stream, const char *format, ...);
...
+  ERR_FAILED_INTERNAL_ASSERTION = 120
+} jerry_fatal_code_t;
+
+/**
+ * Signal the port that jerry experienced a fatal failure from which it cannot
+ * recover.
+ *
+ * @param code gives the cause of the error.
+ *
+ * Note: jerry expects the function not to return.
```
LaszloLango: ditto

targets/default/jerry-port-default.h
```diff
@@ -0,0 +1,42 @@
...
+
+#ifndef JERRY_PORT_DEFAULT_H
+#define JERRY_PORT_DEFAULT_H
+
+#include <stdbool.h>
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif /* !__cplusplus */
```
LaszloLango: `#endif /* __cplusplus */`

targets/default/jerry-port-default.h
```diff
@@ -0,0 +1,42 @@
...
+void jerry_port_default_set_abort_on_fail (bool);
+bool jerry_port_default_is_abort_on_fail (void);
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif /* !__cplusplus */
```
LaszloLango: `#endif /* __cplusplus */`

akiss77: Although that might seem better, we have all our public headers with comments `/* !__cplusplus */`. (see jerry-libc/include/*.h, or jerry-core/jerry-port,h)

LaszloLango: That is a problem. The style should be the following everywhere in Jerry:

```C++
#ifdef NAME
#endif /* NAME */

#ifndef NAME
#endif /* !NAME */

#ifdef NAME
#else /* !NAME */
#endif /* NAME */

#ifndef NAME
#else /* NAME */
#endif /* !NAME */
```

I think we need a Vera++ rule for this sooner or later.

LaszloLango:
I like the PR, but I'd like to hear the opinion of everyone who is involved in #964 before we merge. Let's give some time for everyone to check it. :)

targets/default/jerry-port-default.h
akiss77: I sort of agree about the style-to-be-followed. However, the vera++ rule can be tricky as there are `#if`s where the condition is more complex than just checking whether a macro is defined.

And what's unclear to me what to put after the final `#endif` when there are `#elif`s after the first `#if`?

LaszloLango: I think the comment of `#endif` always should be related to first `#if`

akiss77:
Should anyone be interested, a draft of the Date Port API is prototyped as some patches on top of this PR here: https://github.com/akiss77/jerryscript/tree/port-date (no PR, yet)

Acknowledgement: it does have some overlap with #969 by @franc0is

zherczeg:
LGTM

LaszloLango:
@akiss77, Date Port API patch also looks good for first sight.

LaszloLango:
LGTM

franc0is:
I'm just now catching up with these PRs, they are terrific! I'm thrilled to see the port model moving forward.

PR #1016, opened 2016-04-18
Unifiy the comments of preprocessor conditionals

akiss77:


akiss77:
As discussed at #1015 . This is not necessarily the final version, but consistent at least.

Why I'm saying it's not necessarily the final version: the comments after `#endif`s look sometimes weird if preceeded by an `#else` or `#elif` preprocessor directive. One could argue that the final comment of an `#endif` could repeat that of its directly preceeding directive, i.e., not necessarily the predicate of the `#if`. E.g., 

```c
#ifdef XXX
#endif /* XXX */

#ifndef XXX
#endif /* !XXX */

#if XXX
#elif YYY
#elif ZZZ
#else /* !XXX && !YYY & !ZZZ */
#endif /* !XXX && !YYY & !ZZZ */
```

But that's really just a matter of style. Right now, the PR goes with the originally suggested style. Happy to change it however the reviewers vote. Just let's get it consitent.

zherczeg:
This is ok for me. LGTM

LaszloLango:
Thanks for cleaning up this. LGTM

PR #1018, opened 2016-04-19
Introduce the Date Port API

akiss77:
Replaced `gettimeofday`-related code with `jerry_port_get_current_time` and `jerry_port_get_time_zone` function calls. Moved old code to default port implementation.

Removed `ENABLE_DATE_SYS_CALLS` as date syscalls should "just work".

Fixed DST adjustments: even if `gettimeofday` returns meaningful data in `tz_dsttime`, the value is just a flag (or, according to some sources, a tri-state value: >0 - DST applies, ==0 - DST does not apply, <0 - unknown). Hitherto, the field was simply added to/subtracted from a time value in milliseconds. To use it approximately correctly, the field's value should be multiplied by "millisecs/hour".


akiss77:
Related to the proposal #964 .
Acknowledgement to @franc0is for #969 with which this PR overlaps.

LaszloLango:
LGTM

jerry-core/jerry-port.h
```diff
@@ -71,6 +72,34 @@ typedef enum
...
+ *         false - otherwise
+ */
+bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p);
+
+/**
+ * Get system time
+ *
+ * @return milliseconds since Unix epoch
+ */
+unsigned long int jerry_port_get_current_time (void);
```
zherczeg: milliseconds? not seconds?

LaszloLango: We have to use milliseconds everywhere, so it is the best, so we don't have to use multiplications and divisions. I think ms is good.

zherczeg: The problem is that an uint32 cannot represent ms since Unix epoch. An uint32 can only represent around 47 days in ms.

Is this tested in a 32 bit machine?

LaszloLango: I haven't tested. What about using `uint64_t` for return type?

LaszloLango: Note: I checked and you are right. `unsigned long` isn't enough for 32bit systems.

zherczeg: uint64_t is ok for me. or double

akiss77:
Updated

zherczeg:
LGTM. Please squash commits

franc0is:
Sorry I am just now looking at this. I have a single comment: using a `uint64_t` when we know it will be converted to floating point is a little bit wasteful. Why did you decide against using `double`?

akiss77:
Even if `jerry_port_get_current_time` returns `double`, a "cast" with `DOUBLE_TO_ECMA_NUMBER_T` is still needed, since the number representation may be `float` instead of `double`. (Of course, if numbers are `double`, that macro doesn't cost anything.)

That being said, I tend to agree now that `double` would be/have been better than `uint64_t`. Fortunately, this would take a very small change only.

PR #1019, opened 2016-04-19
Merged target for mbed boards.

knightburton:
Merged the mbed releated targets into one source.
Available targets: FRDM-K64f, Discovery-STM32F4, Discovery-STM32F429ZI, NUCLEO-STM32F4

JerryScript-DCO-1.0-Signed-off-by: Imre Kiss kissi@inf.u-szeged.hu


targets/target-mbed/Makefile.mbed
```diff
@@ -0,0 +1,100 @@
...
+	YOTTA_TARGET  = stm32f429i-disco-gcc
+else ifeq ($(TARGET), nucleo)
+	YOTTA_TARGET  = st-nucleo-f401re-gcc
+	TARGET_DIR   ?= /media/$(USER)/NODE_F401RE
+endif
+
+UPPERC_TARGET = $(shell echo $(TARGET) | tr a-z A-Z) 
+INTERM        = build/obj-mbed-$(TARGET)
+OUTPUT        = build/bin/$(TYPE).mbed$(TARGET)
+COPYTARGET    = targets/target-mbed/libjerry
```
akiss77: I have the problem with the directory name that it unnecessarily duplicates the word "target". We don't follow that scheme under "targets/". Please, just put everything under "targets/mbed/".

targets/target-mbed/Makefile.mbed
```diff
@@ -0,0 +1,100 @@
...
+	 -DEXTERNAL_CMAKE_C_COMPILER=arm-none-eabi-gcc \
+	 -DEXTERNAL_CMAKE_C_COMPILER_ID=GNU \
+	 -DEXTERNAL_COMPILE_FLAGS="$(EXT_CFLAGS)" \
+	 -DEXTERNAL_MEM_HEAP_SIZE_KB=$(JERRYHEAP) \
+	 -DEXTERNAL_PORT_DIR=$(EXT_PORT_DIR)
+
+	make -C $(INTERM) $(TYPE).external
+	cp `cat $(INTERM)/$(TYPE).external/list` $(OUTPUT)/.
+	cp $(OUTPUT)/lib$(TYPE).jerry-core.a $(COPYTARGET)/libjerrycore.a
+	cp $(OUTPUT)/lib$(TYPE).jerry-libm.lib.a $(COPYTARGET)/libfdlibm.a
```
akiss77: As we have specialized fdlibm to jerry recently, its name became jerry-libm. So, I'd suggest that being reflected in the copy target as well, i.e., please copy it as `$(COPYTARGET)/libjerrylibm.a`. (This will need an update in makejerry.cmake as well.)

targets/target-mbed/js/blink.js
```diff
@@ -1,9 +1,10 @@
 var check = 1;
 
-function blink() {
+function blink ()
+{
   var blk = (check > 8) ? 1 : 0;
-  led(1, blk);
+  led(0, blk);
   check = (check >= 10) ? 1 : check+1;
```
akiss77: Style: add spaces around the plus sign.

targets/target-mbed/source/jerry_port.c
```diff
@@ -0,0 +1,141 @@
...
+  va_start (args, format);
+  count = vfprintf (stderr, format, args);
+  va_end (args);
+  return count;
+} /* jerry_port_errormsg */
+
+/**
+ * Provide output character to console implementation for the engine.
+ */
+int jerry_port_putchar (int c) /**< character to put */
```
akiss77: `jerry_port_putchar` has been dropped from the port API, so this implementation is not needed.

targets/target-mbed/source/jerry_port.c
```diff
@@ -0,0 +1,141 @@
...
+{
+  return putchar ((unsigned char) c);
+} /* jerry_port_putchar */
+
+/**
+ * Sets whether 'abort' should be called instead of 'exit' upon exiting with
+ * non-zero exit code in the default implementation of jerry_port_fatal.
+ */
+void jerry_port_default_set_abort_on_fail (bool flag) /**< new value of 'abort on fail' flag */
```
akiss77: Copying over the `*_abort_on_fail` functions from the default port implementation seems to be a misunderstanding. From main.cpp and jerry_run.cpp, it does not seem as if you needed an "do an exit or an abort on fail?" decision. The default port extends the main jerry API with these functions and logic so that the jerry command line interpreter on Linux or OSX can control fatal exit behaviour via command line options. On the boards, however, you can predetermine that. So, I'd suggest drop these two functions and implement `jerry_port_fatal` in a way that simply either calls `abort` or `exit`.

(BTW, the `jerry_port_default_` prefix of these functions hints that they are specific to the default port.)

targets/target-mbed/source/jerry_port.c
```diff
@@ -0,0 +1,141 @@
...
+{
+  struct timeval tv;
+  struct timezone tz;
+
+  /* gettimeofday may not fill tz, so zero-initializing */
+  
+  tz.tz_minuteswest = 0;
+  tz.tz_dsttime = 0;
+
+  gettimeofday (&tv, &tz);
```
akiss77: I'm not familiar with mbed's libc implementation, but from `_gettimeofday` (just some lines below) I guess that `gettimeofday` is just a trampoline to `_gettimeofday`. And that function does not do anything with the time zone parameter. So, I guess you'll have no info about TZ at all. If my hypothesis is right, then this function could be just:
```c
tz_p->offset = 0;
tz_p->daylight_saving_time = 0;
return true;
```
Without any calls to `gettimeofday`.

targets/target-mbed/source/jerry_port.c
```diff
@@ -0,0 +1,141 @@
...
+/**
+ * Implementation of jerry_port_get_current_time.
+ */
+double jerry_port_get_current_time ()
+{
+  struct timeval tv;
+
+  gettimeofday (&tv, NULL);
+
+  return ((double) tv.tv_sec) * 1000.0 + ((double) tv.tv_usec) / 1000.0;
```
akiss77: Similarly here: you might not need `gettimeofday` but you can just move the logic from `_gettimeofday` here. Something like:
```c
return (double) us_ticker_read ();
```
(If I get milli-micro-nano conversions right...)

targets/target-mbed/Makefile.mbed
```diff
@@ -0,0 +1,100 @@
...
+
+else
+  $(error This board ($(board)) is not supported!)
+endif
+
+EXT_CFLAGS := -D__TARGET_MBED_$(UPPERC_TARGET)
+EXT_CFLAGS += -mlittle-endian -mthumb -mcpu=cortex-m4
+EXT_CFLAGS += -Wno-error=format=
+
+EXT_PORT_DIR := source/
```
akiss77: I almost wrote that this isn't going to work since the port directory is the same as the directory of the embedder app sources, and they are going to be built and linked twice (once into libjerry-core and then into the app). Then I realized that embedder app sources are cpp and jerry-core looks for "*.c" only for port implementation, and that saves the day. However, I still consider this (putting port and app code into the same dir) a practice to be avoided.

Please, have a separate directory for the port code and use that for `EXT_PORT_DIR`.

thegecko:
Thanks @knightburton I can confirm this builds.

targets/target-mbed/Makefile.mbed
knightburton: ok, you are right! My bad!

targets/target-mbed/source/jerry_port.c
knightburton: You are right again, i did not digged into myself.

targets/target-mbed/Makefile.mbed
knightburton: @akiss77 The yotta does not like any stuff outside the source folder. If we put the port code into the source/port folder?

akiss77: I have to admit that I'm not familiar with yotta. You already have a directory that's not under "source/", i.e., "js/". Does that cause no problem?

The port implementation will not/should not be compiled by yotta after all but it is compiled by the build system of jerry-core and the resulting object gets incorporated in that library (in `$(COPYTARGET)/libjerrycore.a`). That's why I though it could be separated from "source/".

Do you get some errors if "port/" is not under "source/"?

knightburton: The "js/" directory does not couse problem. The js2c python script int the "../tools/" directory creates a C file from the "js/" directory content. We do not use this dirctory in the jerry or yotta building process.

Yes, I get some undefined reference errors if  the "port/" is not under the "source/" directory.
I get this: https://gist.github.com/knightburton/9237699e7a4f427acc94200b171d117a
but if the "port/" is under the "source/" it is works fine.

targets/mbed/Makefile.mbed
```diff
@@ -0,0 +1,101 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# use TAB-8
+
+CURDIR          = `pwd`
```
akiss77: I'm just realizing that this assignment is here.
- Does make not set it by default? (From make manual: "when GNU make starts (after it has processed any -C options) it sets the variable CURDIR to the pathname of the current working directory")
- It is not referenced anywhere.

Just remove it?

targets/mbed/Makefile.mbed
```diff
@@ -0,0 +1,101 @@
...
+
+else
+  $(error This board ($(board)) is not supported!)
+endif
+
+EXT_CFLAGS := -D__TARGET_MBED_$(UPPERC_TARGET)
+EXT_CFLAGS += -mlittle-endian -mthumb -mcpu=cortex-m4
+EXT_CFLAGS += -Wno-error=format=
+
+EXT_PORT_DIR := "nasty-hack-nonexistent-directory-to-prevent-jerry-core-from-compiling-any-port-implementation"
```
akiss77: Yes, that's me to blame. I've suggested this in an offline discussion. :)
Then, I've had a probably nicer idea to remove the need for the hack. I see that right now this patch touches the targets directory only. But if you'd add just two lines to jerry-core/CMakeLists.txt to enhance the port_dir processing a bit, then things could become cleaner.:
```
 if(NOT ("${PORT_DIR}" STREQUAL ""))
  file(GLOB SOURCE_PORT_FILES ${PORT_DIR}/*.c)
 endif()
```
Then, `EXT_PORT_DIR:=""` should work.
(I haven't tried it out myself, so this is just a best guess. Let me know whether it works.)

targets/mbed/Makefile.mbed
knightburton: Yeah, i forgot this, thanks to remind me!

targets/mbed/Makefile.mbed
knightburton: It's perfectly works for me.

akiss77:
LGTM FWIW

targets/mbed/readme.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+ifeq ($(TARGET), k64f)
+  YOTTA_TARGET  = frdm-k64f-gcc
+  TARGET_DIR   ?= /media/$(USER)/MBED
+else ifeq ($(TARGET), stm32f4)
+  YOTTA_TARGET  = 
+```
+
+Basicly, you can create a new target in this way (If the mbed OS support your board).
```
LaszloLango: typo: `Basically`

targets/mbed/readme.md
```diff
@@ -0,0 +1,68 @@
...
+
+2. The next rule is the `js2c`. This rule calls a `js2c.py` python script from the `jerryscript/targets/tools` and creates the JavaScript builtin file into the `targets/mbed$(TARGET)/source/` folder. This file is the `jerry_targetjs.h`. You can run this rule with the follwoing command:
+
+  - `make -f targets/target-mbed/Makefile.mbed board=$(TARGET) js2c`
+
+3. The last rule is the `yotta`. This rule sets the yotta target and install the mbed-drivers module, install the dependencies for the mbed OS and finaly creates the mbed binary file. The binary file will be genrated at `targets/mbed$(TARGET)/build/$(YOTTA_TARGET)/source/jerry.bin`. You can run this rule with the following command: 
+
+  - `make -f targets/target-mbed/Makefile.mbed board=$(TARGET) yotta`
+
+4. Optional rule: `clean`. This rule is removes the build folder from the target-mbed and jerry. You can run this rule with this command:
```
LaszloLango: "This rule ~~is~~ removes ..." or simply "It removes ..."

targets/mbed/source/jerry_run.cpp
```diff
@@ -13,38 +14,54 @@
...
+    jerry_api_release_object (err_obj_p);
   }
   else
   {
     if ((flags & JERRY_FLAG_PARSE_ONLY) == 0)
     {
-      ret_code = jerry_run ();
+      ret_code = jerry_run (&err_value);
+      jerry_api_string_t *err_str_p = NULL;
+      if (err_obj_p != NULL)
```
LaszloLango: This condition never has to be true, because if `jerry_parse` does not fail, then `error_obj_p` is not set. I'd replace this with an assert:
`JERRY_ASSERT (error_obj_p != NULL);`

targets/mbed/source/jerry_run.cpp
```diff
@@ -13,38 +14,54 @@
...
+        jerry_api_value_t err_value = jerry_api_create_object_value (err_obj_p);
+        err_str_p = jerry_api_value_to_string (&err_value);
+        jerry_api_release_object (err_obj_p);
+      }
+      else if (!jerry_api_value_is_void (&err_value))
+      {
+        err_str_p = jerry_api_value_to_string (&err_value);
+        jerry_api_release_value (&err_value);
+      }
+      if (err_str_p != NULL)
```
LaszloLango: Please add an empty line before the new if statement.

targets/mbed/source/jerry_run.cpp
knightburton: ok, thank you!

targets/mbed/readme.md
knightburton: hmm, shame, thanks.

targets/mbed/source/makejerry.cmake
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015-2016 Samsung Electronics Co., Ltd.
+# Copyright 2015 Samsung Electronics Co., Ltd.
```
LaszloLango: Why? Downgrading the copyright date is odd to me. This file is touched by @akiss77 in 3ba286f3

targets/mbed/source/native_mbed.h
```diff
@@ -13,11 +13,9 @@
...
-#define __NATIVE_MBEDK64F_H__
-
-
-void native_led(int port, int val);
+#ifndef __NATIVE_MBED_H__
+#define __NATIVE_MBED_H__
 
+void native_led (int port, int val);
 
 #endif
```
LaszloLango: Please add a comment to `#endif`
`#endif /* !__NATIVE_MBED_H__ */`

targets/mbed/source/port/jerry_port.c
```diff
@@ -0,0 +1,78 @@
...
+#include <sys/time.h>
+
+#include "jerry-core/jerry-port.h"
+
+#include "mbed-hal/us_ticker_api.h"
+
+/**
+ * Provide log message to filestream implementation for the engine.
+ */
+int jerry_port_logmsg (FILE *stream, /**< stream pointer */
```
LaszloLango: Move the return value to a separate line.

targets/mbed/source/port/jerry_port.c
```diff
@@ -0,0 +1,78 @@
...
+  va_start (args, format);
+  count = vfprintf (stream, format, args);
+  va_end (args);
+  return count;
+} /* jerry_port_logmsg */
+
+/**
+ * Provide error message to console implementation for the engine.
+ */
+int jerry_port_errormsg (const char *format, /**< format string */
```
LaszloLango: ditto

targets/mbed/source/port/jerry_port.c
```diff
@@ -0,0 +1,78 @@
...
+  va_start (args, format);
+  count = vfprintf (stderr, format, args);
+  va_end (args);
+  return count;
+} /* jerry_port_errormsg */
+
+/**
+ * Implementation of jerry_port_fatal.
+ */
+void jerry_port_fatal (jerry_fatal_code_t code)
```
LaszloLango: ditto

targets/mbed/source/port/jerry_port.c
```diff
@@ -0,0 +1,78 @@
...
+{
+  exit (code);
+} /* jerry_port_fatal */
+
+/**
+ * Implementation of jerry_port_get_time_zone.
+ */
+bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p)
```
LaszloLango: ditto

targets/mbed/source/port/jerry_port.c
```diff
@@ -0,0 +1,78 @@
...
+{
+  tz_p->offset = 0;
+  tz_p->daylight_saving_time = 0;
+  return true;
+} /* jerry_port_get_time_zone */
+
+/**
+ * Implementation of jerry_port_get_current_time.
+ */
+double jerry_port_get_current_time ()
```
LaszloLango: ditto

targets/mbed/source/port/jerry_port.c
LaszloLango: Missing comment for the argument

targets/mbed/source/port/jerry_port.c
LaszloLango: Missing comment for argument and for return value.

targets/mbed/source/port/jerry_port.c
LaszloLango: Missing comment for return value

targets/mbed/source/jerry_run.cpp
knightburton: I can't use the 'JERRY_ASSERT', it is not declared in this scope

LaszloLango:
Nice job, great PR!  I had only a few comment. Please fix them when you can.

targets/mbed/Makefile.mbed
```diff
@@ -46,38 +65,36 @@ jerry:
...
+	cd targets/mbed; \
+	 yotta target $(YOTTA_TARGET); \
 	 yotta build
 
-
-flash: check_mbed
-	cp targets/mbedk64f/build/frdm-k64f-gcc/source/jerry.bin \
-	 "$(TARGET_DIR)/."
+flash:
+	ifndef TARGET_DIR
```
stephenkyle-ARM: ifndef and related lines shouldn't be tab-indented, as they're not shell commands, but part of the Makefile.

targets/mbed/source/jerry_run.cpp
knightburton: @LaszloLango  I can't use the 'JERRY_ASSERT', it is not declared in this scope
what should i do? :) I am not sure.

LaszloLango: ops, my bad. Simply use `assert` instead or just remove the `if` statement. `JERRY_ASSERT` is not available from targets folder.

knightburton:
Updated.

LaszloLango:
LGTM

PR #1023, opened 2016-04-21
Change current time from uint64_t to double in Date Port API

akiss77:
As discussed after #1018


LaszloLango:
I'm OK with this change. LGTM

zherczeg:
LGTM. I also preferred double.

PR #1025, opened 2016-04-21
Factor out common error dispatch call functionality into helper

akiss77:


zherczeg:
Nice catch. How much does it affect binary size?

akiss77:
@zherczeg In release LTO, cca. 500 bytes of text gained on arm-linux and 300-500 bytes on arm-mcu. 2K+ in debug.

LaszloLango:
LGTM

zherczeg:
LGTM

PR #1028, opened 2016-04-23
Drop jerry_port_putchar

akiss77:
Recent changes eliminate the need for `jerry_port_putchar`. As port
API discussions don't make it likely that it will ever be needed
again, this patch removes its declaration from jerry-port.h and its
implementations from the port(s).

The related code in jerry-libc is not needed either: whatever `putc`
(and `puts`) can do, `printf` can do as well.


LaszloLango:
LGTM

zherczeg:
LGTM

PR #1029, opened 2016-04-25
Force enums in snapshots to be 32bits wide

franc0is:
Assumptions are hard-coded elsewhere that require these enums to be 32 bits in size, yet it isn't enforced and some compilers will shorten them.

JerryScript-DCO-1.0-Signed-off-by: François Baldassari <francois@pebble.com>


zherczeg:
I think it is bad if we directly store enums in the snapshot. I would prefer to change type to uint8_t rather than adding a high value.

LaszloLango:
@franc0is, I agree with @zherczeg. Do you plan to update your PR?

zherczeg:
After #1168 this change becomes invalid.

zherczeg:
May I close this PR?

zherczeg:
I close this bug. Please feel free to repoen if needed.

PR #1033, opened 2016-04-26
Check whether the aligned pointers value of the internal properties c…

robertsipka:
…an be stored directly

in ecma_value_t or not.

This change affects those internal properties that haven't been checked previously.


LaszloLango:
Please fix the format of the commit message.

robertsipka:
Sure, I've updated the commit message.

zherczeg:
Patch is ok, but I would prefer a bit better description of the commit message.

robertsipka:
Thanks for the review, I've updated the commit message, and I've rebased this patch with the current master.

zherczeg:
LGTM

LaszloLango:
LGTM

PR #1034, opened 2016-04-29
Add context support for memory allocator

zherczeg:
This is the first patch which aims to move all jerry global variables into a context. This patch contains the infrastructure part, and the memory manager is updated to use this context. The JERRY_CONTEXT macro should provide the freedom of using contexts in Jerry.

@franc0is if you have some time please check this patch.

### DIFF
--- a/jerry-core/CMakeLists.txt
+++ b/jerry-core/CMakeLists.txt
@@ -105,16 +105,16 @@ project (JerryCore C ASM)
 # Include directories
  set(INCLUDE_CORE
      ${CMAKE_SOURCE_DIR}/jerry-core
-     ${CMAKE_SOURCE_DIR}/jerry-core/lit
-     ${CMAKE_SOURCE_DIR}/jerry-core/rcs
-     ${CMAKE_SOURCE_DIR}/jerry-core/jmem
-     ${CMAKE_SOURCE_DIR}/jerry-core/vm
-     ${CMAKE_SOURCE_DIR}/jerry-core/ecma/builtin-objects
      ${CMAKE_SOURCE_DIR}/jerry-core/ecma/base
+     ${CMAKE_SOURCE_DIR}/jerry-core/ecma/builtin-objects
      ${CMAKE_SOURCE_DIR}/jerry-core/ecma/operations
+     ${CMAKE_SOURCE_DIR}/jerry-core/jcontext
+     ${CMAKE_SOURCE_DIR}/jerry-core/jmem
+     ${CMAKE_SOURCE_DIR}/jerry-core/jrt
+     ${CMAKE_SOURCE_DIR}/jerry-core/lit
      ${CMAKE_SOURCE_DIR}/jerry-core/parser/js
      ${CMAKE_SOURCE_DIR}/jerry-core/parser/regexp
-     ${CMAKE_SOURCE_DIR}/jerry-core/jrt)
+     ${CMAKE_SOURCE_DIR}/jerry-core/vm)
 
  # Third-party
   # Valgrind
@@ -123,29 +123,29 @@ project (JerryCore C ASM)
 # Sources
  # Jerry core
   file(GLOB SOURCE_CORE_API                   *.c)
-  file(GLOB SOURCE_CORE_LIT                   lit/*.c)
-  file(GLOB SOURCE_CORE_RCS                   rcs/*.c)
-  file(GLOB SOURCE_CORE_MEM                   jmem/*.c)
-  file(GLOB SOURCE_CORE_VM                    vm/*.c)
-  file(GLOB SOURCE_CORE_ECMA_BUILTINS         ecma/builtin-objects/*.c)
   file(GLOB SOURCE_CORE_ECMA_BASE             ecma/base/*.c)
+  file(GLOB SOURCE_CORE_ECMA_BUILTINS         ecma/builtin-objects/*.c)
   file(GLOB SOURCE_CORE_ECMA_OPERATIONS       ecma/operations/*.c)
+  file(GLOB SOURCE_CORE_JCONTEXT              jcontext/*.c)
+  file(GLOB SOURCE_CORE_JMEM                  jmem/*.c)
+  file(GLOB SOURCE_CORE_JRT                   jrt/*.c)
+  file(GLOB SOURCE_CORE_LIT                   lit/*.c)
   file(GLOB SOURCE_CORE_PARSER_JS             parser/js/*.c)
   file(GLOB SOURCE_CORE_PARSER_REGEXP         parser/regexp/*.c)
-  file(GLOB SOURCE_CORE_JRT                   jrt/*.c)
+  file(GLOB SOURCE_CORE_VM                    vm/*.c)
 
   set(SOURCE_CORE_FILES
       ${SOURCE_CORE_API}
-      ${SOURCE_CORE_LIT}
-      ${SOURCE_CORE_RCS}
-      ${SOURCE_CORE_MEM}
-      ${SOURCE_CORE_VM}
-      ${SOURCE_CORE_ECMA_BUILTINS}
       ${SOURCE_CORE_ECMA_BASE}
+      ${SOURCE_CORE_ECMA_BUILTINS}
       ${SOURCE_CORE_ECMA_OPERATIONS}
+      ${SOURCE_CORE_JCONTEXT}
+      ${SOURCE_CORE_JMEM}
+      ${SOURCE_CORE_JRT}
+      ${SOURCE_CORE_LIT}
       ${SOURCE_CORE_PARSER_JS}
       ${SOURCE_CORE_PARSER_REGEXP}
-      ${SOURCE_CORE_JRT})
+      ${SOURCE_CORE_VM})
 
  # Jerry port
  if(NOT ("${PORT_DIR}" STREQUAL ""))
--- a/jerry-core/jcontext/jcontext.c
+++ b/jerry-core/jcontext/jcontext.c
@@ -1,0 +1,48 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jcontext.h"
+
+/** \addtogroup context Jerry context
+ * @{
+ *
+ * \addtogroup context Context
+ * @{
+ */
+
+/**
+ * Global context.
+ */
+jerry_context_t jerry_global_context;
+
+/**
+ * Jerry global heap section attribute.
+ */
+#ifndef JERRY_HEAP_SECTION_ATTR
+#define JERRY_GLOBAL_HEAP_SECTION
+#else /* JERRY_HEAP_SECTION_ATTR */
+#define JERRY_GLOBAL_HEAP_SECTION __attribute__ ((section (JERRY_HEAP_SECTION_ATTR)))
+#endif /* !JERRY_HEAP_SECTION_ATTR */
+
+/**
+ * Global heap.
+ */
+jmem_heap_t jerry_global_heap __attribute__ ((aligned (JMEM_ALIGNMENT))) JERRY_GLOBAL_HEAP_SECTION;
+
+/**
+ * @}
+ * @}
+ */
--- a/jerry-core/jcontext/jcontext.h
+++ b/jerry-core/jcontext/jcontext.h
@@ -1,0 +1,111 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Memory context for JerryScript
+ */
+#ifndef JCONTEXT_H
+#define JCONTEXT_H
+
+#include "jrt.h"
+#include "jmem-allocator.h"
+#include "jmem-config.h"
+
+/** \addtogroup context Jerry context
+ * @{
+ *
+ * \addtogroup context Context
+ * @{
+ */
+
+/**
+ * Calculate heap area size, leaving space for a pointer to the free list
+ */
+#define JMEM_HEAP_AREA_SIZE (JMEM_HEAP_SIZE - JMEM_ALIGNMENT)
+
+/**
+ * Heap structure
+ *
+ * Memory blocks returned by the allocator must not start from the
+ * beginning of the heap area because offset 0 is reserved for
+ * JMEM_CP_NULL. This special constant is used in several places,
+ * e.g. it marks the end of the property chain list, so it cannot
+ * be eliminated from the project. Although the allocator cannot
+ * use the first 8 bytes of the heap, nothing prevents to use it
+ * for other purposes. Currently the free region start is stored
+ * there.
+ */
+typedef struct
+{
+  jmem_heap_free_t first; /**< first node in free region list */
+  uint8_t area[JMEM_HEAP_AREA_SIZE]; /**< heap area */
+} jmem_heap_t;
+
+/**
+ * JerryScript context
+ *
+ * The purpose of this header is storing
+ * all global variables for Jerry
+ */
+typedef struct
+{
+  /**
+   * Memory manager part.
+   */
+  size_t jmem_heap_allocated_size; /**< size of allocated regions */
+  size_t jmem_heap_limit; /**< current limit of heap usage, that is upon being reached,
+                           *   causes call of "try give memory back" callbacks */
+  jmem_heap_free_t *jmem_heap_list_skip_p; /**< This is used to speed up deallocation. */
+  jmem_pools_chunk_t *jmem_free_chunk_p; /**< list of free pool chunks */
+  jmem_free_unused_memory_callback_t jmem_free_unused_memory_callback; /**< Callback for freeing up memory. */
+
+#ifdef JMEM_STATS
+  jmem_heap_stats_t jmem_heap_stats; /**< heap's memory usage statistics */
+  jmem_pools_stats_t jmem_pools_stats; /**< pools' memory usage statistics */
+#endif /* MEM_STATS */
+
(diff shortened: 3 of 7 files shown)


franc0is:
Awesome! I'll take a closer look in a few hours.

jerry-core/mem/mem-context.c
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "mem-context.h"
+
+/** \addtogroup mem Memory allocation
+ * @{
+ *
+ * \addtogroup heap Context
```
akiss77: Is this really heap-related (or just a copy-paste issue)?

jerry-core/mem/mem-context.h
```diff
@@ -0,0 +1,119 @@
...
+ */
+typedef struct
+{
+  /**
+   * Memory manager part.
+   */
+  size_t mem_heap_allocated_size; /**< size of allocated regions */
+  size_t mem_heap_limit; /**< current limit of heap usage, that is upon being reached,
+                          *   causes call of "try give memory back" callbacks */
+  mem_heap_free_t *mem_heap_list_skip_p; /**< This is used to speed up deallocation. */
```
franc0is: while this is ok for a first pass, it is odd to see these variables be part of the context rather than the heap.

jerry-core/mem/mem-context.h
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Jerry global heap.
+ */
+extern mem_heap_t jerry_global_heap;
+
+/**
+ * Provides a reference to a field in the current context.
+ */
+#define JERRY_CONTEXT(field) (jerry_global_context.field)
```
franc0is: I like this, it will make it easier to add more logic here if we need it later.

jerry-core/mem/mem-heap.c
```diff
@@ -199,23 +146,29 @@ static void mem_heap_stat_free_iter ();
 void
 mem_heap_init (void)
 {
-  JERRY_STATIC_ASSERT ((uintptr_t) mem_heap.area % MEM_ALIGNMENT == 0,
+  JERRY_STATIC_ASSERT ((uintptr_t) MEM_HEAP_CONTEXT (area) % MEM_ALIGNMENT == 0,
```
franc0is: Careful with the static asserts. The `MEM_HEAP_CONTEXT` macro could change so it would not resolve statically. Either change it to a runtime assert or add a comment near `MEM_HEAP_CONTEXT` that states something like "this needs to resolve statically as it is used in static asserts"

jerry-core/mem/mem-context.h
```diff
@@ -0,0 +1,119 @@
...
+#ifdef JERRY_VALGRIND_FREYA
+  bool valgrind_freya_mempool_request; /**< Tells whether a pool manager
+                                        *   allocator request is in progress */
+#endif /* JERRY_VALGRIND_FREYA */
+} jerry_context_t;
+
+/**
+ * Jerry global context.
+ */
+extern jerry_context_t jerry_global_context;
```
franc0is: Since you've separated JERRY_CONTEXT and MEM_HEAP_CONTEXT, it would make sense for jerry_context_t to live outside of the `mem` folder eventually.

franc0is:
Looks great :+1:

jerry-core/mem/mem-context.c
zherczeg: I have been thinking what is the best place for the context in Jerry, and for me the allocator is, since even if it is statically allocated, it is still "allocated". It could go to jrt, but for me that does not make sense. It cannot go to ecma, because mem cannot depend on ecma. I am open to better suggestions though.

jerry-core/mem/mem-context.h
zherczeg: The "context" is quite general for me. It should contain everything except heap and lcache (those two are big arrays). My plan for the layout when context is dynamically allocated:

    [ context ][ heap ][ lcache ]

The context pointer points to the start of the heap. The context members are accessed by 

   ((mem_context_t*) heap_ptr)[-1].member

Even on Thumb2 ARM +- 255 offset is supported, so we can access the fields of the context easily (provided the size of the context is < 255, but I do hope it will be). We should not have too many global variables.

Similar to the previous one, I am open to suggestions here.

jerry-core/mem/mem-heap.c
zherczeg: Agreed, I will change this to dynamic assert.

jerry-core/mem/mem-context.c
franc0is: I think `jrt` may be a bit better, but I see what you mean. Your call.

jerry-core/mem/mem-context.c
```diff
@@ -0,0 +1,55 @@
...
+ * Jerry global heap section attribute.
+ */
+#define JERRY_GLOBAL_HEAP_SECTION __attribute__ ((section (JERRY_HEAP_SECTION_ATTR)))
+
+/**
+ * Global heap.
+ */
+mem_heap_t jerry_global_heap __attribute__ ((aligned (MEM_ALIGNMENT))) JERRY_GLOBAL_HEAP_SECTION;
+
+#endif /* !JERRY_HEAP_SECTION_ATTR */
```
akiss77: This a is bit overly complex. You could simplify it as:
```c
#ifndef JERRY_HEAP_SECTION_ATTR
#define JERRY_GLOBAL_HEAP_SECTION
#else /* JERRY_HEAP_SECTION_ATTR */
/**
 * Jerry global heap section attribute.
 */
#define JERRY_GLOBAL_HEAP_SECTION __attribute__ ((section (JERRY_HEAP_SECTION_ATTR)))
#endif /* !JERRY_HEAP_SECTION_ATTR */

/**
 * Global heap.
 */
mem_heap_t jerry_global_heap __attribute__ ((aligned (MEM_ALIGNMENT))) JERRY_GLOBAL_HEAP_SECTION;
```
This would move the double definiton of the variable outside the macro guards and remove duplication.

zherczeg:
I think we should decide where to put this context. After some thinking I would put in the jerry-core/ root directory or a new jerry-core/context/ sub-directory

akiss77:
I would have favoured "jerry-core/jrt" but certainly not "jerry-core/", which leaves me with voting for a new subdir under jerry-core.

LaszloLango:
+1 to a new `jerry-core/context` directory.

zherczeg:
Patch reworked

LaszloLango:
Sorting the source directories in CMake is a good idea :) I'd like to see it in the commit message too. In general, a more descriptive commit message would be great. The code is LGTM.

jerry-core/context/context.c
```diff
@@ -0,0 +1,48 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "context.h"
+
+/** \addtogroup context Jerry context
+ * @{
+ *
+ * \addtogroup context Context
```
akiss77: Is the two levels deep gruoping intentional? Will the context doc group have multiple sub-groups? Or is a single context group enough for everything here?

franc0is:
LGTM

jerry-core/context/context.h
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Jerry global context.
+ */
+extern jerry_context_t jerry_global_context;
+
+/**
+ * Jerry global heap.
+ */
+extern mem_heap_t jerry_global_heap;
```
akiss77: I'd strongly suggest merging most information of `mem_heap_t` and `jerry_global_heap` into `jerry_global_context`. Something along the lines:
```c
typedef struct
{
  // whatever already is here extended with the fields below:
  mem_heap_free_t first; /**< first node in free region list */
  uint8_t *area; /**< heap area */
} jerry_context_t;
```
And then, you could move the relevant code from context.c to mem-heap.c:
```c
/**
+ * Jerry global heap section attribute.
+ */
+#ifndef JERRY_HEAP_SECTION_ATTR
+#define JERRY_GLOBAL_HEAP_SECTION
+#else /* JERRY_HEAP_SECTION_ATTR */
+#define JERRY_GLOBAL_HEAP_SECTION __attribute__ ((section (JERRY_HEAP_SECTION_ATTR)))
+#endif /* !JERRY_HEAP_SECTION_ATTR */
+
+/**
+ * Global heap.
+ */
uint8_t jerry_global_heap_area[MEM_HEAP_AREA_SIZE] __attribute__ ((aligned (MEM_ALIGNMENT))) JERRY_GLOBAL_HEAP_SECTION;
```
And initialize the area pointer in `mem_heap_init` with:
```c
JERRY_CONTEXT (area) = jerry_global_heap_area;
```

This would remove the need for two context globals and two context macros, and would also make setting the heap pointer with something not statically allocated easier later in the future, while keeping the possibility to have the heap put in a separate section right now.

zherczeg: I would prefer to keep two areas because of the performance reasons discussed before. When somebody has a single context pointer, the area should be organized like this [context][heap], and the pointer should point to [heap]. This would allow fast pointer compression and context could be reached easily, because negative absolute offsets are supported on all popular CPUs.

akiss77: I'm not sure I get this concept.

Context is (or, should be) a "thing" that represents all the state of the engine. Engine-managed heap, any caches, or anything else required to execute JS in/by the engine should (must) be part of the context. That should be the only entity that the user of the library has to deal with if s/he wants to change ... well, context. Dealing with the internals -- lile, heap -- should be off limits for the user. Thus, s/he should not pass in a heap pointer or something like that.

My understanding is that we concluded at the workshop that -- at least for now -- we start with the implementation of a single global context, which advances the current state by (1) ensuring that all global state information is in one place, (2) that state is always initialized, and (3) it can be swapped/replaced. (I.e., we don't pass a context object/pointer around the engine all the time but allow a global context struct -- or pointer to such a struct -- to be set by public API calls.)

I don't see this, especially (3), in the above explanation.

Could you please benchmark how the current PR performs (in its current form, without the mentioned context+heap layout optimization) and what performance degradation the suggested approach of merging the two contexts would cause (if any)?

akiss77: BTW, the outlined [context][heap] memory layout cannot be ensured as heap can end up in a completely different section to context (because of `JERRY_HEAP_SECTION_ATTR`).

zherczeg: This concept provides nice optimization opportunities in the future, and I don't think it is worth sacrificing them. I want three contexts: for simple variables, heap, and lookup cache. If you prefer you can put all of them into a super structure (it can be the default), but you can also organize them cleverly. Obviously it is not relevant for static data, but can have nice benefit for other purposes. The point of the current approach is not putting everything into one structure, but to put everything into one file. People can reorganize this data according to their needs.

zherczeg: The described layout is useful when you have a context pointer, not a static context. The implementation there is kind of a default approach, but I expecting to add more layouts in the future. They don't exists at this point, but I don't want to limit ourselves.

akiss77: > I want three contexts: for simple variables, heap, and lookup cache.

That's completely against the concept of context. The lookup cache of one context has no meaning in another context with a different heap. All the three entities you mentioned belong together and form the state of the engine. Just putting everything in one file is no context.

zherczeg: Yes they are no context, but the building blocks of a context. The point is, you take these blocks, and build your own context. One opportunity is defining static variables for each block individually, another is putting all (or some) them in a large structure.

akiss77: - It's still unclear to me who is "you" above. How will it be possible to put these blocks together in different ways? All the code above is strictly jerry internal with noone (no user or port API) having access to it.

- I still don't see why the type `mem_heap_t` is needed. Even with the layout you depicted, `mem_heap_free_t first;` could go into `jerry_context_t` (perhaps with a slightly different name) as it already has several heap-related fields (e.g., `mem_heap_free_t *mem_heap_list_skip_p;`). That would leave the heap as a byte array, and the `MEM_HEAP_CONTEXT` macro would need no argument at all as it would/could only refer to the heap area.

- I'm still interested in the benchmark results.

zherczeg: 1) there will be different modes like the floating point representations (e.g. 32/64 bit doubles) (I expect 3-5 modes maximum). Ports will choose the appropriate mode with a compiler flags (one port likely will use only one mode).

2) The mem_heap_t is very much needed, since the two fields are tightly coupled. Actually the "first" is the first 8 bytes of the heap, and provides nicer access to this 8 bytes than casting uint8_t* all the time.

3) What do you expect here? Why the results should be much different? I expect that there is no difference between the current several globals spread in the code and a few global variables. From compiler viewpoint, nothing changed (and that is the plan). But this is just mode 1. For other modes, there will likely be changes, but their purpose is different. In a future mode with a single global context pointer, your primary aim is having one changable pointer, and not being faster than globals.

LaszloLango: I like @zherczeg's idea. This is a good first step.

LaszloLango:
LGTM

jerry-core/context/context.h
akiss77: >2) The mem_heap_t is very much needed, since the two fields are tightly coupled. Actually the "first" is the first 8 bytes of the heap, and provides nicer access to this 8 bytes than casting uint8_t* all the time.

I don't think that this is right/correct. I've tried to trace down all uses of the `first` field and it is never used as part of the heap. I cannot find a place where it would be referred to via a compact pointer. It is never used in a more tightly coupled way than `mem_heap_list_skip_p`, which is in the "regular" context object.

>3) What do you expect here?

My original suggestion (the first in this thread) was to have a pointer in the context object to a heap area, which is "somewhere"; instead of having a globally allocated byte array for the heap. You've argued against it because of performance reasons. That's why I asked for benchmark results.

zherczeg:
I tried the pointer to heap version. Performance is roughly the same (gain is less than the deviation), but the binary size is increased by 600 bytes. I would still prefer the original.

zherczeg:
Updated and rebased to the latest master.

LaszloLango:
Still LGTM

zherczeg:
Added a long comment why the "first" is in a good place.

zherczeg:
Patch rebased.

dbatyai:
LGTM

jerry-core/jcontext/jcontext.c
```diff
@@ -0,0 +1,48 @@
...
+ * @{
+ *
+ * \addtogroup context Context
+ * @{
+ */
+
+/**
+ * Global context.
+ */
+jerry_context_t jerry_global_context;
```
jiangzidong: Will the name "global_context" be misleading? People may think it's global environment/scope

What about "native_context"?

zherczeg: I am not sure I understand. It is a global environment. This is an initial patch, which collects global variables into a structure, and later this structure can be used to support different global management. E.g. multiple Jerries in one process, optional memory allocation for Jerry, etc.

jiangzidong: Sorry that I didn't put it clearly. Here "context" I think it is similar with V8 context or Node.js's vm.runInXXContext, right? If so, each context will have its own global env,  like **ecma_global_lex_env_p** in jerry-core\ecma\operations\ecma-lex-env.c
The followings are my understanding, please correct me if I am wrong.
jerry_global_context is the default context when we run jerry. in the future, we will support creating new context (let's name it "context2") and leting some js codes run in the context2. In the above case, both jerry_global_context and context2 will have respective ecma_global_lex_env_p. But the name of "jerry_global_context" may let people feel it is the global env.

zherczeg: My plan here is a bit different. Jerry will support more "context models" in the future and these context models will be selected by an appropriate defines. E.g. the current context model (which will likely be the default context model) supports only a single global context. Other context models (e.g. JERRY_CONTEXT_MODEL_USER_CONTEXT_PTR) will defines other variables and structures. The point is, all context models will reuse the structures above for convenience. Hence creating new context models will be easy.

jiangzidong: I am sorry that I didn't fully understand. In a case which has 2 contexts, do the 2 context have 2 different jerry_global_context? or there is only one jerry_global_context in one jerry process? (I think only one jerry_global_context in one jerry process)

zherczeg: The plan is to have more contexts. But the other modes has not implemented yet, so we are quite free to do whatever we want. Those modes which support multiple contexts jerry_global_context will obviously not a good name, so they will define something like jerry_current_context, and this will be a pointer type.

I think a pseudo code will makes this easier to understand:

    #ifdef JERRY_CONTEXT_MODE_SOMETHING1
    // Whatever we want, e.g. a pointer to the current context
    #elif defined JERRY_CONTEXT_MODE_SOMETHING2
    // Whatever we want, e.g. a callback is defined to get the current context
    #else
    // Default context, simply one global context. This is the only mode we have now.
    jerry_context_t global_context;
    #endif

This is not implemented yet, just the general plan. The code will evolve into this direction. So we can define any kind of context modes depending on user requests. Hence we will not be limited to one context model like v8, but we can adapt several different ones.

jiangzidong: I see, Thanks very much

zherczeg: No problem. I admit the plan was not clear from the patch. Different context modes has different advantages / disadvantages, and we don't want to limit choices here. Instead everybody should choose at compile time the best context model for them.

jerry-core/CMakeLists.txt
```diff
@@ -105,16 +105,16 @@ project (JerryCore C ASM)
 # Include directories
  set(INCLUDE_CORE
      ${CMAKE_SOURCE_DIR}/jerry-core
-     ${CMAKE_SOURCE_DIR}/jerry-core/lit
-     ${CMAKE_SOURCE_DIR}/jerry-core/rcs
-     ${CMAKE_SOURCE_DIR}/jerry-core/jmem
-     ${CMAKE_SOURCE_DIR}/jerry-core/vm
-     ${CMAKE_SOURCE_DIR}/jerry-core/ecma/builtin-objects
+     ${CMAKE_SOURCE_DIR}/jerry-core/jcontext
```
akosthekiss: (minor) this violates alpha order, move it down before jmem

jerry-core/CMakeLists.txt
```diff
@@ -123,29 +123,29 @@ project (JerryCore C ASM)
 # Sources
  # Jerry core
   file(GLOB SOURCE_CORE_API                   *.c)
-  file(GLOB SOURCE_CORE_LIT                   lit/*.c)
-  file(GLOB SOURCE_CORE_RCS                   rcs/*.c)
-  file(GLOB SOURCE_CORE_MEM                   jmem/*.c)
-  file(GLOB SOURCE_CORE_VM                    vm/*.c)
-  file(GLOB SOURCE_CORE_ECMA_BUILTINS         ecma/builtin-objects/*.c)
+  file(GLOB SOURCE_CORE_JCONTEXT              jcontext/*.c)
```
akosthekiss: ditto

jerry-core/CMakeLists.txt
```diff
@@ -123,29 +123,29 @@ project (JerryCore C ASM)
...
+  file(GLOB SOURCE_CORE_VM                    vm/*.c)
 
   set(SOURCE_CORE_FILES
       ${SOURCE_CORE_API}
-      ${SOURCE_CORE_LIT}
-      ${SOURCE_CORE_RCS}
-      ${SOURCE_CORE_MEM}
-      ${SOURCE_CORE_VM}
-      ${SOURCE_CORE_ECMA_BUILTINS}
+      ${SOURCE_CORE_JCONTEXT}
```
akosthekiss: ditto

akosthekiss:
Other than the minor stylistic comments, LGTM

zherczeg:
Thank you for the review. The patch and its description is updated.

LaszloLango:
Still LGTM

PR #1039, opened 2016-05-04
Introduce integer ecma-value representation to reduce the double allocations

zherczeg:
The patch is an initial patch to store 28 bit integer values in ecma_values. This is a first patch in a series of patches which exploits these integer numbers to accelerate math performance, since most numbers are small integer numbers in JavaScript. The memory saving is resulted by not allocating doubles for these integer numbers. The follow-up patches will likely not affect memory.

RPi2 results
```

                      Benchmark |          RSS(+ is better) |                   Perf(+ is better)
                     3d-cube.js |   84k ->   68k : +19.048% |   2.176s ->   2.096s :  +3.676%  (+-0.720%) : [+]
         access-binary-trees.js |   72k ->   72k :  +0.000% |   1.200s ->   1.144s :  +4.667%  (+-0.940%) : [+]
             access-fannkuch.js |   32k ->   32k :  +0.000% |   6.366s ->   6.084s :  +4.430%  (+-0.515%) : [+]
                access-nbody.js |   36k ->   36k :  +0.000% |   2.414s ->   2.452s :  -1.574%  (+-1.491%) : [-]
    bitops-3bit-bits-in-byte.js |   24k ->   24k :  +0.000% |   1.378s ->   1.336s :  +3.048%  (+-1.462%) : [+]
         bitops-bits-in-byte.js |   24k ->   24k :  +0.000% |   2.058s ->   1.894s :  +7.969%  (+-0.685%) : [+]
          bitops-bitwise-and.js |   24k ->   24k :  +0.000% |   2.798s ->   2.882s :  -3.002%  (+-1.157%) : [-]
          bitops-nsieve-bits.js |  164k ->  164k :  +0.000% |   3.956s ->   4.042s :  -2.174%  (+-0.743%) : [-]
       controlflow-recursive.js |  124k ->  112k :  +9.677% |   0.808s ->   0.740s :  +8.416%  (+-1.044%) : [+]
                  crypto-aes.js |  104k ->   96k :  +7.692% |   2.898s ->   2.690s :  +7.177%  (+-0.583%) : [+]
                  crypto-md5.js |  168k ->  168k :  +0.000% |   1.194s ->   1.202s :  -0.670%  (+-1.224%) : [≈]
                 crypto-sha1.js |  116k ->  116k :  +0.000% |   1.130s ->   1.144s :  -1.239%  (+-0.998%) : [-]
           date-format-tofte.js |   56k ->   56k :  +0.000% |   1.730s ->   1.648s :  +4.740%  (+-0.962%) : [+]
           date-format-xparb.js |   52k ->   52k :  +0.000% |   0.742s ->   0.690s :  +7.008%  (+-1.154%) : [+]
                 math-cordic.js |   28k ->   24k : +14.286% |   2.456s ->   2.486s :  -1.221%  (+-1.071%) : [-]
           math-partial-sums.js |   24k ->   24k :  +0.000% |   1.346s ->   1.410s :  -4.755%  (+-0.878%) : [-]
          math-spectral-norm.js |   32k ->   32k :  +0.000% |   1.190s ->   1.188s :  +0.168%  (+-0.774%) : [≈]
               string-base64.js |  156k ->  148k :  +5.128% |   5.290s ->   5.178s :  +2.117%  (+-1.191%) : [+]
                string-fasta.js |   40k ->   40k :  +0.000% |   2.702s ->   2.688s :  +0.518%  (+-1.502%) : [≈]
                Geometric mean: |     RSS reduction: 3.108% |                   Speed up: 2.146% (+-0.244%) : [+]

Binaries
master: 173,568
branch: 173,480
```

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -55,8 +55,8 @@
  */
 typedef enum
 {
-  ECMA_TYPE_SIMPLE, /**< simple value */
-  ECMA_TYPE_NUMBER, /**< 64-bit integer */
+  ECMA_TYPE_DIRECT, /**< directly encoded value, a 28 bit signed integer or a simple value */
+  ECMA_TYPE_FLOAT, /**< pointer to a 64 or 32 bit floating point number */
   ECMA_TYPE_STRING, /**< pointer to description of a string */
   ECMA_TYPE_OBJECT, /**< pointer to description of an object */
   ECMA_TYPE___MAX = ECMA_TYPE_OBJECT /** highest value for ecma types */
@@ -68,10 +68,10 @@ typedef enum
 typedef enum
 {
   /**
-   * Empty value is implementation defined value, used for:
-   *   - representing empty value in completion values (see also: ECMA-262 v5, 8.9 Completion specification type);
-   *   - values of uninitialized immutable bindings;
-   *   - values of empty register variables.
+   * Empty value is implementation defined value, used for representing:
+   *   - empty (uninitialized) values
+   *   - immutable binding values
+   *   - special register or stack values for vm
    */
   ECMA_SIMPLE_VALUE_EMPTY,
   ECMA_SIMPLE_VALUE_UNDEFINED, /**< undefined value */
@@ -89,6 +89,11 @@ typedef enum
  * Bit-field structure: type (2) | error (1) | value (29)
  */
 typedef uint32_t ecma_value_t;
+
+/**
+ * Type for directly encoded integer numbers in JerryScript.
+ */
+typedef int32_t ecma_integer_value_t;
 
 #if UINTPTR_MAX <= UINT32_MAX
 
@@ -113,6 +118,50 @@ typedef uint32_t ecma_value_t;
  * Shift for value part in ecma_type_t
  */
 #define ECMA_VALUE_SHIFT 3
+
+/**
+ * Mask for directly encoded values
+ */
+#define ECMA_DIRECT_TYPE_MASK ((1u << ECMA_VALUE_SHIFT) | ECMA_VALUE_TYPE_MASK)
+
+/**
+ * Ecma integer value type
+ */
+#define ECMA_DIRECT_TYPE_INTEGER_VALUE ((0u << ECMA_VALUE_SHIFT) | ECMA_TYPE_DIRECT)
+
+/**
+ * Ecma simple value type
+ */
+#define ECMA_DIRECT_TYPE_SIMPLE_VALUE ((1u << ECMA_VALUE_SHIFT) | ECMA_TYPE_DIRECT)
+
+/**
+ * Shift for directly encoded values in ecma_value_t
+ */
+#define ECMA_DIRECT_SHIFT 4
+
+/**
+ * Maximum integer number for an ecma value
+ */
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
+#define ECMA_INTEGER_NUMBER_MAX 0x7fffff
+#else /* CONFIG_ECMA_NUMBER_TYPE != CONFIG_ECMA_NUMBER_FLOAT32 */
+#define ECMA_INTEGER_NUMBER_MAX 0x7ffffff
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32 */
+
+/**
+ * Minimum integer number for an ecma value
+ */
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
+#define ECMA_INTEGER_NUMBER_MIN -0x7fffff
+#else /* CONFIG_ECMA_NUMBER_TYPE != CONFIG_ECMA_NUMBER_FLOAT32 */
+#define ECMA_INTEGER_NUMBER_MIN -0x8000000
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32 */
+
+/**
+ * Checks whether the integer number is in the integer number range.
+ */
+#define ECMA_IS_INTEGER_NUMBER(num) \
+  (ECMA_INTEGER_NUMBER_MIN <= (num) && (num) <= ECMA_INTEGER_NUMBER_MAX)
 
 /**
  * Internal properties' identifiers.
--- a/jerry-core/ecma/base/ecma-helpers-value.c
+++ b/jerry-core/ecma/base/ecma-helpers-value.c
@@ -207,6 +207,30 @@ ecma_is_value_array_hole (ecma_value_t value) /**< ecma value */
 } /* ecma_is_value_array_hole */
 
 /**
+ * Check if the value is integer ecma-number.
+ *
+ * @return true - if the value contains an integer ecma-number value,
+ *         false - otherwise.
+ */
+inline bool __attr_pure___ __attr_always_inline___
+ecma_is_value_integer_number (ecma_value_t value) /**< ecma value */
+{
+  return (value & ECMA_DIRECT_TYPE_MASK) == ECMA_DIRECT_TYPE_INTEGER_VALUE;
+} /* ecma_is_value_integer_number */
+
+/**
+ * Check if the value is floating-point ecma-number.
+ *
+ * @return true - if the value contains a floating-point ecma-number value,
+ *         false - otherwise.
+ */
+inline bool __attr_pure___ __attr_always_inline___
+ecma_is_value_float_number (ecma_value_t value) /**< ecma value */
+{
+  return (ecma_get_value_type_field (value) == ECMA_TYPE_FLOAT);
+} /* ecma_is_value_float_number */
+
+/**
  * Check if the value is ecma-number.
  *
  * @return true - if the value contains ecma-number value,
@@ -215,7 +239,8 @@ ecma_is_value_array_hole (ecma_value_t value) /**< ecma value */
 inline bool __attr_pure___ __attr_always_inline___
 ecma_is_value_number (ecma_value_t value) /**< ecma value */
 {
-  return (ecma_get_value_type_field (value) == ECMA_TYPE_NUMBER);
+  return (ecma_is_value_integer_number (value)
+          || ecma_is_value_float_number (value));
 } /* ecma_is_value_number */
 
 /**
@@ -275,19 +300,133 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
 inline ecma_value_t __attr_const___ __attr_always_inline___
 ecma_make_simple_value (const ecma_simple_value_t simple_value) /**< simple value */
 {
-  return (((ecma_value_t) (simple_value)) << ECMA_VALUE_SHIFT) | ECMA_TYPE_SIMPLE;
+  return (((ecma_value_t) (simple_value)) << ECMA_DIRECT_SHIFT) | ECMA_DIRECT_TYPE_SIMPLE_VALUE;
 } /* ecma_make_simple_value */
 
 /**
- * Number value constructor
+ * Encode an integer number into an ecma-value without allocating memory
+ *
+ * Note:
+ *   The value must fit into the range of allowed ecma integer values
+ *
+ * @return ecma-value
  */
-ecma_value_t __attr_const___
-ecma_make_number_value (const ecma_number_t *ecma_num_p) /**< number to reference in value */
+inline ecma_value_t __attr_const___ __attr_always_inline___
+ecma_make_integer_value (ecma_integer_value_t integer_value) /**< integer number to be encoded */
 {
-  JERRY_ASSERT (ecma_num_p != NULL);
+  JERRY_ASSERT (ECMA_IS_INTEGER_NUMBER (integer_value));
 
-  return ecma_pointer_to_ecma_value (ecma_num_p) | ECMA_TYPE_NUMBER;
+  return ((ecma_value_t) (integer_value << ECMA_DIRECT_SHIFT)) | ECMA_DIRECT_TYPE_INTEGER_VALUE;
+} /* ecma_make_integer_value */
+
+/**
+ * Allocate and initialize a new float number without checks.
+ *
+ * @return ecma-value
+ */
+static ecma_value_t __attr_const___
+ecma_create_float_number (ecma_number_t ecma_number) /**< value of the float number */
+{
+  ecma_number_t *ecma_num_p = ecma_alloc_number ();
+
+  *ecma_num_p = ecma_number;
+
+  return ecma_pointer_to_ecma_value (ecma_num_p) | ECMA_TYPE_FLOAT;
+} /* ecma_create_float_number */
+
+/**
+ * Create a new NaN value.
+ *
+ * @return ecma-value
+ */
+inline ecma_value_t __attr_always_inline___
+ecma_make_nan_value (void)
+{
+  return ecma_create_float_number (ecma_number_make_nan ());
+} /* ecma_make_nan_value */
+
+/**
+ * Checks whether the passed number is +0.0
+ *
+ * @return true, if it is +0.0, false otherwise
+ */
+static inline bool __attr_const___ __attr_always_inline___
+ecma_is_number_equal_to_positive_zero (ecma_number_t ecma_number) /**< number */
+{
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
+  union
+  {
+    uint32_t u32_value;
+    ecma_number_t float_value;
+  } u;
+
+  u.float_value = ecma_number;
+
+  return u.u32_value == 0;
+#else /* CONFIG_ECMA_NUMBER_TYPE != CONFIG_ECMA_NUMBER_FLOAT32 */
+  union
+  {
+    uint64_t u64_value;
+    ecma_number_t float_value;
+  } u;
+
+  u.float_value = ecma_number;
+
+  return u.u64_value == 0;
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32 */
+} /* ecma_is_number_equal_to_positive_zero */
+
+/**
+ * Encode a number into an ecma-value
+ *
+ * @return ecma-value
+ */
+ecma_value_t
+ecma_make_number_value (ecma_number_t ecma_number) /**< number to be encoded */
+{
+  ecma_integer_value_t integer_value = (ecma_integer_value_t) ecma_number;
+
+  if ((ecma_number_t) integer_value == ecma_number
+      && ((integer_value == 0) ? ecma_is_number_equal_to_positive_zero (ecma_number)
+                               : ECMA_IS_INTEGER_NUMBER (integer_value)))
+  {
+    return ecma_make_integer_value (integer_value);
+  }
+
+  return ecma_create_float_number (ecma_number);
 } /* ecma_make_number_value */
+
+/**
+ * Encode an int32 number into an ecma-value
+ *
+ * @return ecma-value
+ */
+ecma_value_t
+ecma_make_int32_value (int32_t int32_number) /**< int32 number to be encoded */
+{
+  if (ECMA_IS_INTEGER_NUMBER (int32_number))
+  {
+    return ecma_make_integer_value ((ecma_integer_value_t) int32_number);
+  }
+
+  return ecma_create_float_number ((ecma_number_t) int32_number);
+} /* ecma_make_int32_value */
+
+/**
+ * Encode an unsigned int32 number into an ecma-value
+ *
+ * @return ecma-value
+ */
+ecma_value_t
+ecma_make_uint32_value (uint32_t uint32_number) /**< uint32 number to be encoded */
+{
+  if (uint32_number <= ECMA_INTEGER_NUMBER_MAX)
+  {
+    return ecma_make_integer_value ((ecma_integer_value_t) uint32_number);
+  }
+
+  return ecma_create_float_number ((ecma_number_t) uint32_number);
+} /* ecma_make_uint32_value */
 
 /**
  * String value constructor
@@ -333,24 +472,48 @@ ecma_make_error_obj_value (const ecma_object_t *object_p) /**< object to referen
 } /* ecma_make_error_obj_value */
 
 /**
- * Get pointer to ecma-number from ecma value
+ * Get floating point value from an ecma value
  *
- * @return the pointer
+ * @return floating point value
  */
-ecma_number_t *__attr_pure___
+ecma_number_t __attr_pure___
 ecma_get_number_from_value (ecma_value_t value) /**< ecma value */
 {
-  JERRY_ASSERT (ecma_get_value_type_field (value) == ECMA_TYPE_NUMBER);
+  if (ecma_is_value_integer_number (value))
+  {
+    return (ecma_number_t) (((ecma_integer_value_t) value) >> ECMA_DIRECT_SHIFT);
+  }
 
-  return (ecma_number_t *) ecma_get_pointer_from_ecma_value (value);
+  JERRY_ASSERT (ecma_get_value_type_field (value) == ECMA_TYPE_FLOAT);
+
+  return *(ecma_number_t *) ecma_get_pointer_from_ecma_value (value);
 } /* ecma_get_number_from_value */
+
+/**
+ * Get uint32 value from an ecma value
+ *
+ * @return floating point value
+ */
+uint32_t __attr_pure___
+ecma_get_uint32_from_value (ecma_value_t value) /**< ecma value */
+{
+  if (ecma_is_value_integer_number (value))
+  {
+    /* Works with negative numbers as well. */
+    return (uint32_t) (((ecma_integer_value_t) value) >> ECMA_DIRECT_SHIFT);
+  }
+
+  JERRY_ASSERT (ecma_get_value_type_field (value) == ECMA_TYPE_FLOAT);
+
+  return ecma_number_to_uint32 (*(ecma_number_t *) ecma_get_pointer_from_ecma_value (value));
+} /* ecma_get_uint32_from_value */
 
 /**
  * Get pointer to ecma-string from ecma value
  *
  * @return the pointer
  */
-ecma_string_t *__attr_pure___
+inline ecma_string_t *__attr_pure___ __attr_always_inline___
 ecma_get_string_from_value (ecma_value_t value) /**< ecma value */
 {
   JERRY_ASSERT (ecma_get_value_type_field (value) == ECMA_TYPE_STRING);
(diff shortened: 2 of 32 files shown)


zherczeg:
The binary size is decreased because a lot of ecma_alloc_number () calls are eliminated.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,146 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+    uint64_t u64_value;
+    ecma_number_t float_value;
+  } u;
 
-  return ecma_pointer_to_ecma_value (ecma_num_p) | ECMA_TYPE_NUMBER;
+  u.float_value = ecma_num;
+
+  return u.u64_value == 0;
+#endif
+}
```
LaszloLango: Missing comment:
https://travis-ci.org/Samsung/jerryscript/jobs/128006828
./jerry-core/ecma/base/ecma-helpers-value.c:376: error: missing comment at the end of function: /* ecma_is_number_equals_to_positive_zero */

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,146 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+inline ecma_value_t __attr_always_inline___
+ecma_make_float_number_value (ecma_number_t ecma_num) /**< value of the float number */
+{
+  JERRY_ASSERT (((ecma_number_t) (ecma_value_integer_t) ecma_num) != ecma_num);
+
+  return ecma_create_float_number (ecma_num);
+} /* ecma_make_float_number_value */
+
+static inline bool __attr_const___ __attr_always_inline___
+ecma_is_number_equals_to_positive_zero (ecma_number_t ecma_num) /**< number */
```
LaszloLango: Missing comment for this function

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -90,6 +90,13 @@ typedef enum
  */
 typedef uint32_t ecma_value_t;
 
+/**
+ * Type for small integer numbers in JerryScript.
+ *
+ * The size of ecma_value_integer_t and ecma_value_t must be the same.
```
LaszloLango: I think we can check this with a static assert

LaszloLango: I just saw you did it. Please move this comment there.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,146 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+  union
+  {
+    uint32_t u32_value;
+    ecma_number_t float_value;
+  } u;
+
+  u.float_value = ecma_num;
+
+  return u.u32_value == 0;
+#else
```
LaszloLango: missing comment
/* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,146 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+  {
+    uint64_t u64_value;
+    ecma_number_t float_value;
+  } u;
 
-  return ecma_pointer_to_ecma_value (ecma_num_p) | ECMA_TYPE_NUMBER;
+  u.float_value = ecma_num;
+
+  return u.u64_value == 0;
+#endif
```
LaszloLango: `#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32 */`

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,146 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
 
 /**
+ * Encode an int32 number into an ecma-value
+ *
+ * @return ecma-value
+ */
+ecma_value_t
+ecma_make_int32_value (int32_t ecma_num) /**< int32 number to be encoded */
+{
+  if (ecma_num <= ECMA_INTEGER_NUMBER_MAX
```
LaszloLango: Could you write like this:
```C++
if (ECMA_INTEGER_NUMBER_MIN <= ecma_num
    && ecma_num <= ECMA_INTEGER_NUMBER_MAX
```
It's better to read because it's more alike `min <= X <= max`

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +613,173 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+} /* ecma_value_assign_value */
+
+/**
+ * Assign a float number to an ecma-value
+ *
+ * Note:
+ *      value previously stored in the property is freed
+ */
+static void
+ecma_value_assign_float_number (ecma_value_t *value_p, /**< ecma value */
```
LaszloLango: why pointer for value?

LaszloLango: Mark as in&out parameter: /**< [in, out] ecma value */

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +613,173 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+} /* ecma_value_assign_float_number */
+
+/**
+ * Assign a number to an ecma-value
+ *
+ * Note:
+ *      value previously stored in the property is freed
+ */
+void
+ecma_value_assign_number (ecma_value_t *value_p, /**< ecma value */
```
LaszloLango: /**< [in, out] ecma value */

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +613,173 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+} /* ecma_value_assign_number */
+
+/**
+ * Assign an int32 value to an ecma-value
+ *
+ * Note:
+ *      value previously stored in the property is freed
+ */
+void
+ecma_value_assign_int32 (ecma_value_t *value_p, /**< ecma value */
```
LaszloLango: /**< [in, out] ecma value */

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +613,173 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+} /* ecma_value_assign_int32 */
+
+/**
+ * Assign an uint32 value to an ecma-value
+ *
+ * Note:
+ *      value previously stored in the property is freed
+ */
+void
+ecma_value_assign_uint32 (ecma_value_t *value_p, /**< ecma value */
```
LaszloLango: /**< [in, out] ecma value */

zherczeg:
Thank you, fixed.

LaszloLango:
LGTM

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -55,8 +55,8 @@
  */
 typedef enum
 {
-  ECMA_TYPE_SIMPLE, /**< simple value */
-  ECMA_TYPE_NUMBER, /**< 64-bit integer */
+  ECMA_TYPE_NON_PTR_VALUE, /**< directly encoded value, a 28 bit signed integer or a simple value */
+  ECMA_TYPE_FLOAT_NUMBER, /**< pointer to a 64 or 32 bit floating point number */
```
akiss77: Terminology: to me, simpler is better. I'd suggest using `ECMA_TYPE_DIRECT` and perhaps even `ECMA_TYPE_FLOAT`. Even the comment of "non-ptr" starts explaining the type as "DIRECTly encoded"; and we haven't used "VALUE" suffices at the other labels (OBJECT, STRING) either. And its better to name something what it _is_ than what it _isn't_.

We could argue that terminology is unimportant enough not to worry about -- but using the right terms helps understanding IMHO.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -90,6 +90,11 @@ typedef enum
  */
 typedef uint32_t ecma_value_t;
 
+/**
+ * Type for small integer numbers in JerryScript.
+ */
+typedef int32_t ecma_value_integer_t;
```
akiss77: Terminology: I think `ecma_integer_t` or `ecma_integer_value_t` would be closer to the naming schemes we are already using.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -115,6 +120,26 @@ typedef uint32_t ecma_value_t;
...
 
 /**
+ * Flag for simple values
+ */
+#define ECMA_SIMPLE_VALUE_FLAG 0x8u
+
+/**
+ * Shift for integer value part in ecma_value_t
+ */
+#define ECMA_INTEGER_NUMBER_SHIFT 4
```
akiss77: Terminology(?): the use of this macro at `ecma_make_simple_value` reveals that this shift is not specific to integers but to all directly encoded (i.e., non-prt) values. Thus, I think that `ECMA_DIRECT_SHIFT` would be a better name.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -115,6 +120,26 @@ typedef uint32_t ecma_value_t;
 #define ECMA_VALUE_SHIFT 3
 
 /**
+ * Flag for simple values
+ */
+#define ECMA_SIMPLE_VALUE_FLAG 0x8u
```
akiss77: This macro is used both as a mask and as a flag/type indicator. I'd propose splitting this into something like
```c
#define ECMA_DIRECT_VALUE_MASK 0x8u
#define ECMA_DIRECT_VALUE_INTEGER_NUMBER (0u << ECMA_VALUE_SHIFT)
#define ECMA_DIRECT_VALUE_SIMPLE_VALUE (1u << ECMA_VALUE_SHIFT)
```
And then code like `ecma_is_value_integer_number` could become:
```c
return (value & (ECMA_DIRECT_VALUE_MASK | ECMA_VALUE_TYPE_MASK)) == (ECMA_DIRECT_VALUE_INTEGER_NUMBER | ECMA_TYPE_DIRECT);
```
This yields the same binary but the source is more readable IMHO. And this could be more future-proof should any other direct values be used in addition to integers and simple values.

jerry-core/ecma/base/ecma-globals.h
zherczeg: Ok. I used this name first, but I didn't particularly like it. But I am not against it either.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,151 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+
+/**
+ * Create a new float number (usually from NaN).
+ *
+ * @return ecma-value
+ */
+inline ecma_value_t __attr_always_inline___
+ecma_make_float_number_value (ecma_number_t ecma_num) /**< value of the float number */
+{
+  JERRY_ASSERT (((ecma_number_t) (ecma_value_integer_t) ecma_num) != ecma_num);
```
akiss77: What exactly does this assert check? I tried to figure it out but I'm not sure I made it. Might also be worth adding a comment before the assert to explain it.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,151 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+  return (ecma_value_t) ((integer_value << ECMA_INTEGER_NUMBER_SHIFT) | ECMA_TYPE_NON_PTR_VALUE);
+} /* ecma_make_integer_value */
+
+/**
+ * Allocate and initialize a new float number without checks.
+ *
+ * @return ecma-value
+ */
+static ecma_value_t __attr_const___
+ecma_create_float_number (ecma_number_t ecma_num) /**< value of the float number */
```
akiss77: Is this function really needed? `ecma_make_float_number_value` seems to be identical to this function except for an assert check. And no calls to this function seem to violate that check. Couldn't all calls to `ecma_create_float_number` be replaced with a call to `ecma_make_float_number_value`?

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,151 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+  return ecma_create_float_number (ecma_num);
+} /* ecma_make_float_number_value */
+
+/**
+ * Checks whether the passed number is +0.0
+ *
+ * @return true, if it is +0.0, false otherwise
+ */
+static inline bool __attr_const___ __attr_always_inline___
+ecma_is_number_equals_to_positive_zero (ecma_number_t ecma_num) /**< number */
```
akiss77: Grammar: if I'm not mistaken, this should be `ecma_is_number_equal_to_positive_zero`

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +618,173 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+    if (integer_value == 0)
+    {
+      if (ecma_is_number_equals_to_positive_zero (ecma_number))
+      {
+        if (ecma_get_value_type_field (*value_p) != ECMA_TYPE_NON_PTR_VALUE
+            && ecma_get_value_type_field (*value_p) != ECMA_TYPE_OBJECT)
+        {
+          ecma_free_value (*value_p);
+        }
+        *value_p = ecma_make_integer_value (0);
```
akiss77: Is a `return;` not missing from here?

jerry-core/ecma/base/ecma-helpers-value.c
akiss77: Shouldn't actually be better to invert this function and have `ecma_is_number_equal_to_negative_zero`? Both `ecma_make_number_value` and `ecma_value_assign_number` seem to be duplicating functionality like
```c
  if ((ecma_number_t) integer_value == ecma_num)
  {
    if (integer_value == 0)
    {
      if (ecma_is_number_equals_to_positive_zero (ecma_num))
      {
        // do something with integers and return
      }
    }
    else if (ECMA_INTEGER_NUMBER_MIN <= integer_value
             && integer_value <= ECMA_INTEGER_NUMBER_MAX)
    {
      // do the same with integers and return
    }
  }

  // fall back to floats and return
```
All the above could be reduced to (if I didn't miss something):
```c
  if (((ecma_number_t) integer_value == ecma_num)
      && ECMA_INTEGER_NUMBER_MIN <= integer_value
      && integer_value <= ECMA_INTEGER_NUMBER_MAX
      && !ecma_is_number_equal_to_negative_zero (ecma_num))
  {
    // do something with integers and return
  }

  // fall back to floats and return
```

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +618,173 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+ * Assign a new value to an ecma-value
+ *
+ * Note:
+ *      value previously stored in the property is freed
+ */
+void
+ecma_value_assign_value (ecma_value_t *value_p, /**< [in, out] ecma value */
+                         ecma_value_t ecma_value) /**< value to assign */
+{
+  JERRY_ASSERT (ECMA_TYPE_NON_PTR_VALUE == 0);
```
akiss77: What does this assert check? Both operands of the equality check seem to be constant.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +618,173 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+} /* ecma_value_assign_number */
+
+/**
+ * Assign an int32 value to an ecma-value
+ *
+ * Note:
+ *      value previously stored in the property is freed
+ */
+void
+ecma_value_assign_int32 (ecma_value_t *value_p, /**< [in, out] ecma value */
```
akiss77: This function is not used anywhere. We shouldn't introduce dead code.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -587,10 +586,7 @@ ecma_builtin_array_prototype_object_push (ecma_value_t this_arg, /**< this argum
                     ecma_builtin_array_prototype_helper_set_length (obj_p, n),
                     ret_value);
 
-    ecma_number_t *ret_num_p = ecma_alloc_number ();
-    *ret_num_p = n;
-
-    ret_value = ecma_make_number_value (ret_num_p);
+    ret_value = ecma_make_number_value (n);
```
akiss77: Something seems to be sub-optimal here. Some lines above, we have
`ecma_number_t n = ((ecma_number_t) ecma_number_to_uint32 (length_var));`
Is it intentional to convert an uint32 to float first and then most probably convert it back from float?
Or, is it because of this line
`ecma_string_t *n_str_p = ecma_new_ecma_string_from_number (n);`?
I'm not sure right now whether we have a quick(er) implementation for uint32-to-string conversion.

jerry-core/ecma/base/ecma-globals.h
zherczeg: In this case I would prefer the full value because of the name, i.e.:

    #define ECMA_DIRECT_INTEGER_VALUE (0u << ECMA_VALUE_SHIFT) | ECMA_TYPE_DIRECT

zherczeg: Choose ECMA_DIRECT_TYPE_INTEGER_VALUE finally.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: No! The check in ecma_make_float_number_value does not cover all cases, so it would trigger assertion failures for large integer numbers. The ecma_make_float_number_value is designed for non-integer compile time constants.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: Checking negative zero is difficult, because you need to do two comparisons:

uint32_t values[2]
memcpy(values, &double, 8);

negative zero check (little-endian):
  if (values[0] == 0 && values[0] == 0x80000000)

positive zero check:
 if ((values[0] | values[1]) == 0) // only one ARM instruction + branch: orrs low, high

Furthermore converting double to two uint32 values is costly (FPU->Integer core data copy) so I would only do it when absolutely necessary (the converted value is zero).

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
zherczeg: Array length range was very confusing for me. I had the impression that it can go up to UINT32_MAX+1, but the spec says otherwise:

http://www.ecma-international.org/ecma-262/5.1/#sec-15.4

Every Array object has a length property whose value is always a nonnegative integer less than 2^32.

This is not the only place where it is sub-optimal, I think we could improve a lot of code then.

dbatyai: That is true for Array objects, but not for other Objects. They don't have a limit on their length property, and when push() is called on a non-Array object, it may go above 2^32. There is a test for this in test262.

zherczeg: Unfortunately we cannot do anything in this case. The n can be bigger than UINT32_MAX, and the function must append these values as well

    var a = new Array(); a[4294967294]=7; try { a.push(5); }catch(e) {}; print(a[4294967295])

This prints 5, though the push throws an error when setting the length.

zherczeg: What I don't understand why this command is needed: Let n be ToUint32(lenVal).

zherczeg:
Patch updated

jerry-core/ecma/base/ecma-helpers-value.c
akiss77: OK, let's check for positive zeros then. The name of the function is still incorrect, however. (And I'll propose and alternative control structure for `ecma_make_number_value` and `ecma_value_assign_number`.)

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,156 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+      {
+        return ecma_make_integer_value (0);
+      }
+    }
+    else if (ECMA_INTEGER_NUMBER_MIN <= integer_value
+             && integer_value <= ECMA_INTEGER_NUMBER_MAX)
+    {
+      return ecma_make_integer_value (integer_value);
+    }
+  }
```
akiss77: To remove the need for code duplication, the above if could/should be written as
```c
  if (((ecma_number_t) integer_value == ecma_num)
      && (((integer_value == 0) && ecma_is_number_equals_to_positive_zero (ecma_num))
          || (ECMA_INTEGER_NUMBER_MIN <= integer_value && integer_value <= ECMA_INTEGER_NUMBER_MAX)))
  {
    // do something (only once)
  }
```
(The above is still within line length limits.)

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -442,22 +623,149 @@ ecma_copy_value_if_not_object (ecma_value_t value) /**< value description */
...
+    {
+      if (ecma_get_value_type_field (*value_p) != ECMA_TYPE_DIRECT
+          && ecma_get_value_type_field (*value_p) != ECMA_TYPE_OBJECT)
+      {
+        ecma_free_value (*value_p);
+      }
+      *value_p = ecma_make_integer_value (integer_value);
+      return;
+    }
+  }
```
akiss77: Could use the same approach as suggested for `ecma_make_number_value`.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: The above is incorrect, because the ECMA_INTEGER_NUMBER_MIN <= integer_value && integer_value <= ECMA_INTEGER_NUMBER_MAX part includes the integer_value == 0

That is, it is the same as:
```
if (((ecma_number_t) integer_value == ecma_num)
    && (ECMA_INTEGER_NUMBER_MIN <= integer_value && integer_value <= ECMA_INTEGER_NUMBER_MAX)))
```
probably a ?: is better

zherczeg:
Changes are added as a separate commit, but I plan to squash it. Please review.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,143 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+ * @return ecma-value
  */
-ecma_value_t __attr_const___
-ecma_make_number_value (const ecma_number_t *ecma_num_p) /**< number to reference in value */
+inline ecma_value_t __attr_const___ __attr_always_inline___
+ecma_make_integer_value (ecma_integer_value_t integer_value) /**< integer number to be encoded */
+{
+  JERRY_ASSERT (ECMA_IS_INTEGER_NUMBER (integer_value));
+
+  return (ecma_value_t) ((integer_value << ECMA_DIRECT_SHIFT) | ECMA_TYPE_DIRECT);
```
akiss77: This should rather end with ` | ECMA_DIRECT_TYPE_INTEGER_VALUE`

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,143 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+  return ecma_create_float_number (ecma_number);
+} /* ecma_make_float_number_value */
+
+/**
+ * Checks whether the passed number is +0.0
+ *
+ * @return true, if it is +0.0, false otherwise
+ */
+static inline bool __attr_const___ __attr_always_inline___
+ecma_is_number_equals_to_positive_zero (ecma_number_t ecma_number) /**< number */
```
akiss77: Do we intend to leave this english grammar error in the code base?

zherczeg: What is the fix?

jerry-core/vm/vm.c
```diff
@@ -1240,7 +1238,80 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+            }
+
+            if (int_increase != 0)
+            {
+              /* Post operators require the unmodifed number value. */
+              if (base & 0x4)
+              {
+                if (opcode_data & VM_OC_PUT_STACK)
+                {
+                  if (base & 0x1)
```
akiss77: I'm a bit lost with the bit masks. Of course, I know what `& 0x1, 0x2, 0x4` achieves, but what do they represent? Are there no mnemonics (macros) for the fields they test?

zherczeg: This code is similar to the code below just specialized for integers. Perhaps we could rework this in some other patch.

jerry-core/ecma/base/ecma-helpers-value.c
akiss77: I've mentioned this at least twice in previous review rounds (not sure why they don't show up anymore). That's why I asked whether it's been intentionally negliged. Properly, this would be `ecma_is_number_equal_to_positive_zero` or `ecma_number_equals_to_positive_zero`.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -275,21 +300,143 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
...
+ *
+ * This function is intended to create floating point constants when it is
+ * known that ecma_number is not an integer number, e.g. 0.5, NaN, or -infinity.
+ * Calling this function is faster than ecma_make_number_value since no
+ * checks are performed on the input apart from a debug assert.
+ *
+ * @return ecma-value
+ */
+inline ecma_value_t __attr_always_inline___
+ecma_make_float_number_value (ecma_number_t ecma_number) /**< value of the float number */
```
akiss77: I'm still not happy with this function. This is only called from different compilation units and only in the form of
```c
ecma_make_float_number_value (ecma_number_make_nan ())
```
This signals that this function is unnecessary and should be replaced by something like an `ecma_make_float_number_nan`. But I'm not sure what makes it different from `ecma_number_make_nan` then?

jerry-core/vm/vm.c
akiss77: Please, make an issue for this. This is something we should definitely address. Hard-coded constants (flags, bit masks, etc.) should be removed from the code base.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: Ok. Unfortunately comments disappear after code changes. This one will be.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: Currently nothing. But I would prefer a more general form, since we might need it in the future. At that point we might not recognize that ecma_number_make_nan is redundant.

jerry-core/vm/vm.c
zherczeg: Done.

zherczeg:
Since only NaNs are created now, I made  a ecma_make_nan_value function. If we ever need other constants we could continue this discussion.

akiss77:
LGTM

Issue #668, opened 2015-10-14
does get operation really need non constant object?


idkiller:
jerry_api_get_object_native_handle
jerry_api_get_array_index_value
jerry_api_get_object_field_value
jerry_api_get_object_field_value_sz

these functions are get operation but functions need non constant object pointer.
I think constant pointer is better if it can.

for example:

```cpp
  auto cb = [] (const jerry_api_object_t *function_obj_p,
                const jerry_api_value_t *this_p,
                jerry_api_value_t *ret_val_p,
                const jerry_api_value_t args_p[],
                const jerry_api_length_t args_count) -> bool 
  {                                                                             
    uintptr_t p;
    bool err = !jerry_api_get_object_native_handle(const_cast<jerry_api_object_t*>(function_obj_p), &p);
    // use p
  };
  jerry_api_object_t *jfunc = jerry_api_create_external_function(cb);
```
In many case, user need native handle when call there external function,
but object pointer from external function pointer is constant pointer. so user must casting it to non constant object.

What do you think about it?

PR #1047, opened 2016-05-10
Use const object parameters in getter API functions

LaszloLango:
Related issue: #668

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -373,7 +373,7 @@ ecma_op_function_try_lazy_instantiate_property (ecma_object_t *obj_p, /**< the f
  *         NULL (i.e. ecma-undefined) - otherwise.
  */
 ecma_property_t *
-ecma_op_function_object_get_own_property (ecma_object_t *obj_p, /**< the function object */
+ecma_op_function_object_get_own_property (const ecma_object_t *obj_p, /**< the function object */
                                           ecma_string_t *property_name_p) /**< property name */
 {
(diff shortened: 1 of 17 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -392,7 +392,7 @@ ecma_op_function_object_get_own_property (ecma_object_t *obj_p, /**< the functio
   }
   else if (!ecma_get_object_is_builtin (obj_p))
   {
-    prop_p = ecma_op_function_try_lazy_instantiate_property (obj_p, property_name_p);
+    prop_p = ecma_op_function_try_lazy_instantiate_property ((ecma_object_t *) obj_p, property_name_p);
```
LaszloLango: IMHO, this cast makes the patch ugly and non-secure. Any suggestion to avoid this?

LaszloLango:
This is a WIP patch. There is a few const to non-const cast in the code that shouldn't be there.

jerry-core/ecma/operations/ecma-function-object.c
zherczeg: You cannot avoid this, since reading a property (length for example) might also trigger its construction. Hence the object is extended, and an out-of-memory error can stop the engine.

zherczeg:
What is the plan with this patch? Shall we move the property hash map creation into some other place? Or make it error tolerant?

LaszloLango:
@zherczeg, good question. :) I think the API change is still needed and the referred issue is valid and open in a long time, so we should do something with it. Moving the property hash map creation sounds good to me, but currently no idea how to do it. Any specific idea?

zherczeg:
Perhaps we can move it when a property is assigned. But that does not easy either.

LaszloLango:
Won't fix.

PR #1050, opened 2016-05-11
Update API reference

LaszloLango:
* Fixed code style issues
 * Updated the documentation of 'jerry_parse' and 'jerry_run'
 * Fixed broken anchor formats
 * Fixed 'jerry_api_value_t' usage
 * Updated fuction parameters

### DIFF
--- a/docs/API-REFERENCE.md
+++ b/docs/API-REFERENCE.md
@@ -7,7 +7,7 @@ The simplest way to run JavaScript.
 
 ```c
 jerry_completion_code_t
-jerry_run_simple (const char * script_source,
+jerry_run_simple (const jerry_api_char_t *script_source,
                   size_t script_source_size,
                   jerry_flag_t flags);
 ```
@@ -20,18 +20,18 @@ jerry_run_simple (const char * script_source,
 
 ```c
 {
-  const char * script = "print ('Hello, World!');";
+  const jerry_api_char_t *script = "print ('Hello, World!');";
 
-  jerry_run_simple (script, strlen (script), JERRY_FLAG_EMPTY);
+  jerry_run_simple (script, strlen ((const char *) script), JERRY_FLAG_EMPTY);
 }
 ```
 
 **See also**
 
-- [jerry_init](#jerryinit)
-- [jerry_cleanup](#jerrycleanup)
-- [jerry_parse](#jerryparse)
-- [jerry_run](#jerryrun)
+- [jerry_init](#jerry_init)
+- [jerry_cleanup](#jerry_cleanup)
+- [jerry_parse](#jerry_parse)
+- [jerry_run](#jerry_run)
 
 # jerry_init
 
@@ -48,10 +48,11 @@ jerry_init (jerry_flag_t flags);
 
 `flags` - combination of various engine configuration flags:
 
-- `JERRY_FLAG_MEM_STATS` - dump memory statistics;
-- `JERRY_FLAG_ENABLE_LOG` - enable logging;
+- `JERRY_FLAG_EMPTY` - no flags, just initialize in default configuration;
 - `JERRY_FLAG_SHOW_OPCODES` - print compiled byte-code;
-- `JERRY_FLAG_EMPTY` - no flags, just initialize in default configuration.
+- `JERRY_FLAG_MEM_STATS` - dump memory statistics;
+- `JERRY_FLAG_MEM_STATS_SEPARATE` - dump memory statistics and reset peak values after parse;
+- `JERRY_FLAG_ENABLE_LOG` - enable logging;
 
 **Example**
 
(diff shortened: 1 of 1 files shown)


docs/API-REFERENCE.md
```diff
@@ -182,28 +217,25 @@ typedef struct jerry_api_value_t
 
   union
   {
-    bool v_bool;
+    bool.u.v_bool;
```
galpeter: Is this really correct?

docs/API-REFERENCE.md
```diff
@@ -1040,38 +1084,38 @@ Create an external function object.
...
+         const jerry_api_value_t *this_p,
+         jerry_api_value_t *ret_val_p,
          const jerry_api_value_t args_p[],
          const uint16_t args_cnt)
 {
   printf ("native handler called!\n");
 
   ret_val_p->type = JERRY_API_DATA_TYPE_BOOLEAN;
-  ret_val_p->v_bool = true;
+  ret_val_p-.u.v_bool = true;
```
galpeter: Shouldn't this be: `ret_val_p->u.v_bool = ...`?

docs/API-REFERENCE.md
```diff
@@ -1271,31 +1315,31 @@ jerry_api_create_error (jerry_api_error_t error_type,
...
          const uint16_t args_cnt)
 {
-  jerry_api_object_t * error_p = jerry_api_create_error (JERRY_API_ERROR_TYPE,
+  jerry_api_object_t *error_p = jerry_api_create_error (JERRY_API_ERROR_TYPE,
                                                          (jerry_api_char_t * ) "error");
 
   jerry_api_acquire_object (error_p);
   ret_val_p->type = JERRY_API_DATA_TYPE_OBJECT;
-  ret_val_p->v_object = error_p;
+  ret_val_p-.u.v_object = error_p;
```
galpeter: Shouldn't this be: `ret_val_p->u.v_object = ...`?

docs/API-REFERENCE.md
LaszloLango: Of course not. Good catch.

LaszloLango:
@galpeter, I've updated the PR.

docs/API-REFERENCE.md
```diff
@@ -100,29 +101,37 @@ so `jerry_parse` could be invoked only once between `jerry_init` and `jerry_clea
...
 ```c
 bool
-jerry_parse (const char* source_p, size_t source_size);
+jerry_parse (const jerry_api_char_t *source_p,
+             size_t source_size,
+             jerry_api_object_t **error_obj_p);
 ```
 - `source_p` - string, containing source code to parse;
 - `source_size` - size of the string, in bytes.
+- `error_obj_p` - error object (output parameter)
```
zherczeg: Can be null?

docs/API-REFERENCE.md
```diff
@@ -135,29 +144,52 @@ The code should be previously registered through `jerry_parse`.
 
 ```c
 jerry_completion_code_t
-jerry_run (void);
+jerry_run (jerry_api_value_t *error_value_p);
 ```
-
-- returned value - completion code that indicates whether run performed successfully (`JERRY_COMPLETION_CODE_OK`), or an unhandled JavaScript exception occurred (`JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION`).
+- `error_value_p` - error value (output parameter)
+- returned value - completion code that indicates whether run performed successfully
```
zherczeg: completion value

docs/API-REFERENCE.md
```diff
@@ -135,29 +144,52 @@ The code should be previously registered through `jerry_parse`.
...
+  if (!jerry_parse (script, script_size, &error_object_p))
+  {
+    /* Error object must be freed, if parsing failed */
+    jerry_api_release_object (error_object_p);
+  }
+  else
+  {
+    /* Execute Global scope code */
+    jerry_api_value_t error_value = jerry_api_create_void_value ();
+    jerry_completion_code_t return_code = jerry_run (&error_value);
```
zherczeg: Do you need to initialize the return value?

LaszloLango: It is a local variable, so it is better to be initialized. If we use it normally (return_code != JERRY_COMPLETION_CODE_OK) it is unnecessary. I think it is better to initialize always.

docs/API-REFERENCE.md
LaszloLango: No

docs/API-REFERENCE.md
zherczeg: Yes, it is a good prectice, but the question still stands: is it necessary?

LaszloLango: In this example it is not necessary.

akiss77:
LGTM

docs/API-REFERENCE.md
zherczeg: I think we should be careful when writing examples because our exaples can be misleading. So please add a comment that this initialization is not mandatory.

zherczeg:
LGTM after the comment fix.

PR #1058, opened 2016-05-13
Argument literal names were not stored in the snapshot

zherczeg:
Argument literal names were not stored in the snapshot, because the names were replaced by empty strings before that. The snapshot form of tests/jerry/arguments.js fails because of this issue.

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -1991,10 +1991,7 @@ jerry_snapshot_set_offsets (uint8_t *buffer_p, /**< buffer */
         const_literal_end = args_p->const_literal_end;
       }
 
-      for (uint32_t i = 0; i < register_end; i++)
-      {
-        literal_start_p[i] = MEM_CP_NULL;
-      }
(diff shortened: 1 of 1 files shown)


LaszloLango:
LGTM

jerry-core/jerry.c
```diff
@@ -1991,10 +1992,7 @@ jerry_snapshot_set_offsets (uint8_t *buffer_p, /**< buffer */
         const_literal_end = args_p->const_literal_end;
       }
 
-      for (uint32_t i = 0; i < register_end; i++)
-      {
-        literal_start_p[i] = MEM_CP_NULL;
-      }
+      register_clear_start = 0;
```
akiss77: As always, I'd suggest moving declaration as close to initialization as possible. Here, it is possible to do them in one step. (C99 allows declaration later than at the beginning of a block, and we use it often.)

akiss77:
After the `uint32_t register_clear_start  = 0;` merge, LGTM.

PR #1059, opened 2016-05-13
Run JS regression tests in snapshot mode too

LaszloLango:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -304,6 +304,11 @@ test-js.$(1).$(2): build.$$(NATIVE_SYSTEM)
 	$$(Q) $$(call SHLOG,cd $$(OUT_DIR)/$(1)/check/$(2) && $$(ROOT_DIR)/tools/runners/run-test-suite.sh \
           $$(OUT_DIR)/$(1)/jerry \
           $(3),$$(OUT_DIR)/$(1)/check/$(2)/test.log,Testing)
+test-js.$(1).$(2).snapshot: build.$$(NATIVE_SYSTEM)
+	$$(Q) mkdir -p $$(OUT_DIR)/$(1)/check/$(2)
+	$$(Q) $$(call SHLOG,cd $$(OUT_DIR)/$(1)/check/$(2) && $$(ROOT_DIR)/tools/runners/run-test-suite.sh \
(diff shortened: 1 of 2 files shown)


LaszloLango:
`./tests/jerry/arguments.js` fails on the current master in snapshot mode. #1058 fixes this issue.

zherczeg:
I hope after a rebase checks will pass.

akiss77:
I'd suggest some changes. As proposed now, there is a bit too much code duplication and hard-coding. I'd leave the choice, whether to run tests in interpreted mode or from snapshot, to the user.

Notes:
* Before copying `$@` to `ENGINE_ARGS`, you coud check whether there are args left and whether the first is, say, `--snapshot` (or something like that), and if so, set a variable indicating that user requested a snapshot run (and remove that arg from `$@`, of course). This way, you could use the runner script like `tools/runners/run-test-suite.sh ENGINE TESTS [--snapshot] ENGINE_ARGS...`.
* Then, in the test loop body, you could just choose between the two execution modes based on the variable set during the handling of command line args:
```bash
    ( ulimit -t $TIMEOUT; $ENGINE $ENGINE_ARGS $full_test &> $ENGINE_TEMP )
    status_code=$?
``` 
```bash
    SNAPSHOT_TEMP=`mktemp snapshot-out.XXXXXXXXXX`

    ( ulimit -t $TIMEOUT; $ENGINE $ENGINE_ARGS --save-snapshot-for-global $SNAPSHOT_TEMP $full_test &> $ENGINE_TEMP )
     status_code=$?
 
    if [ $status_code -eq 0 ]
    then
        ( ulimit -t $TIMEOUT; $ENGINE $ENGINE_ARGS --exec-snapshot $SNAPSHOT_TEMP &> $ENGINE_TEMP )
        status_code=$?
    fi

    rm -f $SNAPSHOT_TEMP
```
*  Finally, of course, the make system can be adapted to call the runner script both ways.

LaszloLango:
@akiss77, I've updated the PR based on your suggestions. Please check.

akiss77:
LGTM

zherczeg:
LGTM

PR #1061, opened 2016-05-17
Integer acceleration for basic arithmetic and comparison.

zherczeg:


jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -140,6 +141,7 @@ extern ecma_value_t ecma_make_uint32_value (uint32_t);
 extern ecma_value_t ecma_make_string_value (const ecma_string_t *);
 extern ecma_value_t ecma_make_object_value (const ecma_object_t *);
 extern ecma_value_t ecma_make_error_value (ecma_value_t);
+extern ecma_integer_value_t ecma_get_integer_from_value (ecma_value_t);
```
akiss77: Misplaced. Move declaration right above `ecma_get_number_from_value`.

jerry-core/vm/opcodes-ecma-arithmetics.c
```diff
@@ -45,6 +45,69 @@ do_number_arithmetic (number_arithmetic_op op, /**< number arithmetic operation
...
+          ecma_integer_value_t result = left_integer % right_integer;
+
+          if (result != 0 || left_integer > 0)
+          {
+            return ecma_make_integer_value (result);
+          }
+        }
+        break;
+      }
+    }
```
akiss77: Strange not to have a default unreachable assert at the end of this switch. Are there any more arithmetic operations which we intentionally don't want to handle?

jerry-core/vm/opcodes-ecma-bitwise.c
```diff
@@ -43,6 +43,55 @@ do_number_bitwise_logic (number_bitwise_logic_op op, /**< number bitwise logic o
...
+      {
+        uint32_t left_uint32 = (uint32_t) ecma_get_integer_from_value (left_value);
+        ecma_integer_value_t right_integer = ecma_get_integer_from_value (right_value);
+        return ecma_make_uint32_value (left_uint32 >> (right_integer & 0x1f));
+      }
+      case NUMBER_BITWISE_NOT:
+      {
+        return (~right_value) & (ecma_value_t) (~((1 << ECMA_DIRECT_SHIFT) - 1));
+      }
+    }
```
akiss77: Same here for bitwise operations.

jerry-core/vm/opcodes-ecma-arithmetics.c
zherczeg: No. NUMBER_ARITHMETIC_DIVISION is the only unhandled case. I would prefer a compiler error if new cases are added.

jerry-core/vm/opcodes-ecma-bitwise.c
zherczeg: No unhandled bitwise operator. I would prefer a compiler error if new cases are added.

zherczeg:
10% speedup:

```
                     3d-cube.js |   2.168s ->   2.084s :  +3.874%  (+-0.661%) : [+]
                 3d-raytrace.js |   2.298s ->   2.248s :  +2.176%  (+-1.373%) : [+]
         access-binary-trees.js |   1.158s ->   1.124s :  +2.936%  (+-1.243%) : [+]
             access-fannkuch.js |   6.122s ->   5.548s :  +9.376%  (+-0.653%) : [+]
                access-nbody.js |   2.472s ->   2.416s :  +2.265%  (+-0.584%) : [+]
    bitops-3bit-bits-in-byte.js |   1.396s ->   1.002s : +28.224%  (+-0.878%) : [+]
         bitops-bits-in-byte.js |   1.914s ->   1.360s : +28.945%  (+-0.419%) : [+]
          bitops-bitwise-and.js |   3.110s ->   2.734s : +12.090%  (+-1.239%) : [+]
          bitops-nsieve-bits.js |   4.504s ->   3.750s : +16.741%  (+-0.577%) : [+]
       controlflow-recursive.js |   0.780s ->   0.660s : +15.385%  (+-0.000%) : [+]
                  crypto-aes.js |   2.764s ->   2.548s :  +7.815%  (+-1.323%) : [+]
                  crypto-md5.js |   1.254s ->   1.132s :  +9.729%  (+-1.095%) : [+]
                 crypto-sha1.js |   1.194s ->   1.062s : +11.055%  (+-1.140%) : [+]
           date-format-tofte.js |   1.716s ->   1.666s :  +2.914%  (+-1.232%) : [+]
           date-format-xparb.js |   0.720s ->   0.698s :  +3.056%  (+-1.279%) : [+]
                 math-cordic.js |   2.656s ->   2.368s : +10.843%  (+-0.513%) : [+]
           math-partial-sums.js |   1.490s ->   1.404s :  +5.772%  (+-1.236%) : [+]
          math-spectral-norm.js |   1.262s ->   1.156s :  +8.399%  (+-1.116%) : [+]
               string-base64.js |   4.744s ->   4.582s :  +3.415%  (+-0.736%) : [+]
                string-fasta.js |   2.828s ->   2.656s :  +6.082%  (+-0.503%) : [+]
                Geometric mean: |                   Speed up: 9.905% (+-0.234%) : [+]

Binaries (bytes)
master: 171400
branch: 171856
```

jerry-core/vm/opcodes-ecma-arithmetics.c
akiss77: I recall having this debate earlier. Ok then to rely on warnings/errors at compile time.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -219,6 +219,22 @@ ecma_is_value_integer_number (ecma_value_t value) /**< ecma value */
...
+ *         false - otherwise.
+ */
+inline bool __attr_pure___ __attr_always_inline___
+ecma_are_values_integer_numbers (ecma_value_t first_value, /**< first ecma value */
+                                 ecma_value_t second_value) /**< second ecma value */
+{
+  JERRY_STATIC_ASSERT (ECMA_DIRECT_TYPE_INTEGER_VALUE == 0,
+                       ecma_direct_type_integer_value_must_be_zero);
+
+  return ((first_value | second_value) & ECMA_DIRECT_TYPE_MASK) == ECMA_DIRECT_TYPE_INTEGER_VALUE;
```
LaszloLango: Are you sure this is correct? Why not `(first_value & second_value)`?

jerry-core/vm/opcodes-ecma-arithmetics.c
```diff
@@ -45,6 +45,69 @@ do_number_arithmetic (number_arithmetic_op op, /**< number arithmetic operation
...
+
+  if (ecma_are_values_integer_numbers (left_value, right_value))
+  {
+    switch (op)
+    {
+      case NUMBER_ARITHMETIC_ADDITION:
+      {
+        ecma_integer_value_t left_integer = ecma_get_integer_from_value (left_value);
+        ecma_integer_value_t right_integer = ecma_get_integer_from_value (right_value);
+        return ecma_make_int32_value ((int32_t) (left_integer + right_integer));
```
LaszloLango: The mix of `int32_t `and `ecma_integer_value_t `(which is also `int32_t`) is confusing.

jerry-core/vm/opcodes-ecma-bitwise.c
```diff
@@ -43,6 +43,55 @@ do_number_bitwise_logic (number_bitwise_logic_op op, /**< number bitwise logic o
...
+      {
+        return left_value & right_value;
+      }
+      case NUMBER_BITWISE_LOGIC_OR:
+      {
+        return left_value | right_value;
+      }
+      case NUMBER_BITWISE_LOGIC_XOR:
+      {
+        return (left_value ^ right_value) & (ecma_value_t) (~((1 << ECMA_DIRECT_SHIFT) - 1));
```
akiss77: Here, you also assume that `ECMA_TYPE_DIRECT == 0`, if I'm right. If so, it should be asserted.

jerry-core/vm/opcodes-ecma-arithmetics.c
```diff
@@ -45,6 +45,69 @@ do_number_arithmetic (number_arithmetic_op op, /**< number arithmetic operation
...
+      }
+      case NUMBER_ARITHMETIC_REMAINDER:
+      {
+        ecma_integer_value_t left_integer = ecma_get_integer_from_value (left_value);
+        ecma_integer_value_t right_integer = ecma_get_integer_from_value (right_value);
+        if (right_integer != 0)
+        {
+          ecma_integer_value_t result = left_integer % right_integer;
+
+          if (result != 0 || left_integer > 0)
```
akiss77: My feeling is that the second half of the predicte could be `left_integer >= 0`. Right now, `0 % 3` would not make use of this optimization, as both `result` and `left_integer` will be 0. I guess `> 0` was used not to have problems with negative zero, but if I'm not mistaken, if left value were negative zero, it couldn't have been encoded as integer, so we wouldn't run this code path.

jerry-core/vm/opcodes-ecma-bitwise.c
```diff
@@ -43,6 +43,55 @@ do_number_bitwise_logic (number_bitwise_logic_op op, /**< number bitwise logic o
...
+      {
+        ecma_integer_value_t left_integer = ecma_get_integer_from_value (left_value);
+        ecma_integer_value_t right_integer = ecma_get_integer_from_value (right_value);
+        return ecma_make_int32_value ((int32_t) (left_integer << (right_integer & 0x1f)));
+      }
+      case NUMBER_BITWISE_SHIFT_RIGHT:
+      {
+        ecma_integer_value_t left_integer = ecma_get_integer_from_value (left_value);
+        ecma_integer_value_t right_integer = ecma_get_integer_from_value (right_value);
+        return ecma_make_int32_value ((int32_t) (left_integer >> (right_integer & 0x1f)));
```
akiss77: Is there a way for the result not to fit in `ecma_integer_value_t`? (If left value was OK, then by shifting right, we can get smaller values only.) I guess we could call `ecma_make_integer_value` directly. (Same holds for `NUMBER_BITWISE_SHIFT_URIGHT` below.)

jerry-core/vm/opcodes-ecma-equality.c
```diff
@@ -44,6 +44,18 @@ ecma_value_t
...
+  JERRY_ASSERT (!ecma_is_value_error (left_value)
+                && !ecma_is_value_error (right_value));
+
+  if (ecma_are_values_integer_numbers (left_value, right_value))
+  {
+    if (left_value == right_value)
+    {
+      return ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE);
+    }
+    return ecma_make_simple_value (ECMA_SIMPLE_VALUE_FALSE);
```
akiss77: Couldn't we simply have
```c
ecma_make_simple_value (left_value == right_value ? ECMA_SIMPLE_VALUE_TRUE
                                                  : ECMA_SIMPLE_VALUE_FALSE);
```
? (It's still within line length limit, and the pattern is used all over the code base.)

LaszloLango: +1

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: Because any bit is set (logical or), both values cannot be number. If I would use & and the first_value is number, the result is always true regardless of the second number.

jerry-core/vm/opcodes-ecma-arithmetics.c
zherczeg: It is intentional. See the assert above.

jerry-core/vm/opcodes-ecma-bitwise.c
zherczeg: It is checked by ecma_are_values_integer_numbers, but a compile assert never hurts.

zherczeg: There is an assert for ECMA_DIRECT_TYPE_INTEGER_VALUE == 0 above, so I think this is not necessary.

jerry-core/vm/opcodes-ecma-bitwise.c
zherczeg: Good idea. However unsigned right shift cannot use this optimization, because negative number >>> 0 cannot be represented as ecma_integer_value_t.

jerry-core/vm/opcodes-ecma-equality.c
zherczeg: But I would need to use it everywhere in this file for consistency, and (ecma_integer_value_t) left_value <= (ecma_integer_value_t) right_value is too long for checks.

jerry-core/vm/opcodes-ecma-bitwise.c
akiss77: Indeed. `ECMA_DIRECT_TYPE_INTEGER_VALUE` incorporates `ECMA_TYPE_DIRECT`. I've missed that. OK then.

zherczeg:
Patch updated.

jerry-core/ecma/base/ecma-helpers-value.c
LaszloLango: If 'ECMA_DIRECT_TYPE_INTEGER_VALUE' is set, then it is not an integer_number? I'm still confused.
What `(ecma_value & ECMA_DIRECT_TYPE_MASK) == ECMA_DIRECT_TYPE_INTEGER_VALUE` means?

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -164,6 +164,15 @@ typedef int32_t ecma_integer_value_t;
   (ECMA_INTEGER_NUMBER_MIN <= (num) && (num) <= ECMA_INTEGER_NUMBER_MAX)
 
 /**
+ * Maximum integer number for multpily two ecma values
+ */
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
+#define ECMA_INTEGER_MULTIPLY_MAX 0xb50
```
akiss77: I just realize this conditional here. Why is this dependent on float size? The only constraint for `ECMA_INTEGER_MULTIPLY_MAX` that I see in the code is that it should be (less than) the square root of `ECMA_INTEGER_NUMBER_MAX` (rounded down). This should be the same value both for float32 and float64, IMHO.
PS: I also suggest to adapt de description. Something like: "Maximum integer number, which if squared, still fits in ecma_integer_value_t"

jerry-core/ecma/base/ecma-helpers-value.c
akiss77: zherczeg is right here. `ECMA_DIRECT_TYPE_MASK` is 0b1011, while `ECMA_DIRECT_TYPE_INTEGER_VALUE` is 0b0000. If any of `first_value` or `second_value` is not of a direct type or not an integer value, then by or-ing them together, some of the bits in the mask will be 1. You can get all zeros only if both values are of a direct type and that direct type is integer.

jerry-core/vm/opcodes-ecma-equality.c
akiss77: The following still works:
```c
	return ecma_make_simple_value ((ecma_integer_value_t) left_value <= (ecma_integer_value_t) right_value 
	                               ? ECMA_SIMPLE_VALUE_TRUE
								   : ECMA_SIMPLE_VALUE_FALSE);
```

zherczeg: I don't want to introduce new style.

jerry-core/ecma/base/ecma-globals.h
zherczeg: The integer max is also different for float32 and 64 so the square root of max yield a different number for these modes.

jerry-core/ecma/base/ecma-helpers-value.c
LaszloLango: I got it, thanks. The value of `ECMA_DIRECT_TYPE_INTEGER_VALUE` is a bit odd to me, but the code is correct. :)

LaszloLango:
LGTM

akiss77:
LGTM

zherczeg:
Thank you for the reviews

PR #1064, opened 2016-05-17
Shrink test logs by not printing full paths

akiss77:
Both JS test suite and unit test logs can grow large as Makefile
invokes run-test-suite.sh and run-unittests.sh with absolute paths
to engines and test directories, which get then printed quite
often. This patch adds code to the runner scripts to determine the
longest directory path common to the current working directory, the
invoked script, the test directory, and the engine (in case of JS
tests). Then, when a path is to be printed, this common path
component is skipped.

### DIFF
--- a/tools/runners/run-test-suite.sh
+++ b/tools/runners/run-test-suite.sh
@@ -59,6 +59,25 @@ fi
 
 rm -f $TEST_FAILED $TEST_PASSED
 
+ROOT_DIR=""
+CURRENT_DIR=`pwd`
+PATH_STEP=2
+while true
+do
+    TMP_ROOT_DIR=`(echo "$CURRENT_DIR"; echo "$0"; echo "$ENGINE"; echo "$TESTS") | cut -f1-$PATH_STEP -d/ | uniq -d`
+    if [ -z "$TMP_ROOT_DIR" ]
+    then
+        break
+    else
+        ROOT_DIR="$TMP_ROOT_DIR"
+    fi
+    PATH_STEP=$((PATH_STEP+1))
+done
+if [ -n "$ROOT_DIR" ]
+then
+    ROOT_DIR="$ROOT_DIR/"
+fi
+
 tested=1
 failed=0
 passed=0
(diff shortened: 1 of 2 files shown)


akiss77:
This may help #1059 , where Travis CI reported errors because log length has exceeded the limit of 4 MB.

tools/runners/run-test-suite.sh
```diff
@@ -59,6 +59,25 @@ fi
 
 rm -f $TEST_FAILED $TEST_PASSED
 
+ROOT_DIR=""
+CURRENT_DIR=`pwd`
+PATH_STEP=2
+while true
+do
+    TMP_ROOT_DIR=`(echo $CURRENT_DIR; echo $0; echo $ENGINE; echo $TESTS) | cut -f1-$PATH_STEP -d/ | uniq -d`
```
LaszloLango: This might not work with symbolic links

akiss77: Please, explain. Which part, why, and do you have a fix? Did you run into a problem with it? (In the standard setup, i.e., when run by the makefile only, such a situation should not arise.)
PS: Please note that if this loop cannot find a common prefix then it will simply not remove anything from the printed paths. That's all to its effects. It does not change the actual execution at all.

LaszloLango: I haven't checked it yet. I'll check it later. I use symbolic links in my working paths. My prediction is that we need `pwd -P`.

LaszloLango: Works as expected :) My only problem is the missing quotes.
```bash
TMP_ROOT_DIR=`(echo "$CURRENT_DIR"; echo "$0"; echo "$ENGINE"; echo "$TESTS") | cut -f1-${PATH_STEP} -d/ | uniq -d`
```

akiss77:
Quotes added. Updated.

LaszloLango:
Thanks. LGTM

zherczeg:
LGTM

PR #1070, opened 2016-05-18
Don't concatenate multiple JS scripts in jerry

akiss77:
Jerry (the command line tool) has been supporting the execution of
multiple script files for long. However, until now, it simply
concatenated all sources into a single source buffer and
parsed/executed them as one unit. Other JS execution tools (e.g.,
jsc, v8) load and execute separate files as separate units -- but
still in the same execution environment. The most significant
effect of this approach is that the `"use strict;"` directive (or
the absence of it) at the beginning of each JS script file takes
effect as expected (i.e., as if the script was executed alone).
Contrarily, the concatenation-based approach forces the strictness
of the first script on all the rest (i.e., if the first script
starts with `"use strict";` the rest is also executed in a strict
environment even if they did not contain the directive, and vice
versa).

This patch makes the jerry command line tool to load/parse/run one
unit at a time.

Side effects:
- As there is no need for separate file read routines that load one
  file (a snapshot) or concat multiple (JS sources) anymore, those
  routines got merged.
- Both previous read routines used multiple stdio functions
  (`fseek`, `ftell`, and `rewind`). This has been simplified to
  rely on `fread` only to find out the length of the input.
- This simplification made the above mentioned functions
  superfluous in jerry-libc.
- As some error messages had to be touched in this patch, several
  more have been beautified to make them more consistent.
- One small change was needed in `jerry_parse` in jerry-core to
  allow subsequent parsing of multiple sources (without that, an
  assertion was triggered).

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -1698,6 +1698,8 @@ jerry_parse (const jerry_api_char_t *source_p, /**< script source */
 {
   jerry_assert_api_available ();
 
+  vm_finalize ();
+
   int is_show_instructions = ((jerry_flags & JERRY_FLAG_SHOW_OPCODES) != 0);
 
   parser_set_show_instrs (is_show_instructions);
--- a/main-unix.c
+++ b/main-unix.c
@@ -1,4 +1,5 @@
 /* Copyright 2014-2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


akiss77:
A simple example where concatenation causes problems:
nonstrict.js:
```js
a=1;
```
strict.js:
```js
"use strict";
var b=2;
```
print.js:
```js
print("a, b = " + a + ", " + b);
```
When this is executed by `jerry nonstrict.js strict.js print.js`, all is fine, however, `jerry strict.js nonstrict.js print.js` throws a reference error with the concat-based approach.

In the above example, enabling the strict mode for all files causes the problem. It is possible to construct examples when the loss of the strict mode would cause trouble,

jerry-core/jerry.c
```diff
@@ -1698,6 +1698,8 @@ jerry_parse (const jerry_api_char_t *source_p, /**< script source */
 {
   jerry_assert_api_available ();
 
+  vm_finalize ();
```
LaszloLango: finialize but not initialize? This call is a bit odd.

zherczeg:
LGTM if runs well with test_262 es5 branch

jerry-core/jerry.c
akiss77: That's actually "finalize if it has already been initialized"; the check happens within `vm_finalize`. If this call is omitted, `jerry_parse` cannot be invoked multiple times.

LaszloLango: My first thought is that using eval for each file would be a better way.

LaszloLango: The second is that what about `jerry_parse_and_save_snapshot`

akiss77: I tried to trace down similarities/differences between global and eval run, and it turned out that they cannot be forced to act the same. The standard (ECMA-262 5.1ed) has its relevant parts in section 10.4. Even "if the eval code is not being evaluated by a direct call" and we "initialise the execution context as if it was a global execution context", when performing "Declaration Binding Instantiation", we come to a difference that cannot be worked around: "If code is eval code, then let configurableBindings be true else let configurableBindings be false". Our implementation follows these instructions (fortunately).

Long story cut short, eval'd code will always be executed somewhat differently from globally executed code. Thus, it's no good for us.

LaszloLango: I see. Thanks for the detailed explanation.

akiss77: I didn't touch `jerry_parse_and_save_snapshot`. I didn't want to introduce too big changes to the embedder (main-unix.c) -- it might have required command line interface change to enable saving multiple snapshots of multiple input script files. That's why if snapshot saving is enabled, only a single input JS script is allowed, see below:
```c
    if (files_counter != 1)
    {
      jerry_port_errormsg ("Error: --save-snapshot argument works with exactly one script\n");
    }
```
Thus, as invokation count of `jerry_parse_and_save_snapshot` didn't increase, I didn't examine/touch that function either.

LaszloLango: Sorry, I missed that condition.

main-unix.c
```diff
@@ -511,6 +405,11 @@ main (int argc,
           ret_code = jerry_run (&err_value);
         }
       }
+
+      if (ret_code != JERRY_COMPLETION_CODE_OK)
+      {
+        break;
```
LaszloLango: I think err_value should be released at this point.

akiss77: Will not the code starting at line 594 take care of it?

LaszloLango: Right, breaking the for loop is enough.

LaszloLango:
LGTM

akiss77:
@zherczeg, Jerry master and this PR give the exact same results on test262

zherczeg:
Excellent. LGTM

PR #1087, opened 2016-05-20
Optimize ecma_op_general_object_define_own_property

zherczeg:
The return value of ecma_op_general_object_define_own_property is
always true if the currently updated property is configurable.
This allows skipping sever checks in the current code.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects-general.c
+++ b/jerry-core/ecma/operations/ecma-objects-general.c
@@ -572,6 +572,11 @@ ecma_op_general_object_default_value (ecma_object_t *obj_p, /**< the object */
 } /* ecma_op_general_object_default_value */
 
 /**
+ * Special type for ecma_op_general_object_define_own_property.
+ */
+#define ECMA_PROPERTY_TYPE_GENERIC ECMA_PROPERTY_TYPE_DELETED
+
+/**
  * [[DefineOwnProperty]] ecma general object's operation
  *
  * See also:
@@ -592,47 +597,64 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
                 && !ecma_is_lexical_environment (obj_p));
   JERRY_ASSERT (property_name_p != NULL);
 
-  const bool is_property_desc_generic_descriptor = (!property_desc_p->is_value_defined
-                                                    && !property_desc_p->is_writable_defined
-                                                    && !property_desc_p->is_get_defined
-                                                    && !property_desc_p->is_set_defined);
-  const bool is_property_desc_data_descriptor = (property_desc_p->is_value_defined
-                                                 || property_desc_p->is_writable_defined);
-  const bool is_property_desc_accessor_descriptor = (property_desc_p->is_get_defined
-                                                     || property_desc_p->is_set_defined);
+  ecma_property_types_t property_desc_type = ECMA_PROPERTY_TYPE_GENERIC;
+
+  if (property_desc_p->is_value_defined
+      || property_desc_p->is_writable_defined)
+  {
+    /* A property descriptor cannot be both named data and named accessor. */
+    JERRY_ASSERT (!property_desc_p->is_get_defined
+                  && !property_desc_p->is_set_defined);
+    property_desc_type = ECMA_PROPERTY_TYPE_NAMEDDATA;
+  }
+  else if (property_desc_p->is_get_defined
+           || property_desc_p->is_set_defined)
+  {
+    property_desc_type = ECMA_PROPERTY_TYPE_NAMEDACCESSOR;
+  }
+
+  /* These three asserts ensures that a new property is created with the appropriate default flags.
+   * E.g. if is_configurable_defined is false, the newly created property must be non-configurable. */
+  JERRY_ASSERT (property_desc_p->is_configurable_defined || !property_desc_p->is_configurable);
+  JERRY_ASSERT (property_desc_p->is_enumerable_defined || !property_desc_p->is_enumerable);
+  JERRY_ASSERT (property_desc_p->is_writable_defined || !property_desc_p->is_writable);
 
   // 1.
   ecma_property_t *current_p = ecma_op_object_get_own_property (obj_p, property_name_p);
 
-  // 2.
-  bool extensible = ecma_get_object_extensible (obj_p);
-
   if (current_p == NULL)
   {
     // 3.
-    if (!extensible)
+    if (!ecma_get_object_extensible (obj_p))
     {
+      // 2.
       return ecma_reject (is_throw);
     }
 
     // 4.
 
-    // a.
-    if (is_property_desc_generic_descriptor
-        || is_property_desc_data_descriptor)
+    if (property_desc_type != ECMA_PROPERTY_TYPE_NAMEDACCESSOR)
     {
+      // a.
+
+      JERRY_ASSERT (property_desc_type == ECMA_PROPERTY_TYPE_GENERIC
+                    || property_desc_type == ECMA_PROPERTY_TYPE_NAMEDDATA);
+
       ecma_property_t *new_prop_p = ecma_create_named_data_property (obj_p,
                                                                      property_name_p,
                                                                      property_desc_p->is_writable,
                                                                      property_desc_p->is_enumerable,
                                                                      property_desc_p->is_configurable);
 
+
+      JERRY_ASSERT (property_desc_p->is_value_defined
+                    || ecma_is_value_undefined (property_desc_p->value));
+
       ecma_named_data_property_assign_value (obj_p, new_prop_p, property_desc_p->value);
     }
     else
     {
       // b.
-      JERRY_ASSERT (is_property_desc_accessor_descriptor);
 
       ecma_create_named_accessor_property (obj_p,
                                            property_name_p,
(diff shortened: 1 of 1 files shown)


jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -658,71 +658,73 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
   ecma_property_types_t current_type = ECMA_PROPERTY_GET_TYPE (current_p);
   const bool is_current_data_descriptor = (current_type == ECMA_PROPERTY_TYPE_NAMEDDATA);
   const bool is_current_accessor_descriptor = (current_type == ECMA_PROPERTY_TYPE_NAMEDACCESSOR);
+  const bool current_property_configurable = ecma_is_property_configurable (current_p);
```
LaszloLango: Please follow the naming convention and give a name like the other two bool variables above. Ex.: `is_current_property_configurable` or `is_curr_prop_configurable`

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -658,71 +658,73 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
...
   JERRY_ASSERT (is_current_data_descriptor || is_current_accessor_descriptor);
 
-  bool is_every_field_in_desc_also_occurs_in_current_desc_with_same_value = true;
-  if (property_desc_p->is_value_defined)
+  if (!current_property_configurable)
   {
-    if (!is_current_data_descriptor
-        || !ecma_op_same_value (property_desc_p->value,
-                                ecma_get_named_data_property_value (current_p)))
+    bool is_every_field_in_desc_also_occurs_in_current_desc_with_same_value = true;
```
LaszloLango: IMHO, this variable name is way too long. If we touch this anyway, then we can add a better name and put the details into a comment.

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -658,71 +658,73 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
...
-    if (!is_current_data_descriptor
-        || property_desc_p->is_writable != ecma_is_property_writable (current_p))
+    if (property_desc_p->is_writable_defined)
     {
-      is_every_field_in_desc_also_occurs_in_current_desc_with_same_value = false;
+      if (!is_current_data_descriptor
+          || property_desc_p->is_writable != ecma_is_property_writable (current_p))
+      {
+        is_every_field_in_desc_also_occurs_in_current_desc_with_same_value = false;
+      }
```
akiss77: These `if (a) { if (!b || !c) {} }` constructs are a very interesting way of writing `if (a && (!b || !c))`. (Used all over here.)

zherczeg:
Patch updated.

LaszloLango:
Thanks for the update, LGTM

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -658,77 +656,50 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
...
   JERRY_ASSERT (is_current_data_descriptor || is_current_accessor_descriptor);
 
-  bool is_every_field_in_desc_also_occurs_in_current_desc_with_same_value = true;
-  if (property_desc_p->is_value_defined)
+  if (!is_current_configurable)
   {
-    if (!is_current_data_descriptor
-        || !ecma_op_same_value (property_desc_p->value,
-                                ecma_get_named_data_property_value (current_p)))
+    if (property_desc_p->is_configurable
```
akiss77: Moving step 7 here (btw, comment not transfered) introduces an incompatible change. Let's assume the following situation (written in pseudo-code for brewity):

!current.configurable && prop.configurable
&& current.writable == prop.writable
&& current.value == prop.value
&& current.getter == prop.getter
&& current.setter == prop.setter

The original version will return with true, while this PR rejects it.

zherczeg: !current.configurable && prop.configurable - this cannot happen

akiss77: Why is this `if()` checking it then? When reaching this line, we know that current is not configurable and check whether prop is configurable.

zherczeg: You are always confusing me. See the deleted code below in line 702.

zherczeg:
Added asserts which shows the "hidden design rules", and also ensures that they are true. Test262 runs correctly.

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -760,7 +780,7 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
                                                        NULL,
                                                        NULL,
                                                        was_enumerable,
-                                                       was_configurable);
+                                                       is_current_configurable);
```
akiss77: Because of the early return condition above (at step 9), don't we know that is_current_configurable is true for sure?

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -770,51 +790,7 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
...
-
-      if ((property_desc_p->is_get_defined
-           && property_desc_p->get_p != ecma_get_named_accessor_property_getter (current_p))
-          || (property_desc_p->is_set_defined
-              && property_desc_p->set_p != ecma_get_named_accessor_property_setter (current_p)))
-      {
-        // i., ii.
-        return ecma_reject (is_throw);
-      }
+                                                   is_current_configurable);
```
akiss77: same here

jerry-core/ecma/operations/ecma-objects-general.c
akiss77: It seems I cannot put a comment on the enclosing if (it's not in the diff'd region). However, `if (is_current_data_descriptor)` is a badly misleading (although functinoally correct) predicate. It should be `if (!is_property_desc_data_descriptor)` or `if (is_property_desc_accessor_descriptor)`, as the `ecma_create_named_*`  calls don't work with "current" (which is already deleted) but with "property".

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -658,77 +660,52 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
   ecma_property_types_t current_type = ECMA_PROPERTY_GET_TYPE (current_p);
   const bool is_current_data_descriptor = (current_type == ECMA_PROPERTY_TYPE_NAMEDDATA);
   const bool is_current_accessor_descriptor = (current_type == ECMA_PROPERTY_TYPE_NAMEDACCESSOR);
+  const bool is_current_configurable = ecma_is_property_configurable (current_p);
 
   JERRY_ASSERT (is_current_data_descriptor || is_current_accessor_descriptor);
 
-  bool is_every_field_in_desc_also_occurs_in_current_desc_with_same_value = true;
-  if (property_desc_p->is_value_defined)
+  if (!is_current_configurable)
```
akiss77: I still have doubts here about equivalence with original code. Prove me wrong. If current is configurable, you sidestep these early return checks below and will certainly execute step 12, potentially with all the set and assign calls, even if unnecessary.

zherczeg: If configurable is true you can do anything with the property.

I think you should check this code rather then the current implementation:
http://www.ecma-international.org/ecma-262/5.1/#sec-8.12.9

You need a functionally equivalent code for this.

jerry-core/ecma/operations/ecma-objects-general.c
zherczeg: This is tricky here. If is_current_data_descriptor is true, we deleted a data descriptor, and create an accessor. We have no other choice.

zherczeg: Yes, is_current_configurable must be true. That is a clever observation.

jerry-core/ecma/operations/ecma-objects-general.c
zherczeg: I think I understand what you want to know here. This is a tricky code. If is_current_configurable is false we return with TRUE regardless of the descriptor. Hence doing early TRUE return with costly checks does not gain anything from our perspective. We simply update the property (might even with the same values) and return TRUE. Since same_value and those checks are costly, so we are still getting a faster code.

jerry-core/ecma/operations/ecma-objects-general.c
akiss77: As I tracked it down: by the time we reach this statement, we know: 
- `is_property_desc_generic_descriptor == false`
- `is_property_desc_data_descriptor != is_current_data_descriptor`
- `is_current_data_descriptor == true`

The last two implies that `is_property_desc_data_descriptor == false`. From this and the first item of the above list we can also infer that `is_property_desc_accessor_descriptor == true`. And actually _that's_ why we are calling `ecma_create_named_accessor_property` here. That's why I think that `if (is_property_desc_accessor_descriptor)` would be the right predicate. IMHO.

zherczeg:
I am trying to explain what is happening here:

Step 1:
    current = READ_CURRENT_DESCRIPTOR(property_name)

    if (!current) {
      CREATE_NEW_DESCRIPTOR()
      // The create above always used the following "hidden" rule:
      //   property_desc_p->is_configurable_defined
      //   || property_desc_p->is_configurable == true
      // If is_configurable_defined is NOT defined, the new property must be
      // non-configurable because that is the default. Same for the
      // other flags (enumerable, writable, etc.)
      // Now an assertion is added to check this.
    }

    // Otherwise we do the following:
    if (CHANGES_ALLOWED (current, new_descriptor)) {
      DO_CHANGES
      return TRUE;
    }
    return FALSE;

The CHANGES_ALLOWED check above is costly. So we do the following instead:

    if (current->is_configurable()) {
      // We know changes are allowed in this case, 
      // regardless we update to the same or another value
      DO_CHANGES
      return TRUE;
    }
    else {
      // Here we need to do a lot of checks, but this is a rare case:
      if (CHANGES_ALLOWED (current, new_descriptor)) {
        DO_CHANGES
        return TRUE;
      }
      return FALSE;
    }

The simplified form of above looks like the following:

    if (!current->is_configurable()) {
      if (!CHANGES_ALLOWED (current, new_descriptor)) {
        return FALSE;
      }
    }
    DO_CHANGES
    return TRUE;

The DO_CHANGES part depend on that both is_property_desc_data_descriptor and is_property_desc_accessor_descriptor most not be true. This was also in the code, but now there is an assertion to check this another "hidden" rule.

zherczeg:
I have been thinking about this function. It was made in the early phase of Jerry, and never got a real optimization rework. But it is a key function, and it is worth to optimize it. Apart from the data<->accessor switch, everything is heavily optimized now. If that conversion turns out to be a hot path, I will optimize that as well.

zherczeg:
Note: test262 es5 branch and precommit runs cleanly.

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -592,166 +592,162 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
...
-    {
-      is_every_field_in_desc_also_occurs_in_current_desc_with_same_value = false;
-    }
-  }
+  JERRY_ASSERT (current_property_type == ECMA_PROPERTY_TYPE_NAMEDDATA
+                || current_property_type == ECMA_PROPERTY_TYPE_NAMEDACCESSOR);
 
-  if (property_desc_p->is_set_defined)
+  /* If current is configurable, any change is allowed. */
+  if (!is_current_configurable)
```
akiss77: Merge these nested `if`s with `&&`.

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -592,166 +592,162 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
...
+      if (property_desc_type == ECMA_PROPERTY_TYPE_NAMEDDATA)
+      {
+        // 10.
 
-  if (is_every_field_in_desc_also_occurs_in_current_desc_with_same_value)
-  {
-    return ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE);
-  }
+        // a.
+        if (!ecma_is_property_writable (current_p))
```
akiss77: I think this could be expressed in a shorter form by writing
```c
        // 10. a. i. & ii.
        if (!ecma_is_property_writable (current_p) 
            && (property_desc_p->is_writable 
                || (property_desc_p->is_value_defined 
                    && !ecma_op_same_value (property_desc_p->value, ecma_get_named_data_property_value (current_p))))
        {
          return ecma_reject (is_throw);
        }
```

jerry-core/ecma/operations/ecma-objects-general.c
zherczeg: Actually I don't prefer too complex ifs. Hard to read. Do you insist on these changes?

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -592,166 +592,162 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
...
   {
     // 9.
-    if (!ecma_is_property_configurable (current_p))
+    if (!is_current_configurable)
     {
       // a.
       return ecma_reject (is_throw);
     }
 
+    /* The following implementation can be optimized by directly overwrite the
```
akiss77: typo: overwrite -> overwriting

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -257,6 +257,8 @@ typedef enum
 typedef enum
 {
   ECMA_PROPERTY_TYPE_DELETED, /**< deleted property */
+  ECMA_PROPERTY_TYPE_GENERIC = ECMA_PROPERTY_TYPE_DELETED, /**< generic property type
```
akiss77: I'm a bit concerned about polluting this public enum type with `ECMA_PROPERTY_TYPE_GENERIC`, as that marker is only used in and meaningful for `ecma_op_general_object_define_own_property`. What about adding
```c
#define ECMA_PROPERTY_TYPE_GENERIC ECMA_PROPERTY_TYPE_DELETED
```
and
```c
#undef ECMA_PROPERTY_TYPE_GENERIC
```
around `ecma_op_general_object_define_own_property` in `ecma-objects-general.c`?

jerry-core/ecma/operations/ecma-objects-general.c
akiss77: I do, especially here.

As an example, in step 11. a., we also have a 4-line condition which is intuitive to read: "if we want to overwrite a getter with something that's not the same as we already have OR we want to overwrite a setter with something that's not the same as we already have..."

Here, a very similar condition is broken to nested ifs for no apparent reason and makes me spend brain cycles on figuring out what needs special handling. Outcome: nothing. Our condition here is: "if current property is non-configurable but we either want to make it configurable or change its enumerable status..."

There is no else branch either in the inner or in the outer if. So, I see no point in nesting them. (Or do we have a guidline that predicates have to be in disjunctive normal form?)

zherczeg:
Fixed all requests.

akiss77:
Thanks, LGTM

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -860,6 +803,8 @@ ecma_op_general_object_define_own_property (ecma_object_t *obj_p, /**< the objec
   return ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE);
 } /* ecma_op_general_object_define_own_property */
 
+#undef ECMA_PROPERTY_TYPE_GENERIC
```
LaszloLango: This undef looks unnecessary. This isn't a header file and the undef is at the of the source, so won't be any redefinition of it. Did I missed something here?

zherczeg: In all-in-one build it is good if noone references to this define.

LaszloLango:
Still LGTM

PR #1091, opened 2016-05-23
Print each output line with one command in test runners

akiss77:
When running multiple test suites in parallel -- e.g., as it
happens on the CI --, the names of the executed commands and the
results of the executions can/do get printed far from each other,
interrupted by other prints. This can make the reading of the
output hard. This patch prints each line with one echo command,
which makes the interruption of the line less likely.

### DIFF
--- a/tools/runners/run-test-suite.sh
+++ b/tools/runners/run-test-suite.sh
@@ -112,35 +112,31 @@ do
     if [ "$IS_SNAPSHOT" == true ]
     then
         # Testing snapshot
-
         SNAPSHOT_TEMP=`mktemp snapshot-out.XXXXXXXXXX`
 
-        echo -n "[$tested/$total] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS "
-        echo -n "--save-snapshot-for-global $SNAPSHOT_TEMP ${full_test#$ROOT_DIR}: "
-
+        cmd_line="${ENGINE#$ROOT_DIR} $ENGINE_ARGS --save-snapshot-for-global $SNAPSHOT_TEMP ${full_test#$ROOT_DIR}"
         ( ulimit -t $TIMEOUT; $ENGINE $ENGINE_ARGS --save-snapshot-for-global $SNAPSHOT_TEMP $full_test &> $ENGINE_TEMP )
         status_code=$?
 
         if [ $status_code -eq 0 ]
         then
-            echo "$PASS"
-            echo -n "[$tested/$total] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS --exec-snapshot $SNAPSHOT_TEMP: "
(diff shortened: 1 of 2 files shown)


tools/runners/run-test-suite.sh
```diff
@@ -112,35 +112,31 @@ do
...
     then
         # Testing snapshot
-
         SNAPSHOT_TEMP=`mktemp snapshot-out.XXXXXXXXXX`
 
-        echo -n "[$tested/$total] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS "
-        echo -n "--save-snapshot-for-global $SNAPSHOT_TEMP ${full_test#$ROOT_DIR}: "
-
+        cmd_line="${ENGINE#$ROOT_DIR} $ENGINE_ARGS --save-snapshot-for-global $SNAPSHOT_TEMP ${full_test#$ROOT_DIR}"
         ( ulimit -t $TIMEOUT; $ENGINE $ENGINE_ARGS --save-snapshot-for-global $SNAPSHOT_TEMP $full_test &> $ENGINE_TEMP )
```
zherczeg: Since the command line is in a variable now, I think we should combine these lines. This would improve maintenance.

tools/runners/run-unittests.sh
```diff
@@ -59,14 +59,13 @@ UNITTEST_TEMP=`mktemp unittest-out.XXXXXXXXXX`
...
-    echo -n "[$tested/$total] ${unit_test#$ROOT_DIR}: "
-
+    cmd_line="${unit_test#$ROOT_DIR}"
     $unit_test &>$UNITTEST_TEMP
     status_code=$?
 
     if [ $status_code -ne 0 ]
     then
-        echo "FAIL ($status_code)"
+        echo "[$tested/$total] $cmd_line: FAIL ($status_code)"
```
zherczeg: Do we really need this cmd_line variable here?

tools/runners/run-test-suite.sh
akiss77: We cannot easily combine `cmd_line` and the actual execution since (after a recent change) we don't print exactly what we execute but strip some potentially lengthy prefix from the paths. (Otherwise, log would be too long and Travis would kill the process for exceeding log size limits.) That's why we have `${ENGINE#$ROOT_DIR}` and `${full_test#$ROOT_DIR}` above, where `$ROOT_DIR` is the prefix we don't print.

tools/runners/run-unittests.sh
akiss77: I wanted to keep the same pattern in both runner scripts, and also didn't want to duplicate the path strip logic (`#$ROOT_DIR`) to help maintainability. Both it's no big code so we can merge this if you think so.

zherczeg:
Ok. LGTM

LaszloLango:
LGTM

PR #1095, opened 2016-05-25
Improve ecma_op_general_object_put

zherczeg:
Canput is removed, properties are directly created

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects-general.c
+++ b/jerry-core/ecma/operations/ecma-objects-general.c
@@ -246,14 +246,41 @@ ecma_op_general_object_get_property (ecma_object_t *obj_p, /**< the object */
 } /* ecma_op_general_object_get_property */
 
 /**
+ * Checks whether the property name is "length".
+ *
+ * @return true if the property name is length
+ *         false otherwise
+ */
+static inline bool __attr_always_inline___
+ecma_op_general_object_property_name_is_length (ecma_string_t *property_name_p) /**< property name */
+{
+  ecma_string_t *magic_string_length_p = ecma_get_magic_string (LIT_MAGIC_STRING_LENGTH);
+
+  bool property_name_is_length = ecma_compare_ecma_strings (property_name_p,
+                                                            magic_string_length_p);
+
+  ecma_deref_ecma_string (magic_string_length_p);
+
+  return property_name_is_length;
+} /* ecma_op_general_object_property_name_is_length */
+
+/**
  * [[Put]] ecma general object's operation
  *
  * See also:
  *          ECMA-262 v5, 8.6.2; ECMA-262 v5, Table 8
  *          ECMA-262 v5, 8.12.5
+ *          Also incorporates [[CanPut]] ECMA-262 v5, 8.12.4
  *
  * @return ecma value
- *         Returned value must be freed with ecma_free_value
+ *         The returned value must be freed with ecma_free_value.
+ *
+ *         Returns with ECMA_SIMPLE_VALUE_TRUE if the operation is
+ *         successful. Otherwise it returns with an error object
+ *         or ECMA_SIMPLE_VALUE_FALSE.
+ *
+ *         Note: even if is_throw is false, the setter can throw an
+ *         error, and this function returns with that error.
  */
 ecma_value_t
 ecma_op_general_object_put (ecma_object_t *obj_p, /**< the object */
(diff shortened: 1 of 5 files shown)


zherczeg:
Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 1.966 -> 1.883 : 4.193% | 
3d-raytrace.js | 2.135 -> 2.118 : 0.815% | 
access-binary-trees.js | 1.078 -> 0.861 : 20.176% | 
access-fannkuch.js | 4.982 -> 4.302 : 13.650% | 
access-nbody.js | 2.212 -> 2.102 : 4.962% | 
bitops-3bit-bits-in-byte.js | 0.812 -> 0.811 : 0.107% | 
bitops-bits-in-byte.js | 1.142 -> 1.144 : -0.152% | 
bitops-bitwise-and.js | 2.233 -> 1.903 : 14.803% | 
bitops-nsieve-bits.js | 3.203 -> 2.990 : 6.655% | 
controlflow-recursive.js | 0.565 -> 0.563 : 0.375% | 
crypto-aes.js | 2.440 -> 2.139 : 12.338% | 
crypto-md5.js | 1.046 -> 1.015 : 3.026% | 
crypto-sha1.js | 0.970 -> 0.952 : 1.943% | 
date-format-tofte.js | 1.582 -> 1.550 : 2.048% | 
date-format-xparb.js | 0.659 -> 0.665 : -0.975% | 
math-cordic.js | 2.145 -> 2.130 : 0.706% | 
math-partial-sums.js | 1.226 -> 1.148 : 6.365% | 
math-spectral-norm.js | 0.992 -> 0.991 : 0.093% | 
string-base64.js | 4.938 -> 4.980 : -0.862% | 
string-fasta.js | 2.532 -> 2.495 : 1.480% | 
Geometric mean: | 5.020% |

jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -417,16 +417,15 @@ ecma_op_array_object_define_own_property (ecma_object_t *obj_p, /**< the array o
     }
 
     // e.
-    if (index >= old_len_uint32)
+    if (index < UINT32_MAX
+        && index >= old_len_uint32)
```
akiss77: This seems to be a completely unrelated change. Please have this in a separate patch.

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -246,6 +246,25 @@ ecma_op_general_object_get_property (ecma_object_t *obj_p, /**< the object */
 } /* ecma_op_general_object_get_property */
 
 /**
+ * Checks whether the property name is "length".
+ *
+ * @return true if the property name is length
+ *         false otherwise
+ */
+static bool
+ecma_op_general_object_property_name_is_length (ecma_string_t *property_name_p) /**< property name */
```
akiss77: Is this function inlined? Since it is used in release code only once, I'd suggest explicitly asking the compiler to inline this. (`inline` and/or `__attr_always_inline___`)

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -265,179 +284,150 @@ ecma_op_general_object_put (ecma_object_t *obj_p, /**< the object */
...
+    return ret_value;
   }
 
-  JERRY_UNREACHABLE ();
-} /* ecma_op_general_object_can_put */
+  if (is_throw)
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG (""));
+  }
+  return ecma_make_simple_value (ECMA_SIMPLE_VALUE_FALSE);
```
akiss77: isn't this equivalent to `ecma_reject(is_throw)`?

jerry-core/ecma/operations/ecma-objects-general.c
zherczeg: I expect the compielr is clever enough to do it, but I don't want to force it. In debug mode we use it twice btw.

jerry-core/ecma/operations/ecma-objects-general.c
zherczeg: I was wondering about that too.

akiss77:
I cannot put a comment on the header of the function so commenting here:
* `ecma_op_general_object_put` now also incorporates the logic of `[[CanPut]]`, thus it's doc comment should be extended with a note at least, e.g., like "(Also implementing [[CanPut]] ecma general object's operation internally, ECMA-262 v5, 8.12.4.)"
* The return value needs more/better documentation. According to the standard, [[Put]] has no return value, it can only throw an error or return nothing. Thus, this deviaton from the standard's prescribed behaviour should be documented, e.g., like "@return ecma value. True in case of success, False or Error in case of failure, depending on is_throw. As returned value is not always simple, it must be freed with ecma_free_value. (Returning a boolean value deviates from the word of the standard but this is an implementation detail.)".

(Feel free to tweek or style the above suggestions as they fit the comment the best.)

Noteworthy:
* Documentation defects like the one mentioned above are quite frequent in this part of the code.
* A bigger problem is that the return value is actually _not_ always freed by some callers (of `ecma_op_object_put`).

jerry-core/ecma/operations/ecma-objects-general.c
akiss77: I don't get this. Our code base is full of `inline`s and `__attr_always_inline___`s to make sure that performance critical paths are inlined. This patch seems to be focusing on performance. Why not force inlining this time then? Especially for a function which is only used once (I do realize that debug version uses it twice but debug mode shouldn't count when optimizing performance), thus inlining does not only help performance but code size as well (as it makes sure that function call, parameter and return value passing, prolog and epilog don't get generated.

jerry-core/ecma/operations/ecma-objects-general.c
akiss77: Please reuse it.

jerry-core/ecma/operations/ecma-objects-general.c
LaszloLango: IMHO, until the JERRY_ASSERT is not defined as empty the compiler won't inline this because it is called twice. We should use `inline` and `__attr_always_inline___` to force it.

zherczeg:
It turned out that even if is_throw is false, we need to return with an error object if the setter throws an error.

akiss77:
Does that mean that neither of the test suites (tests/jerry, tests/jerry-test-suite, ecma262 es51) caught this? Please, extend our test suite with a case that can check the correct behaviour.

zherczeg:
Our test suite does not check it, but it was the old behaviour as well, and matches to Duktape and Firefox. I did not test it with test262.

zherczeg:
I added a test, and checked the test works without the patch.

tests/jerry/object-defineproperty.js
```diff
@@ -55,3 +55,17 @@ Object.defineProperty(obj, "prop", {
 
 var desc2 = Object.getOwnPropertyDescriptor(obj, "prop");
 assert (desc1.set === setter && desc2.set === undefined);
+
+obj = {}
+
+/* This error is thrown even in non-strict mode. */
+Object.defineProperty(obj, 'f', {
+  set: function(value) { throw 234 },
```
akiss77: Add `;` after 234. Not mandated by the standard but it is good practice not to excercise the semicolon insertion logic of the parser if it's not the feature we want to test.

jerry-core/ecma/operations/ecma-objects-general.c
akiss77: `inline` and `__attr_always_inline___` still missing

zherczeg:
Patch updated.

tests/jerry/object-defineproperty.js
```diff
@@ -55,3 +55,17 @@ Object.defineProperty(obj, "prop", {
 
 var desc2 = Object.getOwnPropertyDescriptor(obj, "prop");
 assert (desc1.set === setter && desc2.set === undefined);
+
+obj = {}
```
akiss77: Sorry for the late hit, semicolon missing here as well

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -265,179 +292,147 @@ ecma_op_general_object_put (ecma_object_t *obj_p, /**< the object */
...
-                                         true, /* Writable */
-                                         true, /* Enumerable */
-                                         true, /* Configurable */
-                                         is_throw); /* Failure handling */
-  }
-
-  JERRY_UNREACHABLE ();
-} /* ecma_op_general_object_put */
+    ecma_object_t *proto_p = ecma_get_object_prototype (obj_p);
+    bool create_new_property = true;
```
LaszloLango: Please use the `is_` prefix for bool variables (at least for the new ones).

zherczeg: I don't think "is" is grammatically ok here.

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -265,179 +292,147 @@ ecma_op_general_object_put (ecma_object_t *obj_p, /**< the object */
...
-  {
-    return ecma_get_object_extensible (obj_p);
-  }
+        uint32_t old_len = ecma_get_uint32_from_value (ecma_get_named_data_property_value (len_prop_p));
 
-  // 7.
-  if (ECMA_PROPERTY_GET_TYPE (inherited_p) == ECMA_PROPERTY_TYPE_NAMEDACCESSOR)
-  {
-    ecma_object_t *setter_p = ecma_get_named_accessor_property_setter (inherited_p);
+        if (index < UINT32_MAX
```
LaszloLango: What happens if `index == UINT32_MAX`? Is it possible to reach the maximum value?

zherczeg: A property with ID UINT32_MAX is not part of an array, just like 0.5 or -3.

    var a = []; a[0.5] = 5; // a.length is zero, the array has no items.
                            // But it has a property which name is 0.5
    var a = []; a[1] = 5; // a.length is two, the array has two items.

zherczeg:
These flaky travis runs are tiring.

jerry-core/ecma/operations/ecma-objects.h
```diff
@@ -30,7 +30,6 @@ extern ecma_value_t ecma_op_object_get (ecma_object_t *, ecma_string_t *);
 extern ecma_property_t *ecma_op_object_get_own_property (ecma_object_t *, ecma_string_t *);
 extern ecma_property_t *ecma_op_object_get_property (ecma_object_t *, ecma_string_t *);
 extern ecma_value_t ecma_op_object_put (ecma_object_t *, ecma_string_t *, ecma_value_t, bool);
-extern bool ecma_op_object_can_put (ecma_object_t *, ecma_string_t *);
```
akiss77: Update copyright year here

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -270,45 +270,6 @@ ecma_op_object_put (ecma_object_t *obj_p, /**< the object */
 } /* ecma_op_object_put */
 
 /**
- * [[CanPut]] ecma object's operation
```
akiss77: Same here (copyright year update needed)

jerry-core/ecma/operations/ecma-objects-general.h
```diff
@@ -35,7 +35,6 @@ extern ecma_value_t ecma_op_general_object_get (ecma_object_t *, ecma_string_t *
 extern ecma_property_t *ecma_op_general_object_get_own_property (ecma_object_t *, ecma_string_t *);
 extern ecma_property_t *ecma_op_general_object_get_property (ecma_object_t *, ecma_string_t *);
 extern ecma_value_t ecma_op_general_object_put (ecma_object_t *, ecma_string_t *, ecma_value_t, bool);
-extern bool ecma_op_general_object_can_put (ecma_object_t *, ecma_string_t *);
```
akiss77: Same here (copyright year update needed)

jerry-core/ecma/operations/ecma-objects-general.c
zherczeg: Even more interesting that all proeprty names are strings. Hence:

    var a = []; a[true] = 5; // a["true"] is 5

zherczeg:
Copyrights added.

LaszloLango:
LGTM

akiss77:
LGTM

PR #1100, opened 2016-05-26
Replace multiple boolean arguments to one uint8_t.

zherczeg:


zherczeg:
Binary size decreased by 1K.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -576,26 +576,13 @@ ecma_get_internal_property (ecma_object_t *object_p, /**< object descriptor */
 ecma_property_t *
 ecma_create_named_data_property (ecma_object_t *object_p, /**< object */
                                  ecma_string_t *name_p, /**< property name */
-                                 bool is_writable, /**< 'Writable' attribute */
-                                 bool is_enumerable, /**< 'Enumerable' attribute */
-                                 bool is_configurable) /**< 'Configurable' attribute */
+                                 uint8_t prop_attributes) /**< property attributes */
```
LaszloLango: I like this idea, but I'd give a name to this. What do you think about `typedef uint8_t prop_attr_t;` or `typedef uint8_t prop_attr_flag_t;`?

LaszloLango: I just saw that we already have a `ecma_property_flags_t`, but it is an enum, so it'll be 32bit long (usually, but not always). In this case I think it is unnecessary to create a new type, but it should be mentioned in the comment.

zherczeg: That is a flagset, not a flag. The type_and_flags in ecma_property_t is also an uint8_t. I wouldn't want to change it.

zherczeg:
Any more thoughts?

jerry-core/ecma/base/ecma-helpers.c
LaszloLango: As I said, you don't have to change the name, just make a comment.

```c
...
uint8_t prop_attributes) /**< property attributes (See: ecma_property_flags_t) */
```

LaszloLango:
LGTM, after the comment fixed.

dbatyai:
LGTM

PR #1103, opened 2016-05-26
Fix ESP8266 build

seanshpark:
This change is to fix build for ESP8266 board with jerry_api_value_t changes.

### DIFF
--- a/targets/esp8266/source/jerry_run.c
+++ b/targets/esp8266/source/jerry_run.c
@@ -24,19 +24,20 @@
 static const char* fn_sys_loop_name = "sysloop";
 
 
-//-----------------------------------------------------------------------------
+/*---------------------------------------------------------------------------*/
 
 int js_entry (const char *source_p, const size_t source_size)
 {
   const jerry_api_char_t *jerry_src = (const jerry_api_char_t *) source_p;
   jerry_completion_code_t ret_code = JERRY_COMPLETION_CODE_OK;
   jerry_flag_t flags = JERRY_FLAG_EMPTY;
+  jerry_api_object_t *err_obj_p = NULL;
 
   jerry_init (flags);
 
   js_register_functions ();
 
-  if (!jerry_parse (jerry_src, source_size))
+  if (!jerry_parse ((jerry_api_char_t *)jerry_src, source_size, &err_obj_p))
   {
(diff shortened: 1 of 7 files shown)


seanshpark:
Though this is just a build fix, it fails to run in the target board for some reason that needs investigation. I'll work on this after this PR lands.

targets/esp8266/source/jerry_run.c
```diff
@@ -45,7 +46,8 @@ int js_entry (const char *source_p, const size_t source_size)
   {
     if ((flags & JERRY_FLAG_PARSE_ONLY) == 0)
     {
-      ret_code = jerry_run ();
+      jerry_api_value_t res;
+      ret_code = jerry_run (&res);
```
LaszloLango: `jerry_api_release_value` should be called if jerry_run failes. The parameter of `jerry_run` is for error value not for result, so the name of the variable is misleading.

targets/esp8266/source/jerry_run.c
```diff
@@ -45,7 +46,8 @@ int js_entry (const char *source_p, const size_t source_size)
   {
     if ((flags & JERRY_FLAG_PARSE_ONLY) == 0)
     {
-      ret_code = jerry_run ();
+      jerry_api_value_t res;
```
LaszloLango: It is not mandatory, but good practice to initialize `res` to `VOID` (`jerry_api_create_void_value`).

targets/esp8266/source/jerry_run.c
seanshpark: got it. thanks.

zherczeg:
LGTM

LaszloLango:
LGTM

LaszloLango:
@seanshpark, please rebase to the current master, so we can land your PR.

seanshpark:
Done rebase, thank you.

seanshpark:
CI fails, last merge is #1104. @zherczeg , can you please check ?

zherczeg:
We have already spent a lot of time reproducing this issue, but no luck so far. It only appears on travis arm build. I submitted another patch, and it disappeared.

seanshpark:
Thank you, rebase and merged.

PR #1104, opened 2016-05-30
Optimize conditional jumps.

zherczeg:
The ecma_op_to_boolean return value is changed to bool for faster
evaluation, and no need to swap operandos of relational compare.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-value.c
+++ b/jerry-core/ecma/base/ecma-helpers-value.c
@@ -59,6 +59,10 @@ JERRY_STATIC_ASSERT (sizeof (uintptr_t) > sizeof (ecma_value_t),
 
 #endif /* ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY */
 
+JERRY_STATIC_ASSERT ((ECMA_SIMPLE_VALUE_FALSE | 0x1) == ECMA_SIMPLE_VALUE_TRUE
+                     && ECMA_SIMPLE_VALUE_FALSE != ECMA_SIMPLE_VALUE_TRUE,
+                     only_the_lowest_bit_must_be_different_for_simple_value_true_and_false);
+
 /**
  * Get type field of ecma value
  *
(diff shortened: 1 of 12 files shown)


jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -564,6 +582,22 @@ ecma_get_object_from_value (ecma_value_t value) /**< ecma value */
...
 
 /**
+ * Invert a boolean value
+ *
+ * @return ecma value
+ */
+inline ecma_value_t __attr_pure___ __attr_always_inline___
+ecma_invert_boolean_value (ecma_value_t value) /**< ecma value */
+{
+  JERRY_STATIC_ASSERT ((ECMA_SIMPLE_VALUE_FALSE | 0x1) == ECMA_SIMPLE_VALUE_TRUE
```
LaszloLango: This static assert should be right after the definition of ECMA_SIMPLE_VALUE_TRUE and ECMA_SIMPLE_VALUE_FALSE.

zherczeg: You cannot do that because they are in the header. And only this function depends on this condition at the moment.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -143,11 +144,13 @@ extern ecma_value_t ecma_make_object_value (const ecma_object_t *);
 extern ecma_value_t ecma_make_error_value (ecma_value_t);
 extern ecma_value_t ecma_make_error_obj_value (const ecma_object_t *);
 extern ecma_integer_value_t ecma_get_integer_from_value (ecma_value_t);
+extern ecma_number_t ecma_get_float_from_value (ecma_value_t value);
```
dbatyai: These new functions were tagged with `__attr_pure___` at the definition, shouldn't it be added here as well?

zherczeg:
Thank you, fixed.

dbatyai:
LGTM

jerry-core/ecma/base/ecma-helpers-value.c
LaszloLango: OK, at least move out of the function. IMHO, it should be placed at the top of the file like in `ecma-helpers.c`

zherczeg:
Moved to the beginning

LaszloLango:
@zherczeg, thanks for the update. LGTM

PR #1099, opened 2016-05-25
API documentation update #1

ktorpi:
JerryScript-DCO-1.0-Signed-off-by: István Kádár ikadar@inf.u-szeged.hu

PR #1069, opened 2016-05-18
Internals doc update after lcache

ktorpi:
Sections after Lcache are updated, sections Property Hashmap, Literal Store and Snapshot are added.

PR #1053, opened 2016-05-11
Documentation update

ktorpi:
Updated section ECMA of JerryScript Internals documentation ending with subsection LCache.

PR #1038, opened 2016-05-04
Documentation update

ktorpi:
Updated JerryScript Internals documentation form section High-Level design to section Virtual Machine.

JerryScript-DCO-1.0-Signed-off-by: István Kádár ikadar@inf.u-szeged.hu

PR #1106, opened 2016-05-31
API documentation update under docs.

ktorpi:
Links to JerryScript API reference page (https://samsung.github.io/jerryscript/API/) are inserted.


zherczeg:
I see it is just delete. The reference points to the removed text?

LaszloLango:
@zherczeg, yes. The markdown file of the API reference was duplicated. This PR makes the further maintenance easier. My only note is that #1099 should be merged at first. LGTM, after that.

zherczeg:
Ok. LGTM then.

lemmaa:
@LaszloLango , in my personal opinion, keeping API documents with source code is better then this. Historically, we moved documents from web site to source code repository. How about find the way to sync documents from source repository to web pages? Composing the web site in `*.md` file is one of reason for it. :)

LaszloLango:
@lemmaa, It is still in the source code repository. It's only moved to the `gh-pages` branch. We already had two md files for the same information (one in the master branch and one in gh-pages branch). The API reference is still in a markdown file and you can make pull requests to change it, so we can do better reviews before land. This is what @ktorpi does in nowadays. See: #1038, #1053, #1069, 1099, #1106 and #1117

I think this was your goal too, if you are referring to this issue: https://github.com/Samsung/iotjs/issues/315.  But please fix me if I misunderstood something. :)

lemmaa:
@LaszloLango, In my comment, `document with source code repository` means in same branch. I think documents must be distributed with source code in single tar ball, when it releases. Is it suitable to make a single tar ball from source code and document in separate branch? 

Form another point of view, version syncs between source code and document also looks complicate on separate branch.

PR #1108, opened 2016-05-31
Wip/sergio/zephyr arduino101

sergioamr:
Added target to compile zephyr against jerryscript.
This patch contains a project in which you can run
a small js test inside the Arduino101.

Follows the Nuttx implementation to replicate a command
line in the Quark SE Lakemont architecture.

Everything is self contained in the targets/arduino_101 folder.

Check the README.md for a more detailed explanation of
how to compile and run it.

Added Qemu & DFU flashing support not requiring JTAG.

JerryScript-DCO-1.0-Signed-off-by: Sergio Martinez <sergio.martinez.rodriguez@intel.com>


LaszloLango:
@sergioamr, could you rebase it to the current master?

sergioamr:
Rebase done, sorry i was not at the office with a dev computer :)

bzsolt:
Please correct the signed-offs to this form: JerryScript-DCO-1.0-Signed-off-by: ...
Details: https://github.com/Samsung/jerryscript/wiki/JerryScript-Developer's-Certificate-of-Origin

zherczeg:
I know little about arduino 101 but the patch makes sense.

sergioamr:
@LaszloLango  Thanks for the labels, i will start adding them on my next commits.

zherczeg:
At some point please rebase it to master and squash the commit.

LaszloLango:
The Sign-off is still incorrect in the commit message. See the travis logs for details:

> Signed-off-by message is incorrect. The following line should be at the end of the 509bbb9 commit's message: 'JerryScript-DCO-1.0-Signed-off-by: Sergio Martinez sergio.martinez.rodriguez@intel.com'.

sergioamr:
@LaszloLango  Removed Adrian's signed off, i wrote that bit on his computer anyway while testing the integration. For my record, how should i add two authors if i squash a small couple of lines commit?, 

JerryScript-DCO-1.0-Signed-off-by: Random A Developer randomA@developer.example.org
JerryScript-DCO-1.0-Signed-off-by: Random B Developer randomB@developer.example.org

targets/arduino_101/scripts/get-zephyr.sh
```diff
@@ -0,0 +1,22 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#sudo apt-get update -q
+#sudo apt-get install -q -y \
+# dependencies
+
+# Clone zephyr to deps
+# git clone https://gerrit.zephyrproject.org/r/zephyr
```
LaszloLango: This file contains only comment lines. Please remove it you don't need it or uncomment the necessary parts.

targets/arduino_101/src/jerry-entry.c
```diff
@@ -0,0 +1,48 @@
...
+ * limitations under the License.
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <stdarg.h>
+
+#include "jerry.h"
+#include "jrt/jrt.h"
```
LaszloLango: I think this include is unnecessary. Do you use something from it in this file?

targets/arduino_101/src/libc_support.c
```diff
@@ -0,0 +1,105 @@
+/* Copyright 2016 Samsung Electronics Co., Ltd.
```
LaszloLango: Why is this file needed? It is a copy-paste of jerry-libc functions, but ports can call jerry-libc functions. Did I miss something here? :)

sergioamr: I have to use the Zephyr's libc which doesn't include those two functions :(

LaszloLango:
@sergioamr, the checker is always checking the last non-empty line of the commit message. That should me match to the author of the commit. If you added the commit and you are the author, then your Sign-off should be the last one.

targets/arduino_101/src/jerry-entry.c
```diff
@@ -0,0 +1,45 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdio.h>
+#include <stdarg.h>
+
+#include "jrt/jrt.h"
```
LaszloLango: IMHO, targets shouldn't include `jrt.h`. Please check `targets/default/jerry-port-default-io.c`. This file only needs two include files: `jerry-port.h` and `stdarg.h`

LaszloLango:
@sergioamr, thanks for the update. LGTM

targets/arduino_101/Makefile.arduino_101
```diff
@@ -0,0 +1,183 @@
...
+all: jerry zephyr
+
+$(EXTERNAL_LIB):
+ifdef V
+	@echo "- JERRY SCRIPT -------------------------------------------------"
+endif
+	mkdir -p $(INTERM)
+	mkdir -p $(OUTPUT)
+	cmake -B$(INTERM) -H./ \
+	 -DENABLE_LTO=OFF \
```
zherczeg: I just want to note that LTO provides a big improvement. At some point you should consider enabling it. All-in-one is also an alternative.

sergioamr: Sounds good to me!, I can add it as an optional parameter too.

targets/arduino_101/src/main-zephyr.c
```diff
@@ -0,0 +1,132 @@
...
+
+#define VERBOSE 0x01
+/**
+ * Jerryscript simple test loop
+ */
+int jerryscript_test() {
+    jerry_completion_code_t ret_code;
+    const char script[] =
+            "var test=0; " \
+            "for (t=100; t<1000; t++) test+=t; " \
```
zherczeg: t is intentionally not var? Can have side effects.

targets/arduino_101/src/main-zephyr.c
```diff
@@ -0,0 +1,132 @@
...
+        { "version", shell_cmd_version },
+        { "test", shell_cmd_test },
+        { "verbose", shell_cmd_verbose },
+        { NULL, NULL } };
+
+void main(void) {
+    printf("Jerry Compilation " __DATE__ " " __TIME__ "\n");
+    shell_register_app_cmd_handler(shell_cmd_handler);
+    shell_init("js> ", commands);
+}
```
zherczeg: This file does not follow the coding style, but I suspect that is intentional.

zherczeg:
In ginereal, the patch is go, so LGTM. I added some comments, but nothing is a requirement.

I would also note that we have also been thinking about how could we manage ports better: keep things in trunk, or a separate project for each port. Nothing is decided so far however. In the meantime we land everything to the trunk, and hoping that things does not go out of hand :)

targets/arduino_101/src/main-zephyr.c
sergioamr: This part goes inside Zephyr as an application, and I was not following the coding style for it.
Looking at the nuttx integration, i see that they also follow JS style and i will change this one to match it.

targets/arduino_101/src/main-zephyr.c
sergioamr: I will change it to "var" to be local in the loop. In some of my tests i was printing t after the loop to see how the global scope worked.

sergioamr:
I adjusted the style, but is this const struct supposed to look like this?
check-vera would fail otherwise.

+const struct shell_cmd commands[] =
 +{
 +  {
 +    "syntax", shell_cmd_syntax_help
 +  }
 +  ,
 +  {
 +    "version", shell_cmd_version
 +  }
 +  ,
 +  {
 +    "test", shell_cmd_test
 +  }
 +  ,
 +  {
 +    "verbose", shell_cmd_verbose
 +  }
 +  ,
 +  {
 +    NULL, NULL
 +  }
 +};

LaszloLango:
@sergioamr create a define for the array elements. Something like this:

```C++
#define SHELL_COMMAND(name,cmd) { name, cmd }

const struct shell_cmd commands[] =
{
  SHELL_COMMAND ("syntax", shell_cmd_syntax_help),
  SHELL_COMMAND ("version", shell_cmd_version),
  ...
}
```

targets/arduino_101/src/libc_support.c
poussa: I still don't fully understand why this file (`libc_support.c`) is included. In the `prj.conf` you are using the newlib libc library which it not the Zephyr minimal nor JerryScript libc. The newlib comes from the compilers that Zephyr supports.

sergioamr: Hi Sakari,

You were right about the libc_support.c. I didn’t need any of that to link. An update is on the pull request without any of the lib_support.c or the asm code.

Sergio

sergioamr:
@LaszloLango Thanks!

zherczeg:
Still LGTM.

poussa:
LGTM. IMO, we are ready to merge.

PR #1112, opened 2016-06-01
Optimize put result phase in the vm.

zherczeg:
Many opcodes has no results, or their results are always pushed onto the stack,
and checking their result type is wasting CPU cycles. Furthermore the
last_completion_value is removed which simplify this third phase.


zherczeg:
Separated into multiple patches.

PR #1116, opened 2016-06-01
Optimize LCache operation.

zherczeg:
The cache stores only real properties now, because storing NULLs has
little benefit according to tests. Since only real properties are
stored now, there is no need to create real references to objects
and property names, which reduces the keeping of dead objects after
garbage collection.

### DIFF
--- a/jerry-core/ecma/base/ecma-lcache.c
+++ b/jerry-core/ecma/base/ecma-lcache.c
@@ -44,19 +44,14 @@ typedef struct
 } ecma_lcache_hash_entry_t;
 
 /**
- * LCache hash value length, in bits
- */
-#define ECMA_LCACHE_HASH_BITS (sizeof (lit_string_hash_t) * JERRY_BITSINBYTE)
-
-/**
  * Number of rows in LCache's hash table
  */
-#define ECMA_LCACHE_HASH_ROWS_COUNT (1ull << ECMA_LCACHE_HASH_BITS)
+#define ECMA_LCACHE_HASH_ROWS_COUNT 256
 
 /**
  * Number of entries in a row of LCache's hash table
  */
-#define ECMA_LCACHE_HASH_ROW_LENGTH (2)
+#define ECMA_LCACHE_HASH_ROW_LENGTH 2
 
 /**
  * LCache's hash table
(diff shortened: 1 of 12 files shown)


zherczeg:
Benchmark | RSS (bytes) | Perf (sec) |
----: | ----: | ----: | 
3d-cube.js | 73728 -> 73728 : 0.000% | 1.769 -> 1.486 : +16.018% | 
3d-raytrace.js | 225280 -> 217088 : +3.636% | 1.987 -> 1.522 : +23.385% | 
access-binary-trees.js | 77824 -> 77824 : 0.000% | 0.860 -> 0.765 : +11.047% | 
access-fannkuch.js | 32768 -> 32768 : 0.000% | 3.785 -> 3.182 : +15.919% | 
access-nbody.js | 32768 -> 32768 : 0.000% | 2.109 -> 1.558 : +26.136% | 
bitops-3bit-bits-in-byte.js | 28672 -> 24576 : +14.286% | 0.828 -> 0.793 : +4.232% | 
bitops-bits-in-byte.js | 24576 -> 24576 : 0.000% | 1.096 -> 1.041 : +4.968% | 
bitops-bitwise-and.js | 28672 -> 24576 : +14.286% | 1.879 -> 1.935 : -2.983% | 
bitops-nsieve-bits.js | 167936 -> 159744 : +4.878% | 2.736 -> 2.501 : +8.583% | 
controlflow-recursive.js | 118784 -> 114688 : +3.448% | 0.553 -> 0.546 : +1.194% | 
crypto-aes.js | 110592 -> 98304 : +11.111% | 1.928 -> 1.470 : +23.762% | 
crypto-md5.js | 180224 -> 180224 : 0.000% | 0.985 -> 0.919 : +6.744% | 
crypto-sha1.js | 118784 -> 118784 : 0.000% | 0.916 -> 0.865 : +5.623% | 
date-format-tofte.js | 61440 -> 53248 : +13.333% | 1.437 -> 1.161 : +19.200% | 
date-format-xparb.js | 57344 -> 53248 : +7.143% | 0.653 -> 0.597 : +8.440% | 
math-cordic.js | 28672 -> 28672 : 0.000% | 2.076 -> 1.964 : +5.435% | 
math-partial-sums.js | 28672 -> 28672 : 0.000% | 1.122 -> 1.065 : +5.082% | 
math-spectral-norm.js | 32768 -> 32768 : 0.000% | 0.957 -> 0.884 : +7.644% | 
string-base64.js | 151552 -> 147456 : +2.703% | 4.923 -> 4.683 : +4.874% | 
string-fasta.js | 40960 -> 40960 : 0.000% | 2.357 -> 2.231 : +5.374% | 
Geometric mean: | +3.887% | +10.390% | 

Binary sizes (bytes)
a5d9701f60:167752
8a4f25f81a:167392

LaszloLango:
Impressive results :)

jerry-core/ecma/base/ecma-lcache.c
```diff
@@ -79,241 +79,175 @@ ecma_lcache_init (void)
...
+  jmem_cpointer_t object_cp;
+  ECMA_SET_NON_NULL_POINTER (object_cp, object_p);
 
-    bool is_cached = ecma_is_property_lcached (prop_p);
+  ecma_lcache_hash_entry_t *entry_p = ecma_lcache_hash_table[ecma_lcache_row_idx (object_cp, prop_name_p)];
 
-    if (!is_cached)
+  for (uint32_t entry_index = 0; entry_index < ECMA_LCACHE_HASH_ROW_LENGTH; entry_index++)
+  {
+    if (entry_p->object_cp == object_cp && entry_p->prop_p == prop_p)
```
LaszloLango: Is it possible to attach the same property to different objects? What I'm trying to say here is can the following code work here?
```C
if (entry_p->prop_p == prop_p)
{
  JERRY_ASSERT (entry_p->object_cp == object_cp);
```

zherczeg: No it is not possible. But we need to check that entry_p->object_cp cannot be not NULL. Hm, but this could be a good idea. Since we only cache valid properties now, we could move the NULL to the property pointer.

zherczeg: I was thinking about it, and moving the NULL to prop_p is inefficient because the lookup() function depends on the object_cp. Here a (entry_p->prop_p == prop_p) check is not enough, since a property can be added, removed, and added to another entry position. This could work though:

    if (entry_p->object_cp != ECMA_NULL_POINTER && entry_p->prop_p == prop_p)

But I think the difference is minor compared to the other one. I will try it though.

LaszloLango: @zherczeg, I see. Thanks for checking.

jerry-core/ecma/base/ecma-lcache.c
```diff
@@ -79,241 +74,169 @@ ecma_lcache_init (void)
...
-  for (uint32_t entry_index = 0; entry_index < ECMA_LCACHE_HASH_ROW_LENGTH; entry_index++)
-  {
-    if (ecma_lcache_hash_table[ row_index ][ entry_index ].object_cp == object_cp
-        && ecma_lcache_hash_table[ row_index ][ entry_index ].prop_p == property_p)
-    {
-      ecma_lcache_invalidate_entry (&ecma_lcache_hash_table[ row_index ][ entry_index ]);
-    }
-  }
-} /* ecma_lcache_invalidate_row_for_object_property_pair */
+  return (size_t) ((ecma_string_hash (prop_name_p) ^ object_cp) & (ECMA_LCACHE_HASH_ROWS_COUNT - 1));
```
LaszloLango: Please help me to understand this line. Maybe some comment would be good. :)

zherczeg: This is randomizing the hash with the object. Before this patch, objects with the same field names evicted each other:

    a[0] = 1; b[0] = 1; c[0] = 1; // c[0] evicted a[0] before. Not anymore

This happened frequently in access-fannkuch test. Now because the hash is xor-ed with the object, properties do not evict each other anymore.

jerry-core/vm/vm.c
```diff
@@ -65,30 +66,41 @@ static ecma_compiled_code_t *__program = NULL;
...
-                 bool is_strict) /**< strict mode */
+                 ecma_value_t property) /**< property name */
 {
   if (unlikely (ecma_is_value_undefined (object) || ecma_is_value_null (object)))
   {
     return ecma_raise_type_error (ECMA_ERR_MSG (""));
   }
 
-  ecma_value_t completion_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY);
+  ecma_value_t completion_value = ecma_op_to_string (property);
```
LaszloLango: IMHO `completion_value` is a bad variable name here. It does not say anything about what contains. I suggest `prop_name_comp_value` instead or something like that.

jerry-core/vm/vm.c
```diff
@@ -65,30 +66,41 @@ static ecma_compiled_code_t *__program = NULL;
...
+    {
+      completion_value = ecma_fast_copy_value (ecma_get_named_data_property_value (property_p));
 
-  ECMA_FINALIZE (property_val);
+      ecma_deref_ecma_string (property_name_p);
+      return completion_value;
+    }
+  }
+
+  completion_value = ecma_op_get_value_object_base (object, property_name_p);
```
LaszloLango: I see now. You reused the same variable, this is why it's name is so general. Not sure it's worth it, but it won't be an objection against the merge from my side. :)

dbatyai:
Could we use the cache in `vm_op_set_value` as well, for example when reassign a value to a property, or is something preventing this?

jerry-core/ecma/base/ecma-lcache.c
LaszloLango: That's tricky :) Great work!

LaszloLango:
LGTM

zherczeg:
Actually ecma_op_object_put in vm_op_set_value has been recently optimized and already using it. But we could probably do something here as well in a follow-up patch.

jerry-core/ecma/base/ecma-lcache.h
```diff
@@ -24,9 +24,8 @@
  */
 
 extern void ecma_lcache_init (void);
-extern void ecma_lcache_invalidate_all (void);
 extern void ecma_lcache_insert (ecma_object_t *, ecma_string_t *, ecma_property_t *);
-extern bool ecma_lcache_lookup (ecma_object_t *, const ecma_string_t *, ecma_property_t **);
+extern ecma_property_t * ecma_lcache_lookup (ecma_object_t *, const ecma_string_t *);
```
dbatyai: Extra space after asterix

dbatyai:
LGTM

PR #1117, opened 2016-06-01
API documentation update #2

ktorpi:
Function descriptions and examples are added to API page.

### DIFF
--- a/03.api.md
+++ b/03.api.md
@@ -358,51 +358,6 @@ jerry_api_gc ();
 ```
 
 
-# jerry_api_value_t
-
-**Summary**
-The data type represents any JavaScript value that can be sent to or received from the engine.
-
-Type of value is identified by `jerry_api_value_t::type`, and can be one of the following:
-
-- `JERRY_API_DATA_TYPE_VOID` - no return value
-- `JERRY_API_DATA_TYPE_UNDEFINED` - JavaScript undefined;
-- `JERRY_API_DATA_TYPE_NULL` - JavaScript null;
-- `JERRY_API_DATA_TYPE_BOOLEAN` - boolean;
-- `JERRY_API_DATA_TYPE_FLOAT32` - number;
-- `JERRY_API_DATA_TYPE_FLOAT64` - number;
-- `JERRY_API_DATA_TYPE_UINT32` - number;
-- `JERRY_API_DATA_TYPE_STRING` - string;
-- `JERRY_API_DATA_TYPE_OBJECT` - reference to JavaScript object.
-
-**Structure**
-
-```c
-typedef struct jerry_api_value_t
-{
-  jerry_api_data_type_t type;
-
-  union
-  {
-    bool v_bool;
-
-    float v_float32;
-    double v_float64;
-
-    uint32_t v_uint32;
-
-    jerry_api_string_t v_string;
-    jerry_api_object_t v_object;
-  } u;
-} jerry_api_value_t;
-```
-
-**See also**
-
-- [jerry_api_eval](#jerry_api_eval)
-- [jerry_api_call_function](#jerry_api_call_function)
-- [jerry_api_construct_object](#jerry_api_construct_object)
-
 # jerry_api_eval
 
 **Summary**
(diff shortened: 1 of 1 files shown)


03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+jerry_api_eval (const jerry_api_char_t *source_p,
+                size_t source_size,
+                bool is_direct,
+                bool is_strict,
+                jerry_api_value_t *retval_p);
+```
+
+- `source_p` - source code to evaluate;
+- `source_size` - length of the source code;
+- `is_direct` - whether to perform `eval` in "direct" mode (possible only if `eval` invocation is performed from native function, called from JavaScript);
```
zherczeg: No, this is incorrect. Direct mode means it is called as "eval" and not through some alias. In this case the context is the same as the caller context. I would not expose this flag, because it has no meaning when called from C. Actually it may messes an unexpected context.

03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+
+    float v_float32;
+    double v_float64;
+
+    uint32_t v_uint32;
+
+    jerry_api_string_t v_string;
+    jerry_api_object_t v_object;
+  } u;
+} jerry_api_value_t;
```
zherczeg: Do we need to expose this structure? I thought there are wrappers for this.

03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+# jerry_api_create_string_sz
+
+**Summary**
+
+Creates a string. The caller should release the string with [jerry_api_release_string](#jerry_api_release_string) when the value becomes unnecessary.
+
+**Prototype**
+```c
+jerry_api_string_t *
+jerry_api_create_string_sz (const jerry_api_char_t *v,
```
zherczeg: I would call this function as jerry_api_create_string.

03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+
+**Summary**
+
+Creates an error object. Caller should release the object with jerry_api_release_object, just when the value becomes unnecessary.
+
+**Prototype**
+
+```c
+jerry_api_object_t *
+jerry_api_create_error_sz (jerry_api_error_t error_type,
```
zherczeg: Ditto.

03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+**See also**
+
+- [jerry_api_create_error](#jerry_api_create_error)
+
+
+# jerry_api_get_boolean_value
+
+**Summary**
+
+Gets the raw bool value form a jerry_api_valu_t value.
```
zherczeg: jerry_api_value_t

03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+
+**Example**
+
+```c
+{
+  jerry_init (JERRY_FLAG_EMPTY);
+  jerry_api_value_t value;
+  ... // create or acquire value
+
+  if (jerry_api_value_is_boolean (&value))
```
zherczeg: Yes, we have these functions. I think we should not expose the union then. I mean we can expose it if necessary, but not document it, because it is internal, and can be changed anytime.

03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+
+```c
+{
+  jerry_init (JERRY_FLAG_EMPTY);
+  jerry_api_value_t value;
+  ... // create or acquire value
+
+  if (jerry_api_value_is_string (&value))
+  {
+    jerry_api_string_t *raw_value = jerry_api_get_string_value (&value);
```
zherczeg: Can we get the actual characters?

03.api.md
```diff
@@ -358,92 +358,973 @@ jerry_api_gc ();
...
+
+**Summary**
+
+Returns whether the given jerry_api_value_t is a void value.
+
+**Prototype**
+
+```c
+bool
+jerry_api_value_is_void (const jerry_api_value_t *value_p);
```
zherczeg: What is a void? There is no such ES5.1 data type only an unary operator.

03.api.md
LaszloLango: IMHO, yes we do. `jerry_api_value_t` is the wrapper of `ecma_value_t`. See the API examples. The user have to write code that access the members of an api value, so we need to document what can be stored in it. But I think it's already there with some comments. Isn't this a duplicate?

03.api.md
LaszloLango: `jerry_api_create_string` already exists

03.api.md
LaszloLango: ditto

03.api.md
LaszloLango: It's the same as `ECMA_SIMPLE_VALUE_EMPTY`, but in API level.

03.api.md
ktorpi: It is possible using function jerry_api_string_to_char_buffer.

zherczeg:
Ok, LGTM

LaszloLango:
LGTM

PR #1119, opened 2016-06-02
Modify the usage of ecma_string_to_utf8_string()

bzsolt:
Parts:
 * The functions return value mostly used in assertion, but this value
    is already checked by the function itself, therefore we can remove some
    local variables, which serves this purpose.
 * Remove the 'should use return value' requirement for this function
   and remove the attribute itself, because it became unused in the project.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -395,10 +395,8 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
     string_desc_p->refs_and_container = ECMA_STRING_CONTAINER_HEAP_ASCII_STRING | ECMA_STRING_REF_ONE;
     const size_t data_size = new_size;
     lit_utf8_byte_t *data_p = (lit_utf8_byte_t *) jmem_heap_alloc_block (data_size);
-    lit_utf8_size_t bytes_copied = ecma_string_to_utf8_string (string1_p, data_p, str1_size);
-    JERRY_ASSERT (bytes_copied == str1_size);
-    bytes_copied = ecma_string_to_utf8_string (string2_p, data_p + str1_size, str2_size);
-    JERRY_ASSERT (bytes_copied == str2_size);
+    ecma_string_to_utf8_bytes (string1_p, data_p, str1_size);
+    ecma_string_to_utf8_bytes (string2_p, data_p + str1_size, str2_size);
 
     string_desc_p->u.ascii_string.size = (uint16_t) new_size;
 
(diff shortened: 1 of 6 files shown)


jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -674,7 +671,7 @@ ecma_string_get_array_index (const ecma_string_t *str_p, /**< ecma-string */
  *
  * @return number of bytes, actually copied to the buffer.
  */
-lit_utf8_size_t __attr_return_value_should_be_checked___
+lit_utf8_size_t
```
zherczeg: Do we use the return value anywhere?

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -740,6 +737,8 @@ ecma_string_to_utf8_string (const ecma_string_t *string_desc_p, /**< ecma-string
   }
 
   JERRY_ASSERT (size <= buffer_size);
+  JERRY_ASSERT (size == ecma_string_get_size (string_desc_p));
```
zherczeg: I think this is checked at the beginning of the function.

Personally I would add another always inline function, which does not have a return value, and checks whether the size is the same as the string length, and use it whenever possible.

The name would be: ecma_string_to_utf8_bytes - it emphasises that we want the strinmg representation and nothing else
And the current function would be renamed to ecma_string_copy_to_utf8_buffer - it emphasises that we have a generic buffer where we combine strings

The first one would be an always inline function without return value, which calls the second. But before you do this change, I would like to hear the opinion of other reviewers.

LaszloLango: @zherczeg, I like your idea.

dbatyai: It would be useful to differentiate the use cases, +1.

bzsolt:
Thanks for the reviews, I've updated the PR.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -741,7 +738,24 @@ ecma_string_to_utf8_string (const ecma_string_t *string_desc_p, /**< ecma-string
...
+ * It is the caller's responsibility to make sure that the string fits in the buffer.
+ * Check if the size of the string is equal with the size of the buffer.
+ */
+void __attr_always_inline___
+ecma_string_to_utf8_bytes (const ecma_string_t *string_desc_p, /**< ecma-string descriptor */
+                           lit_utf8_byte_t *buffer_p, /**< destination buffer pointer
+                                                       * (can be NULL if buffer_size == 0) */
+                           lit_utf8_size_t buffer_size) /**< size of buffer */
+{
+  JERRY_ASSERT (ecma_string_get_size (string_desc_p) == buffer_size);
```
zherczeg: I think this assert is unnecessary, since the <= is checked by ecma_string_copy_to_utf8_buffer and size == buffer_size checks the equality.

zherczeg:
LGTM after that

zherczeg:
LGTM

LaszloLango:
LGTM

PR #1123, opened 2016-06-03
Remove const array and scalar variables from ecma_utf8_string_to_number

akosthekiss:
There is no need for putting `LIT_CHAR_*` constants in const
scalars or arrays, they can be used directly. (There is especially
no need for arrays of 10 elements if only 2 of them are used.)

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-conversion.c
+++ b/jerry-core/ecma/base/ecma-helpers-conversion.c
@@ -341,15 +341,6 @@ ecma_utf8_string_to_number (const lit_utf8_byte_t *str_p, /**< utf-8 string */
 {
   /* TODO: Check license issues */
 
-  const lit_utf8_byte_t dec_digits_range[10] = { '0', '9' };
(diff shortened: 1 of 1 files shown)


zherczeg:
The patch is good, but do we need to use type casting? Does the compiler complains?

LaszloLango:
LGTM, but I agree with @zherczeg. Remove the typecasts if they are not needed.

akosthekiss:
@zherczeg @LaszloLango I've removed the casts and rebased the patch to current master.

You were right, the casts were not needed -- which was a surprise to me. I'll apparently have to reiterate some parts of my C studies...

LaszloLango:
@akosthekiss, thanks for the update. LGTM

zherczeg:
This is the convert everything to "int" rule which is the most annoying rule of C

    short a = 6; a += 2; // Compiler warning. Only the ++ is not a warning...

I would not say it is illogical but definitely not handy.

LGTM

PR #1131, opened 2016-06-09
Introduce JERRY_UNUSED() instead of __attr_unused___

bzsolt:

### DIFF
--- a/jerry-core/jrt/jrt.h
+++ b/jerry-core/jrt/jrt.h
@@ -26,7 +26,6 @@
 /**
  * Attributes
  */
-#define __attr_unused___ __attribute__((unused))
 #define __noreturn __attribute__((noreturn))
 #define __attr_noinline___ __attribute__((noinline))
 #define __attr_return_value_should_be_checked___ __attribute__((warn_unused_result))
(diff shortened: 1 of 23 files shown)


jerry-core/jrt/jrt.h
```diff
@@ -81,6 +80,8 @@ extern void __noreturn jerry_unimplemented (const char *, const char *, const ch
 #define JERRY_ASSERT(x) do { if (false) { (void)(x); } } while (0)
 #endif /* !JERRY_NDEBUG */
 
+#define JERRY_UNUSED(x) (void)(x)
```
LaszloLango: `#define JERRY_UNUSED(x) ((void) (x))`

jerry-core/jrt/jrt.h
```diff
@@ -61,7 +60,7 @@
 #define JERRY_STATIC_ASSERT_GLUE(a, b, c) JERRY_STATIC_ASSERT_GLUE_ (a, b, c)
 #define JERRY_STATIC_ASSERT(x, msg) \
   typedef char JERRY_STATIC_ASSERT_GLUE (static_assertion_failed_, __LINE__, msg) \
-  [ (x) ? 1 : -1 ] __attr_unused___
+  [ (x) ? 1 : -1 ] __attribute__((unused))
```
LaszloLango: Why cannot use JERRY_UNUSED here?

bzsolt: I think it looks weird, so I kept the attribute here.
Do you have a sophisticated solution? :)

tests/unit/test-date-helpers.c
```diff
@@ -32,9 +32,10 @@
  * Unit test's main function.
  */
 int
-main (int __attr_unused___ argc,
-      char __attr_unused___ **argv)
+main (int argc, char **argv)
```
zherczeg: I think you can simply delete both arguments, and just use:

    main ()
    {
    }

Several similar cases below as well.

zherczeg:
I like this patch and I think it is close to ready

LaszloLango:
LGTM

jerry-core/jrt/jrt.h
```diff
@@ -60,8 +59,7 @@
 #define JERRY_STATIC_ASSERT_GLUE_(a, b, c) a ## b ## c
```
zherczeg: Wouldn't it be better to add an extra underscore (_) ?

    #define JERRY_STATIC_ASSERT_GLUE_(a, b, c) a ## b ## _ ## c

bzsolt: That's a good idea, it would be more readable.

zherczeg:
LGTM

PR #1138, opened 2016-06-13
Use 'ecma_value_t' in Jerry API

LaszloLango:

### DIFF
--- a/jerry-core/jerry-api.h
+++ b/jerry-core/jerry-api.h
@@ -43,22 +43,6 @@ typedef enum
 } jerry_completion_code_t;
 
 /**
- * Jerry API data types
- */
-typedef enum
-{
-  JERRY_DATA_TYPE_VOID,      /**< no return value */
-  JERRY_DATA_TYPE_UNDEFINED, /**< undefined */
-  JERRY_DATA_TYPE_NULL,      /**< null */
-  JERRY_DATA_TYPE_BOOLEAN,   /**< bool */
-  JERRY_DATA_TYPE_FLOAT32,   /**< 32-bit float */
-  JERRY_DATA_TYPE_FLOAT64,   /**< 64-bit float */
-  JERRY_DATA_TYPE_UINT32,    /**< number converted to 32-bit unsigned integer */
-  JERRY_DATA_TYPE_STRING,    /**< string */
-  JERRY_DATA_TYPE_OBJECT     /**< object */
-} jerry_data_type_t;
-
-/**
  * Jerry API Error object types
  */
 typedef enum
@@ -103,32 +87,15 @@ typedef struct ecma_string_t jerry_string_t;
 typedef struct ecma_object_t jerry_object_t;
 
 /**
- * Description of an extension function's argument
+ * Description of a JerryScript value
  */
-typedef struct jerry_value_t
-{
-  jerry_data_type_t type; /**< argument data type */
-
-  union
-  {
-    bool v_bool; /**< boolean */
-
-    float v_float32; /**< 32-bit float */
-    double v_float64; /**< 64-bit float */
-
-    uint32_t v_uint32; /**< number converted 32-bit unsigned integer */
-
-    jerry_string_t *v_string; /**< pointer to a JS string */
-    jerry_object_t *v_object; /**< pointer to a JS object */
-
-  } u;
-} jerry_value_t;
+typedef uint32_t jerry_value_t;
 
 /**
  * Type of an external function handler
  */
 typedef bool (*jerry_external_handler_t) (const jerry_object_t *function_obj_p,
-                                          const jerry_value_t *this_p,
+                                          const jerry_value_t this_val,
                                           jerry_value_t *ret_val_p,
                                           const jerry_value_t args_p[],
                                           const jerry_length_t args_count);
@@ -142,7 +109,7 @@ typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
  * Function type applied for each fields in objects
  */
 typedef bool (*jerry_object_field_foreach_t) (const jerry_string_t *field_name_p,
-                                              const jerry_value_t *field_value_p,
+                                              const jerry_value_t field_value_p,
                                               void *user_data_p);
 
 /**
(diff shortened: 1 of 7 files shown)


jerry-core/jerry-api.h
```diff
@@ -153,52 +120,53 @@ jerry_object_t *jerry_get_global (void);
...
+bool jerry_get_boolean_value (const jerry_value_t);
+double jerry_get_number_value (const jerry_value_t);
+jerry_string_t *jerry_get_string_value (const jerry_value_t);
+jerry_object_t *jerry_get_object_value (const jerry_value_t);
 
 /**
  * Converters of 'jerry_value_t'
  */
-jerry_string_t *jerry_value_to_string (const jerry_value_t *);
+jerry_value_t jerry_value_to_string (const jerry_value_t value);
```
zherczeg: "value" is not needed.

jerry-core/jerry-api.h
```diff
@@ -214,7 +182,7 @@ jerry_string_t *jerry_create_string_sz (const jerry_char_t *, jerry_size_t);
 /**
  * Functions of array objects
  */
-bool jerry_set_array_index_value (jerry_object_t *, jerry_length_t, jerry_value_t *);
+bool jerry_set_array_index_value (jerry_object_t *, jerry_length_t, jerry_value_t);
 bool jerry_get_array_index_value (jerry_object_t *, jerry_length_t, jerry_value_t *);
```
zherczeg: Why the get needs jerry_value_t *?

jerry-core/jerry.c
```diff
@@ -278,9 +256,7 @@ jerry_get_object_value (const jerry_value_t *value_p) /**< pointer to api value
 jerry_value_t
 jerry_create_void_value (void)
 {
-  jerry_value_t jerry_val;
-  jerry_val.type = JERRY_DATA_TYPE_VOID;
-  return jerry_val;
+  return ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY);
```
zherczeg: void is UNDEFINED not EMPTY. Empty must not be used externally.

zherczeg: There is a jerry_create_undefined_value. Perhaps this function could be deleted. There is "void" type anyway.

jerry-core/jerry.c
```diff
@@ -620,24 +439,20 @@ jerry_release_object (jerry_object_t *object_p) /**< pointer acquired through je
  *
  * Warning:
  *         Acquired pointer should be released with jerry_release_value
- *
- * @return pointer that may be used outside of the engine
  */
-jerry_value_t *
-jerry_acquire_value (jerry_value_t *value_p) /**< API value */
+void
```
zherczeg: This is incorrect. For strings it might return with a new instance. Although I don't like that idea, because the same string might be copied several times.

jerry-core/jerry-api.h
LaszloLango: That is an output parameter here.

jerry-core/jerry.c
```diff
@@ -774,17 +579,20 @@ jerry_set_array_index_value (jerry_object_t *array_obj_p, /**< array object */
...
 
-  return true;
+  *value_p = get_completion;
+
+  if (!ecma_is_value_error (get_completion))
+  {
+    return true;
+  }
+
+  return false;
```
zherczeg: return !ecma_is_value_error (get_completion);

zherczeg: Btw I am not sure people expects that the value stores the exception. Perhaps we should free the exception and return with undefined if unsuccessful.

jerry-core/jerry.c
```diff
@@ -914,62 +722,25 @@ ecma_value_t
...
+  if (!is_successful)
   {
-    jerry_release_value (&api_arg_values[i]);
+    ret_value = ecma_make_error_value (ret_value);
   }
 
-  JMEM_FINALIZE_LOCAL_ARRAY (api_arg_values);
-
-  return completion_value;
+  return ret_value;
```
zherczeg: This is a different interface compared to the arraz getter. Might be confusing.

tests/unit/test-api.c
```diff
@@ -290,7 +237,10 @@ static bool foreach (const jerry_string_t *name,
 
 #define UNUSED(x) (void)(x)
 
-static bool foreach_exception (const jerry_string_t *name, const jerry_value_t *value, void * user_data)
+static bool
+foreach_exception (const jerry_string_t *name,
```
zherczeg: No comments for these arguments. (Not just here)

jerry-core/jerry.c
LaszloLango: I thought to the same, but not sure what would be the best.

zherczeg:
I think the general direction is very good. However, I feel people will not understand how the exception handling works in Jerry. Namely a value has an exception bit, and a value with an exception bit set cannot be passed to any function as input argument.

I am open to other ideas, but I see two directions at this moment:

- all functions should return with a bool (exception or not), and a value as an output argument. The output contains the exception value in case of exception, but the exception bit is cleared, so they can pass it to other functions.
- all functions return with a jerry value, where the exception bit can be set. There is an API function to check exceptions. For all input arguments the exception bit is cleared manually (so people cannot accidentally pass exception objects as values when setting a field of an object).

jerry-core/jerry.c
LaszloLango: You are right. but the previous implementation is wrong too, because it does not handle the the return value of `jerry_acquire_string`.

LaszloLango:
@zherczeg, I prefer the second, but maybe the first one would be better for newcomers.

zherczeg:
The second is obviously simpler and faster, but it might be confusing. What do others think?

dbatyai:
In my opinion going with the simpler solution is always the best choice, and would also behave similarly to how the engine currently works. The first approach would be the complete opposite, which, the way I see it, could create more confusion than the second one.

LaszloLango:
@zherczeg, I raised an issue for this problem. I think we can do it in a follow up patch, because this PR is big enough and barely related to the problem.

zherczeg:
Ok, LGTM

dbatyai:
LGTM

PR #1139, opened 2016-06-14
Add travis script to build and run jerryscript with zephyr

Akromx16:
Travis will build jerryscript against zephyr and run the "test" command in the virtual console, using named pipe. The output is directed to log.txt which is then checked to see if the command was successful.

JerryScript-DCO-1.0-Signed-off-by: Adrian Moldovan adrian.moldovan@intel.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,19 @@ os: linux
 dist: trusty
 sudo: required
 
+addons:
+  apt:
+    packages:
+      - gcc
+      - gcc-multilib
+      - g++
+      - libc6-dev-i386
+      - g++-multilib
+      - python3-ply
+      - linux-libc-dev
+      - gcc-multilib
+      - build-essential
+
 before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tools/apt-get-install-deps.sh; fi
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -4,6 +4,19 @@ os: linux
...
+    packages:
+      - gcc
+      - gcc-multilib
+      - g++
+      - libc6-dev-i386
+      - g++-multilib
+      - python3-ply
+      - linux-libc-dev
+      - gcc-multilib
+      - build-essential
```
LaszloLango: Please, move these into a new file under the tools folder (`tools/apt-get-install-zephyr-deps.sh`). Similar to `tools/apt-get-install-qemu-arm.sh`. After that you can add it the the `before_install` step in `travis.yml`.

Makefile
```diff
@@ -343,6 +343,10 @@ clean:
 check-signed-off:
 	$(Q) ./tools/check-signed-off.sh
 
+.PHONY: check-signed-off
+build_zephyr:
```
LaszloLango: Use dash instead of underscore in target names.

Makefile
```diff
@@ -343,6 +343,10 @@ clean:
...
+	$(Q) ./tools/build_zephyr.sh
```
LaszloLango: also in filenames

PR #1144, opened 2016-06-14
Added travis script to build and run jerryscript with zephyr

Akromx16:
Travis will build jerryscript against zephyr and run the "test" command in the virtual console, using named pipe. The output is directed to log.txt which is then checked to see if the command was successfull.

JerryScript-DCO-1.0-Signed-off-by: Adrian Moldovan adrian.moldovan@intel.com

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,7 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tools/apt-get-install-deps.sh; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tools/apt-get-install-qemu-arm.sh; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then tools/brew-install-deps.sh; fi
+  - if [[ "$TARGET" == "build-zephyr" ]]; then tools/apt-get-install-zephyr-deps.sh; fi
 
 install:
 
@@ -26,5 +27,8 @@ matrix:
       env: TARGET="build.darwin test-js"
     - os: osx
       env: TARGET=test-unit
+    - os: linux
+      env: TARGET=build-zephyr
   allow_failures:
     - os: osx
+    - env: TARGET=build-zephyr
--- a/Makefile
+++ b/Makefile
@@ -343,6 +343,10 @@ clean:
 check-signed-off:
 	$(Q) ./tools/check-signed-off.sh
 
+.PHONY: build-zephyr
+build-zephyr:
+	$(Q) ./tools/build-zephyr.sh
+
 .PHONY: check-vera
 check-vera:
 	$(Q) $(call SHLOG,./tools/check-vera.sh,$(OUT_DIR)/vera.log,Vera++)
--- a/tools/build-zephyr.sh
+++ b/tools/build-zephyr.sh
@@ -1,0 +1,82 @@
+#!/bin/bash
+
+# Copyright © 2016 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# clone zephyr repo
+git clone https://gerrit.zephyrproject.org/r/zephyr
+cd zephyr
+git checkout tags/v1.4.0
+
+# download the zephyr_sdk installer script
+wget https://nexus.zephyrproject.org/content/repositories/releases/org/\
+zephyrproject/zephyr-sdk/0.8-i686/zephyr-sdk-0.8-i686-setup.run
+
+# extract the zephyr-sdk installer
+chmod +x zephyr-sdk-0.8-i686-setup.run
+mkdir zephyr_sdk_installer && cd zephyr_sdk_installer
+../zephyr-sdk-0.8-i686-setup.run --target $(pwd) --noexec
+
+# install the zephyr_sdk
+mkdir ../zephyr_sdk && cd ../zephyr_sdk
+export ZEPHYR_SDK_INSTALL_DIR=$(pwd)
+cd ../zephyr_sdk_installer
+./setup.sh -d $ZEPHYR_SDK_INSTALL_DIR
+
+# source the zephyr required environment variables
+source ../zephyr-env.sh
+export ZEPHYR_GCC_VARIANT=zephyr
+
+# back to jerryscript dir
+cd ../..
+
+# link asm-generic to asm (otherwise there will be errors with include headers)
+sudo ln -s /usr/include/asm-generic /usr/include/asm
+
+# build jerryscript for qemu
+make -f ./targets/arduino_101/Makefile.arduino_101 BOARD=qemu_x86
+
+# create 2 named pipes to input jerryscript commands and evaluate the output
+mkfifo path.in path.out
+
+# run the qemu executable in background using the named pipes
+./zephyr/zephyr_sdk/sysroots/i686-pokysdk-linux/usr/bin/qemu-system-i386 -m 32 \
+    -cpu qemu32 -no-reboot -nographic -vga none -display none -net none \
+    -clock dynticks -no-acpi -balloon none \
+    -L ./zephyr/zephyr_sdk/sysroots/i686-pokysdk-linux/usr/share/qemu \
+    -bios bios.bin -machine type=pc-0.14 -serial pipe:path \
+    -kernel ./build/qemu_x86/zephyr/zephyr.elf & sleep 5
+
+# listen for any output and direct it to log for further checking
+cat path.out > log.txt & sleep 5
+
+# write the jerryscript "test" command in the input named pipe
+printf "test\r\n" > path.in
+sleep 5
+
+# kill qemu executable process (which is marked as first job that runs in background)
+kill %1
+
+# show the log output and check if the string "Hi JS World!" is present there
+cat log.txt
+if grep "Hi JS World!" log.txt > /dev/null
+then
+    echo -e "\n\nQapla'!    (it means \"Success\" in Klingon)"
+else
+    echo -e "\n\nScript has failed"
+fi
+
+# run the grep command again, so that in the case of not finding the string, the
+# grep command will return a non 0 value, which will make travis fail the build
+grep "Hi JS World!" log.txt > /dev/null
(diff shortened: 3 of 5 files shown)


tools/build-zephyr.sh
```diff
@@ -0,0 +1,40 @@
...
+make -f ./targets/arduino_101/Makefile.arduino_101 BOARD=qemu_x86
+mkfifo path.in path.out
+./zephyr/zephyr_sdk/sysroots/i686-pokysdk-linux/usr/bin/qemu-system-i386 -m 32 -cpu qemu32 -no-reboot -nographic -vga none -display none -net none -clock dynticks -no-acpi -balloon none -L ./zephyr/zephyr_sdk/sysroots/i686-pokysdk-linux/usr/share/qemu -bios bios.bin -machine type=pc-0.14 -pidfile qemu.pid -serial pipe:path -kernel ./build/qemu_x86/zephyr/zephyr.elf & sleep 5
+cat path.out > log.txt & sleep 5
+printf "test\r\n" > path.in
+sleep 5
+kill %1
+cat log.txt
+if grep "Hi JS World!" log.txt > /dev/null; then echo -e "\n\nQapla'!    (it means \"Success\" in Klingon)"; else echo -e "\n\nScript has failed"; fi
+grep "Hi JS World!" log.txt > /dev/null
```
LaszloLango: Please make this file more readable. Add some comment and empty line, break long lines where you can, etc. I'd rename this file, because it not just builds zephyr (ex.: `build-and-test-zephyr-target.sh`).

LaszloLango:
Nice work! I think we can merge it soon.

tools/build-zephyr.sh
```diff
@@ -0,0 +1,76 @@
...
+
+# build jerryscript for qemu
+make -f ./targets/arduino_101/Makefile.arduino_101 BOARD=qemu_x86
+
+# create 2 named pipes to input jerryscript commands and evaluate the output
+mkfifo path.in path.out
+
+# run the qemu executable in background using the named pipes
+./zephyr/zephyr_sdk/sysroots/i686-pokysdk-linux/usr/bin/qemu-system-i386 -m 32 \
+-cpu qemu32 -no-reboot -nographic -vga none -display none -net none -clock dynticks \
```
LaszloLango: I bit nitpicky, but please indent these lines. It makes it more readable.

tools/build-zephyr.sh
```diff
@@ -0,0 +1,76 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# clone zephyr repo
+git clone https://gerrit.zephyrproject.org/r/zephyr
+cd zephyr
+
+# download the zephyr_sdk installer script
+wget https://nexus.zephyrproject.org/content/repositories/releases/org/\
```
LaszloLango: ditto
```
wget https://nexus.zephyrproject.org/content/repositories/releases/org/ \
    zephyrproject/zephyr-sdk/0.8-i686/zephyr-sdk-0.8-i686-setup.run
```

tools/build-zephyr.sh
```diff
@@ -0,0 +1,76 @@
...
+printf "test\r\n" > path.in
+sleep 5
+
+# kill qemu executable process (which is marked as first job that runs in background)
+kill %1
+
+# show the log output and check if the string "Hi JS World!" is present there
+cat log.txt
+if grep "Hi JS World!" log.txt > /dev/null; then echo -e \
+"\n\nQapla'!    (it means \"Success\" in Klingon)"; else echo -e "\n\nScript has failed"; fi
```
LaszloLango:
```bash
if [ grep "Hi JS World!" log.txt > /dev/null ]
then
    echo -e "\n\nQapla'!    (it means \"Success\" in Klingon)"
else
    echo -e "\n\nScript has failed"
fi
```

Makefile
```diff
@@ -343,6 +343,10 @@ clean:
 check-signed-off:
 	$(Q) ./tools/check-signed-off.sh
 
+.PHONY: check-signed-off
+build-zephyr:
```
LaszloLango: I'm still confused here. Is this only build or test it too? Release or debug?

.travis.yml
```diff
@@ -8,6 +8,7 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tools/apt-get-install-deps.sh; fi
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then tools/apt-get-install-qemu-arm.sh; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then tools/brew-install-deps.sh; fi
+  - if [[ "$TARGET" == "build-zephyr" ]]; then tools/apt-get-install-zephyr-deps.sh; fi
```
zherczeg: Don't we need a $TRAVIS_OS_NAME here?

Akromx16:
Thank you for the feedback! I didn't know the PR gets updated automatically once I push to my branch, sometimes I just wanted to push so travis could build to check if it works, without having the final release ready. I believe I should close the PR now and reopen it once I finish completly.

LaszloLango:
@Akromx16, if you want to test the travis build, then you can enable travis for your own fork too. So when you push a new commit to any of your branches, it will test it. You don't have to open a PR for that. But I think it is ok to keep this open and force update it when you are ready.

.travis.yml
Akromx16: The 3rd build fails (the one with toolchain) if I choose TRAVIS_OS_NAME == linux. I think there are some conflicts in the packages installed.

Travis error is:
CMake Error at CMakeLists.txt:16 (project):
  The CMAKE_C_COMPILER:
    arm-linux-gnueabihf-gcc
  is not a full path and was not found in the PATH.
  Tell CMake where to find the compiler by setting either the environment
  variable "CC" or the CMake cache entry CMAKE_C_COMPILER to the full path to
  the compiler, or to the compiler name if it is in the PATH.

Maybe I can change the order of the dependencies scripts...so that I call the install-zephyr.sh before all of them, perhaps this could fix the conflict....Although I thought it was better for that dependency script to be called only for the build in which it's actually building and running the zephyr with jerryscript.

Akromx16:
Yes, that is what I am doing, I am pushing on my own fork, it's just that the PR gets updated everytime I push, (even if sometimes it was for testing travis purposes, and not a final release).

tools/build-zephyr.sh
Akromx16: that wouldn't work.  I tried looking into ways of splitting that as well as indenting, but couldn't find a nice looking one in bash...i am splitting it but i cannot indent it as well (afaik)

Akromx16:
@LaszloLango @zherczeg I have my PR finished, with the changes that I could implement. There were some things that broke and I spend quite some time finding out what the problems were. I believe the zephyr sdk changed as well and caused my build to break (i wasn't changing the branch before so implicitly it was going to master). Thanks for all the feedback, and I will make more changes if there are needed.

Makefile
Akromx16: I am not sure what I should put there. I just follosed the example of check-signed-off and did the same with build-zephyr. It basically builds and runs (to test) it.

tools/build-zephyr.sh
LaszloLango: I see. If the whitespaces breaks the `wget` command then it is OK to write it in one long line. :)

zherczeg:
LGTM However, we might need help for fixing zephyr failures.

akosthekiss:
Sorry for joining the conversation a bit late, I've been OoO for a longer period. I've objections, which I'd like to discuss / get answer for:
- The tools/build.sh is quite monolithic now. It does prerequisites/dependencies installation, building, and testing in one. This is no good. Those deps which can be instaled via system package managers are already handled OK (tools/apt-get-install-*) but those that need other handling should also be separated. The key point is that this "other" install script should not only be useful for Travis CI but developers as well. And devs don't want to run installation at every build. The tools/prerequisites.sh can be a template. Right now, that's used for the setup of STM32 board dependencies. So, please create something similar for zephyr as well.
- There is an "interesting" line in build-zephyr.sh: `sudo ln -s /usr/include/asm-generic /usr/include/asm`. Everything else works in the local directory structure but this one is disturbingly global. Why is this needed? Is this needed on all linux variants? I'd be reloctant to fiddle with my global include dir structure.
- Where are the deps (e.g., zephyr repo, sdk) installed? They shouldn't pollute the main directory of jerry. Other targets (STM32, again) install their deps into the third-party directory. Please, follow that pattern.
- The testing should not happen in the build script either. Moreover: why is it not possible to run the official tests for checking the correctness of the build?
- The make target pattern proposed in this PR does not align with the targets already used. `build.xxx` normally builds all variants (debug and release, all profile variants, etc) for that "xxx" target. And there are targets for all these variants separately as well. `build-zephyr` doesn't really fit this model.

sergioamr:
@akosthekiss  @Akromx16  will go back to university soon and i am not sure he will be able to finish this PR with all those changes. Is it something you could help him with so we get this integration finish? :dancers:

pfalcon:
> Moreover: why is it not possible to run the official tests for checking the correctness of the build?

This is good question I myself would like to find answer to - does jerryscript has any infrastructure to run its testsuite on an *embedded target*? I posted that question to the recently created mailing list: https://mail.gna.org/public/jerryscript-dev/2016-07/msg00000.html . @akosthekiss: If you could shed some light on that, it would be appreciated!


https://mail.gna.org/public/jerryscript-dev/2016-07/msg00000.html

pfalcon:
@sergioamr, @Akromx16 : If nobody else beats me to it, then it's in my backlog ;-). But given concerns raised above, I guess I'd take this PR as a known-working reference, but would try to re-implement it myself trying to make as little changes to existing jerryscript CI structure as possible.

My biggest concern however is it may be a bit heavy "gift" fro jerryscript project - majority of runtime here will (likely) be downloading of Zephyr dependencies, and that may skew JerryScript CI responsiveness. And of course, these dependencies need maintenance (update Zephyr SDK downlaod link, etc.) I'd personally be a bit shy to propose such addition right away, how I do it for the other projects is to set up a separate fork and update it manually. If it proves useful and maintained over some time (couple of months, half-year), then it's a good argumentation to burden upstream with it ;-).

sergioamr:
@pfalcon That is actually a very good suggestion, i didn't think on the heavy work that updating sdks and keeping it current could be for the Jerryscript project. For me to create a new repository with a new open project would also involve to go through a heavy internal process of open sourcing. That would be more work than actually keeping this code running :D

LaszloLango:
I didn't get any mail from the mailing list, but I'm subscribed. Is it just me?

pfalcon:
> For me to create a new repository with a new open project would also involve to go through a heavy internal process of open sourcing.

Fortunately, on github, it's just a click of a "Fork" button, the only trouble then is to make sure to update to upstream regularly (and that's when you push new commits to your fork is when CI on it triggers).

pfalcon:
> I didn't get any mail from the mailing list, but I'm subscribed. Is it just me?

Just in case, I sent it last night (EU time). And according to the archive, it's a second message in the ML, after "hello" one last month. So, maybe check spam folder...

Issue #1141, opened 2016-06-14
Fix error handling in API


LaszloLango:
Details: https://github.com/Samsung/jerryscript/pull/1138#issuecomment-225802981

> 
> I think the general direction is very good. However, I feel people will not understand how the exception handling works in Jerry. Namely a value has an exception bit, and a value with an exception bit set cannot be passed to any function as input argument.
> 
> I am open to other ideas, but I see two directions at this moment:
> 
> - all functions should return with a bool (exception or not), and a value as an output argument. The output contains the exception value in case of exception, but the exception bit is cleared, so they can pass it to other functions.
> - all functions return with a jerry value, where the exception bit can be set. There is an API function to check exceptions. For all input arguments the exception bit is cleared manually (so people cannot accidentally pass exception objects as values when setting a field of an object).

PR #1147, opened 2016-06-15
Update jerry API

LaszloLango:
* Fix error handling (related issue: #1141)
 * Move output paramters to the end of the arguments lists

### DIFF
--- a/jerry-core/jerry-api.h
+++ b/jerry-core/jerry-api.h
@@ -96,9 +96,9 @@ typedef uint32_t jerry_value_t;
  */
 typedef bool (*jerry_external_handler_t) (const jerry_object_t *function_obj_p,
                                           const jerry_value_t this_val,
-                                          jerry_value_t *ret_val_p,
                                           const jerry_value_t args_p[],
-                                          const jerry_length_t args_count);
+                                          const jerry_length_t args_count,
+                                          jerry_value_t *ret_val_p);
 
 /**
  * Native free callback of an object
@@ -197,15 +197,15 @@ bool jerry_is_constructor (const jerry_object_t *);
 bool jerry_is_function (const jerry_object_t *);
 bool jerry_add_object_field (jerry_object_t *, const jerry_char_t *, jerry_size_t, const jerry_value_t, bool);
 bool jerry_delete_object_field (jerry_object_t *, const jerry_char_t *, jerry_size_t);
-bool jerry_get_object_field_value (jerry_object_t *, const jerry_char_t *, jerry_value_t *);
-bool jerry_get_object_field_value_sz (jerry_object_t *, const jerry_char_t *, jerry_size_t, jerry_value_t *);
+jerry_value_t jerry_get_object_field_value (jerry_object_t *, const jerry_char_t *);
+jerry_value_t jerry_get_object_field_value_sz (jerry_object_t *, const jerry_char_t *, jerry_size_t);
 bool jerry_set_object_field_value (jerry_object_t *, const jerry_char_t *, const jerry_value_t);
 bool jerry_set_object_field_value_sz (jerry_object_t *, const jerry_char_t *, jerry_size_t, const jerry_value_t);
 bool jerry_foreach_object_field (jerry_object_t *, jerry_object_field_foreach_t, void *);
 bool jerry_get_object_native_handle (jerry_object_t *, uintptr_t *);
 void jerry_set_object_native_handle (jerry_object_t *, uintptr_t, jerry_object_free_callback_t);
-bool jerry_construct_object (jerry_object_t *, jerry_value_t *, const jerry_value_t[], uint16_t);
-bool jerry_call_function (jerry_object_t *, jerry_object_t *, jerry_value_t *, const jerry_value_t[], uint16_t);
+jerry_value_t jerry_construct_object (jerry_object_t *, const jerry_value_t[], uint16_t);
+jerry_value_t jerry_call_function (jerry_object_t *, jerry_object_t *, const jerry_value_t[], uint16_t);
 
 /**
  * @}
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -125,6 +125,15 @@ jerry_make_api_unavailable (void)
 } /* jerry_make_api_unavailable */
 
 /**
+ * Returns whether the given jerry_value_t is error.
+ */
+bool
+jerry_value_is_error (const jerry_value_t value) /**< api value */
+{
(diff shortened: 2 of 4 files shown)


jerry-core/jerry.c
```diff
@@ -125,9 +125,18 @@ jerry_make_api_unavailable (void)
...
+jerry_value_is_error (const jerry_value_t value) /**< api value */
+{
+  return ECMA_IS_VALUE_ERROR (value);
+} /* jerry_value_is_error */
+
+/**
  * Returns whether the given jerry_value_t is null.
  */
-bool
+inline bool __attr_always_inline___
```
zherczeg: I think this depends on LTO/all-in-one build, and an API user unlikely to have this advantage.

jerry-core/jerry-api.h
```diff
@@ -197,15 +197,15 @@ bool jerry_is_constructor (const jerry_object_t *);
...
+jerry_value_t jerry_get_object_field_value_sz (jerry_object_t *, const jerry_char_t *, jerry_size_t);
 bool jerry_set_object_field_value (jerry_object_t *, const jerry_char_t *, const jerry_value_t);
 bool jerry_set_object_field_value_sz (jerry_object_t *, const jerry_char_t *, jerry_size_t, const jerry_value_t);
 bool jerry_foreach_object_field (jerry_object_t *, jerry_object_field_foreach_t, void *);
 bool jerry_get_object_native_handle (jerry_object_t *, uintptr_t *);
 void jerry_set_object_native_handle (jerry_object_t *, uintptr_t, jerry_object_free_callback_t);
-bool jerry_construct_object (jerry_object_t *, jerry_value_t *, const jerry_value_t[], uint16_t);
-bool jerry_call_function (jerry_object_t *, jerry_object_t *, jerry_value_t *, const jerry_value_t[], uint16_t);
+bool jerry_construct_object (jerry_object_t *, const jerry_value_t[], uint16_t, jerry_value_t *);
+bool jerry_call_function (jerry_object_t *, jerry_object_t *, const jerry_value_t[], uint16_t, jerry_value_t *);
```
zherczeg: Why don't these functions return with jerry_value_t?

LaszloLango: Was not sure about that, but I don't mind to change.

jerry-core/jerry.c
LaszloLango: You are right. This was not a good idea. I'll remove it.

LaszloLango:
@zherczeg, I've upadted the PR. Please check.

jerry-core/jerry.c
```diff
@@ -1098,11 +1100,9 @@ jerry_set_object_native_handle (jerry_object_t *object_p, /**< object to set han
  *      If function is invoked as constructor, it should support [[Construct]] method,
  *      otherwise, if function is simply called - it should support [[Call]] method.
  *
- * @return true, if invocation was performed successfully, i.e.:
- *                - no unhandled exceptions were thrown in connection with the call;
- *         false - otherwise.
+ * @return TODO
```
zherczeg: return todo?

jerry-core/jerry.c
```diff
@@ -1157,29 +1152,19 @@ jerry_invoke_function (bool is_invoke_as_constructor, /**< true - invoke functio
...
 /**
  * Construct new TypeError object
+ *
+ * @return TypeError object value
  */
-static void
-jerry_construct_type_error (jerry_value_t *retval_p) /**< [out] value with constructed
-                                                      *         TypeError object */
+static inline jerry_value_t __attr_always_inline___
+jerry_construct_type_error (void)
```
zherczeg: inline again. I think inlining is important, but it is unclear to me how can we take advantage of that, so I would do it as a separate step. Such change would not even hurt the api.

jerry-core/jerry.c
```diff
@@ -1189,20 +1174,12 @@ jerry_construct_type_error (jerry_value_t *retval_p) /**< [out] value with const
  *      returned value should be freed with jerry_release_value
  *      just when the value becomes unnecessary.
  *
- * @return true, if call was performed successfully, i.e.:
- *                - specified object is a function object (see also jerry_is_function);
- *                - no unhandled exceptions were thrown in connection with the call;
- *         false - otherwise, 'retval_p' contains thrown exception:
- *                  if called object is not function object - a TypeError instance;
- *                  else - exception, thrown during the function call.
+ * @return TODO
```
zherczeg: Fix all ot these please.

zherczeg:
The question is what should happen if we pass error objects as arguments is still open. We definitely need to do something with that before the release.

jerry-core/jerry.c
LaszloLango: sorry, I forgot to update the description.

jerry-core/jerry.c
LaszloLango: This is a static function. It cannot be used outside of `jerry.c`. I think this won't be a problem.

zherczeg:
LGTM

main-unix.c
```diff
@@ -461,12 +462,11 @@ main (int argc,
...
-        if (jerry_call_function (jerry_get_object_value (print_function),
-                                 NULL,
-                                 &ret_val_print,
-                                 args,
-                                 1))
+        jerry_value_t ret_val_print = jerry_call_function (jerry_get_object_value (print_function),
+                                                           NULL,
+                                                           args,
+                                                           1);
+        if (!jerry_value_is_error (ret_val_print))
```
dbatyai: `ret_val_print` should also be released when it's an error, otherwise it may leak.

jerry-core/jerry.c
```diff
@@ -946,35 +954,29 @@ jerry_foreach_object_field (jerry_object_t *object_p, /**< object */
  *      if value was retrieved successfully, it should be freed
  *      with jerry_release_value just when it becomes unnecessary.
  *
- * @return true, if field value was retrieved successfully, i.e. upon the call:
- *                - there is field with specified name in the object;
- *         false - otherwise.
+ * @return ecma value
```
dbatyai: In some places it's `ecma value` and in other places it's `jerry value`. I think the naming should be consistent.
Also, these `@return`s could be a little bit more descriptive in my opinion. I get that this is a simple getter function, but still.

LaszloLango:
@dbatyai, I've updated the PR. I will do a big review and update on the function comments in a later PR, but the comments of touched functions should be ready (more or less), so I hope I fixed all of them. :)

dbatyai:
LGTM

PR #1150, opened 2016-06-15
Add conversion API functions

LaszloLango:


zherczeg:
LGTM.

Deciding what should happen if an error object is an argument applies here as well.

dbatyai:
LGTM

Issue #1054, opened 2016-05-11
Assertion 's == 0' failed in ecma_number_to_utf8_string


renatahodovan:
###### Jerry version:
    Checked revision: 210ba6d
    Build: debug.linux
###### OS: 
    Ubuntu 15.10, x86_64
###### Test case:
```javascript
new String(Number.MAX_VALUE)
```

###### Backtrace:
```
ICE: Assertion 's == 0' failed at jerryscript/jerry-core/ecma/base/ecma-helpers-conversion.c(ecma_number_to_utf8_string):1273.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x000000000048575d in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:60
60    SYSCALL_2
(gdb) bt
# 0  0x000000000048575d in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:60
# 1  0x000000000040037f in raise (sig=6) at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:92
# 2  0x0000000000400351 in abort () at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:78
# 3  0x000000000041d3f1 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/targets/default/jerry-port-default-fatal.c:54
# 4  0x000000000041d032 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:68
# 5  0x000000000041d079 in jerry_assert_fail (assertion=0x49c4ec "s == 0", 
    file=0x49c2d0 "jerryscript/jerry-core/ecma/base/ecma-helpers-conversion.c", 
    function=0x48c5b0 <__func__.3096.lto_priv.420> "ecma_number_to_utf8_string", line=1273)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:95
# 6  0x000000000043ba34 in ecma_number_to_utf8_string (num=1.7976931348623157e+308, buffer_p=0x7fffffffc850 "8e+308", buffer_size=64)
    at jerryscript/jerry-core/ecma/base/ecma-helpers-conversion.c:1273
# 7  0x000000000043bf48 in ecma_new_ecma_string_from_number (num=1.7976931348623157e+308)
[...]

PR #1151, opened 2016-06-16
Fix `errol0_dtoa` for DBL_MAX

wateret:
When `val` is DBL_MAX, we get +inf from `ECMA_NEXT_FLOAT (val)`.
It is necessary to handle this case specially.

Related issue: #1054

JerryScript-DCO-1.0-Signed-off-by: Hanjoung Lee hanjoung.lee@samsung.com

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-errol.c
+++ b/jerry-core/ecma/base/ecma-helpers-errol.c
@@ -164,7 +164,13 @@ ecma_errol0_dtoa (ecma_number_t val, /**< ecma number */
   ecma_high_prec_t high_bound, low_bound;
 
   high_bound.value = mid.value;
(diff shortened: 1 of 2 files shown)


wateret:
I also checked the patch that handles same issue for `errol1_dtoa` - https://github.com/marcandrysco/Errol/commit/b000e1f597013ebdc03a7e3a269c36799765af60#diff-7d7958f2ccee37243fc551994b221d7eR156

But I thought we may not want to use `strcpy`.

LaszloLango:
LGTM, thanks for the fix.

zherczeg:
LGTM

LaszloLango:
@wateret, please rebase to the current master.

wateret:
@LaszloLango Done rebasing :D

PR #1156, opened 2016-06-17
Remove GCC 4.7 build requirement.

tilmannOSG:
The GCC 4.7 check is is a leftover from the times when JerryScript was a C++11 codebase, it was there to ensure that JerryScript is compiled with a GCC version which has sufficient support for C++11.

Today JerryScript is a pure C99 codebase (enforced by building all the sources with -std=c99) and this check is no longer necessary. JerryScript should now build just fine even with rather old versions of GCC.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,17 +26,6 @@ project (Jerry C ASM)
   endif()
  endif()
 
- if(CMAKE_COMPILER_IS_GNUCC)
-  # Require gcc of version >= 4.7.0
-   execute_process(COMMAND ${CMAKE_C_COMPILER} -dumpversion
-                   OUTPUT_VARIABLE GNU_CC_VERSION
-                   OUTPUT_STRIP_TRAILING_WHITESPACE)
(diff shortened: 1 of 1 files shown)


LaszloLango:
typo in commit message (double 'is')
> The GCC 4.7 check is is a ...

LGTM

zherczeg:
LGTM

galpeter:
Just a side note, with gcc 4.6 I'm getting the following error when running `make release.linux`:
```sh
cc1: error: unrecognized command line option ‘-Wpedantic’
cc1: error: -Werror=pedantic: no option -Wpedantic
cc1: error: unrecognized command line option ‘-fno-fat-lto-objects’

```

tilmannOSG:
@galpeter Thanks for heads-up! I think that's easy enough for someone to disable when actually building with an old version of GCC.

A major overhaul of the warning flags is planned with the build system renovation work that's going on right now. I'm hoping we can trim it down to a minimum set of flags which will generally make it easier to build JerryScript with non-GCC compilers implementing the GCC command line interface (e.g. Clang or Intel's ICC)

PR #1159, opened 2016-06-17
Build fix for ALL_IN_ONE and COMPILER_DEFAULT_LIBC

robertsipka:
The build fails with enabled ALL_ON_ONE and enabled COMPILER_DEFAULT_LIBC options,
because the _BSD_SOURCE doesn't defined before including sys/time.h.

In case of glibc, the 'timezone' is not part of C99 but part of POSIX, so we need
to define the _BSD_SOURCE feature test macro before including sys/time.h to get
the prototype for this struct.


LaszloLango:
LGTM

robertsipka:
Rebased with master

zherczeg:
LGTM

PR #1160, opened 2016-06-17
Update API reference

LaszloLango:

### DIFF
--- a/docs/API-REFERENCE.md
+++ b/docs/API-REFERENCE.md
@@ -1,1 +1,2640 @@
-See JerryScript API documentation on [JerryScript home page](https://samsung.github.io/jerryscript/API/).
+# jerry_run_simple
+
+**Summary**
+
+The simplest way to run JavaScript.
+
+**Prototype**
+
+```c
+jerry_completion_code_t
+jerry_run_simple (const jerry_char_t *script_source,
+                  size_t script_source_size,
+                  jerry_flag_t flags);
+```
+
+- `script_source` - source code, it must be a valid utf8 string.
+- `script_source_size` - size of source code buffer, in bytes.
+- return value - completion code, indicating whether execution was successful
+                 (`JERRY_COMPLETION_CODE_OK`), or an unhandled JavaScript exception occurred
+                 (`JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION`).
+
+**Example**
+
+```c
+{
+  const jerry_char_t *script = "print ('Hello, World!');";
+
+  jerry_run_simple (script, strlen ((const char *) script), JERRY_FLAG_EMPTY);
+}
+```
+
+**See also**
+
+- [jerry_init](#jerry_init)
+- [jerry_cleanup](#jerry_cleanup)
+- [jerry_parse](#jerry_parse)
+- [jerry_run](#jerry_run)
+
+
+# jerry_init
+
+**Summary**
+
+Initializes the JerryScript engine, making possible to run JavaScript code and perform operations on
+JavaScript values.
+
+**Prototype**
+
+```c
+void
+jerry_init (jerry_flag_t flags);
+```
+
+`flags` - combination of various engine configuration flags:
+
+- `JERRY_FLAG_EMPTY` - no flags, just initialize in default configuration.
+- `JERRY_FLAG_SHOW_OPCODES` - print compiled byte-code.
+- `JERRY_FLAG_MEM_STATS` - dump memory statistics.
+- `JERRY_FLAG_MEM_STATS_SEPARATE` - dump memory statistics and reset peak values after parse.
+- `JERRY_FLAG_ENABLE_LOG` - enable logging.
+
+**Example**
+
+```c
+{
+  jerry_init (JERRY_FLAG_SHOW_OPCODES | JERRY_FLAG_ENABLE_LOG);
+
+  // ...
+
+  jerry_cleanup ();
+}
+```
+
+**See also**
+
+- [jerry_cleanup](#jerry_cleanup)
+
+# jerry_cleanup
+
+**Summary**
+
+Finish JavaScript engine execution, freeing memory and JavaScript values.
+
+*Note*: JavaScript values, received from engine, will be inaccessible after the cleanup.
+
+**Prototype**
+
+```c
+void
+jerry_cleanup (void);
+```
+
+**See also**
+
+- [jerry_init](#jerry_init)
+
+
+# jerry_parse
+
+**Summary**
+
+Parse specified script to execute in Global scope.
+
+*Note*: Current API doesn't permit replacement or modification of Global scope's code without engine
+restart, which means 'jerry_parse' can be invoked only once between `jerry_init` and `jerry_cleanup`.
+
+**Prototype**
+
+```c
+bool
+jerry_parse (const jerry_char_t *source_p,
+             size_t source_size,
+             jerry_object_t **error_obj_p);
+```
+
+- `source_p` - string, containing source code to parse. It must be a valid utf8 string.
+- `source_size` - size of the string, in bytes.
+- `error_obj_p` - error object (output parameter).
+
+**Example**
+
+```c
+{
+  jerry_init (JERRY_FLAG_EMPTY);
+
+  char script [] = "print ('Hello, World!');";
+  size_t script_size = strlen ((const char *) script);
+
+  jerry_object_t *error_object_p = NULL;
+  if (!jerry_parse (script, script_size, &error_object_p))
+  {
+    /* Error object must be freed, if parsing failed */
+    jerry_release_object (error_object_p);
+  }
+
+  jerry_cleanup ();
+}
+```
+
+**See also**
+
+- [jerry_run](#jerry_run)
+
+
+# jerry_parse_and_save_snapshot
+
+**Summary**
+
+Generate snapshot from the specified source code.
+
+**Prototype**
+
+```c
+size_t
+jerry_parse_and_save_snapshot (const jerry_char_t *source_p,
+                               size_t source_size,
+                               bool is_for_global,
+                               uint8_t *buffer_p,
+                               size_t buffer_size);
+```
+
+- `source_p` - script source, it must be a valid utf8 string.
+- `source_size` - script source size, in bytes.
+- `is_for_global` - snapshot would be executed as global (true) or eval (false).
+- `buffer_p` - buffer to save snapshot to.
+- `buffer_size` - the buffer's size.
+- return value
+  - the size of snapshot, if it was generated succesfully (i.e. there are no syntax errors in source
+    code, buffer size is sufficient, and snapshot support is enabled in current configuration through
+    JERRY_ENABLE_SNAPSHOT)
+  - 0 otherwise.
+
+**Example**
+
+```c
+{
+  jerry_init (JERRY_FLAG_EMPTY);
+
+  static uint8_t global_mode_snapshot_buffer[1024];
+  const char *code_to_snapshot_p = "(function () { return 'string from snapshot'; }) ();";
+
+  size_t global_mode_snapshot_size = jerry_parse_and_save_snapshot ((jerry_char_t *) code_to_snapshot_p,
+                                                                    strlen (code_to_snapshot_p),
+                                                                    true,
+                                                                    global_mode_snapshot_buffer,
+                                                                    sizeof (global_mode_snapshot_buffer));
+}
+```
+
+**See also**
+
+- [jerry_init](#jerry_init)
+- [jerry_exec_snapshot](#jerry_exec_snapshot)
+
+
+# jerry_run
+
+**Summary**
+
+Run code of Global scope.
+
+*Note*: The code should be previously registered through `jerry_parse`.
+
+**Prototype**
+
+```c
+jerry_completion_code_t
+jerry_run (jerry_value_t *error_value_p);
+```
+
+- `error_value_p` - error value (output parameter).
+- return value - completion code that indicates whether run performed successfully.
+  - `JERRY_COMPLETION_CODE_OK` - successful completion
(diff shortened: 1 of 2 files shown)


docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+**Prototype**
+
+```c
+jerry_completion_code_t
+jerry_run_simple (const jerry_char_t *script_source,
+                  size_t script_source_size,
+                  jerry_flag_t flags);
+```
+
+- `script_source` - source code;
```
zherczeg: mention that it needs to be a valid utf8 string

docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+- `JERRY_FLAG_SHOW_OPCODES` - print compiled byte-code;
+- `JERRY_FLAG_MEM_STATS` - dump memory statistics;
+- `JERRY_FLAG_MEM_STATS_SEPARATE` - dump memory statistics and reset peak values after parse;
+- `JERRY_FLAG_ENABLE_LOG` - enable logging;
+
+**Example**
+
+```c
+{
+  jerry_init (JERRY_FLAG_ENABLE_LOG);
```
zherczeg: show a combination with |

docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+**Prototype**
+
+```c
+bool
+jerry_parse (const jerry_char_t *source_p,
+             size_t source_size,
+             jerry_object_t **error_obj_p);
+```
+
+- `source_p` - string, containing source code to parse;
```
zherczeg: utf8

docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+jerry_completion_code_t
+jerry_exec_snapshot (const void *snapshot_p,
+                     size_t snapshot_size,
+                     bool copy_bytecode,
+                     jerry_value_t *retval_p);
+```
+
+- `snapshot_p` - pointer to snapshot
+- `snapshot_size` - size of snapshot
+- `copy_bytecode` - flag, indicating whether the passed snapshot buffer should be copied to the engine's memory. If set the engine should not reference the buffer after the function returns (in this case, the passed buffer could be freed after the call). Otherwise (if the flag is not set) - the buffer could only be freed after the engine stops (i.e. after call to jerry_cleanup).
```
zherczeg: Single line? I think it is hard to read in a normal text editor

docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+
+**Example**
+
+```c
+{
+  jerry_value_t ret_val;
+
+  jerry_completion_code_t status = jerry_eval (str_to_eval,
+                                               strlen (str_to_eval),
+                                               false, false,
```
zherczeg: false in separate line

docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+
+**Summary**
+
+Creates a jerry_value_t with type JERRY_API_DATA_TYPE_VOID representing a void value. The value have to be released by [jerry_release_value](#jerry_release_value).
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_void_value (void);
```
zherczeg: This function is removed, isn't it?

docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+**See also**
+
+- [jerry_release_value](#jerry_release_value)
+
+
+# jerry_create_null_value
+
+**Summary**
+
+Creates a jerry_value_t with type JERRY_API_DATA_TYPE_NULL represneting a null value. The value have to be released by [jerry_release_value](#jerry_release_value).
```
zherczeg: representing

docs/API-REFERENCE.md
```diff
@@ -1 +1,2685 @@
...
+**See also**
+
+- [jerry_release_value](#jerry_release_value)
+
+
+# jerry_create_number_value
+
+**Summary**
+
+Creates a jerry_value_t from the given double parameter. The v_float64 member will be set and the type will be JERRY_API_DATA_TYPE_FLOAT64. The value have to be released by [jerry_release_value](#jerry_release_value).
```
zherczeg: v_float64 member will be ?

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+```c
+jerry_completion_code_t
+jerry_run_simple (const jerry_char_t *script_source,
+                  size_t script_source_size,
+                  jerry_flag_t flags);
+```
+
+- `script_source` - source code, it must be a valid utf8 string.
+- `script_source_size` - size of source code buffer, in bytes.
+- return value - completion code that indicates whether run was performed
```
zherczeg: "whether the"
and it seems the alignment is strange

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+- [jerry_cleanup](#jerry_cleanup)
+- [jerry_parse](#jerry_parse)
+- [jerry_run](#jerry_run)
+
+
+# jerry_init
+
+**Summary**
+
+Initializes JerryScript engine, making possible to run JavaScript code and perform operations on
```
zherczeg: "the JerryScript engine"

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+**See also**
+
+- [jerry_release_value](#jerry_release_value)
+
+
+# jerry_create_object_value
+
+**Summary**
+
+Creates a `jerry_value_t` from the given `jerry_object_t` parameter. The value has to be released
```
zherczeg: I think "convert" is a better word. This is kind of a type cast.

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+- [jerry_release_value](#jerry_release_value)
+
+
+# jerry_acquire_value
+
+**Summary**
+
+Acquires the specified Jerry API value.
+
+For values of string and object types this acquires the underlying data, for all other types it is
```
zherczeg: and number

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+- [jerry_release_string](#jerry_release_string)
+- [jerry_string_to_char_buffer](#jerry_string_to_char_buffer)
+
+
+# jerry_string_to_char_buffer
+
+**Summary**
+
+Copy string characters to specified buffer. It is the caller's responsibility to make sure that
+the string fits in the buffer.
```
zherczeg: I would mention that the result will not be zero terminated.

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+{
+  jerry_init (JERRY_FLAG_EMPTY);
+
+  jerry_char_t field_name[] = "field";
+  jerry_value_t field_value = jerry_create_number_value (3.14);
+
+  jerry_object_t* object_p = jerry_create_object ();
+  jerry_add_object_field (object_p,
+                          field_name,
+                          lit_zt_utf8_string_size (field_name),
```
zherczeg: Is this function exposed by the API? I think strlen is ok.

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+                          field_value, 
+                          true);
+
+  jerry_value_t retrieved_field_value;
+  retrieved_field_value = jerry_get_object_field_value_sz (object_p,
+                                                           field_name,
+                                                           lit_zt_utf8_string_size (field_name));
```
zherczeg: ditto.

zherczeg:
After the minor changes, LGTM.

docs/API-REFERENCE.md
LaszloLango: There is a note of that just below.

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+```c
+jerry_completion_code_t
+jerry_run_simple (const jerry_char_t *script_source,
+                  size_t script_source_size,
+                  jerry_flag_t flags);
+```
+
+- `script_source` - source code, it must be a valid utf8 string.
+- `script_source_size` - size of source code buffer, in bytes.
+- return value - completion code that indicates whether the run was performed successfully
```
dbatyai: `run was performed` feels a bit weird to me. Could we say something like `completion code, indicating whether execution was successful or not`?

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+
+
+# jerry_parse
+
+**Summary**
+
+Parse specified script to execute in Global scope.
+
+*Note*: Current API doesn't permit replacement or modification of Global scope's code without engine
+restart, so `jerry_parse` could be invoked only once between `jerry_init` and `jerry_cleanup`.
```
dbatyai: I would change this to `which means 'jerry_parse' can be invoked ... `

docs/API-REFERENCE.md
LaszloLango: Sure, I'll update them.

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+**See also**
+
+- [jerry_release_value](#jerry_release_value)
+
+
+# jerry_create_error_sz
+
+**Summary**
+
+Creates an error object. Caller should release the object with jerry_release_object, just when
```
dbatyai: There are multiple instances of `just when ...`, please change these to simply say `when ...`

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+
+**Prototype**
+
+```c
+jerry_object_t *
+jerry_get_object_value (const jerry_value_t value);
+```
+
+- `value` - api value
+- return value - the v_object member of the given `jerry_value_t` parameter.
```
dbatyai: This feels a bit too technical. Can this be something like `return value - the object contained in the given 'jerry_value_t' parameter`

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+
+**Prototype**
+
+```c
+jerry_string_t *
+jerry_get_string_value (const jerry_value_t value);
+```
+
+- `value` - api value
+- return value - the v_string member of the given `jerry_value_t` parameter.
```
dbatyai: Same as above.

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+
+
+# jerry_acquire_value
+
+**Summary**
+
+Acquires the specified Jerry API value.
+
+For values of string, number and object types this acquires the underlying data, for all other types
+it is a no-op. The acquired pointer have to be released by [jerry_release_value](#jerry_release_value).
```
dbatyai: `have` -> `has`

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+```
+
+
+# jerry_create_string
+
+**Summary**
+
+Create new JavaScript string.
+
+Upon the JavaScript string becomes unused, all pointers to it should be released using
```
dbatyai: `Upon` feels quirky. Can we say `When the string is no longer needed ...`.
Also, I wouldn't say here that `all pointers should be released`, only that `it has to be released`, as it is a bit misleading. All _acquired_ pointers should be released, but `jerry_acquire_string` already notes this, and that's fine, as it's only relevant there.

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+- [jerry_create_string](#jerry_create_string)
+
+
+# jerry_create_object
+
+**Summary**
+
+Create new JavaScript object, like with `new Object()`.
+
+*Note*: Upon the JavaScript object becomes unused, all pointers to it should be released using
```
dbatyai: Same as above

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+
+- [jerry_release_value](#jerry_release_value)
+
+
+# jerry_create_error_sz
+
+**Summary**
+
+Creates an error object. Caller should release the object with jerry_release_object, just when
+the value becomes unnecessary.
```
dbatyai: Also, can we change `becomes unnecessary` to `is no longer needed`?

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+
+- [jerry_is_function](#jerry_is_function)
+- [jerry_create_external_function](#jerry_create_external_function)
+
+
+# jerry_construct_object
+
+**Summary**
+
+Construct object invoking specified function object as constructor.
```
dbatyai: `Construct object, ...`

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+}
+
+{
+  jerry_object_t *throw_obj_p = jerry_create_external_function (handler);
+  jerry_object_t *glob_obj_p = jerry_get_global ();
+
+  jerry_value_t val = jerry_create_object_value (throw_obj_p);
+
+  // after this, script can invoke the native handler through "error_func ();"
+  // and "error_func" throw a error on called
```
dbatyai: `"error_func" will throw an error when called`?

docs/API-REFERENCE.md
```diff
@@ -1 +1,2640 @@
...
+
+- [jerry_create_array_object](#jerry_create_array_object)
+
+
+# jerry_create_error
+
+**Summary**
+
+Create new JavaScript error object. It should be throwed inside of handle attached to external
+function object.
```
dbatyai: `... throwed from a handle attached to an external function object.`?

LaszloLango:
@zherczeg, @dbatyai, I've updated the PR.

dbatyai:
LGTM

PR #1161, opened 2016-06-17
Add new API function

LaszloLango:
Implemented 'jerry_is_array' and 'jerry_get_array_length' API functions.

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -591,6 +591,40 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
 } /* jerry_get_array_index_value */
 
 /**
+ * Get length of an array object
+ *
+ * Note:
+ *      Returns 0, if the given parameter is not an array object.
+ *
+ * @return length of the given array
+ */
+uint32_t
+jerry_get_array_length (const jerry_object_t *object_p)
+{
+  jerry_assert_api_available ();
+
+  if (!jerry_is_array (object_p))
+  {
+    return 0;
+  }
+
+  jerry_length_t length = 0;
+  ecma_string_t *magic_string_length_p = ecma_get_magic_string (LIT_MAGIC_STRING_LENGTH);
+
+  ecma_value_t len_value = ecma_op_object_get ((jerry_object_t *) object_p, magic_string_length_p);
+  ecma_deref_ecma_string (magic_string_length_p);
+
+  if (ecma_is_value_number (len_value))
+  {
+    length = ecma_number_to_uint32 (ecma_get_number_from_value (len_value));
+  }
+
+  ecma_free_value (len_value);
+
+  return length;
+} /* jerry_get_array_length */
+
+/**
  * Create an error object
  *
  * Note:
@@ -756,6 +790,20 @@ jerry_dispatch_object_free_callback (ecma_external_pointer_t freecb_p, /**< poin
 
   jerry_make_api_available ();
 } /* jerry_dispatch_object_free_callback */
+
+/**
+ * Check if the specified object is an array object.
+ *
+ * @return true  - if the specified object is an array object,
+ *         false - otherwise.
+ */
+bool
+jerry_is_array (const jerry_object_t *object_p) /**< an object */
+{
+  jerry_assert_api_available ();
+
+  return (ecma_get_object_type (object_p) == ECMA_OBJECT_TYPE_ARRAY);
+} /* jerry_is_array */
 
 /**
  * Check if the specified object is a function object.
--- a/jerry-core/jerry-api.h
+++ b/jerry-core/jerry-api.h
@@ -185,6 +185,7 @@ jerry_string_t *jerry_create_string_sz (const jerry_char_t *, jerry_size_t);
  */
 bool jerry_set_array_index_value (jerry_object_t *, jerry_length_t, jerry_value_t);
 bool jerry_get_array_index_value (jerry_object_t *, jerry_length_t, jerry_value_t *);
+uint32_t jerry_get_array_length (const jerry_object_t *);
 
 /**
  * Functions of 'jerry_string_t'
@@ -196,6 +197,7 @@ jerry_size_t jerry_string_to_char_buffer (const jerry_string_t *, jerry_char_t *
 /**
  * General API functions of JS objects
  */
+bool jerry_is_array (const jerry_object_t *);
 bool jerry_is_constructor (const jerry_object_t *);
 bool jerry_is_function (const jerry_object_t *);
 bool jerry_add_object_field (jerry_object_t *, const jerry_char_t *, jerry_size_t, const jerry_value_t, bool);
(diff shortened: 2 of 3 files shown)


jerry-core/jerry.c
```diff
@@ -591,6 +591,49 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
...
+  length = ecma_number_to_uint32 (len_number);
+
+  ECMA_OP_TO_NUMBER_FINALIZE (len_number);
+  ECMA_FINALIZE (len_value);
+
+  ecma_deref_ecma_string (magic_string_length_p);
+  JERRY_ASSERT (!ECMA_IS_VALUE_ERROR (ret_value));
+
+  return length;
+} /* jerry_get_length */
```
zherczeg: Why this separate function is needed?

jerry-core/jerry.c
```diff
@@ -591,6 +591,49 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
...
+
+/**
+ * Get length of an array object
+ *
+ * @return length of the given array
+ */
+jerry_length_t
```
zherczeg: shouldn't it be uint32?

LaszloLango: It is uint32 already. 

```c
/**
 * Jerry's length
 */
typedef uint32_t jerry_length_t;
```

zherczeg: I think it will be not clear for everybody. The standard is clear here, so I don't think we need to complicate things with own defines.

jerry-core/jerry.c
```diff
@@ -591,6 +591,38 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
...
+                  ret_value);
+  ECMA_OP_TO_NUMBER_TRY_CATCH (len_number, len_value, ret_value);
+
+  length = ecma_number_to_uint32 (len_number);
+
+  ECMA_OP_TO_NUMBER_FINALIZE (len_number);
+  ECMA_FINALIZE (len_value);
+
+  ecma_deref_ecma_string (magic_string_length_p);
+  JERRY_ASSERT (!ECMA_IS_VALUE_ERROR (ret_value));
```
dbatyai: It would be better to handle/free the error if we happen to get one. I don't think hoping that people will play by the rules, and only pass valid arrays to this function is enough.

jerry-core/jerry.c
```diff
@@ -591,6 +591,47 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
...
+
+  JERRY_ASSERT (jerry_is_array (object_p));
+
+  jerry_length_t length = 0;
+  ecma_string_t *magic_string_length_p = ecma_get_magic_string (LIT_MAGIC_STRING_LENGTH);
+
+  ecma_value_t len_value = ecma_op_object_get ((jerry_object_t *) object_p, magic_string_length_p);
+  ecma_deref_ecma_string (magic_string_length_p);
+
+  ecma_number_t num_var = ecma_number_make_nan ();
```
dbatyai: I don't think an intermediate variable is necessary.

jerry-core/jerry.c
```diff
@@ -591,6 +591,47 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
...
+
+  if (ecma_is_value_number (len_value))
+  {
+    num_var = ecma_get_number_from_value (len_value);
+    length = ecma_number_to_uint32 (num_var);
+  }
+  else
+  {
+    ecma_value_t to_number_value = ecma_op_to_number (len_value);
```
dbatyai: Shouldn't this be freed?

jerry-core/jerry.c
```diff
@@ -591,6 +591,47 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
...
+ * Get length of an array object
+ *
+ * @return length of the given array
+ */
+uint32_t
+jerry_get_array_length (const jerry_object_t *object_p)
+{
+  jerry_assert_api_available ();
+
+  JERRY_ASSERT (jerry_is_array (object_p));
```
dbatyai: I'm not sure how I feel about this kind of an assert in an API function. I think it would be better to handle the case when the argument is not an array, for example an early `return 0` or something like that (since we can't throw errors here without changing the return type).

zherczeg: Fail-safe API? Might be a good idea, but all functions must follow this layout, which have perf cost.

E.g:
    if (is_number(value))
      val = get_number(value); // Adding an is_number check here slows down the code.

I am not against fail-safe API, but I would like to hear others opinion.

LaszloLango:
@dbatyai, @zherczeg, I've updated the PR.

jerry-core/jerry.c
dbatyai: Basically what I was thinking is, since any object can behave as an array as long as it has a length field, people may end giving specialized objects to this function thinking that it will work every time (but may fail in some extreme circumstances). On the other hand, seeing that other functions also behave in the way, that giving incorrect arguments will yield undefined behavior, it's probably fine to have this do the same. (But I would definitely leave note about it in the documentation)

jerry-core/jerry.c
```diff
@@ -591,6 +591,41 @@ jerry_get_array_index_value (jerry_object_t *array_obj_p, /**< array object */
 } /* jerry_get_array_index_value */
 
 /**
+ * Get length of an array object
+ *
+ * Note:
+ *      Returns 0, if the given parameter is not an array object or
+ *      cannot retrieve the length.
```
zherczeg: If it is an array you can always retreive the length. It is an unconfigurable data property.

jerry-core/jerry.c
```diff
@@ -758,6 +793,20 @@ jerry_dispatch_object_free_callback (ecma_external_pointer_t freecb_p, /**< poin
...
+ *
+ * @return true  - if the specified object is an array object,
+ *         false - otherwise.
+ */
+bool
+jerry_is_array (const jerry_object_t *object_p) /**< an object */
+{
+  jerry_assert_api_available ();
+
+  return (ecma_object_get_class_name ((jerry_object_t *) object_p) == LIT_MAGIC_STRING_ARRAY_UL);
```
zherczeg: ecma_get_object_type (obj_p) == ECMA_OBJECT_TYPE_ARRAY

is faster

zherczeg:
LGTM after these are fixed

jerry-core/jerry.c
LaszloLango: You can pass any 'const jerry_object_t *object_p' to this function

zherczeg: I am sorry I wasn't clear. So:
if array -> return length, always
if not -> return 0

So the correct comment is:
Returns 0, if the given parameter is not an array object

dbatyai:
LGTM

zherczeg:
LGTM

PR #1162, opened 2016-06-21
Enforce pure C99 build.

tilmannOSG:
Building with -std=c99 is not sufficient to enforce a pure C99 build as it still allows GNU extensions which don't conflict with C99 to be used. Add -pedantic to the build options to ensure that the codebase is only using C99 constructs and the build will fail whenever any compiler extension is used. This helps to ensure that JerryScript remains portable and can be built with any C99-compatible C compiler.

Removing -Wpedantic, as any warnings of that type will already trigger an error when building with -pedantic.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -263,7 +263,7 @@ project (Jerry C ASM)
     endforeach()
   endmacro()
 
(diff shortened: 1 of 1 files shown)


LaszloLango:
LGTM

zherczeg:
LGTM

tilmannOSG:
Thanks for the quick review!

PR #1167, opened 2016-06-23
Check error values in API functions

LaszloLango:
Internal functions cannot handle error values, so it must be avoided to
pass error values to the engine.

### DIFF
--- a/jerry-core/jerry-api.h
+++ b/jerry-core/jerry-api.h
@@ -144,6 +144,7 @@ bool jerry_value_to_boolean (const jerry_value_t);
 jerry_value_t jerry_value_to_number (const jerry_value_t);
 jerry_value_t jerry_value_to_object (const jerry_value_t);
 jerry_value_t jerry_value_to_string (const jerry_value_t);
+jerry_value_t jerry_value_remove_error_flag (const jerry_value_t);
 
 /**
  * Create functions of 'jerry_value_t'
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -62,6 +62,8 @@ static jerry_flag_t jerry_flags;
  */
 static bool jerry_api_available;
 
+static const jerry_char_t *error_value_msg_p = (const jerry_char_t *) "argument cannot have an error flag";
+
 /** \addtogroup jerry_extension Jerry engine extension interface
  * @{
  */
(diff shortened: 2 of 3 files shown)


zherczeg:
Please check the travis fails.

jerry-core/jerry.c
```diff
@@ -62,6 +62,8 @@ static jerry_flag_t jerry_flags;
  */
 static bool jerry_api_available;
 
+static const jerry_char_t *error_value_msg_p = (const jerry_char_t *) "argument cannot be an error value";
```
zherczeg: "cannot have an error flag" is better

jerry-core/jerry.c
```diff
@@ -549,6 +551,11 @@ jerry_set_array_index_value (jerry_object_t *array_obj_p, /**< array object */
                              jerry_length_t index, /**< index to be written */
                              jerry_value_t value) /**< value to set */
 {
+  if (ECMA_IS_VALUE_ERROR (value))
```
zherczeg: the "throw exception, otherwise" comment above is incorrect

jerry-core/jerry.c
```diff
@@ -2166,10 +2214,27 @@ jerry_value_to_string (const jerry_value_t value) /**< input value */
...
+  {
+    return jerry_create_object_value (jerry_create_error (JERRY_ERROR_TYPE, error_value_msg_p));
+  }
+
   return ecma_op_to_string (value);
 } /* jerry_value_to_string */
 
 /**
+ * Convert Jerry value to normal value
```
zherczeg: "Remove the error flag" sounds better.

I think we should emphasize that error is a flag, which can be combined with any value.

jerry-core/jerry.c
LaszloLango: fixed

LaszloLango:
@zherczeg, I've updated the PR. Please check.

jerry-core/jerry-api.h
```diff
@@ -144,6 +144,7 @@ bool jerry_value_to_boolean (const jerry_value_t);
 jerry_value_t jerry_value_to_number (const jerry_value_t);
 jerry_value_t jerry_value_to_object (const jerry_value_t);
 jerry_value_t jerry_value_to_string (const jerry_value_t);
+jerry_value_t jerry_value_to_normal_value (const jerry_value_t);
```
zherczeg: what about jerry_value_remove_error_flag? I don't think people will understand the "normal" phrase.

LaszloLango: I don't mind to change it.

jerry-core/jerry.c
```diff
@@ -2115,6 +2148,11 @@ jerry_value_to_boolean (const jerry_value_t value) /**< input value */
 {
   jerry_assert_api_available ();
 
+  if (ECMA_IS_VALUE_ERROR (value))
+  {
+    return false;
+  }
+
```
zherczeg: I think a value with error flag has a logical value. Or you do this for consistency?

zherczeg: Come to think of it, this is not a conversion, so even the internal api has a wrong name. Hence the value is not converted to boolen (e.g. a Boolean object is always true) so perhaps we need another conversion routine.

    var a = new Boolean(false); if (a) print("TRUE"); // yields true, although a to_boolean should yield false.

I would change this function to jerry_get_logical_value

LaszloLango: Yes I do

LaszloLango: IMHO, this should be ToBoolean function of the standard: http://www.ecma-international.org/ecma-262/5.1/index.html#sec-9.2
I think the name is good as is.

zherczeg: Ok

zherczeg:
I was thinking, and I convinced myself that the current approach is right.
LGTM

dbatyai:
LGTM

PR #1165, opened 2016-06-22
Four byte optimized memcpy.

zherczeg:
Binary unchanged.

Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 1.152 -> 1.147 : +0.378% | 
3d-raytrace.js | 1.371 -> 1.315 : +4.092% | 
access-binary-trees.js | 0.674 -> 0.673 : +0.062% | 
access-fannkuch.js | 2.854 -> 2.845 : +0.315% | 
access-nbody.js | 1.386 -> 1.387 : -0.047% | 
bitops-3bit-bits-in-byte.js | 0.630 -> 0.631 : -0.154% | 
bitops-bits-in-byte.js | 0.917 -> 0.917 : -0.063% | 
bitops-bitwise-and.js | 1.920 -> 1.923 : -0.123% | 
bitops-nsieve-bits.js | 2.276 -> 2.240 : +1.578% | 
controlflow-recursive.js | 0.475 -> 0.475 : +0.026% | 
crypto-aes.js | 1.251 -> 1.244 : +0.622% | 
crypto-md5.js | 0.849 -> 0.845 : +0.473% | 
crypto-sha1.js | 0.792 -> 0.789 : +0.432% | 
date-format-tofte.js | 1.023 -> 1.019 : +0.409% | 
date-format-xparb.js | 0.556 -> 0.551 : +0.820% | 
math-cordic.js | 1.669 -> 1.667 : +0.100% | 
math-partial-sums.js | 0.986 -> 0.986 : -0.013% | 
math-spectral-norm.js | 0.705 -> 0.700 : +0.740% | 
string-base64.js | 4.597 -> 2.241 : +51.246% | 
string-fasta.js | 2.145 -> 2.124 : +1.000% | 
Geometric mean: | +4.046% |

PR #1169, opened 2016-06-24
Fix memcpy return value.

zherczeg:


zherczeg:
Should fix #1165

LaszloLango:
I think we should add unit tests for jerry-libc at some point.

LaszloLango:
LGTM

dbatyai:
LGTM

PR #1172, opened 2016-06-29
Doc converter

ktorpi:


tools/update-webpage.sh
```diff
@@ -0,0 +1,94 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if [ $# -ne 1 ]; then
+  echo "Please, specify your gh-pages clone directory: update-webpage <gh-pages clone dir>"
+  exit 1
+fi
+
+gh_pages_dir=$1
+docs_dir="../docs"
```
galpeter: I'm not sure that if we ran it like this way: ``` tools/update-webpage.sh <path>``` the script will find the docs_dir.

tools/update-webpage.sh
```diff
@@ -0,0 +1,94 @@
...
+
+permalink[$GETTING_STARTED_MD]='/getting-started/'
+permalink[$API_REFERENCE_MD]='/api-reference/'
+permalink[$API_EXAMPLES_MD]='/api-example/'
+permalink[$INTERNALS_MD]='/internals/'
+
+order[$GETTING_STARTED_MD]='01.'
+order[$API_REFERENCE_MD]='02.'
+order[$API_EXAMPLES_MD]='03.'
+order[$INTERNALS_MD]='04.'
```
ktorpi: The ordering determines in which order the menus will appear on the webpage (https://ktorpi.github.io/jerryscript/) and I think it would be nice to have e.g. the Getting Started page before the API reference.

tools/update-webpage.sh
ktorpi: That's right. The question is from which directory we want to call it? If there is a "policy" to call these  scripts `tools/<script>.sh`, I'll fix it.

tools/update-webpage.sh
galpeter: How is it generated (I mean what generates the menu)? Is it something done by the github automatically?

ktorpi: It is done by the github automatically. I don't know the details... :(

galpeter: I see. In that case, can't we just have the ordering by default for the input filenames? So that we don't need to hardcode here. My main problem is that with this if I add a new md file I'll need to add everything here also. It would be really great that just by adding the md file this script still works. 

The premalink could be calculated from the md file name. Only the title can't be extracted from the md file name.

ktorpi: I've discussed with the team and I will add the numbers to the original filenames .

tools/update-webpage.sh
galpeter: If I'm correct you can get a scripts base directory with: `dirname $(readlink -f $0)`

ktorpi:
Check the update, please.

tools/update-webpage.sh
```diff
@@ -0,0 +1,83 @@
...
+    link_end_pos = index(line_remain, ")")
+    link = substr(line_remain, 1, link_end_pos-1);
+    line_end = substr(line_remain, link_end_pos)
+
+    # delete underscores form the link
+    gsub(/_/, "", link);
+
+    printf "%s%s%s\n", line_beg, link, line_end
+  }
+  ' $docfile                                             >> $gh_pages_dir/$docfile_base
```
LaszloLango: Is it really necessary to write so many space after '$docfile'? :)

ktorpi: Of course, it's not, but I wanted to allign the right side of the redirection to the ones that above (line 45-53), because I belive the script looks better, and it's more readable this way.

LaszloLango:
LGTM

zherczeg:
LGTM

PR #1173, opened 2016-06-29
Optimize arithmetic in VM.

zherczeg:
Nice performance improvement. Binary size unchanged.

Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 1.138 -> 1.038 : +8.777% | 
3d-raytrace.js | 1.252 -> 1.171 : +6.487% | 
access-binary-trees.js | 0.638 -> 0.621 : +2.690% | 
access-fannkuch.js | 2.842 -> 2.741 : +3.544% | 
access-nbody.js | 1.313 -> 1.165 : +11.278% | 
bitops-3bit-bits-in-byte.js | 0.624 -> 0.586 : +5.942% | 
bitops-bits-in-byte.js | 0.919 -> 0.876 : +4.696% | 
bitops-bitwise-and.js | 1.501 -> 1.460 : +2.718% | 
bitops-nsieve-bits.js | 1.953 -> 1.954 : -0.010% | 
controlflow-recursive.js | 0.458 -> 0.435 : +4.920% | 
crypto-aes.js | 1.228 -> 1.223 : +0.368% | 
crypto-md5.js | 0.786 -> 0.774 : +1.586% | 
crypto-sha1.js | 0.751 -> 0.730 : +2.854% | 
date-format-tofte.js | 0.919 -> 0.913 : +0.661% | 
date-format-xparb.js | 0.507 -> 0.498 : +1.844% | 
math-cordic.js | 1.604 -> 1.553 : +3.170% | 
math-partial-sums.js | 0.877 -> 0.814 : +7.224% | 
math-spectral-norm.js | 0.669 -> 0.614 : +8.206% | 
string-base64.js | 2.163 -> 2.112 : +2.346% | 
string-fasta.js | 2.021 -> 1.993 : +1.384% | 
Geometric mean: | +4.082% |


jerry-core/vm/vm.c
```diff
@@ -1651,6 +1657,36 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
         }
         case VM_OC_ADD:
         {
+          if (ecma_are_values_integer_numbers (left_value, right_value))
```
tilmannOSG: For the newly added code to the VM_OC_ADD and VM_OC_SUB cases there seems to be a lot of code duplication (minus the actual add/sub operations and invocations of the respective helper functions). This probably could be factored out into a static inline helper function without a negative performance impact. Not sure it's worth the hassle though, I guess it's up to you whether you want to go that route :)

tilmannOSG:
LGTM from my side although my knowledge of that part of the code is still fairly limited :)

jerry-core/vm/vm.c
zherczeg: Could work, but I need to make a separate patch to try that out and see what is happening.

LaszloLango:
Nice improvement, LGTM

jerry-core/vm/vm.c
zherczeg: It seems the helper function requires several parameters, so it does not look really nice. Macros make hard to debug things.

The VM is the most critical element of the engine performance, so we use techniques here which is not used in the other parts of Jerry. E.g. gotos, and code duplications. I know this increases the maintenance burden (and binary size), but the gain here outweights the costs. For binary size optimizations, we usually focus on rarely used but large parts of the code, e.g. json support.

tilmannOSG: That's reasonable, I agree that introducing macros here would not be beneficial to the maintainability of the code.  
Since you don't see any code size regressions I think it's fine to go ahead as is.

dbatyai:
LGTM

PR #1177, opened 2016-06-30
Jerry API update

LaszloLango:


jerry-core/parser/js/js-parser.c
```diff
@@ -2265,24 +2264,22 @@ parser_parse_script (const jerry_char_t *source_p, /**< source code */
  * Note:
  *      returned error object should be freed with jerry_release_object
  */
-jsp_status_t
-parser_parse_eval (const jerry_char_t *source_p, /**< source code */
+ecma_object_t *
```
zherczeg: I would prefer return with an ecma_value_t.

jerry-core/jerry-api.h
```diff
@@ -77,50 +88,67 @@ typedef uint32_t jerry_size_t;
...
  */
 typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
 
 /**
- * Function type applied for each fields in objects
+ * Function type applied for each data property of an object
  */
-typedef bool (*jerry_object_field_foreach_t) (const jerry_string_t *field_name_p,
-                                              const jerry_value_t field_value,
+typedef bool (*jerry_object_field_foreach_t) (const jerry_value_t property_name_p,
```
zherczeg: jerry_object_property_foreach_t

jerry-core/jerry-api.h
```diff
@@ -77,50 +88,67 @@ typedef uint32_t jerry_size_t;
...
+void jerry_cleanup (void);
+void jerry_register_magic_strings (const jerry_char_ptr_t *, uint32_t, const jerry_length_t *);
+void jerry_get_memory_limits (size_t *, size_t *);
+void jerry_gc (void);
+
+/**
+ * Parser and executor functions
+ */
+bool jerry_run_simple (const jerry_char_t *, size_t, jerry_init_flag_t);
+jerry_value_t jerry_parse (const jerry_char_t *, size_t); // true value if success
```
zherczeg: comment format?

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
...
+jerry_value_t jerry_create_boolean (bool);
+jerry_value_t jerry_create_error (jerry_error_t, const jerry_char_t *);
+jerry_value_t jerry_create_error_sz (jerry_error_t, const jerry_char_t *, jerry_size_t);
+jerry_value_t jerry_create_external_function (jerry_external_handler_t);
+jerry_value_t jerry_create_number (double);
+jerry_value_t jerry_create_null (void);
+jerry_value_t jerry_create_object (void);
+jerry_value_t jerry_create_string (const jerry_char_t *);
+jerry_value_t jerry_create_string_sz (const jerry_char_t *, jerry_size_t);
+jerry_value_t jerry_create_undefined (void);
```
zherczeg: I would organize this a bit more logical order. Primitive types first, number, string, object later.

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
...
 
 /**
  * Functions of array objects
  */
-bool jerry_set_array_index_value (jerry_object_t *, jerry_length_t, jerry_value_t);
-bool jerry_get_array_index_value (jerry_object_t *, jerry_length_t, jerry_value_t *);
-uint32_t jerry_get_array_length (const jerry_object_t *);
+jerry_value_t jerry_set_array_index_value (const jerry_value_t, uint32_t, const jerry_value_t);
```
zherczeg: Does it needs to be array? I think this should work with any objects.

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
...
+jerry_value_t jerry_set_own_data_property (const jerry_value_t, const jerry_value_t, const jerry_value_t);
+jerry_value_t jerry_get_data_property (const jerry_value_t, const jerry_value_t);
+jerry_value_t jerry_get_own_data_property (const jerry_value_t, const jerry_value_t);
+
+jerry_value_t jerry_define_own_data_property (const jerry_value_t,
+                                              const jerry_value_t,
+                                              bool,
+                                              bool,
+                                              bool,
+                                              const jerry_value_t);
```
zherczeg: We don't need one line for each type.

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
...
-bool jerry_foreach_object_field (jerry_object_t *, jerry_object_field_foreach_t, void *);
-bool jerry_get_object_native_handle (jerry_object_t *, uintptr_t *);
-void jerry_set_object_native_handle (jerry_object_t *, uintptr_t, jerry_object_free_callback_t);
-jerry_value_t jerry_construct_object (jerry_object_t *, const jerry_value_t[], uint16_t);
-jerry_value_t jerry_call_function (jerry_object_t *, jerry_object_t *, const jerry_value_t[], uint16_t);
+bool jerry_has_property (const jerry_value_t, const jerry_value_t);
+bool jerry_has_own_property (const jerry_value_t, const jerry_value_t);
+bool jerry_delete_property (const jerry_value_t, const jerry_value_t);
+
+jerry_value_t jerry_set_data_property (const jerry_value_t, const jerry_value_t, const jerry_value_t);
```
zherczeg: Does it set data property only? I mean, if a set accessor is defined, does it not call this accessor? If it works with accessors as well, I would call it jerry_set_property_value

jerry-core/jerry-api.h
zherczeg: Btw data_property name is correct here.

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
...
+                                                  const jerry_value_t,
+                                                  bool,
+                                                  bool,
+                                                  const jerry_value_t,
+                                                  const jerry_value_t);
+
+jerry_value_t jerry_function_call (const jerry_value_t, const jerry_value_t, const jerry_value_t[], jerry_size_t);
+jerry_value_t jerry_object_construct (const jerry_value_t, const jerry_value_t[], jerry_size_t);
+
+jerry_value_t jerry_get_object_keys (const jerry_value_t);
```
zherczeg: I think jerry_foreach_object_field should be after this, since they belong to the same group.

zherczeg:
@poussa and @lemmaa the API update patch is here. Could you give your feedback about the changes in jerry-api.h? Of course you can give feedback for other files as well but since you use the api, we would value your opinion about the API itself.

jerry-core/jerry-api.h
LaszloLango: IMHO, it's easier to search for something when it is in lexical order.

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
...
+                                              const jerry_value_t);
+jerry_value_t jerry_define_own_accessor_property (const jerry_value_t,
+                                                  const jerry_value_t,
+                                                  bool,
+                                                  bool,
+                                                  const jerry_value_t,
+                                                  const jerry_value_t);
+
+jerry_value_t jerry_function_call (const jerry_value_t, const jerry_value_t, const jerry_value_t[], jerry_size_t);
+jerry_value_t jerry_object_construct (const jerry_value_t, const jerry_value_t[], jerry_size_t);
```
zherczeg: I would use function_call and construct_object

jerry-core/jerry.c
```diff
@@ -119,526 +124,565 @@ jerry_make_api_available (void)
...
- * Returns whether the given jerry_value_t is string.
+ * Register external magic string array
  */
-bool
-jerry_value_is_string (const jerry_value_t value) /**< api value */
+void
+jerry_register_magic_strings (const jerry_char_ptr_t *ex_str_items, /**< character arrays, representing
+                                                                     *   external magic strings' contents */
+                              uint32_t count,                       /**< number of the strings */
+                              const jerry_length_t *str_lengths)    /**< lengths of the strings */
```
zherczeg: I think our commenting style does not use alignment.

jerry-core/jerry.c
```diff
@@ -119,526 +124,565 @@ jerry_make_api_available (void)
...
  *
- * @return pointer that may be used outside of the engine
+ * @return result of eval, may be error value.
  */
-jerry_string_t *
-jerry_acquire_string (jerry_string_t *string_p) /**< pointer passed to function */
+jerry_value_t
+jerry_eval (const jerry_char_t *source_p, /**< source code */
+            size_t source_size, /**< length of source code */
+            bool is_strict) /**< perform eval as it is called from strict mode code */
```
zherczeg: I think the comment should be:
/**< source must conform with strict mode */
we call this from C, so strict mode is meaningless in that context

jerry-core/jerry.c
```diff
@@ -119,526 +124,565 @@ jerry_make_api_available (void)
...
- * Note:
- *      For values of string and object types this releases the underlying data,
- *      for all other types it is a no-op.
+ * @return true  - if the error flag of the specified value is true,
+ *         false - otherwise.
  */
-void
-jerry_release_value (jerry_value_t value) /**< API value */
+bool
+jerry_value_is_error (const jerry_value_t value) /**< api value */
```
zherczeg: I would prefer jerry_value_has_error_flag, since strictly speaking, we don't have "error value". Any value can be an error with the error flag.

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
 jerry_value_t jerry_value_remove_error_flag (const jerry_value_t);
```
zherczeg: Perhaps jerry_value_clear_error_flag is a better name here. And we could add a set_error_flag as well.

jerry-core/jerry.c
```diff
@@ -119,526 +124,565 @@ jerry_make_api_available (void)
...
- * @return pointer to created array object
+ * @return true or false.
  */
-jerry_object_t *
-jerry_create_array_object (jerry_size_t size) /**< size of array */
+bool
+jerry_get_boolean_value (const jerry_value_t value) /**< api value */
 {
-  JERRY_ASSERT (size > 0);
+  JERRY_ASSERT (jerry_value_is_boolean (value));
```
zherczeg: We could return with false if the value is not object to be fail-safe. Or that might be a bad idea, I don't know.

jerry-core/jerry.c
```diff
@@ -119,526 +124,565 @@ jerry_make_api_available (void)
...
-  jerry_object_t *obj_p = ecma_get_object_from_value (new_array_completion);
+/**
+ * Get number from the specified value as a double.
+ *
+ * @return stored number as double
+ */
+double
+jerry_get_number_value (const jerry_value_t value) /**< api value */
+{
+  JERRY_ASSERT (jerry_value_is_number (value));
```
zherczeg: Return with +0 if value is not a number.

jerry-core/jerry.c
```diff
@@ -119,526 +124,565 @@ jerry_make_api_available (void)
...
+/**
+ * Create an array object value
+ *
+ * Note:
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return value of the constructed array object
+ */
+jerry_value_t
+jerry_create_array (jerry_size_t size) /**< size of array */
```
zherczeg: size must be uint32_t

jerry-core/jerry.c
```diff
@@ -652,893 +696,858 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
+  if (!ecma_is_value_object (array_obj_val))
   {
-    is_successful = false;
+    return jerry_create_type_error ();
   }
 
-  ecma_free_value (delete_completion);
-
-  ecma_deref_ecma_string (field_name_str_p);
+  ecma_string_t *str_idx_p = ecma_new_ecma_string_from_uint32 ((uint32_t) index);
```
zherczeg: You can create the string on the stack.

jerry-core/jerry-api.h
LaszloLango: Currently it checks the argument, but we can make it more general. What about 'jerry_set_value_at_index'?

jerry-core/jerry-api.h
LaszloLango: We have to write it because it would be longer then 120 chars. What wrong with the name? What do you suggest?

jerry-core/jerry-api.h
zherczeg: by index?

jerry-core/jerry-api.h
zherczeg: This is allowed for interfaces:

    jerry_value_t jerry_define_own_data_property (const jerry_value_t, const jerry_value_t,
                                                  bool, bool, bool, const jerry_value_t);

poussa:
@jimmy-huang @grgustaf FYI changes coming. PTAL.

jerry-core/jerry-api.h
galpeter: `jerry_set_property_index` ?

tilmannOSG:
@franc0is @HBehrens We would appreciate your feeback on the new API as well :)

grgustaf:
One of @zherczeg 's comments above reminded me, it seems odd that jerry_get_array_length returns uint32_t but [gs]et_array_index_value take jerry_length_t. But I'll admit the myriad aliases for ints confuse me.

Is that as it should be?

jerry-core/jerry-api.h
```diff
@@ -147,71 +173,83 @@ jerry_value_t jerry_value_to_string (const jerry_value_t);
...
+jerry_size_t jerry_get_string_size (const jerry_value_t);
+jerry_length_t jerry_get_string_length (const jerry_value_t);
+jerry_size_t jerry_string_to_char_buffer (const jerry_value_t, jerry_char_t *, jerry_size_t);
 
 /**
  * General API functions of JS objects
  */
-bool jerry_is_array (const jerry_object_t *);
-bool jerry_is_constructor (const jerry_object_t *);
-bool jerry_is_function (const jerry_object_t *);
```
poussa: Why are we removing these 3 jerry_is_* functions? How to use the new API to achieve the same functionality?

zherczeg: There are new functions ofr that:

    bool jerry_value_is_array (const jerry_value_t);

the point is we don't export the real types of values anymore, all of them are jerry_value_t

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -315,7 +319,7 @@ ecma_op_object_delete (ecma_object_t *obj_p, /**< the object */
     {
       JERRY_ASSERT (false);
```
dbatyai: If this is a valid code path, than this should probably be removed. Otherwise, it would be better to use `JERRY_UNREACHABLE`.

jerry-core/vm/vm.c
```diff
@@ -220,22 +218,16 @@ vm_run_global (ecma_value_t *error_value_p) /**< [out] error value */
...
-    *error_value_p = ret_value;
-    ret_code = JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION;
-  }
-  else
+  if (!ECMA_IS_VALUE_ERROR (ret_value))
   {
     ecma_free_value (ret_value);
-    *error_value_p = ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
-    ret_code = JERRY_COMPLETION_CODE_OK;
+    ret_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE);
```
dbatyai: Why not return the actual value instead of replacing it with a `true`? The result may prove to be useful.

jerry-core/jerry.c
```diff
@@ -2090,172 +2094,86 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
-
-/**
- * Call the ToString ecma builtin operation on the api value.
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
  *
  * Note:
- *      - Can throw error, so return value should be checked with 'jerry_value_is_error'.
- *      - Return value must be freed with 'jerry_release_value'.
+ *      if called native handler returns true, then dispatcher just returns value received
+ *      through 'return value' output argument, otherwise - throws the value as an exception.
```
dbatyai: Could this note get cleaned up a bit? Capitalization, `otherwise ` in a new line, etc.

jerry-core/vm/vm.c
LaszloLango: good idea

jerry-core/jerry-api.h
```diff
@@ -77,51 +89,110 @@ typedef uint32_t jerry_size_t;
...
+bool jerry_run_simple (const jerry_char_t *, size_t, jerry_init_flag_t);
+jerry_value_t jerry_parse (const jerry_char_t *, size_t);
+jerry_value_t jerry_run (void);
+jerry_value_t jerry_eval (const jerry_char_t *, size_t, bool);
 
 /**
  * Get the global context
  */
-jerry_object_t *jerry_get_global (void);
+jerry_value_t jerry_get_global (void);
```
zherczeg: I would use jerry_get_global_object since it is not clear that global is an object anymore.

jerry-core/jerry-api.h
```diff
@@ -130,12 +201,29 @@ bool jerry_value_is_string (const jerry_value_t);
...
+/**
  * Getter functions of 'jerry_value_t'
  */
 bool jerry_get_boolean_value (const jerry_value_t);
 double jerry_get_number_value (const jerry_value_t);
-jerry_string_t *jerry_get_string_value (const jerry_value_t);
-jerry_object_t *jerry_get_object_value (const jerry_value_t);
+
+/**
+ * Functions of string values
```
zherczeg: of -> for? It sounds better to me.

jerry-core/jerry-api.h
```diff
@@ -144,74 +232,70 @@ bool jerry_value_to_boolean (const jerry_value_t);
 jerry_value_t jerry_value_to_number (const jerry_value_t);
 jerry_value_t jerry_value_to_object (const jerry_value_t);
 jerry_value_t jerry_value_to_string (const jerry_value_t);
-jerry_value_t jerry_value_remove_error_flag (const jerry_value_t);
```
zherczeg: I would export ecma_op_to_primitive as well.

jerry-core/jerry-api.h
```diff
@@ -144,74 +232,70 @@ bool jerry_value_to_boolean (const jerry_value_t);
...
-jerry_length_t jerry_get_string_length (const jerry_string_t *);
-jerry_size_t jerry_string_to_char_buffer (const jerry_string_t *, jerry_char_t *, jerry_size_t);
+jerry_value_t jerry_set_property (const jerry_value_t, const jerry_value_t, const jerry_value_t);
+jerry_value_t jerry_set_own_property (const jerry_value_t, const jerry_value_t, const jerry_value_t);
+jerry_value_t jerry_set_property_by_index (const jerry_value_t, uint32_t, const jerry_value_t);
+jerry_value_t jerry_get_property (const jerry_value_t, const jerry_value_t);
+jerry_value_t jerry_get_own_property (const jerry_value_t, const jerry_value_t);
+jerry_value_t jerry_get_property_by_index (const jerry_value_t , uint32_t);
+
+void jerry_init_property_descriptor (jerry_property_descriptor_t *);
```
zherczeg: perhaps jerry_init_property_descriptor_fields would be more consistent

jerry-core/jerry.c
```diff
@@ -119,525 +124,571 @@ jerry_make_api_available (void)
...
 /**
- * Returns whether the given jerry_value_t is error.
+ * Construct new TypeError object
+ *
+ * @return TypeError object value
  */
-bool
-jerry_value_is_error (const jerry_value_t value) /**< api value */
+static inline jerry_value_t __attr_always_inline___
+jerry_create_type_error (void)
```
zherczeg: I don't think it is worth to inline this, since this is a reare event, and not worth optimizing it for performance.

jerry-core/jerry.c
```diff
@@ -119,525 +124,571 @@ jerry_make_api_available (void)
...
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return acquired api value
+ */
+jerry_value_t
+jerry_acquire_value (jerry_value_t value) /**< API value */
+{
+  jerry_assert_api_available ();
+
+  return ecma_copy_value (value);
```
zherczeg: fast_copy

jerry-core/jerry.c
```diff
@@ -119,525 +124,571 @@ jerry_make_api_available (void)
...
+ *
+ * Note:
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return value of the constructed array object
+ */
+jerry_value_t
+jerry_create_array (jerry_size_t size) /**< size of array */
+{
+  JERRY_ASSERT (size > 0);
```
zherczeg: Size can be 0 I think.

jerry-core/jerry.c
```diff
@@ -119,525 +124,571 @@ jerry_make_api_available (void)
...
+
+/**
+ * Create a jerry_value_t representing a boolean value from the given boolean parameter.
+ *
+ * @return value of the created boolean
+ */
+jerry_value_t
+jerry_create_boolean (bool value) /**< bool value from which a jerry_value_t will be created */
+{
+  return (value ? ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE)
```
zherczeg: Please use ecma_make_boolean_value.

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
-  ecma_value_t obj_val = ecma_make_object_value (object_p);
-
-  return ecma_op_is_callable (obj_val);
-} /* jerry_is_function */
+  return ecma_make_object_value (ecma_op_create_object_object_noarg ());
+} /* jerry_create_object */
 
 /**
- * Check if the specified object is a constructor function object.
+ * Create a string
```
zherczeg: Create string from a valid CESU8 string.

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
     return false;
   }
 
-  bool is_successful = true;
+  ecma_property_t *prop_p = ecma_op_object_get_property (ecma_get_object_from_value (obj_val),
+                                                         ecma_get_string_from_value (prop_name_val));
 
-  ecma_string_t *field_name_str_p = ecma_new_ecma_string_from_utf8 ((lit_utf8_byte_t *) field_name_p,
-                                                                    (lit_utf8_size_t) field_name_size);
+  return (prop_p != NULL ? ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE)
```
zherczeg: Please use ecma_make_boolean_value.

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
-    is_successful = false;
+    return false;
   }
 
-  ecma_free_value (set_completion);
-  ecma_deref_ecma_string (field_name_str_p);
+  ecma_property_t *prop_p = ecma_op_object_get_own_property (ecma_get_object_from_value (obj_val),
+                                                             ecma_get_string_from_value (prop_name_val));
+
+  return (prop_p != NULL ? ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE)
```
zherczeg: Ditto.

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
   jerry_assert_api_available ();
 
-  uintptr_t handle_value;
-
-  bool does_exist = ecma_get_external_pointer_value (object_p,
-                                                     ECMA_INTERNAL_PROPERTY_NATIVE_HANDLE,
-                                                     &handle_value);
-
-  if (does_exist)
+  if (!ecma_is_value_object (obj_val)
```
zherczeg: Please check error as well.

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
-    JERRY_ASSERT (this_arg_p == NULL);
-    JERRY_ASSERT (jerry_is_constructor (function_object_p));
+  jerry_assert_api_available ();
 
-    call_completion = ecma_op_function_construct (function_object_p,
-                                                  args_p,
-                                                  args_count);
-  }
-  else
+  if (!jerry_has_own_property (obj_val, prop_name_val))
```
zherczeg: Please check error.

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
+  ecma_property_descriptor_t prop_desc = ecma_make_empty_property_descriptor ();
 
-  status = jerry_convert_eval_completion_to_retval (completion, retval_p);
+  prop_desc.is_value_defined = prop_desc_p->is_value_defined;
+  prop_desc.value = prop_desc_p->value;
 
-  return status;
-} /* jerry_eval */
+  prop_desc.is_writable_defined = prop_desc_p->is_writable_defined;
+  prop_desc.is_writable = prop_desc_p->is_writable;
```
zherczeg: You cannot trust the is_writable flag. The correct form is:

    prop_desc.is_writable_defined = prop_desc_p->is_writable_defined;
    prop_desc.is_writable = prop_desc_p->is_writable_defined ? prop_desc_p->is_writable : false;

Similar cases below.

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
+  if (ECMA_IS_VALUE_ERROR (obj_val)
+      || ECMA_IS_VALUE_ERROR (prop_name_val))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p));
+  }
+  else if (!ecma_is_value_object (obj_val)
+           || !ecma_is_value_string (prop_name_val))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG (wrong_args_msg_p));
+  }
```
zherczeg: Please also check that the property descriptor is not both a data nd accessor property. That is:

    if (((writable_defined && writable) || has_value) && (has_getter || has_setter)) -> ERROR

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
 
-  ecma_value_t completion = ecma_op_eval_chars_buffer ((const lit_utf8_byte_t *) source_p,
-                                                       source_size,
-                                                       is_direct,
-                                                       is_strict);
+  ecma_property_descriptor_t prop_desc = ecma_make_empty_property_descriptor ();
 
-  status = jerry_convert_eval_completion_to_retval (completion, retval_p);
+  prop_desc.is_value_defined = prop_desc_p->is_value_defined;
+  prop_desc.value = prop_desc_p->value;
```
zherczeg: What about the accessor properties?

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
+  prop_desc_p->is_enumerable = prop_desc.is_enumerable;
+  prop_desc_p->is_enumerable_defined = prop_desc.is_enumerable_defined;
+  prop_desc_p->is_writable = prop_desc.is_writable;
+  prop_desc_p->is_writable_defined = prop_desc.is_writable_defined;
+
+  prop_desc_p->is_get_defined = prop_desc.is_get_defined;
+  prop_desc_p->is_set_defined = prop_desc.is_set_defined;
+  prop_desc_p->is_value_defined = prop_desc.is_value_defined;
+
+  prop_desc_p->getter = ecma_make_object_value (prop_desc.get_p);
```
zherczeg: is_getter defined? NULL check?

jerry-core/jerry.c
```diff
@@ -651,893 +702,885 @@ jerry_create_error (jerry_error_t error_type, /**< type of error */
...
+{
+  JERRY_ASSERT (args_count == 0 || args_p != NULL);
+
+  if (ECMA_IS_VALUE_ERROR (func_obj_val)
+      || ECMA_IS_VALUE_ERROR (this_val))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p));
+  }
+  else if (!ecma_is_value_object (func_obj_val)
+           || (!ecma_is_value_object (this_val)
```
zherczeg: I think it can be primitive type in strict mode.

jerry-core/jerry-api.h
```diff
@@ -77,51 +89,110 @@ typedef uint32_t jerry_size_t;
...
  */
-jerry_object_t *jerry_get_global (void);
+jerry_value_t jerry_get_global (void);
 
 /**
  * Checker functions of 'jerry_value_t'
  */
+bool jerry_value_is_array (const jerry_value_t);
 bool jerry_value_is_boolean (const jerry_value_t);
-bool jerry_value_is_error (const jerry_value_t);
```
jiangzidong: I think we should remain jerry_value_is_error, because Error is in the same level as Array, Function, Boolean, Number, ... They are all js builtin-types

zherczeg: I feel this can be misleading. Error is a type of object but any value can be thrown. Hence people might think if "jerry_value_is_error" returns with false, there is no error, but this is not true. The opposite is also true, so having an error object does not mean an error is occured. Perhaps jerry_value_is_error_object could be better. What do you think?

zherczeg: E.g:

    var a = new Error("No error"); // This is an error object, but not an error
    throw true; // This is a true value, but also represents an error
    throw void 0; // This is an undefined value, but also represents an error

jiangzidong: Yes, it is confusing that Error type and value with error flag.  
1) the new apis return jerry_value_t instead of a boolean, and we should check if it run correctly by jerry_value_has_error_flag(), right?
2) But the api: jerry_create_error, it will set the error flag after create the error object. How should we create a pure error type object

zherczeg: 1) Yes. You can also chain operations, and just check the last value. This is a clever use of the API.
2) Normally you want to throw an error object so I think it is reasonable to set the error flag. If you want to manipulate this flag you can use jerry_value_set_error_flag / jerry_value_clear_error_flag with any value. It sets the flag for that particular value. Therefore the flag belongs to the value, not the referenced object. Hence the same error object can be referenced from a jerry_value which has the error flag and another value which doesn't have this flag in the same time.

Is this ok for you?

jiangzidong: Personally, I just feel a little strange that **create_error = create error object + set error flag**
I understand it is convenient for api users, but it may also confuse some one.
I suggest that we emphasize the **different between error type and error flag, and the behavior of jerry_create_error** in some api docs

Thanks for your detailed explanation!

LaszloLango: In the current implementation the create_error functions also set the error flag.

zherczeg: I am sorry if it was confusing. Yes, when you create an error object, the flag is automatically set for convenience. And we will explain this in the documentation of course.

jerry-core/jerry.c
LaszloLango: fixed

docs/03.API-EXAMPLE.md
```diff
@@ -180,36 +163,21 @@ main (int argc, char * argv[]) {
...
-- `v_uint32` (if JERRY_API_DATA_TYPE_UINT32) - 32 bit unsigned integer value;
-- `v_string` (if JERRY_API_DATA_TYPE_STRING) - pointer to string;
-- `v_object` (if JERRY_API_DATA_TYPE_OBJECT) - pointer to object.
-
-Abstract values, to be sent to or received from the engine are described with the structure.
+## Step 5. Description of JerryScript value descriptors
 
-Pointers to strings or objects and values should be released just when become unnecessary, using `jerry_release_string` or `jerry_release_object` and `jerry_release_value`, correspondingly.
+JerryScript value can be a boolean, number, null, object, string or undefined. The value has an error flag,
+that indicates whether is an error or not. Every type can be an error not only objects.
```
zherczeg: Every type has an error flag not only objects.

docs/03.API-EXAMPLE.md
```diff
@@ -425,22 +376,16 @@ main (int argc, char * argv[])
...
 
-  jerry_value_t eval_ret;
-
   /* Evaluate script */
-  status = jerry_eval (script,
-                       script_size,
-                       false,
-                       false,
-                       &eval_ret);
+  jerry_value_t eval_ret = jerry_eval (script, script_size, false);
```
zherczeg: status = jerry_has_error_flag (eval_ret);

LaszloLango: `status` variable is removed from the example. Place it back?

jerry-core/ecma/operations/ecma-objects.c
zherczeg: There are several such asserts in legacy code, perhaps we could fix all of them in a separate patch.

dbatyai: Okay.

docs/03.API-EXAMPLE.md
zherczeg: If return value matters.

jerry-core/parser/js/js-parser-internal.h
```diff
@@ -19,6 +19,7 @@
 
 #include "common.h"
 
+#include "byte-code.h"
```
zherczeg: Why do we need this?

jerry-core/vm/vm.c
```diff
@@ -226,39 +210,29 @@ static const uint16_t vm_decode_table[] =
  *
  * @return completion code
  */
-jerry_completion_code_t
-vm_run_global (ecma_value_t *error_value_p) /**< [out] error value */
+ecma_value_t
+vm_run_global (const ecma_compiled_code_t *bytecode_p)
```
zherczeg: Missing comment.

jerry-core/vm/vm.c
zherczeg: I agree.

targets/zephyr/Makefile.zephyr
```diff
@@ -1,4 +1,5 @@
 # Copyright © 2016 Intel Corporation
+ * Copyright 2016 University of Szeged.
```
zherczeg: Why do you add this?

zherczeg:
I think this is a great job!
LGTM after fixing those small issues.

jerry-core/parser/js/js-parser-internal.h
LaszloLango: There are some `CBC_` opcode below

targets/zephyr/Makefile.zephyr
LaszloLango: Sorry, there is a bug in my licence updater script.

docs/03.API-EXAMPLE.md
LaszloLango: It doesn't really matters in these small examples.

jerry-core/vm/vm.c
LaszloLango: What about `jerry_run`? Should it give back the result of `vm_run_global` instead of true?

LaszloLango:
@zherczeg, @dbatyai, I've updated the PR. Please check

docs/03.API-EXAMPLE.md
```diff
@@ -180,36 +163,21 @@ main (int argc, char * argv[]) {
...
-- `v_uint32` (if JERRY_API_DATA_TYPE_UINT32) - 32 bit unsigned integer value;
-- `v_string` (if JERRY_API_DATA_TYPE_STRING) - pointer to string;
-- `v_object` (if JERRY_API_DATA_TYPE_OBJECT) - pointer to object.
-
-Abstract values, to be sent to or received from the engine are described with the structure.
+## Step 5. Description of JerryScript value descriptors
 
-Pointers to strings or objects and values should be released just when become unnecessary, using `jerry_release_string` or `jerry_release_object` and `jerry_release_value`, correspondingly.
+JerryScript value can be a boolean, number, null, object, string or undefined. The value has an error flag,
+that indicates whether is an error or not. Every type has an error flag not only objects.
```
dbatyai: It should also be noted that the error flag should be cleared before the value is passed as an argument.

jerry-core/jerry-api.h
```diff
@@ -32,15 +33,26 @@ extern "C"
...
+ * Jerry init flags
  */
 typedef enum
 {
-  JERRY_COMPLETION_CODE_OK                       = 0, /**< successful completion */
-  JERRY_COMPLETION_CODE_UNHANDLED_EXCEPTION      = 1, /**< exception occured and it was not handled */
-  JERRY_COMPLETION_CODE_INVALID_SNAPSHOT_VERSION = 2, /**< snapshot version mismatch */
-  JERRY_COMPLETION_CODE_INVALID_SNAPSHOT_FORMAT  = 3, /**< snapshot format is not valid */
-} jerry_completion_code_t;
+  JERRY_INIT_EMPTY              = (0u),      /**< empty flag set */
```
dbatyai: Maybe this would sound better as `JERRY_INIT_DEFAULT`, but I'm okay with `_EMPTY` as well. Up to you.

docs/03.API-EXAMPLE.md
dbatyai: And that created error objects automatically have the flag set.

dbatyai:
LGTM otherwise.

jiangzidong:
@LaszloLango Hi, can I assume that it is the API 1.0, and will not change a lot in this period?

zherczeg:
We are hoping this API will be stable. Perhaps we add more functions, but not removing them. Unless something serious comes up. Hopefully not.

jiangzidong:
Thanks! and is our next step is Jerry Port API #964?:)

LaszloLango:
@jiangzidong, yes. Port API patch will be the next :)

PR #1181, opened 2016-07-01
Fix issue when you don't flash the ARC and the x86 waits

sergioamr:
- Fixed problem when the x86 hangs waiting for the ARC to
be initialized. If you flash as arduino_101 instead of
arduino_101_factory this will happen and you will not get
a command line and you don't have anything on the arc.

- Changed default build to arduino_101_factory which is
the current recommended method of flashing the arduino_101

- Updated documentation to explain a few bits on how to flash
the board and conform to the v1.4.0 way of flashing the device.

- Added a helper to run the dfu-util command directly on make.

- Fixing small details like the path to jerryscript changed from jerry to jerryscript.

JerryScript-DCO-1.0-Signed-off-by: Sergio Martinez sergio.martinez.rodriguez@intel.com


zherczeg:
LGTM

LaszloLango:
LGTM

PR #1182, opened 2016-07-01
targets/zephyr: Add support for NXP FRDM-K64F board.

pfalcon:
Tested to boot and run "test" one-liner.

### DIFF
--- a/targets/zephyr/Makefile.zephyr
+++ b/targets/zephyr/Makefile.zephyr
@@ -66,7 +66,7 @@ CONFIG_TOOLCHAIN_VARIANT = x86
 CPU = x86
 EXT_CFLAGS += -march=pentium
 EXT_CFLAGS += -mpreferred-stack-boundary=2 -mno-sse
-else ifeq ($(BOARD),qemu_cortex_m3)
+else ifeq ($(BOARD),$(filter $(BOARD),qemu_cortex_m3 frdm_k64f))
 CONFIG_TOOLCHAIN_VARIANT = arm
 CPU = arm7-m
 EXT_CFLAGS += -march=armv7-m -mthumb -mcpu=cortex-m3 -mabi=aapcs
--- a/targets/zephyr/README.md
+++ b/targets/zephyr/README.md
@@ -25,7 +25,8 @@ harmony
 
 2. Target boards/emulations
 
-Following Zephyr boards were tested: qemu_x86, qemu_cortex_m3, arduino_101.
+Following Zephyr boards were tested: qemu_x86, qemu_cortex_m3, arduino_101,
+frdm_k64f.
 
 
 #### 2. Prepare Zephyr


pfalcon:
@sergioamr, @poussa: FYI.

sergioamr:
LGTM

targets/zephyr/Makefile.zephyr
```diff
@@ -70,7 +70,7 @@ CONFIG_TOOLCHAIN_VARIANT = x86
 CPU = x86
 EXT_CFLAGS += -march=pentium
 EXT_CFLAGS += -mpreferred-stack-boundary=2 -mno-sse
-else ifeq ($(BOARD),qemu_cortex_m3)
+else ifeq ($(BOARD),$(filter $(BOARD),qemu_cortex_m3 frdm_k64f))
 CONFIG_TOOLCHAIN_VARIANT = arm
```
tilmannOSG: Might be worth to add a separate case for the FRDM-K64F board with more aggressive compiler flags, tuning for the Cortex-M4F to allow the compiler to take advantage of the native DSP and floating-point instructions which are available with ARMv7E-M, e.g. -mcpu=cortex-m4 -mfpu=fpv4-sp-d16.  
Native floating-point support will definitely make a performance difference for JerryScript :)

tilmannOSG:
LGTM

zherczeg:
LGTM

targets/zephyr/Makefile.zephyr
pfalcon: > Might be worth to add a separate case for the FRDM-K64F board with more aggressive compiler flags

I agree that would be nice follow-up target. This one is however a small patch to add support of a  first physical ARM/Cortex-M board for Zephyr port (previously it was tested with just QEMU).

And when doing such CPU-specific optimizations, it would be nice to reuse machinery Zephyr already has for that, instead of building parallel if-else ladders in JerryScript makefile. I know that work on better integration with 3rd-party libs is ongoing in Zephyr, so another reason not to go for "premature optimization" at this spot. Hope that makes sense.

Thanks for review!

pfalcon:
Rebased.

targets/zephyr/Makefile.zephyr
tilmannOSG: I see, thanks for providing the the necessary Zephyr context. I agree that ideally the compiler flags should be specified in a single location for Zephyr and then just cascade down to JerryScript. Perfectly fine to address this in a later patch once Zephyr has the appropriate support in place :)

LaszloLango:
LGTM

tilmannOSG:
Looks like this one is ready to go in.  
@pfalcon Can you please do another rebase on master?

pfalcon:
Rebased, thanks.

Issue #1184, opened 2016-07-04
Test failure: jerry --exec-snapshot date-tostring triggers assertion


tilmannOSG:
In the Travis CI ARM build of https://github.com/Samsung/jerryscript/pull/1182, the following test failed:  
[37/266] build/bin/debug.linux/jerry  --exec-snapshot date-tostring.t9Fkk0uDHJ.snapshot: FAIL (1)
Script error: assertion failed  
The full log is available at https://travis-ci.org/Samsung/jerryscript/jobs/142165136.  
It's pretty much impossible that https://github.com/Samsung/jerryscript/pull/1182 actually caused this failure since it only changed a Zephyr-specific Makefile and a README, e.g. no code changes.  
This is either a recently introduced non-deterministic regression or an issue with our test harness.

zherczeg:
Actually this flaky test is there for a long time, and improssible to reproduce the crash. Even if it is reported by travis, and the binary is downloaded, the crash does not happen.

tilmannOSG:
I see, thanks for the clarification. Definitely sounds like something non-deterministic. Might be worth to run the test with AddressSanitizer/MemorySanitzer/Valgrind to get to the bottom of this. I would guess it's probably an uninitialized variable or a pointer in a hash table causing this non-determinism.  
Did this show up on x86-64 as well? Or is it specific to ARM?

zherczeg:
We spent some time to figure this out (with valgrind), but no success at all. Might have some dependency which is not available in our environment. It is a date test after all.

akosthekiss:
What to do with this issue? Close with "who knows why?" or keep it open as a reminder (but still, who knows why?...)

LaszloLango:
IMHO, keep open for a while.

zherczeg:
We could not reproduce it on other machines so far. Very difficult to debug.

tilmannOSG:
@zherczeg Mentioned to me that this issues has not triggered a single time after the build system rewrite. Could very well have been a bug in the test harness of the old build system. Closing for now, feel free to reopen in case the issue shows up again.

tilmannOSG:
Reopening, managed to trigger the bug again: https://travis-ci.org/Samsung/jerryscript/jobs/159634805  
Should be possible to reproduce with a507c08.

LaszloLango:
It looks like this was fixed. Feel free to reopen if needed.

PR #1187, opened 2016-07-05
Fix signed left shift in vm_loop().

### LINKED ISSUE #1174: Compile error in vm.c with GCC 6
I've got a compile error in the vm.c with GCC 6.x:
```
jerry-core/vm/vm.c: In function ‘vm_loop’:
jerry-core/vm/vm.c:1222:57: error: left shift of negative value [-Werror=shift-negative-value]
               if (int_value <= (ECMA_INTEGER_NUMBER_MIN << ECMA_DIRECT_SHIFT))
                                                         ^~
cc1: all warnings being treated as errors
```

tilmannOSG:
Signed left shift operations are undefined in C. Add constants for the minimum/maximum integer value which are already shifted. Technically, the constant for the shifted maximum value is not required, adding it for consistency/increased readability.

The bug was detected by -Wshift-negative-value both with GCC 6.x and Clang.

This fixes #1174.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -143,19 +143,27 @@ typedef int32_t ecma_integer_value_t;
  * Maximum integer number for an ecma value
  */
 #if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
-#define ECMA_INTEGER_NUMBER_MAX 0x7fffff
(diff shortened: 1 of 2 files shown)


tilmannOSG:
The Travis CI results should actually be all green. They are red because #1184 has struck again.

zherczeg:
Just re-run the travis. That error is pretty frequent these days. It is there for a long time, but time-to-time its frequency increases or decreases.

zherczeg:
LGTM. Compilers tends to overdo things.

LaszloLango:
LGTM

PR #1190, opened 2016-07-07
targets/zephyr: Update to new Jerry API and improve REPL interaction.

pfalcon:
Using jerry_eval() instead of jerry_run_simple() allows to store (and then access)
variables in a global environment. E.g. following now works:

    a = 1
    print(a)

### DIFF
--- a/targets/zephyr/src/main-zephyr.c
+++ b/targets/zephyr/src/main-zephyr.c
@@ -21,7 +21,7 @@
 #include <misc/printk.h>
 #include <misc/shell.h>
 
-#include "jerry.h"
+#include "jerry-api.h"
 
 #if defined (CONFIG_STDOUT_CONSOLE)
 #include <stdio.h>
@@ -41,19 +41,19 @@ static unsigned char flags = 0;
  */
 int jerryscript_test ()
 {
-  jerry_completion_code_t ret_code;
+  jerry_value_t ret_val;
 
   const char script[] =
   "var test=0; " \
   "for (var t=100; t<1000; t++) test+=t; " \
   "print ('Hi JS World! '+test);";
 
-  printf ("Script [%s]\n",script);
-  ret_code = jerry_run_simple ((jerry_char_t *) script,
+  printf ("Script [%s]\n", script);
+  ret_val = jerry_eval ((jerry_char_t *) script,
       strlen (script),
-      JERRY_FLAG_EMPTY);
+      false);
 
-  return ret_code;
+  return jerry_value_has_error_flag (ret_val) ? -1 : 0;
 } /* jerryscript_test */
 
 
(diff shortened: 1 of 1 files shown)


pfalcon:
This follows Unix port (main-unix.c) and API example doc.

pfalcon:
@sergioamr, @poussa: FYI.

zherczeg:
Sounds like a good idea. The run_simple initializes and shuts down the engine, so it cannot return with any meaningful value, since jerry heap has been freed already.

zherczeg:
LGTM

poussa:
LGTM

LaszloLango:
This won't work after #1177 
@pfalcon, would you like to land it now or wait for the API update first?

pfalcon:
@LaszloLango : If #1177 is close to be merged (it seems to be), I guess it makes sense to land it first, then I'll update this patch. Thanks.

sergioamr:
LGTM. I would like to be able to have the Travis integration working, so changes on the API get highlighted and do not break the build.

The situation is that @Akromx16 will go back to university really soon and i am not sure if he would be able to finish his integration on https://github.com/Samsung/jerryscript/pull/1144

Could someone with travis experience help on that PR?

LaszloLango:
#1177 is merged. `docs/02.API-EXAMPLES.md` is up to date, but `docs/02.API-REFERENCE.md` is not. Let me know, if you need any help with the update.

pfalcon:
@LaszloLango : Thanks, refactored to the new API, rebased to master as of now.

zherczeg:
LGTM

targets/zephyr/src/main-zephyr.c
```diff
@@ -158,6 +158,7 @@ const struct shell_cmd commands[] =
 void main (void)
 {
   printf ("Jerry Compilation " __DATE__ " " __TIME__ "\n");
+  jerry_init (JERRY_INIT_EMPTY);
   shell_register_app_cmd_handler (shell_cmd_handler);
   shell_init ("js> ", commands);
 } /* main */
```
LaszloLango: `jerry_cleanup` should be called before the end of main

LaszloLango:
LGTM after `jerry_cleanup` call is added.

sergioamr:
lgtm

targets/zephyr/src/main-zephyr.c
pfalcon: @LaszloLango : That's not the case with Zephyr. Generally, a typical embedded app would work as an infinite loop (it "exits" via hardware reset), and specifically with Zephyr, shell_init() just sets up background task and exit, all actual input passed to callbacks at later time. So, it's not ok to deinitialize Jerry after call to shell_init(), and per above, no need to deinitialize at all. Hope that makes sense. I'm adding a comment to the patch with this info.

pfalcon:
Added comment to the source why jerry_cleanup() is not called, to address @LaszloLango's comment above.

targets/zephyr/src/main-zephyr.c
sergioamr: @LaszloLango

shell_init will spawn a fiber that performs the shell task.
task_fiber_start(stack, STACKSIZE, shell, 0, 0, 7, 0);

If you clean zephyr there it will just crash.

With the change of running code on eval i think pfalcon is right on not cleaning up after execution.

LaszloLango: @pfalcon, @sergioamr, OK, thanks for the explanation.

LaszloLango:
LGTM

PR #1201, opened 2016-07-13
Implement IO port API

LaszloLango:
Related issue: #964

### DIFF
--- a/jerry-core/jrt/jrt.h
+++ b/jerry-core/jrt/jrt.h
@@ -86,9 +86,9 @@ extern void __noreturn jerry_unimplemented (const char *, const char *, const ch
 #define JERRY_LOG(lvl, ...) \
   do \
   { \
-    if (lvl <= jerry_debug_level && jerry_log_file) \
+    if (lvl <= jerry_debug_level) \
     { \
-      jerry_port_logmsg (jerry_log_file, __VA_ARGS__); \
+      jerry_port_log (JERRY_LOG_LEVEL_DEBUG, __VA_ARGS__); \
     } \
   } \
   while (0)
@@ -109,8 +109,8 @@ extern void __noreturn jerry_unimplemented (const char *, const char *, const ch
 #define JERRY_DDDLOG(...) JERRY_DLOG (__VA_ARGS__)
 #endif /* JERRY_ENABLE_LOG */
 
-#define JERRY_ERROR_MSG(...) jerry_port_errormsg (__VA_ARGS__)
-#define JERRY_WARNING_MSG(...) JERRY_ERROR_MSG (__VA_ARGS__)
+#define JERRY_ERROR_MSG(...) jerry_port_log (JERRY_LOG_LEVEL_ERROR, __VA_ARGS__)
+#define JERRY_WARNING_MSG(...) jerry_port_log (JERRY_LOG_LEVEL_WARNING, __VA_ARGS__)
 
 /**
  * Mark for unreachable points and unimplemented cases
--- a/targets/esp8266/user/jerry_port.c
+++ b/targets/esp8266/user/jerry_port.c
@@ -17,34 +17,36 @@
 #include <stdio.h>
 #include <stdarg.h>
 
-
 /**
- * Provide log message to filestream implementation for the engine.
+ * Provide console message implementation for the engine.
  */
-int jerry_port_logmsg (FILE* stream, const char* format, ...)
+void
+jerry_port_console (const char *format, /**< format string */
+                    ...) /**< parameters */
 {
   va_list args;
-  int count = 0;
   va_start (args, format);
-  // TODO, uncomment when vfprint link is ok
-  //count = vfprintf (stream, format, args);
+  /* TODO, uncomment when vprint link is ok */
+  /* vfprintf (stdout, format, args); */
   va_end (args);
-  return count;
-}
+} /* jerry_port_console */
 
 /**
- * Provide error message to console implementation for the engine.
+ * Provide log message implementation for the engine.
  */
-int jerry_port_errormsg (const char* format, ...)
+void
+jerry_port_log (jerry_log_level_t level, /**< log level */
+                const char *format, /**< format string */
+                ...)  /**< parameters */
 {
+  (void) level; /* ignore log level */
+
   va_list args;
-  int count = 0;
   va_start (args, format);
-  // TODO, uncomment when vprint link is ok
-  //count = vprintf (format, args);
+  /* TODO, uncomment when vprint link is ok */
+  /* vprintf (stderr, format, args); */
   va_end (args);
-  return count;
-}
+} /* jerry_port_log */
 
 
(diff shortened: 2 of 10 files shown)


LaszloLango:
@pfalcon, @sergioamr  please check the Zephyr port. I hope my speculative update work as it should.

LaszloLango:
Note: after this PR, we will change all the printf calls inside the engine to the related port API function (`jerry_port_log` and `jerry_port_console`)

LaszloLango:
cc: @jiangzidong

jerry-core/jrt/jrt.h
```diff
@@ -109,7 +109,7 @@ extern void __noreturn jerry_unimplemented (const char *, const char *, const ch
 #define JERRY_DDDLOG(...) JERRY_DLOG (__VA_ARGS__)
 #endif /* JERRY_ENABLE_LOG */
 
-#define JERRY_ERROR_MSG(...) jerry_port_errormsg (__VA_ARGS__)
+#define JERRY_ERROR_MSG(...) jerry_port_log (JERRY_LOG_LEVEL_ERROR, __VA_ARGS__)
 #define JERRY_WARNING_MSG(...) JERRY_ERROR_MSG (__VA_ARGS__)
```
galpeter: Shouldn't the warning use warning level?

sergioamr:
Will check it in my morning tomorrow :)

jiangzidong:
It will bring convenience to porting work. Thanks

targets/esp8266/user/jerry_port.c
```diff
@@ -17,34 +17,36 @@
...
+void
+jerry_port_console (const char *format, /**< format string */
+                    ...) /**< parameters */
 {
   va_list args;
-  int count = 0;
   va_start (args, format);
   // TODO, uncomment when vfprint link is ok
-  //count = vfprintf (stream, format, args);
+  // vfprintf (stdout, format, args);
```
zherczeg: please use /* */ comment style

targets/esp8266/user/jerry_port.c
```diff
@@ -17,34 +17,36 @@
...
+                ...)  /**< parameters */
 {
+  (void) level; /* ignore log level */
+
   va_list args;
-  int count = 0;
   va_start (args, format);
   // TODO, uncomment when vprint link is ok
-  //count = vprintf (format, args);
+  // vprintf (stderr, format, args);
```
zherczeg: Ditto

zherczeg:
LGTM after the comment fixes.

dbatyai:
LGTM

jerry-core/jrt/jrt.h
```diff
@@ -86,9 +86,9 @@ extern void __noreturn jerry_unimplemented (const char *, const char *, const ch
 #define JERRY_LOG(lvl, ...) \
   do \
   { \
-    if (lvl <= jerry_debug_level && jerry_log_file) \
+    if (lvl <= jerry_debug_level) \
```
akosthekiss: IMHO, this check should not happen here but in the port implementation (if that impl cares about it). Likewise, `jerry_debug_level` shouldn't even be defined in jerry.c. Everything `JERRY_LOG` should do is just pass things on to `jerry_port_log` implemented in the port for it to decide what to do with it (as `lvl` is passed on as well).

In some sense, it should work similarly to the "fatal" logic. As far ar jerry-core is concerned, there is only a single `jerry_port_fatal` function to be called. However, the default port implementation adds some extras to it, namely the "abort_on_fail" functionality. However, that's a port-specific thing. Therefore, the `static bool abort_on_fail` variable is defined in jerry-port-default-fatal.c, and accessor methods `jerry_port_default_set_abort_on_fail` and `jerry_port_default_is_abort_on_fail` in jerry-port-default.h.

I can imagine a similar scenario with the logs. There could be default port-specific accessors, e.g., `jerry_port_default_set_log_level` and `jerry_port_default_get_log_level` in jerry-port-default.h and their implementation (along with any variables needed for that logic, say `debug_level` or `log_level`) in jerry-port-default-io.c. And then the line `(void) level; /* default port implementation ignores the log level */` will go away (actually, the presence of this line already signals the problem that log levels are not checked at the right place, IMHO).

Additonal notes:
* `JERRY_D[D[D]]LOG` macros call `JERRY_LOG` with 1, 2, and 3 as level parameters. This does not seem right in the light of the new log level enum.
* There are still some `printf`s in the code (e.g., jrt-fatals.c, js-parser.c, js-parser-util.c, ecma-builtin-global.c, jmem-poolman.c, jmem-heap.c, js-parser-statm.c), even if mostly macro-guarded. When will these be rewritten to call the `_log` and `_console` functions?

akosthekiss: (Sorry, missed the "Note: after this PR," part. Cancel my last Q.)

sergioamr:
tested branch on my arduino101. LGTM

jerry-core/jrt/jrt.h
LaszloLango: I disagree with the first part. `JERRY_D[D[D]]LOG` macros are all on debug level. There where no multiple debug levels in the proposed (and accepted) IO port API proposal (see: #964).

I think moving the debug level to the port is might be a good idea. I'll try to do it in a followup work.

LaszloLango:
@sergioamr, thanks

jerry-core/jrt/jrt.h
akosthekiss:
```c
#ifdef JERRY_ENABLE_LOG
#define JERRY_LOG(lvl, ...) \
  do \
  { \
    if (lvl <= jerry_debug_level) \
    { \
      jerry_port_log (JERRY_LOG_LEVEL_DEBUG, __VA_ARGS__); \
    } \
  } \
  while (0)

#define JERRY_DLOG(...) JERRY_LOG (1, __VA_ARGS__)
#define JERRY_DDLOG(...) JERRY_LOG (2, __VA_ARGS__)
#define JERRY_DDDLOG(...) JERRY_LOG (3, __VA_ARGS__)
#else /* !JERRY_ENABLE_LOG */
#define JERRY_DLOG(...) \
  do \
  { \
    if (false) \
    { \
      jerry_ref_unused_variables (0, __VA_ARGS__); \
    } \
  } while (0)
#define JERRY_DDLOG(...) JERRY_DLOG (__VA_ARGS__)
#define JERRY_DDDLOG(...) JERRY_DLOG (__VA_ARGS__)
#endif /* JERRY_ENABLE_LOG */
```
This is current status. `JERRY_D[D[D]]LOG` macros expand to `JERRY_LOG (]0-3], ...)` depending on `JERRY_ENABLE_LOG` settings. The first parameter should be the debug level you just mentioned, not integers.

LaszloLango: @akosthekiss, as I said. I'll do it in a follow up work. Don't worry. It's on my list to clean this up.

PR #1207, opened 2016-07-14
Optimize ecma_string_get_array_index

dbatyai:

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -554,27 +554,83 @@ bool
 ecma_string_get_array_index (const ecma_string_t *str_p, /**< ecma-string */
                              uint32_t *out_index_p) /**< [out] index */
 {
-  bool is_array_index = true;
-  if (ECMA_STRING_GET_CONTAINER (str_p) == ECMA_STRING_CONTAINER_UINT32_IN_DESC)
+  const ecma_string_container_t type = ECMA_STRING_GET_CONTAINER (str_p);
+
+  if (type == ECMA_STRING_CONTAINER_UINT32_IN_DESC)
   {
-    *out_index_p = str_p->u.uint32_number;
+    const uint32_t index = str_p->u.uint32_number;
+    *out_index_p = index;
+    return index != UINT32_MAX;
+  }
+  else if (type == ECMA_STRING_CONTAINER_MAGIC_STRING)
+  {
+    return false;
   }
   else
   {
-    ecma_number_t num = ecma_string_to_number (str_p);
-    *out_index_p = ecma_number_to_uint32 (num);
+    lit_utf8_size_t size;
+    const lit_utf8_byte_t *raw_str_p;
 
-    ecma_string_t *to_uint32_to_string_p = ecma_new_ecma_string_from_uint32 (*out_index_p);
+    if (unlikely (type == ECMA_STRING_CONTAINER_MAGIC_STRING_EX))
+    {
+      size = lit_get_magic_string_ex_size (str_p->u.magic_string_ex_id);
+      raw_str_p = lit_get_magic_string_ex_utf8 (str_p->u.magic_string_ex_id);
+    }
+    else
+    {
+      JERRY_ASSERT (type == ECMA_STRING_CONTAINER_HEAP_UTF8_STRING);
 
-    is_array_index = ecma_compare_ecma_strings (str_p,
-                                                to_uint32_to_string_p);
+      size = str_p->u.utf8_string.size;
+      raw_str_p = (const lit_utf8_byte_t *) (str_p + 1);
+    }
 
-    ecma_deref_ecma_string (to_uint32_to_string_p);
+    if (*raw_str_p == LIT_CHAR_0)
+    {
+      *out_index_p = 0;
+      return size == 1;
+    }
+
+    if (size > 10)
+    {
+      return false;
+    }
+
+    uint32_t index = 0;
+    const lit_utf8_size_t end = (size == 10) ? 9 : size;
+
+    for (lit_utf8_size_t i = 0; i < end; i++)
+    {
+      if (raw_str_p[i] > LIT_CHAR_9 || raw_str_p[i] < LIT_CHAR_0)
+      {
+        return false;
+      }
+
+      index = (index * 10) + (uint32_t) (raw_str_p[i] - LIT_CHAR_0);
+    }
+
+    if (size == 10)
+    {
+      if (index > UINT32_MAX / 10
+          || raw_str_p[9] > LIT_CHAR_9
+          || raw_str_p[9] < LIT_CHAR_0)
+      {
+        return false;
+      }
+
+      index *= 10;
+      const uint32_t digit = (uint32_t) (raw_str_p[9] - LIT_CHAR_0);
+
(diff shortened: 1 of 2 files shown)


dbatyai:
Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 0.944 -> 0.944 : -0.019% | 
3d-raytrace.js | 1.114 -> 1.117 : -0.253% | 
access-binary-trees.js | 0.582 -> 0.583 : -0.202% | 
access-fannkuch.js | 2.365 -> 2.361 : +0.165% | 
access-nbody.js | 1.064 -> 1.065 : -0.119% | 
bitops-3bit-bits-in-byte.js | 0.582 -> 0.582 : +0.005% | 
bitops-bits-in-byte.js | 0.869 -> 0.869 : +0.001% | 
bitops-bitwise-and.js | 0.939 -> 0.940 : -0.092% | 
bitops-nsieve-bits.js | 1.799 -> 1.791 : +0.493% | 
controlflow-recursive.js | 0.379 -> 0.379 : -0.028% | 
crypto-aes.js | 1.114 -> 1.116 : -0.161% | 
crypto-md5.js | 0.733 -> 0.734 : -0.043% | 
crypto-sha1.js | 0.695 -> 0.696 : -0.087% | 
date-format-tofte.js | 0.865 -> 0.867 : -0.192% | 
date-format-xparb.js | 0.455 -> 0.454 : +0.164% | 
math-cordic.js | 1.318 -> 1.320 : -0.174% | 
math-partial-sums.js | 0.692 -> 0.692 : +0.034% | 
math-spectral-norm.js | 0.584 -> 0.584 : -0.024% | 
string-base64.js | 2.053 -> 2.055 : -0.128% | 
string-fasta.js | 1.761 -> 1.355 : +23.066% | 
Geometric mean: | +1.270% | 

Binary sizes (bytes)
35c0869ef5:156236
ae50973fb9:156236

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -554,27 +554,49 @@ bool
 ecma_string_get_array_index (const ecma_string_t *str_p, /**< ecma-string */
                              uint32_t *out_index_p) /**< [out] index */
 {
-  bool is_array_index = true;
-  if (ECMA_STRING_GET_CONTAINER (str_p) == ECMA_STRING_CONTAINER_UINT32_IN_DESC)
+  uint32_t index;
+  const ecma_string_container_t type = ECMA_STRING_GET_CONTAINER (str_p);
+  if (type == ECMA_STRING_CONTAINER_UINT32_IN_DESC)
```
zherczeg: Newline before if.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -554,27 +554,49 @@ bool
...
+    for (lit_utf8_size_t i = 0; i < size; i++)
+    {
+      if (raw_str_p[i] > LIT_CHAR_9 || raw_str_p[i] < LIT_CHAR_0)
+      {
+        return false;
+      }
 
-  is_array_index = is_array_index && (*out_index_p != ECMA_MAX_VALUE_OF_VALID_ARRAY_INDEX);
+      index *= 10;
+      index += (uint32_t) (raw_str_p[i] - LIT_CHAR_0);
```
zherczeg: What I don't see here is the overflow and leading zero checks. E.g:

0001 or 11111111111111111111111111 are not valid array indicies.

zherczeg:
Nice catch btw.

jerry-core/ecma/base/ecma-helpers-string.c
zherczeg: I mean array[1] is the same as array["1"] but array[1] is not the same as array["0001"] or array["+1"]. Property names are always strings. So a valid array index needs to be /0/ or /[1-9][0-9]*/ where the number is lower than UINT32_MAX-1. ( <= 4294967294). So any decimal number whose length is <= 9 digits are valid, and we need the max check for 10 digit numbers. Strings longer than 10 characters cannot be valid array indicies.

LaszloLango:
LGTM

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -554,27 +554,63 @@ bool
...
 
-    is_array_index = ecma_compare_ecma_strings (str_p,
-                                                to_uint32_to_string_p);
+      size = str_p->u.utf8_string.size;
+      raw_str_p = (const lit_utf8_byte_t *) (str_p + 1);
+    }
 
-    ecma_deref_ecma_string (to_uint32_to_string_p);
-  }
+    if (size > 10
```
zherczeg: I would put the zero check first:

    if (*raw_str_p == LIT_CHAR_0) return size == 1;

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -554,27 +554,63 @@ bool
...
+    for (lit_utf8_size_t i = 0; i < size; i++)
+    {
+      if (raw_str_p[i] > LIT_CHAR_9 || raw_str_p[i] < LIT_CHAR_0)
+      {
+        return false;
+      }
+
+      const uint32_t tmp = (index * 10) + (uint32_t) (raw_str_p[i] - LIT_CHAR_0);
+
+      if (unlikely (tmp < index))
```
zherczeg: Such constructs are usually optimized out by compilers. I had similar bugs before.

I would do something like this:
1) end =size == 10 ? 9 : size;
2) the usual for loop [0..end] -- we know the number fits in index.
3) if size == 10, then do the appropriate checks. This is really a rare case.
 if (index > UINT32_MAX / 10) return false;
 if (index == UINT32_MAX / 10) -- another check
 else return true;

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -554,27 +554,81 @@ bool
...
+        return false;
+      }
+
+      index = (index * 10) + (uint32_t) (raw_str_p[i] - LIT_CHAR_0);
+    }
+
+    if (size == 10)
+    {
+      if (index > UINT32_MAX / 10
+          || raw_str_p[9] > LIT_CHAR_9 || raw_str_p[9] < LIT_CHAR_0)
```
zherczeg: Style issue: two || in one line.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -554,27 +554,81 @@ bool
...
+      {
+        return false;
+      }
+
+      index *= 10;
+      const uint32_t digit = (uint32_t) (raw_str_p[9] - LIT_CHAR_0);
+
+      if (index > UINT32_MAX - digit)
```
zherczeg: I think this should be >=, because UINT32_MAX is also an invalid array index.

Perhaps we could remove ECMA_MAX_VALUE_OF_VALID_ARRAY_INDEX macro, and in the 

    if (type == ECMA_STRING_CONTAINER_UINT32_IN_DESC)

case just simply check that index is not UINT32_MAX. I don't like this definition:

    #define ECMA_MAX_VALUE_OF_VALID_ARRAY_INDEX ((uint32_t) (-1))

zherczeg:
LGTM

PR #1208, opened 2016-07-14
Remove arbitrary case conversion tests.

zherczeg:
Since upper/lower case conversions are unspecified by the standard,
we convert ASCII characters only, and a few other characters for
testing purposes. Because these are just random cases, it is better
to remove them before the release. At some point we could add a
unicode compatible case conversion which can be enabled at compile
time.


LaszloLango:
LGTM

dbatyai:
LGTM

PR #1210, opened 2016-07-15
Fix bug in jerry_string_to_char_buffer.

jiangzidong:
Str_len can smaller than buffer_size.

```
  if (ecma_string_get_size (str_p) > buffer_size)
  {
    return 0;
  }
```
It should be ">" in the comparison.

And also change the related code in unit-test-api.c. 
Before, in the `foreach`, it only sunk into
```
  if (sz == 0)
  {
    JERRY_ASSERT (!strncmp ((const char *) user_data, "user_data", 9));
    return true;
  }
```

### DIFF
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -999,7 +999,7 @@ jerry_string_to_char_buffer (const jerry_value_t value, /**< input string value 
 
   ecma_string_t *str_p = ecma_get_string_from_value (value);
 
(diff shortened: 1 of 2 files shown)


zherczeg:
LGTM

LaszloLango:
LGTM

LaszloLango:
@jiangzidong, please rebase to the current master, so we can land it properly.

PR #1225, opened 2016-07-25
Move built-in, external string, and interpreter globals to the global context

zherczeg:
There are still globals left.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.c
@@ -20,6 +20,7 @@
 #include "ecma-globals.h"
 #include "ecma-helpers.h"
 #include "ecma-objects.h"
+#include "jcontext.h"
 #include "jrt-bit-fields.h"
 
 #define ECMA_BUILTINS_INTERNAL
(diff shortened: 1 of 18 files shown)


jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -798,56 +838,6 @@ ecma_builtin_dispatch_construct (ecma_object_t *obj_p, /**< built-in object */
 } /* ecma_builtin_dispatch_construct */
 
 /**
- * Dispatcher of built-in routines
- *
- * @return ecma value
- *         Returned value must be freed with ecma_free_value.
- */
-static ecma_value_t
-ecma_builtin_dispatch_routine (ecma_builtin_id_t builtin_object_id, /**< built-in object' identifier */
```
galpeter: If I see this correctly this was just moved above a few lines. Do we really need this to be moved?

zherczeg: Not necessary, but I would prefer to do this. It is better from logical perspective.

dbatyai:
LGTM

jerry-core/parser/regexp/re-compiler.c
```diff
@@ -614,25 +612,26 @@ re_compile_bytecode (const re_compiled_code_t **out_bytecode_p, /**< [out] point
...
+
+      cache_idx = JERRY_CONTEXT (re_cache_idx)++;
 
-      if (cached_bytecode_p != NULL)
+      /* The garbage collector might run during the byte code
+       * allocations above and it may free this entry. */
+      if (JERRY_CONTEXT (re_cache)[cache_idx] != NULL)
       {
-        ecma_bytecode_deref ((ecma_compiled_code_t *) cached_bytecode_p);
+        ecma_bytecode_deref ((ecma_compiled_code_t *) JERRY_CONTEXT (re_cache)[cache_idx]);
```
LaszloLango: style of `JERRY_CONTEXT (...)[...]` is mixed. There is a space after `)` in some place. I prefer this style (without space). Please unify that.

LaszloLango:
LGTM after the style is fixed.

PR #1229, opened 2016-07-28
Change the CONFIG_ECMA_NUMBER_TYPE default value to CONFIG_ECMA_NUMBE…

robertsipka:
…R_FLOAT64.


jerry-core/CMakeLists.txt
```diff
@@ -171,7 +170,9 @@ elseif(FEATURE_PROFILE STREQUAL "minimal")
       CONFIG_ECMA_COMPACT_PROFILE_DISABLE_JSON_BUILTIN)
 
 else()
-  message(FATAL_ERROR "FEATURE_PROFILE='${FEATURE_PROFILE}' doesn't supported")
```
zherczeg: Can't we use elseif like above?

robertsipka: Sure, I've updated the patch.

zherczeg:
LGTM

jerry-core/CMakeLists.txt
```diff
@@ -143,13 +143,12 @@ else()
...
-  set(DEFINES_JERRY ${DEFINES_JERRY} CONFIG_ECMA_NUMBER_TYPE=CONFIG_ECMA_NUMBER_FLOAT64)
-
 # Compact profile
-elseif(FEATURE_PROFILE STREQUAL "compact")
-  set(DEFINES_JERRY ${DEFINES_JERRY} CONFIG_ECMA_COMPACT_PROFILE)
+if(FEATURE_PROFILE STREQUAL "compact")
+  set(DEFINES_JERRY
+  ${DEFINES_JERRY}
+  CONFIG_ECMA_COMPACT_PROFILE
+  CONFIG_ECMA_NUMBER_TYPE=CONFIG_ECMA_NUMBER_FLOAT32)
```
tilmannOSG: Was this line added by mistake? (CONFIG_ECMA_NUMBER_TYPE=CONFIG_ECMA_NUMBER_FLOAT32)

This would give us:
   * Minimal profile: FLOAT64
   * Compact profile: FLOAT32
   * Full profile: FLOAT64

I think we want the default to be FLOAT64 in all three profiles :)

tilmannOSG:
LGTM (minus feedback above)

Can you please mention in the commit message that we're changing the default to FLOAT64 to ensure compliance with ECMAScript 5.1 which requires numbers to be represented in double precision floating-point format?

I'm thinking we should not expose FLOAT32 as a CMake option to have a higher barrier to ensure that only experienced users who understand the performance/precision trade-off of going with FLOAT32 (and are willing to sacrifice ECMAScript 5.1 conformance) will actually use this.

robertsipka:
@tilmannOSG : Thanks, I've updated this patch based on your comments.

tilmannOSG:
@robertsipka Thanks, looks good!

dbatyai:
LGTM

PR #1237, opened 2016-08-03
Fix logging of error message in `jerry_fatal`

akosthekiss:
As the comment of syscall error mentions, we should not try to log
(or print) anything in that case.


zherczeg:
LGTM

LaszloLango:
LGTM

PR #1239, opened 2016-08-04
Cleanup jerry's assert-like routines and macros

akosthekiss:
Until now, jerry had 3 different assert-like routines:
`jerry_assert_fail`, `jerry_unreachable`, and `jerry_unimplemented`,
and 3 corresponding macros (`JERRY_ASSERT`, `JERRY_UNREACHABLE`,
and `JERRY_UNIMPLEMENTED`). They had some irregularities, namely:

* All of them had a string parameter, although `jerry_unreachable`
  never got anything there but NULL.
* Both `jerry_unreachable` and `jerry_unimplemented` checked its
  string parameter for NULL, although it was always NULL for the
  first one and never NULL for the second.
* `jerry_unreachable` is just a regular assert with a fixed error
  message (i.e., control should not have got here), however, the
  expansion of its corresponding macro in debug and release modes
  differs from the behaviour of `JERRY_ASSERT`: `JERRY_ASSERT` is
  a no-op in release, however, `JERRY_UNREACHABLE` was triggering
  a crash even there.
* Moreover, `JERRY_UNIMPLEMENTED` was almost never used anymore but
  in a few places (where often an `#ifdef` selected between
  `JERRY_UNIMPLEMENTED` and `JERRY_UNREACHABLE`).

Because of the above, this patch makes the following changes:

* Drops `JERRY_UNIMPLEMENTED` completely and whereever it was still
  used, replaces it with `JERRY_UNREACHABLE`. As a consequence, the
  `jerry_unimplemented` function and the `ERR_UNIMPLEMENTED_CASE`
  fatal error code are also removed.
* Makes `JERRY_UNREACHABLE` expand to no-op in release builds.
  (Actually, to `__builtin_unreachable ()` to avoid warnings.) As
  a consequence, makes both `jerry_assert_fail` and
  `jerry_unreachable` be guarded by `#ifndef JERRY_NDEBUG`. Also,
  changes `jerry_unreachable` not to expect a string parameter.
* Rewrites `TEST_ASSERT` not to rely on `jerry_assert_fail` as
  `TEST_ASSERT` has to work in release builds as well. This also
  allows changing the error message not to mention "ICE", which
  would misleadingly suggest an assert within the engine, but
  "TEST" instead.

As a side-effect of the cleanup, some refactorings happened in
jrt.h:

* Removed the definition of the unnecessary `__extension__` macro.
* Re-used `JERRY_UNUSED` and `unlikely` where possible.
* Moved some parts of the file around.
* Fixed some comments (`/**` should only be used for the docstring
  of a single entity, for groups header comments, the regular `/*`
  should be used).


jerry-core/jrt/jrt.h
```diff
@@ -73,114 +80,57 @@
...
-#define JERRY_UNIMPLEMENTED_REF_UNUSED_VARS(comment, ...) \
-  do \
-  { \
-    if (false) \
-    { \
-      jerry_ref_unused_variables (0, __VA_ARGS__); \
-    } \
-    jerry_unimplemented (comment, NULL, NULL, 0); \
-  } while (0)
+#define JERRY_ASSERT(x) do { if (false) { JERRY_UNUSED (x); } } while (0)
```
zherczeg: You one-lined a macro here, but expanded a macro at the end. I think we should use the same style everwhere, and I prefer the later.

jerry-core/jrt/jrt.h
```diff
@@ -73,114 +80,57 @@
...
-#ifdef JERRY_ENABLE_LOG
-#define JERRY_DLOG(...) jerry_port_log (JERRY_LOG_LEVEL_DEBUG, __VA_ARGS__)
-#define JERRY_DDLOG(...) jerry_port_log (JERRY_LOG_LEVEL_TRACE, __VA_ARGS__)
-#else /* !JERRY_ENABLE_LOG */
-#define JERRY_DLOG(...) \
+#define JERRY_ASSERT(x) \
   do \
   { \
-    if (false) \
+    if (unlikely (!(x))) \
```
zherczeg: change false to unlikely? This will increase code size heavily in release.

akosthekiss: Don't let be misguided by the misalignment of the diff. Unfortunately, the diff incorrectly compares `JERRY_ASSERT` to `JERRY_LOG`, must be becose of the movements of the code. moreover, this is debug code (`#ifndef JERRY_NDEBUG`) so no effect on release, and `JERRY_ASSERT` was always using `__builtin_expect(..., 0)` in its `if`. (See the rewritten `JERRY_DLOG` at line 123.)

jerry-core/jrt/jrt.h
akosthekiss: This is intentional. Whereever the macro did something "meaningful", I did expand it. But at places where the macro is just a no-op, even if written in a complicated form, I prefer to write it in one line.

zherczeg: But the style is quite strict. Please expand it.

jerry-core/jrt/jrt.h
zherczeg: Ok.

jerry-core/jrt/jrt.h
akosthekiss: Expanded them all, maintaining the opinion that it was unnecessary and ugly.

akosthekiss:
updated

zherczeg:
LGTM

LaszloLango:
@akosthekiss, please rebase to the current master

akosthekiss:
rebased and conflicts solved

LaszloLango:
@akosthekiss, thanks for the update. LGTM

zherczeg:
LGTM

PR #1241, opened 2016-08-04
Add NULL pointer check to jmem_pools_free

grgustaf:

### DIFF
--- a/jerry-core/jmem/jmem-poolman.c
+++ b/jerry-core/jmem/jmem-poolman.c
@@ -131,6 +131,8 @@ jmem_pools_alloc (void)
 void __attribute__((hot))
 jmem_pools_free (void *chunk_p) /**< pointer to the chunk */
 {
+  JERRY_ASSERT (chunk_p != NULL);
+
   jmem_pools_chunk_t *const chunk_to_free_p = (jmem_pools_chunk_t *) chunk_p;
 
   VALGRIND_DEFINED_SPACE (chunk_to_free_p, JMEM_POOL_CHUNK_SIZE);


grgustaf:
I'm not sure that this is the right fix, but if you can afford the check, it seems like a pretty good idea. :)

I'm actually working against an older commit (3828f0) because of Zephyr build breakage, but I ran into a problem where I see jmem_pools_free getting called with a NULL pointer. This then tries to write the current free list to NULL->next_p, and sets the free list to NULL (losing all free chunks). So it showed up for me as a memory leak, which eventually fills up all my memory.

The backtrace showed line 2414 of vm.c (in vm_loop) as having made the call, but I haven't looked into it further than that:
free_left_value:
      ecma_fast_free_value (left_value);

So somehow that left_value was NULL there I guess.

Some additional background that may not be any help:

What I have going on in my code is a setInterval implementation, within the JS callback, I call a C API I've written, which in turn calls jerry_set_property to set a float property on an existing JS object. To call that, it has to do jerry_create_number and jerry_create_string first, then releases them after set_property. But if I comment out the string stuff and even the set_property, the memory leak remains. If I also comment out the number stuff, the memory leak goes away. So initially I thought this was a memory leak within number allocation, but instead for some reason it results in the main vm_loop making a bad free call.

zherczeg:
Left_value cannot be NULL, since it is an uint32 data, and 0 represent integer 0. I would like to see an example to reproduce this bug (that is needed for the fix anyway).

grgustaf:
I can understand the desire for a test case. :) I'm not sure when I'll have time to try to create though.

Sorry, I didn't realize how many levels of inline functions there are between ecma_fast_free_value and jmem_pools_free, so I don't know how it's getting called with NULL exactly. I can't print out left_value in the debugger because it's optimized out. I will try to build w/o optimization.

But here's a backtrace snippet showing the one of the times that jmem_pools_free is called with NULL:
```
#0  jmem_pools_free (chunk_p=0x0) at .../jerryscript/jerry-core/jmem/jmem-poolman.c:165
#1  0x40016150 in vm_loop (frame_ctx_p=frame_ctx_p@entry=0xa80100f4 <__TASKA_stack+1908>)
    at .../jerryscript/jerry-core/vm/vm.c:2414
#2  0x400165ff in vm_execute (frame_ctx_p=frame_ctx_p@entry=0xa80100f4 <__TASKA_stack+1908>, arg_p=arg_p@entry=0x0, 
    arg_list_len=<optimized out>, arg_list_len@entry=0) at .../jerryscript/jerry-core/vm/vm.c:2592
#3  0x400167c9 in vm_run_with_inline_stack (frame_ctx_p=frame_ctx_p@entry=0xa80100f4 <__TASKA_stack+1908>, arg_p=arg_p@entry=0x0, 
    arg_list_len=arg_list_len@entry=0) at .../jerryscript/jerry-core/vm/vm.c:2637
#4  0x40016861 in vm_run (bytecode_header_p=bytecode_header_p@entry=0xa8009940 <jmem_heap+928>, this_binding_value=<optimized out>, 
    lex_env_p=lex_env_p@entry=0xa80095c8 <jmem_heap+40>, is_eval_code=false, arg_list_p=0x0, arg_list_len=0)
    at .../jerryscript/jerry-core/vm/vm.c:2709
#5  0x400168ae in vm_run_global (bytecode_p=0xa8009940 <jmem_heap+928>)
    at .../jerryscript/jerry-core/vm/vm.c:218
#6  0x400128ce in jerry_run (func_val=func_val@entry=2818611075) at .../jerryscript/jerry-core/jerry.c:337
#7  0x40026cd1 in main (argc=<optimized out>, argv=<optimized out>) at .../main.c:68
```

grgustaf:
OK, I pragma'd out optimization in vm.c but I haven't reproduced that exact case yet. However, I now see this line in vm.c (1385 in my older commit) also resulting in jmem_pools_free(0):

```
          if (!(opcode_data & (VM_OC_PUT_STACK | VM_OC_PUT_BLOCK)))
          {
            ecma_fast_free_value (result);
```

The value of "result" was 1 and that leads to jmem_pools_free getting called with 0 apparently. I guess that makes it a float (type 1) but with a 0 pointer. Not sure how it *got* that way.

An easy test for anyone to do though is:
$ gdb <image>
(gdb) b jmem_pools_free if chunk_p == 0
(gdb) c

If that breakpoint ever fires, we've got a problem I think. You won't see any symptoms if the free list is empty (other than writing to NULL->next_p).

zherczeg:
Yes, your analysis is correct, the result cannot be 1. I think we need to figure out where that 1 is produced. I mean the free issue is just a symptom not the bug. You probably need a write watchpoint to "result" and condition to stop only if it receives the 1 value.

grgustaf:
So this probably doesn't help much, but this second one I'm able to get it to stop on vm.c:1385, is within "case VM_OC_CALL:" where it does result = *(--stack_top_p). It's getting set to one there. I presume that doesn't help much - now the question is how did that get on the stack. Any suggestion on how I'd track that down?

grgustaf:
OK, I figured out the problem. It's a bug in my code (shocking! :D)... I have an API that is still returning "true" like in the good old days of JerryScript API, instead of jerry_value_t. That's where the 1 came from.

Now... would it make sense to validate the return values of called functions better in JerryScript?

zherczeg:
Well, good question. We already made protection of passing values with error flag to property setters and function calls. But I think no amount of protection can protect from user errors, and each checks cost performance. Finding the  right amount of check is difficult.

grgustaf:
In my opinion, accepting this pull request would not be a bad thing. It's essentially just changing the (undocumented) definition of jmem_pools_free to say "it's okay to pass NULL pointers, I'll just ignore them". And that would mean that cases like this where callers don't do input validation, and can end up calling it with NULL in cases of invalid input, would effectively be doing some input validation.

Another thing to consider might be adding some input validation checks under a STRICT_CHECKING define or something, that could be turned on during development to try to catch problems like this. Would that make sense?

grgustaf:
I meant to point out that the corresponding jmem_pools_alloc has a NULL pointer check (for whether there are any free blocks) so I don't think having such a check here is a problem performance-wise.

akosthekiss:
Just askin': wouldn't an `assert (chunk_p != NULL);` work here well? In a debug build, it would bring attention to the issue, and in release, it would cost nothing.

(@grgustaf: I did not follow all the discussion, so just double checking now: you are not calling internal undocumented functions of jerry-core (like jmem_pools_free) from outside the engine, are you?)

grgustaf:
@akosthekiss, no I'm not. I just meant "documented" for internal users, like a line of text specifying the intended interface of the function.

You're right, an assert would be helpful, if nothing else. I don't think it's a big cost for some safety in release though... losing your free chunks list is just so bad. :)

zherczeg:
+1 for assert

grgustaf:
Here's that assert.

zherczeg:
LGTM

akosthekiss:
LGTM

akosthekiss:
You've got two LGTMs, please make sure that your patch is rebased on latest master.

akosthekiss:
@grgustaf ping for PR rebase

akosthekiss:
(wrong button pressed, sorry)

PR #1242, opened 2016-08-04
Refactor flag additions in CMakeLists

akosthekiss:
CMakeLists already contains macros to ease adding compilation and
warning flags. This patch:
* Ensures that they are used whereever possible.
* Adds more macros to help with other flags as well.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,14 +75,36 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/bin/")
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib/")
 
 # Compile/link flags
+# Helper macros
+macro(jerry_add_flags VAR)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -134,69 +156,50 @@ else()
 endif()
 
 # Compiler / Linker flags
-set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fno-builtin")
+jerry_add_compile_flags(-fno-builtin)
 if(("${PLATFORM}" STREQUAL "DARWIN"))
-  set(LINKER_FLAGS_COMMON "-lSystem")
+  jerry_add_link_flags(-lSystem)
```
akosthekiss: This rewrite and the rewrite 2 lines below are not equivalent to their original versions. However, I suspect that the old versions (which don't append to LINKER_FLAGS_COMMON but set it) are incorrect (as LINKER_FLAGS_COMMON might have already been set for LTO).

LaszloLango:
LGTM

zherczeg:
LGTM

PR #1244, opened 2016-08-05
Fix run-tests.py: don't override the return value of a testrun

bzsolt:

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -112,51 +112,51 @@ def create_binary(buildoptions):
     return 0
 
 def run_jerry_tests():
+    ret_build = ret_test = 0
     for job in jerry_tests_options:
-        ret = create_binary(job.build_args)
-
-        if not ret:
-            test_cmd = [TEST_RUNNER_SCRIPT, get_binary_path(job.out_dir), JERRY_TESTS_DIR]
-            if job.test_args:
-                test_cmd.extend(job.test_args)
-
-            ret = run_check(test_cmd)
-        else:
+        ret_build = create_binary(job.build_args)
+        if ret_build:
             break
 
-    return ret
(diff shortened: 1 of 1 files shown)


tools/run-tests.py
```diff
@@ -114,46 +114,49 @@ def create_binary(buildoptions):
...
-            test_cmd = [TEST_RUNNER_SCRIPT, get_binary_path(job.out_dir), JERRY_TESTS_DIR]
-            if job.test_args:
-                test_cmd.extend(job.test_args)
+        test_cmd = [TEST_RUNNER_SCRIPT, get_binary_path(job.out_dir), JERRY_TESTS_DIR]
+        if job.test_args:
+            test_cmd.extend(job.test_args)
 
-            ret = run_check(test_cmd)
-        else:
+        ret = run_check(test_cmd)
```
LaszloLango: Why not only use OR for the actual result instead of break at the first failure?

bzsolt: I think it's easier to find the failing testcases, but if you would like I change this behaviour and aggregate the return values of every steps.

akosthekiss: I second to aggregation

LaszloLango: I thinks it is not that hard. Please change if you have time for this issue.

bzsolt: It's true, but if I only changed here, the next cycle of the loop will override this aggregated value if the build success, hiding the failing testrun.
And if we also aggregate the return value of ```create_binary()``` and also remove the related ```break``` statement, in case of an unsuccessful build the script would like to run a non-existing binary.

I've updated the patch, now the testrun breaks immediately when the build was unsuccessful, otherwise aggregate the return codes of test-steps.

zherczeg:
LGTM

LaszloLango:
LGTM

PR #1246, opened 2016-08-05
Integrate JS parser module better with JRT

akosthekiss:
The fact that the JS parser had been developed as a separate
component for a while is still visible from some macros that mirror
things from JRT. This patch removes those duplicates and makes the
JS parser rely on jrt.h. (The removed macros are: `PARSER_DEBUG`,
`PARSER_INLINE`, `PARSER_NOINLINE`.)


LaszloLango:
LGTM

zherczeg:
LGTM

PR #1249, opened 2016-08-06
Follow-up refactoring of logging-related parts

akosthekiss:
This patch:
* Ensures that all calls to `jerry_port_log` in jerry-core happen
  via macros defined in jrt.h. Also, it unifies the names of those
  macros: as `JERRY_ERROR_MSG` and `JERRY_WARNING_MSG` gave a good
  pattern that was well aligned with the naming scheme of the log
  level enum, `JERRY_DLOG` and `JERRY_DDLOG` were rewritten to
  `JERRY_DEBUG_MSG` and `JERRY_TRACE_MSG`.
* Ensures that all debug logging code parts of jerry-core (i.e.,
  memory statistics, JS byte-code dumps, and RegExp byte-code
  dumps) are guarded by macros: `JMEM_STATS`,
  `PARSER_DUMP_BYTE_CODE`, and `REGEXP_DUMP_BYTE_CODE`, which in
  turn are controled by cmake build system feature flags
  `FEATURE_MEM_STATS`, `FEATURE_PARSER_DUMP`, and
  `FEATURE_REGEXP_DUMP`.
* Ensures that all debug logging functionalities can be controled
  during run time (provided that they were enabled during build
  time): the engine has `JERRY_INIT_MEM_STATS[_SEPARATE]`,
  `JERRY_INIT_SHOW_OPCODES`, `JERRY_INIT_SHOW_REGEXP_OPCODES` init
  flags, and the default unix/linux command line app has
  corresponding command line switches.`
* Drops `FEATURE_LOG`, `JERRY_ENABLE_LOG`, and
  `JERRY_INIT_ENABLE_LOG`, as their name was misleadingly general,
  even though they mostly controled the regexp engine only. The
  above-mentioned `*REGEXP*` things mostly act as their
  replacements.
* Updates build and test tool scripts and documentation.


akosthekiss:
follow up as discussed in #1205

zherczeg:
Pretty big patch. Shall we wait after the release?

akosthekiss:
We shouldn't. The `FEATURE_LOG`, `JERRY_ENABLE_LOG`, `JERRY_INIT_ENABLE_LOG` things are quite misleading. We shouldn't leave them in 1.0.

jerry-core/jerry.c
```diff
@@ -143,27 +143,36 @@ jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
...
+#endif /* !PARSER_DUMP_BYTE_CODE */
+  }
+
+  if (flags & JERRY_INIT_SHOW_REGEXP_OPCODES)
+  {
+#ifndef REGEXP_DUMP_BYTE_CODE
+    flags &= (jerry_init_flag_t) ~JERRY_INIT_SHOW_REGEXP_OPCODES;
+
+    JERRY_WARNING_MSG ("Ignoring JERRY_INIT_SHOW_REGEXP_OPCODES flag "
+                       "because of !REGEXP_DUMP_BYTE_CODE configuration.\n");
```
zherczeg: Why two lines? Only one line above.

akosthekiss: Written in one line would exceed the 120 chars/line limit.

zherczeg:
I don't see much risk here. The patch is mostly just renaming. The API change is minor, and happens before the release.

LGTM

akosthekiss:
updated to master and resolved conflicts

jerry-main/main-unix.c
```diff
@@ -187,6 +188,11 @@ main (int argc,
       flags |= JERRY_INIT_SHOW_OPCODES;
       jerry_port_default_set_log_level (JERRY_LOG_LEVEL_DEBUG);
     }
+    else if (!strcmp ("--show-regexp-opcodes", argv[i]))
+    {
+      flags |= JERRY_INIT_SHOW_REGEXP_OPCODES;
+      jerry_port_default_set_log_level (JERRY_LOG_LEVEL_DEBUG);
```
LaszloLango: Shouldn't it be JERRY_LOG_LEVEL_TRACE ?

akosthekiss: It's true that the regexp engine can log trace level messages but they are mostly "extras" and most of the "important" logs go as debug level. Thus, I wanted to keep the log level implied by "--show-regexp-opcodes" at debug level only as that seamed like a sensible default and aligns with the default implied by other command line switches. Still, should someone want to see the "extras", "--show-regexp-opcodes --log-level 3" is there to use.

LaszloLango:
LGTM

PR #1251, opened 2016-08-07
Let the build script use a default for toolchain

akosthekiss:
The `cmake` directory already contains several toolchain files for
various platforms (operating system + architecture). However,
`tools/build.py` does not define a toolchain file for cmake unless
explicitly specified. This patch changes the script to look into
the `cmake` directory for a file named
`toolchain_$(os)_$(arch).cmake` and, if found, pass that to cmake
by default.

OS and arch are determined by `os.uname()`. As Linux on Raspberry
Pi identifies itself as "armv7l", the legacy "armv7l-hf" arch name
is shortened to "armv7l". This way, building jerry on RPi
(natively, not cross) becomes possible by simply running
`tools/build.py` without any extra options.


zherczeg:
LGTM

LaszloLango:
LGTM

PR #1255, opened 2016-08-09
Fix the OSX build

akosthekiss:
Apple does not support staticaly built applications on OSX, and all
dynamically built apps have to be linked against the System lib
(i.e., `-static` should not be used and `-lSystem` is a must).
As System contains all libc and libm functions, building (and
linking) the minimal jerry-libc and jerry-libm libs makes no sense.
Moreover, if JERRY_LIBC is ON, the compiler will use the jerry-libc
headers but will link the libc functions from System, which causes
heavy confusion and segfaults at run time.

Thus, this patch changes the build system to disable the building
of jerry-libc and jerry-libm, and enables the use of system
libraries when building on OSX.

Fallout: It turned out that jerry-libm is not standard-conforming
in `pow`: `pow(1,INFINITY)` (and sign variants) should return `1`
according to the libm standard. However, the `pow` in jerry-libm
returns `NAN` - seems to be an fdlibm legacy. Interestingly, ES5.1
standard differs from libm standard when in comes to `pow` and also
expects `NaN` in this special case. And unfortunately, the
implementation of `Math.pow` simply passes its arguments to the
libm's `pow` and returns its result, which only works if jerry-libm
is used but not if a (standard-conforming) system libm is used.

As a temporary solution, this patch disables the conformance checks
of the affected cases in the test suites.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -44,6 +44,13 @@ endif()
 if("${PLATFORM}" STREQUAL "DARWIN")
   set(ENABLE_LTO "OFF")
   set(ENABLE_ALL_IN_ONE "ON")
+  set(JERRY_LIBC "OFF")
+  set(JERRY_LIBM "OFF")
+  set(COMPILER_DEFAULT_LIBC "ON")
+endif()
+
+if(JERRY_LIBC AND COMPILER_DEFAULT_LIBC)
+  message(FATAL_ERROR "JERRY_LIBC and COMPILER_DEFAULT_LIBC is enabled at the same time!")
 endif()
 
 # Status messages
@@ -105,11 +112,6 @@ set(CMAKE_C_FLAGS_RELEASE "-Os")
 # Architecture-specific compile/link flags
 jerry_add_compile_flags(${FLAGS_COMMON_ARCH})
 jerry_add_flags(CMAKE_EXE_LINKER_FLAGS ${FLAGS_COMMON_ARCH})
-
-# Use jerry libc
-if(JERRY_LIBC AND COMPILER_DEFAULT_LIBC)
-  message(FATAL_ERROR "JERRY_LIBC and COMPILER_DEFAULT_LIBC is enabled at the same time!")
-endif()
 
 # LTO
 if(ENABLE_LTO)


akosthekiss:
With this patch OSX Travis jobs go green!

zherczeg:
I am not sure the right way to support OSX is removing tests. I think we should check whether the target is OSX by compiler directives and add a separate code path for it.

akosthekiss:
I might not have been clear enough in the commit message. The core of the problem is not OSX specific. Only the OSX brought the issue to light. Our Math.pow implementation is incorrect. We get correct results only becasue our libm implementation is alsó incorrect. E.g., should we use the default libm on linux, the pow asserts would fail there as well.

akosthekiss:
For future reference:
  - math.h pow specification: http://pubs.opengroup.org/onlinepubs/9699919799/functions/pow.html#
  - ES5.1 pow specification: http://www.ecma-international.org/ecma-262/5.1/#sec-15.8.2.13

zherczeg:
I understand the problem, but disabling tests does not seem a solution. Perhaps we should fix the math function to check this corner case and does not call pow in this case. I think pow is only used in one place in Jerry..

akosthekiss:
First of all, I think that jerry-libm's pow must be fixed (actually, I'm on it). If it calls itself a libm then it should adhere the specification. Then, from that it will follow that jerry-core's Math.pow implementation will have to be changed as it will not be able to rely on the bug in jerry-libm.

The question is whether we should leave the problematic tests enabled until we fix the above, as we know that a valid build (with system libm) can fail on them.

zherczeg:
I may not understand the problem,. but according to the description, there is a corner case which does not work with standard libm. (As for me jerry libm is not a libm, as jerry libc is not a libc, so it does not need to follow other libms).

So why don't we just add:

    Math.pow C callback:
    if (corner_case) {
      return NaN;
    } else {
      return pow();
    }

This would work regardless of libm. And we can keep the tests.

akosthekiss:
> As for me jerry libm is not a libm, as jerry libc is not a libc, so it does not need to follow other libms

I disagree. jerry-libc *is* a restricted libc, i.e., when a functionality is implemented, it works like a standard libc, the only difference is that not everything is implemented that can be found in all libc's. (Actually, only those parts are implemented, which are needed by jerry-core and jerry-main.) The same should hold for jerry-libm. If it contains an implementation for a math function, it must follow the spec.

If we declared that jerry-libc and jerry-libm could deviate from the standards, that would lead to a dangerous path. That would put heavy burden on jerry-core to have different code paths/variants depending on which libc/libm are used. But jerry-core must stay libc/libm-invariant.

> according to the description, there is a corner case which does not work with standard libm

Well, I'd put it that way: there is at least one difference in the spec of pows for C an JS, which got revealed now. As the wording of the specs differs a bit, it needs some comparison work to confirm that this is the only difference.

zherczeg:
Consider printf for jerry libc: it supports only a subset of the standard printf format specifiers, and has non-standard behaviour when you use a flag which is not supported. What is the difference?

akosthekiss:
What do you mean by non-standard behaviour?

zherczeg:
case LIBC_PRINTF_ARG_LENGTH_TYPE_HIGHL:
    {
      assert (!"unsupported length field L");
    }

I don't think a printf should crash at all.

akosthekiss:
BTW, there is one more issue in pow, just found it:
```
jerry> Math.pow(1,NaN)
1
jerry> Math.pow(2,NaN)
NaN
```
According to Ecma spec, both should be NaN. But for some reason, 15.08.02.13-001.js tests the second case only and so passes.

akosthekiss:
The assert failure is not a valid counter example. That assert states: "this is not implemented". jerry-libc's printf is a restricted printf, but whatever is implemented, conforms to the fully implemented printf. If you hit the assert, that means that you tried to reach beyond the restrictions.

This does not hold for pow. It implements something, does return a result *but* that result is not standard conforming.

zherczeg:
But it implements something: a crash. Please don't think jerry libc/libm as a restricted libc/libm. These are just libraries which has some in common with libc/libm, but they have a lot of freedom.

Anyway the fix should be in the ecma-builtin-math.c / ecma_builtin_math_object_pow special casing the issue. So jerry-libm will never be called with those arguments anymore, so it can be unimplemented (assert?).

akosthekiss:
No. Heavily disagree. A crash is a very serious signal of misuse. It gives no result but aborts execution. Please don't think that jerry libc/libm have a lot of freedom. The only freedom they have is that they don't have to implement everything.

But let's go a bit more formal, so that the message gets through:
restricted(r,b) iff foreach x: computes(r,x) => r(x) == b(x)
(where computes(r,x) iff r terminates correctly on input x)

And now:
restricted(jerry-libc,libc)
!restricted(jerry-libm,libm)  <--- This is a big problem.

And putting assert into jerry-libm's pow would not make much sense. It should be much cleaner to return the right value instead.

zherczeg:
But the "right" value is not what ES expects, if I understand you correctly. Then adding this support would only be dead code (which increase binary size), since we need special casing the issue anyway.

akosthekiss:
If #1258 gets accepted/landed, that solves the pow issue and I'll be able to update this PR afterwards not to change the tests.

akosthekiss:
Updated to master. Now that #1258 got landed, the tests don't need to be touched and the patch shrinks to a simple CMakeLists modification.

LaszloLango:
LGTM

zherczeg:
LGTM

PR #1256, opened 2016-08-09
Fix the use of mktemp in tool scripts

akosthekiss:
The manual of `mktemp` states that "TEMPLATE must contain at least
3 consecutive 'X's in last component." Linux implementation seems
to be relaxed about this and accepts and rewrites X's even inside
the template, but mktemp of OSX is more strict and handles trailing
X's only. This patch makes sure that mktemp templates work on both
OS's.


zherczeg:
LGTM

LaszloLango:
LGTM

PR #1259, opened 2016-08-10
Remove needless toolchain files and variable settings.

robertsipka:
Unnecessary to set CMAKE_SYSTEM_NAME and  CMAKE_SYSTEM_PROCESSOR set, these values are set by default.
Its also able to use any C99-compliant compiler bye use the default CC enviroment variable or set the
CMAKE_C_COMPILER value using the build script cmake-param argument.


zherczeg:
LGTM

akosthekiss:
LGTM

robertsipka:
Thanks, I've rebased it with master.

PR #1260, opened 2016-08-11
targets/zephyr/Makefile.zephyr: Use zephyr_getline module for line input.

pfalcon:
The original implementation used shell facility, but it was designed for
a unix shell like input, and automatically tokenized it into
space-separated "words", with limit of 10 (i.e. 9 spaces per line). For
JavaScript input, it is quite easy to have more than 9 spaces per line,
and get error:

Too many parameters (max 10)

After consultation with upstream
(https://jira.zephyrproject.org/browse/ZEP-532) it was decided that the
best approach is to skip using shell facility and use Zephyr console
facility. That however requires some Zephyr-specific boilerplate code.
This code was implemented as reusable modules in
https://github.com/pfalcon/zephyr_console_helpers repository, to be
usable for other console-based projects too. zephyr_getline.h/c in this
commits are direct imports from this repository.

### DIFF
--- a/targets/zephyr/src/Makefile
+++ b/targets/zephyr/src/Makefile
@@ -19,4 +19,4 @@ endif
 # Adding path for jerry script APIs
 ZEPHYRINCLUDE += -I$(JERRY_INCLUDE)
 
-obj-y += main-zephyr.o
+obj-y += main-zephyr.o getline-zephyr.o
--- a/targets/zephyr/src/getline-zephyr.c
+++ b/targets/zephyr/src/getline-zephyr.c
@@ -1,0 +1,58 @@
+/*
+ * Copyright (c) 2016 Linaro
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <zephyr.h>
+#include <uart.h>
+#include <drivers/console/uart_console.h>
+#include "getline-zephyr.h"
+
+/* While app processes one input line, Zephyr will have another line
+   buffer to accumulate more console input. */
+static struct uart_console_input line_bufs[2];
+
+static struct nano_fifo free_queue;
+static struct nano_fifo used_queue;
+
+char *zephyr_getline(void)
+{
+  static struct uart_console_input *cmd;
+
+  /* Recycle cmd buffer returned previous time */
+  if (cmd != NULL)
+  {
+    nano_fifo_put(&free_queue, cmd);
+  }
+
+  cmd = nano_fifo_get(&used_queue, TICKS_UNLIMITED);
+  return cmd->line;
+}
+
+void zephyr_getline_init(void)
+{
+  int i;
+
+  nano_fifo_init(&used_queue);
+  nano_fifo_init(&free_queue);
+  for (i = 0; i < sizeof(line_bufs) / sizeof(*line_bufs); i++)
+  {
+    nano_fifo_put(&free_queue, &line_bufs[i]);
+  }
+
+  /* Zephyr UART handler takes an empty buffer from free_queue,
+     stores UART input in it until EOL, and then puts it into
+     used_queue. */
+  uart_register_input(&free_queue, &used_queue, NULL);
+}
--- a/targets/zephyr/src/main-zephyr.c
+++ b/targets/zephyr/src/main-zephyr.c
@@ -20,119 +20,19 @@
 #include <zephyr.h>
 #include <misc/printk.h>
 #include <misc/shell.h>
+#include "getline-zephyr.h"
 
 #include "jerry-api.h"
 
-#if defined (CONFIG_STDOUT_CONSOLE)
-#include <stdio.h>
-#define PRINT       printf
-#else
-#include <misc/printk.h>
-#define PRINT       printk
-#endif
-
-static char *source_buffer = NULL;
-static unsigned char flags = 0;
 static jerry_value_t print_function;
 
-#define VERBOSE 0x01
-
-/**
- * Jerryscript simple test loop
- */
-int jerryscript_test ()
+static int shell_cmd_handler (char *source_buffer)
 {
-  jerry_value_t ret_val;
-
-  const char script[] =
-  "var test=0; " \
-  "for (var t=100; t<1000; t++) test+=t; " \
-  "print ('Hi JS World! '+test);";
-
-  printf ("Script [%s]\n", script);
-  ret_val = jerry_eval ((jerry_char_t *) script,
-      strlen (script),
-      false);
-
-  return jerry_value_has_error_flag (ret_val) ? -1 : 0;
-} /* jerryscript_test */
-
-
-static int shell_cmd_verbose (int argc, char *argv[])
-{
-  printf ("Enable verbose \n");
-  flags |= VERBOSE;
-  return 0;
-} /* shell_cmd_verbose */
-
-
-static int shell_cmd_syntax_help (int argc, char *argv[])
-{
-  printf ("version jerryscript & zephyr versions\n");
-  return 0;
-} /* shell_cmd_syntax_help */
-
-
-static int shell_cmd_version (int argc, char *argv[])
-{
-  uint32_t version = sys_kernel_version_get ();
-
-  printf ("Jerryscript API %d.%d\n", JERRY_API_MAJOR_VERSION, JERRY_API_MINOR_VERSION);
-
-  printk ("Zephyr version %d.%d.%d\n", SYS_KERNEL_VER_MAJOR (version),
-    SYS_KERNEL_VER_MINOR (version),
-    SYS_KERNEL_VER_PATCHLEVEL (version));
-  return 0;
-} /* shell_cmd_version */
-
-
-static int shell_cmd_test (int argc, char *argv[])
-{
-  return jerryscript_test ();
-} /* shell_cmd_test */
-
-
-static int shell_cmd_handler (int argc, char *argv[])
-{
-  if (argc <= 0)
-  {
-    return -1;
-  }
-
-  unsigned int size = 0;
-  for (int t = 0; t < argc; t++)
-  {
-    size += strlen (argv[t]) + 1;
-  }
-
-  source_buffer = (char *) malloc (size);
-
-  char *d = source_buffer;
-  unsigned int len;
-
-  for (int t = 0; t < argc; t++)
-  {
-    len = strlen (argv[t]);
-    memcpy (d, argv[t], len);
-    d += len;
-    *d = ' ';
-    d++;
-  }
-
-  * (d - 1) = '\0';
-
-  if (flags & VERBOSE)
-  {
-    printf ("[%s] %lu\n", source_buffer, strlen (source_buffer));
-  }
-
   jerry_value_t ret_val;
 
   ret_val = jerry_eval ((jerry_char_t *) source_buffer,
     strlen (source_buffer),
     false);
-
-  free (source_buffer);
 
   if (jerry_value_has_error_flag (ret_val))
   {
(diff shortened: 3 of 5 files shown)


pfalcon:
@sergioamr : Well, so I guess enough uncertainty, we've got to make this change ;-). I've coded minimal versions of getchar()/gets() like functions in https://github.com/pfalcon/zephyr_console_helpers , to be reusable for other projects too.

There're more things to remove from shell API based implementation, submitting this interim patch to get an ack before doing more work.

targets/zephyr/src/zephyr_getline.c
```diff
@@ -0,0 +1,56 @@
...
+/* While app processes one input line, Zephyr will have another line
+   buffer to accumulate more console input. */
+static struct uart_console_input line_bufs[2];
+
+static struct nano_fifo free_queue;
+static struct nano_fifo used_queue;
+
+char *zephyr_getline(void)
+{
+	static struct uart_console_input *cmd;
```
akosthekiss: Don't we need to initialize this to NULL for the first time?

targets/zephyr/src/Makefile
```diff
@@ -19,4 +19,4 @@ endif
 # Adding path for jerry script APIs
 ZEPHYRINCLUDE += -I$(JERRY_INCLUDE)
 
-obj-y += main-zephyr.o
+obj-y += main-zephyr.o zephyr_getline.o
```
akosthekiss: Nitpicking: inconsistent naming scheme of files... dash or underscore? starts with zephyr or ends with zephyr?

(Note: the third filename in the src dir is also somewhat misleading, jerry-entry.c contains only i/o port api implementation.)

targets/zephyr/src/main-zephyr.c
```diff
@@ -186,11 +150,13 @@ void main (void)
...
 
-  shell_register_app_cmd_handler (shell_cmd_handler);
-  shell_init ("js> ", commands);
-  /* Don't call jerry_cleanup() here, as shell_init() returns after setting
-     up background task to process shell input, and that task calls
-     shell_cmd_handler(), etc. as callbacks. This processing happens in
-     the infinite loop, so JerryScript doesn't need to be de-initialized. */
+  while (1) {
+    char *s;
+    printf("js> "); fflush(stdout);
```
akosthekiss: (style) one statement per line

targets/zephyr/src/main-zephyr.c
```diff
@@ -186,11 +150,13 @@ void main (void)
     printf ("Error: could not look up print function, expression results won't be printed\n");
   }
 
-  shell_register_app_cmd_handler (shell_cmd_handler);
-  shell_init ("js> ", commands);
-  /* Don't call jerry_cleanup() here, as shell_init() returns after setting
-     up background task to process shell input, and that task calls
-     shell_cmd_handler(), etc. as callbacks. This processing happens in
-     the infinite loop, so JerryScript doesn't need to be de-initialized. */
+  while (1) {
```
akosthekiss: (style) put curly brace on a new line

targets/zephyr/src/zephyr_getline.c
```diff
@@ -0,0 +1,56 @@
...
+
+	/* Recycle cmd buffer returned previous time */
+	if (cmd != NULL) {
```
akosthekiss: (sytle) General comment to this whole file: it does not follow project style guidelines (2 space indentation, curly braces on separate lines, space before parentheses, etc.) Being a port implementation, I'm tempted not to be too picky, but the other files in this port do follow the style, so I guess this one should do too.

targets/zephyr/src/zephyr_getline.c
pfalcon: ANSI C guarantees it to be NULL. If it's initialized to NULL, it goes to .data section instead of (virtual) .bss section, growing firmware image size, ultimately with a chance to overflow it for example from 128K-2 to 128K+2, meaing that it no longer can run on an MCU with 128K of FlashROM.

targets/zephyr/src/zephyr_getline.c
pfalcon: As mentioned in the commit message, this file is a direct import from another project, zephyr_console_helpers, which tries to follow Zephyr codestyle (in the hope that it may be eventually merged in Zephyr tree). I can reformat it per JrS style guide, but that may only complicated future syncs from zephyr_console_helpers (these aren't expected to be frequent, and expected reason would be Zephyr API change for example).

targets/zephyr/src/zephyr_getline.c
akosthekiss: OK with me then

targets/zephyr/src/Makefile
pfalcon: It similar matter to codestyle mentioned above -  zephyr_getline.o tries to follow zephyr's conventions, which uses underscores in filenames. I agree it's inconsistent, but it'a about same kind of inconsistency as, for example mbed port, which require C++ at all. I can change it if JerryScript maintainers between "ease of pulling updates from upstream" and "following JrS own conventions" choose the latter.

targets/zephyr/src/main-zephyr.c
pfalcon: fixed

targets/zephyr/src/main-zephyr.c
pfalcon: fixed

targets/zephyr/src/main-zephyr.c
```diff
@@ -186,11 +150,14 @@ void main (void)
...
-     shell_cmd_handler(), etc. as callbacks. This processing happens in
-     the infinite loop, so JerryScript doesn't need to be de-initialized. */
-} /* main */
+  while (1)
+  {
+    char *s;
+    printf("js> ");
+    fflush(stdout);
+    s = zephyr_getline ();
+    shell_cmd_handler (s);
```
akosthekiss: This now calls the `shell_cmd_handler` function directly and leaves no chance for `shell_cmd_syntax_help`, `shell_cmd_version`, `shell_cmd_test`, or `shell_cmd_verbose` to be called. Will these functions remain dead code (and the `commands` array dead data)?

targets/zephyr/src/zephyr_getline.c
akosthekiss: I'm a bit undecided, so let others speak up, too.

sergioamr:
After fixing the style issues, it's LGTM 

I don't understand why i didnt' do it that way in the first place :-/. So thanks!

One caveat that the uart driver has, is that you cannot send too much data in one go since it will start dropping it and you will lose text really fast. So no cut & paste from a terminal.

That is why i have to do a full implementation on my ihex reader project using the USB ACM driver.

My use case involves not having lines in the future and trying to stream as fast as i can to the device in different data interfaces with uglify Javascript and Snapshots.

On my case, it really feels like a tiny computer now :)
https://www.youtube.com/watch?v=L2pF48BOPBc

targets/zephyr/src/main-zephyr.c
sergioamr: The test is something i quite use for automatic testing and show that Javascript works without having to type a long function. But i am not particularly attached to any code there :D

pfalcon: @akosthekiss : That's kind of "yet to be done" things I mentioned in the first comment. As @sergioamr is positive about this change, I'll post an updated version soon which removes all unused functions.

pfalcon: @sergioamr : Yep, that's kind of regression, and why I didn't post such patch before ;-). Well, we need to do something about 9-space limitation, and those commands were adhoc anyway (not available in other ports, what to do if there's a variable named "test", etc.). I'll update README, and simple test statement will be there for quick copy-paste.

pfalcon:
> I don't understand why i didnt' do it that way in the first place :-/. So thanks!

Well, I'd say using Zephyr shell was a natural first step, and I'm happy that raised some discussion in Zephyr tracker/mailing list first on how to proceed further.

> One caveat that the uart driver has, is that you cannot send too much data in one go since it will start dropping it and you will lose text really fast. So no cut & paste from a terminal.

I tested it on FRDM-K64F and well, normal-sized paste, e.g. your test statement, works for me. However, if I try to paste something long (or just press and keep Ctrl+V repeat a paste), it stop accepting chars, apparently after 256 chars, which is a size of Zephyr (non-configurable) line buffer. Only board reset helps then. Well, this is definitely another Zephyr issue, which needs to be explored further (and why I wanted to keep Zephyr-specific stuff in a separate module, to patch only it).

> On my case, it really feels like a tiny computer now :) https://www.youtube.com/watch?v=L2pF48BOPBc

Looks nice!

sergioamr:
@pfalcon Check this report from me:
https://jira.zephyrproject.org/browse/ZEP-467

That is exactly why i did have to rewrite my uart interaction on my ihex uploader to be able to dynamically allocate memory and not hit the memory limit.

If you see the problem still happening, you could answer back and say that you can see the issue in a different platform. I don't mind it losing data, but the crashing is not good.

pfalcon:
Ok, the patch updated, main-zephyr.c should be fully cleaned up now. README.md was also updated. The only remaining issue is codestyle of zephyr_getline.c - to remind, that was taken as is from https://github.com/pfalcon/zephyr_console_helpers , which follows Zephyr code conventions.

targets/zephyr/src/main-zephyr.c
```diff
@@ -159,21 +59,16 @@ static int shell_cmd_handler (int argc, char *argv[])
...
-  SHELL_COMMAND (NULL, NULL)
-};
-
-
 void main (void)
 {
-  printf ("Jerry Compilation " __DATE__ " " __TIME__ "\n");
+  uint32_t zephyr_ver = sys_kernel_version_get ();
+  printf ("JerryScript build: " __DATE__ " " __TIME__ "\n");
+  printf ("Jerryscript API %d.%d\n", JERRY_API_MAJOR_VERSION, JERRY_API_MINOR_VERSION);
```
akosthekiss: Why not JerryScript as one line above?

akosthekiss:
The patch looks quite OK to me now. The only remaining question is the coding style. As this is JerryScript repository, do we require all port code to conform to Jerry coding style? Or, as port code is often integrated in a project with different coding stlye, is it OK to let it follow the integrator's style? But then, is it OK to have mixed style code in a port directory? Opinions? @zherczeg

zherczeg:
I think the rule is "allowed" but not "preferred". Mixing is disallowed, so each port must follow one coding style.

targets/zephyr/src/main-zephyr.c
pfalcon: Oops, copy-paste from previous code, meant to fix it, but forgot.

pfalcon:
Ok, given the codestyle comments above, I've renamed zephyr_getline.[ch] to getline-zephyr.[ch], which is consistent with existing main-zephyr.c. The content of it still stays as provided by the upstream project, https://github.com/pfalcon/zephyr_console_helpers, to facilitate future syncing. Let me know if that's acceptable compromise.

pfalcon:
@akosthekiss, @zherczeg, @LaszloLango : Can you please have a look at the updated patch and see if it's good enough or need more work? Thanks.

zherczeg:
It is ok for me. LGTM

targets/zephyr/src/getline-zephyr.c
```diff
@@ -0,0 +1,56 @@
...
+/* While app processes one input line, Zephyr will have another line
+   buffer to accumulate more console input. */
+static struct uart_console_input line_bufs[2];
+
+static struct nano_fifo free_queue;
+static struct nano_fifo used_queue;
+
+char *zephyr_getline(void)
+{
+	static struct uart_console_input *cmd;
```
LaszloLango: Please avoid tab characters. Use 2 space indentation just like in the "old" code of zephyr port.

sergioamr:
Hi @pfalcon 

I modified the check-vera.sh on my first commit to be able to test on any folder and adjust the style:
https://github.com/Samsung/jerryscript/blob/master/tools/check-vera.sh

Now in retrospective i realise that the return parameter should be on the next line but vera did not pick up on that when i first did the port. :?

targets/zephyr/src/getline-zephyr.c
pfalcon: Ok, getline-zephyr.c reformatted to JrS codestyle.

akosthekiss:
LGTM

(BTW, @sergioamr , would you like Zephyr port to be bound by JerryScript guidelines? It might be possible to add the Zephyr target directory to the vera and cppcheck runner scripts. But then there is no excuse, the port code must really conform :) . Note: there is no previous example for that, i.e., a port to be checked automatically. @zherczeg, @LaszloLango: Any thoughts on that?)

sergioamr:
@akosthekiss @poussa Well, is up to you guys. If you enforce the ports to be on the particular coding style. I am not too picky myself, years of IDE auto-format makes a person very relax on coding style :D. 

I though there were talks about moving the targets outside JerryScript, so it really depends on that move. 

If they are part of JS, they should look like JS. But I understand that having code from two systems with different coding style makes a difficult decision on which one to use for which file.

zherczeg:
No ports checked before.

akosthekiss:
Yeah, multiple sources make life hard. Was just a wild idea.

pfalcon:
> Yeah, multiple sources make life hard. Was just a wild idea.

Exactly, and that's why I, well, spent some time in "discussion" of this issue above. Just imagine for example of a bare-metal Cortex-M port. It would require a CMSIS library from MCU vendor, and there's no sense in trying to reformat that. It's of course not clear if such port would be mergeable, but illustrates an idea that when dealing with dependencies, it's probably more productive to accept pluralism. (I of course don't discuss the need for consistent codestyle in project's own code, that's obvious.)

akosthekiss:
We've had such ports (for STM32Fx boards). They required a bunch of vendor-supplied code, indeed. However, actually, those libraries were not incorporated into the JerryScript repository but a script downloaded them into the `third-party` directory at build time. (And that was not only because of style but also because of licensing differences.)

Anyway, this debate does not bring much gain anymore. The PR got two LGTMs, so it's ready for merging. Please, make sure that your patch is rebased on latest master.

pfalcon:
Rebased, thanks.

PR #1265, opened 2016-08-13
Add target build option to CMakeLists.txt

seanshpark:
Adding EXTERNAL_BUILD option for targets build as
previous EXTERNAL_BUILD_ENTRY_FILE is removed
so that we can build target specific codes calling jerry interfaces.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -226,3 +226,8 @@ endif()
 if(UNITTESTS)
   add_subdirectory(tests/unit)
 endif()
+
+# External target build
+if(EXTERNAL_BUILD)
+  add_subdirectory(${EXTERNAL_BUILD})
+endif()


seanshpark:
@akosthekiss , @robertsipka PTAL
I'm working on fixing esp8266 target build and with this it can be done.

zherczeg:
LGTM

akosthekiss:
@seanshpark I'll have to take a closer look at the esp8266 target, but I have some slight doubts.
  - If we go this path, the CMakeLists file has to be updated at the beginning as well. We declare and document all config variables at the start of the file (e.g., `set(UNITTESTS     OFF CACHE BOOL "Use unit tests?")`) and print them as well so that we always know what our configuration is (e.g., `message(STATUS "UNITTESTS                 " ${UNITTESTS})`).
  - However, I'm not sure we have to go this path. Why do we have to inject anything into the build system of Jerry? The port has components that need to be built on their own anyway, doesn't it? Why can't those parts be built there, in that directory, with the port's build system? (But I admit, I'm lacking some context, I'll have to look up the esp8266 target better.)

seanshpark:
@akosthekiss , well, as I wrote, `EXTERNAL_BUILD_ENTRY_FILE` is not available anymore. This variable was used to build files with same setting that Jerry is built that call JerryScript APIs but exist in targets folder. I mean, I would like to build and drop a library along with Jerry build. 

> If we go this path, the CMakeLists file has to be updated at the beginning as well. We declare and document all config variables at the start of the file (e.g., set(UNITTESTS OFF CACHE BOOL "Use unit tests?")) and print them as well so that we always know what our configuration is (e.g., message(STATUS "UNITTESTS " ${UNITTESTS})).

Sorry but I can't catch what this is. Can you point to some code or document? I don't think my patch affects this much... :)

> However, I'm not sure we have to go this path. Why do we have to inject anything into the build system of Jerry?

Reason is what I wrote above. There was `EXTERNAL_BUILD_ENTRY_FILE` and now it's gone.

> The port has components that need to be built on their own anyway, doesn't it? 

Some but not all... Something like this... (1) JerryScript output libraries (2) target library that calls JerryScript APIs and exposes some APIs, (3) target codes that call (2). Purpose is to build (2). 

> Why can't those parts be built there, in that directory, with the port's build system? (But I admit, I'm lacking some context, I'll have to look up the esp8266 target better.)

When I first wrote targets codes, when (2) and (3) was one, build didn't go well.

seanshpark:
I would like to check again building (2) and (3) at once. If that works, I don't need this patch.

seanshpark:
I get this kind of error with building (2) files inside (3).
```
In file included from ../../../jerry-core/jerry-api.h:22:0,
                 from jerry_extapi.c:23:
/opt/Espressif/ESP8266_SDK/extra_include/stdint.h:84:20: error: conflicting types for 'int32_t'
 typedef signed int int32_t;
                    ^
In file included from /opt/Espressif/ESP8266_SDK/include/espressif/esp_common.h:9:0,
                 from jerry_extapi.c:19:
/opt/Espressif/ESP8266_SDK/include/espressif/c_types.h:17:29: note: previous declaration of 'int32_t' was here
 typedef signed long         int32_t;
                             ^
```

But can handle these problems with giving some defines , for example in `jerry_run.c`
```
//#include "esp_common.h"
#define TENSILICA 1
#define __have_long64 0
#define __int_fast64_t_defined 1
#include <stdlib.h>
#include <stdio.h>
```

Simple blinking with Jerry seems working.

@akosthekiss, "The port has components that need to be built on their own anyway, doesn't it?" seems sort of working, but not sure about the define values.  I'll follow your decision.

akosthekiss:
@seanshpark The error log you've copied looks very strange to me. Do I get it right that the ESP8266 SDK has two headers (extra_include/stdint.h and include/espressif/esp_common.h) that provide conflicting definitions for int32_t? This almost sounds like an SDK error. Or do I get something wrong? (Unfortunately, I don't have the SDK on my box right now.)

zherczeg:
Is there any plans with this patch?

zherczeg:
Friendly ping :)

seanshpark:
Sorry for late responding. I may have to close this PR. It needs time to look inside this problem but my situation doesn't let me to do so.

PR #1269, opened 2016-08-15
Fix JMEM_HEAP_END_OF_LIST for pointers larger than 32 bits

akosthekiss:
For 64 bit pointers, 0xffffffff as end-of-list marker pointer does
not work, as the marker is expected to be greater than the start
address of the heap area (and also max 0xffffffff away), but on 64
bit systems, the heap can start on higher addresses, i.e., above
0x100000000. This patch ensures that the marker is set correctly on
all systems.

### DIFF
--- a/jerry-core/jmem/jmem-heap.c
+++ b/jerry-core/jmem/jmem-heap.c
@@ -87,7 +87,7 @@ void jmem_heap_valgrind_freya_mempool_request (void)
 /**
  * End of list marker.
  */
-#define JMEM_HEAP_END_OF_LIST ((jmem_heap_free_t *const) ~((uint32_t) 0x0))
+#define JMEM_HEAP_END_OF_LIST ((uint32_t) 0xffffffff)
 
 #if UINTPTR_MAX > UINT32_MAX
 #define JMEM_HEAP_GET_OFFSET_FROM_ADDR(p) ((uint32_t) ((uint8_t *) (p) - JERRY_HEAP_CONTEXT (area)))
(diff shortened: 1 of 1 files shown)


akosthekiss:
Interestingly, this issue only appeared on OSX and did not occur on x64/Linux. When running `./build/bin/jerry tests/jerry/parser-oom.js`, jerry crashed with the following message and stack trace:

```
ICE: Assertion 'jmem_is_heap_pointer (current_p)' failed at jerry-core/jmem/jmem-heap.c(jmem_heap_alloc_block_internal):242.
```

```
#0  jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerry-core/jrt/jrt-fatals.c:37
#1  0x0000000100001eae in jerry_assert_fail (
    assertion=0x1000d666d "jmem_is_heap_pointer (current_p)", 
    file=0x1000ce9bf "jerry-core/jmem/jmem-heap.c", function=0x1000d662a "jmem_heap_alloc_block_internal", line=242)
    at jerry-core/jrt/jrt-fatals.c:86
#2  0x00000001000ba9f6 in jmem_heap_alloc_block_internal (size=49168)
    at jerry-core/jmem/jmem-heap.c:242
#3  0x00000001000942c7 in jmem_heap_gc_and_alloc_block (size=49168, 
    ret_null_on_error=true)
    at jerry-core/jmem/jmem-heap.c:346
#4  0x000000010009ee4a in jmem_heap_alloc_block_null_on_error (size=49168)
    at jerry-core/jmem/jmem-heap.c:408
#5  parser_malloc (context_p=0x7fff5fbfec30, size=49168)
    at jerry-core/parser/js/js-parser-mem.c:42
```

After instrumenting the loop of the slow path of `jmem_heap_alloc_block_internal` with `printf`s, I got the following output:

```
end_of_list	= 0xffffffff
current_p	= 0x104f06278
current_p	= 0x104f13fc0
current_p	= 0x104f7a340
current_p	= 0x1ffffffff
```

Clearly, we wanted to put the EOL marker into `current_p` just didn't succeed: heap was on high addresses, EOL address was lower, when the offset of the EOL was calculated (with `JMEM_HEAP_GET_OFFSET_FROM_ADDR`), it was truncated to 32 bits, and when it was transformed back to address again (with JMEM_HEAP_GET_ADDR_FROM_OFFSET`), it became different.

LaszloLango:
LGTM

zherczeg:
I don't think this is the right approach. I think we should change

    JMEM_HEAP_GET_OFFSET_FROM_ADDR (JMEM_HEAP_END_OF_LIST)

to

    JMEM_HEAP_END_OF_LIST

There is one case where a few line needs to be changed as well.

akosthekiss:
patch updated according to the review (all CI jobs are green, OS X included!)

zherczeg:
I like this much better, LGTM.

PR #1272, opened 2016-08-17
Force to use SSE in 32-bit environment instead of the 387 unit

bzsolt:
This patch fixes the failing unittests, which fails in 32-bit mode.
These compiler-options adjusted through the default toolchain file.


zherczeg:
LGTM

LaszloLango:
LGTM

Issue #1274, opened 2016-08-18
Large string truncated without error


svaarala:
In trying to do some performance testing I noted that when trying to build a large string it gets silently truncated to zero bytes:

```
var txt = 'xy';
for (var i = 0; i < 20; i++) {
    print(i, txt.length);
    txt = txt + txt;
}
print(txt.length);
```

This should print:

```
0 2
1 4
2 8
3 16
4 32
5 64
6 128
7 256
8 512
9 1024
10 2048
11 4096
12 8192
13 16384
14 32768
15 65536
16 131072
17 262144
18 524288
19 1048576
2097152
2097152
```

But prints:

```
0 2
1 4
2 8
3 16
4 32
5 64
6 128
7 256
8 512
9 1024
10 2048
11 4096
12 8192
13 16384
14 32768
15 0
16 0
17 0
18 0
19 0
0
```

This is probably caused by an internal limitation on string length? What was confusing and is a bit dangerous is that the result is silently truncated rather than triggering a `RangeError`.

What seems to happen is an overflow in the length field because the result is not always zero length. For example:
```
x = 'x'; while (x.length < 65530) x += 'y';
print(x.length);
x += 'dummydummy';
print(x.length);
print(x);
```

This prints:
```
65530
4
xyyy
```

So that the contents are from the beginning of the original string before appending.

Assuming this is a simple 16-bit overflow, the string append processing should check for overflow and cause a RangeError if there would be an overflow. Without the error the behavior is a bit dangerous; in performance testing it caused the test to run but with incorrect inputs, for instance.

zherczeg:
Thank you for the report, I will fix this in due course. There was only an assert for this.

zherczeg:
Since 1275 is accepted, we can close this bug.

svaarala:
Just to be clear, the 16-bit string length limitation is quite understandable for low memory environments, but the silent length wrapping is dangerous because a script can proceed silently with incorrect data. Typically engines throw e.g. a `RangeError` when encountering internal limits and I think that would be appropriate here.

PR #1275, opened 2016-08-19
Add long string support.

zherczeg:
An option for fixing the string issue: support strings with uint32 length. Not necessarily the best option.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -809,7 +809,10 @@ typedef struct
  */
 typedef enum
 {
-  ECMA_STRING_CONTAINER_HEAP_UTF8_STRING, /**< actual data is on the heap as an utf-8 (cesu8) string */
+  ECMA_STRING_CONTAINER_HEAP_UTF8_STRING, /**< actual data is on the heap as an utf-8 (cesu8) string
+                                           *   maximum size is 2^16. */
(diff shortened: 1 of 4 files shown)


LaszloLango:
Is this related to #1274?

zherczeg:
Kind of an alternative. Not sure this is the best approach, so this is just a proposal.

zherczeg:
The performance change is just noise, but the binary size is increased by 600 bytes (156952->157512). Question is, do we want this?

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -872,32 +875,33 @@ typedef struct ecma_string_t
...
+    lit_magic_string_ex_id_t magic_string_ex_id; /**< identifier of an external magic string */
+    ecma_value_t lit_number; /**< literal number (note: not a regular string type) */
+    uint32_t common_field; /**< for zeroing and comparison in some cases */
   } u;
 } ecma_string_t;
 
 /**
+ * Long ECMA string-value descriptor
+ */
+typedef struct ecma_long_string_t
```
LaszloLango: Please write only `typedef struct`

LaszloLango:
I think we need this change. LGTM after the style fix.

dbatyai:
This will surely be useful and in my opinion the trade-off is definitely worth it. LGTM

PR #1276, opened 2016-08-22
Regexp.prototype.exec should generate return array with [[DefineOwnProperty]], not [[Put]]

yichoi:
The spec (ES5-15.10.6.2) explicitly states that it uses [[DefineOwnProperty]], not [[Put]].

Releated issue: https://github.com/Samsung/jerryscript/issues/1078
In this case, the behavior of two operations are different,
because of the indexed property previously defined on Array.prototype.

(And the last line of 'tests/jerry/regression-test-issue-787.js' should not throw TypeError.)

### DIFF
--- a/jerry-core/ecma/operations/ecma-regexp-object.c
+++ b/jerry-core/ecma/operations/ecma-regexp-object.c
@@ -1434,15 +1434,13 @@ ecma_regexp_exec_helper (ecma_value_t regexp_value, /**< RegExp object */
           capture_value = ecma_make_string_value (capture_str_p);
         }
 
-        ECMA_TRY_CATCH (put_value,
-                        ecma_op_object_put (result_array_obj_p,
-                                            index_str_p,
-                                            capture_value,
-                                            true),
-                        ret_value);
-        ECMA_FINALIZE (put_value);
+        ecma_property_t *prop_p;
+        prop_p = ecma_create_named_data_property (result_array_obj_p,
+                                                  index_str_p,
+                                                  ECMA_PROPERTY_CONFIGURABLE_ENUMERABLE_WRITABLE);
+        ecma_set_named_data_property_value (prop_p, capture_value);
 
-        ecma_free_value (capture_value);
+        JERRY_ASSERT (!ecma_is_value_object (capture_value));
         ecma_deref_ecma_string (index_str_p);
       }
 
--- a/tests/jerry/regression-test-issue-1078.js
+++ b/tests/jerry/regression-test-issue-1078.js
@@ -13,6 +13,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-Array.prototype.push(Math.sin);
+Array.prototype.splice(Function.prototype, 1, this);
 Object.freeze(Array.prototype);
-String.prototype.match(String.prototype);
+var res = (new String("Hello")).split(new RegExp());
+assert(res.length == 5);


jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1434,10 +1434,13 @@ ecma_regexp_exec_helper (ecma_value_t regexp_value, /**< RegExp object */
...
+        ecma_property_descriptor_t array_item_prop_desc = ecma_make_empty_property_descriptor ();
+        array_item_prop_desc.is_value_defined = true;
+        array_item_prop_desc.value = capture_value;
         ECMA_TRY_CATCH (put_value,
-                        ecma_op_object_put (result_array_obj_p,
+                        ecma_op_object_define_own_property (result_array_obj_p,
                                             index_str_p,
-                                            capture_value,
+                                            &array_item_prop_desc,
                                             true),
```
zherczeg: What about using the 'put' but without the TRY_CATCH and simply free the return value? Something like:

    ecma_fast_free_value (ecma_op_object_put (result_array_obj_p, ...))

yichoi: RegExp.prototype.exec must return matched result array.

But if the `put` is used instead of the `define_own_property`,
its [[CanPut]] operation fails (because of the read-only property "0" of Array.prototype),
and [[Put]] operation throws exception without actually putting matched result into array.

So I think 'define_own_property' must be used in this case.

zherczeg: The standard only describes the observable behaviour, but implementations are free to do anything as long as the visible effect is the same as described in the standard. In this case if I understand correctly, you wish to discard the result of the property set. The easiest way (least binary size, fastest) to do this is calling the put, and discarding the result.

yichoi: No, I do **NOT** wish to discard the result of property set.
RegExp.prototype.exec **should not ignore** the matched result,
but it should return matched array correctly.

And I understand that the implementation can be different with the spec.
The reason that I mentioned the spec was just to clarify the explanation.

In fact, in this case, the behavior of implementation exactly follows the spec description.
If we use `ecma_op_object_put` here,
it tries looking the prototype object to check if the property exists. (same with [[CanPut]])
And if the property exists in prototype object, (and the property is non-writable),
`ecma_op_object_put` does not create new property. 
(see jerry-core/ecma/operations/ecma-objects-general.c:334)

However, RegExp.prototype.exec must return the matched result array,
regardless of the property defined on Array prototype object. 
For example, below sample code should return the array [H,e,l,l,o]. 
And if we use `ecma_op_object_define_own_property`, it prints the correct result. 
```
Array.prototype.splice(0, 1, this);
Object.freeze(Array.prototype);
print((new String("Hello")).split(new RegExp()));
```
However if we use `ecma_op_object_put`,
it cannot add properties 'h', 'e', 'l', 'l', 'o' into result array (it just throws exception), 
so the printed result is wrong. 
(In fact, the unexpected property defined in prototype object is returned instead,
so the program get killed in random place.)

zherczeg: Oh, it is even more easier then. Just use ecma_create_named_data_property / ecma_set_named_data_property_value. You can even delete the ecma_free_value (capture_value)

zherczeg: An example here:
https://github.com/Samsung/jerryscript/blob/master/jerry-core/vm/vm.c#L1015

yichoi: Oh, your new suggestion is more simpler than using define_own_property.
I will upload a new patch.

Just to be clear, as I understand, `ecma_free_value (capture_value)` can be removed because `ecma_set_named_data_property` *DOES* pass the ownership of `capture_value`, is it correct? Or are there any other reasons? I am little confused because in [jerryScript internals document](http://jerryscript.net/internals/), there's a statement like this: "When the value is passed to a function as an argument, the ownership of it will not pass, the called function have to make an own copy of the value".

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1434,15 +1434,12 @@ ecma_regexp_exec_helper (ecma_value_t regexp_value, /**< RegExp object */
...
-                                            true),
-                        ret_value);
-        ECMA_FINALIZE (put_value);
-
-        ecma_free_value (capture_value);
+        ecma_property_t *prop_p;
+        prop_p = ecma_create_named_data_property (result_array_obj_p,
+                                                  index_str_p,
+                                                  ECMA_PROPERTY_CONFIGURABLE_ENUMERABLE_WRITABLE);
+        ecma_set_named_data_property_value (prop_p, capture_value);
```
zherczeg: I would add an assert before this line that capture_value is not object. This is pretty obvious from the code above, but it does not hurt release code and adds extra safety.

zherczeg:
LGTM after the assert is added.

LaszloLango:
LGTM

PR #1278, opened 2016-08-24
Correct the documentation

bzsolt:
- Fix some style issue, typos, and examples
- Follow the variable naming conventions
- Fix tables both in the project and on the webpage

### DIFF
--- a/docs/01.GETTING-STARTED.md
+++ b/docs/01.GETTING-STARTED.md
@@ -1,4 +1,4 @@
-### Setting Up Prerequisites
+## Setting up prerequisites
 
 Currently, only Ubuntu 14.04+ is officially supported as primary development environment.
 
@@ -19,38 +19,41 @@ sudo apt-get install gcc g++ gcc-arm-none-eabi cmake cppcheck vera++ python
 
 To make our scripts run correctly, several shell utilities should be available on the system:
 
-- `find`
 - `awk`
+- `bc`
+- `find`
+- `sed`
 
-### Building Jerryscript
+## Building JerryScript
 
-##### To build debug version for Linux:
+**To build debug version for Linux**
 
 ```bash
 python tools/build.py --debug
 ```
 
-##### To build debug version for Linux without LTO (Link Time Optimization):
+**To build debug version for Linux without LTO (Link Time Optimization)**
 
 ```bash
 python tools/build.py --debug --lto=off
 ```
 
-##### Add custom arguments to CMake:
+**Add custom arguments to CMake**
 
 ```bash
 python tools/build.py --cmake-param=CMAKE_PARAM
 ```
 
-##### Set a profile mode (full|minimal):
+**Set a profile mode (full, minimal)**
 
 ```bash
 python tools/build.py --feature=full|minimal
 ```
 
-##### Use (jerry|compiler-default|external libc) libc:
+**Use (jerry, compiler-default, external) libc**
 
 The default libc is jerry-libc, but you can use compiler-default libc or an external libc:
+
 - compiler-default libc:
 
 ```bash
@@ -63,7 +66,7 @@ python tools/build.py --jerry-libc=off --compiler-default-libc=on
 python tools/build.py --jerry-libc=off --compiler-default-libc=off --compile-flag="-I/path/to/libc/include"
 ```
 
-##### Add toolchain file:
+**Add toolchain file**
 
 The ```cmake``` dir already contains some usable toolchain files, which you can use in the following format:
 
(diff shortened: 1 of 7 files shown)


docs/01.GETTING-STARTED.md
```diff
@@ -77,72 +80,65 @@ For example the cross-compile to RaspberryPi 2 is something like this:
...
-##### To get a list of all the available buildoptions for Linux:
+**To get a list of all the available buildoptions for Linux**
 
 ```bash
 python tools/build.py --help
 ```
 
-### Checking patch
+
+## Checking patch
```
zherczeg: Two newlines here?

docs/03.API-EXAMPLE.md
```diff
@@ -214,6 +214,7 @@ print_value (const jerry_value_t value)
     jerry_char_t str_buf_p[req_sz];
 
     jerry_string_to_char_buffer (value, str_buf_p, req_sz);
+    str_buf_p[req_sz] = '\0';
```
zherczeg: Isn't the jerry_char_t str_buf_p[req_sz]; declaration above is a C++11 feature?

docs/03.API-EXAMPLE.md
```diff
@@ -283,6 +284,12 @@ main (int argc, char * argv[])
       len++;
     }
 
+    /* If the command starts with 'quit', break the loop */
+    if (!strncmp (cmd, "quit", strlen ("quit")))
```
zherczeg: I think we can use strcmp simply and break only for "quit". Or is there a reason for this prefix check?

docs/04.INTERNALS.md
```diff
@@ -135,57 +135,61 @@ Byte-codes can be placed into four main categories.
 
 Byte-codes of this category serve for placing objects onto the stack. As there are many instructions representing multiple atomic tasks in CBC, there are also many instructions for pushing objects onto the stack according to the number and the type of the arguments. The following table list a few of these opcodes with a brief description.
 
-<div class="CSSTableGenerator" markdown="block">
+<span class="CSSTableGenerator" markdown="block">
 
-| byte-code                            | description                                       |
+| byte-code                            | description |
+| ------------------------------------ | ----------- |
```
zherczeg: I think the pipes should be aligned.

docs/04.INTERNALS.md
```diff
@@ -135,57 +135,61 @@ Byte-codes can be placed into four main categories.
...
 
 ### Call Byte-codes
 
 The byte-codes of this category perform calls in different ways.
 
-<div class="CSSTableGenerator" markdown="block">
+<span class="CSSTableGenerator" markdown="block">
 
 | byte-code                            | description |
+| ------------------------------------ | ----------- |
```
zherczeg: Ditto

docs/01.GETTING-STARTED.md
bzsolt: Just a left over, thanks.

docs/03.API-EXAMPLE.md
bzsolt: It's possible, but this declaration method is also used in [main-unix.c:453](https://github.com/Samsung/jerryscript/blob/master/jerry-main/main-unix.c#L453)

docs/03.API-EXAMPLE.md
bzsolt: Yes, because the `cmd` contains a `\n` at the end and the `strcmp` doesn't find them equal, so this is the reason why the `strncmp` is used instead of `strcmp`.

docs/04.INTERNALS.md
bzsolt: Sorry, but I don't fully understand, how should I align the pipes? To the bigger column in the table or to the average line width? Now they are aligned to the average line width.

LaszloLango:
LGTM

docs/03.API-EXAMPLE.md
zherczeg: Fine then.

docs/03.API-EXAMPLE.md
zherczeg: But something like quitten also quits. Perhaps we could check "quit" or "quit\n" and be happy.

bzsolt: Ok, I've updated the PR according your review.

jerry-libc/jerry-libc.c
```diff
@@ -176,7 +176,7 @@ CALL_PRAGMA (GCC diagnostic pop)
 
 /** Compare two strings. return an integer less than, equal to, or greater than zero
      if s1 is found, respectively, to be less than, to match, or be greater than s2.  */
-int
+int __attr_used___
 strcmp (const char *s1, const char *s2)
```
bzsolt: We should add this function attribute, otherwise the compiler discards `strcmp`.

zherczeg: Is this fix the -O3 issue?

zherczeg: Perhaps we should land this as a different patch, since it is not related to documentation.

bzsolt: This modification is unrelated to the -O3 issue. It's needed for this example, because compile of this code produces the following message: `'strcmp' referenced in section '.text' of <tmp object file> defined in discarded section '.text' of jerry-libc.c.o (symbol from plugin)`
Note: If the libraries and this example compiled without LTO, everything works fine.
I would like to provide a working example with default settings, so I would restore the solution with `strncmp`.
I agree with you, your example (after `> quitten = 3` the program will terminate) shows that this is not the best solution, but this is a simple example code. And force the compiler to keep this symbol at all costs is also not the best solution.
What do you think?

zherczeg: strncmp(str, "quit", strlen("quit") + 1)

All strings are terminated by \0 in C, so adding one to the strlen is ok

bzsolt: I've updated this part.

zherczeg:
LGTM

PR #1281, opened 2016-08-25
Make jerry shell more intuitive

yichoi:
Related issue: https://github.com/Samsung/jerryscript/issues/1280

- REPL mode also prints each error messages, not only each return values.
- If read_file fails, execution stops right away.

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -121,6 +121,33 @@ print_help (char *name)
                       name);
 } /* print_help */
 
+static void
+print_unhandled_exception (jerry_value_t error_value)
+{
+  assert (jerry_value_has_error_flag (error_value));
+
+  jerry_value_clear_error_flag (&error_value);
+  jerry_value_t err_str_val = jerry_value_to_string (error_value);
+  jerry_size_t err_str_size = jerry_get_string_size (err_str_val);
+  jerry_char_t err_str_buf[256];
+
+  if (err_str_size >= 256)
+  {
+    const char msg[] = "[Error message too long]";
+    err_str_size = sizeof (msg) / sizeof (char) - 1;
+    memcpy (err_str_buf, msg, err_str_size);
+  }
+  else
+  {
+    jerry_size_t sz = jerry_string_to_char_buffer (err_str_val, err_str_buf, err_str_size);
+    assert (sz == err_str_size);
+  }
+  err_str_buf[err_str_size] = 0;
+
+  jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Script Error: unhandled exception: %s\n", err_str_buf);
+  jerry_release_value (err_str_val);
+} /* print_unhandled_exception */
+
 int
 main (int argc,
       char **argv)
@@ -335,6 +362,7 @@ main (int argc,
       if (source_p == NULL)
       {
         ret_value = jerry_create_error (JERRY_ERROR_COMMON, (jerry_char_t *) "");
+        break;
       }
 
       if (is_save_snapshot_mode)
@@ -427,14 +455,21 @@ main (int argc,
         /* Evaluate the line */
         jerry_value_t ret_val_eval = jerry_eval (buffer, len, false);
 
-        /* Print return value */
-        const jerry_value_t args[] = { ret_val_eval };
-        jerry_value_t ret_val_print = jerry_call_function (print_function,
-                                                           jerry_create_undefined (),
-                                                           args,
-                                                           1);
+        if (!jerry_value_has_error_flag (ret_val_eval))
+        {
+          /* Print return value */
+          const jerry_value_t args[] = { ret_val_eval };
+          jerry_value_t ret_val_print = jerry_call_function (print_function,
+                                                             jerry_create_undefined (),
+                                                             args,
+                                                             1);
+          jerry_release_value (ret_val_print);
+        }
+        else
+        {
+          print_unhandled_exception (ret_val_eval);
+        }
 
-        jerry_release_value (ret_val_print);
         jerry_release_value (ret_val_eval);
       }
     }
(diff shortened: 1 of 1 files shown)


jerry-main/main-unix.c
```diff
@@ -121,6 +121,26 @@ print_help (char *name)
...
+{
+  assert (jerry_value_has_error_flag (error_value));
+
+  jerry_value_clear_error_flag (&error_value);
+  jerry_value_t err_str_val = jerry_value_to_string (error_value);
+
+  jerry_char_t err_str_buf[256];
+  jerry_size_t err_str_size = jerry_get_string_size (err_str_val);
+
+  assert (err_str_size < 256);
```
zherczeg: I don't think this must be true. Custom errors can have longer than 256 size.

zherczeg: Either don't print the error when the size is > 256 or dynamically allocate a buffer.

yichoi: In fact, this code has already been used to print custom errors for file-mode.
What I did was just moving the location of the code into the function.

If there're no historical reason for this, I am willing to change it.

zherczeg: Ancient code rots :( Would be good to change it.

akosthekiss: Beware that dynamic buffer allocation will most probably not be possible as we don't have malloc in jerry-main (as it must work with jerry-libc, which lacks dynamic memory support functions).

yichoi:
As @akosthekiss said, it was hard to find any way to allocate dynamic buffer in jerry-main.
So I just made main-unix.c to print unified error message when original message was longer than 256.
Please tell me if there are better ways.

LaszloLango:
LGTM

jerry-main/main-unix.c
```diff
@@ -121,6 +121,33 @@ print_help (char *name)
...
+static void
+print_unhandled_exception (jerry_value_t error_value)
+{
+  assert (jerry_value_has_error_flag (error_value));
+
+  jerry_value_clear_error_flag (&error_value);
+  jerry_value_t err_str_val = jerry_value_to_string (error_value);
+  jerry_size_t err_str_size = jerry_get_string_size (err_str_val);
+
+  jerry_char_t err_str_buf[256];
```
zherczeg: Move this definition one line above.

jerry-main/main-unix.c
```diff
@@ -121,6 +121,33 @@ print_help (char *name)
...
+  if (err_str_size >= 256)
```
zherczeg: >= 255 (you need an extra byte for the 0).
For me this is quite ok.

jerry-main/main-unix.c
```diff
@@ -121,6 +121,33 @@ print_help (char *name)
...
+print_unhandled_exception (jerry_value_t error_value)
+{
+  assert (jerry_value_has_error_flag (error_value));
+
+  jerry_value_clear_error_flag (&error_value);
+  jerry_value_t err_str_val = jerry_value_to_string (error_value);
+  jerry_size_t err_str_size = jerry_get_string_size (err_str_val);
+  jerry_char_t err_str_buf[256];
+
+  if (err_str_size >= 255)
```
akosthekiss: Is `err_str_size` the number of valid characters, i.e., not counting any terminating zeros? (Seems so!) If so, then a 255 char long string is still perfectly valid as the terminating zero will be written to the 256th byte of `err_str_buf` (which is index 255). Thus, this should either be `err_str_size >= 256` or `err_str_size > 255`.

Note: this is contradicting with a previous comment of @zherczeg

jerry-main/main-unix.c
```diff
@@ -121,6 +121,33 @@ print_help (char *name)
...
+  {
+    const char msg[] = "[Long message]";
```
akosthekiss: (thinking aloud) Truncating the too long message would be better than completely replacing it. Too bad that we cannot copy out only a part of the string with `jerry_string_to_char_buffer`.

zherczeg: This is just a simple shell, having such limitations are perfectly fine to me. Adding a truncated get could be an API extension, the challenge is breaking UTF-8 characters. But I don't think it will happen before 1.0.

yichoi:
It seemed that `err_str_size` is the number of valid characters (without null character), like @akosthekiss said. So I've changed the length check condition again.
If @zherczeg and others are fine with this, please review this PR again.

jerry-main/main-unix.c
```diff
@@ -121,6 +121,33 @@ print_help (char *name)
...
+  assert (jerry_value_has_error_flag (error_value));
+
+  jerry_value_clear_error_flag (&error_value);
+  jerry_value_t err_str_val = jerry_value_to_string (error_value);
+  jerry_size_t err_str_size = jerry_get_string_size (err_str_val);
+  jerry_char_t err_str_buf[256];
+
+  if (err_str_size >= 256)
+  {
+    const char msg[] = "[Long message]";
```
zherczeg: I would change this:
[Error message too long]

jerry-main/main-unix.c
```diff
@@ -121,6 +121,33 @@ print_help (char *name)
...
+    err_str_size = sizeof (msg) / sizeof (char);
```
akosthekiss: Well, _this_ needs some adjustment, if I'm not mistaken. Correctly, this should be `err_str_size = sizeof (msg) / sizeof (char) - 1;`, as the current code does count the terminating zero as well. We don't want to copy that terminating zero and we want to get `err_str_buf[err_str_size] = 0;` at line 145 put that 0 to the correct index. (Right now, it would write a 0 after the already existing terminating zero.)

Example:
```c
#include <stdio.h>

int main()
{
    const char msg[] = "1234567";
    int size = sizeof(msg) / sizeof(char);
    printf("size(%s) = %d\n", msg, size);
}
```
prints: "size(1234567) = 8"

zherczeg: what about strlen(msg) ?

akosthekiss: There is nothing wrong with sizeof() / sizeof(), it just have to be used carefully. It still has the advantage over strlen() of being a compile time constant and not requiring any code execution at runtime.

zherczeg:
LGTM

yichoi:
Maybe I was too worried about increasing binary size :P
Uploaded a new patch.

PR #1283, opened 2016-08-25
Numbers should be permitted as property name in object literal

yichoi:
Related issue: https://github.com/Samsung/jerryscript/issues/1282

Binary size doesn't change. (at least for linux/release build)

### DIFF
--- a/jerry-core/vm/opcodes.c
+++ b/jerry-core/vm/opcodes.c
@@ -153,7 +153,8 @@ opfunc_set_accessor (bool is_getter, /**< is getter accessor */
                      ecma_value_t accessor) /**< accessor value */
 {
   ecma_object_t *object_p = ecma_get_object_from_value (object);
-  ecma_string_t *accessor_name_p = ecma_get_string_from_value (accessor_name);
+  JERRY_ASSERT (ecma_is_value_string (accessor_name) || ecma_is_value_number (accessor_name));
+  ecma_string_t *accessor_name_p = ecma_get_string_from_value (ecma_op_to_string (accessor_name));
   ecma_property_t *property_p = ecma_find_named_property (object_p, accessor_name_p);
 
   if (property_p != NULL && ECMA_PROPERTY_GET_TYPE (property_p) != ECMA_PROPERTY_TYPE_NAMEDACCESSOR)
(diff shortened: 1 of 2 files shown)


jerry-core/vm/opcodes.c
```diff
@@ -153,7 +153,23 @@ opfunc_set_accessor (bool is_getter, /**< is getter accessor */
                      ecma_value_t accessor) /**< accessor value */
 {
   ecma_object_t *object_p = ecma_get_object_from_value (object);
-  ecma_string_t *accessor_name_p = ecma_get_string_from_value (accessor_name);
+  ecma_string_t *accessor_name_p;
+  bool need_cleanup = false;
+  if ((accessor_name & ECMA_VALUE_TYPE_MASK) == ECMA_TYPE_STRING)
```
zherczeg: Please use the helper macros from ecma-helpers.h. Also I think we can simply use the to_string in the non-string case, since that is very rare. We can use an assert since the conversion must be successful (in es5.1 at least).

jerry-core/vm/opcodes.c
```diff
@@ -153,7 +153,16 @@ opfunc_set_accessor (bool is_getter, /**< is getter accessor */
                      ecma_value_t accessor) /**< accessor value */
 {
   ecma_object_t *object_p = ecma_get_object_from_value (object);
-  ecma_string_t *accessor_name_p = ecma_get_string_from_value (accessor_name);
```
LaszloLango: I think it would be better if we'd call `ecma_op_to_string` every time and add an assert to check the type of accessor_name. Something like this:

```C++
JERRY_ASSERT (ecma_is_value_string (accessor_name) || ecma_is_value_number (accessor_name));
ecma_string_t *accessor_name_p = ecma_get_string_from_value (ecma_op_to_string (accessor_name));
```

In this case there is no need to check the type before the `ecma_deref_ecma_string (accessor_name_p)` call.

LaszloLango:
LGTM

zherczeg:
LGTM

zherczeg:
Could you rebase this to the current master? After landing, please rebase another one where you have two LGTMs.

PR #1285, opened 2016-08-25
[[DefaultValue]] should be called twice when setting object as Array length

yichoi:
Related issue: https://github.com/Samsung/jerryscript/issues/1284

It didn't hurt the sunspider performance after wrapping the code with `ecma_is_value_object` check.

### DIFF
--- a/jerry-core/ecma/operations/ecma-array-object.c
+++ b/jerry-core/ecma/operations/ecma-array-object.c
@@ -198,6 +198,13 @@ ecma_op_array_object_define_own_property (ecma_object_t *obj_p, /**< the array o
     uint32_t new_len_uint32 = ecma_number_to_uint32 (new_len_num);
 
     // d.
+    if (ecma_is_value_object (property_desc_p->value))
+    {
+      ecma_value_t compared_num_val = ecma_op_to_number (property_desc_p->value);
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -198,6 +198,12 @@ ecma_op_array_object_define_own_property (ecma_object_t *obj_p, /**< the array o
     uint32_t new_len_uint32 = ecma_number_to_uint32 (new_len_num);
 
     // d.
+    if (ecma_is_value_object (property_desc_p->value))
+    {
+      ecma_value_t compared_num_val = ecma_op_to_number (property_desc_p->value);
+      new_len_num = ecma_get_number_from_value (compared_num_val);
+      ecma_free_value (compared_num_val);
+    }
```
zherczeg: Add a newline after this "if". (the next if is not an else-if).

LaszloLango:
LGTM

zherczeg:
LGTM

zherczeg:
Please rebase this as well. And add a comment after rebase because github does not send notification otherwise.

yichoi:
@zherczeg, rebase is done.

PR #1289, opened 2016-08-26
All-in-one build friendly jerry-libm.

ktorpi:
Define constants that would collide in case of all-in-one build of jerry-libm are moved to local const variables.

### DIFF
--- a/jerry-libm/acos.c
+++ b/jerry-libm/acos.c
@@ -127,3 +127,18 @@ acos (double x)
     return 2.0 * (df + w);
   }
 } /* acos */
+
+#undef one
+#undef pi
+#undef pio2_hi
+#undef pio2_lo
+#undef pS0
+#undef pS1
+#undef pS2
+#undef pS3
+#undef pS4
+#undef pS5
+#undef qS1
+#undef qS2
+#undef qS3
+#undef qS4
--- a/jerry-libm/asin.c
+++ b/jerry-libm/asin.c
@@ -136,3 +136,19 @@ asin (double x)
     return -t;
   }
 } /* asin */
+
+#undef one
+#undef huge
+#undef pio2_hi
+#undef pio2_lo
+#undef pio4_hi
+#undef pS0
+#undef pS1
+#undef pS2
+#undef pS3
+#undef pS4
+#undef pS5
+#undef qS1
+#undef qS2
+#undef qS3
+#undef qS4
--- a/jerry-libm/scalbn.c
+++ b/jerry-libm/scalbn.c
@@ -89,3 +89,8 @@ scalbn (double x, int n)
   __HI (x) = (hx & 0x800fffff) | (k << 20);
   return x * twom54;
 } /* scalbn */
+
+#undef two54
+#undef twom54
+#undef huge
+#undef tiny
--- a/jerry-libm/trig.c
+++ b/jerry-libm/trig.c
@@ -1055,3 +1055,43 @@ tan (double x)
     return __kernel_tan (y[0], y[1], 1 - ((n & 1) << 1)); /*   1 -- n even, -1 -- n odd */
   }
 } /* tan */
+
+#undef zero
+#undef half
+#undef one
+#undef two24
+#undef twon24
+#undef invpio2
+#undef pio2_1
+#undef pio2_1t
+#undef pio2_2
+#undef pio2_2t
+#undef pio2_3
+#undef pio2_3t
+#undef S1
+#undef S2
+#undef S3
+#undef S4
+#undef S5
+#undef S6
+#undef C1
+#undef C2
+#undef C3
+#undef C4
+#undef C5
+#undef C6
+#undef T0
+#undef T1
+#undef T2
+#undef T3
+#undef T4
+#undef T5
+#undef T6
+#undef T7
+#undef T8
+#undef T9
+#undef T10
+#undef T11
+#undef T12
+#undef pio4
+#undef pio4lo
--- a/jerry-libm/atan.c
+++ b/jerry-libm/atan.c
@@ -160,3 +160,17 @@ atan (double x)
     return (hx < 0) ? -z : z;
   }
 } /* atan */
+
+#undef aT0
+#undef aT1
+#undef aT2
+#undef aT3
+#undef aT4
+#undef aT5
+#undef aT6
+#undef aT7
+#undef aT8
+#undef aT9
+#undef aT10
+#undef one
+#undef huge
--- a/jerry-libm/atan2.c
+++ b/jerry-libm/atan2.c
@@ -201,3 +201,10 @@ atan2 (double y, double x)
     }
   }
 } /* atan2 */
+
+#undef tiny
+#undef zero
+#undef pi_o_4
+#undef pi_o_2
+#undef pi
+#undef pi_lo
--- a/jerry-libm/ceil.c
+++ b/jerry-libm/ceil.c
@@ -128,3 +128,5 @@ ceil (double x)
   __LO (x) = i1;
   return x;
 } /* ceil */
+
+#undef huge
--- a/jerry-libm/exp.c
+++ b/jerry-libm/exp.c
@@ -205,3 +205,15 @@ exp (double x) /* default IEEE double exp */
     return y * twom1000;
   }
 } /* exp */
+
+#undef one
+#undef huge
+#undef twom1000
+#undef o_threshold
+#undef u_threshold
+#undef invln2
+#undef P1
+#undef P2
+#undef P3
+#undef P4
+#undef P5
--- a/jerry-libm/floor.c
+++ b/jerry-libm/floor.c
@@ -127,3 +127,5 @@ floor (double x)
   __LO (x) = i1;
   return x;
 } /* floor */
+
+#undef huge
--- a/jerry-libm/fmod.c
+++ b/jerry-libm/fmod.c
@@ -232,3 +232,5 @@ fmod (double x, double y)
   }
   return x; /* exact output */
 } /* fmod */
+
+#undef one
--- a/jerry-libm/log.c
+++ b/jerry-libm/log.c
@@ -185,3 +185,15 @@ log (double x)
     }
   }
 } /* log */
+
+#undef zero
+#undef ln2_hi
+#undef ln2_lo
+#undef two54
+#undef Lg1
+#undef Lg2
+#undef Lg3
+#undef Lg4
+#undef Lg5
+#undef Lg6
+#undef Lg7
--- a/jerry-libm/pow.c
+++ b/jerry-libm/pow.c
@@ -73,7 +73,6 @@
  * to produce the hexadecimal values shown.
  */
 
-static const double one = 1.0;
 static const double bp[] =
 {
   1.0,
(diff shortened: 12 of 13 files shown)


zherczeg:
LGTM. I think at some point we could move all constants into all functions, like "one".

akosthekiss:
What are the implications of the change on text and/or data section sizes, if any? (I know that compilers _should_ be clever enough to optimize out const variables in functions, but I've seen the opposite in practice in some cases.) When doing the assessment, please provide numbers for LTO, ALL_IN_ONE, and non-all-in-one-non-lto builds as well.

ktorpi:
Here are the numbers you asked for. There is no difference according to `size`.

|   text   |  data  |   bss  |   dec   |  hex   | |
|---------|---------|---------|---------|--------|----|
| 238469   |   16 | 2626032 | 2864517 | 2bb585 | LTO, allinone build friendly libm |
| 238469   |  16 | 2626032 | 2864517  | 2bb585 | LTO, master|
| 237219   |   16 |2626032 |2863267  |2bb0a3  |ALL_IN_ONE, allinone build friendly libm|
| 237219    |  16 |2626032 |2863267  |2bb0a3  |ALL_IN_ONE, master|
|259252    |  32 |2626032 |2885316  |2c06c4  |NON_LTO, NON_ALL_IN_ONE, allinone build friendly libm|
 259252   |   32 |2626032 |2885316  |2c06c4  |NON_LTO, NON_ALL_IN_ONE, master|

zherczeg:
Convincing for me. I am not sure we should change the coding style though.

jerry-libm/scalbn.c
```diff
@@ -33,14 +33,15 @@
  * exponentiation or a multiplication.
  */
 
-#define two54  1.80143985094819840000e+16 /* 0x43500000, 0x00000000 */
 #define twom54 5.55111512312578270212e-17 /* 0x3C900000, 0x00000000 */
```
akosthekiss: Please make sure that the `#define` -> `const double` rewrite is applied consistently, i.e., no constants remain `#define`d.

jerry-libm/acos.c
```diff
@@ -53,24 +53,24 @@
...
-#define pS5      3.47933107596021167570e-05 /* 0x3F023DE1, 0x0DFDF709 */
-#define qS1     -2.40339491173441421878e+00 /* 0xC0033A27, 0x1C8A2D4B */
-#define qS2      2.02094576023350569471e+00 /* 0x40002AE5, 0x9C598AC8 */
-#define qS3     -6.88283971605453293030e-01 /* 0xBFE6066C, 0x1B8D0159 */
-#define qS4      7.70381505559019352791e-02 /* 0x3FB3B8C5, 0xB12E9282 */
-
 double
 acos (double x)
 {
+  const double one     =  1.0;                        /* 0x3FF00000, 0x00000000 */
```
akosthekiss: Not exactly the same format as with the `#define`. It's just some trailing zeros, I know, but why the change?

jerry-libm/asin.c
```diff
@@ -58,26 +58,26 @@
...
-#define pS5      3.47933107596021167570e-05 /* 0x3F023DE1, 0x0DFDF709 */
-#define qS1     -2.40339491173441421878e+00 /* 0xC0033A27, 0x1C8A2D4B */
-#define qS2      2.02094576023350569471e+00 /* 0x40002AE5, 0x9C598AC8 */
-#define qS3     -6.88283971605453293030e-01 /* 0xBFE6066C, 0x1B8D0159 */
-#define qS4      7.70381505559019352791e-02 /* 0x3FB3B8C5, 0xB12E9282 */
-
 double
 asin (double x)
 {
+  const double one     =  1.0;                        /* 0x3FF00000, 0x00000000 */
```
akosthekiss: ditto

jerry-libm/scalbn.c
```diff
@@ -33,14 +33,15 @@
...
-#define huge   1.0e+300
-#define tiny   1.0e-300
 
 double
 scalbn (double x, int n)
 {
+  const double tiny = 1.0e-300;
+  const double huge = 1.0e+300;
+  const double two54 = 1.80143985094819840000e+16;
```
akosthekiss: Comment on binary representation got removed (should not).

jerry-libm/trig.c
```diff
@@ -190,6 +184,11 @@ static const int ipio2[] =
 static int
 __kernel_rem_pio2 (double *x, double *y, int e0, int nx, int prec)
 {
+  const double zero   = 0.0;
+  const double one    = 1.0;
```
akosthekiss: Floating point representation changed (to an equivalent one, but changed), and comment on binary representation got removed (both for zero and one).

jerry-libm/trig.c
```diff
@@ -456,26 +455,29 @@ static const int npio2_hw[] =
...
-#define pio2_2  6.07710050630396597660e-11 /* 0x3DD0B461, 0x1A600000 */
-#define pio2_2t 2.02226624879595063154e-21 /* 0x3BA3198A, 0x2E037073 */
-#define pio2_3  2.02226624871116645580e-21 /* 0x3BA3198A, 0x2E000000 */
-#define pio2_3t 8.47842766036889956997e-32 /* 0x397B839A, 0x252049C1 */
-
 static int
 __ieee754_rem_pio2 (double x, double *y)
 {
+  const double zero  = 0.0;
+  const double half  = 0.5;                        /* 0x3FE00000, 0x00000000 */
```
akosthekiss: ditto for zero, floating point representation changed for half

jerry-libm/trig.c
```diff
@@ -630,16 +632,17 @@ __ieee754_rem_pio2 (double x, double *y)
...
-#define S2    8.33333333332248946124e-03 /* 0x3F811111, 0x1110F8A6 */
-#define S3   -1.98412698298579493134e-04 /* 0xBF2A01A0, 0x19C161D5 */
-#define S4    2.75573137070700676789e-06 /* 0x3EC71DE3, 0x57B1FE7D */
-#define S5   -2.50507602534068634195e-08 /* 0xBE5AE5E6, 0x8A2B9CEB */
-#define S6    1.58969099521155010221e-10 /* 0x3DE5D93A, 0x5ACFD57C */
-
 static double
 __kernel_sin (double x, double y, int iy)
 {
+  const double half =  0.5;                        /* 0x3FE00000, 0x00000000 */
```
akosthekiss: ditto

jerry-libm/trig.c
```diff
@@ -699,16 +702,17 @@ __kernel_sin (double x, double y, int iy)
...
-#define C2  -1.38888888888741095749e-03 /* 0xBF56C16C, 0x16C15177 */
-#define C3   2.48015872894767294178e-05 /* 0x3EFA01A0, 0x19CB1590 */
-#define C4  -2.75573143513906633035e-07 /* 0xBE927E4F, 0x809C52AD */
-#define C5   2.08757232129817482790e-09 /* 0x3E21EE9E, 0xBDB4B1C4 */
-#define C6  -1.13596475577881948265e-11 /* 0xBDA8FAE9, 0xBE8838D4 */
-
 static double
 __kernel_cos (double x, double y)
 {
+  const double one =  1.0;
```
akosthekiss: ditto

jerry-libm/trig.c
```diff
@@ -776,25 +780,26 @@ __kernel_cos (double x, double y)
...
-#define T10     7.14072491382608190305e-05 /* 3F12B80F, 32F0A7E9 */
-#define T11    -1.85586374855275456654e-05 /* BEF375CB, DB605373 */
-#define T12     2.59073051863633712884e-05 /* 3EFB2A70, 74BF7AD4 */
-#define pio4    7.85398163397448278999e-01 /* 3FE921FB, 54442D18 */
-#define pio4lo  3.06161699786838301793e-17 /* 3C81A626, 33145C07 */
-
 static double
 __kernel_tan (double x, double y, int iy)
 {
+  const double one    =  1.0;
```
akosthekiss: ditto

akosthekiss:
@ktorpi, what architecture/OS were these results measured on? If x86_64, would it be possible to do the same measurements on ARM as well?

jerry-libm/acos.c
ktorpi: The reason is that the same literal value was presented in different forms in libm (e.g. 1.0, 1.00000000), so I tried to unify the form and I've chosen (at least in my opinion) the more readable one. Of course, if it's more suitable to use original form, I'll modify the patch.

akosthekiss: Almost all constants are given in the `0.12345678901234567890e+12 /* 0x12345678, 0x9abcdef0 */` format. If changing representation anywhere, I'd suggest moving in the unified direction.

ktorpi:
Results on pi:

|   text   |  data  |   bss  |   dec   |  hex   | |
|---------|---------|---------|---------|--------|----|
| 156672	  |   16 | 	2623740 | 	2780428	 | 2a6d0c | LTO, allinone build friendly libm |
|  156622	   |  16 | 2623740 | 	2780378	 | 2a6cda | LTO, master|
| 155354	   |  16	 | 2623736 | 	2779106	 | 2a67e2  | ALL_IN_ONE, allinone build friendly libm |
| 155296	   |  16	  | 2623736	 | 2779048	 | 2a67a8	 | ALL_IN_ONE, master|
| 162303	   |  24 | 	2623744 | 	2786071 |	 2a8317 | NON_LTO, NON_ALL_IN_ONE, allinone build friendly libm|
| 162243	   |  24	 | 2623744 | 	2786011	 | 2a82db  | NON_LTO, NON_ALL_IN_ONE, master|

akosthekiss:
Thanks for the results. They align with previous experiences, unfortunately. The 50-60 bytes increase (not a big one, though) signals that gcc for arm puts the const double variables in the constant pool, and does not treat them exactly as double literals. This might also have some impact on performace, but as our benchmarks are not math-heavy as I recall, I don't expect to measure big difference.

I see some options:
- Accept this approach. (Issue with that analyzed above.)
- Factor out all defines to the internal header. (That would separate definitions from their uses, which makes it hard to maintain.)
- A new idea: add `#undef`s to the end of source files for all locally `#define`d macros. (This would keep the numbers as literals and still enable the all-in-one concatenation of the sources.)

Opinions?

LaszloLango:
I'd prefer the 1) or the 2) option.

zherczeg:
Don't hate me but I prefer option 3.

akosthekiss:
I'd also second the 3rd option -- I'm sorry that I suggested it so late

LaszloLango:
2 against 1, so let's use `#undef` :)

zherczeg:
LGTM

jerry-libm/asin.c
```diff
@@ -136,3 +136,33 @@ asin (double x)
...
+#undef pS1
+#undef pS2
+#undef pS3
+#undef pS4
+#undef pS5
+#undef qS1
+#undef qS2
+#undef qS3
+#undef qS4
+#undef huge
```
akosthekiss: copy-paste duplication of undefs

akosthekiss:
if asin.c is fixed, LGTM

akosthekiss:
please, rebase to latest master

ktorpi:
rebased

PR #1293, opened 2016-08-29
FunctionExpression name binding should be immutable

yichoi:
Related issue: https://github.com/Samsung/jerryscript/issues/1292
Please tell me if there's better way to find whether `CBC_INITIALIZE_VAR` is initializing function expression name binding or not.

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -530,6 +530,8 @@ vm_init_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
   uint16_t register_end;
   jmem_cpointer_t *literal_start_p = frame_ctx_p->literal_start_p;
   bool is_strict = ((frame_ctx_p->bytecode_header_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE) != 0);
+  jmem_cpointer_t self_reference;
+  ECMA_SET_NON_NULL_POINTER (self_reference, bytecode_header_p);
 
   /* Prepare. */
   if (!(bytecode_header_p->status_flags & CBC_CODE_FLAGS_FULL_LITERAL_ENCODING))
@@ -602,11 +604,7 @@ vm_init_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
           ecma_string_t *name_p = JMEM_CP_GET_NON_NULL_POINTER (ecma_string_t,
                                                                 literal_start_p[literal_index]);
 
-          vm_var_decl (frame_ctx_p, name_p);
-
           READ_LITERAL_INDEX (value_index);
-
-          ecma_object_t *ref_base_lex_env_p = ecma_op_resolve_reference_base (frame_ctx_p->lex_env_p, name_p);
 
           if (value_index < register_end)
           {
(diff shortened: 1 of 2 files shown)


jerry-core/vm/vm.c
```diff
@@ -620,11 +619,26 @@ vm_init_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
                                                      literal_start_p[value_index]);
           }
 
-          /* TODO: check the return value */
-          ecma_op_put_value_lex_env_base (ref_base_lex_env_p,
-                                          name_p,
-                                          is_strict,
-                                          lit_value);
+          if ((ecma_get_lex_env_type (frame_ctx_p->lex_env_p) == ECMA_LEXICAL_ENVIRONMENT_DECLARATIVE)
```
zherczeg: I don't really understand this check, could you explain to me?

zherczeg:
Personally I would just capture the self-reference, since self-reference is only possible with function expressions. So just decode the literal with ECMA_GET_NON_NULL_POINTER, and if it is the same as the byte code you get a match. Perhaps we could extract this decoding out from the actual function call.

jerry-core/vm/vm.c
yichoi: I borrowed the check code from here:
https://github.com/Samsung/jerryscript/blob/master/jerry-core/parser/js/js-parser.c#L871

zherczeg: I was talking about the ecma_get_lex_env_type part, not the other condition.

yichoi: That check is added because `value_index == const_literal_end` check wasn't enough for global environment code. The `value_index` of first function declaration in global environment was same as `const_literal_end`, so I wanted to exclude the global environment.
Anyway since it seems to be hard to understand the code, I am trying to modify this check code.

zherczeg:
Let me help a bit. In the beginning of vm_init_loop() create a compressed pointer from byte_code_p. If literal_start_p[literal_index] is the same as this compressed pointer, you got a self reference, and create an immutable binding from it.

yichoi:
Thanks for your help, @zherczeg. If I understand correctly, I think you mean `bytecode_header_p` instead of `bytecode_p`, and `literal_start_p[value_index]` instead of `literal_start_p[literal_index]`. I uploaded a new commit.

jerry-core/vm/vm.c
```diff
@@ -618,11 +614,28 @@ vm_init_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
                                                      literal_start_p[value_index]);
           }
 
-          /* TODO: check the return value */
-          ecma_op_put_value_lex_env_base (ref_base_lex_env_p,
-                                          name_p,
-                                          is_strict,
-                                          lit_value);
+          jmem_cpointer_t self_reference;
+          ECMA_SET_NON_NULL_POINTER (self_reference, bytecode_header_p);
```
zherczeg: I would move this out from the loop at the beginning of the function.

zherczeg:
Yes I think something like this.

It is worth noting that the original code seems inefficient though, since we create a binding, then resolve it and set its value. This should be reworked in a follow-up patch to allow create and initialize a binding (add an extra default value parameter to ecma_op_create_immutable_binding ecma_op_create_mutable_binding).

yichoi:
Moved the initialization of `self_reference` into the beginning of the function, and rebased the commit to the latest master.

zherczeg:
LGTM

LaszloLango:
LGTM

PR #1294, opened 2016-08-29
Use C99's variable length array instead of alloca()

bsdelf:
"alloc.h" is a non-standard header file, and won't compile on FreeBSD (at least). To make this project more portable, the standard "stdlib.h" should be preferred.

However, since "-std=c99" is enabled globally in "CMakeLists.txt", use C99's "variable length array" is much easier.

Reference:
1. https://en.wikibooks.org/wiki/C_Programming/Platform_Reference/alloc.h
2. https://en.wikipedia.org/wiki/Variable-length_array
3. http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf
4. https://gcc.gnu.org/onlinedocs/gcc/Variable-Length.html

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -36,8 +36,6 @@
 #include "vm.h"
 #include "vm-stack.h"
 
-#include <alloca.h>
-
 /** \addtogroup vm Virtual machine
  * @{
  *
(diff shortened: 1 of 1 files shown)


zherczeg:
Is the generated code the same as with alloca?

bsdelf:
I checked the sha256 digest of the ELF binaries, with or without this patch, "bin/jerry" are same. 

As for "vm.c.o", the disassembly of the function "vm_run_with_alloca" are same, but looks like the debug sections are different, thus the digest changed.

bsdelf:
kkk

zherczeg:
LGTM
Please squash the commits and rebase it to master.

LaszloLango:
LGTM

zherczeg:
Thank you, landed.

PR #1301, opened 2016-08-30
Fix bug in for-in bytecode replacement

yichoi:
Related issue: https://github.com/Samsung/jerryscript/issues/1300

`CBC_PUSH_PROP_LITERAL_REFERENCE` leaves 2 more values on stack, 
and it doesn't assign anything into lhs.
But in this case, the assign operation is necessary, and 2 elements should be poped from the stack.
so I think `CBC_ASSIGN_PROP_LITERAL` should be used in here.


zherczeg:
Nice catch. Probably a left-over from some copy-paste.
LGTM

LaszloLango:
LGTM

yichoi:
Rebased the code.

PR #1302, opened 2016-08-30
Change return value to ecma_value_t for getting internal properties

zherczeg:
Start reworking the property handling.

This patch has a negligible perf change (+0.18%), but reduces binary size by 600 bytes: 157304 -> 156744.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects-arguments.c
+++ b/jerry-core/ecma/operations/ecma-objects-arguments.c
@@ -80,10 +80,6 @@ ecma_op_create_arguments_object (ecma_object_t *func_obj_p, /**< callee function
 
   // 1.
 
-  // 4.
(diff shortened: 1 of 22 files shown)


jerry-core/ecma/operations/ecma-objects-arguments.c
```diff
@@ -99,84 +95,80 @@ ecma_op_create_arguments_object (ecma_object_t *func_obj_p, /**< callee function
 
   ecma_property_descriptor_t prop_desc = ecma_make_empty_property_descriptor ();
 
-  if (bytecode_data_p != NULL)
```
zherczeg: This check is removed because it is unnecessary. The result is a lot of code realignment, so the changes below are mostly caused by that.

LaszloLango:
LGTM

dbatyai:
LGTM

PR #1304, opened 2016-08-30
Improve the Curie build file generator a bit

galpeter:
Works with python 2 and 3.

### DIFF
--- a/targets/curie_bsp/setup.py
+++ b/targets/curie_bsp/setup.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python
 
 # Copyright 2016 Intel Corporation
+# Copyright 2016 Samsung Electronics Co., Ltd.
+# Copyright 2016 University of Szeged
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


zherczeg:
@jiangzidong could you try this patch?

targets/curie_bsp/setup.py
```diff
@@ -14,138 +16,147 @@
...
-    out_put = '''
+def build_soft_links(project_path, jerry_path):
+    """ Creates soft links into the @project_path. """
+
+    if not os.path.exists(project_path):
+        os.makedirs(project_path)
+
+    links = [
+        { # arc
+            'src': 'targets/curie_bsp/jerry_app/arc',
```
akosthekiss: Why no `os.path.join()` for 'src'?

targets/curie_bsp/setup.py
```diff
@@ -14,138 +16,147 @@
...
+        src = os.path.join(jerry_path, link['src'])
+        link_name = os.path.join(project_path, link['link_name'])
+        if not os.path.islink(link_name):
+            os.symlink(src, link_name)
+            print("Created symlink '{link_name}' -> '{src}'".format(src=src, link_name=link_name))
+
+
+def find_sources(root_dir, sub_dir):
+    """
+    Find .c and .S file inside the @root_dir/@sub_dir directory.
```
akosthekiss: (typo) file -> files

targets/curie_bsp/setup.py
galpeter: Good question! IMHO it's readable, and now I noticed that below there is a join for 'quark' and 'include' (also there is not variable part in the path here and if for whatever reason you want to run this on windows the `/` paths will work)

targets/curie_bsp/setup.py
```diff
@@ -159,62 +170,83 @@ def create_makefile_in_curie(curie_path):
...
-        print 'Usage:'
-        print sys.argv[0] + ' [full or relative path of Curie_BSP]'
-    else:
-        file_dir = os.path.dirname(os.path.abspath(__file__))
-        jerry_path = os.path.join(file_dir, "..", "..")
-        curie_path = os.path.join(os.getcwd(), sys.argv[1])
-        main(curie_path, jerry_path)
+    import sys
+
+    if len(sys.argv) < 2:
```
akosthekiss: Why the change from `!=` to `<`? What would be the meaning of accepted optional command line arguments in `sys.argv[2:]`?

galpeter: Just wanted to be a bit more fault tolerant (accidentally putting two paths there :)) but you are right it would be better to leave it as `!=`

galpeter:
@akosthekiss updated based on your comments.

jiangzidong:
@zherczeg sure, I will try the patch on device next morning :)

LaszloLango:
@galpeter, please fix the Sign-off in the commit message

galpeter:
ohh... yeah forgot that thing.

targets/curie_bsp/setup.py
```diff
@@ -159,62 +170,83 @@ def create_makefile_in_curie(curie_path):
...
+    quark_include_paths = [
+        'include',
+        'jerryscript',
+        'jerryscript/jerry-libm/include',
+        'jerryscript/targets/curie_bsp/include'
+    ] + list(jerry_data['dirs'])
+
+    quark_includes = [
+        '-Wno-error',
+    ] + ['-I$(PROJECT_PATH)/quark/%s' % dir for dir in quark_include_paths]
```
akosthekiss: Do we need `$(PROJECT_PATH)` here or do we have that already from the `project_path` variable?
And why not `['-I%s' % os.path.join(project_path, 'quark', dir) for dir in quark_include_paths]`?

targets/curie_bsp/setup.py
```diff
@@ -159,62 +170,83 @@ def create_makefile_in_curie(curie_path):
...
+    jerry_data = build_jerry_data(jerry_path)
+    jerry_objects = build_obj_y(jerry_data['sources'])
+    jerry_defines = jerry_data['cflags']
+    jerry_build_dirs = build_mkdir(jerry_data['dirs'])
+
+    quark_include_paths = [
+        'include',
+        'jerryscript',
+        'jerryscript/jerry-libm/include',
+        'jerryscript/targets/curie_bsp/include'
```
akosthekiss: If we use `os.path.join` everywhere, might be nice to use it here as well.

targets/curie_bsp/setup.py
galpeter: I wanted to keep the generated code as close as possible to the previous version. Also I don't know if there would be any side effect if we change it to the `project_path` one.

akosthekiss: might be good to hear @jiangzidong 's opinion

jiangzidong: no side effect here. it is ok to replace $(PROJECT_PATH)

jiangzidong:
@galpeter the patch works well on real device (and I also tested the` ['-I%s' % os.path.join(project_path, 'quark', dir) for dir in quark_include_paths]`)
LGTM, thanks!

galpeter:
@jiangzidong that's good to hear. I'll do a final round of update.

galpeter:
@akosthekiss everything should be fixed now. PTAL.

akosthekiss:
LGTM

LaszloLango:
LGTM

akosthekiss:
Please, update to latest master to get landed

galpeter:
rebased to the latest master.

PR #1307, opened 2016-08-31
Fix warnings reported on OS X

akosthekiss:
OS X build regularly reports some 39 warnings falling in 3 major
categories:

* "static function '...' is used in an inline function with
  external linkage [-Wstatic-in-inline]": Some semantics around
  `inline` have changed between C89 and C99, and gcc and clang seem
  to disagree on how strict they should be about them. Solution
  chosen is to use `-Wnostatic-in-inline` command line option for
  clang.

* "implicit conversion turns floating-point number into integer:
  'double' to 'bool' [-Wfloat-conversion]": `if (fmod (..., ...))`
  was used at different places, which is not nice anyway, thus the
  return value is compared explicitly against `ECMA_NUMBER_ZERO`.

* "format string is not a string literal [-Wformat-nonliteral]":
  Console and log port I/O functions have a printf-like interface,
  and the default implementations actually pass both format string
  and the remaining arguments to a vfprintf. However, clang is
  strict about the format string parameter of vfprintf and expects
  a literal there. By annotating the port I/O functions with
  `__attribute__ ((format (printf, ..., ...)))`, clang will check
  the format string being a literal string earlier, when the port
  functions are called, and will not complain within them when
  vfprintf is called.
  (Actually, this has revealed an incorrect format string, which
  has been fixed as well.)

(There were also some single conversion errors not listed above.)

The patch was tested on OS X (where all warnings disappeared), but
it should help clang compilation on other OS's as well.

### DIFF
--- a/jerry-core/ecma/operations/ecma-regexp-object.c
+++ b/jerry-core/ecma/operations/ecma-regexp-object.c
@@ -734,7 +734,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
 
         while (*bc_p == RE_OP_ALTERNATIVE)
         {
-          JERRY_TRACE_MSG (", jump: %d");
+          JERRY_TRACE_MSG (", jump: %d", offset);
           bc_p++;
           offset = re_get_value (&bc_p);
           bc_p += offset;
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -183,7 +183,7 @@ if(CMAKE_COMPILER_IS_GNUCC)
   endif()
   jerry_add_compile_warnings(logical-op)
 else()
-  jerry_add_compile_flags(-Wno-nested-anon-types)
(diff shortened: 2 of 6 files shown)


tilmannOSG:
@akosthekiss Nice! I've seen those warnings as well when building with Clang on Linux, thanks for fixing them!

LGTM

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -734,7 +734,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
 
         while (*bc_p == RE_OP_ALTERNATIVE)
         {
-          JERRY_TRACE_MSG (", jump: %d");
+          JERRY_TRACE_MSG (", jump: %d", offset);
```
LaszloLango: good catch :)

LaszloLango:
LGTM

zherczeg:
LGTM

PR #1313, opened 2016-09-01
Add documentation to port API

LaszloLango:

### DIFF
--- a/docs/05.PORT-API.md
+++ b/docs/05.PORT-API.md
@@ -1,0 +1,210 @@
+# Reference
+
+## Termination
+
+It is questionable whether a library should be able to terminate an application. Any API function can signal an error (ex.: cannot allocate memory), so the engine use the termination approach with this port function.
+
+```c
+/**
+ * Signal the port that jerry experienced a fatal failure from which it cannot
+ * recover.
+ *
+ * @param code gives the cause of the error.
+ *
+ * Note: jerry expects the function not to return.
+ *
+ * Example: a libc-based port may implement this with exit() or abort(), or both.
+ */
+void jerry_port_fatal (jerry_fatal_code_t code);
+```
+
+Error codes
+
+```c
+typedef enum
+{
+  ERR_OUT_OF_MEMORY = 10,
+  ERR_SYSCALL = 11,
+  ERR_REF_COUNT_LIMIT = 12,
+  ERR_FAILED_INTERNAL_ASSERTION = 120
+} jerry_fatal_code_t;
+```
+
+## I/O
+
+These are the only I/O functions jerry calls.
+
+```c
+/**
+ * Print a string to the console. The function should implement a printf-like
+ * interface, where the first argument specifies a format string on how to
+ * stringify the rest of the parameter list.
+ *
+ * This function is only called with strings coming from the executed ECMAScript
+ * wanting to print something as the result of its normal operation.
+ *
+ * It should be the port that decides what a "console" is.
+ *
+ * Example: a libc-based port may implement this with vprintf().
+ */
+void jerry_port_console (const char *fmt, ...);
+
+/**
+ * Jerry log levels. The levels are in severity order
+ * where the most serious levels come first.
+ */
+typedef enum
+{
+  JERRY_LOG_LEVEL_ERROR,    /**< the engine will terminate after the message is printed */
+  JERRY_LOG_LEVEL_WARNING,  /**< a request is aborted, but the engine continues its operation */
+  JERRY_LOG_LEVEL_DEBUG,    /**< debug messages from the engine, low volume */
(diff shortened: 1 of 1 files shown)


docs/05.PORT-API.md
```diff
@@ -0,0 +1,287 @@
...
+  ERR_OUT_OF_MEMORY = 10,
+  ERR_SYSCALL = 11,
+  ERR_REF_COUNT_LIMIT = 12,
+  ERR_FAILED_INTERNAL_ASSERTION = 120
+} jerry_fatal_code_t;
+```
+
+## I/O
+
+These should be the only I/O functions jerry calls. All calls to `printf`, `vfprintf`, and `putchar` should be rewritten to one of these, depending on what jerry wants to print.
```
zherczeg: These are the only functions jerry calls.
The rest can be deleted.

zherczeg:
LGTM

docs/05.PORT-API.md
```diff
@@ -0,0 +1,287 @@
+# Reference
+
+## Termination
+
+It is questionable whether a library should be able to terminate an application. However, as of now, we only have the concept of completion code around `jerry_parse` and `jerry_run`. Most of the other API functions have no way of signaling an error. So, we keep the termination approach with this port function.
```
akosthekiss: Do we still have the concept of completion code around jerry_parse and jerry_run?

docs/05.PORT-API.md
```diff
@@ -0,0 +1,287 @@
...
+double jerry_port_get_current_time (void);
+```
+
+# How to port JerryScript
+
+This section describes the default port implementation which was created for Unix based systems.
+
+## jerry-port-default.h
+
+This header containt the decleration of port specific helper functions for termination and I/O.
```
akosthekiss: contains, declaration

docs/05.PORT-API.md
```diff
@@ -0,0 +1,287 @@
...
+
+```C
```
akosthekiss: Use consistent annotation of example language throughout the document. Either capital-C or small-c.

docs/05.PORT-API.md
```diff
@@ -0,0 +1,287 @@
...
+ * Get system time
+ *
+ * @return milliseconds since Unix epoch
+ */
+double jerry_port_get_current_time (void);
+```
+
+# How to port JerryScript
+
+This section describes the default port implementation which was created for Unix based systems.
```
akosthekiss: I'm not sure that we should copy the default implementation verbatim here. From previous experiences, it seems that port creators often get confused by the default port-specific public helper methods (e.g., whether to abort or exit on failure). So, I think we should provide a simplified version here.

docs/05.PORT-API.md
LaszloLango: Good question. AFAIK, "out of memory" errors can happen in `jerry_create_...` calls. @zherczeg ?

docs/05.PORT-API.md
LaszloLango: I don't mind to simplify the documentation.

docs/05.PORT-API.md
zherczeg: Yes, anything that allocates memory (property setting, etc.).

akosthekiss: I mean, completion code is an old concept, which has been removed, AFAIK. Those functions return jerry_value_t nowadays. So that seems definitely like an outdated sentence.

docs/05.PORT-API.md
akosthekiss: I'd vote for providing a simpler example then. We just have to make sure that the example code works, of course...

LaszloLango:
@akosthekiss, @zherczeg: I've updated the PR. Please check.

docs/05.PORT-API.md
```diff
@@ -0,0 +1,210 @@
+# Reference
+
+## Termination
+
+It is questionable whether a library should be able to terminate an application. Some of the API functions may signal an error (ex.: cannot allocate memory). So, we keep the termination approach with this port function.
```
akosthekiss: Only some of the API functions can signal an error ... for the others, we keep ...

akosthekiss:
If my last comment gets fixed, LGTM

docs/05.PORT-API.md
```diff
@@ -0,0 +1,210 @@
+# Reference
+
+## Termination
+
+It is questionable whether a library should be able to terminate an application. Any API function can signal an error (ex.: cannot allocate memory), so the engine use the termination approach with this port function.
```
zherczeg: I would remove the "It is questionable whether a library should be able to terminate an application. " because the next sentence has no logical relation to it.

PR #1317, opened 2016-09-02
Add missing jerry-libm include path to jerry-core target.

robertsipka:
Add the 'jerry-libm/include' path to specified include directories to jerry-core target in case of enabled jerry-libm build.


akosthekiss:
LGTM

LaszloLango:
LGTM

PR #1320, opened 2016-09-05
Improve error messages

LaszloLango:
* Print location on parser errors.
 * Do not print messages on parse error if FEATURE_ERROR_MESSAGES is not set.
 * Minor style fixes

### DIFF
--- a/jerry-core/parser/js/js-parser.c
+++ b/jerry-core/parser/js/js-parser.c
@@ -1803,7 +1803,7 @@ static ecma_compiled_code_t *
 parser_parse_source (const uint8_t *source_p, /**< valid UTF-8 source code */
                      size_t size, /**< size of the source code */
                      int strict_mode, /**< strict mode */
-                     parser_error_location *error_location) /**< error location */
+                     parser_error_location_t *error_location_p) /**< error location */
 {
   parser_context_t context;
   ecma_compiled_code_t *compiled_code;
@@ -1811,9 +1811,9 @@ parser_parse_source (const uint8_t *source_p, /**< valid UTF-8 source code */
   context.error = PARSER_ERR_NO_ERROR;
   context.allocated_buffer_p = NULL;
 
-  if (error_location != NULL)
+  if (error_location_p != NULL)
   {
-    error_location->error = PARSER_ERR_NO_ERROR;
+    error_location_p->error = PARSER_ERR_NO_ERROR;
   }
 
   context.status_flags = PARSER_NO_REG_STORE | PARSER_LEXICAL_ENV_NEEDED | PARSER_ARGUMENTS_NOT_NEEDED;
@@ -1907,11 +1907,11 @@ parser_parse_source (const uint8_t *source_p, /**< valid UTF-8 source code */
                          context.allocated_buffer_size);
     }
 
-    if (error_location != NULL)
+    if (error_location_p != NULL)
     {
-      error_location->error = context.error;
-      error_location->line = context.token.line;
-      error_location->column = context.token.column;
+      error_location_p->error = context.error;
+      error_location_p->line = context.token.line;
+      error_location_p->column = context.token.column;
     }
 
     compiled_code = NULL;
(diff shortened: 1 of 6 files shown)


jerry-core/parser/js/js-parser.c
```diff
@@ -2236,19 +2236,56 @@ parser_parse_script (const uint8_t *source_p, /**< source code */
...
+                                                    (lit_utf8_byte_t *) line_str_p,
+                                                    ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32);
+
+    char col_str_p[ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32];
+    uint32_t col_len = ecma_uint32_to_utf8_string (parser_error.column,
+                                                   (lit_utf8_byte_t *) col_str_p,
+                                                   ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32);
+
+    uint32_t msg_size = err_str_size + line_len + col_len + 20;
+    char error_msg_p[msg_size];
```
zherczeg: I think we should allocate memory with DEFINE_LOCAL_ARRAY

zherczeg:
LGTM

tilmannOSG:
@LaszloLango Do I understand correctly that this enables better error messages for parser errors?  
How about adding a couple regression tests for that? :)

LaszloLango:
@tilmannOSG, yes, it makes error messages better. I'll try to write some test for it.

LaszloLango:
@tilmannOSG, I've added a unittest for this. Please check.

tilmannOSG:
@LaszloLango Thanks!  
I'll leave the technical review to someone else who's more familiar with this part of the codebase :)

jerry-core/parser/js/js-parser.c
```diff
@@ -2236,19 +2236,62 @@ parser_parse_script (const uint8_t *source_p, /**< source code */
...
+    ecma_value_t error_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY);
+
+    JMEM_DEFINE_LOCAL_ARRAY (error_msg_p, msg_size, char);
+    char *err_msg_pos_p = error_msg_p;
+
+    strncpy (err_msg_pos_p, err_str_p, err_str_size);
+    err_msg_pos_p += err_str_size;
+
+    strncpy (err_msg_pos_p, " [line: ", 8);
+    err_msg_pos_p += 8;
```
akosthekiss: There are a bit too much hard-coded string and length constants around here that have to be kept in sync manually and easy to get wrong. (IMHO) Something like:
```c
#define POS_START  " [line: "
#define POS_MIDDLE ", column: "
#define POS_END    "]"
uint32_t msg_size = err_str_size + line_len + col_len + (uint32_t) sizeof (POS_START) + (uint32_t) sizeof (POS_MIDDLE) + (uint32_t) sizeof (POS_END);
strncpy (err_msg_pos_p, POS_START, sizeof (POS_START));
err_msg_pos_p += sizeof (POS_START);
// etc...
```
might be easier to maintain.

jerry-main/main-unix.c
```diff
@@ -144,7 +144,7 @@ print_unhandled_exception (jerry_value_t error_value)
   }
   err_str_buf[err_str_size] = 0;
 
-  jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Script Error: unhandled exception: %s\n", err_str_buf);
+  jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Unhandled Exception: %s\n", err_str_buf);
```
akosthekiss: I'm not fully convinced about the necessity of this change. In jerry-main, all our diagnostics are prefixed: if something is not completely OK, we print something with "Warning: " prefix, if something went totally wrong in jerry-main from which we cannot recover, we use "Error: " prefix, and if some show-stopper issue comes from the script, we prefix it with "Script Error: ". This way it is easier to distinguish whether source of the error is in the tool or in the executed script.
(Note: the same "Script Error: " prefix is used in the JS assertion handler.)

jerry-core/parser/js/js-parser.c
LaszloLango: sizeof (POS_START) return with length+1, so I have to put -1 everywhere. Are you sure about this?

jerry-main/main-unix.c
LaszloLango: My problem with this is the too many colons in the output.

`Script Error: unhandled expection: SyntaxError: ...`

IMHO, "Unhandled Exception" is obviously refers to the source of the error. What about 'Unhandled Script Error: SyntaxError: ...' or only 'Script Error: SyntaxError: ...'?

LaszloLango:
@akosthekiss, I've updated the PR. Please check.

jerry-main/main-unix.c
```diff
@@ -144,7 +144,7 @@ print_unhandled_exception (jerry_value_t error_value)
   }
   err_str_buf[err_str_size] = 0;
 
-  jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Script Error: unhandled exception: %s\n", err_str_buf);
+  jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Unhandled Script Error: %s\n", err_str_buf);
```
akosthekiss: From the list of alternatives you've suggested, I'd go for "Script Error: ", so even "Unhandled" might be removed.

akosthekiss:
LGTM if jerry-main message is fixed and once Travis CI gets green again

PR #1322, opened 2016-09-05
Add tolerant mode to Signed-off-by check

akosthekiss:
In tolerant mode, only the existence of the Signed-off-by line is
checked but the name and address is not matched against the author
of the commit.

The tools/run-check.py script is extended to allow calling the
check script in tolerant mode and also to detect whether it is
being run by Travis CI.

The Travis CI config has been modified to check PRs in strict mode
but use tolerant mode for merges to master. This should enable the
use of the Merge button on the GitHub web interface. (The PR is
strictly checked when a contributor opens it but when a committer
merges it via the web interface and GitHub rewrites author details
from the contributor's profile, master will not turn red either.)

### DIFF
--- a/tools/check-signed-off.sh
+++ b/tools/check-signed-off.sh
@@ -15,6 +15,49 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Usage
+function print_usage
+{
+ echo "Usage: $0 [--help] [--tolerant]"
+}
+
+function print_help
+{
+ echo "$0: Check Signed-off-by message of the latest commit"
+ echo ""
+ print_usage
+ echo ""
+ echo "Optional arguments:"
+ echo "  --help            print this help message"
+ echo "  --tolerant        check the existence of the message only but don't"
+ echo "                    require the name and email address to match the author"
+ echo "                    of the commit"
+ echo ""
+ echo "The last line of every commit message must follow the form of:"
+ echo "'JerryScript-DCO-1.0-Signed-off-by: NAME EMAIL', where NAME and EMAIL must"
+ echo "match the name and email address of the author of the commit (unless in"
+ echo "tolerant mode)."
+}
+
+# Processing command line
+TOLERANT="no"
+while [ "$#" -gt 0 ]
+do
+ if [ "$1" == "--help" ]
+ then
+  print_help
+  exit 0
+ elif [ "$1" == "--tolerant" ]
+ then
+  TOLERANT="yes"
+  shift
+ else
+  print_usage
+  exit 1
+ fi
+done
+
+# Determining latest commit
 parent_hashes=(`git show -s --format=%p HEAD | head -1`)
 
 if [ "${#parent_hashes[@]}" -eq 1 ]
@@ -28,16 +71,27 @@ else
  exit 1
 fi
 
-author_name=`git show -s --format=%an $commit_hash`
-author_email=`git show -s --format=%ae $commit_hash`
-required_signed_off_by_line="JerryScript-DCO-1.0-Signed-off-by: $author_name $author_email"
+# Checking the last line
 actual_signed_off_by_line=`git show -s --format=%B $commit_hash | sed '/^$/d' | tr -d '\015' | tail -n 1`
 
-if [ "$actual_signed_off_by_line" != "$required_signed_off_by_line" ]
+if [ "$TOLERANT" == "no" ]
 then
- echo -e "\e[1;33mSigned-off-by message is incorrect. The following line should be at the end of the $commit_hash commit's message: '$required_signed_off_by_line'. \e[0m"
+ author_name=`git show -s --format=%an $commit_hash`
+ author_email=`git show -s --format=%ae $commit_hash`
+ required_signed_off_by_line="JerryScript-DCO-1.0-Signed-off-by: $author_name $author_email"
 
- exit 1
+ if [ "$actual_signed_off_by_line" != "$required_signed_off_by_line" ]
+ then
+  echo -e "\e[1;33mSigned-off-by message is incorrect. The following line should be at the end of the $commit_hash commit's message: '$required_signed_off_by_line'. \e[0m"
+  exit 1
+ fi
+else
+  echo -e "\e[1;33mWarning! The name and email address of the author of the $commit_hash commit is not checked in tolerant mode! \e[0m"
+  if echo "$actual_signed_off_by_line" | grep -q -v '^JerryScript-DCO-1.0-Signed-off-by:'
+  then
+   echo -e "\e[1;33mSigned-off-by message is incorrect. The following line should be at the end of the $commit_hash commit's message: '$required_signed_off_by_line'. \e[0m"
+   exit 1
+  fi
 fi
 
 exit 0
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ install:
 script: "python tools/run-tests.py $OPTS"
 
 env:
-  - OPTS="--check-signed-off --check-cppcheck --check-vera"
+  - OPTS="--check-signed-off-travis --check-cppcheck --check-vera"
   - OPTS="--jerry-tests --jerry-test-suite"
   - OPTS="--jerry-tests --jerry-test-suite --toolchain=cmake/toolchain_linux_armv7l.cmake" TIMEOUT=300
   - OPTS=--buildoption-test
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 
 import argparse
+import os
 import subprocess
 import sys
 from settings import *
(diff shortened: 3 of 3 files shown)


tilmannOSG:
@akosthekiss Are you sure that we actually need this? E.g. when I merged https://github.com/Samsung/jerryscript/pull/1307 using the web interface no rewrite of the author information happened. I assume if someone has an email address associated with their GitHub account that is different from the one that is being used for the commit in the pull request the rewrite happens?

LaszloLango:
IMHO, we need this change.

tilmannOSG:
@LaszloLango Generally I'm not opposed to the change but I would like to get a better understanding of why it is actually needed :)  
Do we have any examples of merges where the Author field was rewritten?

tools/check-signed-off.sh
```diff
@@ -28,16 +69,27 @@ else
...
 
- exit 1
+ if [ "$actual_signed_off_by_line" != "$required_signed_off_by_line" ]
+ then
+  echo -e "\e[1;33mSigned-off-by message is incorrect. The following line should be at the end of the $commit_hash commit's message: '$required_signed_off_by_line'. \e[0m"
+  exit 1
+ fi
+else
+  echo -e "\e[1;33mWarning! The name and email address of the author of the $commit_hash commit is not checked in tolerant mode! \e[0m"
+  if echo "$actual_signed_off_by_line" | grep -q -v '^JerryScript-DCO-1.0-Signed-off-by:'
```
LaszloLango: I think this check should be a RegEx that checks the format after the colon (`firstname lastname [...] email@domain`). But that might cause similar problems to strict mode.

LaszloLango:
My only problem with the PR is that what happens if someone has a fake name in their GitHub account? Do we want wrong DCO-s on the master or should we check it manually before every commit?

LaszloLango:
@tilmannOSG, sorry. No offense :) In #1307 you merged a PR of Akos and he has the right settings in his GitHub account (Same name and e-mail address).

LaszloLango:
@tilmannOSG, yes we had examples to prove this behavior, but I cannot see back enough in Travis history page. I'll try to find them somehow.

LaszloLango:
@tilmannOSG, https://travis-ci.org/Samsung/jerryscript/builds/125007792 and https://travis-ci.org/Samsung/jerryscript/builds/122994701

akosthekiss:
@tilmannOSG: AFAIK, the rewrite happens whenever we merge with the "Squash and merge" button from the web. The new commit will have both the name and the email address coming from the GitHub profile of the user who opened the PR. At least, that was our experience hitherto. If the user's profile settings are the same as the settings of their git client locally, then all is good (as for my settings). Otherwise, the glitch happens as the commit author details change but the signed-off-by line does not.

@LaszloLango: I'm not sure whether we have any more chances to validate the identity of the contributor right now. All we can check automatically is whether the commit's signed-off-by message and author metadata match. Whether they are valid or even meaningful, can only be done by the reviewers. (E.g., if I set my local git to user.name Bla Bla and user.email to bla@bla.com, abd create a commit with a Bla Bla signed-off-by line, all checks will pass. But hopefully someone will ask me to use valid ID and update my commit.) And with this patch, the automated full strict sign check still happens for all opened PRs (up to the above detailed extent).

tools/check-signed-off.sh
akosthekiss: I'm not sure it's worth experimenting with a complex regex here. There will always be some exceptions we are not thinking of. How many parts may the name be composed of (of, van den, Jr., Sr., III)? What characters may the name consist of (ascii only, unicode accented chars for some crazy east European languages)? Dashes, dots, apostrophes, etc. And the email regex is easy to get bad as well. So, I'd leave this as is, even if a bit sloppy (or tolerant...)

LaszloLango: @akosthekiss, right. It isn't worth it. It will be good in this form.

tilmannOSG:
@akosthekiss I see, thanks for the clarification.  
Obviously we could also keep the process as is and reject contributions where the email address of the GitHub account doesn't match the email address of the commit. That would raise the barrier for contributions though.  
I'm not a lawyer but based on my understanding of the DCO what matters is the Signed-off-by line in the commit message. E.g. this line ties the contribution to a real name. So this should still work even if the author associated to the GitHub account is different.  
However, the real name used in the Signed-off-by line should match the real name associated with the GitHub account (it would be odd if they don't match). How about we modify the relaxed check to check for matching real names but allow a different email address?

akosthekiss:
@tilmannOSG: The problem is, till now, we haven't required the GitHub name/email to match the git user name/email, not even in the current process. We used to land patches by hand where no such restriction is/was imposed. We've always only be looking for the git commit author vs signed-off-by line match. And I've seen people around in the community where at least one of the details don't match: I've seen people with no email address published, with a different primary public email address, with a slightly different name (artificial example: Ákos Kiss on GitHub vs Akos Kiss in git config -- it's not like that in reality but better keep exemplifying myself). So, I'd still propose to keep the final check relaxed. (Before the PR gets landed, it will be checked with full strictness anyway.)

tilmannOSG:
@akosthekiss Ok, I see. Let's stay with the relaxed check for now then. We can always refine the process over time :)  
We should still encourage contributors to use the same email address they use for the DCO line as their primary GitHub address (or the other way around) to get similar results like we did with the old manual merge process.

akosthekiss:
@tilmannOSG (I'm not arguing purely for the sake of argument, just trying to make sure that everyone has the same understanding of the proposed change.)

Currently, we have a PR workflow like this:
```py
open PR
# assume that all test suites pass
if signed_off_by does not match commit_author:
    travis signals red for PR
    stop until fixed
else: # signed_off_by matches commit_author, thus travis signals green for PR
    if land manually:
        PR gets closed
        travis will signal green for master
    else: # land via web
        PR gets closed
        if signed_off matches github_profile:
            travis will signal green for master
        else: # signed_off_by does not match github_profile
            travis will signal red for master
```

This will change to:
```py
open PR
# assume that all test suites pass
if signed_off_by does not match commit_author:
    travis signals red for PR
    stop until fixed
else: # signed_off_by matches commit_author, thus travis signals green for PR
    land manually or via web
    PR gets closed
    travis will signal green for master
```

I.e., if travis signals green for a PR then landing that PR either way will not turn the master red. But the way (=strictness) of checking the PR does _not_ get changed with this patch.

tilmannOSG:
@akosthekiss Sorry, my last comment might have been a bit unclear. What I wanted to say is that with the new process we'll have a less consistent history (commits where Author and DCO line slightly deviate if the primary GitHub address of the contributor doesn't match the information in the DCO line). This happens even though we still enforce a matching Author and DCO line through Travis CI in the originating pull request.  
Ultimately that is the price for moving to the web interface based workflow. I'm hoping that frequent contributors will adjust their GitHub primary email address (or DCO line) to minimize those inconsistencies between the Author and DCO line. For one-off contributions I guess it's still reasonable to just accept them as is.

akosthekiss:
@tilmannOSG Ah, I get the point now. We are on the same page then.

I guess/hope that we have discussed the ideology behind the change. Still looking for technical reviews and verdicts :)

tools/check-signed-off.sh
```diff
@@ -15,6 +15,47 @@
...
+
+if [ "$1" == "--tolerant" ]
+then
+ TOLERANT="yes"
+ shift
+else
+ TOLERANT="no"
+fi
+
+if [ "$#" -gt 0 ]
```
tilmannOSG: Print help and exit if there's at least one argument? I guess that should be "-gt 1" instead?  
My understanding is that the intended behavior is as follows:

- No argument: Check in strict mode
- --tolerant: Check in tolerant mode
- --help: Print help
- More than 1 argument: Print help and exit with error code  

The help should probably say something like "[--help | --tolerant]" to emphasize that an invocation with "--help --tolerant" will error out.

tools/check-signed-off.sh
```diff
@@ -15,6 +15,47 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# Usage
+function usage
+{
+ echo "$0: Check Signed-off-by message of latest commit"
```
tilmannOSG: latest -> the latest

tools/check-signed-off.sh
```diff
@@ -15,6 +15,47 @@
...
+ echo ""
+ echo "Usage: $0 [--help] [--tolerant]"
+ echo ""
+ echo "Optional arguments:"
+ echo "  --help            print this help message"
+ echo "  --tolerant        check the existense of the message only but don't"
```
tilmannOSG: Typo: existense -> exisence

tilmannOSG:
@akosthekiss LGTM (minus added comments) :)

akosthekiss:
In addition to the typo/English fixes, I've updated the command line processing, hopefully it became better. @tilmannOSG PTAL

tilmannOSG:
@akosthekiss Thank you, fixes look good!

PR #1327, opened 2016-09-06
Make tools/settings.py lighter

akosthekiss:
It should not import things it does not use and it should define
settings data only (i.e., definition of functions should not happen
there).


zherczeg:
LGTM

tilmannOSG:
LGTM

PR #1328, opened 2016-09-06
Make jerry-main's link options configurable

akosthekiss:
Until now,
* the link map of jerry-main has always been generated, even though
  it is not used that often;
* static linking has always been forced except for the OSX target.

This patch adds configuration options for these features.

### DIFF
--- a/jerry-main/CMakeLists.txt
+++ b/jerry-main/CMakeLists.txt
@@ -17,23 +17,34 @@ cmake_minimum_required (VERSION 2.8.12)
 set(JERRY_NAME jerry)
 project (${JERRY_NAME} C)
 
+# Optional build settings
+set(ENABLE_LINK_MAP    OFF CACHE BOOL "Enable generating a link map file?")
+set(ENABLE_STATIC_LINK ON  CACHE BOOL "Enable static linking?")
+
+if("${PLATFORM}" STREQUAL "DARWIN")
+  set(ENABLE_STATIC_LINK "OFF")
+endif()
+
+# Status messages
+message(STATUS "ENABLE_LINK_MAP           " ${ENABLE_LINK_MAP})
+message(STATUS "ENABLE_STATIC_LINK        " ${ENABLE_STATIC_LINK})
+
 # Sources
 # Jerry standalone
 set(SOURCE_JERRY_STANDALONE_MAIN main-unix.c)
 
 # Generate map file
-if("${PLATFORM}" STREQUAL "DARWIN")
-  set(MAP_FILE_FLAGS "-Xlinker -map")
-else()
-  set(MAP_FILE_FLAGS "-Xlinker -Map")
+if(ENABLE_LINK_MAP)
+  if("${PLATFORM}" STREQUAL "DARWIN")
+    set(LINKER_FLAGS_COMMON "${LINKER_FLAGS_COMMON} -Xlinker -map -Xlinker jerry.map")
+  else()
+    set(LINKER_FLAGS_COMMON "${LINKER_FLAGS_COMMON} -Xlinker -Map -Xlinker jerry.map")
+  endif()
 endif()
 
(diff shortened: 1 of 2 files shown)


zherczeg:
LGTM

tilmannOSG:
@akosthekiss Is there any particular reason why we can't do static builds on macOS? Not sure whether it makes sense to do dynamic builds of JerryScript in the first place due to the self-contained nature (e.g. except for a tiny amount of glibc code nothing should be linked in). IMHO we should just default to static builds on all platforms and not offer an option to enable/disable static builds unless we can come up with a strong motivation for dynamic builds of JerryScript.  
Regarding the link map I think we should just keep that as is, the build time overhead is negligible and it's useful information for people interested JerryScript code size. E.g. not worth to introduce an additional option for that :)  
Don't get me wrong, I highly appreciate your efforts in improving the build system even further :)

akosthekiss:
@tilmannOSG: For the non-static builds, I have to reasons:
* We've had it before as well :)
* Actually, OSX simply does not support static binaries. You have to link against the standard lib framework and you have to do it dynamically. https://developer.apple.com/library/mac/qa/qa1118/_index.html That's why we've put the Darwin goard around the `-static` linker flag originally. (And that's why we always build without jerry-libc and jerry-libm on OSX.) At least, I haven't found any other way to build the jerry command line tool successfully.
* I have a Tizen box that has gcc and cmake, so theoretically it should be able to build jerry, but it turns out that it ships with libc.so only, no libc.a (and jerry-libc does not work there yet). Thus, the only way to build jerry-main is via dynamic linking.

There may be other platforms as well that we don't even have access to that may have one of the above limitations. I've proposed the change to make jerry useful for more users potentially. (And let's not forget that jerry-main is actually an example application "only", to show how the jerry-core lib may be used in a posix CLI environment. Other targets will/do implement their own tool around the lib anyway, and they will decide what linking mode they choose. Btw, I'm not even sure whether IoT.js forces static builds always; I've seen recent discussions about that.)

As for the link map: I agree that it can we a useful info for a developer. But it's not an artifact that a regular build system produces. My impression with the recent build system reworks was that we try to align with "tradition" in terms of what a build system produces and what not if invoked "just so", which usually means no extras that a non-dev user doesn't need. That's why I propose to put it behind an option.

tilmannOSG:
@akosthekiss I see, thanks for the clarification! It sounds like there are enough reasons to factor this out into a separate build option :)  
Regarding the link map, I would argue that for a project with such a strong focus on low-end microcontrollers, it is reasonable to produce data like this during the build to help the developer evaluate the effects of his/her changes on code size (in the same spirit like we require performance/code size data as part of the pull requests). I agree that for a user that just wants to build JerryScript this file is not relevant but then again it's unlikely that he/she will actually notice the file or the overhead of generating it. 
With 28 options (29 with the flag for static builds) tools/build.py --help is pretty much a "wall of text" even today and IMHO we should have a high bar for adding additional options exposed at this level (the goal is to have an easy to use build system after all :) ). I don't think a link map flag should be exposed at this level.  
This is probably a separate discussion but maybe we should consider having a different set of flags which are targeted specifically at JerryScript developers? (and not shown by default when invoking tools/build.py --help)

akosthekiss:
@tilmannOSG I'll update the patch then. But a question first (your final paragraph made me think): do you suggest to remove the configurability of link map from build.py only (and keep it in the CMakeLists with an ON default), or remove it from CMakeLists as well (and make it configurable only when we figure out how to add developer-targeted flags to build.py)?

tilmannOSG:
@akosthekiss I'm thinking as a compromise and to make progress on this maybe just keep the patch as is? And then we open an issue to discuss how to hide the more advanced developer-only options? What do you think?

akosthekiss:
works with me (as this requires the least effort from me :) )

tilmannOSG:
@akosthekiss Ok, I think we can proceed with the merge then :)  
LGTM from my side.  
I'll open an issue on the advanced options shortly.

PR #1330, opened 2016-09-07
Remove several ecma_op_object_get_[own_]property calls.

zherczeg:
The ecma_op_object_get_[own_]property calls should be phased out from
the project eventually and virtual properties should be introduced instead.

Negligible perf improvement (+0.282%) and 200 byte binary size reduction (157080->156824)

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
@@ -72,7 +72,6 @@ ecma_builtin_regexp_prototype_compile (ecma_value_t this_arg, /**< this argument
   }
   else
   {
-    ecma_string_t *pattern_string_p = NULL;
     uint16_t flags = 0;
 
     if (ecma_is_value_object (pattern_arg)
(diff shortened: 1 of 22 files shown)


jerry-core/ecma/operations/ecma-lex-env.c
```diff
@@ -270,19 +270,21 @@ ecma_op_get_binding_value (ecma_object_t *lex_env_p, /**< lexical environment */
 
     ecma_object_t *binding_obj_p = ecma_get_lex_env_binding_object (lex_env_p);
 
-    if (ecma_op_object_get_property (binding_obj_p, name_p) == NULL)
+    ecma_value_t result = ecma_op_object_find (binding_obj_p, name_p);
+
+    if (!ecma_is_value_found (result))
```
LaszloLango: Not sure, but I think we should call `ecma_value_free (result)` here.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -194,35 +163,204 @@ ecma_op_object_get_own_property (ecma_object_t *obj_p, /**< the object */
...
+ * @return ecma value if property is found
+ *         ECMA_SIMPLE_VALUE_NOT_FOUND if property is not found
+ *         Returned value must be freed with ecma_free_value
+ */
+ecma_value_t
+ecma_op_object_find (ecma_object_t *object_p, /**< the object */
+                     ecma_string_t *property_name_p) /**< property name */
+{
+  /* Circular reference is possible in JavaScript and testing it is complicated. */
+  int max_depth = 128;
```
LaszloLango: static const?

LaszloLango: My bad, leave as is.

LaszloLango:
LGTM

jerry-core/ecma/operations/ecma-lex-env.c
zherczeg: No because it is a simple value, so no need to free it.

LaszloLango: I mean if it is not an ECMA_SIMPLE_VALUE_NOT_FOUND then it might be anything else. There is not guarantee that is a simple value. What about adding an assert to `ecma_is_value_found` to check the parameter is always a simple value?

zherczeg: Oh no, we return with the simple value if it is "found".

LaszloLango: Description of `ecma_op_object_find` says
> "Returned value must be freed with ecma_free_value"

LaszloLango: Moreover almost everywhere you are using ECMA_TRY_CATCH and ECMA_FINALIZE where ECMA_FINALIZE calls `ecma_value free`

zherczeg: For any direct value calling free is optional. I am still unsure what is the issue here.

LaszloLango: My bad, code is good here.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -89,39 +89,45 @@ ecma_builtin_regexp_prototype_compile (ecma_value_t this_arg, /**< this argument
 
         /* Get source. */
         ecma_string_t *magic_string_p = ecma_get_magic_string (LIT_MAGIC_STRING_SOURCE);
-        ecma_property_t *prop_p = ecma_op_object_get_property (target_p, magic_string_p);
-        pattern_string_p = ecma_get_string_from_value (ecma_get_named_data_property_value (prop_p));
+        ecma_value_t source_value = ecma_op_object_get_own_data_prop (target_p, magic_string_p);
         ecma_deref_ecma_string (magic_string_p);
+        pattern_string_p = ecma_get_string_from_value (source_value);
+        ecma_deref_ecma_string (pattern_string_p);
```
dbatyai: Derefing `pattern_string_p` immediately after acquiring it feels a bit weird to me. Is this intentional?

zherczeg: That is a tricky code. The pattern_string_p is defined in the outside scope and wants to be a reference to an existing string depending on various conditions. Since target_p still has the reference to the original string, dereferencing this string does not destroy it. This logic was in the original code.

dbatyai: I see, but still, using a string we don't have reference to is just asking for trouble. Could you please move the deref after the `re_compile_bytecode` call in line `141`?

dbatyai:
LGTM

Issue #883, opened 2016-02-15
Missing "make install" on linux-x64 target


zoobab:
Hi,

I am missing a "make install" in cmake or autotools to install the final binaries on the system (I am on a linux 64 bits ubuntu machine).

I presume jerryscript binaries are those:
```
release.linux
release.linux-cp
release.linux-cp_minimal
release.linux-cp_minimal-mem_stats
release.linux-mem_stats
release.linux-mem_stress_test
```
However, to be installed a linux system, those are weird binary names. What about renaming them to:
```
jerryscript
jerryscript-cp
jerryscript-cp_minimal
jerryscript-cp_minimal-mem_stats
jerryscript-mem_stats
jerryscript-mem_stress_test
```
Furthermore, there are no simple way for a newbie to understand how this JS VM works, I tried on the command line:
```
root@7448a5be284f:/mnt/jerryscript/build2# cat ../example.js 
{
  jerry_init (JERRY_FLAG_ENABLE_LOG);

  char script [] = "print ('Hello, World!');";
  jerry_parse (script, strlen (script));

  jerry_run ();

  jerry_cleanup ();
}
root@7448a5be284f:/mnt/jerryscript/build2# ./release.linux ../example.js 
root@7448a5be284f:/mnt/jerryscript/build2# 
```

Having API documentation is good, but it looks like C code, while people expect JS code (a la nodejs).

akiss77:
@zoobab Where did you get that example.js from?... It's nowhere in the project tree AFAIK. It looks like a C source indeed. If you are looking for a first JS file to execute, try tests/hello.js.

LaszloLango:
Related PR #1192

PR #1335, opened 2016-09-08
Enable `make install`

akosthekiss:
With the new build system, the conventional `cmake && make` already
works. However, the last step, i.e., `make install` was still
missing (didn't work). This patch adds the `install()` commands to
CMakeLists that are required to generate the `install` target in
the Makefile.

### DIFF
--- a/jerry-core/CMakeLists.txt
+++ b/jerry-core/CMakeLists.txt
@@ -208,3 +208,6 @@ endif()
 if (JERRY_LIBM)
   target_include_directories(${JERRY_CORE_NAME} SYSTEM PRIVATE "${CMAKE_SOURCE_DIR}/jerry-libm/include")
 endif()
+
+install(TARGETS ${JERRY_CORE_NAME} DESTINATION lib)
+install(FILES jerry-api.h jerry-port.h DESTINATION include)
--- a/jerry-libc/CMakeLists.txt
+++ b/jerry-libc/CMakeLists.txt
@@ -47,3 +47,6 @@ add_library(${JERRY_LIBC_NAME} STATIC ${SOURCE_LIBC} ${TARGET_SPECIFIC_LIBC_SOUR
 target_compile_definitions(${JERRY_LIBC_NAME} PRIVATE ${DEFINES_LIBC})
 target_include_directories(${JERRY_LIBC_NAME} PRIVATE ${INCLUDE_LIBC})
 target_include_directories(${JERRY_LIBC_NAME} SYSTEM PUBLIC "${CMAKE_SOURCE_DIR}/jerry-libc/include")
+
+install(TARGETS ${JERRY_LIBC_NAME} DESTINATION lib)
+install(DIRECTORY ${INCLUDE_LIBC}/include/ DESTINATION include/jerry-libc)
--- a/jerry-libm/CMakeLists.txt
+++ b/jerry-libm/CMakeLists.txt
@@ -37,3 +37,6 @@ set_property(TARGET ${JERRY_LIBM_NAME}
 
 target_include_directories(${JERRY_LIBM_NAME} PRIVATE ${INCLUDE_LIBM})
 target_include_directories(${JERRY_LIBM_NAME} INTERFACE ${INCLUDE_LIBM})
+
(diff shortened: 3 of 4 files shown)


zherczeg:
I don't get the purpose of this patch. We don't compile with make, we compile with build.py, and I don't think we want to install anything.

robertsipka:
I think that you might add the unittests as well.

galpeter:
@zherczeg we use the build.py partially because you don't need to call the cmake directly. However if others using the project maybe want to skip that python file an directly use the cmake and additionally install the whole thing. Also I think there was an issue requesting 'make install' functionality somewhere.

akosthekiss:
At the core of our build system we have exactly that: cmake and make.

Actually, even the old build system was based on these two tools. But the problem with the old system was that it abused the whole cmake+make concept: it could not have been used in a way that would be expected by someone who is experienced in their use (and, say, packaging). The CMakeLists and Makefile was completely developer-focused, produced multiple build artifacts for multiple build configurations, etc.

However, for any "well-behaving" project, an "ordinary" user (who is not knee-deep in the project's development and does not want to do regular rebuilds an regression tests, etc.) usually does/expects:
* Run some configuration script, which normally detects what system the project is to be configured for, allows turning components and features on/off, specifies some locations (e.g., out-of-source-tree build directory): for automake-based systems, this is `configure OPTS`, for cmake-based projects, this is `cmake OPTS`.
* Build the project, create its build artifacts (executables, libraries) in the build directory: almost always, this is just `make [-j] [-C BUILDDIR]`, as the configuration tools generate Makefiles.
* If the user is "ordinary", they might want to use the project's results not straight from the build directory but from some well-known places, some of them are perhaps even on PATH or LD_LIBRARY_PATH, like "/usr/{bin,lib,include}", "/usr/local/{bin,lib,include}", or if noaccess is given to system dirs, "/home/user/local/{bin,lib,include}". (This has some extra nice features as well: e.g., the build dir is cluttered with object files and extra temporary outputs, it's not nice for the "final" artifacts to live in that mess; the build dir might be cleaned, and then all the built apps and libs and headers are gone, you cannot rely on them anymore; it's better to have something deployed to PATH that's proved to work, e.g., passed all the regression tests: the build dir cannot guarantee that.) This is what `make [DESTDIR=/not/usr/local] [-C BUILDDIR]` is for.

And right now, we have no support for this final step. Our CMakeLists have improved a lot, and we have build.py to ease our lives as developer (which does the first two steps  from above), but we have to think about others as well.

(E.g., I've tried to wrap jerry in an rpm package, which works quite well with a `cmake && make && make install` style workflow, but right now the final step has to be written with terrible manual copies.)

LaszloLango:
@galpeter, related issue #883

akosthekiss:
@robertsipka, I don't think that unittests are an install artifact. I'm not sure I'd like to have "/usr/bin/test-libm,test-api" on my system.

zherczeg:
I would wait with such installation until a Linux distribution wants to add jerry as a library for their package manager. I know they have lots of strict requirements.

akosthekiss:
@zherczeg Please, give examples. As, according to my best understanding, `make install` is not something of a Linux-specific expectation. It's the norm. (BTW, most distros and packaging systems -- e.g., RPM -- don't impose any (neither strict nor loose) requirements on the project itself, but work with the the sources (and build system) of the project as is, and wrap it according to their needs, e.g., parameterising configure for some specific directory layout, etc.) I'm insisting to add the `make install` feature.

tilmannOSG:
@akosthekiss Thanks for implementing support for this!  
Looking at the changes I assume invoking the install target will install the JerryScript binary, the static libraries and the header files, correct?  
I guess installing to an arbitrary location with -DCMAKE_INSTALL_PREFIX=\<path...\> will just work?

akosthekiss:
@tilmannOSG: correct. And install location can be specified either the way you've mentioned or by passing DESTDIR=<path> to make install.

tilmannOSG:
@akosthekiss Great! LGTM from my side.

LaszloLango:
LGTM

PR #1331, opened 2016-09-07
Make build.py's help output less of a "wall of text"

akosthekiss:
`tools/build.py --help` prints a lot of output, which is sometimes
hard to comprehend, especially since some described options are
useful for developers only.

This patch:
* shortens how options are displayed in the output with the help of
  the "metavar" feature of the argparse module ("--option {on, off}"
  becomes "--option X" and the choices are listed in the description
  only); and
* hides all developer-specific options from "--help", and reveals
  them only for "--devhelp".

As a result, help output becomes more readable, especially for
ordinary users, and needs less space.

Some additional changes:
* the options are ordered in a more logical way (although that's
  somewhat subjective);
* help strings start with lowercase letters (it aligns better with
  argparse's overall style); and
* rename "--cpointer_32_bit" to "--cpointer-32bit" (underscores are
  alien to command line options).

JerryScript-DCO-1.0-Signed-off-by: Akos Kiss akiss@inf.u-szeged.hu

PR #1336, opened 2016-09-09
Set the number of simultaneous jobs of the build.

robertsipka:

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -16,6 +16,7 @@
 # limitations under the License.
 
 import argparse
+import multiprocessing
 import shutil
 import subprocess
 import sys
@@ -43,6 +44,7 @@ def get_arguments():
     parser = argparse.ArgumentParser(parents=[devhelp_preparser])
     parser.add_argument('-v', '--verbose', action='store_const', const='ON', default='OFF', help='increase verbosity')
     parser.add_argument('--clean', action='store_true', default=False, help='clean build')
+    parser.add_argument('-j', '--jobs', metavar='N', action='store', type=int, default=multiprocessing.cpu_count() + 1, help='Allowed N build jobs at once (default: %(default)s)')
     parser.add_argument('--debug', action='store_const', const='Debug', default='Release', dest='build_type', help='debug build')
     parser.add_argument('--builddir', metavar='DIR', action='store', default=BUILD_DIR, help='specify output directory (default: %(default)s)')
     parser.add_argument('--lto', metavar='X', choices=['on', 'off'], default='on', help='enable link-time optimizations (%(choices)s; default: %(default)s)')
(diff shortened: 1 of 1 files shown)


akosthekiss:
Let this wait #1331

tools/build.py
```diff
@@ -59,6 +60,7 @@ def add_build_args(parser):
     parser.add_argument('--jerry-cmdline', choices=['on', 'off'], default='on', help='Build jerry command line tool (default: %(default)s)')
     parser.add_argument('--static-link', choices=['on', 'off'], default='on', help='Enable static linking of jerry command line tool (default: %(default)s)')
     parser.add_argument('--link-map', choices=['on', 'off'], default='off', help='Enable the generation of a link map file for jerry command line tool (default: %(default)s)')
+    parser.add_argument('--parallel-build', action='store',type=int, default=multiprocessing.cpu_count()+1, help='Allowed jobs at once (default: %(default)s)')
```
dbatyai: How about also adding a `-j` flag for it, just to be more conventional? :)

akosthekiss: And then, it's better off with `--jobs`

galpeter: we could add both `-j` and `--jobs` as one the argument parser supports it.

galpeter: Also missing spaces after `,` and around `+`. Also will the help really work? it seems to me that it should fail as we have an int default value and try to place it into the help as a string.

tools/build.py
```diff
@@ -128,7 +130,7 @@ def configure_build(arguments):
     return subprocess.call(cmake_cmd)
 
 def build_jerry(arguments):
-    return subprocess.call(['make', '--no-print-directory','-j', '-C', BUILD_DIR])
+    return subprocess.call(['make', '--no-print-directory','-j', str(arguments.parallel_build), '-C', BUILD_DIR])
```
galpeter: What will happen if I have a `0` value for the argument?

tools/build.py
robertsipka: Yes, it will work, it prints that default integer value.

tools/build.py
robertsipka: It will prints the 'make: the `-j' option requires a positive integral argument' message, and the build will fail.

tools/build.py
```diff
@@ -63,6 +64,7 @@ def devhelp(help):
     parser.add_argument('--static-link', metavar='X', choices=['on', 'off'], default='on', help='enable static linking of jerry command line tool (%(choices)s; default: %(default)s)')
     parser.add_argument('--strip', metavar='X', choices=['on', 'off'], default='on', help='strip release binary (%(choices)s; default: %(default)s)')
     parser.add_argument('--unittests', action='store_const', const='ON', default='OFF', help='build unittests')
+    parser.add_argument('-j', '--jobs', metavar='N', action='store', type=int, default=multiprocessing.cpu_count() + 1, help='Allowed N jobs at once (default: %(default)s)')
```
akosthekiss: I'd suggest:
* "Allowed N build jobs..."
* moving the option up under '--clean'

robertsipka:
Thanks, I've updated the patch based on your comments.

akosthekiss:
LGTM

LaszloLango:
LGTM

PR #1338, opened 2016-09-11
Improve the linking of libraries

akosthekiss:
Although both jerry-libc and jerry-libm have configuration options
that enable/disable their build, in practice, only jerry-libc can be
replaced with the system (compiler-default) libc. If jerry-libm is
disabled, the build of jerry-main fails, as there is no way to
instruct the linker to link the system libm to the binary. (The
build system does have a way to pass flags to the linker, but those
flags are listed before the linked objects. For the references to
get resolved correctly, the libraries to be linked have to be
specified _after_ the objects.)

This patch adds the EXTERNAL_LINK_LIBS configuration option to
CMakeLists, which ensures that the specified libraries get
correctly passed to the linker. (E.g, replacing jerry-libm with
system libm becomes possible with
`JERRY_LIBM=OFF EXTERNAL_LINK_LIBS='-lm'`.)

Additionally, the patch also makes the following related changes:

* Removes the COMPILER_DEFAULT_LIBC configuration option, as it is
  (almost) always the opposite of JERRY_LIBC. Moreover, its name is
  misleading: its only role is to add `-nostdlib` to the linker
  flags.

* Makes use of transitive library dependencies: if a library has
  another library as dependency, and it is linked to a binary, its
  dependency is linked as well. Thus, jerry-libc, jerry-libm, and
  any external libraries are added to jerry-core as dependency, and
  then only jerry-core is linked to executables (cmake will take
  care of the rest).

* build.py and run-tests.py follow up the changes, along with some
  minor syntax changes.

* Moves static linking option to global CMakeLists, as unit test
  binaries should be linked the same way as jerry-main.

* Adds EXTERNAL_COMPILER_FLAGS and EXTERNAL_LINKER_FLAGS as last to
  the flag list, to allow user override of (nearly) anything.

The patch speculatively follows up the build system changes in the
mbed, riot-stm32f4, and zephyr targets.


akosthekiss:
Example use cases:

Default, e.g., statically linked Linux binaries:
JERRY_LIBC=ON JERRY_LIBM=ON ENABLE_STATIC_LINK=ON

Linux or Tizen binaries dynamically linked to system libs:
JERRY_LIBC=OFF JERRY_LIBM=OFF EXTERNAL_LINK_LIBS='-lm' ENABLE_STATIC_LINK=OFF

OS X binaries:
JERRY_LIBC=OFF JERRY_LIBM=OFF ENABLE_STATIC_LINK=OFF
('-lSystem' is automatically added by the build system)

Experiment results on x86-64/Linux:

* tools/build.py: (default opts)
  * ldd: not a dynamic executable
  * size: 233573 text, 16 data, 2626072 bss
* tools/build.py: --jerry-libc=off
  * ldd: not a dynamic executable
  * size: 1012949 text, 7764 data, 2635776 bss
* tools/build.py: --jerry-libc=off --jerry-libm=off --link-lib='-lm'
  * ldd: not a dynamic executable
  * size: 1843941 text, 7900 data, 2635776 bss
* tools/build.py: --static-link=off
  * ldd: not a dynamic executable
  * size: 236517 text, 16 data, 2626072 bss
* tools/build.py: --jerry-libc=off --static-link=off
  * ldd: linux-vdso.so.1, libc.so.6, /lib64/ld-linux-x86-64.so.2
  * size: 232376 text, 704 data, 2626136 bss
* tools/build.py: --jerry-libc=off --jerry-libm=off --link-lib='-lm' --static-link=off
  * ldd: linux-vdso.so.1, libc.so.6, libm.so.6, /lib64/ld-linux-x86-64.so.2
  * size: 219850 text, 848 data, 2626136 bss

robertsipka:
I like this patch, especially that you removed the the COMPILER_DEFAULT_LIBC option, it was quite confusing. Could you please also update the GETTING-STARTED documentation?

jerry-core/CMakeLists.txt
```diff
@@ -201,13 +201,19 @@ add_library(${JERRY_CORE_NAME} STATIC ${SOURCE_CORE})
 target_compile_definitions(${JERRY_CORE_NAME} PUBLIC ${DEFINES_JERRY})
 target_include_directories(${JERRY_CORE_NAME} PUBLIC ${INCLUDE_CORE})
 
-if (JERRY_LIBC)
+set(JERRY_LIBS )
```
LaszloLango: Is the space before the left parenthesis intentional?

akosthekiss: Yes, JERRY_LIBS is set to empty string. Without that space it just looked like a predicate ("if(JERRY_LIBS)") to me.

LaszloLango:
LGTM

akosthekiss:
@sergioamr Could you please take a look at the zephyr-related part? I did my best guess not to break the build there but could not try it in practice.

sergioamr:
@poussa Sakari, i will be in the thingmonk conference this week so i will probably don't have access to build this. Could you have a look? Thanks

dbatyai:
LGTM

akosthekiss:
@sergioamr @poussa Any feedback that would block landing this?

sergioamr:
@akosthekiss Sorry, still at Thinkmonk conference. Tomorrow i will be back in the office.

poussa:
Sorry here as well - did not had time for this recently.

sergioamr:
LGTM, sorry about the delay

akosthekiss:
@sergioamr Thanks a lot!

PR #1343, opened 2016-09-12
Add DCO check to gh-pages branch

### LINKED ISSUE #1342: gh-pages branch needs an automatic DCO check
On master we are using Travis CI to do various checks around the DCO line in the commit messages. We need to enable the same functionality for the gh-pages branch as well.

akosthekiss:
Resolves #1342

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,13 @@
+language: c
+os: linux
+
+install:
+  - wget https://raw.githubusercontent.com/${TRAVIS_REPO_SLUG}/master/tools/check-signed-off.sh
+  - chmod u+x check-signed-off.sh
+script: ./check-signed-off.sh --travis
+
+branches:
+  only:
+  - master
+  - gh-pages
+  - /.*/


akosthekiss:
Strange, Travis does not seem to get triggered for the PR. Worked for me on my branch...

tilmannOSG:
LGTM  
Maybe we have to enable it somewhere?

LaszloLango:
Looks like it is not working as expected.

akosthekiss:
How so? I see no Travis "green check mark" icon here.
BTW, I wrote to support, but got no answer yet.

akosthekiss:
Got no answer but found explanation: https://docs.travis-ci.com/user/customizing-the-build#Safelisting-or-blocklisting-branches
And with some experimentation, Travis CI seems to get triggered for gh-pages as well.

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+os: linux
+
+install:
+  - wget https://raw.githubusercontent.com/${TRAVIS_REPO_SLUG}/master/tools/check-signed-off.sh
+  - chmod u+x check-signed-off.sh
+script: ./check-signed-off.sh --travis
+
+branches:
+  only:
+  - master
```
LaszloLango: Do we really need this line?

.travis.yml
```diff
@@ -0,0 +1,13 @@
...
+  - gh-pages
+  - /.*/
```
LaszloLango: and this?

akosthekiss:
@LangoLaszlo: To be honest, this is right now in a "works for me" status.

A bit more detail: as the Travis docs state, gh-pages is excluded from the built branches by default. However, to my best understanding, there is no such thing in .travis.yml like "branches: include:", we only have "branches: only:". That is, if I list anything under "branches: only:",  it has to be complete (anything not listed there will not be built/tested). If I only list "gh-pages" here, then no other branch will be built.

My usual workflow and why it would be a problem to me: I have Travis enabled for my own fork as well, so whenever I do something locally and push it to my fork, Travis tests that push. Only if my Travis reports me OK, then do I open a pull request (where origin's Travis will test it again, of course). Moreover, I try not to work on the master branch directly but create a new branch for every fix/feature/pull request. If I work on the web site, gh-pages takes over the role of the master branch. Thus, I branch off gh-pages under a new name, work on it, push it to my fork, wait for the green signal, and open a pull request against origin:gh-pages. However, as the name of the branch is _not_ gh-pages, my fork's Travis does not test it _unless_ that name is on its whitelist as well. As we cannot guess or prescribe in advance what names we will use, I went for a catch-all regex.

So, this is the reason for the regex part.

However, when I've tried to have the regex _only_, origin's Travis still didn't do anything. And there was a comment in the doc that because of some legacy reasons, all active branches have to have a .travis.yml (or something like that). And here comes the experimental and "works for me" part: I've added the exact names of the two key active branches of the project (master and gh-pages), and lo!, origin's Travis got triggered...

LaszloLango:
@akosthekiss it's odd, because there are two .travis.yml in this project after this PR. One in gh-pages in one in the master. Moving "master" and the regexp branch into the other travis.yml might work, but you cannot do it in one PR. The other strange thing is the option in Travis UI "Build only if .travis.yml is present" is enabled, but we need "branches:" in the yml. IMHO we can land this PR in the current form. LGTM

akosthekiss:
@LaszloLango: We need "branches:" because even if "build only if .travis.yml is present" is enabled, gh-pages is excluded by default. So we have to add that back via whitelisting. (That's in the linked docs.)

The other thing is : I don't see a real issue from having two different .travis.yml files in the master and gh-pages branches: they are actually completely different repositories merged into one (and I don't think that master's .travis.yml would need any adjustment).

To me, the strange thing was that a single catch-all regex was not enough in the gh-pages .travis.yml. Anyway, whatever...

Thanks for the review!

PR #1347, opened 2016-09-13
Move DCO from the Wiki into the source tree.

tilmannOSG:
All the files related to licensing/contributing should be directly included in the project repository.

### DIFF
--- a/DCO.md
+++ b/DCO.md
@@ -1,0 +1,44 @@
+# JerryScript Developer's Certificate of Origin
+
+The JerryScript project uses the signed-off-by language and process, to give us a clear chain of trust for every patch received.
+
+> By making a contribution to this project, I certify that:
+
+> (a)	The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or
+
+> (b)	The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or
+
+> (c)	The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.
+
+> (d)	I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project, under the same open source license.
+
+
+## Using the Signed-Off-By Process
+
+We have the same requirements for using the signed-off-by process as the Linux kernel. In short, you need to include a signed-off-by tag in every patch:
+
+"JerryScript-DCO-1.0-Signed-off-by:" this is a developer's certification that he or she has the right to submit the patch for inclusion into the project. It is an agreement to the Developer's Certificate of Origin (above). **Code without a proper signoff cannot be merged into the mainline.**
+
+You should use your real name and email address in the format below:
+
+> JerryScript-DCO-1.0-Signed-off-by: Random J Developer random@developer.example.org
+
+
+## How to add the DCO line to every single commit automatically
+
+It is easy to forget adding the DCO line to the end of every commit message. Fortunately there is a nice way to do it automatically. Once you've cloned the repository into your local machine, you can add `prepare commit message hook` in `.git/hooks` directory like this:
+
+```
+#!/usr/bin/env python
+
+import sys
+
+commit_msg_filepath = sys.argv[1]
+
+with open(commit_msg_filepath, "r+") as f:
+	content = f.read()
+	f.seek(0, 0)
+	f.write("%s\n\nJerryScript-DCO-1.0-Signed-off-by: <Your Name> <Your Email>" % content)
+```
+
+Please refer [Git Hooks](http://git-scm.com/book/en/v2/Customizing-Git-Git-Hooks) for more information.


akosthekiss:
The proposed file name (DCO.md) is quite unusual. I mean, we have been working with the project for some time now, so we certainly know what DCO stands for, but for a newcomer, it does not necessarily look familiar. I'd rather look for a CONTRIBUTING.md in the project root folder.

PS (for the future): Some files (e.g., README.md, LICENSE, CMakeLists.txt, Doxyfile, and this DCO|CONTRIBUTING.md) are rightfully in the root directory. However, I would not add any more here in the future but in the docs dir.

DCO.md
```diff
@@ -0,0 +1,44 @@
...
+> (c)	The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.
+
+> (d)	I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project, under the same open source license.
+
+
+### Using the Signed-Off-By Process
+
+We have the same requirements for using the signed-off-by process as the Linux kernel. In short, you need to include a signed-off-by tag in every patch:
+
+"Signed-off-by:" this is a developer's certification that he or she has the right to submit the patch for inclusion into the project. It is an agreement to the Developer's Certificate of Origin (above). **Code without a proper signoff cannot be merged into the mainline.**
```
akosthekiss: The beginning of the paragraph is misleading. It's not "Signed-off-by:" but "JerryScript-DCO-1.0-Signed-off-by:", as stated below as well. This might seem picky, but we have seen problems from such slight misinterpretations...

DCO.md
```diff
@@ -0,0 +1,44 @@
...
+
+You should use your real name and email address in the format below:
+
+> JerryScript-DCO-1.0-Signed-off-by: Random J Developer random@developer.example.org
+
+
+#### How to add DCO every single commit automatically.
+
+It is easy to forget adding DCO end of every commit message. Fortunately there is a nice way to do it automatically. Once you've clone the repository into your local machine, you can add `prepare commit message hook` in `.git/hooks` directory like this:
```
akosthekiss: (syntax)
adding DCO +to the+ end of
Once you've clone+d+

DCO.md
```diff
@@ -0,0 +1,44 @@
...
+> (a)	The contribution was created in whole or in part by me and I have the right to submit it under the open source license indicated in the file; or
+
+> (b)	The contribution is based upon previous work that, to the best of my knowledge, is covered under an appropriate open source license and I have the right under that license to submit that work with modifications, whether created in whole or in part by me, under the same open source license (unless I am permitted to submit under a different license), as indicated in the file; or
+
+> (c)	The contribution was provided directly to me by some other person who certified (a), (b) or (c) and I have not modified it.
+
+> (d)	I understand and agree that this project and the contribution are public and that a record of the contribution (including all personal information I submit with it, including my sign-off) is maintained indefinitely and may be redistributed consistent with this project, under the same open source license.
+
+
+### Using the Signed-Off-By Process
```
akosthekiss: strange heading level(s): file starts with heading level 1, then this one is level 3, and gets followed by a level 4 heading

DCO.md
```diff
@@ -0,0 +1,44 @@
...
+We have the same requirements for using the signed-off-by process as the Linux kernel. In short, you need to include a signed-off-by tag in every patch:
+
+"Signed-off-by:" this is a developer's certification that he or she has the right to submit the patch for inclusion into the project. It is an agreement to the Developer's Certificate of Origin (above). **Code without a proper signoff cannot be merged into the mainline.**
+
+You should use your real name and email address in the format below:
+
+> JerryScript-DCO-1.0-Signed-off-by: Random J Developer random@developer.example.org
+
+
+#### How to add DCO every single commit automatically.
```
akosthekiss: should we (or shouldn't we) use Title Case for this heading as well?

DCO.md
tilmannOSG: Good catch, fixed :)

DCO.md
tilmannOSG: Fixed.

DCO.md
tilmannOSG: Changed to a level 1 heading at the beginning of the document and then level 2 headings for consecutive sections.

DCO.md
tilmannOSG: Changed to a level 2 heading.

tilmannOSG:
@akosthekiss Thanks for the review! I guess we should do a thorough review of all the Wiki pages eventually :)  
The idea is that CONTRIBUTING.md links to DCO.md (see #1348). In theory we could embed DCO.md into CONTRIBUTING.md but since the content of DCO.md has legal implications it's beneficial too keep it isolated in a separate file, e.g. in the same spirit like LICENSE (makes things easier if we want to change the contents of the DCO in the future).  
I fully agree that we should have a high barrier for adding additional files to the project root.

akosthekiss:
I'm not sure that - in its current form - the file is meeting that barrier.

IMHO:
- I agree that there is some legal stuff in the current doc but the rest is discussion and explanation and guidance. Things I'd expect from the CONTRIBUTING.md.
- Thus, I'd split out the legal part ("By making a contribution to this project, I certify that ...") into a file named "[JerryScript-]DCO-1.0[.md]". (Extra reason behind that is that the Signed-off-by line is referring to a version 1.0 of the DCO but the version number did not appear anywhere until now.)
- The remaining part ("The JerryScript project uses the signed-off-by language ..." and "We have the same requirements for using the signed-off-by process ...") could go to CONTRIBUTING.md's DCO section.
- The technical tips'n'tricks part could go to the end of CONTRIBUTING.md, which is already proposed to contain some technical help.

Thus, I'd suggest merging this PR into #1348.

LaszloLango:
+1 to merge this into #1348

tilmannOSG:
@akosthekiss @LaszloLango Thanks for the feedback!  
Merging this into #1348 sounds like a good idea to me, I'll proceed with that.

tilmannOSG:
Closing, content has been moved into #1348.

PR #1349, opened 2016-09-13
Remove obsolete dependencies from install scripts

akosthekiss:
The new build script does not support ninja anymore, and we haven't
been building bare-metal arm targets for long. This patch removes
now-obsolete dependencies from install scripts.

This can also help to reduce the load on Travis CI (as it keeps
installing dependencies for each build job over and over again).

### DIFF
--- a/tools/apt-get-install-deps.sh
+++ b/tools/apt-get-install-deps.sh
@@ -17,6 +17,6 @@
 
 sudo apt-get update -q
 sudo apt-get install -q -y \
-    make cmake ninja-build \
-    gcc gcc-arm-none-eabi \
+    make cmake \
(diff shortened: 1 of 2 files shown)


LaszloLango:
LGTM

tilmannOSG:
LGTM  
I think we should consider to keep building with Ninja though :) (although I guess that means adding yet another option to build.py). Or we change build.py to check whether Ninja is installed and if so use it instead of Make?  
Regarding the bare metal support I think in the long-term it would be nice to have Travis CI testing support for the bare metal targets as well :) (e.g. testing the NuttX/RIOT/Zephr/mbed OS ports using QEMU)

PR #1353, opened 2016-09-16
Add license checking

akosthekiss:
It's stated in the Guidelines that all contributions must be under
the Apache License 2.0. To avoid potential mistakes from manual
reviews, this patch adds the check-license.py script to
automatically check all source files for license headers.

Travis CI is also configured to run the check.

Fallout: it turned out that some files already in the code base
either miss a license header or have some minor typo differences.
The patch fixes up these deficiences.


akosthekiss:
Target contributors, could you please take a look at the files that got license added/fixed (license text, copyright year(s) and holder(s)), and give your approval if they are OK with you?

@seanshpark: esp8266
@knightburton: mbed
@pfalcon: zephyr

pfalcon:
Yes, zephyr changes look find by me. Some files were authorized by @sergioamr (I'm sure he's not against punctuation changes, just to keep in loop).

sergioamr:
Thanks @pfalcon , LGTM

knightburton:
Yes, the mbed is looks good to me. However @polaroi8d is also worked on this.

polaroi8d:
Thanks @knightburton for the mention, look good to me.

zherczeg:
LGTM

seanshpark:
1) ` targets/esp8266/gen_misc.sh` is from ESP developers without changes and I don't know we can add the license.
2) disagree on adding license texts to small `main.js`, `blink.js` files as they will eat up the `RO` memories, so alternatively is there any possible way to add one line comment pointing to the `LICENSE` file?

akosthekiss:
@seanshpark 1) I don't know exactly what to do then as guidelines state explicitly that only something with Apache 2.0 license can go in the code base. We surely cannot put own copyright notice on it as a start. What about the license, where is the code originating from, what is it's original license?

2) What about adding a comment remover to js2c? Then all sources could bear license header in their source form but when put in the C array, it would be shrinked to a "minimal" size. (All targets could benefit from this approach, not only ESP.)

seanshpark:
@akosthekiss , 1) @somang-park, I think you can decide. Original LICENSE is bottom side of `https://github.com/Samsung/jerryscript/blob/master/targets/esp8266/LICENSE` file
2) good idea :)

tilmannOSG:
@seanshpark Is there a particular reason for including gen_misc.sh in the JerryScript source tree? (since you mentioned that it is unmodified) E.g. can't we just let the Makefiles of the ESP8266 port reference this file in the ESP8266 SDK without keeping a copy in our source tree?

seanshpark:
@tilmannOSG , I hope someone could look inside the script.

akosthekiss:
Updated the PR. The ESP target has been removed from the list of checked directories as its status needs further review but that should not block the introduction of automated license checks.

zherczeg:
LGTM

PR #1357, opened 2016-09-17
Fix doxygen-related documentation issues

akosthekiss:

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -43,7 +43,7 @@
 /**
  * Assert that specified object type value is valid
  *
- * @param object's implementation-defined type
+ * @param type object's implementation-defined type
  */
 #ifndef JERRY_NDEBUG
 #define JERRY_ASSERT_OBJECT_TYPE_IS_VALID(type) \
--- a/jerry-core/parser/js/js-parser-internal.h
+++ b/jerry-core/parser/js/js-parser-internal.h
@@ -273,6 +273,18 @@ typedef struct
 #endif /* PARSER_DUMP_BYTE_CODE */
 } parser_context_t;
 
+/**
+ * @}
+ * @}
+ * @}
+ *
+ * \addtogroup mem Memory allocation
+ * @{
+ *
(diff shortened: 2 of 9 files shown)


LaszloLango:
LGTM

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -42,7 +42,7 @@
 /**
  * Assert that specified object type value is valid
  *
- * @param object's implementation-defined type
+ * @param type object's implementation-defined type
```
zherczeg: Why do you change this?

jerry-core/parser/js/js-parser-internal.h
```diff
@@ -273,6 +273,19 @@ typedef struct
...
+ * @}
+ */
+
+/** \addtogroup mem Memory allocation
+ * @{
+ *
+ * \addtogroup mem_parser Parser memory manager
+ * @{
+ */
+
```
zherczeg: Do we need this amount fo grouping?

jerry-core/ecma/operations/ecma-objects.c
akosthekiss: Because `@param` needs the name of the parameter it is documenting (which is type in this case).

jerry-core/parser/js/js-parser-internal.h
akosthekiss: It's needed because the implementations of the functions defined in this header are put in these different groups. So either the header should group them separately just like the sources do, or the sources should put all the implementations in the same group as the header did.

zherczeg: I get it. But adding these extra lines to the header does not looks nice. Is there a way to put these comments into one line?

akosthekiss: I could shrink things a bit (some empty line vanished) but this is the limit we can reach (provided that we follow our own style used for doc comments).
Patch updated.

zherczeg: It seems improved. Ok, leave it this way.

PR #1358, opened 2016-09-17
Unify (and fix) the naming convention of structures

akosthekiss:
Several `struct`s are defined as `typedef struct x_t { } x_t`. This
is bad style. If the struct has no self-recursive pointer member,
then `typedef struct { } x_t` is enough. Otherwise, the `struct`
needs a name but it should be different from the name of the
typedef: `typedef struct x_s { } x_t` is a widely used naming
convention. This patch applies this style consistently throughout
the code base.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -533,7 +533,7 @@ typedef enum
  * Description of ECMA-object or lexical environment
  * (depending on is_lexical_environment).
  */
-typedef struct ecma_object_t
+typedef struct
 {
   /** type : 3 bit : ecma_object_type_t or ecma_lexical_environment_type_t
                      depending on ECMA_OBJECT_FLAG_BUILT_IN_OR_LEXICAL_ENV
@@ -891,7 +891,7 @@ typedef enum
 /**
  * ECMA string-value descriptor
  */
-typedef struct ecma_string_t
+typedef struct
 {
   /** Reference counter for the string */
   uint16_t refs_and_container;
--- a/jerry-core/jmem/jmem-allocator.h
+++ b/jerry-core/jmem/jmem-allocator.h
@@ -103,9 +103,9 @@ typedef struct
 /**
  * Node for free chunk list
  */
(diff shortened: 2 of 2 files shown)


jerry-core/vm/vm-stack.h
```diff
@@ -43,7 +43,7 @@ typedef struct
 /**
  * ECMA stack frame
  */
-typedef struct vm_stack_frame_t
+typedef struct
 {
   struct vm_stack_frame_t *prev_frame_p;      /**< previous frame */
```
LaszloLango: `vm_stack_frame_s`

akosthekiss: You're right. No idea how I missed that. Fixed.

jerry-core/parser/js/js-parser-internal.h
```diff
@@ -118,9 +118,9 @@ typedef struct
  * All data allocated by the parser is
  * stored in parser_data_pages in the memory.
  */
-typedef struct parser_mem_page_t
+typedef struct parser_mem_page_s
```
zherczeg: I would rather keep these. The _s suffix is very unusual to me. Furthermore in C struct x; and typedef ... x; are different types.

akosthekiss: I'm open to discussions. I've seen quite some opinion wars "out there", I surely don't want to start one. If there are more voting for `typedef x_t { ... } x_t;` then I can drop this. (Or just keep that part of the patch that removes the struct tags where not needed. Or... add struct tags everywhere. Really, the key motivation for this PR was consistency throughout the code base.)

zherczeg: Please keep the current one.

LaszloLango: I'm Ok with the current form. Just keep that part of the patch that removes the struct tags where not needed.

akosthekiss:
Updated according to reviews.

LaszloLango:
still LGTM

zherczeg:
LGTM

PR #1360, opened 2016-09-18
Sync cmake and build.py defaults

akosthekiss:
* Cmake had LTO OFF, build.py had LTO ON by default. ON became the
  common setting.
* Cmake had snapshot save/exec OFF, build.py had them ON by
  default. OFF became the common setting.


LaszloLango:
LGTM

tools/build.py
```diff
@@ -51,8 +51,8 @@ def devhelp(help):
     parser.add_argument('--all-in-one', metavar='X', choices=['on', 'off'], default='off', help='all-in-one build (%(choices)s; default: %(default)s)')
     parser.add_argument('--profile', metavar='PROFILE', choices=['full', 'minimal'], default='full', help='specify the profile (%(choices)s; default: %(default)s)')
     parser.add_argument('--error-messages', metavar='X', choices=['on', 'off'], default='off', help='enable error messages (%(choices)s; default: %(default)s)')
-    parser.add_argument('--snapshot-save', metavar='X', choices=['on', 'off'], default='on', help='enable saving snapshot files (%(choices)s; default: %(default)s)')
-    parser.add_argument('--snapshot-exec', metavar='X', choices=['on', 'off'], default='on', help='enable executing snapshot files (%(choices)s; default: %(default)s)')
+    parser.add_argument('--snapshot-save', metavar='X', choices=['on', 'off'], default='off', help='enable saving snapshot files (%(choices)s; default: %(default)s)')
+    parser.add_argument('--snapshot-exec', metavar='X', choices=['on', 'off'], default='off', help='enable executing snapshot files (%(choices)s; default: %(default)s)')
```
zherczeg: Why snapshot is disabled by default?

akosthekiss: Answer 1: CMakeLists.txt and build.by were out of sync. I had to choose which one to change, I chose build.py in this case.

Answer 2: All `FEATURE_` CMake configuration settings are OFF by default. I think/hope that's on purpose, to keep every feature that's not strictly needed for ES5.1 conformance and would increase running time, memory consumption, or binary size as opt-in (i.e., needing explicit action to get it enabled).

zherczeg: I think snapshot is enabled currently, so some people might be surprised but lets disable it.

zherczeg:
LGTM

PR #1361, opened 2016-09-18
Remove unnecessary jerry internal header file.

bsdelf:
For developers who want to check version numbers of jerry snapshot binaries, it is nature to include "jerry-snapshot.h" and use macro `JERRY_SNAPSHOT_VERSION`. Therefore "jerry-snapshot.h" should not rely on any jerry internal stuff.

### DIFF
--- a/jerry-core/jerry-snapshot.h
+++ b/jerry-core/jerry-snapshot.h
@@ -17,8 +17,6 @@
 #ifndef JERRY_SNAPSHOT_H
 #define JERRY_SNAPSHOT_H
 
-#include "ecma-globals.h"
-
 /**
  * Snapshot header
  */


akosthekiss:
The removed include is apparently not needed there, that's good.

I'm confused by the justification of the PR, however, that jerry-snapshot.h is being accessed from "outside". AFAIK, this has always been an internal header, on which users of the library should not rely. Am I wrong?

cc @zherczeg

zherczeg:
Without header, who defines uint32_t ? This is not a native type. Perhaps we could move this type to jerry-internal.h, and leave the snapshot version only here.

bsdelf:
@zherczeg You are right on `uint32_t`. But why not treat "jerry-snapshot.h" like "elf.h", and even prepend a "magic" field to the struct, so that jerry snapshot can be easily identified?

akosthekiss:
Note: `uint32_t` is a standard type in C99, defined in `<stdint.h>`. We can include it safely here as well, as we are including the same header elsewhere, too.

zherczeg:
To tell the truth jerry-snapshot.h is not designed to be a public header. It was created when the parser needed it, and I hoped it will go away at some point (it could now).

However if we want make it a public header we could think about what we want there. Providing the snapshot version is reasonable for me, but the header format is not. This would solve the dependency issue.

bsdelf:
Thanks for the explaination. If jerry-snapshot.h is not designed to be public, the pull request can be closed now.

zherczeg:
Still, the question is open: whether we want to add the supported snapshot version to the API.

zherczeg:
I am closing this issue. Please reopen if needed.

PR #1368, opened 2016-09-21
Fix compiler error about between type and format string

chokobole:
IoT.js just integrated with jerryscript updated, and the build for Nuttx was failed. So my colleague opened an issue https://github.com/Samsung/iotjs/issues/398

While re-enabling build, I found out several messages regarding type conversion because of platform depenency like below.

```bash
/home/chokobole/Workspace/iotjs/iotjs/deps/jerry/jerry-core/ecma/operations/ecma-regexp-object.c:615:11: error: format '%d' expects argument of type 'int', but argument 3 has type 'uint32_t' [-Werror=format=]
           JERRY_TRACE_MSG ("num_of_ranges=%d, ch1=%d, ch2=%d, curr_ch=%d; ",
```

~~To workaround errors, I used standard format string macro defined at inttypes.h, but I don't know if there have better solutions.~~

**FIX**

*Accepting @zherczeg and @akosthekiss 's opinion, I just simply casted around the type according to format string.*

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.c
@@ -113,7 +113,7 @@ ecma_builtin_global_object_print (ecma_value_t this_arg, /**< this argument */
                                                               0,
                                                               JERRY_BITSINBYTE);
 
-        jerry_port_console ("\\u%02x%02x", byte_high, byte_low);
+        jerry_port_console ("\\u%02x%02x", (unsigned int) byte_high, (unsigned int) byte_low);
       }
     }
 
--- a/jerry-core/ecma/operations/ecma-regexp-object.c
+++ b/jerry-core/ecma/operations/ecma-regexp-object.c
@@ -415,7 +415,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
         }
 
         ecma_char_t ch = lit_utf8_read_next (&str_curr_p);
-        JERRY_TRACE_MSG ("Period matching '.' to %d: ", (uint32_t) ch);
+        JERRY_TRACE_MSG ("Period matching '.' to %u: ", (unsigned int) ch);
 
         if (lit_char_is_line_terminator (ch))
         {
@@ -608,8 +608,9 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
         {
           ecma_char_t ch1 = re_canonicalize (re_get_char (&bc_p), is_ignorecase);
           ecma_char_t ch2 = re_canonicalize (re_get_char (&bc_p), is_ignorecase);
-          JERRY_TRACE_MSG ("num_of_ranges=%d, ch1=%d, ch2=%d, curr_ch=%d; ",
-                           num_of_ranges, ch1, ch2, curr_ch);
+          JERRY_TRACE_MSG ("num_of_ranges=%u, ch1=%u, ch2=%u, curr_ch=%u; ",
+                           (unsigned int) num_of_ranges, (unsigned int) ch1,
+                           (unsigned int) ch2, (unsigned int) curr_ch);
 
           if (curr_ch >= ch1 && curr_ch <= ch2)
           {
@@ -644,7 +645,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
         uint32_t backref_idx;
 
         backref_idx = re_get_value (&bc_p);
-        JERRY_TRACE_MSG ("Execute RE_OP_BACKREFERENCE (idx: %d): ", backref_idx);
+        JERRY_TRACE_MSG ("Execute RE_OP_BACKREFERENCE (idx: %u): ", (unsigned int) backref_idx);
         backref_idx *= 2;  /* backref n -> saved indices [n*2, n*2+1] */
         JERRY_ASSERT (backref_idx >= 2 && backref_idx + 1 < re_ctx_p->num_of_captures);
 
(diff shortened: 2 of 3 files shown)


jerry-libc/jerry-libc-defs.h
```diff
@@ -20,6 +20,7 @@
 #include <stdbool.h>
 #include <stddef.h>
 
+
```
yichoi: Is it necessary ?

yichoi:
DCO is not included in your patch, so Travis shows fail message.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1148,7 +1148,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
       }
       default:
       {
-        JERRY_TRACE_MSG ("UNKNOWN opcode (%d)!\n", (uint32_t) op);
+        JERRY_TRACE_MSG ("UNKNOWN opcode (%" PRIu8 ")!\n", op);
```
akosthekiss: `op` is of type `re_opcode_t`, which is an enum, thus its size is implementation specific. Therefore `op` should be cast to `uint8_t` here to be portable.

jerry-core/jerry-api.h
```diff
@@ -21,6 +21,7 @@
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
+#include <inttypes.h>
```
akosthekiss: We should not pollute the API header, especially as the macros are only needed by the implementation. This include has a better place in jrt-libc-includes.h .

jerry-core/ecma/builtin-objects/ecma-builtin-global.c
```diff
@@ -113,7 +113,7 @@ ecma_builtin_global_object_print (ecma_value_t this_arg, /**< this argument */
                                                               0,
                                                               JERRY_BITSINBYTE);
 
-        jerry_port_console ("\\u%02x%02x", byte_high, byte_low);
+        jerry_port_console ("\\u%" PRIu32 "02%" PRIu32 "02x", byte_high, byte_low);
```
zherczeg: Is this a standard format?

akosthekiss: AFAIK, yes: http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/inttypes.h.html

akosthekiss: I have to revise myself. @zherczeg is right that something is not OK here.
`"\\u%02" PRIx32 "%02" PRIx32`
is the correct form (it is PRIx32 not PRIu32, and it's replacing the "x" in the format string)

jerry-core/parser/regexp/re-compiler.c
```diff
@@ -264,7 +264,7 @@ re_parse_alternative (re_compiler_ctx_t *re_ctx_p, /**< RegExp compiler context
       }
       case RE_TOK_CHAR:
       {
-        JERRY_TRACE_MSG ("Compile character token: %c, qmin: %d, qmax: %d\n",
+        JERRY_TRACE_MSG ("Compile character token: %" PRIu32 ", qmin: %" PRIu32 ", qmax: %" PRIu32 "\n",
```
akosthekiss: This will not produce the same output as the original: the first argument should not be rewritten from c to PRIu32 but the argument itself should be cast to char.

jerry-core/ecma/builtin-objects/ecma-builtin-global.c
zherczeg: What about simply casting the arguments to (int) ? Especially since a byte is not an int32 type.

akosthekiss: For the sake of precision: `byte_high` and `byte_low` _are_ of `uint32_t` type. However, basically, I start agreeing with you: it might be best keeping the "simple" format strings, using non-macro format specifiers and up-casting the arguments to int or unsigned int. (I guess, the existing format strings could work in most cases, perhaps some "%d"s might need a rewrite to "%u" to keep sign-correctness.)
@zherczeg Is this aligned with your ideas?

jerry-core/jerry-api.h
chokobole: @akosthekiss  okay i will do reflect this.

jerry-core/parser/regexp/re-compiler.c
chokobole: @akosthekiss  I agree with both of  them and i will reflect it, too.

jerry-core/ecma/builtin-objects/ecma-builtin-global.c
zherczeg: Yes. This is just debug print, we should not overdo it.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -415,7 +415,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
         }
 
         ecma_char_t ch = lit_utf8_read_next (&str_curr_p);
-        JERRY_TRACE_MSG ("Period matching '.' to %d: ", (uint32_t) ch);
+        JERRY_TRACE_MSG ("Period matching '.' to %u: ", ch);
```
zherczeg: I think we should add (unsigned int) here.

zherczeg:
LGTM after the last type cast added.

yichoi:
ping @zherczeg @LaszloLango @akosthekiss

zherczeg:
Still LGTM

akosthekiss:
LGTM

PR #1371, opened 2016-09-21
Switch cppcheck to C99 mode

akosthekiss:
From the beginning, we have been configuring cppcheck to check its
input as C++ source. However, the transition to C99 happened a
while ago. This patch switches cppcheck into C99 mode.

Some related changes:
* Progress reporting of cppcheck just clutters the output and makes
  warnings hard to discover. Thus, this patch puts cppcheck into a
  quieter mode where it prints anything only if a non-suppressed
  warning is found.
* The default warning format of cppcheck is a bit different from
  usual compiler error/warning format. This patch configures
  cppcheck to use a more familiar warning template.

### DIFF
--- a/tools/check-cppcheck.sh
+++ b/tools/check-cppcheck.sh
@@ -35,9 +35,11 @@ do
 done
 
 cppcheck -j$CPPCHECK_JOBS --force \
- --language=c++ --std=c++11 \
+ --language=c --std=c99 \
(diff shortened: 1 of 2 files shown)


akosthekiss:
With this patch, all the remaining warning suppressions are jerry-libm related, i.e., jerry-core is cppcheck 1.61-clean.

zherczeg:
LGTM

PR #1372, opened 2016-09-21
Echo executed test commands in run-tests.py

akosthekiss:
run-tests.py used to echo the build commands it executed, but did
not behave the same way when it executed test commands. To make its
behavior more traceable, this patch adds echo for test commands as
well.


zherczeg:
LGTM

PR #1373, opened 2016-09-21
Fix compiler warnings in unit tests

akosthekiss:

### DIFF
--- a/tests/unit/test-libm.c
+++ b/tests/unit/test-libm.c
@@ -33,13 +33,25 @@ check_int (const char *expr, int computed, int expected)
   passed &= result;
 } /* check_int */
 
+typedef union
+{
+  double value;
+  uint64_t bits64;
+  uint32_t bits32[2];
+} double_bits_t;
+
 static void
 check_double (const char *expr, double computed, double expected)
 {
-  uint64_t *computed_bits_p = (uint64_t *) &computed;
-  uint64_t *expected_bits_p = (uint64_t *) &expected;
+  double_bits_t computed_bits;
+  double_bits_t expected_bits;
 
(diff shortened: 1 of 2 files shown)


akosthekiss:
Note: not all compilers generate warnings during the compilation of unit tests, but OSX clang does. Some examples:
```
tests/unit/test-libm.c:42:49: warning: format specifies type 'unsigned long' but the argument has type 'uint64_t' (aka 'unsigned long long') [-Wformat]
tests/unit/test-api.c:74:55: warning: format specifies type 'void *' but the argument has type 'const jerry_value_t *' (aka 'const unsigned int *') [-Wformat-pedantic]
```

tests/unit/test-libm.c
```diff
@@ -39,7 +39,14 @@ check_double (const char *expr, double computed, double expected)
   uint64_t *computed_bits_p = (uint64_t *) &computed;
   uint64_t *expected_bits_p = (uint64_t *) &expected;
 
-  printf ("%s = 0x%lx [expected=0x%lx] ", expr, *computed_bits_p, *expected_bits_p);
+  uint32_t computed_hi = (uint32_t) (*computed_bits_p >> 32);
+  uint32_t computed_lo = (uint32_t) *computed_bits_p;
+  uint32_t expected_hi = (uint32_t) (*expected_bits_p >> 32);
+  uint32_t expected_lo = (uint32_t) *expected_bits_p;
```
zherczeg: There is a better way to do this with unions. Define a union with a double and an uint32_t [2] array, and do the conversion with it.

akosthekiss: I've rewritten it as suggested. See updated patch. (Note, however, that it's not as big/little-endian-safe. But it's just the printing, so no big deal.)

zherczeg:
LGTM

PR #1374, opened 2016-09-22
Generalize the zephyr build system

daniel-thompson:
I was trying to run JerryScript on a new board and it annoyed me having to modify the makefiles for each and every new board.

With these chances JerryScript for Zephyr can be compiled out-of-the-box on a much wider variety of Zephyr platforms.

There are 21 platforms in $ZEPHYR_BASE/boards. Without these patches 4 build, with the patches it rises to 13 (and there are no regressions).

### DIFF
--- a/targets/zephyr/Makefile.zephyr
+++ b/targets/zephyr/Makefile.zephyr
@@ -47,6 +47,11 @@ endif
 
 INTERM     = build/$(BOARD)/obj-$(BOARD)
 OUTPUT     = build/$(BOARD)
+DOTCONFIG  = $(OUTPUT)/zephyr/.config
+
+include $(DOTCONFIG)
+override ARCH = $(subst ",,$(CONFIG_ARCH))
+# (this comment is to close the mismatched quote) "
 
 -include $(ZEPHYR_BASE)/boards/$(BOARD_NAME)/Makefile.board
 -include $(ZEPHYR_BASE)/scripts/Makefile.toolchain.$(ZEPHYR_GCC_VARIANT)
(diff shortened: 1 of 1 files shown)


targets/zephyr/Makefile.zephyr
```diff
@@ -111,6 +131,9 @@ BUILD_CONFIG = O="$(OUTPUT)/zephyr" V=$(V) USER_LIBS="$(LIBS)" USER_LIB_INCLUDE_
 .PHONY: all
 all: jerry zephyr
 
+$(DOTCONFIG):
```
akosthekiss: Shouldn't this recipe be called `initconfig`, just like the other similar recipes below (e.g., `showconfig`, `mrproper`, `flash`, etc.)? The file itself (i.e., `$(DOTCONFIG)`) is not a dependency anywhere anyway, as far as I can telll.

targets/zephyr/Makefile.zephyr
```diff
@@ -47,6 +47,11 @@ endif
 
 INTERM     = build/$(BOARD)/obj-$(BOARD)
 OUTPUT     = build/$(BOARD)
+DOTCONFIG  = $(OUTPUT)/zephyr/.config
+
+include $(DOTCONFIG)
+override ARCH = $(subst ",,$(CONFIG_ARCH))
```
akosthekiss: Why is this an `override`? Can `ARCH` come from the command line?

targets/zephyr/Makefile.zephyr
```diff
@@ -47,6 +47,11 @@ endif
 
 INTERM     = build/$(BOARD)/obj-$(BOARD)
 OUTPUT     = build/$(BOARD)
+DOTCONFIG  = $(OUTPUT)/zephyr/.config
+
+include $(DOTCONFIG)
```
akosthekiss: If I get it right, then `.config` is included unconditionally, and it seems that `ARCH` comes from that file, and then `CPU` is defined based on `ARCH`. Thus, if the file does not exist yet, then almost none of the "build" recipes can be invoked since the evaluation of `CPU` would trigger an error. So, `make -f Makefile.zephyr .config` (or `make -f Makefile.zephyr initconfig`, as suggested in this review) seems to be a mandatory first step in the build process. If so, then the readme should be updated as well.

daniel-thompson: These patches do not introduce any change to the build process (from user's point of view).

An include file is an implicit build target and, because there is a rule that tells make how to generate $(DOTCONFIG), it can be derived automatically.

targets/zephyr/Makefile.zephyr
daniel-thompson: To be honest I doubt the override is actually needed. However the equivalent code in the zephyr Makefile uses override so I decided it was better to act in the same manner as zephyr normally would.

targets/zephyr/Makefile.zephyr
daniel-thompson: As mentioned about $(DOTCONFIG) becomes a dependancy because we include it. If we renamed this build target then the build would break (it would fail when it tried to include the .config file).

targets/zephyr/Makefile.zephyr
akosthekiss: I admit, I was not aware of the implicit build target implications of `include`. It's OK then as it is.

akosthekiss:
LGTM (pending on squash and original port contributor feedback)

sergioamr:
LGTM

daniel-thompson:
I'm just looking at the squash request that came up during review.

Are you absolutely sure you want the first patch (a bug fix that prevent -Werrors from breaking the build) with the rest of the build system enhancements?

akosthekiss:
@daniel-thompson If you think that the two patches are better separated then please split one of them out into a new pull request. As of late, we are using the "big green magic button" of github to land pull requests, which is doing squash-and-merge (that means one PR -> one patch in master).

akosthekiss:
@daniel-thompson do you plan to split out one of the patches into a separate PR?

daniel-thompson:
@akosthekiss: Thanks for the poke! Sorry!

PR #1376, opened 2016-09-23
Add IRC notification to Travis CI configuration

akosthekiss:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,3 +26,20 @@ before_install:
 install:
 
 script: "python tools/run-tests.py $OPTS"
+
+# The channel name "chat.freenode.net#jerryscript"
+# is encrypted against Samsung/jerryscript
+# to prevent IRC spam of forks
+#
+# travis encrypt -r "Samsung/jerryscript" "chat.freenode.net#jerryscript"
+notifications:
+  irc:
+    channels:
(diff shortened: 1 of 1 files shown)


akosthekiss:
With this patch, everyone can get notified on irc if a patch lands in origin. Not necessarily a "must have" feature, but might help following the progress of the master branch for all project contributors.

While experimenting with the Travis feature, I already received messages on my test channel like:
```
travis-ci has joined (~travis-ci@ec2-50-16-37-223.compute-1.amazonaws.com)
travis-ci: jerryscript (travis-notify-irc@ea7491e): Akos Kiss - Add IRC notification to Travis CI configuration [passed]
travis-ci: Commit: https://github.com/akosthekiss/jerryscript/compare/52c39d3b921d...ea7491ec7792
travis-ci: Build: https://travis-ci.org/akosthekiss/jerryscript/builds/162278619
travis-ci has left ()
```

Note: Travis has issues with notifications and forks: .tramis.yml exists in the forks as well, and forks can also have Travis enabled, thus we could get a flood of messages to the irc channel about any and all experimental development results. However, if the channel name is encrypted with the original repo's public key (even though it is not a secret!), then only the Travis running for the original repo will be able to decrypt it and use it for notification. Trick borrowed from https://github.com/mozilla/build-tools/commit/c13a297 (Actually, this is the currently suggested best practice.)

PR #1377, opened 2016-09-24
Revoke exec permission from non-scripts

akosthekiss:
Several non-script files have been added with exec permission to
the repository. This patch revokes the erroneous permissions.


akosthekiss:
For future reference: I've used the following command to find the executable non-script files:
```bash
 find . -type f -perm +111 -not \( -name '*.sh' -or -name '*.py' \) | grep -v .git
```

galpeter:
Could do also in the future an automated travis check for this?

akosthekiss:
@galpeter Not sure that the above command is generic enough. Who knows when does a non-.sh or non-.py script get added to the project.

PR #1378, opened 2016-09-25
Improve the stdin handling of the command line tool

akosthekiss:
This patch improves the command line tool in two ways:
* The tool can be instructed to read and execute a script from
  stdin by invoking it as `jerry -`.
* The tool can be instructed to suppress its prompt when in repl
  mode by invoking it as `jerry --no-prompt`.

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -45,11 +45,19 @@ static const uint8_t *
 read_file (const char *file_name,
            size_t *out_size_p)
 {
-  FILE *file = fopen (file_name, "r");
-  if (file == NULL)
+  FILE *file;
+  if (!strcmp ("-", file_name))
   {
-    jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Error: failed to open file: %s\n", file_name);
-    return NULL;
+    file = stdin;
+  }
+  else
+  {
+    file = fopen (file_name, "r");
+    if (file == NULL)
+    {
+      jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Error: failed to open file: %s\n", file_name);
+      return NULL;
+    }
   }
 
   size_t bytes_read = fread (buffer, 1u, sizeof (buffer), file);
@@ -117,6 +125,7 @@ print_help (char *name)
                       "  --exec-snapshot FILE\n"
                       "  --log-level [0-3]\n"
                       "  --abort-on-fail\n"
+                      "  --no-prompt\n"
                       "\n",
                       name);
 } /* print_help */
@@ -183,6 +192,7 @@ main (int argc,
   const char *save_snapshot_file_name_p = NULL;
 
   bool is_repl_mode = false;
+  bool no_prompt = false;
 
   for (i = 1; i < argc; i++)
   {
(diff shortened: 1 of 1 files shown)


jerry-main/main-unix.c
```diff
@@ -45,11 +45,19 @@ static const uint8_t *
 read_file (const char *file_name,
            size_t *out_size_p)
 {
-  FILE *file = fopen (file_name, "r");
-  if (file == NULL)
+  FILE *file;
+  if (strcmp("-", file_name))
```
galpeter: Other parts the `strcmp` is used wiht `!`. Can't we just have the same variant here?

jerry-main/main-unix.c
```diff
@@ -276,6 +286,14 @@ main (int argc,
     {
       jerry_port_default_set_abort_on_fail (true);
     }
+    else if (!strcmp ("--no-prompt", argv[i]))
+    {
+      no_prompt = true;
+    }
+    else if (!strcmp ("-", argv[i]))
```
galpeter: Do we really need this case? Can't we just use the `else` case for this?

jerry-main/main-unix.c
```diff
@@ -407,7 +425,7 @@ main (int argc,
 
   if (is_repl_mode)
   {
-    const char *prompt = "jerry> ";
+    const char *prompt = !no_prompt ? "jerry> " : "";
```
galpeter: Why it is good to disable the prompt?

akosthekiss:
The/my reason for this patch is the following:

Before this patch:

```bash
$ echo '5+7' | ./build/bin/jerry
jerry> 12
jerry> 
```

Example invocations after this patch:

```bash
$ echo '5+7' | ./build/bin/jerry -
$ echo '5+7' | ./build/bin/jerry --no-prompt
12
$ cat tests/hello.js | ./build/bin/jerry -
Hello JerryScript!
$ cat tests/hello.js | ./build/bin/jerry --no-prompt
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
undefined
Hello JerryScript!
undefined
```

I.e., `-` is useful if one wants to feed a whole script into jerry from stdin, which is responsible for printing its own messages (if any), and does not want to see the result of each statement converted to string and printed.

`--no-prompt` is useful in "eval mode", i.e., if we want jerry to evaluate an expression just like it does in repl mode but don't want to see the "jerry>" prompts interleaved with the results.

jerry-main/main-unix.c
akosthekiss: We can. I've negated the condition.

jerry-main/main-unix.c
akosthekiss: Without this `if`, execution does not reach the final `else` branch because the "error handling if" (just below) catches everything that starts with a dash and has not been recognized hitherto.

galpeter: That is true but we could also simply add a length check to the next `else if` imho. However I'm not really sure which one is better.

zherczeg:
I would prefer --stsdin than a dash.

akosthekiss:
@zherczeg, I disagree here. The single dash is a very unixish tradition to denote stdin as an input. AFAIK, a lot of standard tools support it (tar, cat, diff, file, etc.). I would rather go with tradition instead of a newly invented long option name.

zherczeg:
LGTM but there is one confusing thing. It seems we can have multiple dashes and can be mixed with file names. For me a dash should be an exclusive input option.

akosthekiss:
@zherczeg Traditionally, you can open stdin just as many times as any other file, a mix them as well (give `cat - hello.js -` a try).

Thanks for the reviews.

galpeter:
lgtm

PR #1379, opened 2016-09-29
Update jerry-libc unreachable asserts

akosthekiss:
Newer compilers (especially clang) warn (and fail) on
`assert (!"message");` constructs typically used to assert on
unreachable code paths (multiple occurrences in jerry-libc). A more
up-to-date approach is to use `assert (false && "message");`. This
patch applies the pattern to all relevant asserts in jerry-libc.

### DIFF
--- a/jerry-libc/jerry-libc-printf.c
+++ b/jerry-libc/jerry-libc-printf.c
@@ -253,7 +253,7 @@ libc_printf_write_d_i (FILE *stream, /**< stream pointer */
 
     case LIBC_PRINTF_ARG_LENGTH_TYPE_HIGHL:
     {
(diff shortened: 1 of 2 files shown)


akosthekiss:
JerryScript builds fine with clang-3.5 on x86-64/Ubuntu 14.04 with `CC=clang-3.5 tools/build.py --lto=off`. However, with `CC=clang-3.8` errors appear, e.g.:
```
jerryscript/jerry-libc/target/posix/jerry-libc-target.c:143:16: error: implicit conversion turns string literal into bool: 'char [17]' to 'bool'
      [-Werror,-Wstring-conversion]
      assert (!"unsupported mode");
```
This patch fixes these issues.

PR #1380, opened 2016-09-29
Simplify the way libgcc is linked

akosthekiss:


akosthekiss:
Interestingly enough, I've measured cca. 300 B size reduction on arm cross binaries, which I cannot completely explain.

zherczeg:
LGTM

PR #1385, opened 2016-09-30
Drop `-gdwarf-4` from compiler flags

akosthekiss:
`-g` is enough (most modern compilers use DWARF4 as default, at
least, e.g., GCC has switched from DWARF2 in 4.8), more future
proof (if a compiler moves to a newer version, it wont get forced
to a fixed older version), and less restrictive (some compilers
handle the "standard" `-g` only but don't understand its variants).


dbatyai:
LGTM

PR #1391, opened 2016-10-10
Added target to compile JerryScript beside Particle Device Firmware o…

robertsipka:
…n Photon board.

### DIFF
--- a/targets/particle/Makefile.particle
+++ b/targets/particle/Makefile.particle
@@ -1,0 +1,79 @@
+# Copyright 2016 Samsung Electronics Co., Ltd.
+# Copyright 2016 University of Szeged
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+PARTICLE_FIRMWARE := $(CURDIR)/../particle/firmware
+
+JERRYDIR ?= $(CURDIR)
+JERRYHEAP  ?= 16
+
+BUILD_DIR  ?= $(JERRYDIR)/build/particle
+
+EXT_CFLAGS += -mlittle-endian -mthumb -mcpu=cortex-m4
+EXT_CFLAGS += -Wno-error=format=
+
+.PHONY: jerrycore jerry-main flash clean
+
+PARTICLE_BUILD_CONFIG = \
+	 INCLUDE_DIRS=$(JERRYDIR)/jerry-core \
+	 LIBS=jerry-core \
+	 PLATFORM=photon \
+	 LIB_DIRS=$(BUILD_DIR)/lib \
+	 APPDIR=$(JERRYDIR)/targets/particle/source \
+	 TARGET_FILE=jerry_main \
+	 TARGET_DIR=$(BUILD_DIR) \
+	 LDFLAGS=--specs=nano.specs
+
+all: jerrycore jerry-main
+
+jerrycore:
+	mkdir -p $(BUILD_DIR)
+	cmake -B$(BUILD_DIR) -H./ \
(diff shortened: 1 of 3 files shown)


zherczeg:
LGTM

targets/particle/README.md
```diff
@@ -0,0 +1,91 @@
+### About
+
+This folder contains files to run JerryScript beside Particle Device Firmware on Photon board.
+It  runs a mini example, blinking an LED which is the "Hello World" example of the microcontroller universe.
+
+### How to build
+
+#### 1. Preface / Directory structure
+
+Assume `harmony` as the path to the projects to build.
```
LaszloLango: "harmony"?

targets/particle/README.md
```diff
@@ -0,0 +1,91 @@
...
+```
+jerry_main.bin
+```
+
+That’s the binary what we’ll be flashing with dfu-util.
+
+
+#### 4. Flashing
+
+Make sure you put your Photon in DFU mode.
```
LaszloLango: It would be good to add a short description how to put the device into DFU mode.

targets/particle/Makefile.particle
```diff
@@ -0,0 +1,78 @@
...
+all: jerrycore jerry-main
+
+jerrycore:
+	mkdir -p $(BUILD_DIR)
+	cmake -B$(BUILD_DIR) -H./ \
+	 -DJERRY_LIBC=OFF \
+	 -DJERRY_LIBM=OFF \
+	 -DJERRY_CMDLINE=OFF \
+	 -DUNITTESTS=OFF \
+	 -DENABLE_LTO=ON \
```
akosthekiss: Is it intentional to have lto and all-in-one build at the same time?

targets/particle/source/main.cpp
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Loop code for particle firmware
+ */
+void
+loop ()
+{
+  init_jerry ();
+
+  // flash the D7 LED on and off
```
LaszloLango: `/* Turn on and off the D7 LED */`

targets/particle/Makefile.particle
```diff
@@ -0,0 +1,78 @@
...
+	 TARGET_DIR=$(BUILD_DIR) \
+	 LDFLAGS=--specs=nano.specs
+
+all: jerrycore jerry-main
+
+jerrycore:
+	mkdir -p $(BUILD_DIR)
+	cmake -B$(BUILD_DIR) -H./ \
+	 -DJERRY_LIBC=OFF \
+	 -DJERRY_LIBM=OFF \
```
akosthekiss: As far as I can remember, some (one or two) math routines are used even in case of the minimal profile. However, I cannot find any `-lm` switches in the build. What do I miss?

targets/particle/source/main.cpp
```diff
@@ -0,0 +1,149 @@
...
+  Serial.println ("Beginning Listening mode test!");
+} /* setup */
+
+/**
+ * Loop code for particle firmware
+ */
+void
+loop ()
+{
+  init_jerry ();
```
akosthekiss: Is jerry initialized and cleaned in every loop iteration? Does particle/photon application model have no concept of initialization and cleanup, so that the loop could consist of test_jerry only?

targets/particle/README.md
robertsipka: I think its an accidentally brought folder name from iotjs project. I renamed it to 'root'.

targets/particle/README.md
robertsipka: It is described in the 'Update/Install the Particle Firmware' section.

targets/particle/Makefile.particle
robertsipka: The building of the jerry_main target invokes ARM GCC C++ Linker. The GCC implementation achieved this by automatically linking libm in a C++ build.
Although it really doesn't links to jerrycore itself, but in this case it will not be used without the jerry_main app.

targets/particle/Makefile.particle
robertsipka: No, its not intentional, thanks to notice it, I only want to enable the LTO.

targets/particle/source/main.cpp
robertsipka: Thanks, I've changed this comment.

targets/particle/README.md
LaszloLango: My bad, I missed that part :)

Issue #1387, opened 2016-10-05
Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed in jmem_heap_finalize


renatahodovan:
Version: 9915307
OS: Linux-4.4.0-38-generic-x86_64-with-debian-stretch-sid

Load the attached test with debug jerry:

```js
'use strict'
function undefined () {}
```

Output:

```
ICE: Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed at JERRY_PATH/jerry-core/jmem/jmem-heap.c(jmem_heap_finalize):177.
Error: ERR_FAILED_INTERNAL_ASSERTION
Aborted
```

Backtrace:

```
#0  0x00000000005c57c8 in raise ()
#1  0x00000000005c597a in abort ()
#2  0x000000000057d175 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at JERRY_PATH/targets/default/jerry-port-default-fatal.c:54
#3  jerry_fatal (code=code@entry=ERR_FAILED_INTERNAL_ASSERTION) at JERRY_PATH/jerry-core/jrt/jrt-fatals.c:62
#4  0x000000000057d350 in jerry_assert_fail (assertion=<optimized out>, file=<optimized out>, function=<optimized out>, line=<optimized out>) at JERRY_PATH/jerry-core/jrt/jrt-fatals.c:86
#5  0x0000000000406255 in jmem_heap_finalize () at JERRY_PATH/jerry-core/jmem/jmem-heap.c:177
#6  jmem_finalize () at JERRY_PATH/jerry-core/jmem/jmem-allocator.c:62
#7  jerry_cleanup () at JERRY_PATH/jerry-core/jerry.c:191
#8  main (argc=<optimized out>, argv=<optimized out>) at JERRY_PATH/jerry-main/main-unix.c:509
```

robertsipka:
#1397 is landed.

PR #1397, opened 2016-10-18
Fix for Issue #1387.

robertsipka:
Check the return value of 'ecma_op_put_value_lex_env_base' function.

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -625,11 +625,11 @@ vm_init_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
 
             ecma_object_t *ref_base_lex_env_p = ecma_op_resolve_reference_base (frame_ctx_p->lex_env_p, name_p);
 
-            /* TODO: check the return value */
-            ecma_op_put_value_lex_env_base (ref_base_lex_env_p,
-                                            name_p,
-                                            is_strict,
-                                            lit_value);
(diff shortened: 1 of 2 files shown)


LaszloLango:
Please add the testcase too.

robertsipka:
Sorry, I've added the test case too.

jerry-core/vm/vm.c
```diff
@@ -625,11 +625,11 @@ vm_init_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
-            /* TODO: check the return value */
-            ecma_op_put_value_lex_env_base (ref_base_lex_env_p,
-                                            name_p,
-                                            is_strict,
-                                            lit_value);
+            ecma_value_t put_value_result = ecma_op_put_value_lex_env_base (ref_base_lex_env_p,
+                                                                            name_p,
+                                                                            is_strict,
+                                                                            lit_value);
+            ecma_free_value (put_value_result);
```
zherczeg: Hm, shouldn't this be always successful?

robertsipka: As I see, based on the standards, this function will call the 'ecma_op_set_mutable_binding' method (in this case base must be a reference whose base is an environment record), and because this will be an attempt to change the value of an immutable binding, and the strict param is also true, the binding cannot be set and throw a TypeError exception.

zherczeg: Ok.

zherczeg:
LGTM

robertsipka:
Thanks, I've rebased it with master.

PR #1399, opened 2016-10-19
Remove several internal property types for primitive objects.

zherczeg:
Slight performance improvement: +0.741% (these code paths are not coverd by SunSpider)

However binary size drops 152652->153004 but I don't really understand it since I simplified and deleted code.


zherczeg:
date-format-xparb.js: 5.323%
string-base64.js: 4.063%

Some tests who depends on native object types heavily has measurable speedup.

PR #1403, opened 2016-10-20
Print source code with a given context when a parse error happens

LaszloLango:
Example source:
```javascript
  var s = "hello";
  if (s)
    a[,] = s;
```
Output:
```
  var s = "hello";
  if (s)
    a[,] = s;
  ~~~~^
  Script Error: SyntaxError: Primary expression expected. [line: 3, column: 5]
```

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -130,8 +130,81 @@ print_help (char *name)
                       name);
 } /* print_help */
 
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+
+/**
+ * Check whether an error is a SyntaxError or not
+ *
+ * @return true  - if param is SyntaxError
+ *         false - otherwise
+ */
+static bool
+jerry_value_is_syntax_error (jerry_value_t error_value) /**< error value */
+{
+  if (!jerry_value_is_object (error_value))
+  {
+    return false;
+  }
+
+  jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *)"name");
+  jerry_value_t error_name = jerry_get_property (error_value, prop_name);
+  jerry_release_value (prop_name);
+
+  if (jerry_value_has_error_flag (error_name)
+      || !jerry_value_is_string (error_name))
+  {
+    return false;
+  }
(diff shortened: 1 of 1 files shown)


jerry-main/main-unix.c
```diff
@@ -130,8 +130,77 @@ print_help (char *name)
...
+  jerry_release_value (prop_name);
+
+  if (jerry_value_has_error_flag (error_name)
+      || !jerry_value_is_string (error_name))
+  {
+    return false;
+  }
+
+  jerry_size_t err_str_size = jerry_get_string_size (error_name);
+  jerry_char_t err_str_buf[32];
```
zherczeg: You get the size, but the buf is only 32 bytes long.

jerry-main/main-unix.c
```diff
@@ -150,8 +219,101 @@ print_unhandled_exception (jerry_value_t error_value)
...
+            j++;
+            i++;
+          }
+          num_str[j] = '\0';
+
+          err_line = str_to_uint (num_str);
+
+          if (strncmp ((char *) (err_str_buf + i), ", column: ", 10))
+          {
+            break; /* wrong position info format */
```
LaszloLango: IMHO, break is better here, because on a return there is no error message printing, so if you don't check the exit value, then you don't even notice the program failure. Check the end of the function.

jerry-main/main-unix.c
```diff
@@ -150,8 +219,101 @@ print_unhandled_exception (jerry_value_t error_value)
...
+          num_str[j] = '\0';
+
+          err_col = str_to_uint (num_str);
+          break;
+        }
+      } /* for */
+
+      if (err_line != 0 && err_col != 0)
+      {
+        uint32_t curr_line = 1;
```
LaszloLango: Not sure it will be much simpler, because we still have to check a few special case when we cannot print 2*context_size.

zherczeg:
Hm ok. This is a utility function anyway, might be not worth to overdo this.

LGTM

dbatyai:
LGTM

PR #1407, opened 2016-10-25
build: Adopt outputexports

daniel-thompson:
This patch simplifies the integration with the zephyr build system
whilst at the same time enriching the set of build targets to allow
zephyr to be configured using its kbuild features.

It works by exploiting "make outputexports", a feature of the zephyr
build system that makes the zephyr compiler configuration available to
other build systems in an easily accessible manner.

Whilst looking at the build we also correct the implementation of clean
so that it no longer destroys any custom zephyr configuration. Like any
other zephyr application one must use "make pristine" to remove the
config too.

JerryScript-DCO-1.0-Signed-off-by: Daniel Thompson daniel.thompson@linaro.org

### DIFF
--- a/targets/zephyr/Makefile.zephyr
+++ b/targets/zephyr/Makefile.zephyr
@@ -19,11 +19,7 @@ ifeq ($(.DEFAULT_GOAL),)
 endif
 
 BOARD ?= arduino_101
-BOARD_NAME ?= arduino_101
-
-ifeq ($(BOARD),qemu_x86)
-BOARD_NAME = qemu_x86
-endif
+export BOARD
 
 TARGET_ZEPHYR ?= ./targets/zephyr
 TARGET_ZEPHYR_SRC_DIR = $(TARGET_ZEPHYR)/src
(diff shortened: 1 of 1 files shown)


daniel-thompson:
@pfalcon: I think you were interested in this...

pfalcon:
@daniel-thompson : Thanks for submitting it!

Put simple, @daniel-thompson's patch allows to build JrS for any board supported by Zephyr, present or future.

I built and booted JrS on the following Zephyr boards:

~~~
qemu_x86
qemu_cortex_m3
frdm_k64f
~~~

I tested building for the following boards (don't have them to boot):

~~~
96b_nitrogen
altera_max10
arduino_101
arduino_due
~~~

Few other boards are tried didn't build because they don't have full support in Zephyr, few other due to lack of RAM.

So, definitely +1 from me.

@sergioamr , @poussa : FYI.

daniel-thompson:
Of the 21 boards supported by mainline zephyr:

13 build correctly (only 12 build correctly before this patch)
4 build but fail to link because there is insufficient RAM and/or FLASH
3 fail to build because JerryScript tickles a compiler bug in the ARC toolchain
1 fails because CONFIG_UART_INTERRUPT_DRIVEN is set in the .config (I suspect this is a zephyr bug, the config option is not requested by prj.conf)

LaszloLango:
LGTM

zherczeg:
LGTM

PR #1408, opened 2016-10-26
Improve Js2C converter

LaszloLango:
* Rename 'jerry_targetjs.h' to 'jerry-targetjs.h',
   because we use dashes in file names instead of underscores.
 * Made destination and js souce directory configurable.
 * Updated esp8266 target to the recent changes.
 * Updated mbed and mbedos5 target to the recent changes.

### DIFF
--- a/targets/tools/js2c.py
+++ b/targets/tools/js2c.py
@@ -14,14 +14,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-#  This file converts ./js/*.js to a C-array in ./source/jerry_targetjs.h file
+#  This file converts ./js/*.js to a C-array in ./source/jerry-targetjs.h file
 
-import sys
+import argparse
 import glob
 import os
 import re
-
-import argparse
+import sys
 
 special_chars = re.compile(r'[-\\?\'".]')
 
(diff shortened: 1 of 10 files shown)


targets/tools/js2c.py
```diff
@@ -122,7 +127,7 @@ def exportOneFile(path, name):
 def exportOneName(name):
     writeLine(fout, '{ ' + name + '_n, ' + name + '_s, ' + name + '_l }, \\', 1)
 
-files = glob.glob(SRC_PATH + '*.js')
+files = glob.glob(SRC_PATH + '/*.js')
```
galpeter: This could also be joined (os.path.join) together.

targets/tools/js2c.py
```diff
@@ -78,21 +79,25 @@ def removeWhitespaces(code):
 
 '''
 
-OUT_PATH = './source/'
-SRC_PATH = './js/'
+OUT_PATH = './source'
+SRC_PATH = './js'
```
galpeter: I see no point having this, just move them into the add_argument call.

targets/tools/js2c.py
```diff
@@ -14,14 +14,15 @@
...
 
-import sys
+import argparse
 import glob
 import os
 import re
+import sys
 
-import argparse
+from os import path
```
galpeter: We could simply use: `os.path.join` then there is no need for this import :)

LaszloLango:
@galpeter, thanks for the review. I've updated the PR.

galpeter:
LGTM

zherczeg:
LGTM

PR #1411, opened 2016-10-26
zephyr: Revamp jerry-port.c, add more functions.

pfalcon:
The file was named jerry-entry.c and not even built. Rename for clarity,
and add jerry_port_fatal(), jerry_port_get_current_time(),
jerry_port_get_time_zone() functions. User-visible result is that
if Date builtin is enabled (e.g. if building "full" config, its
Date.now() method can be used to measure relative time difference,
e.g. for benchmarking).

### DIFF
--- a/targets/zephyr/src/jerry-port.c
+++ b/targets/zephyr/src/jerry-port.c
@@ -1,4 +1,5 @@
 /* Copyright 2016 Intel Corporation
+ * Copyright 2016 Linaro Limited
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -14,6 +15,8 @@
  */
 
 #include <stdarg.h>
+
+#include <zephyr.h>
 
 #include "jerry-port.h"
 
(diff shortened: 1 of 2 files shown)


pfalcon:
@sergioamr , @poussa : FYI.

targets/zephyr/src/jerry-port.c
```diff
@@ -46,3 +49,40 @@ jerry_port_log (jerry_log_level_t level, /**< log level */
...
+jerry_port_get_current_time ()
+{
+  int64_t us = sys_tick_get() * sys_clock_us_per_tick;
+  return (double) us / 1000;
+} /* jerry_port_get_current_time */
+
+/**
+ * Dummy function to get the time zone.
+ *
+ * @return false
```
galpeter: In the code it returns true.

sergioamr:
lgtm

poussa:
lgtm

targets/zephyr/src/jerry-port.c
zherczeg: LGTM after this fix

pfalcon: Fixed.

pfalcon:
Fixed comment and rebased on the master.

PR #1413, opened 2016-10-27
target: zephyr: Fix build when USE_CCACHE=1

daniel-thompson:
Currently if USE_CCACHE=1 is set to one than the Zephyr build system
will include ccache in the CC variable. ccache cannot be passed to
cmake using EXTERNAL_CMAKE_C_COMPILER (even if we fix the quoting)
because it fails a PATH reachability test within cmake. For that
reasons we solve the build failure simply by removing ccahce from CC
before we pass it into cmake.

The problem described about is a regression due to commit ac1bf19c90d6
("build: Adopt outputexports") and the fix contained in this patch
restores the original behaviour (zephyr exploits ccache, jerryscript
library build does not).

JerryScript-DCO-1.0-Signed-off-by: Daniel Thompson daniel.thompson@linaro.org


zherczeg:
LGTM

LaszloLango:
LGTM

PR #1416, opened 2016-10-30
Add how to install the modules

polaroi8d:

### DIFF
--- a/targets/mbedos5/README.md
+++ b/targets/mbedos5/README.md
@@ -50,6 +50,26 @@ jshint is used to statically check your JavaScript code, as part of the build pr
 This ensures that pins you are using in your code are available on your chosen target
 platform.
 
+### Python modules
+
+There are multiple Python module dependencies which can be installed with the
+following command invoked from your jerryscript/targets/mbedos5/tools directory:
+
+```bash
+pip install -r requirements.txt
+```
+
+If you have no pip installed but your are on an apt-managed system (e.g.,
+Ubuntu), issue the following command first:
+
+```bash
+apt-get -y install python-pip
+```
+
+(Otherwise, visit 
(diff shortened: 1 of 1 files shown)


targets/mbedos5/README.md
```diff
@@ -50,6 +50,15 @@ jshint is used to statically check your JavaScript code, as part of the build pr
 This ensures that pins you are using in your code are available on your chosen target
 platform.
 
+### Python modules
+
+There are Python module dependencies which can be installed with the following command: 
+	`pip install -r requirements.txt`
+Use this command in your .../jerryscript/targets/mbedos5/tools directory.
```
akosthekiss: If you want this command to stand out on its own then this is not the proper markdown syntax. Right now, the pip command is inlined. Either add extra empty lines before and after the command (and remove the backticks from around the command), or add triple-backticks around the command (but don't indent it then) (as seen in the Quick Start section).

Moreover, I think you have used TAB for indenting. Please, don't do that. TABs are evil. Use spaces (if needed).

targets/mbedos5/README.md
```diff
@@ -50,6 +50,15 @@ jshint is used to statically check your JavaScript code, as part of the build pr
...
 platform.
 
+### Python modules
+
+There are Python module dependencies which can be installed with the following command: 
+	`pip install -r requirements.txt`
+Use this command in your .../jerryscript/targets/mbedos5/tools directory. 
+
+If there is no pip installed, use the following command:
+	`apt-get -y install python-pip`
```
akosthekiss: Same here.

Additionally, this only works if the user is on an apt-based system like Ubuntu. A generic description can be found on pip's homepage: https://pip.pypa.io/en/stable/installing/ Might be worth including it in the description.

polaroi8d:
Thanks for the review @akosthekiss, Im updated the PR.

PR #1417, opened 2016-10-31
Ensure that `Math.random()` really covers `[0,1)` interval

### LINKED ISSUE #1414: Math.random() returns range [0, 0.5) instead of [0, 1.0]
This was reported on Pebble forums - https://forums.pebble.com/t/rocky-js-math-random-bug/25359

It looks like the problem was introduced with https://github.com/Samsung/jerryscript/commit/6d6c913bdc3bb08e2937691f75449cf248b1a790

The range of the rand() function in jerry-libc is only the positive int32_t values.  However, the code in ecma_builtin_math_object_random assumes the range of rand() is the full uint32_t space, since RAND_MAX is > 0x10000.

akosthekiss:
Legacy implementation made incorrect assumptions on how many bits
of useful information is returned by libc's `rand()`. If `RAND_MAX`
had more than 16 useful bits, it assumed that `rand()` could return
32 useful random bits. However, e.g., jerry-libc's `RAND_MAX` is
`0x7fffffff`, which denotes 31 useful bits only. The consequence
was that `Math.random()` covered only the lower half of the
standard-mandated `[0,1)` interval.

This path fixes the error and always uses the exact value of
`RAND_MAX` to compute the random value, which will thus fully cover
`[0,1)`.

Fixes #1414

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-math.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-math.c
@@ -507,28 +507,11 @@ static ecma_value_t
 ecma_builtin_math_object_random (ecma_value_t this_arg) /**< 'this' argument */
 {
   JERRY_UNUSED (this_arg);
-  uint32_t rnd = 1;
-  uint32_t reps_count;
-#if RAND_MAX < 0x100
-  reps_count = 4;
-#elif RAND_MAX < 0x10000
-  reps_count = 2;
-#else /* RAND_MAX >= 0x10000 */
-  reps_count = 1;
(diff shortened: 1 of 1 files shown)


akosthekiss:
With this PR, I get the following results:
```
jerry> for (i = 0; i < 10; i++) print(Math.random())
0.24784175818786028
0.4764318028464914
0.38931416952982545
0.20325033459812403
0.028375181369483467
0.9016734985634685
0.4264974128454924
0.142021032050252
0.947486779652536
0.4103130353614688
```
Seems OK to me

zherczeg:
LGTM

PR #1430, opened 2016-11-14
Add API functions to create string from a valid UTF-8 string.

robertsipka:
I think this patch needs to be followed by another, which contains symbol (functions names, types, ...) renames.  Because CESU-8 is a variant of UTF-8, the names was left utf8 and mentioned the CESU-8 only in comments after #616, but I think this is very confusing right now. For example the 'ecma_new_ecma_string_from_utf8' function check that the string is a cesu-8 encoded string at the first step, so it doesn't accept an utf-8 encoded string which contains 4 bytes long characters, and throw an assert.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -216,6 +216,120 @@ ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 stri
 } /* ecma_new_ecma_string_from_utf8 */
 
 /**
+ * Allocate a new ecma-string and initialize it from the utf8 string argument.
+ * All 4-bytes long unicode sequences are converted into two 3-bytes long sequences.
+ *
+ * @return pointer to ecma-string descriptor
+ */
+ecma_string_t *
+ecma_new_ecma_string_from_utf8_converted_to_cesu8 (const lit_utf8_byte_t *string_p, /**< utf-8 string */
+                                                   lit_utf8_size_t string_size) /**< utf-8 string size */
+{
+  JERRY_ASSERT (string_p != NULL || string_size == 0);
+
+  ecma_string_t *string_desc_p = NULL;
+
+  ecma_length_t string_length = 0;
+  lit_utf8_size_t converted_string_size = 0;
+  lit_utf8_size_t pos = 0;
+
+  /* Calculate the required length and size information of the converted cesu-8 encoded string */
+  while (pos < string_size)
+  {
+    if ((string_p[pos] & LIT_UTF8_1_BYTE_MASK) == LIT_UTF8_1_BYTE_MARKER)
+    {
+      pos++;
+    }
+    else if ((string_p[pos] & LIT_UTF8_2_BYTE_MASK) == LIT_UTF8_2_BYTE_MARKER)
+    {
+      pos += 2;
+    }
+    else if ((string_p[pos] & LIT_UTF8_3_BYTE_MASK) == LIT_UTF8_3_BYTE_MARKER)
+    {
+      pos += 3;
+    }
+    else
+    {
+      JERRY_ASSERT ((string_p[pos] & LIT_UTF8_4_BYTE_MASK) == LIT_UTF8_4_BYTE_MARKER);
+      pos += 4;
+      converted_string_size += 2;
+    }
+
+    string_length++;
+  }
+
+  JERRY_ASSERT (pos == string_size);
+
+  if (converted_string_size == 0)
+  {
+    return ecma_new_ecma_string_from_utf8 (string_p, string_size);
+  }
+  else
+  {
+    converted_string_size += string_size;
+
+    JERRY_ASSERT (lit_is_utf8_string_valid (string_p, string_size));
+
+    lit_utf8_byte_t *data_p;
+
+    if (likely (string_size <= UINT16_MAX))
+    {
+      string_desc_p = jmem_heap_alloc_block (sizeof (ecma_string_t) + converted_string_size);
+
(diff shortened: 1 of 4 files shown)


jerry-core/lit/lit-strings.c
```diff
@@ -797,3 +798,73 @@ bool lit_compare_utf8_strings_relational (const lit_utf8_byte_t *string1_p, /**<
...
+{
+  JERRY_ASSERT (buf_p && buf_size);
+
+  lit_utf8_size_t size = 0;
+  lit_utf8_size_t addition_size = 0;
+
+  while (size < buf_size)
+  {
+    if ((buf_p[size] & LIT_UTF8_4_BYTE_MASK) == LIT_UTF8_4_BYTE_MARKER)
+    {
```
zherczeg: Somehow we should return with this information.

jerry-core/jerry.c
```diff
@@ -914,6 +914,56 @@ jerry_create_object (void)
...
+  jerry_assert_api_available ();
+
+  ecma_string_t *ecma_str_p = NULL;
+  lit_utf8_size_t dest_buf_size = 0;
+
+  dest_buf_size = lit_calculate_cesu8_size_from_utf8 ((lit_utf8_byte_t *) str_p, (lit_utf8_size_t) str_size);
+
+  if ((lit_utf8_size_t) str_size != dest_buf_size)
+  {
+    lit_utf8_byte_t dest_buf[dest_buf_size];
```
robertsipka: Indeed, but I think we should use JMEM_DEFINE_LOCAL_ARRAY also and it wouldn't require too much modification, what is your opinion?

jerry-core/jerry.c
```diff
@@ -914,6 +914,56 @@ jerry_create_object (void)
...
+                                  jerry_size_t str_size) /**< string size */
+{
+  jerry_assert_api_available ();
+
+  ecma_string_t *ecma_str_p = NULL;
+  lit_utf8_size_t dest_buf_size = 0;
+
+  dest_buf_size = lit_calculate_cesu8_size_from_utf8 ((lit_utf8_byte_t *) str_p, (lit_utf8_size_t) str_size);
+
+  if ((lit_utf8_size_t) str_size != dest_buf_size)
```
robertsipka: It will be handle all utf8 encoded string.

jerry-core/jerry.c
```diff
@@ -914,6 +914,57 @@ jerry_create_object (void)
...
+  {
+    JMEM_DEFINE_LOCAL_ARRAY (dest_buf, dest_buf_size, lit_utf8_byte_t);
```
zherczeg: I still don't think we should allocate memory here. We could allocate memory for the final string directly. It is a bit more complicated, but I think memory reduction is critcal.

robertsipka:
@zherczeg : Thanks for the advice, I've modified this patch based on your comments.

jerry-core/lit/lit-strings.h
```diff
@@ -135,4 +135,6 @@ ecma_char_t lit_utf8_peek_prev (const lit_utf8_byte_t *);
 void lit_utf8_incr (const lit_utf8_byte_t **);
 void lit_utf8_decr (const lit_utf8_byte_t **);
 
+bool lit_utf8_contains_four_bytes_unicode (const lit_utf8_byte_t *, lit_utf8_size_t);
```
zherczeg: I liked the original approach that the size is also computed by this function, and it can be passed to ecma_new_ecma_string_from_utf8_converted_to_cesu8 to avoid double scan.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -216,6 +216,111 @@ ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 stri
 } /* ecma_new_ecma_string_from_utf8 */
 
 /**
+ * Allocate new ecma-string and fill it with characters from the utf8 string
+ * It converts all 4-bytes long unicode sequence to two 3-bytes long sequence
```
zherczeg: Allocate a new ecma-string and initialize it from the utf8 string argument.
All 4-bytes long unicode sequences are converted into two 3-bytes long sequences.

zherczeg:
The patch is definitely going in the right direction. Some finishing touch is still needed.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -216,6 +216,120 @@ ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 stri
...
+    {
+      JERRY_ASSERT ((string_p[size] & LIT_UTF8_4_BYTE_MASK) == LIT_UTF8_4_BYTE_MARKER);
+      size += 4;
+      conv_size += 2;
+    }
+
+    str_length++;
+  }
+
+  JERRY_ASSERT (size == string_size);
```
zherczeg: If this is true, there is no need to computing it except for debugging purposes. This would simplify the for loop above.

robertsipka: I've renamed the misleading variable names.

zherczeg:
I think the patch is ok. @LaszloLango do you think we should update the documentation and some api tests in this or a follow-up patch?

robertsipka:
Thanks, I've rebased it with master. @LaszloLango : documentation and some api tests in a follow-up patch is ok for you too?

LaszloLango:
Yes, follow-up patch is fine. Just don't forget to do it. :)

zherczeg:
LGTM
Plase also reference this patch for the appropriate issue.

PR #1436, opened 2016-11-16
target: zephyr: Allow parallel build of Zephyr

galak:
If we are building JerryScript from some other makefile that passes
down a -j option to make we should respect that when we build Zephyr.

JerryScript-DCO-1.0-Signed-off-by: Kumar Gala <kumar.gala@linaro.org>

### DIFF
--- a/targets/zephyr/Makefile.zephyr
+++ b/targets/zephyr/Makefile.zephyr
@@ -106,7 +106,7 @@ zephyr: jerry
 ifdef V
 	@echo "- ZEPHYR -------------------------------------------------------"
 endif
-	make -f $(TARGET_ZEPHYR)/Makefile $(BUILD_CONFIG)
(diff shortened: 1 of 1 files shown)


zherczeg:
Travis reported an error:

Signed-off-by message is incorrect. The following line should be at the end of the b057840 commit's message: 'JerryScript-DCO-1.0-Signed-off-by: Kumar Gala kumar.gala@linaro.org'.

galak:
I'm not sure I understand what's missing. I have a JerryScript-DCO in the commit message as the last thing.

zherczeg:
I think the problem is the <> around the email address.

galak:
Ah, thanks.. was missing that bit.

zherczeg:
LGTM

PR #1439, opened 2016-11-17
Change array length into a virtual property.

zherczeg:
It seems this patch has a slight performance degradation (-0.5%) but it opens up future optimization options so I think it is worth landing it.

### DIFF
--- a/jerry-core/ecma/operations/ecma-array-object.c
+++ b/jerry-core/ecma/operations/ecma-array-object.c
@@ -86,16 +86,16 @@ ecma_op_create_array_object (const ecma_value_t *arguments_list_p, /**< list of 
   }
 
 #ifndef CONFIG_DISABLE_ARRAY_BUILTIN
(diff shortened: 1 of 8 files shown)


jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -230,12 +217,13 @@ ecma_op_array_object_set_length (ecma_object_t *object_p, /**< the array object
     current_len_uint32 = ecma_delete_array_properties (object_p, new_len_uint32, old_len_uint32);
   }
 
-  ecma_value_assign_uint32 (&len_prop_value_p->value, current_len_uint32);
+  ext_object_p->u.array.length = current_len_uint32;
```
LaszloLango: Is this correct? `array.length` is ecma_value_t

zherczeg: After the conversion to virtual value, it is an uint32. When you get array.length it is converted to ecma_value.

zherczeg:
Changed to uint32_t

PR #1442, opened 2016-11-18
Fix fopen call on non-existing files.

### LINKED ISSUE #1437: jerry save-snaphot-for-global hangs
I build Jerry on Linux but when I run:

```
./jerry --save-snapshot-for-global t.js output.snp
```

The program just hangs/loops.  If there are additional diagnostics I can run, please let me know.

LaszloLango:
The main implementation of unix platform hangs, if the engine was
built with jerry-libc. Open system call returns with a negative
error code if cannot open the file, but fopen must return NULL
when error happens.

Fixes #1437.


zherczeg:
LGTM

PR #1444, opened 2016-11-21
Add API documentation for create string from a valid UTF8 string functions.

robertsipka:


tests/unit/test-api.c
```diff
@@ -339,6 +339,13 @@ main (void)
   TEST_ASSERT (sz == 0);
   jerry_release_value (args[0]);
 
+  // Test create_jerry_string_from_utf8 with 4-byte long unicode sequences
+  args[0] = jerry_create_string_from_utf8 ((jerry_char_t *) "𝔡𝔢𝔣𝔤");
```
zherczeg: I would prefer for using \x.. byte sequence instead of real characters (since you don't know the character code of these characters). A four byte long sequence should be part of it as well.

robertsipka:
Thanks, I've updated this patch.

zherczeg:
LGTM

LaszloLango:
LGTM

PR #1445, opened 2016-11-24
Reduce memory consumption of property names.

zherczeg:
Large memory improvements. E.g. bitops-nsieve-bits.js 28%, crypto-sha1.js 26%, crypto-md5.js 23%, ...

Binary size is unchanged (152544 bytes), however unlikely is that with such a big patch.

Performance is decreased because of the increased complexity by 1%. Most tests are not changed much, but bitops-bitwise-and.js is decreased by 9%.

Benchmark | Heap (bytes) | RSS (bytes) | Perf (sec) |
----: | ----: | ----: | ----: | 
3d-cube.js | 16376 -> 16376 : 0.000% | 57344 -> 57344 : 0.000% | 0.869 -> 0.858 : +1.230% | 
3d-raytrace.js | 160872 -> 128496 : +20.125% | 212992 -> 176128 : +17.308% | 1.050 -> 1.042 : +0.775% | 
access-binary-trees.js | 32760 -> 32760 : 0.000% | 53248 -> 53248 : 0.000% | 0.567 -> 0.574 : -1.258% | 
access-fannkuch.js | 5624 -> 5376 : +4.410% | 20480 -> 20480 : 0.000% | 2.181 -> 2.203 : -0.972% | 
access-nbody.js | 6672 -> 6488 : +2.758% | 28672 -> 28672 : 0.000% | 1.069 -> 1.081 : -1.207% | 
bitops-3bit-bits-in-byte.js | 928 -> 920 : +0.862% | 16384 -> 16384 : 0.000% | 0.588 -> 0.588 : +0.084% | 
bitops-bits-in-byte.js | 880 -> 872 : +0.909% | 16384 -> 16384 : 0.000% | 0.871 -> 0.875 : -0.465% | 
bitops-bitwise-and.js | 568 -> 560 : +1.408% | 16384 -> 16384 : 0.000% | 1.053 -> 1.153 : -9.469% | 
bitops-nsieve-bits.js | 138536 -> 98528 : +28.879% | 155648 -> 114688 : +26.316% | 1.418 -> 1.389 : +2.001% | 
controlflow-recursive.js | 912 -> 904 : +0.877% | 81920 -> 81920 : 0.000% | 0.393 -> 0.398 : -1.295% | 
crypto-aes.js | 49184 -> 40944 : +16.753% | 81920 -> 73728 : +10.000% | 0.953 -> 0.962 : -1.019% | 
[...]

PR #1457, opened 2016-11-28
target: mbedos5: Add return value to setTimeout/setInterval and implement clearTimeout/clearInterval

janjongboom:
In the mbedos5 target the setTimeout and setInterval functions are not on spec, as they return 'undefined' instead of an event ID. Also clearTimeout and clearInterval are not implemented, so scheduled events cannot be canceled. This patch changes the behavior of the set* functions, and implements clear* functions.

Partial fix for #1445 

/cc @thegecko @matthewelse


zherczeg:
LGTM

thegecko:
Ack

PR #1460, opened 2016-11-29
Add an API function to get the utf8 length of a cesu8 string.

robertsipka:


jerry-core/lit/lit-strings.c
```diff
@@ -316,6 +316,42 @@ lit_get_utf8_size_of_cesu8_string (const lit_utf8_byte_t *cesu8_buf_p, /**< cesu
...
+    ecma_char_t ch;
+    offset += lit_read_code_unit_from_utf8 (cesu8_buf_p + offset, &ch);
+
+    if (lit_is_code_point_utf16_high_surrogate (ch) && (offset < cesu8_buf_size))
+    {
+      ecma_char_t next_ch;
+      offset += lit_read_code_unit_from_utf8 (cesu8_buf_p + offset, &next_ch);
+
+      if (!lit_is_code_point_utf16_low_surrogate (next_ch))
+      {
```
zherczeg: This is not working in case of a high, high, low surrogate case.

robertsipka: Indeed, I missed this case, thanks.

jerry-core/lit/lit-strings.c
```diff
@@ -316,6 +316,48 @@ lit_get_utf8_size_of_cesu8_string (const lit_utf8_byte_t *cesu8_buf_p, /**< cesu
...
+
+    if (lit_is_code_point_utf16_high_surrogate (ch) && (offset < cesu8_buf_size))
+    {
+      ecma_char_t next_ch;
+      lit_utf8_size_t next_ch_size = lit_read_code_unit_from_utf8 (cesu8_buf_p + offset, &next_ch);
+
+      if (lit_is_code_point_utf16_high_surrogate (next_ch))
+      {
+        continue;
+      }
```
zherczeg: Is there a way to optimize this without decoding the character twice?

Perhaps we could "remember" the previous character, and if we found a low surrogate, we check that the previous character was a high surrogate.

robertsipka: Yes, its a good idea, thanks, I've updated this patch.

zherczeg:
I like this new algorithm.

LGTM

docs/02.API-REFERENCE.md
```diff
@@ -1223,6 +1224,44 @@ jerry_get_string_length (const jerry_value_t value);
...
+returns with utf-8 string length instead of the cesu-8 string length.
+
+**Prototype**
+
+```c
+jerry_length_t
+jerry_get_utf8_string_length (const jerry_value_t value);
+```
+
+- `value` - api value
```
LaszloLango: `value` - input string value

robertsipka:
Thanks, I've updated and rebased this patch with master.

PR #1465, opened 2016-11-30
Add API function to check if the specified feature is enabled.

robertsipka:


jerry-core/jerry.c
```diff
@@ -519,6 +519,103 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
...
+ * Check if the specified feature is enabled.
+ *
+ * @return true  - if the specified feature is enabled,
+ *         false - otherwise.
+ */
+bool jerry_is_feature_enabled (const jerry_feature_t feature)
+{
+  switch (feature)
+  {
+    case JERRY_FEATURE_ERROR_MESSAGES:
```
galpeter: What if we use the following format?
```
switch (feature)
{
#ifdef FEATURE
    case FEATURE: return true;
#endif /* FEATURE */
...
    default: return false;
}
```
It would be less lines and probably more readable without multiple returns in one block. Also maybe we could add the case-es with a macro, just to avoid the typing the feature names multiple times. What do you think?

zherczeg: Apart from the incorrect style the idea is not bad

    switch (feature)
    {
    #ifdef FEATURE1
       case FEATURE1:
    #endif /* FEATURE1 */
    #ifdef FEATURE2
       case FEATURE2:
    #endif /* FEATURE2 */
       {
         return true;
       }
       default:
       {
         return false;
       }
    }

This could work as long as at least one feature is enabled.

Or just using a conditional form:

    return (false
    #ifdef FEATURE1
            || feature == FEATURE1
    #endif /* FEATURE1 */
    #ifdef FEATURE2
            || feature == FEATURE2
    #endif /* FEATURE2 */
            || false);

docs/02.API-REFERENCE.md
```diff
@@ -22,6 +22,21 @@ Possible types of an error:
...
+ - JERRY_FEATURE_ERROR_MESSAGES - error messages
+ - JERRY_FEATURE_VALGRIND - valgrind support
+ - JERRY_FEATURE_VALGRIND_FREYA - valgrind-freya support
+ - JERRY_FEATURE_CPOINTER_32_BIT - 32 bit compressed pointers
+ - JERRY_FEATURE_MEM_STRESS_TEST - mem-stress test
+ - JERRY_FEATURE_MEM_STATS - memory statistics
+ - JERRY_FEATURE_PARSER_DUMP - parser byte-code dumps
+ - JERRY_FEATURE_REGEXP_DUMP - regexp byte-code dumps
+ - JERRY_FEATURE_SNAPSHOT_SAVE - saving snapshot files
+ - JERRY_FEATURE_SNAPSHOT_EXEC - executing snapshot files
```
zherczeg: I don't think JERRY_FEATURE_VALGRIND, JERRY_FEATURE_VALGRIND_FREYA and JERRY_FEATURE_MEM_STRESS_TEST has any meaningful advantage. We can make them public later if ever needed.

jerry-core/jerry.c
```diff
@@ -519,6 +519,56 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
...
+#ifdef REGEXP_DUMP_BYTE_CODE
+    case JERRY_FEATURE_REGEXP_DUMP:
+#endif /* REGEXP_DUMP_BYTE_CODE */
+#ifdef JERRY_ENABLE_SNAPSHOT_SAVE
+    case JERRY_FEATURE_SNAPSHOT_SAVE:
+#endif /* JERRY_ENABLE_SNAPSHOT_SAVE */
+#ifdef JERRY_ENABLE_SNAPSHOT_EXEC
+    case JERRY_FEATURE_SNAPSHOT_EXEC:
+#endif /* JERRY_ENABLE_SNAPSHOT_EXEC */
+    {
```
LaszloLango: The default case will handle it and returns false. I think the code is good. I like it.

LaszloLango: IMHO every case should have their own return statement, so the default case could handle it. Like:

```c
...
#ifdef JERRY_ENABLE_SNAPSHOT_SAVE
     case JERRY_FEATURE_SNAPSHOT_SAVE:
     {
       return true;
     }
#endif /* JERRY_ENABLE_SNAPSHOT_SAVE */
#ifdef JERRY_ENABLE_SNAPSHOT_EXEC
     case JERRY_FEATURE_SNAPSHOT_EXEC:
     {
       return true;
     }
#endif /* JERRY_ENABLE_SNAPSHOT_EXEC */
    default:
     {
       return false;
     }
```

LaszloLango: Other solution could be to add a dummy member to `jerry_feature_t` like

```c
  ...
  JERRY_FEATURE_SNAPSHOT_EXEC /**< Executing snapshot files. */
  JERRY_FEATURE__COUNT /**< Number of features. NOTE: must be at the end of the list. */
} jerry_feature_t;
```

So you can write:

```c
...
#ifdef JERRY_ENABLE_SNAPSHOT_EXEC
     case JERRY_FEATURE_SNAPSHOT_EXEC:
#endif /* JERRY_ENABLE_SNAPSHOT_EXEC */
     case JERRY_FEATURE__COUNT:
     {
       JERRY_ASSERT (feature != JERRY_FEATURE__COUNT); /* not a real feature */
       return true;
     }
    default:
     {
       return false;
     }
```

But I don't like this solution too much.

docs/02.API-REFERENCE.md
robertsipka: Thanks the reviews. I removed the mentioned features from the list, and chose an intermediate solution from your suggestions. Please check it.

jerry-core/jerry.c
```diff
@@ -519,6 +519,41 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
...
+#ifdef REGEXP_DUMP_BYTE_CODE
+          || feature == JERRY_FEATURE_REGEXP_DUMP
+#endif /* REGEXP_DUMP_BYTE_CODE */
+#ifdef JERRY_ENABLE_SNAPSHOT_SAVE
+          || feature == JERRY_FEATURE_SNAPSHOT_SAVE
+#endif /* JERRY_ENABLE_SNAPSHOT_SAVE */
+#ifdef JERRY_ENABLE_SNAPSHOT_EXEC
+          || feature == JERRY_FEATURE_SNAPSHOT_EXEC
+#endif /* JERRY_ENABLE_SNAPSHOT_EXEC */
+          );
```
zherczeg: Not sure this is correct according to the current style. Thus I recommended || false); @LaszloLango ?

robertsipka: @zherczeg : The cppcheck will fails with the following error msg: same expression on both sides of '||'.

zherczeg: Well, if this passes vera checks we can live with it.

LaszloLango: IMHO the style is good. I think this is the best way to do this.

docs/02.API-REFERENCE.md
```diff
@@ -22,6 +22,18 @@ Possible types of an error:
  - JERRY_ERROR_TYPE - type error
  - JERRY_ERROR_URI - URI error
 
+## jerry_feature_t
+
+Possible types of a feature:
```
zherczeg: Possible feature types:

zherczeg: Or perhaps compile time enabled feature types:

docs/02.API-REFERENCE.md
```diff
@@ -917,6 +929,38 @@ jerry_value_is_undefined (const jerry_value_t value)
 
 - [jerry_release_value](#jerry_release_value)
 
+## jerry_is_feature_enabled
+
+**Summary**
+
+Returns wheither the specified feature is enabled or not.
```
zherczeg: whether the specified compile time feature is enabled.

jerry-core/jerry-api.h
```diff
@@ -68,6 +68,18 @@ typedef enum
   JERRY_ERROR_URI        /**< URIError */
 } jerry_error_t;
 
+typedef enum
+{
+  JERRY_FEATURE_ERROR_MESSAGES, /**< Error messages. */
```
zherczeg: The comment style for fields does not start with capital letters and no full stop at the end. Please check the coding style guidelines.

robertsipka:
@zherczeg : Thanks, I've updated and rebased it with master.

zherczeg:
LGTM

Issue #1386, opened 2016-10-03
new Date() fails for any date in October on leap years


HBehrens:
The return value of `new Date(2016, 9, day)` (or any other leap year) is an invalid date object (`console.log(…) --> "Invalid Date"`) since the binary search in `ecma_date_make_day()` is erroneous.
We do have a working fix, covered by a brute-force unit-test that creates dates in the range from 1950...2050 (that previously failed for every day that follow the aforementioned pattern), but due to time constraints, I won't be able to put up a PR before mid next week.

If for some reason somebody needs a fix before, here's are possible fix:

```
@@ -781,13 +781,16 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
       delta = delta / 2;
     }

-    if (ecma_date_month_from_time (time + step) < mn)
+    time += step * ECMA_DATE_MS_PER_DAY;
+
+    /* Make sure time represents a day in the previous month */
+    if (ecma_date_month_from_time (time) >= mn)
     {
-      time += step * ECMA_DATE_MS_PER_DAY;
+      time -= ECMA_DATE_MS_PER_DAY * 31;
     }

-    /* Get the month's first day */
-    while (ecma_date_month_from_time (time) < mn)
+    /* Get the month's first day, don't check for <= as the month wraps */
+    while (ecma_date_month_from_time (time) != mn)
     {
       time += ECMA_DATE_MS_PER_DAY;
     }
```

Also, these are the tests, we verified this fix against:

```
+void test_rocky_api_util__ecma_date_make_day(void) {
+  cl_assert_equal_d(16861, ecma_date_make_day(2016, 2, 1));  // JerryScript's unit-test
+  cl_assert_equal_d(-25294, ecma_date_make_day(1900, 9, 1));  // not a leap year!
+  cl_assert_equal_d(17075, ecma_date_make_day(2016, 8, 31)); // Sept-31 == Oct-01
+  cl_assert_equal_d(17075, ecma_date_make_day(2016, 9, 1));  // Oct-01
+  cl_assert_equal_d(17045, ecma_date_make_day(2016, 8, 1));  // Sept-01
+}
+
+void test_rocky_api_util__ecma_date_make_day_list(void) {
+  int fail_count = 0;
+  for(int y = 1950; y < 2050; y++) {
+    for(int m = 0; m < 12; m++) {
+      for (int d = 1; d < 32; d++) {
+        const ecma_number_t result = ecma_date_make_day(y, m, d);
[...]

PR #1466, opened 2016-12-01
new Date error solved

LeeHayun:
I solved the problem mentioned in issue #1386 .
(new Date() fails for any date in October on leap years)
The issue was raised long time before, 
but the problem was not solved for a long time.

Therefore, i solved the problem, and
In addition, i enhanced the algorithm. (Change binary search to simple algorithm)
The algorithm uses ecma standard api. 

I did two unit tests (bug fix, performance).
Following is code is used for unit tests on IoT.js.

```javascript
fail_count=0;
for (y=1950; y<2050; y++) {
  for (m=0; m<12; m++) {
    for (d=1; d<32; d++) {
      var date = new Date(y, m, d);
      if (isNaN(date)) {
        fail_count++;
      }
    }
  }
}
console.log(fail_count);
```

```javascript
var t0 = new Date().getTime();
for (y=1950; y<2050; y++) {
  for (m=0; m<12; m++) {
    for (d=1; d<32; d++) {
      var date = new Date(y, m, d);
    }
  }
}
var t1 = new Date().getTime();
var delay = t1 - t0;
console.log(delay + "ms");
```

This table is results of unit tests.

  /  | Original JerryScript | Modified JerryScript
--- | -------------------- | ---------------------
fail_count | 775 | 0
delay (ms) | 33275 | 16823

### DIFF
--- a/tests/jerry/regression-test-issue-1386.js
+++ b/tests/jerry/regression-test-issue-1386.js
@@ -1,0 +1,24 @@
+// Copyright 2016 Hayun Lee (lhy920806@gmail.com)
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+for (var y = 1950; y < 2050; y++) {
+  for (var m = 0; m < 12; m++) {
+    var last_date = new Date(y, m, 1).getDay ();
+    assert (!isNaN (last_date));
+    for (var d = 1; d < 32; d++) {
+      assert (last_date == new Date(y, m, d).getDay ());
+      last_date = (last_date + 1) % 7;
+    }
+  }
+}
--- a/tests/unit/test-date-helpers.c
+++ b/tests/unit/test-date-helpers.c
@@ -162,6 +162,8 @@ main ()
   TEST_ASSERT (ecma_date_make_day (1970, 1, 35) == 65);
   TEST_ASSERT (ecma_date_make_day (1970, 13, 35) == 430);
   TEST_ASSERT (ecma_date_make_day (2016, 2, 1) == 16861);
+  TEST_ASSERT (ecma_date_make_day (2016, 8, 31) == 17075);
+  TEST_ASSERT (ecma_date_make_day (2016, 9, 1) == 17075);
 
   /* ecma_number_t ecma_date_make_date (day, time) */
 
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
@@ -432,7 +432,8 @@ ecma_date_week_day (ecma_number_t time) /**< time value */
     return time; /* time is NaN */
   }
 
-  return (ecma_number_t) fmod ((ecma_date_day (time) + 4), 7);
+  ecma_number_t week_day = (ecma_number_t) fmod ((ecma_date_day (time) + 4), 7);
+  return (week_day < 0) ? 7 + week_day : week_day;
 } /* ecma_date_week_day */
 
 /**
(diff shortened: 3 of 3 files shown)


LaszloLango:
Something isn't right with your branch. Please rebase to the current mater and add unit test to it.

LeeHayun:
@LaszloLango I modified PR and PR message according to your request. Please review about this.

LaszloLango:
@LeeHayun, I'm still missing the test from the PR. Please add it to your commit too, not only to the description. Add a test to `tests/unit/test-date-helpers.c` or to `tests/jerry/regression-test-issue-1386.js`. Both would be the best. The commit message should be improved too. Please add the related issue number to the commit message like "Fixes #1386" and add a bit more detailed description.

LeeHayun:
@LaszloLango I added unit-test code, and regression-test code as you said.
Javascript code which i wrote above is just for testing by myself, so please never mind.
In addition, i modified my commit message as you said. (e.g., Fixes #1386 ~ )
Please check it out. Thanks.

tests/unit/test-date-helpers.c
```diff
@@ -162,6 +162,8 @@ main ()
   TEST_ASSERT (ecma_date_make_day (1970, 1, 35) == 65);
   TEST_ASSERT (ecma_date_make_day (1970, 13, 35) == 430);
   TEST_ASSERT (ecma_date_make_day (2016, 2, 1) == 16861);
+  TEST_ASSERT (ecma_date_make_day (2016, 8, 31) == 17075);
+  TEST_ASSERT (ecma_date_make_day (2016, 9, 1) == 17075);
```
LaszloLango: Different dates must not have the same day value.

esevan: @LaszloLango  ecma_data_make_day (2016, 8, 31) must be same as ecma_data_make_day (2016,9,1) since there's no 2016/9/31 so that it's regarded as 2016/10/1.

LaszloLango: My bad, both values means October 1 of 2016.

tests/jerry/regression-test-issue-1386.js
```diff
@@ -0,0 +1,22 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+for (y=1950; y<2050; y++) {
+  for (m=0; m<12; m++) {
+    for (d=1; d<32; d++) {
+      var date = new Date(y, m, d);
+      assert(!isNaN(date));
```
LeeHayun: @zherczeg The day test loop is from 1 to 31 so that it wouldn't be incremented on such date like (2/31->3/1). Moreover, ecma_date_make_day() returns NaN when time value is not equal to relative day from (y,m,1). This function implemented by [ECMA-262 v5, 15.9.1.12](http://www.ecma-international.org/ecma-262/5.1/#sec-15.9.1.12) So, step 7 in this link guarantees that the date is incremented by one if it's correct.

zherczeg: I think we should find a way to test this since the original problem was arised from untested values.

What about reset the counter when d = 1 (before the d loop) ?

LeeHayun:
@zherczeg I modified regression test code according to your request.

In addition, I found a bug in ecma_date_week_day(). This function returns minus value when return week_day before 1970 years. For example, value of new Date(1950, 1, 1).getDay() is -4. But, in ECMA standard, 15.9.1.6 defines 'WeekDay(t) = (Day(t) + 4) modulo 7' and 5.2 describes 'The notation “x modulo y” (y must be finite and nonzero) computes a value k of the same sign as y (or zero) ...' for modulo operation.

So, I fixed ecma_date_week_day() to comply with standard. Please check it.

zherczeg:
Excellent. This is why it is worth writing complex tests :)

A question: was it intentional not following jerry coding style in the test case?

LeeHayun:
@zherczeg Thank you. Can you tell me detailedly what part not follows jerry coding style?

tests/jerry/regression-test-issue-1386.js
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+for (var y = 1950; y < 2050; y++) {
+  for (var m = 0; m < 12; m++) {
+    var last_date = new Date(y, m, 1).getDay();
```
zherczeg: Please put a space before the open bracket. Same for all function calls.

tests/jerry/regression-test-issue-1386.js
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+for (var y = 1950; y < 2050; y++) {
+  for (var m = 0; m < 12; m++) {
+    var last_date = new Date(y, m, 1).getDay ();
```
LeeHayun: @zherczeg I put a space before the open bracket for all function calls. Please check it.

zherczeg:
LGTM. Nice improvement.

PR #1471, opened 2016-12-05
Check the compile time enabled feature via API function

robertsipka:


LaszloLango:
This patch will increase the binary size.

zherczeg:
This is just partial solution. There a lot of lines below like:

     else if (!strcmp ("--mem-stats", argv[i]))
     {
       flags |= JERRY_INIT_MEM_STATS;
       jerry_port_default_set_log_level (JERRY_LOG_LEVEL_DEBUG);
     }

These options could be checked as well.

robertsipka:
@LaszloLango : Yes, this patch will increase the binary size, but I planning to add a minimal version of the main-unix source file in a follow-up patch to run tests and measurements with it. 
@zherczeg : I've updated this patch. I hope that I understood you correctly and changed the code parts and checked the features in that way you suggested.

zherczeg:
I think it is enough to run the benchmarks with the minimal tool.

jerry-core/jerry.c
```diff
@@ -140,36 +140,6 @@ jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
   /* Zero out all members. */
   memset (&JERRY_CONTEXT (JERRY_CONTEXT_FIRST_MEMBER), 0, sizeof (jerry_context_t));
 
-  if (flags & (JERRY_INIT_MEM_STATS | JERRY_INIT_MEM_STATS_SEPARATE))
```
zherczeg: It is still true that if JERRY_INIT_MEM_STATS_SEPARATE is set, JERRY_INIT_MEM_STATS must be enabled as well.

jerry-main/main-unix.c
```diff
@@ -370,49 +372,99 @@ main (int argc,
...
       }
+      else
+      {
+        jerry_port_default_set_log_level (JERRY_LOG_LEVEL_WARNING);
 
-      save_snapshot_file_name_p = argv[i];
+        jerry_port_log (JERRY_LOG_LEVEL_WARNING,
+                        "Ignoring 'save-snapshot' option because this feature is disabled!\n");
+
+        jerry_port_log (JERRY_LOG_LEVEL_WARNING,
```
zherczeg: Do we need two jerry_port_log call?

jerry-core/jerry.c
robertsipka: Yes its true, I set this under the 'mem-stats-separate' option earlier, but it's better to set here.

jerry-main/main-unix.c
```diff
@@ -513,25 +573,28 @@ main (int argc,
...
-    }
+      size_t snapshot_size;
+      const uint8_t *snapshot_p = read_file (exec_snapshot_file_names[i], &snapshot_size);
 
-    if (jerry_value_has_error_flag (ret_value))
-    {
-      break;
+      if (snapshot_p == NULL)
+      {
+        ret_value = jerry_create_error (JERRY_ERROR_COMMON, (jerry_char_t *) "");
```
zherczeg: You could add an error message here.

robertsipka: The value of the `snapshot_p` assigned by the the `read_file` function before this condition. As I saw it, this function is add an error message before returns with a NULL and because in this case we breaks from the for loop here, we reach a condition which checks if the 'ret_value' value is an error value. In case of an error it calls the 'print_unhandled_exception' function which also gave a detailed error message.

robertsipka: I've wrote a descriptive information of the 'message' property of this error `ret_value` object.

zherczeg:
LGTM

PR #1474, opened 2016-12-07
Remove INSTANTIATED_MASK_32_63 internal property type.

zherczeg:
No real perf (-0.330%) change, binary size is the same (152544 bytes).

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.c
@@ -35,6 +35,46 @@
 static void ecma_instantiate_builtin (ecma_builtin_id_t id);
 
 /**
+ * Helper definition for ecma_builtin_property_list_references.
+ */
+typedef const ecma_builtin_property_descriptor_t *ecma_builtin_property_list_reference_t;
+
+/**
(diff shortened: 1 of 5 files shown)


jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -426,47 +500,20 @@ ecma_builtin_try_to_instantiate_property (ecma_object_t *object_p, /**< object *
 
   uint32_t index = (uint32_t) (curr_property_p - property_list_p);
 
-  JERRY_ASSERT (index < 64);
+  JERRY_ASSERT (index < 96);
```
zherczeg: Actually this could be extended to 256 at some point, but I want to start with this limit. Each of these uses can be extended further by modifying this code.

zherczeg: I mean we literally need to modify all of these instances, so if it would be a general constant, and it would be increased, we would not notice issues with the code.

tilmannOSG:
Thanks for adding the comments! I'll leave the review of the technical aspects of the change to people more familiar with the code :)

PR #1476, opened 2016-12-07
Add feature to remove parser.

sergioamr:
Remove the parser so we can save space if we are only interested in running snapshots.
Removing the parser enables the snapshot execution.

JerryScript-DCO-1.0-Signed-off-by: Sergio Martinez sergio.martinez.rodriguez@intel.com

### DIFF
--- a/jerry-core/CMakeLists.txt
+++ b/jerry-core/CMakeLists.txt
@@ -21,6 +21,7 @@ set(FEATURE_CPOINTER_32_BIT OFF    CACHE BOOL   "Enable 32 bit compressed pointe
 set(FEATURE_ERROR_MESSAGES  OFF    CACHE BOOL   "Enable error messages?")
 set(FEATURE_MEM_STATS       OFF    CACHE BOOL   "Enable memory statistics?")
 set(FEATURE_MEM_STRESS_TEST OFF    CACHE BOOL   "Enable mem-stress test?")
+set(FEATURE_PARSER_DISABLE  OFF    CACHE BOOL   "Disable the parser and 'eval' and use only snapshots?")
 set(FEATURE_PARSER_DUMP     OFF    CACHE BOOL   "Enable parser byte-code dumps?")
 set(FEATURE_PROFILE         "es5.1" CACHE STRING "Profile types: es5.1, minimal, es2015-subset")
 set(FEATURE_REGEXP_DUMP     OFF    CACHE BOOL   "Enable regexp byte-code dumps?")
@@ -35,6 +36,7 @@ message(STATUS "FEATURE_CPOINTER_32_BIT   " ${FEATURE_CPOINTER_32_BIT})
 message(STATUS "FEATURE_ERROR_MESSAGES    " ${FEATURE_ERROR_MESSAGES})
 message(STATUS "FEATURE_MEM_STATS         " ${FEATURE_MEM_STATS})
 message(STATUS "FEATURE_MEM_STRESS_TEST   " ${FEATURE_MEM_STRESS_TEST})
+message(STATUS "FEATURE_PARSER_DISABLE    " ${FEATURE_PARSER_DISABLE})
 message(STATUS "FEATURE_PARSER_DUMP       " ${FEATURE_PARSER_DUMP})
 message(STATUS "FEATURE_PROFILE           " ${FEATURE_PROFILE})
 message(STATUS "FEATURE_REGEXP_DUMP       " ${FEATURE_REGEXP_DUMP})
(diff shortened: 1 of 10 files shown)


jerry-core/parser/js/js-parser.c
```diff
@@ -20,6 +20,8 @@
 #include "jcontext.h"
 #include "js-parser-internal.h"
 
+#ifndef CONFIG_DISABLE_PARSER_BUILTIN
```
zherczeg: Parser is not exactly a built-in. JERRY_DISABLE_PARSER would sound better. Shouldn't we guard the other files too?

sergioamr: Sounds good, i was not certain how to call it. Let me change it,
I checked on the rest of the code and you guys don't guard the includes, that is why is there after the js-parser-internal.h

zherczeg:
I mean all files in parser/js should have guards. I know the linker can eliminate dead code, but I would prefer a complete solution. Btw how much is the binary size gain on your system?

sergioamr:
It is quite a good saving. 

Parser On
126468 build/arduino_101/zephyr/zephyr.bin

Parser Off
93524 build/arduino_101/zephyr/zephyr.bin

I will add the guards to the rest of the files.

jerry-core/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ set(FEATURE_PARSER_DUMP     OFF    CACHE BOOL   "Enable parser byte-code dumps?"
 set(FEATURE_REGEXP_DUMP     OFF    CACHE BOOL   "Enable regexp byte-code dumps?")
 set(FEATURE_SNAPSHOT_SAVE   OFF    CACHE BOOL   "Enable saving snapshot files?")
 set(FEATURE_SNAPSHOT_EXEC   OFF    CACHE BOOL   "Enable executing snapshot files?")
+set(FEATURE_PARSER_DISABLE  OFF    CACHE BOOL   "Disable the parser and 'eval' and use only snapshots?")
```
zherczeg: It seems to me these features are in alphabetic order except the first few (perhaps that should be fixed in a separate patch). You should put this before FEATURE_PARSER_DUMP.

jerry-core/CMakeLists.txt
```diff
@@ -42,6 +43,7 @@ message(STATUS "FEATURE_PARSER_DUMP       " ${FEATURE_PARSER_DUMP})
 message(STATUS "FEATURE_REGEXP_DUMP       " ${FEATURE_REGEXP_DUMP})
 message(STATUS "FEATURE_SNAPSHOT_SAVE     " ${FEATURE_SNAPSHOT_SAVE})
 message(STATUS "FEATURE_SNAPSHOT_EXEC     " ${FEATURE_SNAPSHOT_EXEC})
+message(STATUS "FEATURE_PARSER_DISABLE    " ${FEATURE_PARSER_DISABLE})
```
zherczeg: Ditto.

jerry-core/CMakeLists.txt
```diff
@@ -140,6 +142,12 @@ elseif(NOT FEATURE_PROFILE STREQUAL "es2015-subset")
   message(FATAL_ERROR "FEATURE_PROFILE='${FEATURE_PROFILE}' isn't supported")
 endif()
 
+# Disable parser and enable snapshots
+if(FEATURE_PARSER_DISABLE)
+  set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_DISABLE_PARSER)
+  set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_ENABLE_SNAPSHOT_EXEC)
+endif()
+
```
zherczeg: It seems these checks also follows the order in the feature list.

jerry-core/parser/js/js-parser.c
```diff
@@ -2305,8 +2310,15 @@ parser_parse_script (const uint8_t *source_p, /**< source code */
...
   }
-
   return ecma_make_simple_value (ECMA_SIMPLE_VALUE_TRUE);
+#else /* !JERRY_DISABLE_PARSER */
+  JERRY_UNUSED (source_p);
+  JERRY_UNUSED (size);
+  JERRY_UNUSED (is_strict);
+  JERRY_UNUSED (bytecode_data_p);
+
+  return ecma_raise_syntax_error ("");
```
zherczeg: Please add a valid error message enclosed in ECMA_ERR_MSG, since disabling the parser does not mean disabling error messages.

jerry-core/CMakeLists.txt
sergioamr: Sounds reasonable

PR #1478, opened 2016-12-08
Create a minimal jerry cmdline tool to run benchmarks with it.

robertsipka:


jerry-main/minimal-main-unix.c
```diff
@@ -0,0 +1,210 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
```
zherczeg: I would call this file main-unix-minimal.c

jerry-main/minimal-main-unix.c
```diff
@@ -0,0 +1,210 @@
...
+  return (const uint8_t *) buffer;
+} /* read_file */
+
+/**
+ * Provide the 'assert' implementation for the engine.
+ *
+ * @return true - if only one argument was passed and the argument is a boolean true.
+ */
+static jerry_value_t
+assert_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
```
zherczeg: Do we need to support asserts? This variant is not for testing. We can add it later if needed though.

robertsipka:
@zherczeg : You're right, we don't need to support the asserts right now for minimal testing. I've updated this patch with the requested changes.

zherczeg:
LGTM

robertsipka:
@zherczeg : Thanks, I've modified this patch again with removed the for-loops and the version info, and reduced the argument count, so it's closer to be a minimal tool now. Please check it.

zherczeg:
I have been thinking about a little different approach. I would not remove the ability to run multiple js-es, just not parsing them. I.e.

    if (argc <= 1 || (argc == 2 && (argv[1] == "-h" || argv[1] == "--help")))
      print_help ();

    for (int i = 1; i < argc; i++)
      run(argv[i]);

(This is just a pseudo code, not copy it). Hence if not -h or --help is passed, we just load and execute every argument as JS files. We don't try to interpret them.

robertsipka:
@zherczeg : I like your suggestion. I've modified this patch by keeping the ability to run multiple js-es and also rebased it with master.

jerry-main/main-unix-minimal.c
```diff
@@ -0,0 +1,128 @@
...
+    return JERRY_STANDALONE_EXIT_CODE_OK;
+  }
+
+  jerry_init (JERRY_INIT_EMPTY);
+  jerry_value_t ret_value = jerry_create_undefined ();
+
+  const char *file_name;
+  int i = 1;
+
+  for (; i < argc; i++)
```
LaszloLango: `i` is not used outside the `for`, so please combine these lines.

`for (int i = 1; i < argc; i++)`

robertsipka:
@LaszloLango : Thanks! I've updated it.

PR #1482, opened 2016-12-09
Set js-parser feature in the build script.

robertsipka:
This patch also contains:
- Renamed `FEATURE_PARSER_DISABLE` to `FEATURE_JS_PARSER`.
- Fixed the build error that occurred in case of disabled js-parser.


galpeter:
lgtm (I like the `FEATURE_JS_PARSER` name:))

zherczeg:
There is one thing I am worry about. If somebody just make an all-in-one build, and forget to explicitly define this variable, he will not get a parser and will be surprised. Due to a similar reason we changed float32 default to float64 default. I think we should check somewhere that either snapshot or parser must be enabled, and use # error to signal when both are disabled. If somebody uses the python build system, this could never happen, but if somebody just takes all sources and tries to compile, it may.

robertsipka:
@zherczeg : I agree with you. I've added an #error to signal when both are disabled.

jerry-core/jerry.c
```diff
@@ -49,6 +49,10 @@ JERRY_STATIC_ASSERT ((int) ECMA_ERROR_COMMON == (int) JERRY_ERROR_COMMON
                      && (int) ECMA_ERROR_URI == (int) JERRY_ERROR_URI,
                      ecma_standard_error_t_must_be_equal_to_jerry_error_t);
 
+#if !defined (JERRY_JS_PARSER) && !defined (JERRY_ENABLE_SNAPSHOT_EXEC)
+#error JS_PARSER or JERRY_ENABLE_SNAPSHOT_EXEC must be defined!
+#endif
```
zherczeg: Comment missing after endif.

jerry-core/jerry.c
```diff
@@ -49,6 +49,10 @@ JERRY_STATIC_ASSERT ((int) ECMA_ERROR_COMMON == (int) JERRY_ERROR_COMMON
                      && (int) ECMA_ERROR_URI == (int) JERRY_ERROR_URI,
                      ecma_standard_error_t_must_be_equal_to_jerry_error_t);
 
+#if !defined (JERRY_JS_PARSER) && !defined (JERRY_ENABLE_SNAPSHOT_EXEC)
+#error JS_PARSER or JERRY_ENABLE_SNAPSHOT_EXEC must be defined!
```
galpeter: Should `JS_PARSER` be `JERRY_JS_PARSER` ?

robertsipka:
@zherczeg , @galpeter : Thanks, I've added these minor fixes.

PR #1483, opened 2016-12-10
Add #if control around GCC builtin functions so that the code can be compiled using non-GCC compilers.

t-harvey:
The header says it all -- only some of the GCC builtin functions are protected by "#ifdef (__GNUC__)", and this change would protect the builtin functions in jrt.h and assert.h.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.c
@@ -343,7 +343,7 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg, /**< this argument 
         while (string_curr_p > start_p)
         {
           ecma_char_t current_char = *(--string_curr_p);
-          ecma_number_t current_number;
+          ecma_number_t current_number = ECMA_NUMBER_MINUS_ONE;
 
           if ((current_char >= LIT_CHAR_LOWERCASE_A && current_char <= LIT_CHAR_LOWERCASE_Z))
           {
--- a/jerry-core/jrt/jrt.h
+++ b/jerry-core/jrt/jrt.h
@@ -43,8 +43,13 @@
 /*
  * Conditions' likeliness, unlikeliness.
  */
-#define likely(x) __builtin_expect (!!(x), 1)
-#define unlikely(x) __builtin_expect (!!(x) , 0)
+#ifdef __GNUC__
+#define likely(x)       __builtin_expect(!!(x), 1)
+#define unlikely(x)     __builtin_expect(!!(x), 0)
+#else /* !__GNUC__ */
+#define likely(x)       (x)
+#define unlikely(x)     (x)
+#endif /* __GNUC__ */
 
 /*
  * Normally compilers store const(ant)s in ROM. Thus saving RAM.
@@ -108,7 +113,11 @@ void __noreturn jerry_unreachable (const char *file, const char *function, const
     } \
   } while (0)
 
+#ifdef __GNUC__
 #define JERRY_UNREACHABLE() __builtin_unreachable ()
+#else /* !__GNUC__ */
+#define JERRY_UNREACHABLE()
+#endif /* __GNUC__ */
 #endif /* !JERRY_NDEBUG */
 
 /**
--- a/jerry-libc/include/assert.h
+++ b/jerry-libc/include/assert.h
@@ -28,7 +28,7 @@ extern "C"
 #define assert(x) \
   do \
   { \
-    if (__builtin_expect (!(x), 0)) \
+    if (!(x)) \
     { \
       fprintf (stderr, "%s:%d: %s: Assertion `%s' failed.", __FILE__, __LINE__, __func__, #x); \
       abort (); \


jerry-core/jrt/jrt.h
```diff
@@ -1,4 +1,5 @@
-/* Copyright JS Foundation and other contributors, http://js.foundation
+/* Copyright 2014-2016 Samsung Electronics Co., Ltd.
+ * Copyright 2016 University of Szeged.
  *
```
robertsipka: Please don't change the copyrights.

jerry-core/jrt/jrt.h
```diff
@@ -43,6 +44,10 @@
 /*
  * Conditions' likeliness, unlikeliness.
  */
+#ifndef (__GNUC__)
```
robertsipka: Don't enclose the proceeding identifier with brackets.

robertsipka:
Please fix the DCO in the commit message. Check travis log for details:
https://travis-ci.org/jerryscript-project/jerryscript/jobs/182762008

jerry-core/jrt/jrt.h
```diff
@@ -43,6 +44,10 @@
...
+#define __builtin_expect(expression, expected_value) (expression)
```
zherczeg: No space after !

jerry-libc/include/assert.h
```diff
@@ -25,6 +26,11 @@ extern "C"
 #endif /* __cplusplus */
 
 #ifndef NDEBUG
+
+#ifndef (__GNUC__)
+#define __builtin_expect(expression, expected_value) (expression)
+#endif /* !__GNUC__ */
```
zherczeg: We could remove this if we follow @robertsipka suggestion.

jerry-core/jrt/jrt.h
```diff
@@ -108,7 +113,11 @@ extern void __noreturn jerry_unreachable (const char *, const char *, const uint
     } \
   } while (0)
 
+#ifdef(__GNUC__)
 #define JERRY_UNREACHABLE() __builtin_unreachable ()
+#else /* !__GNUC__ */
+#define JERRY_UNREACHABLE() abort(); /* but what if they haven't included stdlib?!? */
```
zherczeg: No, this isn't exactly working that way. __builtin_unreachable is NOT a function call, it just notifies the compiler, that the code after that is unreachable, and it can optimize the code that way. The compiler is very serious about it even in debug mode, so we needed to add an assert to capture such issues instead of random crashes.

Hm, perhaps the best would be an empty macro in relase, and an ASSERT in debug. Perhaps the ASSERT is added automatically in debug.

tilmannOSG: I think there's some misunderstanding here :)  
`__builtin_unreachable()` is a built-in function which causes undefined behavior when called during actual execution. Since it's a function that never returns the compiler can make assumptions based on that (but so can the compiler about `abort()` which also never returns).  The current behavior of `JERRY_UNREACHABLE()` works just fine in debug builds with a non-GCC compiler so we don't need to worry about that. However, for release builds we need an alternative to `__builtin_unreachable()` since this is a GCC-specific function.  
`abort()` provides us exactly that: It terminates the program in a controlled manner (rather than just crashing the program) and since `abort()` is part of the C standard library this will work just fine with any C99-compliant compiler.  
Makes sense? :)  
(BTW `assert()` is typically implemented on top of `abort()` e.g. it prints some details about the location of the assertion in the code and then just calls `abort()`)

tilmannOSG: One thing I forgot to mention: The key difference between `__builtin_unreachable()` and `abort()` is that calling `__builtin_unreachable()` yields undefined behavior while calling `abort()` is perfectly valid, e.g. not triggering any kind of undefined behavior. The compiler obviously can optimize much more aggressively when undefined behavior is involved. However, since the goal here is to have a solution that works in a standard C99 environment without relying on any compiler-specific built-in functions, `abort()` seems to be the closest match to `__builtin_unreachable()` in terms of intended behavior. Losing some optimization opportunities is essentially the price that needs to be paid to get a solution that works in any standard C99 environment.

zherczeg: I have some memories about `__builtin_unreachable()`, when a bad code actually reached this point, and no return code was generated by the GCC. Hence the code entered into the next function. Hard to debug. This was the reason while we added an assert before `__builtin_unreachable()`, so we can capture these issues in debug mode.

jerry-core/jrt/jrt.h
```diff
@@ -108,7 +113,14 @@ extern void __noreturn jerry_unreachable (const char *, const char *, const uint
     } \
   } while (0)
 
+#ifdef __GNUC__
 #define JERRY_UNREACHABLE() __builtin_unreachable ()
+#else /* !__GNUC__ */
+#define JERRY_UNREACHABLE() { \
+#include <stdlib.h> \
+abort(); \
```
zherczeg: The question still stands: do we need to do anything here?

    extern void __noreturn jerry_unreachable (const char *, const char *, const uint32_t);

In debug mode, the unreachable function calls a real function, which yields an error and terminates the process. In release mode the undefinied behaviour is expected. Hence why we just do nothing here? Why do we increase the binary size by adding abort calls here?

tilmannOSG:
@t-harvey Can you please change `JERRY_UNREACHABLE()` to become a "no op" and squash all your changes into a single commit? I think after that the PR is ready to be merged :)

jerry-libc/include/assert.h
```diff
@@ -25,6 +25,11 @@ extern "C"
 #endif /* __cplusplus */
 
 #ifndef NDEBUG
+
+#ifndef __GNUC__
+#define __builtin_expect(expression, expected_value) (expression)
+#endif /* !__GNUC__ */
```
robertsipka: As I see it this is the main reason why this patch hasn't been merge yet. The assert function uses __builtin_expect (below in a condition). 
@zherczeg , @tilmannOSG : Should this be remained in this form, or should the assert function be defined differently, or do you have any other ideas?

t-harvey: Guys -- I thought the only reason this hadn't yet been merged was b/c I forgot to ping Tillman that I had added the DCO message (sorry, Tillman! :-)  -- what am I missing/what can I do to fix this other problem?  (I had assumed that if the code passed Travis that it was good to go...?)

jerry-core/ecma/builtin-objects/ecma-builtin-global.c
```diff
@@ -343,7 +343,7 @@ ecma_builtin_global_object_parse_int (ecma_value_t this_arg, /**< this argument
         while (string_curr_p > start_p)
         {
           ecma_char_t current_char = *(--string_curr_p);
-          ecma_number_t current_number;
+          ecma_number_t current_number = ECMA_NUMBER_MINUS_ONE;
 
           if ((current_char >= LIT_CHAR_LOWERCASE_A && current_char <= LIT_CHAR_LOWERCASE_Z))
           {
```
zherczeg: I would simply add an assert around lit_char_is_decimal_digit (current_char).

jerry-libc/include/assert.h
zherczeg: I would prefer to replace the __builtin_expect() to unlikely() here. The project should not rely on __builtin_expect.

tilmannOSG: I agree with @zherczeg, since the only direct use of `__builtin_expect()` is in `assert.h` this is best fixed by just replacing this invocation with `unlikely()`.

jerry-core/ecma/builtin-objects/ecma-builtin-global.c
tilmannOSG: @zherczeg While that might work to eliminate the compiler warning, it seems explicitly initializing the variable is still the better option (we should initialize all variables by default to avoid any bugs due to uninitialized variables).

tilmannOSG:
@t-harvey If you just replace the `__builtin_expect()` call in `assert.h` with a call to `unlikely()` then we can go ahead and merger this pull request :)

jerry-libc/include/assert.h
t-harvey: @tilmannOSG && @zherczeg -- I'd like to propose a different  way of looking at this:

The GCC constructs _are_ useful (to GCC :-); removing them seems like we're getting rid of functionality.  Since, with this #def, they behave just fine for every other compiler, why not leave them?

Secondly, "unlikely" is only defined in jrt.h; using it all over the place forces (read: makes it a really good idea :-) us to reorganize things in .h files (for this change, because now we suddenly have to include jrt.h in assert.h, which makes no obvious sense) -- these kinds of reorganizations are required maintenance for an evolving project, but kind of out of the scope for this otherwise simple change (since the bigger fix requires system architectural knowledge, discussion of division, what/when/where to include what, etc... :-).

...so for the record: I'm not trying to get out of work by suggesting that we leave this code as written :-)...but doing this correctly will take a lot of work (and reduce functionality!), so I _am_ actually saying that we should check this in as is...

tilmannOSG: @t-harvey Good point, I think in this case the pragmatic solution is to just drop the `__builtin_expect()` use from `assert.h`. Since this is only affects the code in jerry-libc (rather than jerry-core) and most of the `assert()` calls seem to be in the `printf()` implementation (which is not really performance-critical code) the performance impact should be negligible. Builds with asserts enabled are expected to have a noticeable performance overhead anyway.

zherczeg:
Ok. I haven't noticed we are talking about libc. Yes, just drop it.

tilmannOSG:
@t-harvey So if you just remove the `__builtin_expect()` call from `assert.h` (and the `__builtin_expect` macro in the current patch) then we're good to go to get this finally merged :)

jerry-core/jrt/jrt.h
```diff
@@ -108,7 +113,11 @@ void __noreturn jerry_unreachable (const char *file, const char *function, const
     } \
   } while (0)
 
+#ifdef __GNUC__
 #define JERRY_UNREACHABLE() __builtin_unreachable ()
+#else /* !__GNUC__ */
+#define JERRY_UNREACHABLE()
+#endif /* !__GNUC__ */
```
zherczeg: #endif /* __GNUC__ */

PR #1468, opened 2016-12-02
target: mbedos5: Update to mbed OS 5.2.3

janjongboom:
Update mbed OS target to latest version of mbed OS (5.2.3). Also rename all files in jerryscript-mbed-drivers/ to include -js.cpp, as our build tools now generate warnings for C++ files with the same name. mbed-events library is now mainlined, so no longer required to pull this library in as a separate dependency. ``main.cpp`` is auto-generated and should not be included in this repository.

JerryScript-DCO-1.0-Signed-off-by: Jan Jongboom janjongboom@gmail.com

/cc @thegecko @matthewelse

Verified against:

- mbed-js-easy-connect
- mbed-js-example
- mbed-js-repl-example
- mbed-js-ble-example

When switching to this branch for an existing mbed JS program, you might need to run:

```
$ cd build/jerryscript/targets/mbedos5
$ mbed deploy
$ rm -rf mbed-events
```

Although probably taken care off by gulp.

PR #1484, opened 2016-12-12
Remove the inline flag of ArrayBuffer helper functions

jiangzidong:
related issue #1468

Latest GCC compiler will report an error when the inline function definition is in a different source file
so we remove the inline flag

### DIFF
--- a/jerry-core/ecma/operations/ecma-arraybuffer-object.c
+++ b/jerry-core/ecma/operations/ecma-arraybuffer-object.c
@@ -103,7 +103,7 @@ ecma_arraybuffer_new_object (ecma_length_t length) /**< length of the arraybuffe
  *
  * @return ecma_length_t, the length of the arraybuffer
  */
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-arraybuffer-object.c
```diff
@@ -103,7 +103,7 @@ ecma_arraybuffer_new_object (ecma_length_t length) /**< length of the arraybuffe
  *
  * @return ecma_length_t, the length of the arraybuffer
  */
-inline ecma_length_t __attr_pure___ __attr_always_inline___
+inline ecma_length_t __attr_pure___
```
zherczeg: Why the inline keyword is not removed?

janjongboom:
LGTM, fixes the build error against mbedos5 target with GCC 4.9.3.

zherczeg:
LGTM

PR #1485, opened 2016-12-13
target: mbedos5: Define JERRY_JS_PARSER macro

janjongboom:
Define JERRY_JS_PARSER macro, which is now required to be defined. Fixes build issues against mbedos5 target.


LaszloLango:
LGTM

PR #1486, opened 2016-12-13
Add section about license headers and copyright notices to the Contri…

tilmannOSG:
…bution Guidelines.

Clarify that each contributed file needs to have a proper license header and that the addition of individual copyright notices is not permitted.


zherczeg:
LGTM

PR #1490, opened 2016-12-13
Warning fixes for Clang build.

robertsipka:


jerry-core/lit/lit-magic-strings.c
```diff
@@ -84,7 +84,7 @@ lit_get_magic_string_size (lit_magic_string_id_t id) /**< magic string id */
 lit_magic_string_id_t
 lit_get_magic_string_size_block_start (lit_utf8_size_t size) /**< magic string size */
 {
-  static const lit_magic_string_id_t const lit_magic_string_size_block_starts[] JERRY_CONST_DATA =
+  static const lit_magic_string_id_t lit_magic_string_size_block_starts[] JERRY_CONST_DATA =
```
zherczeg: Is it still a const data?

robertsipka: Yes, it is.

jerry-core/lit/lit-magic-strings.c
```diff
@@ -183,7 +183,7 @@ lit_is_utf8_string_magic (const lit_utf8_byte_t *string_p, /**< utf-8 string */
   while (first <= last)
   {
     int middle = ((first + last) / 2); /**< mid point of search */
-    int compare = memcmp (lit_get_magic_string_utf8 (middle), string_p, string_size);
+    int compare = memcmp (lit_get_magic_string_utf8 ((lit_magic_string_id_t) middle), string_p, string_size);
```
tilmannOSG: Sure that a cast is the right fix here? Casting should only be a last resort if there are no other options :)

jerry-core/lit/lit-magic-strings.c
```diff
@@ -228,7 +228,7 @@ lit_is_utf8_string_pair_magic (const lit_utf8_byte_t *string1_p, /**< first utf-
   while (first <= last)
   {
     int middle = ((first + last) / 2); /**< mid point of search */
-    const lit_utf8_byte_t *middle_string_p = lit_get_magic_string_utf8 (middle);
+    const lit_utf8_byte_t *middle_string_p = lit_get_magic_string_utf8 ((lit_magic_string_id_t) middle);
```
tilmannOSG: Sure that a cast is the right fix here? Casting should only be a last resort if there are no other options :)

zherczeg:
LGTM

jerry-core/lit/lit-magic-strings.c
tilmannOSG: Discussed this offline. Conclusion: A better approach is to rewrite the binary search algorithm to use unsigned integers rather than signed integers for the indices, this avoids the need add a cast for the `lit_get_magic_string_utf8()` call.

robertsipka:
@tilmannOSG, @zherczeg : Thank you for clarifying this, It's really a better approach.
I've modified this patch based on your suggestions.

jerry-core/lit/lit-magic-strings.c
```diff
@@ -222,12 +222,12 @@ lit_is_utf8_string_pair_magic (const lit_utf8_byte_t *string1_p, /**< first utf-
   }
 
   /**< The string must be in this id range. */
-  int first = (int) lit_get_magic_string_size_block_start (total_string_size);
-  int last = (int) (lit_get_magic_string_size_block_start (total_string_size + 1) - 1);
+  lit_utf8_size_t first = lit_get_magic_string_size_block_start (total_string_size);
+  lit_utf8_size_t last = lit_get_magic_string_size_block_start (total_string_size + 1) - 1;
```
zherczeg: The -1 at the end should be removed, isn't it?

robertsipka:
@zherczeg : Yes, thanks, I've updated this patch.

Issue #1493, opened 2016-12-14
target: mbedos5: source/pins.cpp is no longer generated against mbed OS 5.2


janjongboom:
Go to build/jerryscript/targets/mbedos5, then run `make source/pins.cpp`.

This fails, because the `Target` constructor has been removed in mbed OS 5.2 now; which is used in https://github.com/jerryscript-project/jerryscript/blob/master/targets/mbedos5/tools/generate_pins.py.

theotherjimmy:
The `Target` constructor has been replaced by `target` from `tools/targets.json` it should have the same signature.

theotherjimmy:
I take it back. What you want is `Target.get_target`

PR #1494, opened 2016-12-15
target: mbedos5: Update generate_pins.py

janjongboom:
to use the new Python API for querying targets, and use the new location of target PinNames.h file. Previously `make source/pins.cpp` just failed silently, and returned an empty file. Therefore using pin names from JS (`LED1`) would silently fail when building JerryScript against mbed OS 5.2.3. We missed this because it does not happen when you upgrade an existing mbed OS 5.1 project.

@matthewelse @thegecko 

Fix for #1493


thegecko:
LGTM

janjongboom:
@LaszloLango Done ^

PR #1495, opened 2016-12-15
Remove redundant extern keywords from function declarations/definitions.

tilmannOSG:
Extern keywords on function declarations/definitions provide no additional value since function declarations/definitions default to external linkage in C99, e.g. removing them won't change the semantics of the program.

The extern keywords were essentially a legacy from the early days of the project. This commit cleans this up across the whole codebase in one go to minimize history disruption.

The bulk of the changes in this commit were produced by a custom clang-tidy checker.

Note that variables declarations carrying the extern keyword are untouched by this commit since there the presence of the keyword actually has an impact on the semantics of the program.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.h
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.h
@@ -26,24 +26,24 @@
  * @{
  */
 
-extern ecma_value_t
+ecma_value_t
 ecma_builtin_helper_object_to_string (const ecma_value_t);
-extern ecma_value_t
+ecma_value_t
 ecma_builtin_helper_get_to_locale_string_at_index (ecma_object_t *, uint32_t);
-extern ecma_value_t
+ecma_value_t
 ecma_builtin_helper_object_get_properties (ecma_object_t *, bool);
-extern ecma_value_t
+ecma_value_t
 ecma_builtin_helper_array_concat_value (ecma_object_t *, uint32_t *, ecma_value_t);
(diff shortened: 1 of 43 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-helpers.h
```diff
@@ -93,36 +93,36 @@ typedef enum
...
-extern ecma_number_t ecma_date_hour_from_time (ecma_number_t);
-extern ecma_number_t ecma_date_min_from_time (ecma_number_t);
-extern ecma_number_t ecma_date_sec_from_time (ecma_number_t);
-extern ecma_number_t ecma_date_ms_from_time (ecma_number_t);
-extern ecma_number_t ecma_date_make_time (ecma_number_t, ecma_number_t, ecma_number_t, ecma_number_t);
-extern ecma_number_t ecma_date_make_day (ecma_number_t, ecma_number_t, ecma_number_t);
-extern ecma_number_t ecma_date_make_date (ecma_number_t, ecma_number_t);
-extern ecma_number_t ecma_date_time_clip (ecma_number_t);
-extern ecma_number_t ecma_date_timezone_offset (ecma_number_t);
-
```
zherczeg: Interesting, this newline should not be a change. Git diff bug?

tilmannOSG: Yeah, there's no change in that line (I just double checked). Probably some heuristic of diff to form a larger block of changes.

PR #1500, opened 2016-12-21
[API] Introduce jerry_parse_and_save_literals()

bzsolt:
This function can be used to save literals into a specific file in a list or C format.
These literals are valid identifiers, and doesn't match to any magic-string.
The '--save-literals-list-format FILE' and '--save-literals-c-format FILE'
options are used to save into the given file, when the snapshot-save is enabled.
The saved literals are sorted by size and lexicographically.
The C-format is useful for jerry_register_magic_strings() to generate the array
of external magic strings.


jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,421 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+
+  if (buffer_p + str_size <= buffer_end_p)
+  {
+    strncpy ((char *) buffer_p, chars, str_size);
+    *total_bytes += str_size;
+
+    return str_size;
+  }
+
+  /* Move the pointer behind the buffer to prevent further write */
```
zherczeg: to prevent further writes.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,421 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+static lit_utf8_size_t
+append_ecma_string_to_buffer (uint8_t *buffer_p, /**< buffer */
+                              uint8_t *buffer_end_p, /**< the end of the buffer */
+                              size_t *total_bytes, /**< copied bytes */
+                              ecma_string_t *string_p, /**< ecma-string */
+                              lit_utf8_size_t str_size) /**< string size */
+{
+  lit_utf8_size_t bytes_copied = 0;
+
+  JMEM_DEFINE_LOCAL_ARRAY (str_buffer_p, str_size, lit_utf8_byte_t);
```
zherczeg: Can we use ECMA_STRING_TO_UTF8_STRING here?

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,421 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+ * @return true, if the ecma-string is a valid identifier,
+ *         false - otherwise.
+ */
+static bool
+ecma_string_is_valid_identifier (ecma_string_t *string_p)
+{
+  lit_utf8_size_t str_size = ecma_string_get_size (string_p);
+  bool result = false;
+
+  JMEM_DEFINE_LOCAL_ARRAY (str_buffer_p, str_size, lit_utf8_byte_t);
```
zherczeg: Ditto?

tests/unit/test-api.c
```diff
@@ -923,5 +923,56 @@ main (void)
...
   }
 
+  /* Save literals */
+  if (jerry_is_feature_enabled (JERRY_FEATURE_SNAPSHOT_SAVE))
+  {
+    /* C format generation */
+    jerry_init (JERRY_INIT_EMPTY);
+
+    static jerry_char_t literal_buffer_c[256];
+    static const char *code_for_c_format_p = "var object = { aa:'min', Bb:'max' };";
```
zherczeg: Would be good to have a 

    aaa:'xzy0'

tests/unit/test-api.c
```diff
@@ -923,5 +923,56 @@ main (void)
...
+
+    size_t literal_sizes_c_format = jerry_parse_and_save_literals ((jerry_char_t *) code_for_c_format_p,
+                                                                   strlen (code_for_c_format_p),
+                                                                   false,
+                                                                   literal_buffer_c,
+                                                                   sizeof (literal_buffer_c),
+                                                                   true);
+    TEST_ASSERT (literal_sizes_c_format == 153);
+
+    static const jerry_char_t expected_c_format[256] =
```
zherczeg: Can we align it after the = sign?

bzsolt:
I've updated the PR.

docs/02.API-REFERENCE.md
```diff
@@ -3204,3 +3205,59 @@ jerry_exec_snapshot (const void *snapshot_p,
...
+                               size_t source_size,
+                               bool is_strict,
+                               uint8_t *buffer_p,
+                               size_t buffer_size,
+                               bool is_c_format);
+```
+
+- `source_p` - script source, it must be a valid utf8 string.
+- `source_size` - script source size, in bytes.
+- `is_strict` - strict mode
```
zherczeg: All other sentences stops with dot. Please add it here

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
 /**
  * @}
  */
+
+#ifdef JERRY_ENABLE_SNAPSHOT_SAVE
+
+/**
+ * ====================== Functions for literal saving ==========================
+ */
```
zherczeg: Are other sections have these ====================== ?

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+                                 lit_utf8_size_t begin, /**< begin index */
+                                 lit_utf8_size_t end) /**< end index */
+{
+  lit_utf8_size_t mid = (begin + end) / 2;
+
+  if (begin < end)
+  {
+    jerry_save_literals_sort_helper (src, dest, begin, mid);
+    jerry_save_literals_sort_helper (src, dest, mid + 1, end);
+    jerry_save_literals_merging_helper (src, dest, begin, mid, end);
```
zherczeg: Just a question: is there a way to avoid recursion?

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+  }
+} /* jerry_save_literals_sort_helper */
+
+/**
+ * Append characters to the specified buffer.
+ *
+ * @return number of bytes, actually copied to the buffer.
+ */
+static lit_utf8_size_t
+append_chars_to_buffer (uint8_t *buffer_p, /**< buffer */
```
zherczeg: Why these functions have no jerry_ prefix?

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+                                                              ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32);
+
+  JERRY_ASSERT (utf8_str_size <= ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32);
+
+  lit_utf8_size_t bytes_copied = append_chars_to_buffer (buffer_p,
+                                                         buffer_end_p,
+                                                         total_bytes,
+                                                         (const char *) uint32_to_str_buffer,
+                                                         utf8_str_size);
+  return bytes_copied;
```
zherczeg: We don't need to create a variable, just return with the value.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+      {
+        result = false;
+        break;
+      }
+      lit_utf8_incr (&str_start_p);
+    }
+  }
+  else
+  {
+    result = false;
```
zherczeg: result is initialized to false above

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+} /* ecma_string_is_valid_identifier */
+
+#endif /* JERRY_ENABLE_SNAPSHOT_SAVE */
+
+/**
+ * Copy certain string literals into the given buffer in a specified format,
+ * which are valid identifiers and none of them are magic string.
+ *
+ * @return size of the literal-list in bytes, at most equal to the buffer size,
+ *          if the source parsed successfully and the list of the literals isn't empty,
```
zherczeg: extra space in alignment

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+    string_list_p = JMEM_CP_GET_POINTER (ecma_lit_storage_item_t, string_list_p->next_cp);
+  }
+
+  if (literal_count == 0)
+  {
+    return 0;
+  }
+
+  ecma_string_t *literal_array[literal_count];
+  ecma_string_t *literal_array_sorted[literal_count];
```
zherczeg: I would change these to real allocation, since stack consumption might be too big.

jerry-core/jerry-snapshot.c
zherczeg: Btw heapsort does not need recuresive implementation.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,409 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+  if (is_c_format)
+  {
+    /* Save literal count */
+    buffer_p += append_chars_to_buffer (buffer_p,
+                                        buffer_end_p,
+                                        &total_bytes,
+                                        (const char *) "jerry_length_t literal_count = ",
+                                        0);
+
+    buffer_p += append_number_to_buffer (buffer_p, buffer_end_p, &total_bytes, literal_count);
```
zherczeg: I think total_bytes is a redundant argument since it can be computed as buffer_p-buffer_start_p. Passing less arguments is generally faster. Just save the buffer_p into a buffer_start_p.

zherczeg: The function could also return with the new buffer_p instead of the size, which would also reduce the code size, since there is no preserving the original value and an extra addition.

jerry-main/main-unix.c
```diff
@@ -354,6 +356,10 @@ main (int argc,
   bool is_save_snapshot_mode_for_global_or_eval = false;
   const char *save_snapshot_file_name_p = NULL;
 
+  bool is_save_literals_mode = false;
+  bool is_save_literals_mode_in_c_format_or_list = false;
+  const char *save_literals_file_name_p = NULL;
```
zherczeg: Question: can snapshot and save literals be done in one step? It is great if it is possible. Othoerwise only one file name parameter is enough.

jerry-core/jerry-snapshot.c
bzsolt: Yes, at [jerry.c:1867](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/jerry.c#L1867).

jerry-core/jerry-snapshot.c
bzsolt: At `jerry_parse_and_save_snapshot` there are the same style, which one is nicer?

jerry-main/main-unix.c
bzsolt: Yes, it can be done in one step.

jerry-core/jerry-snapshot.c
zherczeg: No extra space. The other is probably a style error.

jerry-core/jerry-snapshot.c
bzsolt: In this file there are other 'helper' functions without `jerry_` prefix (e.g. `snapshot_add_compiled_code`), but ok, I'll prefix them.

jerry-core/jerry-snapshot.c
bzsolt: I think in case of merge sort is not possible. So you advice to use a heapsort algorithm instead of the current merge sort?

zherczeg: Not a complicated algorithm for an array of data:

https://rosettacode.org/wiki/Sorting_algorithms/Heapsort#C

jerry-core/jerry-snapshot.c
bzsolt: When the buffer is insufficient, we cannot determine the exact amount of the used bytes just from the pointers. But if we don't want to save anything in the mentioned case, we can omit this variable.

bzsolt: Ok, meanwhile I've removed this argument.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,433 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+                                                            num_of_nodes,
+                                                            node_idx,
+                                                            2 * node_idx + 1,
+                                                            2 * node_idx + 2);
+    if (max_idx == node_idx)
+    {
+      break;
+    }
+
+    ecma_string_t *swap  = literals[node_idx];
```
zherczeg: swap_p or tmp_p

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,433 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+  while (lit_idx <= (num_of_literals - 2) / 2)
+  {
+    jerry_save_literals_down_heap (literals, num_of_literals, lit_idx--);
+  }
+
+  for (lit_idx = 0; lit_idx < num_of_literals; lit_idx++)
+  {
+    const lit_utf8_size_t last_idx = num_of_literals - lit_idx - 1;
+
+    ecma_string_t *swap = literals[last_idx];
```
zherczeg: Ditto.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,433 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+    ecma_string_t *swap  = literals[node_idx];
+    literals[node_idx] = literals[max_idx];
+    literals[max_idx] = swap;
+
+    node_idx = max_idx;
+  }
+} /* jerry_save_literals_down_heap */
+
+/**
+ * Helper function for a heapsort algorithm.
```
zherczeg: I like the simplicity of this algorithm. No need any large stack.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,433 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+  if (buffer_p > buffer_end_p)
+  {
+    return buffer_p;
+  }
+
+  const lit_utf8_size_t str_size = (string_size == 0) ? (lit_utf8_size_t) strlen (chars) : string_size;
+
+  if (buffer_p + str_size <= buffer_end_p)
+  {
+    strncpy ((char *) buffer_p, chars, str_size);
```
zherczeg: You should use memcpy. The strncpy stops if it encounters a 0, which isn't possible at this moment (string is identifier), but we might add other rules in the future, and this copy will be a difficult to debug.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,433 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+
+    return buffer_p + str_size;
+  }
+
+  /* Move the pointer behind the buffer to prevent further writes. */
+  /* Distance from the buffer represents the unused bytes. */
+  return (buffer_end_p + (buffer_end_p - buffer_p));
```
zherczeg: ??? Distance from the buffer represents the unused bytes ??? - I don't understand this comment.
Isn't it represent the needed bytes of the last append?

What about simply return with:

    return buffer_end_p + 1;

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,433 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+/**
+ * Append characters to the specified buffer.
+ *
+ * @return the position of the buffer pointer after copy.
+ */
+static uint8_t *
+jerry_append_chars_to_buffer (uint8_t *buffer_p, /**< buffer */
+                              uint8_t *buffer_end_p, /**< the end of the buffer */
+                              const char *chars, /**< string */
+                              lit_utf8_size_t string_size) /**< string size */
```
zherczeg: This function looks better without the extra argument :)

jerry-core/jerry-snapshot.c
bzsolt: So, the `buffer_end_p - buffer_p` represents the remaining bytes in the buffer.
When we encounter a string which doesn't fit into the available space 
we move the `buffer_p` after the `buffer_end_p`.
For the saving process, we need to know the exact amount of used bytes in the buffer and
we can compute it easily when we shifted the `buffer_p` according to the unused bytes,
see [jerry-snapshot.c:1049](https://github.com/jerryscript-project/jerryscript/pull/1500/files#diff-9266e41ebf8c1d213c41e0a627aa6714R1049)

zherczeg: > we move the buffer_p after the buffer_end_p

    return buffer_end_p + 1;

This exactly do this. I don't really understand the last sentence.

docs/02.API-REFERENCE.md
```diff
@@ -3204,3 +3205,59 @@ jerry_exec_snapshot (const void *snapshot_p,
 - [jerry_init](#jerry_init)
 - [jerry_cleanup](#jerry_cleanup)
 - [jerry_parse_and_save_snapshot](#jerry_parse_and_save_snapshot)
+
+
+## jerry_parse_and_save_literals
+
+**Summary**
+
+Collect and save the used literals from the specified source code.
```
LaszloLango: I don't understand how it works. Please add more description here.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,430 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+} /* jerry_append_number_to_buffer */
+
+/**
+ * Check whether the passed ecma-string is a valid identifier.
+ *
+ * @return true, if the ecma-string is a valid identifier,
+ *         false - otherwise.
+ */
+static bool
+ecma_string_is_valid_identifier (ecma_string_t *string_p)
```
LaszloLango: `const ecma_string_t *string_p`

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,430 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+ *
+ * @return size of the literal-list in bytes, at most equal to the buffer size,
+ *         if the source parsed successfully and the list of the literals isn't empty,
+ *         0 - otherwise.
+ */
+size_t
+jerry_parse_and_save_literals (const jerry_char_t *source_p, /**< script source */
+                               size_t source_size, /**< script source size */
+                               bool is_strict, /**< strict mode */
+                               uint8_t *buffer_p, /**< buffer to save literals to */
```
LaszloLango: uint8_t *buffer_p, /**< [out] buffer to save literals to */

LaszloLango: I think out_buffer_p or result_buffer_p would be better names (Only my personal preference)

jerry-main/main-unix.c
```diff
@@ -610,25 +658,47 @@ main (int argc,
...
-          ret_value = jerry_create_error (JERRY_ERROR_COMMON, (jerry_char_t *) "");
+          size_t snapshot_size = jerry_parse_and_save_snapshot ((jerry_char_t *) source_p,
+                                                                source_size,
+                                                                is_save_snapshot_mode_for_global_or_eval,
+                                                                false,
+                                                                snapshot_save_buffer,
+                                                                JERRY_BUFFER_SIZE);
+          if (snapshot_size == 0)
+          {
+            ret_value = jerry_create_error (JERRY_ERROR_COMMON, (jerry_char_t *) "");
```
LaszloLango: Please don't leave the message empty. (I know it was empty before.)

jerry-main/main-unix.c
```diff
@@ -610,25 +658,47 @@ main (int argc,
...
-          fclose (snapshot_file_p);
+          const size_t literal_buffer_size = jerry_parse_and_save_literals ((jerry_char_t *) source_p,
+                                                                            source_size,
+                                                                            false,
+                                                                            snapshot_save_buffer,
+                                                                            JERRY_BUFFER_SIZE,
+                                                                            is_save_literals_mode_in_c_format_or_list);
+          if (literal_buffer_size == 0)
+          {
+            ret_value = jerry_create_error (JERRY_ERROR_COMMON, (jerry_char_t *) "");
```
LaszloLango: ditto

jerry-main/main-unix.c
```diff
@@ -610,25 +658,47 @@ main (int argc,
...
+          }
+          else
+          {
+            FILE *literal_file_p = fopen (save_literals_file_name_p, "w");
+            fwrite (snapshot_save_buffer, sizeof (uint8_t), literal_buffer_size, literal_file_p);
+            fclose (literal_file_p);
```
LaszloLango: I'd add this file saving code the example source code of the `jerry_parse_and_save_literal` function in API docs.

tests/unit/test-api.c
```diff
@@ -923,5 +923,61 @@ main (void)
...
+    static jerry_char_t literal_buffer_c[256];
+    static const char *code_for_c_format_p = "var object = { aa:'min', Bb:'max', aaa:'xzy0' };";
+
+    size_t literal_sizes_c_format = jerry_parse_and_save_literals ((jerry_char_t *) code_for_c_format_p,
+                                                                   strlen (code_for_c_format_p),
+                                                                   false,
+                                                                   literal_buffer_c,
+                                                                   sizeof (literal_buffer_c),
+                                                                   true);
+    TEST_ASSERT (literal_sizes_c_format == 203);
```
LaszloLango: 203? why is it so big for such a small example?

jerry-core/jerry-snapshot.c
bzsolt: Yes, the effect is the same, but how many bytes would you save when the buffer is insufficient?

tests/unit/test-api.c
bzsolt: That is the size of the C-style format, please look 3 lines below for the expected result.

jerry-core/jerry-snapshot.c
zherczeg: I still don't understand the purpose of the computation, please explain it to me.

bzsolt: We shift the buffer pointer by the number of the unused bytes in the buffer instead of 1,
thus we can determine the number of the used bytes at the end of `jerry_parse_and_save_literals()`.

zherczeg: Hey, but that is why I suggested returning with buffer_end_p + 1, and not buffer_p + 1. So we get a pointer which is right after the end of the buffer. Am I missing something?

zherczeg: I think we return with 0 if the buffer is too small, we have no plans to return with the required number of bytes.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
  */
+
+#ifdef JERRY_ENABLE_SNAPSHOT_SAVE
+
+/**
+ * ====================== Functions for literal saving ==========================
+ */
+
+/**
+ * Compare two ecma_string by size, then lexicographically.
```
zherczeg: ecma_strings

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+      && jerry_save_literals_compare (literals[max_idx], literals[child_idx2]))
+  {
+    max_idx = child_idx2;
+  }
+
+  return max_idx;
+} /* jerry_save_literals_heap_max */
+
+/**
+ * Helper function for a heapsort algorithm.
```
zherczeg: for the heapsort

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+  if (lit1_size == lit2_size)
+  {
+    return ecma_compare_ecma_strings_relational (literal1, literal2);
+  }
+
+  return (lit1_size < lit2_size);
+} /* jerry_save_literals_compare */
+
+/**
+ * Helper function for a heapsort algorithm.
```
zherczeg: for the heapsort

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+static uint8_t *
+jerry_append_ecma_string_to_buffer (uint8_t *buffer_p, /**< buffer */
+                                    uint8_t *buffer_end_p, /**< the end of the buffer */
+                                    ecma_string_t *string_p) /**< ecma-string */
+{
+  uint8_t *new_buffer_p = NULL;
+
+  ECMA_STRING_TO_UTF8_STRING (string_p, str_buffer_p, str_buffer_size);
+
+  /* Append the string to the buffer */
```
zherczeg: Dot after sentence.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+    return 0;
+  }
+
+  ecma_free_value (parse_status);
+  ecma_bytecode_deref (bytecode_data_p);
+
+  ecma_lit_storage_item_t *string_list_p = JERRY_CONTEXT (string_list_first_p);
+  lit_utf8_size_t literal_count = 0;
+
+  /* Count the valid and non-magic identifiers in the list */
```
zherczeg: Dot after sentence.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+                                      is_strict,
+                                      &bytecode_data_p);
+
+  if (ECMA_IS_VALUE_ERROR (parse_status))
+  {
+    ecma_free_value (parse_status);
+    return 0;
+  }
+
+  ecma_free_value (parse_status);
```
zherczeg: I alittle optimization here:

    bool error = ECMA_IS_VALUE_ERROR (parse_status);
    ecma_free_value (parse_status);

    if (error)
    {
      return 0;
    }

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+  while (string_list_p != NULL)
+  {
+    for (int i = 0; i < ECMA_LIT_STORAGE_VALUE_COUNT; i++)
+    {
+      if (string_list_p->values[i] != JMEM_CP_NULL)
+      {
+        ecma_string_t *literal_p = JMEM_CP_GET_NON_NULL_POINTER (ecma_string_t,
+                                                                 string_list_p->values[i]);
+        /* We don't save a literal which isn't a valid identifier
+           or it's a magic string */
```
zherczeg: Dot after sentence.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+        if (ecma_string_is_valid_identifier (literal_p)
+            && ecma_get_string_magic (literal_p) == LIT_MAGIC_STRING__COUNT)
```
zherczeg: I would check ecma_get_string_magic (literal_p) == LIT_MAGIC_STRING__COUNT first.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+  {
+    for (int i = 0; i < ECMA_LIT_STORAGE_VALUE_COUNT; i++)
+    {
+      if (string_list_p->values[i] != JMEM_CP_NULL)
+      {
+        ecma_string_t *literal_p = JMEM_CP_GET_NON_NULL_POINTER (ecma_string_t,
+                                                                 string_list_p->values[i]);
+
+        if (ecma_string_is_valid_identifier (literal_p)
+            && ecma_get_string_magic (literal_p) == LIT_MAGIC_STRING__COUNT)
```
zherczeg: Ditto.

jerry-core/jerry-snapshot.c
```diff
@@ -630,3 +632,425 @@ jerry_exec_snapshot (const void *snapshot_p, /**< snapshot */
...
+        {
+          literal_array[literal_idx++] = literal_p;
+        }
+      }
+    }
+
+    string_list_p = JMEM_CP_GET_POINTER (ecma_lit_storage_item_t, string_list_p->next_cp);
+  }
+
+  /* Sort the strings by size at first, then lexicographically */
```
zherczeg: Please add dots after each sentence.

jerry-main/main-unix.c
```diff
@@ -485,6 +491,38 @@ main (int argc,
...
+      {
+        jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Error: no file specified for %s\n", argv[i - 1]);
+        print_usage (argv[0]);
+        return JERRY_STANDALONE_EXIT_CODE_FAIL;
+      }
+
+      if (jerry_is_feature_enabled (JERRY_FEATURE_SNAPSHOT_SAVE))
+      {
+        is_save_literals_mode = true;
+        is_save_literals_mode_in_c_format_or_list = !strcmp ("--save-literals-c-format", argv[i - 1]);
```
zherczeg: I think we can simply check the length, since the two string lengths are different.

bzsolt: I've updated the PR according to your review, but there is another occurence the mentioned comparison at [main-unix.c:445](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-main/main-unix.c#L445) in save-snapshot section. It could be modified in a follow-up patch.

jerry-main/main-unix.c
```diff
@@ -485,6 +491,38 @@ main (int argc,
...
+      {
+        jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Error: no file specified for %s\n", argv[i - 1]);
+        print_usage (argv[0]);
+        return JERRY_STANDALONE_EXIT_CODE_FAIL;
+      }
+
+      if (jerry_is_feature_enabled (JERRY_FEATURE_SNAPSHOT_SAVE))
+      {
+        is_save_literals_mode = true;
+        is_save_literals_mode_in_c_format_or_list = strlen ("--save-literals-c-format") == strlen (argv[i - 1]);
```
zherczeg: Unfortunately this is looks less nicely. Please revert to the original, this is not time critical. Sorry for the extra work.

PR #1504, opened 2016-12-21
Skip out license-check in case of test262 tests

bzsolt:
Can break testruns, when the test262 folder already present.

### DIFF
--- a/tools/check-license.py
+++ b/tools/check-license.py
@@ -17,6 +17,7 @@
 import os
 import re
 import sys
+from settings import *
 
 
 license = re.compile(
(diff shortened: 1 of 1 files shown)


zherczeg:
LGTM

tools/check-license.py
```diff
@@ -46,7 +46,8 @@
 ]
 
 exclude_dirs = [
-    'targets/esp8266'
+    'targets/esp8266',
+    'tests/test262',
```
LaszloLango: It would be better to use `TEST262_TEST_SUITE_DIR` from settings.py, but it is an absolute path. Should be get a relative path from PROJECT_DIR and TEST262_TEST_SUITE_DIR. @galpeter, can we do it easily in python?

PR #1505, opened 2016-12-22
Isolate GCC specific code

t-harvey:
Folks,
This is the second half of our attempt to enable Jerryscript for compilers other than GCC -- in this case, the TI compilers.

Copied from the commit message:

    These changes are designed to enable the TI compilers to compile
    Jerryscript.  The changes include:
    
    CMakeLists.txt: we added conditionals around GCC-specific flags, added
    support for TI flags, and removed the always-on debugging flag (-g)
    
    /toolchain_mcu_tim4f.cmake: new toolchain file uses TI-specific parameters
    
    jerry-api.h: the sys/types.h file was #include'd but never used, so we
    removed it
    
    jrt-types.h: ditto
    
    jerry-port-default-date.c: the TI toolchain doesn't include
    sys/time.h, so we guarded uses of the package

...thanks for reviewing these changes!
Tim

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -48,13 +48,35 @@ if("${PLATFORM}" STREQUAL "DARWIN")
   set(ENABLE_STATIC_LINK "OFF")
 endif()
 
+
+if(CMAKE_C_COMPILER_ID MATCHES "GNU")
+  set(USING_GCC 1)
+endif()
+
+if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+  set(USING_CLANG 1)
+endif()
+
+if(CMAKE_C_COMPILER_ID MATCHES "TI")
+  set(USING_TI 1)
+endif()
+
 # Status messages
 message(STATUS "CMAKE_BUILD_TYPE          " ${CMAKE_BUILD_TYPE})
 message(STATUS "CMAKE_SYSTEM_NAME         " ${CMAKE_SYSTEM_NAME})
 message(STATUS "CMAKE_SYSTEM_PROCESSOR    " ${CMAKE_SYSTEM_PROCESSOR})
 message(STATUS "ENABLE_ALL_IN_ONE         " ${ENABLE_ALL_IN_ONE})
 message(STATUS "ENABLE_LTO                " ${ENABLE_LTO})
-message(STATUS "ENABLE_STATIC_LINK        " ${ENABLE_STATIC_LINK})
+
+if(USING_TI)
+# If using a compiler that _only_ does static linking, inform the user
+# of the discrepancy in settings.
+  set(ENABLE_STATIC_LINK "ON")
+  message(STATUS "ENABLE_STATIC_LINK        " ${ENABLE_STATIC_LINK} " (ONLY OPTION FOR THIS COMPILER)")
+else()
+  message(STATUS "ENABLE_STATIC_LINK        " ${ENABLE_STATIC_LINK})
+endif()
+
 message(STATUS "ENABLE_STRIP              " ${ENABLE_STRIP})
 message(STATUS "JERRY_CMDLINE             " ${JERRY_CMDLINE})
 message(STATUS "JERRY_CMDLINE_MINIMAL     " ${JERRY_CMDLINE_MINIMAL})
@@ -91,7 +113,7 @@ endmacro()
 macro(jerry_add_compile_warnings)
   foreach(_warning ${ARGV})
     jerry_add_compile_flags(-W${_warning})
-    if(CMAKE_COMPILER_IS_GNUCC)
+    if(USING_GCC)
       jerry_add_compile_flags(-Werror=${_warning})
     endif()
   endforeach()
@@ -107,20 +129,27 @@ jerry_add_flags(CMAKE_EXE_LINKER_FLAGS ${FLAGS_COMMON_ARCH})
 
 # Enable static build
 if(ENABLE_STATIC_LINK)
-  jerry_add_link_flags("-static")
+  if (USING_GCC OR USING_CLANG)
+    jerry_add_link_flags("-static")
+  endif()
 endif()
 
 # LTO
 if(ENABLE_LTO)
-  jerry_add_compile_flags(-flto)
-  jerry_add_link_flags(-flto)
-  if(CMAKE_COMPILER_IS_GNUCC)
+  if (USING_GCC OR USING_CLANG)
+    jerry_add_compile_flags(-flto)
+    jerry_add_link_flags(-flto)
+  endif()
+  if(USING_GCC)
     if(NOT "${PLATFORM}" STREQUAL "DARWIN")
       jerry_add_compile_flags(-fno-fat-lto-objects)
     endif()
     # Use gcc-ar and gcc-ranlib to support LTO
     set(CMAKE_AR "gcc-ar")
     set(CMAKE_RANLIB "gcc-ranlib")
+  endif()
+  if(USING_TI)
+    jerry_add_link_flags(-lto)
   endif()
 endif()
 
@@ -130,7 +159,9 @@ if(${PORT_DIR} STREQUAL "${CMAKE_SOURCE_DIR}/targets/default" AND NOT JERRY_LIBC
 endif()
 
 # Compiler / Linker flags
-jerry_add_compile_flags(-fno-builtin)
+if (USING_GCC OR USING_CLANG)
+  jerry_add_compile_flags(-fno-builtin)
+endif()
 if(("${PLATFORM}" STREQUAL "DARWIN"))
   jerry_add_link_flags(-lSystem)
 else()
@@ -143,17 +174,18 @@ if(JERRY_LIBC)
 endif()
 
 # Turn off stack protector
+if (USING_GCC OR USING_CLANG)
 jerry_add_compile_flags(-fno-stack-protector)
+endif()
 
-# Debug information
-jerry_add_compile_flags(-g)
+if (USING_GCC OR USING_CLANG)
+  jerry_add_compile_warnings(all extra format-nonliteral init-self conversion sign-conversion format-security missing-declarations)
+  jerry_add_compile_flags(-Wno-stack-protector -Wno-attributes)
+endif()
 
-jerry_add_compile_warnings(all extra format-nonliteral init-self conversion sign-conversion format-security missing-declarations)
-jerry_add_compile_flags(-Wno-stack-protector -Wno-attributes)
-
-if(CMAKE_COMPILER_IS_GNUCC)
+if(USING_GCC)
   jerry_add_compile_warnings(logical-op)
-else()
+elseif(USING_CLANG)
   jerry_add_compile_flags(-Wno-nested-anon-types -Wno-static-in-inline)
 endif()
 
@@ -162,11 +194,17 @@ if(JERRY_LIBC)
 endif()
 
 # C
-jerry_add_compile_flags(-std=c99 -pedantic)
+if (USING_GCC OR USING_CLANG)
+  jerry_add_compile_flags(-std=c99 -pedantic)
+elseif(USING_TI)
+  jerry_add_compile_flags(--c99)
+endif()
 
 # Strip binary
 if(ENABLE_STRIP AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
-  jerry_add_link_flags(-s)
+  if (USING_GCC OR USING_CLANG)
+    jerry_add_link_flags(-s)
+  endif()
 endif()
 
 # External compiler & linker flags
--- a/cmake/toolchain_mcu_tim4f.cmake
+++ b/cmake/toolchain_mcu_tim4f.cmake
@@ -1,0 +1,33 @@
+# Copyright JS Foundation and other contributors, http://js.foundation 
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+include(CMakeForceCompiler)
+
+set(CMAKE_SYSTEM_NAME MCU)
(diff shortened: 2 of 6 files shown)


CMakeLists.txt
```diff
@@ -88,10 +88,25 @@ macro(jerry_add_compile_flags)
   jerry_add_flags(CMAKE_C_FLAGS ${ARGV})
 endmacro()
 
+# SHOULD THIS BE, SAY, AT THE TOP OF THE FILE?
+if(CMAKE_C_COMPILER_ID MATCHES "GNU")
```
robertsipka: I'm not sure that this condition is necessary, maybe you should use CMAKE_COMPILER_IS_GNUCC to check if the compiler is a variant of gcc.

CMakeLists.txt
```diff
@@ -88,10 +88,25 @@ macro(jerry_add_compile_flags)
...
+  set(USING_GCC 1)
+  set(USING_GCC_OR_CLANG 1)
+endif()
+
+if(CMAKE_C_COMPILER_ID MATCHES "Clang")
+  set(USING_CLANG 1)
+  set(USING_GCC_OR_CLANG 1)
```
robertsipka: I think the USING_GCC_OR_CLANG variable is unnecessary, instead you should check if the compiler is a variant of gcc or a Clang compiler.

CMakeLists.txt
```diff
@@ -102,22 +117,28 @@ macro(jerry_add_link_flags)
 endmacro()
 
 # build mode specific compile/link flags
+if(USE_GCC)
 set(CMAKE_C_FLAGS_RELEASE "-Os")
```
robertsipka: I think you should also specify this optimization level for Clang.

CMakeLists.txt
```diff
@@ -102,22 +117,28 @@ macro(jerry_add_link_flags)
...
+endif()
 
 # Architecture-specific compile/link flags
 jerry_add_compile_flags(${FLAGS_COMMON_ARCH})
 jerry_add_flags(CMAKE_EXE_LINKER_FLAGS ${FLAGS_COMMON_ARCH})
 
 # Enable static build
 if(ENABLE_STATIC_LINK)
-  jerry_add_link_flags("-static")
+  if (USING_GCC_OR_CLANG)
```
robertsipka: If the static linking isn't supported with TI compilers you should set this value to OFF and send a message that It has been disabled, because the compilers doesn't support it and this condition will be needless.

CMakeLists.txt
```diff
@@ -102,22 +117,28 @@ macro(jerry_add_link_flags)
...
+    jerry_add_link_flags("-static")
+  endif()
 endif()
 
 # LTO
 if(ENABLE_LTO)
-  jerry_add_compile_flags(-flto)
-  jerry_add_link_flags(-flto)
-  if(CMAKE_COMPILER_IS_GNUCC)
+  if (USING_GCC_OR_CLANG)
```
robertsipka: ditto.

robertsipka:
Thank you for doing this patch, I think this patch is a step in the right direction to build JerryScript with other compilers, like TI compilers.

CMakeLists.txt
```diff
@@ -146,17 +169,18 @@ if(JERRY_LIBC)
 endif()
 
 # Turn off stack protector
+if (USING_GCC_OR_CLANG)
 jerry_add_compile_flags(-fno-stack-protector)
+endif()
 
-# Debug information
-jerry_add_compile_flags(-g)
```
robertsipka: Do you add this flag somewhere else?

CMakeLists.txt
t-harvey: I was working off of the cmake documentation -- https://cmake.org/cmake/help/v3.0/variable/CMAKE_LANG_COMPILER_ID.html -- is there a better/conventional way to do this?

CMakeLists.txt
t-harvey: I didn't; --  I wasn't confident in this change.  :-) 

I understand that -g was put in as a default variable b/c of an earlier bug in the code (maybe in jerry-main?...my memory doesn't get better as I age :-) -- and I know that interim fixes have made setting -g as the default unnecessary.

I like the flow that has the user set the cmake variables CMAKE_<LANG>_<DEBUG|RELEASE> to include/exclude settings -- this, I think, matches user expectations...

...but, as with all of these, I'll gladly make an argument and start a discussion, but I've got no ego in this: if I'm wrong here, I'll happily fix it (if we put -g back as an always-set flag, does it go back here, or, if not, where?)...

CMakeLists.txt
robertsipka: Yes, it's correct in both cases, I was wondering how to avoid this `USING_GCC` variable, since the CMAKE_COMPILER_IS_GNUCC is also true when CMAKE_C_COMPILER_ID is
"GNU",  but I think you're right and using directly the CMAKE_C_COMPILER_ID to determine the GNU compiler is closely aligned with the conventional way.

CMakeLists.txt
t-harvey: It's actually the other way: we _only_ support static linking.  :-)

This means that we don't have a "-static" flag, which is why it's behind the GCC/Clang conditional...

CMakeLists.txt
t-harvey: You're right -- I missed this one...  The TI compiler flag is slightly different, and I've added it, here.

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -313,7 +313,7 @@ ecma_op_general_object_define_own_property (ecma_object_t *object_p, /**< the ob
   JERRY_ASSERT (property_desc_p->is_writable_defined || !property_desc_p->is_writable);
 
   /* 1. */
-  ecma_extended_property_ref_t ext_property_ref;
+  ecma_extended_property_ref_t ext_property_ref = {0,0};
```
t-harvey: Somehow, my push caused Travis to report that this variable could be used before being initialized, so I just added the obvious initialization...

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -313,7 +313,7 @@ ecma_op_general_object_define_own_property (ecma_object_t *object_p, /**< the ob
   JERRY_ASSERT (property_desc_p->is_writable_defined || !property_desc_p->is_writable);
 
   /* 1. */
-  ecma_extended_property_ref_t ext_property_ref = {0,0};
+  ecma_extended_property_ref_t ext_property_ref = {{0},0};
```
t-harvey: ...and now Travis says this is wrong:

./jerry-core/ecma/operations/ecma-objects-general.c:316: error: brace should be placed on a separate line

...I assume that this is a style checker, objecting to the braces?  This code, however, looks like the right way to write it: am I not following a standard, or is there a bug in the style-checking code?

zherczeg:
Please rebase this patch to master.

CMakeLists.txt
robertsipka: I see :). If the ENABLE STATIC_LINK is off, please set its value to ON and write a STATUS message that the ENABLE_STATIC_LINK has been enabled.

CMakeLists.txt
robertsipka: Seems legit, it's not necessary to compiling a release build with debug flags.

t-harvey:
Robert,
In re ENABLE_STATIC_LINK -- so I played with a number of error messages, and I never got one I liked: they either fell in unexpected places (so the user isn't looking for them), or they interrupted other output (they just got in the way).

I compromised by hanging a comment off of the current printout of flags: 

...normally, you'd see this:

-- ENABLE_ALL_IN_ONE         OFF
-- ENABLE_LTO                ON
-- ENABLE_STATIC_LINK        ON
-- ENABLE_STRIP              ON
-- JERRY_CMDLINE             OFF

...if the compiler doesn't do dynamic linking, you get this:

-- ENABLE_ALL_IN_ONE         OFF
-- ENABLE_LTO                ON
-- ENABLE_STATIC_LINK        ON (ONLY OPTION FOR THIS COMPILER)
-- ENABLE_STRIP              ON
-- JERRY_CMDLINE             OFF

...is this okay?

I've squashed all of the commits, so if everything looks good, this is ready for pulling.

zherczeg:
The #1483 should be landed first since this PR depends on that.

robertsipka:
@t-harvey : Yes, this message form is perfect, thanks. 
I would not set the USING_GCC_OR_CLANG variable. Could you use "USING_GCC OR USING_CLANG" in the affected condition instead of "USING_GCC_OR_CLANG"?  LGTM (informally) after these changes.

LaszloLango:
@t-harvey, please rebase to the current master.

CMakeLists.txt
```diff
@@ -101,27 +123,39 @@ macro(jerry_add_link_flags)
   jerry_add_flags(LINKER_FLAGS_COMMON ${ARGV})
 endmacro()
 
+# build mode specific compile/link flags
+if(USING_GCC OR USING_CLANG)
+set(CMAKE_C_FLAGS_RELEASE "-Os")
+endif()
+
```
robertsipka: This is not needed anymore, see #1513.

t-harvey: Removed...  I had a little trouble getting the rebase done (and I had a failure in the runtime tests, which was weird, as it went away over the course of the afternoon :-), but it should be ready to go now...

cmake/toolchain_mcu_tim4f.cmake
```diff
@@ -0,0 +1,33 @@
...
+
+SET (CMAKE_C_FLAGS_DEBUG_INIT          "-g")
+SET (CMAKE_C_FLAGS_MINSIZEREL_INIT     "-o4 -mf0 -DNDEBUG")
+SET (CMAKE_C_FLAGS_RELEASE_INIT        "-o4 -DNDEBUG")
+SET (CMAKE_C_FLAGS_RELWITHDEBINFO_INIT "-o2 -g")
+
+SET (CMAKE_CXX_FLAGS_DEBUG_INIT          "-g")
+SET (CMAKE_CXX_FLAGS_MINSIZEREL_INIT     "-o4 -mf0 -DNDEBUG")
+SET (CMAKE_CXX_FLAGS_RELEASE_INIT        "-o4 -DNDEBUG")
+SET (CMAKE_CXX_FLAGS_RELWITHDEBINFO_INIT "-o2 -g")
```
LaszloLango: Please add new line at the end of the file

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -313,7 +313,7 @@ ecma_op_general_object_define_own_property (ecma_object_t *object_p, /**< the ob
   JERRY_ASSERT (property_desc_p->is_writable_defined || !property_desc_p->is_writable);
 
   /* 1. */
-  ecma_extended_property_ref_t ext_property_ref;
+  ecma_extended_property_ref_t ext_property_ref = {{0},0};
```
LaszloLango: The Vera++ rule is not clever enough to recognize this. But the correct style would be `{ { 0 }, 0 }`
I suggest to use a define to workaround this problem.

```diff
diff --git a/jerry-core/ecma/base/ecma-globals.h b/jerry-core/ecma/base/ecma-globals.h
index 99b6da3..fa87e91 100644
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -468,6 +468,11 @@ typedef struct
 #define ECMA_PROPERTY_SEARCH_DEPTH_LIMIT 128
 
 /**
+ * Empty extended property reference
+ */
+#define ECMA_EMPTY_EXTENDED_PROPERTY_REFERENCE { { 0 }, 0 }
+
+/**
  * Property reference. It contains the value pointer
  * for real, and the value itself for virtual properties.
  */
diff --git a/jerry-core/ecma/operations/ecma-objects-general.c b/jerry-core/ecma/operations/ecma-objects-general.c
index c24f356..5b479b2 100644
--- a/jerry-core/ecma/operations/ecma-objects-general.c
+++ b/jerry-core/ecma/operations/ecma-objects-general.c
@@ -313,7 +313,7 @@ ecma_op_general_object_define_own_property (ecma_object_t *object_p, /**< the ob
   JERRY_ASSERT (property_desc_p->is_writable_defined || !property_desc_p->is_writable);
 
   /* 1. */
-  ecma_extended_property_ref_t ext_property_ref;
+  ecma_extended_property_ref_t ext_property_ref = ECMA_EMPTY_EXTENDED_PROPERTY_REFERENCE;
   ecma_property_t current_prop;
 
   current_prop = ecma_op_object_get_own_property (object_p,
```

zherczeg:
> The #1483 should be landed first since this PR depends on that.
Just to not forget about it.

CMakeLists.txt
```diff
@@ -48,13 +48,35 @@ if("${PLATFORM}" STREQUAL "DARWIN")
...
 message(STATUS "CMAKE_BUILD_TYPE          " ${CMAKE_BUILD_TYPE})
 message(STATUS "CMAKE_SYSTEM_NAME         " ${CMAKE_SYSTEM_NAME})
 message(STATUS "CMAKE_SYSTEM_PROCESSOR    " ${CMAKE_SYSTEM_PROCESSOR})
 message(STATUS "ENABLE_ALL_IN_ONE         " ${ENABLE_ALL_IN_ONE})
 message(STATUS "ENABLE_LTO                " ${ENABLE_LTO})
-message(STATUS "ENABLE_STATIC_LINK        " ${ENABLE_STATIC_LINK})
+
+if(USING_TI)
+# if using a compiler that _only_ does static linking, inform the user
+# of the discrepancy in settings
```
tilmannOSG: Please start the comment with an upper case letter and end with a full stop :)

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -313,7 +313,9 @@ ecma_op_general_object_define_own_property (ecma_object_t *object_p, /**< the ob
   JERRY_ASSERT (property_desc_p->is_writable_defined || !property_desc_p->is_writable);
 
   /* 1. */
-  ecma_extended_property_ref_t ext_property_ref;
+  /* this #def just gets around the syntax/style checker...*/
```
tilmannOSG: Please start the comment with an upper case letter as well :)

tilmannOSG:
@t-harvey Thanks for the update, this PR is now ready to be merged (can be merged once #1483 has landed).

LaszloLango:
@t-harvey, #1483 is merged. Please rebase, so we can merge this PR too.

PR #1506, opened 2016-12-22
[API] Improve the performance of the external magic id search

bzsolt:
After this patch, we have to provide external strings ordered by size and lexicographically.
We can do this with jerry_parse_and_save_literals() (#1500).

### DIFF
--- a/jerry-core/lit/lit-magic-strings.c
+++ b/jerry-core/lit/lit-magic-strings.c
@@ -157,6 +157,24 @@ lit_magic_strings_ex_set (const lit_utf8_byte_t **ex_str_items, /**< character a
     lit_utf8_size_t string_size = lit_zt_utf8_string_size (lit_get_magic_string_ex_utf8 (id));
     JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] == string_size);
     JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] <= LIT_MAGIC_STRING_LENGTH_LIMIT);
+
+    /**
+     * Check whether the strings are sorted by size and lexicographically,
+     * e.g., "Bb" < "aa" < "aaa" < "xyz0".
+     */
+    if (id > 0)
+    {
+      const lit_magic_string_ex_id_t prev_id = id - 1;
+      const lit_utf8_size_t prev_string_size = lit_get_magic_string_ex_size (prev_id);
+      JERRY_ASSERT (prev_string_size <= string_size);
+
+      if (prev_string_size == string_size)
+      {
+        const lit_utf8_byte_t *prev_ex_string_p = lit_get_magic_string_ex_utf8 (prev_id);
+        const lit_utf8_byte_t *curr_ex_string_p = lit_get_magic_string_ex_utf8 (id);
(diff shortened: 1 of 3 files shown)


jerry-core/lit/lit-magic-strings.c
```diff
@@ -157,6 +157,22 @@ lit_magic_strings_ex_set (const lit_utf8_byte_t **ex_str_items, /**< character a
     lit_utf8_size_t string_size = lit_zt_utf8_string_size (lit_get_magic_string_ex_utf8 (id));
     JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] == string_size);
     JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] <= LIT_MAGIC_STRING_LENGTH_LIMIT);
+
+    /* Check the order of the strings  */
```
LaszloLango: I think we should add more description here. What is the right order? Maybe with a small example.

jerry-core/lit/lit-magic-strings.c
```diff
@@ -157,6 +157,25 @@ lit_magic_strings_ex_set (const lit_utf8_byte_t **ex_str_items, /**< character a
...
+      const lit_magic_string_ex_id_t prev_id = id - 1;
+      const lit_utf8_size_t prev_string_size = lit_get_magic_string_ex_size (prev_id);
+      JERRY_ASSERT (prev_string_size <= string_size);
+
+      if (prev_string_size == string_size)
+      {
+        const lit_utf8_byte_t *prev_ex_string_p = lit_get_magic_string_ex_utf8 (prev_id);
+        const lit_utf8_byte_t *curr_ex_string_p = lit_get_magic_string_ex_utf8 (id);
+        const int string_compare = memcmp (prev_ex_string_p, curr_ex_string_p, string_size);
+        JERRY_ASSERT (string_compare < 0);
```
zherczeg: JERRY_ASSERT (memcmp (prev_ex_string_p, curr_ex_string_p, string_size) < 0)

jerry-core/lit/lit-magic-strings.c
```diff
@@ -264,22 +283,57 @@ lit_magic_string_ex_id_t
...
+        first = middle + 1;
       }
+      else
+      {
+        last = middle;
+      }
+    }
+    else if (found_size_range)
+    {
+      break;
```
zherczeg: Are you sure this is correct?

a
b
c
d
e
ff
gg

Search e.

First, you find d, and found_size_range is set to 1. Then you find ff, and the loop aborts. If I understand correctly.

bzsolt: I've fixed it, thanks.

tests/unit/test-api.c
```diff
@@ -150,8 +150,8 @@ handler_construct (const jerry_value_t func_obj_val, /**< function object */
  */
 #define JERRY_MAGIC_STRING_ITEMS \
   JERRY_MAGIC_STRING_DEF (GLOBAL, global) \
```
robertsipka: The `global` is already in the list of ECMA magic strings, so it doesn't have to be an external magic string because the defined magic string id will be found at first. But I think it could be remove in a follow-up patch, otherwise LGTM (informally).

PR #1508, opened 2017-01-02
Fix urls in the example

kkristof:
The urls in the docs/03.API-EXAMPLE.md was pointing to
the old samsung.github.io pages.

JerryScript-DCO-1.0-Signed-off-by: Kristof Kosztyo kkristof@inf.u-szeged.hu


galpeter:
lgtm

PR #1512, opened 2017-01-04
target: mbedos5: mbed CLI refuses to build if no .mbed file present

### LINKED ISSUE #1511: Problem with building jerryscript on mbed os 5
I try the tutorial below to build jerryscript to an mbed os 5 target.
https://developer.mbed.org/blog/entry/Using-mbed-libraries-with-JerryScript/

In the Setting up section of the tutorial, you can see:

<------------------------
Setting up

The following instructions guide you through the process of setting up an environment where you can run JavaScript on mbed OS:

Open a terminal and navigate to the folder where you cloned the JerryScript repository.
Run cd targets / mbedos5.
Run make getlibs - this will pull in mbed OS and all related repositories.
Run pip install -r ./tools/requirements.txt - this will install Python modules required by the build scripts.
------------------------->
The problem is that when I run the 'make getlibs' command, the following error message appears.

<------------------------
[mbed] ERROR: Could not find mbed program in current path "/ home / parallels / jerryscript / targets / mbedos5".
[mbed] ERROR: Change the current directory to a valid mbed program or use the 'global' option to set global configuration.

mbed config root.
[mbed] ERROR: Could not find mbed program in current path "/ home / parallels / jerryscript / targets / mbedos5".
[mbed] ERROR: Change the current directory to a valid mbed program or use the 'global' option to set global configuration.

make: *** [.mbed] Error 255
------------------------->

So I opened the Makefile and looked at the getlibs configuration section.

<------------------------
getlibs: .mbed

.mbed:
mbed config root.
mbed toolchain GCC_ARM
mbed target $ (BOARD)
mbed deploy
------------------------->

I would like to know what to do if the above error occurs.

Thanks in advance

janjongboom:
With the newest release of mbed CLI (1.0.0) it error's when no .mbed file is present in the project root. This breaks builds for JerryScript on mbed OS 5. With this patch we create this file and builds succeed again. We cannot use `mbed new` because it also creates a new git repository in targets/mbedos5 folder which is not what we want.

/cc @thegecko @LMESTM @zherczeg @limjunsung

Fixes #1511

### DIFF
--- a/targets/mbedos5/.gitignore
+++ b/targets/mbedos5/.gitignore
@@ -2,6 +2,7 @@ mbed-os
 mbed-events
 .build
 .mbed
+.temp/
 mbed_settings.py
 js/pins.js
 source/pins.cpp
--- a/targets/mbedos5/Makefile
+++ b/targets/mbedos5/Makefile
@@ -80,6 +80,7 @@ endif
 getlibs: .mbed
 
 .mbed:
+	echo 'ROOT=.' > .mbed
 	mbed config root .
 	mbed toolchain GCC_ARM
 	mbed target $(BOARD)


LMESTM:
working fine thx !

thegecko:
LGTM

zherczeg:
LGTM

limjunsung:
It seems to be working, but the error message still appears...

$ make getlibs

[mbed] ERROR: Could not find mbed program in current path "/Users/limjunsung/jerryscript/targets/mbedos5".
[mbed] ERROR: Change the current directory to a valid mbed program, set the current directory as an mbed program with 'mbed config root .', or use the '--global' option to set global configuration.
echo 'ROOT=.' > .mbed
mbed config root .
[mbed] . now set as default root in program "mbedos5"
mbed toolchain GCC_ARM
[mbed] GCC_ARM now set as default toolchain in program "mbedos5"
mbed target
[mbed] No default target set in program "mbedos5"
mbed deploy
[mbed] WARNING: Program "mbedos5" in "/Users/limjunsung/jerryscript/targets/mbedos5" does not use source control management.
[mbed] WARNING: To fix this you should use "mbed new ." in the root of your program.
[mbed] Adding library "mbed-os" from "https://github.com/ARMmbed/mbed-os" at rev #d5de476f74dd

thegecko:
Thanks for the feedback, the output you describe is as expected and is a known limitation of the mbed build system. https://github.com/ARMmbed/mbed-cli/issues/283

PR #1513, opened 2017-01-05
 Change the default build type to MinSizeRel.

robertsipka:
The CMAKE_C_FLAGS_RELEASE  had been overwritten earlier to change the optimization level (from -O3 to -Os), but missed to add -DNDEBUG flag. The MinSizeRel build type ensure these flags and it is exactly the same that was the original plan.


zherczeg:
LGTM

PR #1515, opened 2017-01-09
Remove `global` from the list of external magic strings in the api test

robertsipka:
The `global` is already in the list of ECMA magic strings, so it doesn't have to be
an external magic string because the defined magic string id will be found at first.


zherczeg:
LGTM

LaszloLango:
IMHO it is intentional. It tests that there is no error if you add a string that is already existing as an internal magic string. Someone who use the API does not have to know what are the internal magic strings. I'd leave as is.

zherczeg:
What shall we do with PR?

robertsipka:
I agree with @LaszloLango, we should leave it. I would like to separate the API tests to several test files in the future where I will also take care of these test cases. I'am closing it now.

PR #1517, opened 2017-01-10
Refactor the ecma_string_substr function

robertsipka:


jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1779,87 +1779,44 @@ ecma_string_substr (const ecma_string_t *string_p, /**< pointer to an ecma strin
...
-      return ecma_new_ecma_string_from_utf8 (start_p, (lit_utf8_size_t) (end_p - start_p));
-    }
-
-    /**
-     * I. Dump original string to plain buffer
-     */
-    ecma_string_t *ecma_string_p;
-
-    JMEM_DEFINE_LOCAL_ARRAY (utf8_str_p, buffer_size, lit_utf8_byte_t);
+    return ecma_new_ecma_string_from_utf8 (NULL, 0);
```
zherczeg: Just return with an empty magic string.

robertsipka: Thanks for the reviews, I've updated this patch.

PR #1518, opened 2017-01-10
target: mbedos5: Sort magic strings

janjongboom:
Magic strings are expected to be sorted by length, then alphabetically after https://github.com/jerryscript-project/jerryscript/pull/1506 landed. This breaks the mbedos5 target against JerryScript master as it emits the pins in the order that it finds the pins. This patch sorts the pin names.

/cc @thegecko

### DIFF
--- a/targets/mbedos5/tools/generate_pins.py
+++ b/targets/mbedos5/tools/generate_pins.py
@@ -173,7 +173,11 @@ if __name__ == "__main__":
     # enumerate pins from PinNames.h
     pins = enumerate_pins(pins_file, ['./tools'] + list(includes), defines)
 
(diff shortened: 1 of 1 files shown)


janjongboom:
To verify:

* Build [mbed-js-example](https://github.com/armmbed/mbed-js-example) against JerryScript master. Fails with: `Assertion 'prev_string_size <= string_size' failed at ../../jerry-core/lit/
lit-magic-strings.c(lit_magic_strings_ex_set):169.`
* Build with this patch enabled, program runs as expected.

zherczeg:
LGTM. Thank you for the update.

Issue #1519, opened 2017-01-10
checking instanceof against a native external function gives TypeError instead of true/false


grgustaf:
Here's a transcript with bin/jerry that demonstrates the issue:

```
jerry> function foo() { }
undefined
jerry> typeof foo
function
jerry> typeof assert
function
jerry> null instanceof foo
false
jerry> null instanceof assert
Script Error: TypeError
jerry> bar = new foo()
[object Object]
jerry> bar instanceof foo
true
jerry> bar instanceof assert
Script Error: TypeError
```

I'm wanting to write a JS constructor in C, and I think it would probably work except for this problem. What happens here to assert also happens to my function.

zherczeg:
I made fix in #1520. However, you still need to manually create prototypes. See the regression tests.

grgustaf:
OK I can confirm the patch does fix the problem as stated. Interestingly, it doesn't seem to fix the same problem for an external function *I* add myself though. I've done basically this:

```c
jerry_value_t myfunc(const jerry_value_t ...) {
    return jerry_create_undefined();
}
...
jerry_value_t global = jerry_get_global_object();
jerry_value_t func = jerry_create_external_function(myfunc);
jerry_value_t name = jerry_create_string("EmptyObject");
jerry_set_property(global, name, func);
...
```

Then in JavaScript I see that `null instanceof EmptyObject` still gives me a TypeError. I can create an object `foo = new EmptyObject()` but `foo instanceof EmptyObject` also gives me a TypeError.

So apparently this is a slightly different problem than with assert, but this is the one I'm really looking to solve. :) I can open a new issue if you prefer.

grgustaf:
Now I see what you meant about manually setting prototypes. The patch fixes `null instanceof assert` but `({} instanceof assert)` still gives TypeError unless you set a prototype.

But even if I set a prototype: `EmptyObject.prototype = {}` in JS, checking `null instanceof EmptyObject` still gives TypeError. Also, if I try to set the prototype on `func` in C, it the .prototype still shows up as undefined in JS, and still gives a TypeError, though `jerry_set_prototype` returns success.

zherczeg:
> Now I see what you meant about manually setting prototypes. The patch fixes null instanceof assert but ({} instanceof assert) still gives TypeError unless you set a prototype.

This is the expected.

I will need to check the rest. There are lot of rules here where you need to return with type error or false. Yes, instanceof is expected to return with type error in many cases.

[...]

PR #1520, opened 2017-01-11
HasInstance should accept the prototype of external functions.

zherczeg:


LaszloLango:
Please add "Fixes #1519" to the commit message.

zherczeg:
Done.

tests/jerry/function-external.js
```diff
@@ -0,0 +1,41 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+try {
+  ({} instanceof assert);
+  assert (false);
+} catch(e) {
+}
```
LaszloLango: I'd add `assert(e instanceof TypeError);` here like in other tests. IMHO the file name should be `regression-test-issue-1519.js`, because this is a bug fix for a GitHub issue.

tests/jerry/function-external.js
```diff
@@ -0,0 +1,41 @@
...
+  assert (!({} instanceof assert));
+} catch(e) {
+  assert (false);
+}
+
+try {
+  ({} instanceof Math.sin);
+  assert (false);
+} catch(e) {
+}
```
LaszloLango: `assert(e instanceof TypeError);`

tests/jerry/function-external.js
zherczeg: Actually no. The effect will be the same for that test, and that is expected. This patch actually adds a feature.

tests/jerry/function-external.js
```diff
@@ -0,0 +1,41 @@
...
+try {
+  ({} instanceof assert);
+  assert (false);
+} catch(e) {
+}
+
+assert.prototype = {}
+
+try {
+  assert (!({} instanceof assert));
```
tilmannOSG: No space before the opening parenthesis of the function call please since this is JS code.

tests/jerry/function-external.js
```diff
@@ -0,0 +1,41 @@
...
+} catch(e) {
+  assert (false);
```
tilmannOSG: Same here and all the other assert() calls in the test.

tests/jerry/regression-test-issue-736.js
```diff
@@ -31,7 +31,7 @@ try {
 }
 
 try {
-  eval("var x; x instanceof assert;");
+  eval("var x = {}; x instanceof assert;");
```
tilmannOSG: This looks like an unrelated change (bug fix of a test). Shouldn't this be a separate commit? If you discovered the issue after adding the new functionality in this patch then it can be part of this commit but should be mentioned in the commit message.

zherczeg: It is a related test, which fails after this patch since `undefined instanceof assert` does not throw error anymore. This is the correct behaviour. However, the throwing is tested here, so x needs to be an object.

zherczeg:
Thank you for the review, patch updated.

tests/jerry/regression-test-issue-736.js
tilmannOSG: Thanks for the clarification!

PR #1523, opened 2017-01-12
Improve some wording in reference counting documentation

grgustaf:

### DIFF
--- a/docs/06.REFERENCE-COUNTING.md
+++ b/docs/06.REFERENCE-COUNTING.md
@@ -11,7 +11,7 @@ by `jerry_release_value`.
   /* The value stored in the 'global' variable contains a live
    * reference to the global object. The system also keeps its
    * own live reference to the global object. These two references
-   * are indepent, and both must be destroyed before the global
+   * are independent, and both must be destroyed before the global
    * object can be freed. */
 
   jerry_release_value (global);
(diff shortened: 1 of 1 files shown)


docs/06.REFERENCE-COUNTING.md
```diff
@@ -48,16 +48,17 @@ following code is an **INCORRECT WAY** of releasing the 3.14 value.
 
 JerryScript API functions returning with a `jerry_value_t` always
 return with a new live reference. Passing a `jerry_value_t` to
-an API function never releases its reference. The next example
-shows this behaviour through property getting and setting.
+an API function never releases its reference (except for
+`jerry_release_value`, obviously). The next example shows this
+behaviour through property getting and setting.
```
grgustaf: Maybe this is silly, but 'never' is such a strong word. :)

docs/06.REFERENCE-COUNTING.md
```diff
@@ -76,20 +77,21 @@ shows this behaviour through property getting and setting.
 
   /* Property setting is the same. */
 
+  jerry_value_t new_prop_value = jerry_create_number (2.718);
```
grgustaf: I thought this made it more clear to show where new_prop_value comes from, like in all the API doc samples.

If it was passed in from another C function, it might *not* need to be released here, but in the caller. So this shows it was acquired right here and needs to be released here. If you agree, maybe the same thing should be added to the "simplest form" snippet right below this example too, but I didn't do that.

docs/06.REFERENCE-COUNTING.md
zherczeg: Feel free to change it, but emphasize that you can count on this, unless explicitly stated otherwise (at this moment there is otherwise).

docs/06.REFERENCE-COUNTING.md
zherczeg: Ok for me.

docs/06.REFERENCE-COUNTING.md
```diff
@@ -76,20 +77,21 @@ shows this behaviour through property getting and setting.
...
   jerry_value_t result = jerry_set_property (..., new_prop_value);
 
-  /* If the property set is successful a new reference is created
+  /* If the property set is successful, a new reference is created
    * for the value referenced by new_prop_value. The new_prop_value
-   * reference must be released regardless the operation is
+   * reference must be released regardless of whether the operation is
```
zherczeg: move "is" to the next line.

zherczeg:
LGTM

PR #1524, opened 2017-01-13
Copy the characters of an UTF-8 encoded substring into a specified buffer

robertsipka:


jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1060,6 +1060,135 @@ ecma_substring_copy_to_cesu8_buffer (const ecma_string_t *string_desc_p, /**< ec
 } /* ecma_substring_copy_to_cesu8_buffer */
 
 /**
+ * Convert ecma-string's contents to an utf-8 string, extract the parts of the converted string beetween the specified
```
zherczeg: between

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1060,6 +1060,135 @@ ecma_substring_copy_to_cesu8_buffer (const ecma_string_t *string_desc_p, /**< ec
...
+  }
+  else
+  {
+    end_pos -= start_pos;
+    while (start_pos--)
+    {
+      ecma_char_t ch;
+      lit_utf8_size_t code_unit_size = lit_read_code_unit_from_utf8 (cesu8_str_p, &ch);
+      cesu8_str_p += code_unit_size;
+      if (lit_is_code_point_utf16_high_surrogate (ch))
```
zherczeg: Can't we avoid double decoding here?

robertsipka: @zherczeg : Thanks, I've updated this part with a check that we decoded the next unicode code unit earlier or not.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1060,6 +1060,168 @@ ecma_substring_copy_to_cesu8_buffer (const ecma_string_t *string_desc_p, /**< ec
...
+        next_ch_size = lit_read_code_unit_from_utf8 (cesu8_str_p, &next_ch);
+        if (lit_is_code_point_utf16_low_surrogate (next_ch))
+        {
+          JERRY_ASSERT (code_unit_size == next_ch_size);
+          cesu8_str_p += code_unit_size;
+          next_ch = 0;
+          next_ch_size = 0;
+        }
+      }
+    }
```
zherczeg: Ok, this is really complex. I am sorry I wasn't right. Double check is rare anyway, so the old algorithm is good as well.

robertsipka: Yes, it's true. Ok, I've revoked these changes.

PR #1529, opened 2017-01-16
Remove 'OBJECT_ID' from builtin headers.

LaszloLango:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.inc.h
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.inc.h
@@ -16,11 +16,6 @@
 /*
  * Array.prototype built-in description
  */
-
-#ifndef OBJECT_ID
-# define OBJECT_ID(builtin_object_id)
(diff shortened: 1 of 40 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.inc.h
```diff
@@ -34,7 +29,6 @@
 #endif /* !ROUTINE */
 
 /* Object identifier */
-OBJECT_ID (ECMA_BUILTIN_ID_ARRAY_PROTOTYPE)
```
zherczeg: Patch looks good, but why do we leave the comment? I don't see the purpose.

LaszloLango: Sorry, I missed the comments. It was not intentional. :) I'll fix it.

LaszloLango:
Thanks for the reviews. @tilmannOSG, I've updated the commit message. Please check.

tilmannOSG:
@LaszloLango Thanks, looks good!

PR #1541, opened 2017-01-24
Use default system allocator on 32bit systems, if JERRY_SYSTEM_ALLOCATOR is defined.

LaszloLango:


LaszloLango:
Tested on x86_64
```bash
tools/build.py --error-messages=on --debug --clean --jerry-libc=off --compile-flag="-m32" --cpointer-32bit=on
/mnt/work/Samsung/jerryscript/tools/runners/run-test-suite.sh /mnt/work/Samsung/jerryscript/build/bin/jerry /mnt/work/Samsung/jerryscript/tests/jerry
/mnt/work/Samsung/jerryscript/tools/runners/run-test-suite.sh /mnt/work/Samsung/jerryscript/build/bin/jerry /mnt/work/Samsung/jerryscript/tests/jerry-test-suite/es51-profile-list
for f in `ls tests/jerry` ; do tmp=$(valgrind build/bin/jerry "tests/jerry/$f" 2>&1 | grep "no leaks are possible") ; if [ -z "$tmp" ] ; then echo "MEMORY LEAK in $f" ; else echo "no leak in $f" ; fi ; done
for f in `cat tests/jerry-test-suite/es51-profile-list` ; do tmp=$(valgrind build/bin/jerry "tests/jerry-test-suite/$f" 2>&1 | grep "no leaks are possible") ; if [ -z "$tmp" ] ; then echo "MEMORY LEAK in $f" ; else echo "no leak in $f" ; fi ; done
```

The only error is `tests/jerry/parser-oom.js`, because the engine can allocate enough memory and does not throw out of memory error as expected.

dbatyai:
Seems to me that we still statically allocate space for `jerry_global_heap` at `jcontext.c:39`. We should do something about that as well, since we won't be needing it with the system allocator.

LaszloLango:
@dbatyai I've updated the PR.

zherczeg:
This should be a compile time enabled feature, not mandatory . Am I right?

LaszloLango:
Yes, it is a compile time feature, but changes the current behavior of JERRY_CPOINTER_32_BIT. If it is set then the engine will use the system allocator instead of jerry allocator.

zherczeg:
Why can't we choose what we want to combine? I expect this will increase memory consumption. Have you tried it?

LaszloLango:
@zherczeg, I've updated the PR.

jerry-core/CMakeLists.txt
```diff
@@ -194,6 +196,12 @@ if(FEATURE_SNAPSHOT_SAVE)
   set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_ENABLE_SNAPSHOT_SAVE)
 endif()
 
+# Enable system allocator
+if(FEATURE_SYSTEM_ALLOCATOR)
+  set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_SYSTEM_ALLOCATOR)
+  set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_CPOINTER_32_BIT)
```
dbatyai: Instead of sneakily enabling `JERRY_CPOINTER_32_BIT` here, I'd suggest setting `FEATURE_CPOINTER_32_BIT` somewhere before line 35, this way the status message will correctly print if the feature is enabled or not.

dbatyai:
We should also do something about the mem-stat build, either send a warning/error that it is unsupported with the system allocator enabled, or just update it and make it work.

LaszloLango:
@dbatyai, thanks for the review. I've updated the PR. I added a TODO comment for the mem-stat support with system allocators.

PR #1553, opened 2017-01-31
Append external libraries separately not as a string

bzsolt:


bzsolt:
The current behavior can lead to a link-error, when the second and further libraries doesn't prefixed with `-l`.

PR #1569, opened 2017-02-09
add typedarray routine: reduce and reduceRight

jiangzidong:

### DIFF
--- a/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
+++ b/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
@@ -355,6 +355,166 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
 } /* ecma_builtin_typedarray_prototype_map */
 
 /**
+ * Reduce and reduceRight routines share a similar structure.
+ * And we use 'is_right' to distinguish between them.
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_typedarray_prototype_reduce_with_direction (ecma_value_t this_arg, /**< this argument */
+                                                         ecma_value_t cb_func_val, /**< callback function */
+                                                         ecma_value_t initial_val, /**< initial value */
+                                                         bool is_right) /**< choose order, true is reduceRight */
+{
+  if (!ecma_is_typedarray (this_arg))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Argument 'this' is not a TypedArray."));
+  }
+
+  if (!ecma_op_is_callable (cb_func_val))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Callback function is not callable."));
+  }
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (this_arg);
+  uint32_t len = ecma_typedarray_get_length (obj_p);
+
+  if (len == 0)
+  {
+    if (ecma_is_value_undefined (initial_val))
+    {
+      return ecma_raise_type_error (ECMA_ERR_MSG ("Initial value cannot be undefined."));
+    }
+
+    return ecma_copy_value (initial_val);
+  }
+
+  JERRY_ASSERT (len > 0);
+
+  ecma_value_t accumulator = ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
+  uint32_t index = is_right ? (len - 1) : 0;
+
+  if (ecma_is_value_undefined (initial_val))
+  {
+    accumulator = ecma_op_typedarray_get_index_prop (obj_p, index);
+
+    JERRY_ASSERT (ecma_is_value_number (accumulator));
+
+    if (is_right)
+    {
+      if (index == 0)
+      {
+        return accumulator;
+      }
+
+      index--;
+    }
+    else
+    {
+      index++;
+
+      if (index == len)
+      {
+        return accumulator;
+      }
+    }
+  }
+  else
+  {
+    accumulator = ecma_copy_value (initial_val);
+  }
+
+  ecma_object_t *func_object_p = ecma_get_object_from_value (cb_func_val);
+
+  while (true)
+  {
+    ecma_value_t current_index =  ecma_make_uint32_value (index);
+    ecma_value_t get_value = ecma_op_typedarray_get_index_prop (obj_p, index);
+    ecma_value_t call_args[] = { accumulator, get_value, current_index, this_arg };
+
+    JERRY_ASSERT (ecma_is_value_number (get_value));
+
+    ecma_value_t call_value = ecma_op_function_call (func_object_p,
+                                                     ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED),
+                                                     call_args,
+                                                     4);
+
+    ecma_fast_free_value (accumulator);
+    ecma_fast_free_value (get_value);
+    ecma_fast_free_value (current_index);
+
+    if (ECMA_IS_VALUE_ERROR (call_value))
+    {
+      return call_value;
+    }
+
+    accumulator = call_value;
+
+    if (is_right)
+    {
+      if (index == 0)
+      {
+        break;
+      }
+
+      index--;
+    }
+    else
+    {
+      index++;
+
+      if (index == len)
+      {
+        break;
+      }
+    }
+  }
+
+  return accumulator;
+} /* ecma_builtin_typedarray_prototype_reduce_with_direction */
+
+/**
+ * The %TypedArray%.prototype object's 'reduce' routine
+ *
+ * See also:
+ *          ES2015, 22.2.3.19
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_typedarray_prototype_reduce (ecma_value_t this_arg, /**< this argument */
+                                          ecma_value_t cb_func_val, /**< callback function */
+                                          ecma_value_t initial_val) /**< initial value */
+{
+  return ecma_builtin_typedarray_prototype_reduce_with_direction (this_arg,
+                                                                  cb_func_val,
+                                                                  initial_val,
+                                                                  false);
+} /* ecma_builtin_typedarray_prototype_reduce */
+
+/**
+ * The %TypedArray%.prototype object's 'reduceRight' routine
+ *
+ * See also:
+ *          ES2015, 22.2.3.20
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_typedarray_prototype_reduce_right (ecma_value_t this_arg, /**< this argument */
+                                                ecma_value_t cb_func_val, /**< callback function */
+                                                ecma_value_t initial_val) /**< initial value */
+{
+  return ecma_builtin_typedarray_prototype_reduce_with_direction (this_arg,
+                                                                  cb_func_val,
+                                                                  initial_val,
+                                                                  true);
+} /* ecma_builtin_typedarray_prototype_reduce_right */
+
+/**
  * @}
  * @}
  * @}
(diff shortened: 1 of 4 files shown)


LaszloLango:
> ./jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c:395: error: there should be exactly one space after right parentheses
> ./jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c:447: error: there should be no blank characters after opening parentheses

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,155 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
 } /* ecma_builtin_typedarray_prototype_map */
 
 /**
+ * The routine reduce and reduceRight share the similar structure.
```
zherczeg: Reduce and reduceRight routines share a similar structure.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,155 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+ * And we use 'is_right' to distinguish them
```
zherczeg: between them

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,155 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+  }
+  else
+  {
+    accumulator = ecma_copy_value (initial_val);
+  }
+
+  ecma_object_t *func_object_p = ecma_get_object_from_value (cb_func_val);
+  ecma_value_t ret_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY);
+  bool not_finish = is_right ? (index >= 0) : (index < len);
+  uint32_t _index;
```
zherczeg: Do you need this variable? Please never start a variable name with underscore.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,155 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+  ecma_object_t *obj_p = ecma_get_object_from_value (this_arg);
+  uint32_t len = ecma_typedarray_get_length (obj_p);
+
+  if (len == 0 && ecma_is_value_undefined (initial_val))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Initial value cannot be undefined."));
+  }
+
+  ecma_value_t accumulator = ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
+  int64_t index = is_right ? (len - 1) : 0;
```
zherczeg: int64? As far as I know we cannot allocate such amount of memory. And int64 is not a native type on 32 bit machines.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,155 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+
+    ECMA_FINALIZE (call_value);
+
+    ecma_fast_free_value (get_value);
+    ecma_fast_free_value (current_index);
+
+    if (is_right)
+    {
+      index--;
+      not_finish = (index >= 0);
```
zherczeg: What about simply break if index == 0 ? Before index-- of course.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
jiangzidong: I refer the Array's reduceRight routine and use int64_t. 
I will use "if index == 0" to avoid using signed int type.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,175 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+
+    accumulator = ecma_op_typedarray_get_index_prop (obj_p, index);
+
+    JERRY_ASSERT (ecma_is_value_number (accumulator));
+
+    if (is_right)
+    {
+      if (index == 0)
+      {
+        not_finish = false;
```
zherczeg: My point was to use break statement here, so the not_finish could be eliminated from the loop. There is no need for an else case as well.

jiangzidong: ok, I remove the not_finish in the loop, and change the var name "start_while" before while, which is to indicate whether the while loop should be skipped.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,179 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+                      ecma_op_function_call (func_object_p,
+                                             ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED),
+                                             call_args,
+                                             4),
+                      ret_value);
+
+      ecma_fast_free_value (accumulator);
+      accumulator = ecma_copy_value (call_value);
+
+      ECMA_FINALIZE (call_value);
```
zherczeg: Question: is it worth to optimize this by removing ECMA_TRY_CATCH and ecma_copy_value ?

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,179 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Callback function is not callable."));
+  }
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (this_arg);
+  uint32_t len = ecma_typedarray_get_length (obj_p);
+
+  if (len == 0 && ecma_is_value_undefined (initial_val))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Initial value cannot be undefined."));
+  }
```
zherczeg: I think it would make this code more robust if we would handle len == 0 here (return with exception or the accumulator), and can assume that len > 0 after this point.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,179 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+
+    accumulator = ecma_op_typedarray_get_index_prop (obj_p, index);
+
+    JERRY_ASSERT (ecma_is_value_number (accumulator));
+
+    if (is_right)
+    {
+      if (index == 0)
+      {
+        start_while = false;
```
zherczeg: what about return accumulator ?

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
jiangzidong: I'm not sure I totally understand your question. Do you mean sth like `accumulator = ecma_op_function_call(...)`
But the ecma_op_function_call might throw exception, because the callback_fn is provided by user. I think ECMA_TRY_CATCH  is needed here.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
jiangzidong: Yes, I can remove all `start_while`, and directly return the accumulator

jiangzidong:
@zherczeg I removed the flags to check inside/before the while loop. I still retained the ECMA_TRY_CATCH. About the question about "removing ECMA_TRY_CATCH and ecma_copy_value", could you please explain in more details.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,174 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+                    ecma_op_function_call (func_object_p,
+                                           ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED),
+                                           call_args,
+                                           4),
+                    ret_value);
+
+    ecma_fast_free_value (accumulator);
+    accumulator = ecma_copy_value (call_value);
+
+    ECMA_FINALIZE (call_value);
```
zherczeg: Yes you could:

    ecma_value_t ret_value = ecma_op_function_call (....)

    ecma_fast_free_value (get_value);
    ecma_fast_free_value (current_index);

    if (ECMA_IS_VALUE_ERROR (ret_value))
    {
       ecma_free_value (accumulator);
       return ret_value;
    }

while (ecma_is_value_empty (ret_value)) -> while(true)



This could probably reduce binary size.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,174 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+  }
+
+  if (ecma_is_value_empty (ret_value))
+  {
+    ret_value = ecma_copy_value (accumulator);
+  }
+
+  ecma_free_value (accumulator);
+
+  return ret_value;
```
zherczeg: from if (ecma_is_value_empty (ret_value)) ... could be replaced by

    return accumulator;

Since no error is possible here.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -355,6 +355,174 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
...
+
+  if (!ecma_op_is_callable (cb_func_val))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Callback function is not callable."));
+  }
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (this_arg);
+  uint32_t len = ecma_typedarray_get_length (obj_p);
+
+
```
zherczeg: Only one newline please.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
jiangzidong: Oh, I know what you mean. update soon

tests/jerry-test-suite/es2015/22/22.02/22.02.03/22.02.03-016.js
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var total = new Float32Array([-1.5, 0, 1.5, 2]).reduceRight(function(a, b) {
+  return a - b ;
```
zherczeg: One more thing: space before semicolon

PR #1573, opened 2017-02-09
Documentation update

LaszloLango:

### DIFF
--- a/docs/01.GETTING-STARTED.md
+++ b/docs/01.GETTING-STARTED.md
@@ -45,10 +45,10 @@ python tools/build.py --debug --lto=off
 python tools/build.py --cmake-param=CMAKE_PARAM
 ```
 
-**Set a profile mode (full, minimal)**
+**Set a profile mode (ES5.1, subset of ES2015, minimal)**
 
 ```bash
-python tools/build.py --feature=full|minimal
+python tools/build.py --profile=es5.1|es2015-subset|minimal
 ```
 
 **Use (jerry, compiler-default, external) libc**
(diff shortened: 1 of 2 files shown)


docs/01.GETTING-STARTED.md
```diff
@@ -81,6 +81,36 @@ For example the cross-compile to RaspberryPi 2 is something like this:
...
+```bash
+python tools/build.py --cpointer-32bit=on
+```
+
+*Note*: There is no compression/decompression on 32 bit systems, if enabled.
+
+**Change default heap size (512K)**
+
+```bash
+python tools/build.py --mem-heap=256K
```
bzsolt: Just `--mem-heap=256`, this option accepts integers.

docs/01.GETTING-STARTED.md
```diff
@@ -81,6 +81,36 @@ For example the cross-compile to RaspberryPi 2 is something like this:
...
+```
+
+If you would like to use more than 512K, then you must enable the 32 bit compressed pointers.
+
+```bash
+python tools/build.py --cpointer-32bit=on --mem-heap=1024K
```
bzsolt: `--mem-heap=1024`

docs/01.GETTING-STARTED.md
```diff
@@ -81,6 +81,36 @@ For example the cross-compile to RaspberryPi 2 is something like this:
 python tools/build.py --toolchain=cmake/toolchain_linux_armv7l.cmake
 ```
 
+**Use system memory allocator**
+
+```bash
+python tools/build.py --system-allocator=on --jerry-libc=off
```
bzsolt: I think it would be good to add `--compile-flag=-m32` to the command or add some hints about how to build to 32bit.

docs/01.GETTING-STARTED.md
```diff
@@ -81,6 +81,36 @@ For example the cross-compile to RaspberryPi 2 is something like this:
...
+python tools/build.py --mem-heap=256K
+```
+
+If you would like to use more than 512K, then you must enable the 32 bit compressed pointers.
+
+```bash
+python tools/build.py --cpointer-32bit=on --mem-heap=1024K
+```
+
+*Note*: Memory allocator of JerryScript will allocate the full heap, when the engine starts.
```
dbatyai: This isn't completely correct, as our memory allocator has no responsibility to allocate the heap. The heap will be allocated statically at compile time.

docs/02.API-REFERENCE.md
```diff
@@ -1316,10 +1316,13 @@ jerry_get_utf8_string_length (const jerry_value_t value);
 
 **Summary**
 
-Copy the characters of a string into a specified buffer. The
-'\0' character could occur in character buffer. Returns 0,
-if the value parameter is not a string or the buffer is
-not large enough for the whole string.
+Copy the characters of a string into a specified cesu-8 buffer.
+The '\0' character could occur in character buffer. Returns 0,
```
dbatyai: in _the_

docs/01.GETTING-STARTED.md
LaszloLango: @bzsolt, this is why I wrote a note. :)

LaszloLango:
@bzsolt, @dbatyai thanks for the reviews. I've updated the PR.

PR #1577, opened 2017-02-10
Improve error message

bsdelf:
Many error messages of JerryScript only tell there is an error but without further context, which is vague and makes it difficult to debug.

This patch helps the situtation, for errors related to unexisting properties or bad references, it can remind the property/variable names at least.

With this patch, for the following testing code (also works for `node`):

```javascript
'use strict';

var print = print || console.log;

var printError = (function (i) {
  return function (head, error) {
    print(i + (i <= 9 ? ' ' : ''), head, Array(30 - head.length).join(' '), error);
      print(Array(100).join('-'));
      ++i;
  };
})(1);

[ 'x = 1;',
  'x += 1;',
  'var y = x;',
  'x.y = 1;',
  'var x; x.y;',
  'var x; x[1];',
  'var x; x.y = 1;',
  'var x; x[0] = 1;',
  'var x = null; x[\'y\'];',
  'null.y = 1;'
].forEach(function (src) {
  try {
    eval(src);
  } catch (e) {
    printError(src, e);
  }
});
```

The expected output would be:

```
1  x = 1;                         ReferenceError: x is not defined
---------------------------------------------------------------------------------------------------
2  x += 1;                        ReferenceError: x is not defined
---------------------------------------------------------------------------------------------------
3  var y = x;                     ReferenceError: x is not defined
---------------------------------------------------------------------------------------------------
4  x.y = 1;                       ReferenceError: x is not defined
---------------------------------------------------------------------------------------------------
5  var x; x.y;                    TypeError: Cannot read property 'y' of undefined
---------------------------------------------------------------------------------------------------
6  var x; x[1];                   TypeError: Cannot read property '1' of undefined
---------------------------------------------------------------------------------------------------
7  var x; x.y = 1;                TypeError: Cannot set property 'y' of undefined
---------------------------------------------------------------------------------------------------
8  var x; x[0] = 1;               TypeError: Cannot set property '0' of undefined
---------------------------------------------------------------------------------------------------
9  var x = null; x['y'];          TypeError: Cannot read property 'y' of null
---------------------------------------------------------------------------------------------------
10 null.y = 1;                    TypeError: Cannot set property 'y' of null
---------------------------------------------------------------------------------------------------

```

### DIFF
--- a/jerry-core/ecma/operations/ecma-exceptions.h
+++ b/jerry-core/ecma/operations/ecma-exceptions.h
@@ -51,6 +51,9 @@ typedef enum
 ecma_object_t *ecma_new_standard_error (ecma_standard_error_t error_type);
 ecma_object_t *ecma_new_standard_error_with_message (ecma_standard_error_t error_type, ecma_string_t *message_string_p);
 ecma_value_t ecma_raise_standard_error (ecma_standard_error_t error_type, const lit_utf8_byte_t *msg_p);
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+ecma_value_t ecma_raise_standard_error_with_format (ecma_standard_error_t error_type, const char *msg_p, ...);
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
 ecma_value_t ecma_raise_common_error (const char *msg_p);
 ecma_value_t ecma_raise_eval_error (const char *msg_p);
 ecma_value_t ecma_raise_range_error (const char *msg_p);
--- a/jerry-core/ecma/operations/ecma-get-put-value.c
+++ b/jerry-core/ecma/operations/ecma-get-put-value.c
@@ -52,7 +52,15 @@ ecma_op_get_value_lex_env_base (ecma_object_t *ref_base_lex_env_p, /**< referenc
   /* 3. */
   if (unlikely (is_unresolvable_reference))
   {
-    return ecma_raise_reference_error (ECMA_ERR_MSG ("Cannot resolve reference."));
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+    ecma_value_t var_name_val = ecma_make_string_value (var_name_string_p);
+    ecma_value_t error_value = ecma_raise_standard_error_with_format (ECMA_ERROR_REFERENCE,
+                                                                      "% is not defined",
+                                                                      var_name_val);
+#else /* !JERRY_ENABLE_ERROR_MESSAGES */
+    ecma_value_t error_value = ecma_raise_reference_error (NULL);
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
+    return error_value;
   }
 
   /* 5. */
@@ -149,7 +157,15 @@ ecma_op_put_value_lex_env_base (ecma_object_t *ref_base_lex_env_p, /**< referenc
     /* 3.a. */
     if (is_strict)
     {
-      return ecma_raise_reference_error (ECMA_ERR_MSG ("Cannot resolve reference."));
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+      ecma_value_t var_name_val = ecma_make_string_value (var_name_string_p);
+      ecma_value_t error_value = ecma_raise_standard_error_with_format (ECMA_ERROR_REFERENCE,
+                                                                        "% is not defined",
+                                                                        var_name_val);
+#else /* !JERRY_ENABLE_ERROR_MESSAGES */
+      ecma_value_t error_value = ecma_raise_reference_error (NULL);
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
+      return error_value;
     }
     else
     {
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -96,7 +96,15 @@ vm_op_get_value (ecma_value_t object, /**< base object */
 
   if (unlikely (ecma_is_value_undefined (object) || ecma_is_value_null (object)))
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Base object cannot be null or undefined."));
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+    ecma_value_t error_value = ecma_raise_standard_error_with_format (ECMA_ERROR_TYPE,
+                                                                      "Cannot read property '%' of %",
+                                                                      property,
+                                                                      object);
+#else /* !JERRY_ENABLE_ERROR_MESSAGES */
+    ecma_value_t error_value = ecma_raise_type_error (NULL);
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
+    return error_value;
   }
 
   ecma_value_t prop_to_string_result = ecma_op_to_string (property);
@@ -136,8 +144,20 @@ vm_op_set_value (ecma_value_t object, /**< base object */
 
     if (ECMA_IS_VALUE_ERROR (to_object))
     {
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+      ecma_free_value (to_object);
+
+      ecma_value_t error_value = ecma_raise_standard_error_with_format (ECMA_ERROR_TYPE,
+                                                                        "Cannot set property '%' of %",
+                                                                        property,
+                                                                        object);
+      ecma_free_value (property);
+
+      return error_value;
+#else /* !JERRY_ENABLE_ERROR_MESSAGES */
       ecma_free_value (property);
       return to_object;
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
     }
 
     object = to_object;
@@ -1082,16 +1102,9 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
             ref_base_lex_env_p = ecma_op_resolve_reference_base (frame_ctx_p->lex_env_p,
                                                                  name_p);
 
-            if (ref_base_lex_env_p != NULL)
-            {
-              result = ecma_op_get_value_lex_env_base (ref_base_lex_env_p,
-                                                       name_p,
-                                                       is_strict);
-            }
-            else
-            {
-              result = ecma_raise_reference_error (ECMA_ERR_MSG ("Cannot resolve reference."));
-            }
+            result = ecma_op_get_value_lex_env_base (ref_base_lex_env_p,
+                                                     name_p,
+                                                     is_strict);
 
             if (ECMA_IS_VALUE_ERROR (result))
             {
--- a/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/jerry-core/ecma/operations/ecma-exceptions.c
@@ -13,7 +13,9 @@
  * limitations under the License.
  */
 
+#include <stdarg.h>
 #include "ecma-builtins.h"
+#include "ecma-conversion.h"
 #include "ecma-exceptions.h"
 #include "ecma-gc.h"
 #include "ecma-globals.h"
@@ -155,6 +157,95 @@ ecma_raise_standard_error (ecma_standard_error_t error_type, /**< error type */
 
   return ecma_make_error_obj_value (error_obj_p);
 } /* ecma_raise_standard_error */
+
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+
+/**
+ * Raise a standard ecma-error with the given format string and arguments.
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value
+ */
+ecma_value_t
+ecma_raise_standard_error_with_format (ecma_standard_error_t error_type, /**< error type */
+                                       const char *format, /**< format string */
+                                       ...) /**< ecma-values */
+{
+  JERRY_ASSERT (format != NULL);
+
+  ecma_string_t *error_msg_p = ecma_get_magic_string (LIT_MAGIC_STRING__EMPTY);
+  ecma_string_t *string1_p;
+  ecma_string_t *string2_p;
+
+  const char *start_p = format;
+  const char *end_p = format;
+
+  va_list args;
+
+  va_start (args, format);
+
+  while (*end_p)
+  {
+    if (*end_p == '%')
+    {
+      /* Concat template string. */
+      if (end_p > start_p)
+      {
+        string1_p = error_msg_p;
+        string2_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) start_p,
+                                                    (lit_utf8_size_t) (end_p - start_p));
+        error_msg_p = ecma_concat_ecma_strings (string1_p, string2_p);
+        ecma_deref_ecma_string (string1_p);
+        ecma_deref_ecma_string (string2_p);
+      }
+
+      /* Convert an argument to string without side effects. */
+      ecma_string_t *arg_string_p;
+      const ecma_value_t arg_val = va_arg (args, ecma_value_t);
+      if (unlikely (ecma_is_value_object (arg_val)))
+      {
+        ecma_object_t *arg_object_p = ecma_get_object_from_value (arg_val);
+        lit_magic_string_id_t class_name = ecma_object_get_class_name (arg_object_p);
+        arg_string_p = ecma_get_magic_string (class_name);
+      }
+      else
+      {
+        jerry_value_t str_val = ecma_op_to_string (arg_val);
+        arg_string_p = ecma_get_string_from_value (str_val);
+      }
+
+      /* Concat argument. */
+      string1_p = error_msg_p;
+      string2_p = arg_string_p;
+      error_msg_p = ecma_concat_ecma_strings (string1_p, string2_p);
+      ecma_deref_ecma_string (string1_p);
+      ecma_deref_ecma_string (string2_p);
+
+      start_p = end_p + 1;
+    }
+
+    end_p++;
+  }
+
+  va_end (args);
+
+  /* Concat reset of template string. */
+  if (start_p < end_p)
+  {
+    string1_p = error_msg_p;
+    string2_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) start_p,
+                                                (lit_utf8_size_t) (end_p - start_p));
+    error_msg_p = ecma_concat_ecma_strings (string1_p, string2_p);
+    ecma_deref_ecma_string (string1_p);
+    ecma_deref_ecma_string (string2_p);
+  }
+
+  ecma_object_t *error_obj_p = ecma_new_standard_error_with_message (error_type, error_msg_p);
+  ecma_deref_ecma_string (error_msg_p);
+  return ecma_make_error_obj_value (error_obj_p);
+} /* ecma_raise_standard_error_with_format */
+
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
 
 /**
  * Raise a common error with the given message.
--- a/jerry-core/ecma/operations/ecma-reference.c
+++ b/jerry-core/ecma/operations/ecma-reference.c
@@ -123,7 +123,15 @@ ecma_op_resolve_reference_value (ecma_object_t *lex_env_p, /**< starting lexical
     lex_env_p = ecma_get_lex_env_outer_reference (lex_env_p);
   }
 
-  return ecma_raise_reference_error (ECMA_ERR_MSG ("Cannot resolve reference."));
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+  ecma_value_t name_val = ecma_make_string_value (name_p);
+  ecma_value_t error_value = ecma_raise_standard_error_with_format (ECMA_ERROR_REFERENCE,
+                                                                    "% is not defined",
+                                                                    name_val);
+#else /* !JERRY_ENABLE_ERROR_MESSAGES */
+  ecma_value_t error_value = ecma_raise_reference_error (NULL);
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
+  return error_value;
 } /* ecma_op_resolve_reference_value */
 
 /**
--- a/jerry-core/parser/js/js-parser.c
+++ b/jerry-core/parser/js/js-parser.c
@@ -2345,13 +2345,6 @@ parser_send_breakpoints (parser_context_t *context_p, /**< context */
 
 #endif /* JERRY_DEBUGGER */
 
-#define PARSE_ERR_POS_START       " [line: "
-#define PARSE_ERR_POS_START_SIZE  ((uint32_t) sizeof (PARSE_ERR_POS_START) - 1)
-#define PARSE_ERR_POS_MIDDLE      ", column: "
-#define PARSE_ERR_POS_MIDDLE_SIZE ((uint32_t) sizeof (PARSE_ERR_POS_MIDDLE) - 1)
-#define PARSE_ERR_POS_END         "]"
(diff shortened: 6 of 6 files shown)


jerry-core/ecma/operations/ecma-exceptions.h
```diff
@@ -59,6 +59,16 @@ ecma_value_t ecma_raise_syntax_error (const char *msg_p);
 ecma_value_t ecma_raise_type_error (const char *msg_p);
 ecma_value_t ecma_raise_uri_error (const char *msg_p);
 
+#define ECMA_DEFINE_ERROR_VALUE(var_name, error_type, msg_fmt, ...) \
+  ecma_value_t var_name = ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY); \
+  do { \
+    const int msg_size = snprintf (NULL, 0, msg_fmt, __VA_ARGS__) + 1; \
+    JMEM_DEFINE_LOCAL_ARRAY (msg_p, msg_size, char); \
+    snprintf (msg_p, (size_t) msg_size, msg_fmt, __VA_ARGS__); \
```
zherczeg: Does jerry libc has snprintf?

zherczeg:
In general the focus of the JerryScript project is low memory consumption and small binary size, so we don't give too much error context by default. Hence we need a compile time flag to enable this feature.

Additional note: we are working on a debugger in #1557, that will hopefully improves debugging experience.

jerry-core/vm/vm.c
```diff
@@ -96,7 +96,16 @@ vm_op_get_value (ecma_value_t object, /**< base object */
 
   if (unlikely (ecma_is_value_undefined (object) || ecma_is_value_null (object)))
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Base object cannot be null or undefined."));
+    ecma_string_t *property_name_p = ecma_get_string_from_value (property);
+    ECMA_STRING_TO_UTF8_STRING (property_name_p, property_name_utf8_p, property_name_size);
```
dbatyai: Personally, I would prefer if the messages were created using the string helper functions found in `ecma-helpers-string.c`, instead of using intermediate buffers and snprintf, if it's possible. Maybe like defining a few magic strings, and using `ecma_concat_ecma_strings` to create the final message.

dbatyai:
The `JERRY_ENABLE_ERROR_MESSAGES` already exists and would be suitable here as well, so no need for a new one.

jerry-core/vm/vm.c
zherczeg: It would be use create a helper function than a helper macro.

bsdelf: Indeed, use `ecma_concat_ecma_strings` to handle string is better than use plain C functions, but the code might be tedious to compose messages like "Cannot read property '%s' of %s". Would you have a suggestion?

zherczeg: Yes, please create a new helper function which can do the concatenation.

bsdelf:
Check about the updated patch.

The newly introduced helper function is:

```c
ecma_string_t *ecma_join_ecma_strings (ecma_string_t **string_array, uint32_t n);
```

I'm not sure whether it is a good idea to let the join function dereference the input ecma-strings.

Otherwise, we might need another helper function:

 ```c
void ecma_deref_ecma_strings (ecma_string_t **string_array, uint32_t n);
```

File: jerry-core/ecma/operations/ecma-exceptions.h [hunk 3]
```diff
@@ -59,6 +59,13 @@ ecma_value_t ecma_raise_syntax_error (const char *msg_p);
 ecma_value_t ecma_raise_type_error (const char *msg_p);
 ecma_value_t ecma_raise_uri_error (const char *msg_p);
 
+#define ECMA_DEFINE_ERROR_VALUE(var_name, error_type, error_msg_p) \
+  ecma_value_t var_name = ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY); \
+  do { \
+    ecma_object_t *error_obj_p = ecma_new_standard_error_with_message (error_type, error_msg_p); \
+    var_name = ecma_make_error_obj_value (error_obj_p); \
+  } while (0)
```
dbatyai: Could you please remove this macro? It doesn't seem to do much anymore, outside of replacing one line for another.

dbatyai:
In my opinion it's completely fine for `ecma_join_ecma_strings` to dereference the argument strings.

zherczeg:
I agree, binary size matters.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -699,6 +699,31 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
 } /* ecma_concat_ecma_strings */
 
 /**
+ * Join an array of ecma-strings and release them.
+ *
+ * @return a newly joined ecma-string
```
zherczeg: newly created

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -699,6 +699,31 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+ */
+ecma_string_t *
+ecma_join_ecma_strings (ecma_string_t **string_array, uint32_t n)
+{
+  JERRY_ASSERT (string_array != NULL);
+  JERRY_ASSERT (n > 0);
+
+  uint32_t i = 0;
+
+  for (; i < n - 1; ++i)
```
zherczeg: for (uint32_t i = 0; ...)

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -699,6 +699,31 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
 /**
+ * Join an array of ecma-strings and release them.
+ *
+ * @return a newly joined ecma-string
+ */
+ecma_string_t *
+ecma_join_ecma_strings (ecma_string_t **string_array, uint32_t n)
+{
+  JERRY_ASSERT (string_array != NULL);
+  JERRY_ASSERT (n > 0);
```
zherczeg: What about n > 1 ?

jerry-core/ecma/base/ecma-helpers-string.c
zherczeg: i++

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -699,6 +699,31 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+
+  uint32_t i = 0;
+
+  for (; i < n - 1; ++i)
+  {
+    ecma_string_t *string1_p = string_array[i];
+    ecma_string_t *string2_p = string_array[i + 1];
+    string_array[i + 1] = ecma_concat_ecma_strings (string1_p, string2_p);
+    ecma_deref_ecma_string (string1_p);
+    ecma_deref_ecma_string (string2_p);
```
zherczeg: You don't need to overwrite the string array, just keep the final string in a local variable.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -699,6 +699,31 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
 } /* ecma_concat_ecma_strings */
 
 /**
+ * Join an array of ecma-strings and release them.
+ *
+ * @return a newly joined ecma-string
+ */
+ecma_string_t *
+ecma_join_ecma_strings (ecma_string_t **string_array, uint32_t n)
```
zherczeg: Should be const array.

jerry-core/ecma/operations/ecma-get-put-value.c
```diff
@@ -52,7 +52,22 @@ ecma_op_get_value_lex_env_base (ecma_object_t *ref_base_lex_env_p, /**< referenc
...
+    ecma_string_t *string_array[] =
+    {
+      var_name_string_p,
+      ecma_get_magic_string (LIT_MAGIC_STRING_SPACE_CHAR),
+      ecma_get_magic_string (LIT_MAGIC_STRING_ERROR_IS_NOT_DEFINED),
+    };
+
+    ecma_string_t *msg_str_p = ecma_join_ecma_strings (string_array, JERRY_NUM_OF_ELEMENTS (string_array));
+    ECMA_DEFINE_ERROR_VALUE (error_value, ECMA_ERROR_REFERENCE, msg_str_p);
+    ecma_deref_ecma_string (msg_str_p);
```
zherczeg: If we use this utility solely for creating errors, we could create the error object in the utility to reduce binary size.

jerry-core/lit/lit-magic-strings.inc.h
```diff
@@ -27,6 +27,7 @@ LIT_MAGIC_STRING_FIRST_STRING_WITH_SIZE (1, LIT_MAGIC_STRING_NEW_LINE_CHAR)
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_NEW_LINE_CHAR, "\n")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_SPACE_CHAR, " ")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_DOUBLE_QUOTE_CHAR,"\"")
+LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_SINGLE_QUOTE_CHAR,"'")
```
zherczeg: I don't think it is a good idea to add magic strings for error handling, but I am curious about others opinion.

bsdelf: What about use `JERRY_ENABLE_ERROR_MESSAGES` macro to enable specific magic strings at compile time?

zherczeg: My problem is conceptual, no error messages are magic string at this point. Magic strings cover basic JS functionality. @LaszloLango what do you think?

LaszloLango: I thinks the concept is that frequently used strings should be magic strings. Increasing the number of magic strings, when the error messages are disabled is not good. I'm OK with using frequent error messages as magic string if `JERRY_ENABLE_ERROR_MESSAGES` is enabled.

zherczeg: Errors are usually short living objects. Magic strings are designed for property names which likely present for longer times. But having more magic strings does not hurt either.

Ok, we need a system which allows conditional magic string definition. There is no point adding error message strings when they never used. It would just increase binary size.

LaszloLango: > Errors are usually short living objects.

Good point. Both solutions are good to me (remove or guard).

> There is no point adding error message strings when they never used. It would just increase binary size.

I agree.

bsdelf: Tunes out introduce a format function to compose error message can make things easier:

```c
ecma_format_ecma_string ("Cannot read property '%' of %", prop_str_p, name_str_p)
```

- No need to involve magic string.
- The format string can be easily found with `grep`.
- Also, with this util [parser_parse_script](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/parser/js/js-parser.c#L2391) can be simplified.

@zherczeg

zherczeg: Excellent idea.

LaszloLango: Sounds great.

bsdelf:
Format function:

```c
ecma_string_t * ecma_format_ecma_strings (const char *format, ...);
```

Supported placeholder:
- `%`: ecma-string.
- `%%`: single percent sign.

Pitfalls:
- `ecma_format_ecma_strings("%%", str_prefix_p, str_p)`
- `ecma_format_ecma_strings("%%%", str_surrounded_by_percent_p)`

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -709,6 +710,79 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+                                                (lit_utf8_size_t) (end_p - start_p));
+    string_p = ecma_concat_ecma_strings (string1_p, string2_p);
+    ecma_deref_ecma_string (string1_p);
+    ecma_deref_ecma_string (string2_p);
+  }
+
+  va_end (args);
+
+  return string_p;
+} /* ecma_format_ecma_strings */
```
zherczeg: We should add a JERRY_ENABLE_ERROR_MESSAGES macro around this function.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -709,6 +710,79 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
 } /* ecma_concat_ecma_strings */
 
 /**
+ * Format ecma-strings.
+ *
+ * @return a newly created ecma-string
+ */
+ecma_string_t *
+ecma_format_ecma_strings (const char * format, /**< format string */
+                          ...) /**< ecma-strings */
```
zherczeg: I never particularly like `...`, for me it looks like a crude hack. I would prefer a constant array of ecma values. I am curious about others opinion.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -709,6 +710,79 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+      {
+        end_p++;
+
+        string1_p = string_p;
+        string2_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) start_p,
+                                                    (lit_utf8_size_t) (end_p - start_p));
+        string_p = ecma_concat_ecma_strings (string1_p, string2_p);
+        ecma_deref_ecma_string (string1_p);
+        ecma_deref_ecma_string (string2_p);
+      }
```
zherczeg: Do we need to support double `%` sign? This function is not a general printf, we probably don't need to overcomplicate things.

jerry-core/ecma/base/ecma-helpers-string.c
LaszloLango: I have no problems with it. :)

jerry-core/ecma/base/ecma-helpers-string.c
bsdelf: That would be best.

jerry-core/ecma/base/ecma-helpers-string.c
bsdelf: Variadic function is more readable for caller, and numbers of placeholders/ arguments can be easily checked.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -708,6 +709,69 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+
+/**
+ * Format ecma-strings.
+ *
+ * @return a newly created ecma-string
+ */
+ecma_string_t *
+ecma_format_ecma_strings (const char *format, /**< format string */
+                          ...) /**< ecma-strings */
```
zherczeg: ecma values

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -708,6 +709,69 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+  while (*end_p)
+  {
+    if (*end_p == '%')
+    {
+      string1_p = string_p;
+      string2_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) start_p,
+                                                  (lit_utf8_size_t) (end_p - start_p));
+      string_p = ecma_concat_ecma_strings (string1_p, string2_p);
+      ecma_deref_ecma_string (string1_p);
+      ecma_deref_ecma_string (string2_p);
```
zherczeg: We could skip this if end_p == start_p

jerry-core/vm/vm.c
```diff
@@ -96,7 +96,24 @@ vm_op_get_value (ecma_value_t object, /**< base object */
 
   if (unlikely (ecma_is_value_undefined (object) || ecma_is_value_null (object)))
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Base object cannot be null or undefined."));
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+    ecma_value_t obj_to_string_result = ecma_op_to_string (object);
+    ecma_string_t *obj_name_p = ecma_get_string_from_value (obj_to_string_result);
```
zherczeg: The ecma_op_to_string can have side effects (can run a JS code) and might not be successful. The side effects worries me most, since we might break ES5 compatibility. What about returning with the class of an object? That is always successful.

jerry-core/vm/vm.c
```diff
@@ -96,7 +96,24 @@ vm_op_get_value (ecma_value_t object, /**< base object */
...
+
+    ecma_value_t prop_to_string_result = ecma_op_to_string (property);
```
zherczeg: Again, if property is an object, we should limit this conversion.

zherczeg:
Perhaps we could just pass all ecma values to the utility function and let it do the string conversion if needed. Objects could simply be represented with their class.

bsdelf:
Patch updated.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -172,6 +172,7 @@ ecma_string_t *ecma_new_ecma_string_from_magic_string_id (lit_magic_string_id_t
 ecma_string_t *ecma_new_ecma_string_from_magic_string_ex_id (lit_magic_string_ex_id_t id);
 ecma_string_t *ecma_new_ecma_length_string (void);
 ecma_string_t *ecma_concat_ecma_strings (ecma_string_t *string1_p, ecma_string_t *string2_p);
+ecma_string_t *ecma_format_ecma_values (const char *format, ...);
```
LaszloLango:
```c
#ifdef JERRY_ENABLE_ERROR_MESSAGES
ecma_string_t *ecma_format_ecma_values (const char *format, ...);
#endif /* JERRY_ENABLE_ERROR_MESSAGES */
```

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -708,6 +710,133 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+        ecma_object_t *arg_object_p = ecma_get_object_from_value (arg_val);
+        lit_magic_string_id_t class_name = ecma_object_get_class_name (arg_object_p);
+        arg_string_p = ecma_get_magic_string (class_name);
+      }
+      else
+      {
+        if (ecma_is_value_string (arg_val))
+        {
+          arg_string_p = ecma_get_string_from_value (arg_val);
+          ecma_ref_ecma_string (arg_string_p);
```
LaszloLango: Why do we need to increase the reference here?

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -708,6 +710,133 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+      {
+        string1_p = string_p;
+        string2_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) start_p,
+            (lit_utf8_size_t) (end_p - start_p));
+        string_p = ecma_concat_ecma_strings (string1_p, string2_p);
+        ecma_deref_ecma_string (string1_p);
+        ecma_deref_ecma_string (string2_p);
+      }
+
+      /* Convert an argument to string without side effect. */
```
zherczeg: effects.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -708,6 +710,133 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+      const ecma_value_t arg_val = va_arg (args, ecma_value_t);
+      if (unlikely (ecma_is_value_object (arg_val)))
+      {
+        ecma_object_t *arg_object_p = ecma_get_object_from_value (arg_val);
+        lit_magic_string_id_t class_name = ecma_object_get_class_name (arg_object_p);
+        arg_string_p = ecma_get_magic_string (class_name);
+      }
+      else
+      {
+        if (ecma_is_value_string (arg_val))
```
zherczeg: I think we can simply use the to_string in all other cases since non-object types can be converted to string without side effects.

jerry-core/vm/vm.c
```diff
@@ -136,8 +144,20 @@ vm_op_set_value (ecma_value_t object, /**< base object */
 
     if (ECMA_IS_VALUE_ERROR (to_object))
     {
+      ecma_free_value (to_object);
+
+#ifdef JERRY_ENABLE_ERROR_MESSAGES
+      ecma_string_t *error_msg_p = ecma_format_ecma_values ("Cannot set property '%' of %", property, object);
+      ecma_object_t *error_obj_p = ecma_new_standard_error_with_message (ECMA_ERROR_TYPE, error_msg_p);
+      ecma_value_t error_value = ecma_make_error_obj_value (error_obj_p);
+      ecma_deref_ecma_string (error_msg_p);
```
zherczeg: These four lines are repeated all the time. I would suggest to add the error type as a first argument for ecma_format_ecma_values and create the error there to reduce the binary size. Or introduce another helper. Perhaps the name could be ecma_new_standard_error_with_format.

jerry-core/ecma/base/ecma-helpers-string.c
bsdelf: Because `ecma_get_magic_string` will increase the reference count, while `ecma_get_string_from_value` won't. To make algorithm consistent, we increase reference count for the later case, and dereference them after concatation.

LaszloLango: I see, thanks for the explanation.

jerry-core/vm/vm.c
bsdelf: I prefer the latter option.

bsdelf: However, if we add a helper function in "ecma-exceptions.c" to raise ecma-error, why not merge `ecma_format_ecma_values` into that function?

zherczeg: This was option 1, or I misunderstand something.

bsdelf: I see...

bsdelf:
Patch updated. @zherczeg @LaszloLango

jerry-core/vm/vm.c
```diff
@@ -136,8 +144,20 @@ vm_op_set_value (ecma_value_t object, /**< base object */
...
+                                                                        property,
+                                                                        object);
+#else /* !JERRY_ENABLE_ERROR_MESSAGES */
+      ecma_value_t error_value = ecma_raise_type_error (NULL);
+#endif /* JERRY_ENABLE_ERROR_MESSAGES */
+
       ecma_free_value (property);
-      return to_object;
+
+      return error_value;
```
zherczeg: I see the point of recreating the error when messages are enabled, but not when they are disabled. Please update this ifdef to not change the original behavior.

bsdelf:
Partially reverted. @zherczeg @LaszloLango

Issue #1297, opened 2016-08-29
Change name of public API header?


akosthekiss:
The 1.0 release is approaching, but before it happens, we might reconsider the name of the public API header of the JerryScript library. Now, it is called jerry-api.h, which is somewhat unusual. For an ordinary library user, writing `#include <jerry.h>` might be more intuitive.

The time of the release is close and the change might need adaptation at various places of the project, but we should think about it since whatever we choose will most probably stay with us for good (or, for a long time, at least), as the name of the header is also part of the fixed API.

(Note: I'm not proposing to change the name of jerry-port.h. That's OK, IMHO, as it is important for and used by port developers only.)

zherczeg:
I don't mind to rename the header. I think the question is how risky this change is.

LaszloLango:
I think it is not the right time to change the header name. Such change implies an RC3. It will delay the release if we rename it.

tilmannOSG:
In terms of release I think it's too late for an invasive change like that, we should only cherry-pick bug fixes for critical bugs which would block the release for one reason or another :)  
In terms of header naming I think "jerryscript.h" would be even more appropriate since after all the project is called JerryScript not Jerry :)

akosthekiss:
OK, let's wait with renamings then. (BTW: calling the resulting library libjerry-core.a is also a bit unnatural, libjerry.a or libjerryscript.a would seem more clean.)

Removing the issue from the 1.0 release milestone but leaving it open as a reminder that we might still consider something like this in the future. (Feel free to close it if you disagree, though.)

tilmannOSG:
Yes, I'd very much in favor of changing the header name to "jerryscript.h" and the library name to "libjerryscript.a" in the future :)  
I think we should consider this for the next stable release after 1.0.

PR #1579, opened 2017-02-13
Add jerryscript.h and mark jerry-api.h as deprecated

galpeter:
Add a notification for JerryScript API users that the jerry-api.h
will be removed in the future and should use the jerry.h header
instead.

Also update the examples in the docs and the targets where
the jerry-api.h is used.


jerry-core/jerry-api.h
```diff
@@ -16,6 +16,10 @@
 #ifndef JERRY_API_H
 #define JERRY_API_H
 
+#ifndef JERRY_H
+#pragma message("using jerry-api.h directly is deprecated, please use jerry.h")
+#endif
+
```
zherczeg: Why don't we rename jerry-api.h to jerry.h?

zherczeg:
#1297

Please rename it to jerryscript.h

galpeter:
@zherczeg  I've updated the PR

PR #1585, opened 2017-02-16
[ES2015]add typedarray routine: reverse

jiangzidong:

### DIFF
--- a/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
+++ b/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
@@ -22,6 +22,7 @@
 #include "ecma-conversion.h"
 #include "ecma-function-object.h"
 #include "ecma-typedarray-object.h"
+#include "ecma-arraybuffer-object.h"
 #include "ecma-try-catch-macro.h"
 #include "jrt.h"
 #include "jrt-libc-includes.h"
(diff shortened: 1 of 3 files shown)


zherczeg:
I have been thinking about it before, but since the type of the input and output are the same, can't we copy data with memcpy? Conversions are slow. This could also be used by the filter algorithm (to reduce memory consumption).

jiangzidong:
It makes sense. I will try it. @zherczeg

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -609,6 +610,49 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
...
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Argument 'this' is not a TypedArray."));
+  }
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (this_arg);
+  uint32_t len = ecma_typedarray_get_length (obj_p);
+  uint32_t middle = len / 2;
+  ecma_object_t *arraybuffer_p = ecma_typedarray_get_arraybuffer (obj_p);
+  lit_utf8_byte_t *buffer = ecma_arraybuffer_get_buffer (arraybuffer_p);
+  ecma_length_t offset = ecma_typedarray_get_offset (obj_p);
```
zherczeg: What about adding offset to buffer? Then there is no need for +offset.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -609,6 +610,49 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
...
+  uint8_t shift = ecma_typedarray_get_element_size_shift (obj_p);
+  uint8_t element_size = (uint8_t) (1 << shift);
+
+  for (uint32_t lower = 0; lower < middle; lower++)
+  {
+    uint32_t upper = len - lower - 1;
+    ecma_length_t lower_pos = (lower << shift) + offset;
+    ecma_length_t upper_pos = (upper << shift) + offset;
+    lit_utf8_byte_t *lower_p = buffer + lower_pos;
```
zherczeg: lit_utf8_byte_t *lower_p = buffer + (lower << shift)

You can also make lower and upper byte oriented:

    middle <<= shift
    for (uint32_t lower = 0; lower < middle; lower += element_size)

Shift with variable right-value is never efficient, especially on x86.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -609,6 +610,48 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
...
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (this_arg);
+  uint32_t len = ecma_typedarray_get_length (obj_p);
+  ecma_object_t *arraybuffer_p = ecma_typedarray_get_arraybuffer (obj_p);
+  lit_utf8_byte_t *buffer = (ecma_arraybuffer_get_buffer (arraybuffer_p)
+                             + ecma_typedarray_get_offset (obj_p));
+  uint8_t shift = ecma_typedarray_get_element_size_shift (obj_p);
+  uint8_t element_size = (uint8_t) (1 << shift);
+  uint32_t middle = (len / 2) << shift;
+  uint32_t buffer_length = len << shift;
```
zherczeg: we could rename buffer_length to buffer_last and:

    uint32_t buffer_last = (len << shift) - element_size

    uint32_t upper = buffer_last - lower

PR #1588, opened 2017-02-16
Add eval support for JerryScript debugger

zherczeg:

PR #1592, opened 2017-02-17
Fix for python debugger client

rtakacs:
Clear the 'result' buffer when its length is equal to the 'expected' message length.


rtakacs:
Closing since this fix is already contained by #1588

PR #1601, opened 2017-02-21
Remove shadowed declarations, undefined identifiers, and specify argument types where it is required.

robertsipka:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -783,10 +783,10 @@ ecma_builtin_array_prototype_object_shift (ecma_value_t this_arg) /**< this argu
     if (ecma_is_value_empty (ret_value))
     {
       len--;
-      ecma_string_t *index_str_p = ecma_new_ecma_string_from_uint32 (len);
+      ecma_string_t *len_str_p = ecma_new_ecma_string_from_uint32 (len);
 
       /* 8. */
-      ECMA_TRY_CATCH (del_value, ecma_op_object_delete (obj_p, index_str_p, true), ret_value);
+      ECMA_TRY_CATCH (del_value, ecma_op_object_delete (obj_p, len_str_p, true), ret_value);
 
       /* 9. */
       ECMA_TRY_CATCH (set_length_value,
@@ -797,7 +797,7 @@ ecma_builtin_array_prototype_object_shift (ecma_value_t this_arg) /**< this argu
 
       ECMA_FINALIZE (set_length_value);
       ECMA_FINALIZE (del_value);
-      ecma_deref_ecma_string (index_str_p);
+      ecma_deref_ecma_string (len_str_p);
     }
 
     ECMA_FINALIZE (first_value);
@@ -1426,11 +1426,9 @@ ecma_builtin_array_prototype_object_splice (ecma_value_t this_arg, /**< this arg
   }
 
   /* 8-9. */
-  uint32_t k;
+  uint32_t k = 0;
 
-  for (uint32_t del_item_idx, k = 0;
-       k < delete_count && ecma_is_value_empty (ret_value);
-       k++)
+  for (uint32_t del_item_idx; k < delete_count && ecma_is_value_empty (ret_value); k++)
   {
     /* 9.a */
     del_item_idx = k + start;
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date.c
@@ -319,11 +319,11 @@ ecma_builtin_date_parse (ecma_value_t this_arg, /**< this argument */
       else if (date_str_curr_p < date_str_end_p
                && (*date_str_curr_p == '+' || *date_str_curr_p == '-'))
       {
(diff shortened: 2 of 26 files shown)


jerry-main/main-unix.c
```diff
@@ -627,7 +627,7 @@ main (int argc,
 
   if (jerry_is_feature_enabled (JERRY_FEATURE_SNAPSHOT_EXEC))
   {
-    for (int i = 0; i < exec_snapshots_count; i++)
+    for (i = 0; i < exec_snapshots_count; i++)
```
LaszloLango: I thinks won't be right. You will override the `i` of the outer `for` statement here. This is unsafe and might lead to infinite loop. Please rename the variable instead, eg.: `for (int j = 0; j < exec_snapshots_count; j++)`

jerry-main/main-unix.c
```diff
@@ -652,7 +652,7 @@ main (int argc,
 
   if (!jerry_value_has_error_flag (ret_value))
   {
-    for (int i = 0; i < files_counter; i++)
```
LaszloLango: ditto

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -783,10 +783,10 @@ ecma_builtin_array_prototype_object_shift (ecma_value_t this_arg) /**< this argu
     if (ecma_is_value_empty (ret_value))
     {
       len--;
-      ecma_string_t *index_str_p = ecma_new_ecma_string_from_uint32 (len);
+      ecma_string_t *idx_str_p = ecma_new_ecma_string_from_uint32 (len);
```
zherczeg: This seems to be a len_str_p

jerry-core/ecma/builtin-objects/ecma-builtin-date.c
```diff
@@ -319,7 +319,6 @@ ecma_builtin_date_parse (ecma_value_t this_arg, /**< this argument */
       else if (date_str_curr_p < date_str_end_p
                && (*date_str_curr_p == '+' || *date_str_curr_p == '-'))
       {
-        ecma_length_t remaining_length;
         remaining_length = lit_utf8_string_length (date_str_curr_p,
```
zherczeg: This should have a different name.

jerry-main/main-unix.c
robertsipka: That for loop ends before this statement, I've just reuse the `i` variable here.

galpeter: Can't we just move the top level `i` into the required `for`'s init list?

robertsipka:
Thanks, I've updated this patch based on your suggestions.

jerry-core/ecma/builtin-objects/ecma-builtin-json.c
```diff
@@ -934,11 +934,11 @@ ecma_builtin_json_stringify (ecma_value_t this_arg, /**< 'this' argument */
                                    array_length,
                                    ret_value);
 
-      uint32_t array_length = ecma_number_to_uint32 (array_length_num);
+      uint32_t conv_array_length_num = ecma_number_to_uint32 (array_length_num);
       uint32_t index = 0;
 
       /* 4.b.ii */
-      while ((index < array_length) && ecma_is_value_empty (ret_value))
+      while ((index < conv_array_length_num) && ecma_is_value_empty (ret_value))
```
zherczeg: If this is used only once, we could put the ecma_number_to_uint32 (array_length_num) here.

jerry-main/main-unix.c
```diff
@@ -744,7 +743,7 @@ main (int argc,
     const char *prompt = !no_prompt ? "jerry> " : "";
     bool is_done = false;
 
-    jerry_value_t global_obj_val = jerry_get_global_object ();
+    global_obj_val = jerry_get_global_object ();
```
zherczeg: I think this value is initialized to jerry_get_global_object (). Is there no reference counting problem here?

jerry-core/vm/vm.c
```diff
@@ -753,9 +753,9 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
           {
             case VM_OC_GET_LITERAL_LITERAL:
             {
-              uint16_t literal_index;
-              READ_LITERAL_INDEX (literal_index);
-              READ_LITERAL (literal_index, right_value);
+              uint16_t literal_literal_index;
```
zherczeg: second_literal_index

jerry-core/parser/js/js-parser-util.c
```diff
@@ -192,9 +192,9 @@ parser_flush_cbc (parser_context_t *context_p) /**< context */
       {
         literal_index = context_p->last_cbc.third_literal_index;
 
-        lexer_literal_t *literal_p = PARSER_GET_LITERAL (literal_index);
+        lexer_literal_t *third_literal_p = PARSER_GET_LITERAL (literal_index);
```
zherczeg: We should reuse literal_p since literal_index is also reused.

jerry-main/main-unix.c
robertsipka: It's freed earlier, but I've renamed it to avoid the misleading.

PR #1603, opened 2017-02-21
Completely remove internal property support.

zherczeg:
Small binary size reduction: 137740 -> 137620
No other effect.

### DIFF
--- a/jerry-core/lit/lit-magic-strings.h
+++ b/jerry-core/lit/lit-magic-strings.h
@@ -34,6 +34,12 @@ typedef enum
 #include "lit-magic-strings.inc.h"
 #undef LIT_MAGIC_STRING_DEF
 #undef LIT_MAGIC_STRING_FIRST_STRING_WITH_SIZE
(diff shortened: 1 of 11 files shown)


jerry-core/lit/lit-magic-strings.h
```diff
@@ -34,6 +34,11 @@ typedef enum
 #include "lit-magic-strings.inc.h"
 #undef LIT_MAGIC_STRING_DEF
 #undef LIT_MAGIC_STRING_FIRST_STRING_WITH_SIZE
+  LIT_REAL_MAGIC_STRING__COUNT, /**< number of real magic strings */
```
LaszloLango: Not a real issue, but the phase "real magic strings" is ambiguous to me. I'd add more description to this or rename to something else, like `common`, `general` or ` standard magic strings`

zherczeg: Renamed to LIT_NON_INTERNAL_MAGIC_STRING__COUNT and INTERNAL prefix is added to the new strings as well.

yichoi:
LGTM

PR #1608, opened 2017-02-22
Fix pylint install on Travis CI

LaszloLango:
Followup fix after #1602


.travis.yml
```diff
@@ -49,6 +49,13 @@ before_install:
   - if [[ "$TRAVIS_OS_NAME" == "linux" ]] && [[ "$INSTALL_QEMU_ARM" == "yes" ]]; then tools/apt-get-install-qemu-arm.sh; fi
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then tools/brew-install-deps.sh; fi
   - if [[ "$OPTS" == "--test262" ]]; then sudo timedatectl set-timezone America/Los_Angeles; fi
+  - if [[ "$OPTS" == *"--check-pylint"* ]];
+    then
+      git clone https://github.com/PyCQA/pylint.git -b pylint-1.6.5 "$PWD/third-party/pylint";
```
galpeter: What if we do the following? ` pip install pylint==1.6.5`
Then yo don't need to manually call clone/setup.

LaszloLango: I'll check it. Thanks for the tip.

LaszloLango: It works as expected and much simpler, so I've updated the PR. Please check.

galpeter:
lgtm

PR #1612, opened 2017-02-23
Rework address setup in WebSocket Debugger Client

robertsipka:


jerry-debugger/jerry-client-ws.html
```diff
@@ -972,15 +972,25 @@
...
+
+    if (ipAddr == "")
     {
-      appendLog("IP address expected");
-      return true;
+      ipAddr = "localhost";
     }
 
-    appendLog("Connect to: " + args[2]);
+    if (ipAddr.match(":")) {
```
zherczeg: Can we use a regex here? Something like:

    /^.*:\d+$/

If match requires whole string match then ^ and $ is not needed.

robertsipka: Good idea, I've changed it.

jerry-debugger/jerry-client-ws.html
```diff
@@ -972,15 +972,25 @@
...
+
+    if (ipAddr == "")
     {
-      appendLog("IP address expected");
-      return true;
+      ipAddr = "localhost";
     }
 
-    appendLog("Connect to: " + args[2]);
+    if (ipAddr.match(/.*:\d/)) {
```
zherczeg: Put { into the next line.

robertsipka: Fixed

Issue #1613, opened 2017-02-23
Assertion 'ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_GLOBAL)' failed in ecma_object_get_class_name


renatahodovan:
###### Jerry version:

```
Checked revision: 62a581f
Build: debug.linux
```
###### OS:

```
Ubuntu 16.04.2 LTS
```
###### Test case:

```javascript
new Int16Array(Float32Array.prototype)
```
###### Build command:
```
tools/build.py --clean --debug --profile=es2015-subset --compile-flag=-m32 --system-allocator=on --jerry-libc=off
```

###### Backtrace:

```
ICE: Assertion 'ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_GLOBAL)' failed at jerryscript/jerry-core/ecma/operations/ecma-objects.c(ecma_object_get_class_name):1762.
Error: ERR_FAILED_INTERNAL_ASSERTION
bt
#0  0xf7ffdbe9 in __kernel_vsyscall ()
#1  0x080f21d7 in raise ()
#2  0x080b9d17 in abort ()
#3  0x08057732 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/targets/default/jerry-port-default-fatal.c:53
#4  0x0807076b in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:61
#5  0x0807079e in jerry_assert_fail (assertion=0x8139e20 "ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_GLOBAL)", file=0x8139780 "jerryscript/jerry-core/ecma/operations/ecma-objects.c", function=0x812c5c4 <__func__.4429.lto_priv.425> "ecma_object_get_class_name", line=1762) at jerryscript/jerry-core/jrt/jrt-fatals.c:85
#6  0x0807b98d in ecma_object_get_class_name (obj_p=0x837e958) at jerryscript/jerry-core/ecma/operations/ecma-objects.c:1762
#7  0x08076237 in ecma_is_arraybuffer (target=137881947) at jerryscript/jerry-core/ecma/operations/ecma-arraybuffer-object.c:146
[...]

PR #1617, opened 2017-02-24
Fix issue #1613

jiangzidong:
I forget to implement the get_class_name_method of arraybuffer.prototype and typedarray.prototype.

For now, we don't need to distinguish their class name from general object, so I just set their class name as "object". In the future, when we need to know the specific class name, we can add some magic lit and modify them.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -1667,6 +1667,38 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
 } /* ecma_op_object_get_property_names */
 
 /**
+ * The function is used in the assert of ecma_object_get_class_name
+ */
+inline static bool
+ecma_object_check_class_name_is_object (ecma_object_t *obj_p) /**< object */
+{
+#ifndef JERRY_NDEBUG
+  return (ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_GLOBAL)
+#ifndef CONFIG_DISABLE_ARRAYBUFFER_BUILTIN
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_ARRAYBUFFER_PROTOTYPE)
+#endif /* !CONFIG_DISABLE_ARRAYBUFFER_BUILTIN */
+#ifndef CONFIG_DISABLE_TYPEDARRAY_BUILTIN
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_TYPEDARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_INT8ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT8ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_INT16ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT16ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_INT32ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT32ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT32ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT8CLAMPEDARRAY_PROTOTYPE)
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT64ARRAY_PROTOTYPE)
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */
+#endif /* !CONFIG_DISABLE_TYPEDARRAY_BUILTIN */
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_OBJECT_PROTOTYPE));
+#else /* JERRY_NDEBUG */
+  JERRY_UNUSED (obj_p);
+  return true;
+#endif /* !JERRY_NDEBUG */
+} /* ecma_object_check_class_name_is_object */
+
+/**
  * Get [[Class]] string of specified object
  *
  * @return class name magic string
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1729,6 +1729,23 @@ ecma_object_get_class_name (ecma_object_t *obj_p) /**< object */
...
+          case ECMA_BUILTIN_ID_INT16ARRAY_PROTOTYPE:
+          case ECMA_BUILTIN_ID_UINT16ARRAY_PROTOTYPE:
+          case ECMA_BUILTIN_ID_INT32ARRAY_PROTOTYPE:
+          case ECMA_BUILTIN_ID_UINT32ARRAY_PROTOTYPE:
+          case ECMA_BUILTIN_ID_FLOAT32ARRAY_PROTOTYPE:
+          case ECMA_BUILTIN_ID_UINT8CLAMPEDARRAY_PROTOTYPE:
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
+          case ECMA_BUILTIN_ID_FLOAT64ARRAY_PROTOTYPE:
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */
+#endif /* !CONFIG_DISABLE_TYPEDARRAY_BUILTIN */
```
zherczeg: Perhaps we could simply replace this with a default case.

jiangzidong: Do you mean to remove the assert in the default case?

zherczeg: Yes, it seems that is the best option.

zherczeg: Hm, further thinking. We should keep the assert in the default, but extend it to cover all types.

jiangzidong: But I think it is useful to put assert here. for example, it helps us to catch the issue. If we extend other types in the future, and forget the implement their get_class_name, it will alert us.

I prefer to reserve the assert

jiangzidong: @zherczeg sorry that I missed your lastest comments. Yes, we both agree to reserve the assert. Let me try extending it.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1759,7 +1759,25 @@ ecma_object_get_class_name (ecma_object_t *obj_p) /**< object */
...
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT16ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_INT32ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT32ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT32ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT8CLAMPEDARRAY_PROTOTYPE)
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT64ARRAY_PROTOTYPE)
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */
+#endif /* !CONFIG_DISABLE_TYPEDARRAY_BUILTIN */
+                         );
```
zherczeg: Please add ECMA_BUILTIN_ID_OBJECT_PROTOTYPE before the closing `); ` as well, and delete that case.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1759,7 +1755,25 @@ ecma_object_get_class_name (ecma_object_t *obj_p) /**< object */
...
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT16ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_INT32ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT32ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT32ARRAY_PROTOTYPE)
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT8CLAMPEDARRAY_PROTOTYPE)
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT64ARRAY_PROTOTYPE)
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */
+#endif /* !CONFIG_DISABLE_TYPEDARRAY_BUILTIN */
+                          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_OBJECT_PROTOTYPE));
```
jiangzidong: @zherczeg The code here will cause " error: embedding a directive within macro arguments is not portable [-Werror]"

I couldn't find a proper way to extend the assert for now. What about reverting it to multiple cases?

zherczeg: Do you mean create a small helper function, and call it from the assert? That is possible.

jiangzidong: No, I means to list them as switch-case, like I did before, But the helper function is a good idea, and will solve my problem. Thanks

zherczeg: Sorry I misunderstood you.

Yes, a debug only helper function which returns with bool should solve the issue.

jiangzidong: Now I added the helper function, what do you think of it?

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1667,6 +1667,33 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
...
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_INT32ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT32ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT32ARRAY_PROTOTYPE)
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_UINT8CLAMPEDARRAY_PROTOTYPE)
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_FLOAT64ARRAY_PROTOTYPE)
+#endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */
+#endif /* !CONFIG_DISABLE_TYPEDARRAY_BUILTIN */
+          || ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_OBJECT_PROTOTYPE));
+} /* ecma_object_check_class_name_is_object */
```
zherczeg: Please add the ndebug check around this function.

jiangzidong: If I put the NDEBUG check around the whole function, it will cause "implicit declaration ..." in the `JERRY_ASSERT (ecma_object_check_class_name_is_object (obj_p));`

So i choose to put the check around the function body, if NDEBUG, just return true.

Issue #1616, opened 2017-02-23
Segmentation fault in ecma_arraybuffer_new_object


renatahodovan:
###### Jerry version:

```
Checked revision: 62a581f
Build: debug.linux
```
###### OS:

```
Ubuntu 16.04.2 LTS
```
###### Test case:

```javascript
var a = new ArrayBuffer (0xfffffffe)
```

###### Build command:
```
tools/build.py --clean --debug --profile=es2015-subset --compile-flag=-m32 --system-allocator=on --jerry-libc=off
```
###### Backtrace:

```
Program received signal SIGSEGV, Segmentation fault.
0x080d6c8a in __memset_sse2_rep ()
(gdb) bt
#0  0x080d6c8a in __memset_sse2_rep ()
#1  0x08076159 in ecma_arraybuffer_new_object (length=4294967294) at jerryscript/jerry-core/ecma/operations/ecma-arraybuffer-object.c:99
#2  0x080760d4 in ecma_op_create_arraybuffer_object (arguments_list_p=0xffffc064, arguments_list_len=1) at jerryscript/jerry-core/ecma/operations/ecma-arraybuffer-object.c:71
#3  0x08059d51 in ecma_builtin_arraybuffer_dispatch_construct (arguments_list_p=0xffffc064, arguments_list_len=1) at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-arraybuffer.c:94
#4  0x0809aea4 in ecma_builtin_dispatch_construct (obj_p=0x837e7b8, arguments_list_p=0xffffc064, arguments_list_len=1) at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.inc.h:312
#5  0x0807c943 in ecma_op_function_construct (func_obj_p=0x837e7b8, arguments_list_p=0xffffc064, arguments_list_len=1) at jerryscript/jerry-core/ecma/operations/ecma-function-object.c:712
#6  0x08062909 in opfunc_construct.lto_priv.245 (frame_ctx_p=0xffffc084) at jerryscript/jerry-core/vm/vm.c:443
#7  0x0805ae1d in vm_execute (frame_ctx_p=0xffffc084, arg_p=0x0, arg_list_len=0) at jerryscript/jerry-core/vm/vm.c:2656
[...]

PR #1618, opened 2017-02-24
Make sure the byteLength of typedarray will not overflow

jiangzidong:
fix issue #1616 

When allocate new arraybuffer, it need to malloc a buffer. 
if the length is larger than `(UINTPTR_MAX - JMEM_ALIGNMENT - sizeof (ecma_extended_object_t) + 1)`
then the alloc size will be exceed `UINTPTR_MAX`. Since the arg type of size in jmem_heap_alloc_block_internalis `size_t`, size value will overflow.

In this patch, we will check the length to prevent from overflow

build cmd `tools/build.py --clean --debug --profile=es2015-subset --compile-flag=-m32 --system-allocator=on --jerry-libc=off`

`var a = new ArrayBuffer(4294967264)` will cause OUT_OF_MEM
`var a = new ArrayBuffer(4294967265)` will cause range_error

### DIFF
--- a/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
+++ b/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
@@ -321,6 +321,12 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
   ecma_value_t ret_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY);
 
   ecma_value_t new_typedarray = ecma_op_create_typedarray_with_type_and_length (obj_p, len);
+
+  if (ECMA_IS_VALUE_ERROR (new_typedarray))
+  {
+    return new_typedarray;
+  }
+
   ecma_object_t *new_obj_p = ecma_get_object_from_value (new_typedarray);
 
   for (uint32_t index = 0; index < len && ecma_is_value_empty (ret_value); index++)
@@ -331,7 +337,7 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
 
     ECMA_TRY_CATCH (mapped_value, ecma_op_function_call (func_object_p, cb_this_arg, call_args, 3), ret_value);
 
-    ecma_value_t set_status = ecma_op_typedarray_set_index_prop (new_obj_p, index, mapped_value);
+    bool set_status = ecma_op_typedarray_set_index_prop (new_obj_p, index, mapped_value);
 
     if (!set_status)
     {
@@ -574,25 +580,33 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
   if (ecma_is_value_empty (ret_value))
   {
     ecma_value_t new_typedarray = ecma_op_create_typedarray_with_type_and_length (obj_p, pass_num);
-    ecma_object_t *new_obj_p = ecma_get_object_from_value (new_typedarray);
 
-    for (uint32_t index = 0; index < pass_num && ecma_is_value_empty (ret_value); index++)
-    {
-      ecma_value_t set_status = ecma_op_typedarray_set_index_prop (new_obj_p, index, *(pass_value_list + index));
-
-      if (!set_status)
-      {
-        ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("error in typedarray set"));
-      }
-    }
-
-    if (ecma_is_value_empty (ret_value))
+    if (ECMA_IS_VALUE_ERROR (new_typedarray))
     {
       ret_value = new_typedarray;
     }
     else
     {
-      ecma_free_value (new_typedarray);
+      ecma_object_t *new_obj_p = ecma_get_object_from_value (new_typedarray);
+
+      for (uint32_t index = 0; index < pass_num && ecma_is_value_empty (ret_value); index++)
+      {
+        bool set_status = ecma_op_typedarray_set_index_prop (new_obj_p, index, *(pass_value_list + index));
+
+        if (!set_status)
+        {
+          ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("error in typedarray set"));
+        }
+      }
+
+      if (ecma_is_value_empty (ret_value))
+      {
+        ret_value = new_typedarray;
+      }
+      else
+      {
+        ecma_free_value (new_typedarray);
+      }
     }
   }
 
--- a/jerry-core/ecma/operations/ecma-arraybuffer-object.c
+++ b/jerry-core/ecma/operations/ecma-arraybuffer-object.c
@@ -21,6 +21,7 @@
 #include "ecma-gc.h"
 #include "ecma-globals.h"
 #include "ecma-helpers.h"
+#include "jmem.h"
 
 #ifndef CONFIG_DISABLE_ARRAYBUFFER_BUILTIN
 
@@ -66,6 +67,11 @@ ecma_op_create_arraybuffer_object (const ecma_value_t *arguments_list_p, /**< li
     {
       return ret;
     }
+  }
+
+  if (length > UINT32_MAX - sizeof (ecma_extended_object_t) - JMEM_ALIGNMENT + 1)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid ArrayBuffer length."));
   }
 
   ecma_object_t *object_p = ecma_arraybuffer_new_object (length);
--- a/jerry-core/ecma/operations/ecma-typedarray-object.c
+++ b/jerry-core/ecma/operations/ecma-typedarray-object.c
@@ -205,18 +205,28 @@ set_typedarray_element (lit_utf8_byte_t *dst, /**< the location in the internal 
  *
  * See also: ES2015 22.2.1.2.1
  *
- * @return pointer to the new typedarray object
+ * @return ecma value of the new typedarray object
+ *         Returned value must be freed with ecma_free_value
  */
-static ecma_object_t *
+static ecma_value_t
 ecma_typedarray_create_object_with_length (ecma_length_t array_length, /**< length of the typedarray */
                                            ecma_object_t *proto_p, /**< prototype object */
                                            uint8_t element_size_shift, /**< the size shift of the element length */
                                            lit_magic_string_id_t class_id) /**< class name of the typedarray */
 {
+  if (array_length > (UINT32_MAX >> element_size_shift))
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Maximum typedarray size is reached."));
+  }
+
   ecma_length_t byte_length = array_length << element_size_shift;
 
-  ecma_value_t new_arraybuffer_p = ecma_make_object_value (ecma_arraybuffer_new_object (byte_length));
+  if (byte_length > UINT32_MAX - sizeof (ecma_extended_object_t) - JMEM_ALIGNMENT + 1)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Maximum typedarray size is reached."));
+  }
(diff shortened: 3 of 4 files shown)


jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -321,6 +321,12 @@ ecma_builtin_typedarray_prototype_map (ecma_value_t this_arg, /**< this argument
   ecma_value_t ret_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY);
 
   ecma_value_t new_typedarray = ecma_op_create_typedarray_with_type_and_length (obj_p, len);
+
+  if (ECMA_IS_VALUE_ERROR (new_typedarray))
+  {
+    ret_value = ecma_copy_value (new_typedarray);
+  }
```
zherczeg: Shouldn't we add a break here?

jerry-core/ecma/operations/ecma-arraybuffer-object.c
```diff
@@ -68,6 +69,11 @@ ecma_op_create_arraybuffer_object (const ecma_value_t *arguments_list_p, /**< li
     }
   }
 
+  if (length + sizeof (ecma_extended_object_t) + JMEM_ALIGNMENT - 1 <= length)
```
zherczeg: These expressions might be optimized out by the compiler. I saw such things before. Why not simply check the return value of alloc? The ecma_arraybuffer_new_object could return with NULL if the allocation is not successful.

zherczeg: The compiler optimization issue remains, so we need to figure out something better.

What about:

    if (length >= UINT32_MAX - sizeof (ecma_extended_object_t))

jiangzidong: what about `if (length >= UINT32_MAX - sizeof (ecma_extended_object_t)) - JMEM_ALIGNMENT + 1`

if we don't use the system allocator, the overflow may also happen during the alignment.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
jiangzidong: it's not in a loop

jerry-core/ecma/operations/ecma-arraybuffer-object.c
zherczeg: Ok. Btw redardlee of allocator, all Jerry pointers must be aligned to JMEM_ALIGNMENT. The posix_memalign can ensure this.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
zherczeg: The problem is, the ecma_object_t *new_obj_p = ecma_get_object_from_value (new_typedarray) below might not work, so we need to bail out some way. Perhaps a return could help.

jiangzidong: The code below should work. if the new_typedarray is error value, we can also get the object_t from it. 

But if we need to skip the code, we can add `else`

zherczeg: Not necessarily. If there is no memory, the error object is a null value with an error flag (that does not require memory allocation).

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -303,6 +312,12 @@ ecma_typedarray_create_object_with_typedarray (ecma_object_t *typedarray_p, /**<
   else
   {
     ecma_length_t byte_length = array_length << element_size_shift;
+
```
zherczeg: Question: byte_length = array_length << element_size_shift this shift could also lead to overflow (resulting a 0 length).

jiangzidong: Yes, I should check every shift operation. A big project.
what about using `array_length>>(sizeof (array_length) - shift) ==0` to indicate there is no overflow

zherczeg: What about:

    if (array_length > (UINT32_MAX >> element_size_shift)) // overflow

jiangzidong: Sure. Can we use `//` to add comments? I only see `/* */` in the source code

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -205,18 +205,28 @@ set_typedarray_element (lit_utf8_byte_t *dst, /**< the location in the internal
  *
  * See also: ES2015 22.2.1.2.1
  *
- * @return pointer to the new typedarray object
+ * @return ecma value
```
LaszloLango: `@return ecma value of the new typedarray object`

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -229,17 +239,18 @@ ecma_typedarray_create_object_with_length (ecma_length_t array_length, /**< leng
...
+  return ecma_make_object_value (object_p);
 } /* !ecma_typedarray_create_object_with_length */
 
 /**
  * Create a TypedArray object by given buffer, offset, and array_length
  *
  * See also: ES2015 22.2.1.5
  *
- * @return pointer to the new typedarray object
+ * @return ecma value
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -277,17 +289,18 @@ ecma_typedarray_create_object_with_buffer (ecma_object_t *arraybuffer_p, /**< th
...
  * Create a TypedArray object by given another TypedArray object
  *
  * See also: ES2015 22.2.1.3
  *
- * @return pointer to the new typedarray object
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value
  */
-static ecma_object_t *
+static ecma_value_t
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -205,18 +205,28 @@ set_typedarray_element (lit_utf8_byte_t *dst, /**< the location in the internal
...
 ecma_typedarray_create_object_with_length (ecma_length_t array_length, /**< length of the typedarray */
                                            ecma_object_t *proto_p, /**< prototype object */
                                            uint8_t element_size_shift, /**< the size shift of the element length */
                                            lit_magic_string_id_t class_id) /**< class name of the typedarray */
 {
+  if (array_length > (UINT32_MAX >> element_size_shift))
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("byte_length overflows"));
```
zherczeg: Maximum typed array size is reached.

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -205,18 +205,28 @@ set_typedarray_element (lit_utf8_byte_t *dst, /**< the location in the internal
...
+  }
+
   ecma_length_t byte_length = array_length << element_size_shift;
 
-  ecma_value_t new_arraybuffer_p = ecma_make_object_value (ecma_arraybuffer_new_object (byte_length));
+  if (byte_length > UINT32_MAX - sizeof (ecma_extended_object_t) - JMEM_ALIGNMENT + 1)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid ArrayBuffer length."));
```
zherczeg: Maximum typed array size is reached.

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -306,7 +319,18 @@ ecma_typedarray_create_object_with_typedarray (ecma_object_t *typedarray_p, /**<
   }
   else
   {
+    if (array_length > (UINT32_MAX >> element_size_shift))
+    {
+      return ecma_raise_range_error (ECMA_ERR_MSG ("byte_length overflows"));
```
zherczeg: Maximum typed array size is reached.

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -306,7 +319,18 @@ ecma_typedarray_create_object_with_typedarray (ecma_object_t *typedarray_p, /**<
...
+    }
+
     ecma_length_t byte_length = array_length << element_size_shift;
+
+    if (byte_length > UINT32_MAX - sizeof (ecma_extended_object_t) - JMEM_ALIGNMENT + 1)
+    {
+      return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid ArrayBuffer length."));
```
zherczeg: Maximum typed array size is reached.

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -382,11 +406,18 @@ ecma_op_typedarray_from (ecma_value_t items_val, /**< the source array-like obje
   uint32_t len = ecma_number_to_uint32 (len_number);
 
   /* 14 */
-  ecma_object_t *new_typedarray_p = ecma_typedarray_create_object_with_length (len,
+  ecma_value_t new_typedarray_val = ecma_typedarray_create_object_with_length (len,
```
zherczeg: Do we have a val psotfix for values? I think that is the default.

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -638,6 +669,12 @@ ecma_op_create_typedarray (const ecma_value_t *arguments_list_p, /**< the arg li
           ECMA_OP_TO_NUMBER_TRY_CATCH (num3, arg3, ret);
           int32_t new_length = ecma_number_to_int32 (num3);
           new_length = (new_length > 0) ? new_length : 0;
+
+          if ((uint32_t) new_length > (UINT32_MAX >> element_size_shift))
+          {
+            ret = ecma_raise_range_error (ECMA_ERR_MSG ("byte_length overflows"));
```
zherczeg: Maximum typed array size is reached.

jerry-core/ecma/operations/ecma-typedarray-object.c
jiangzidong: And I will also remove other val psotfix. Thanks

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -574,25 +580,33 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
...
     else
     {
-      ecma_free_value (new_typedarray);
+      ecma_object_t *new_obj_p = ecma_get_object_from_value (new_typedarray);
+
+      for (uint32_t index = 0; index < pass_num && ecma_is_value_empty (ret_value); index++)
+      {
+        ecma_value_t set_status = ecma_op_typedarray_set_index_prop (new_obj_p, index, *(pass_value_list + index));
+
+        if (!set_status)
```
zherczeg: Hm, I am not sure I understand this. set_status is an ecma_value but here a !set_status is checked. Is it a bool?

jiangzidong: My fault. the return type of ecma_op_typedarray_set_index_prop  should be bool

jiangzidong:
@zherczeg @LaszloLango I updated the patch. @akosthekiss I added the regression case. Sorry for the delay

Issue #1605, opened 2017-02-21
Assertion 'ecma_get_value_type_field (value) == ECMA_TYPE_OBJECT' failed in ecma_get_object_from_value


bzsolt:
###### Environment: ######
``` 
Ubuntu 16.04
gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4)
Jerry: 92eaea2
```

###### Build command: ######
`tools/build.py --debug --profile=es2015-subset --compile-flag=-m32 --system-allocator=on --jerry-libc=off`

###### Test case: ######
```javascript
var a = new Int8Array();
```

###### Error: ######
```
Assertion 'ecma_get_value_type_field (value) == ECMA_TYPE_OBJECT' failed at jerryscript/jerry-core/ecma/base/ecma-helpers-value.c(ecma_get_object_from_value):605
```

###### Backtrace: ######
```
#0  0xf7ffdbe9 in __kernel_vsyscall ()
#1  0x080f2847 in raise ()
#2  0x080b9f87 in abort ()
#3  0x080649d2 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/targets/default/jerry-port-default-fatal.c:53
#4  0x0807a3d4 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:61
#5  0x0807a407 in jerry_assert_fail (assertion=0x8140030 "ecma_get_value_type_field (value) == ECMA_TYPE_OBJECT", 
    file=0x813fdec "jerryscript/jerry-core/ecma/base/ecma-helpers-value.c", 
    function=0x8129334 <__func__.3617.lto_priv.629> "ecma_get_object_from_value", line=609)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:85
#6  0x080a1703 in ecma_get_object_from_value (value=0)
    at jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:609
#7  0x08080d32 in ecma_typedarray_get_arraybuffer (typedarray_p=0x837fbb8)
    at jerryscript/jerry-core/ecma/operations/ecma-typedarray-object.c:462
#8  0x080a702e in ecma_gc_mark.lto_priv.647 (object_p=0x837fbb8)
    at jerryscript/jerry-core/ecma/base/ecma-gc.c:284
[...]

PR #1619, opened 2017-02-24
Fix issue #1605

jiangzidong:
Fix issue #1605 

We should avoid allocating other objects during we create a typedarray.

Because typearray objects has a member: arraybuffer. when we run gc_mark, it should also mark the arraybuffer. But during we create a typedarray, the typedarray.arraybuffer is not set yet, and it is a wild pointer. We just avoid gc during typedarray creataion.


zherczeg:
No other objects should be allocated before a typed array is fully initialized.

Sorry for the typo.

jiangzidong:
@zherczeg Thanks, I will change other PRs. Btw, Can i press the `squash and merge` button then change the commit message? I haven't done that before, I am not sure...

About the check fail, it tells the `./run-test --jerry-debugger` failed. But it works fine in my host machine (64bit ubuntu 14.04)
Can we re-run the check?

zherczeg:
Yes you can. Sometimes I do that as well, but I would be happier if I wouldn't have to update others commit messages :)

PR #1623, opened 2017-02-27
Refactor the generator scripts for unicode tables

robertsipka:
Extract the source code generator methods into a separated `unicode_c_source.py` script.
Fix the generator scripts to make them compatible with both Python2 and Python3.
Remove pylint warnings.


tools/unicode_c_source.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+class Source(object):
+    def __init__(self, filepath):
+        self.__filepath = filepath
+        self.__header = LICENSE
+        self.__data = ""
+
+    def complete_header(self, completion):
+        header = self.__header[:len(self.__header) - 3]
```
galpeter: If we are removing the last three characters always, why don't we leave out from the string in the first place? Or even better do we need to remove those characters at all? can't we just append the `completion` as a comment and leave the license string alone?

tools/unicode_c_source.py
```diff
@@ -0,0 +1,74 @@
...
+        header += " *\n" + completion
+        header += " */\n\n"
```
galpeter: Personally I'm not really fond of adding strings together. I would use a list add any required lines then join them together with `\n`-s

tools/unicode_c_source.py
```diff
@@ -0,0 +1,74 @@
...
+        self.__header = header
+
+    def add_table(self, table, table_name, table_type, table_descr):
+        self.__data += (table_descr + "\n"
+                        "static const " + table_type + " jerry_" + table_name + "[] JERRY_CONST_DATA =\n"
+                        "{\n" + format_code(table, 1) + "\n};\n\n")
```
galpeter: If the `__data` is a list then this could look like this:
```
self.__data.append(table_descr)
self.__data.append("static const %s jerry_%s[] JERRY_CONST_DATA" % (table_type, table_name))
self.__data.append("{")
self.__data.append(format_code(table, 1))
self.__data.append("}")
self.__data.append("") # for an extra empty line
```
then when writing out the data we can just simple join them together. What do you think?

robertsipka: @galpeter : I like it, thanks! It's a prettier solution and I've changed these parts based on your suggestions.

robertsipka:
@galpeter : thanks for your suggestions, I've updated the patch.

tools/unicode_c_source.py
```diff
@@ -0,0 +1,78 @@
...
+    def add_table(self, table, table_name, table_type, table_descr):
+        self.__data.append(table_descr)
+        self.__data.append("static const %s jerry_%s[] JERRY_CONST_DATA =" % (table_type, table_name))
+        self.__data.append("{")
+        self.__data.append(format_code(table, 1))
+        self.__data.append("};")
+        self.__data.append("") # for an extra empty line
+
+
+
```
galpeter: Why so many empty lines? Just have one empty line between methods inside a class

tools/unicode_case_conversion.py
```diff
@@ -46,7 +217,11 @@ def parse_unicode_sequence(raw_data):
             continue
 
         # Convert it to unicode code point (from hex value without 0x prefix)
-        result += unichr(int(unicode_char, 16))
+        try:
+            result += unichr(int(unicode_char, 16))
+        except NameError:
+            result += chr(int(unicode_char, 16))
```
galpeter: I'm not sure about this, but: maybe we should extract the `int` call before the try-catch. What do you think?

robertsipka: I've updated it accordingly.

galpeter:
lgtm

PR #1628, opened 2017-02-28
Add delete all breakpoints feature to debugger (javascript)

polaroi8d:
Now you can able to use `delete all` command, to clear the breakpoints from the whole project. 

### DIFF
--- a/jerry-debugger/jerry-client-ws.html
+++ b/jerry-debugger/jerry-client-ws.html
@@ -828,7 +828,23 @@ function DebuggerClient(address)
   {
     breakpoint = activeBreakpoints[index];
 
-    if (!breakpoint)
+    if (index == "all")
+    {
+      var found = false;
+
+      for (var i in activeBreakpoints)
+      {
+        delete activeBreakpoints[i];
+        found = true;
+      }
+
+      if (!found)
+      {
+        appendLog("No active breakpoints.")
+      }
+    }
+
+    else if (!breakpoint)
     {
       appendLog("No breakpoint found with index " + index);
       return;


jerry-debugger/jerry-client-ws.html
```diff
@@ -828,7 +828,23 @@
...
+        delete activeBreakpoints[i];
+        found = true;
+      }
+
+      if (!found)
+      {
+        appendLog("No active breakpoints.")
+      }
+    }
+
```
robertsipka: I think this form would be better, whats your opinion?
activeBreakpoints.length != 0 ? activeBreakpoints.splice(0) : appendLog("No active breakpoints.")

polaroi8d: You can't use splice() function in the activeBreakpoints object.

jerry-debugger/jerry-client-ws.html
```diff
@@ -828,7 +828,23 @@
...
+      for (var i in activeBreakpoints)
+      {
+        delete activeBreakpoints[i];
+        found = true;
+      }
+
+      if (!found)
+      {
+        appendLog("No active breakpoints.")
+      }
```
zherczeg: Isn't there an easier way to check whether activeBreakpoints is empty?

jerry-debugger/jerry-client-ws.html
robertsipka: It's my bad, I thought that activeBreakpoints is an array.

jerry-debugger/jerry-client-ws.html
robertsipka: You can checked with `Object.keys(acviteBreakpoints).length`

zherczeg: That is more complicated. Keep the bool then.

PR #1629, opened 2017-02-28
Fix typos in parser error messages.

renatahodovan:

### DIFF
--- a/jerry-core/parser/js/js-parser-util.c
+++ b/jerry-core/parser/js/js-parser-util.c
@@ -731,11 +731,11 @@ parser_error_to_string (parser_error_t error) /**< error code */
     }
     case PARSER_ERR_NUMBER_TOO_LONG:
     {
-      return "Number too long.";
+      return "Number is too long.";
     }
     case PARSER_ERR_REGEXP_TOO_LONG:
     {
-      return "Regexp too long.";
+      return "Regexp is too long.";
     }
     case PARSER_ERR_UNTERMINATED_MULTILINE_COMMENT:
     {
@@ -767,11 +767,11 @@ parser_error_to_string (parser_error_t error) /**< error code */
     }
     case PARSER_ERR_EVAL_NOT_ALLOWED:
     {
-      return "Eval is not allowed to use here in strict mode.";
+      return "Eval is not allowed to be used here in strict mode.";
     }
     case PARSER_ERR_ARGUMENTS_NOT_ALLOWED:
     {
-      return "Arguments is not allowed to use here in strict mode.";
+      return "Arguments is not allowed to be used here in strict mode.";
     }
     case PARSER_ERR_DELETE_IDENT_NOT_ALLOWED:
     {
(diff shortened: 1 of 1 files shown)


jerry-core/parser/js/js-parser-util.c
```diff
@@ -731,11 +731,11 @@ parser_error_to_string (parser_error_t error) /**< error code */
     }
     case PARSER_ERR_NUMBER_TOO_LONG:
     {
-      return "Number too long.";
+      return "Number is too long.";
```
zherczeg: I know this is not correct grammatically but this kind of short form is frequently used, doesn't it?

renatahodovan: Maybe. But we should use consistent messages and adapt the previous two accordingly. Is the short version preferred then?

zherczeg: Based on the others probably not.

jerry-core/parser/js/js-parser-util.c
```diff
@@ -923,7 +923,7 @@ parser_error_to_string (parser_error_t error) /**< error code */
     }
     case PARSER_ERR_NON_STRICT_ARG_DEFINITION:
     {
-      return "Non strict argument definition.";
+      return "Non-strict argument definition.";
```
zherczeg: Strict is enabled by "use strict" and not "use-strict" so perhaps we shouldn't add dashes.

renatahodovan: "use strict" is an ecmascript directive (which is grammatically correct in English). However, the "non-strict" expression is a different language construct and it's incorrect without the dash (btw it's used at 328 distinct places in the project and  in the test262 repo in this dashed form).

zherczeg:
Would you like to fix some fuzzer bugs?

renatahodovan:
Would be nice (in a distant future), but right now I have other priorities, like finding and reporting further bugs ;) (Seriously, my expertise lies in fuzzer technologies, not in JS engine internals. My goal is to help strengthening the project by finding implementation issues as early as possible.)

zherczeg:
But finding and fixing a bug has even more positive effect on a project. So I encourage you to fix some of those bugs.

pfalcon:
@renatahodovan : Out of curiosity, thru how many false positives you have to wade thru to find one real issue, as submitted to the project. (I myself very surprised by the amount of high-quality cases found by fuzzing, but I wonder what's there that I don't see).

renatahodovan:
@pfalcon False positives as such (i.e., something that gets misdiagnosed as an issue but actually isn't) are quite rare - perhaps segfaults caused by infinite recursions classify so. Test cases generated "superfluously" are another story, however, they are part of the nature of fuzz testing. And that ratio (non-failing tests vs failing tests) highly depends on the quality/characteristic of the used test generator. Generally speaking, total randomisation (e.g., random bytes that do not follow any syntax) results in less useful (e.g., issue- triggering) test cases, since they will be mostly discarded in an early stage of the input processing. On the contrary, generators following the rules (syntax or type expectations) too closely are creating tests that are already covered by the project's test suite and don't fail. So, the key is to find the golden mean.

To give some numbers, the latest generator I've been using resulted in a failure in every 30-50 tests, approximately. I've been running it every night during the last week on my workstation and it has executed cca. 8 million test cases (on Jerry, at least). From these issues, 11 were unknown and unique (8 of them are reported already, and 3 are still in my report queue).

As you can see, many of the crashes are duplicates but fortunately, I have a framework (Fuzzinator) that filters the unique ones and prepares them for reporting, automatically.

pfalcon:
@renatahodovan : Thanks for the detailed response. I never tried fuzzing myself and know about it only superficially, but your reply confirms my impression that it takes quite some work and experience with it to get good outcome from it. The dichotomy you pin-point - completely random vs "generative" and the fact that neither of these 2 extremes is productive, also helps to organize my thoughts better. Of false positives, we got reports from someone trying to fuzz MicroPython about problems with things like "99**99999999", which get faithfully computed for a minute or two before leading to OOM (completely random input with AFL was used IIRC).

Anyway, thanks for sharing, will be interesting to look into your Fuzzinator on a calm day, the reports here are pretty impressive.

PR #1632, opened 2017-03-01
Allow to skip certain JS-tests under `tests/jerry/` and `tests/jerry-…

bzsolt:
…test-suite/`

Add `--skip-list` option to tools/run-tests.py.

Additional modifications:
 - The test `parser-oom.js` doesn't work when system-allocator is enabled,
   hence we skip it in these jobs.
 - The sanitizer-tasks become mandatory.

### DIFF
--- a/tools/runners/run-test-suite.sh
+++ b/tools/runners/run-test-suite.sh
@@ -15,7 +15,7 @@
 # limitations under the License.
 
 # Usage:
-#       ./tools/runners/run-test-suite.sh ENGINE TESTS [--snapshot] ENGINE_ARGS....
+#       ./tools/runners/run-test-suite.sh ENGINE TESTS [--skip-list=item1,item2] [--snapshot] ENGINE_ARGS....
 
 TIMEOUT=${TIMEOUT:=5}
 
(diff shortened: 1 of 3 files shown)


zherczeg:
Shouldn't we skip es2015 tests?

bzsolt:
I don't know any reason to skip those tests.

zherczeg:
Ah, so we always test with all features?

bzsolt:
Actually yes, the tested combinations listed in [tools/run-tests.py](https://github.com/jerryscript-project/jerryscript/blob/master/tools/run-tests.py#L86)

jiangzidong:
Do you mean we should always run ed2015-subset jerry for --jerry-tests?

bzsolt:
The `--jerry-test-suite` option runs the es2015-subset, because these tests are under the `tests/jerry-test-suite/`. If the profile isn't specified, the `es5.1` is the default.

tools/runners/run-test-suite.sh
```diff
@@ -64,6 +71,12 @@ else
     exit 1
 fi
 
+# Remove the skipped tests from list
+for test in "${SKIP_LIST[@]}"
```
LaszloLango: Please use uppercase variable names in bash scripts. (`TEST`)

tools/runners/run-test-suite.sh
```diff
@@ -64,6 +71,12 @@ else
...
+do
+    ( sed -i "/$test/d" $TEST_FILES )
+done
+
 total=$(cat $TEST_FILES | wc -l)
```
LaszloLango: Could you fix this too if you are there? :)

jiangzidong:
Could you let --skip-list to support a list file, like the profile-list in the jerry-test-suite, ? Then I can check in the `run-tests.py`, if the profile is NOT es2015, then I will add the cmd arg: `--skip-list es2015-test-list`

zherczeg:
Ok for me, but at some point we should implement a common test selection system.

bzsolt:
@LaszloLango I've updated the PR.
@jiangzidong So the `--skip-list` should support the 'single-file' and a 'list-based' skip too?
It's possible, if necessary, currently you can skip an entire folder under `tests/jerry-test-suite` (or `tests/jerry`) by
`tools/run-tests.py --jerry-test-suite --skip-list='es2015\/'`.

jiangzidong:
OK, it is good to support skipping a sub folder. LGTM :)

Issue #1633, opened 2017-03-02
Assertion 'ECMA_PROPERTY_GET_TYPE (property) == ECMA_PROPERTY_TYPE_NAMEDDATA || ECMA_PROPERTY_GET_TYPE (property) == ECMA_PROPERTY_TYPE_NAMEDACCESSOR || ECMA_PROPERTY_GET_TYPE (property) == ECMA_PROPERTY_TYPE_VIRTUAL' failed in ecma_is_property_configurable


renatahodovan:
###### Jerry version:

```
Checked revision: 92f74f3
Build: debug.linux
```
###### OS:

```
Ubuntu 16.04.2 LTS
```
###### Test case:

```javascript

delete new Uint16Array()[0]
```

###### Build command:
```
./tools/build.py --clean --debug --compile-flag=-m32 --system-allocator=on --jerry-libc=off
```

###### Backtrace:

```
ICE: Assertion 'ECMA_PROPERTY_GET_TYPE (property) == ECMA_PROPERTY_TYPE_NAMEDDATA || ECMA_PROPERTY_GET_TYPE (property) == ECMA_PROPERTY_TYPE_NAMEDACCESSOR || ECMA_PROPERTY_GET_TYPE (property) == ECMA_PROPERTY_TYPE_VIRTUAL' failed at jerryscript/jerry-core/ecma/base/ecma-helpers.c(ecma_is_property_configurable):1197.
Error: ERR_FAILED_INTERNAL_ASSERTION
bt
#0  0xf7ffdbe9 in __kernel_vsyscall ()
#1  0x080f65c7 in raise ()
#2  0x080ba557 in abort ()
#3  0x08059ae5 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/targets/default/jerry-port-default-fatal.c:53
#4  0x0806fbd6 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:61
[...]

PR #1637, opened 2017-03-02
Check if property is not found before delete it.

jiangzidong:
Check if property is not found before delete it.

Fix issue #1633 


jiangzidong:
I will add the test case after #1632 is landed

zherczeg:
I would prefer to add real description to these commits not Fix issue. In the description the Fix issue should be present, but not in the title.

zherczeg:
A test file is still missing :)

jiangzidong:
@zherczeg I know, I will add the case as soon as #1627 merged, since #1627 contain the jerry-tests framework for es2015. Sorry for the dalay

jiangzidong:
@zherczeg The test file is added.

PR #1639, opened 2017-03-03
Require JERRY_JS_PARSER macro to be defined

robertsipka:


jerry-core/jerry.c
```diff
@@ -50,9 +50,11 @@ JERRY_STATIC_ASSERT ((int) ECMA_ERROR_COMMON == (int) JERRY_ERROR_COMMON
                      && (int) ECMA_ERROR_URI == (int) JERRY_ERROR_URI,
                      ecma_standard_error_t_must_be_equal_to_jerry_error_t);
 
-#if !defined (JERRY_JS_PARSER) && !defined (JERRY_ENABLE_SNAPSHOT_EXEC)
-#error JERRY_JS_PARSER or JERRY_ENABLE_SNAPSHOT_EXEC must be defined!
-#endif /* !JERRY_JS_PARSER && !JERRY_ENABLE_SNAPSHOT_EXEC */
+#ifndef JERRY_JS_PARSER
+#error JERRY_JS_PARSER must be defined with 0 (disabled) or 1 (enabled)
+#elif (JERRY_JS_PARSER == 0) && (!defined (JERRY_ENABLE_SNAPSHOT_EXEC) || JERRY_ENABLE_SNAPSHOT_EXEC == 0)
```
zherczeg: !(JERRY_JS_PARSER) instead of (JERRY_JS_PARSER == 0)

I wouldn't change the JERRY_ENABLE_SNAPSHOT_EXEC check, since we only check whether it is present. The JERRY_JS_PARSER is a special exception because people don't expect that the parser can be disabled.

robertsipka: @zherczeg : Thanks, I've updated it.

PR #1642, opened 2017-03-03
Fix the time computations of mbed ports

loki04:
The mbed ports used wrong time computations for the
'jerry_port_get_current_time' function. Microseconds was used instead of
milliseconds.

A quick fix has been done in the Yotta-based mbed port, and an extended one
- with correction - works in the mbed OS 5 port now.

JerryScript-DCO-1.0-Signed-off-by: Gabor Loki loki@inf.u-szeged.hu

### DIFF
--- a/targets/mbed/source/port/jerry_port.c
+++ b/targets/mbed/source/port/jerry_port.c
@@ -76,10 +76,12 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
 /**
  * Implementation of jerry_port_get_current_time.
  *
- * @return current timer's counter value in microseconds 
+ * @return current timer's counter value in milliseconds
  */
 double
 jerry_port_get_current_time ()
 {
-  return (double) us_ticker_read ();
+  /* Note: if the target has its own RTC, this value should be extended by the
+   * RTC's one. */
+  return (double) us_ticker_read () / 1000;
 } /* jerry_port_get_current_time */
--- a/targets/mbedos5/source/jerry_port_mbed.c
+++ b/targets/mbedos5/source/jerry_port_mbed.c
@@ -90,10 +90,34 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
 /**
  * Implementation of jerry_port_get_current_time.
  *
- * @return current timer's counter value in microseconds
+ * @return current timer's counter value in milliseconds
  */
 double
 jerry_port_get_current_time ()
 {
-  return (double) us_ticker_read ();
+  static uint64_t last_tick = 0;
+  static time_t last_time = 0;
+  static uint32_t skew = 0;
+
+  uint64_t curr_tick = us_ticker_read (); /* The value is in microseconds. */
+  time_t curr_time = time(NULL); /*  The value is in seconds. */
+  double result = curr_time * 1000;
+
+  /* The us_ticker_read () has an overflow for each UINT_MAX microseconds
+   * (~71 mins). For each overflow event the ticker-based clock is about 33
+   * milliseconds fast. Without a timer thread the milliseconds part of the
+   * time can be corrected if the difference of two get_current_time calls
+   * are within the mentioned 71 mins. Above that interval we can assume
+   * that the milliseconds part of the time is negligibe.
+   */
+  if (curr_time - last_time > (time_t)(((uint32_t)-1) / 1000000)) {
+    skew = 0;
+  } else if (last_tick > curr_tick) {
+    skew = (skew + 33) % 1000;
+  }
+  result += (curr_tick / 1000 - skew) % 1000;
+
+  last_tick = curr_tick;
+  last_time = curr_time;
+  return result;
 } /* jerry_port_get_current_time */


LaszloLango:
CC: @thegecko

thegecko:
@janjongboom

targets/mbed/source/port/jerry_port.c
```diff
@@ -76,10 +76,12 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
...
- * @return current timer's counter value in microseconds 
+ * @return current timer's counter value in milliseconds
  */
 double
 jerry_port_get_current_time ()
 {
-  return (double) us_ticker_read ();
+  /* Note: if the target has its own RTC, this value should be extended by the
+   * RTC's one. */
+  return (double) us_ticker_read () / 1000;
```
zherczeg: Shouldn't we have the same algorithm here?

targets/mbedos5/source/jerry_port_mbed.c
```diff
@@ -90,10 +90,36 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
...
+
+  uint32_t curr_tick = us_ticker_read (); /* The value is in microseconds. */
+  time_t curr_time = time(NULL); /*  The value is in seconds. */
+  double result = curr_time * 1000;
+
+  /* The us_ticker_read () has an overflow for each UINT_MAX microseconds
+   * (~71 mins). For each overflow event the ticker-based clock is about 33
+   * milliseconds fast. Without a timer thread the milliseconds part of the
+   * time can be corrected if the difference of two get_current_time calls
+   * are within the mention 71 mins (note: because the comparision of two
```
zherczeg: mentioned, comparison

targets/mbedos5/source/jerry_port_mbed.c
```diff
@@ -90,10 +90,36 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
...
+   * uint32 numbers, this interval has to be diveded by 2). Above that
```
zherczeg: divided

targets/mbedos5/source/jerry_port_mbed.c
```diff
@@ -90,10 +90,36 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
...
+   * interval we can assume that the milliseconds part of the time is
+   * negligibe.
+   */
+  if (curr_time - last_time > 1800) {
+    skew = 0;
+  } else if (last_tick > curr_tick) {
+    skew = (skew + 33) % 1000;
+  }
+  result += (curr_tick / 1000 - skew) % 1000;
```
zherczeg: What happens if 2 hours passed without calling this function?

targets/mbed/source/port/jerry_port.c
loki04: This is in the Yotta-based part of mbed. It was fix based on the mbed API, not tested on any board. I do not want to put any complex computation here without testing.

targets/mbedos5/source/jerry_port_mbed.c
loki04: There will be a clock skew in the millisecond part of time. Since the seconds are counted by the RTC, there will be no skew there.
In general we can assume that if two hours passed you are not interested in the correct milliseconds.

zherczeg:
Good to me. @janjongboom any comments?

janjongboom:
@zherczeg I'll test this out tomorrow.

targets/mbedos5/source/jerry_port_mbed.c
```diff
@@ -90,10 +90,36 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
...
+  /* The us_ticker_read () has an overflow for each UINT_MAX microseconds
+   * (~71 mins). For each overflow event the ticker-based clock is about 33
+   * milliseconds fast. Without a timer thread the milliseconds part of the
+   * time can be corrected if the difference of two get_current_time calls
+   * are within the mentioned 71 mins (note: because the comparison of two
+   * uint32 numbers, this interval has to be divided by 2). Above that
+   * interval we can assume that the milliseconds part of the time is
+   * negligibe.
+   */
+  if (curr_time - last_time > 1800) {
```
janjongboom: Where does `1800` come from? Shouldn't we have something closer to 2130?

loki04: Well, it should 2147 sec (~35 mins) if we are about to be as correct as possible. Probably the comment does not say anything about the 1800 seconds (30 mins) interval, but it could be a more user friendly threshold than the 2147. It is easier to accept the 30 mins limit than the 35 mins one in order no to rely on milliseconds based timers above that interval.
Do you have any other option?

targets/mbedos5/source/jerry_port_mbed.c
```diff
@@ -90,10 +90,36 @@ jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< timezone pointer */
...
- * @return current timer's counter value in microseconds
+ * @return current timer's counter value in milliseconds
  */
 double
 jerry_port_get_current_time ()
 {
-  return (double) us_ticker_read ();
+  static uint32_t last_tick = 0;
+  static time_t last_time = 0;
+  static uint32_t skew = 0;
```
loki04: I think the only thing which we might consider is that the threads. If the JerryScript can execute JS in parallel threads safely, this function should be reviewed again. I can imagine that some kind of lock will be done around the port functions in this case no to rely the port itself.
On the other hand, static variables in functions can be seen in all over the ports (even in mbed).
What kind of improvements do you see about this static variables not to be so confusing?

janjongboom:
My suggestions:

1. Change the data type used for comparison to `uint64_t` so we have the full 71 minutes.
2. Don't limit it to 30 minutes, but keep it at 71, no need to artificially limit this.
3. I don't like `static` variables in functions, but I'll leave that up to the JS maintainers.

loki04:
@janjongboom , it has been updated based on your suggestions (1, 2).
@zherczeg , do you have any comment on the 3rd suggestion of @janjongboom ?

zherczeg:
Since it is port specific I have no objections for static declarations.

In the core code, writable static variables are not allowed, they must be part of the jerry context. But the context is not prepared for port usage.

loki04:
Is there anything else what I can do for this PR?

janjongboom:
Sorry for the late reply. I'm on holiday right now. PR looks good, can be merged.

Issue #1600, opened 2017-02-21
Cleanup Python warnings


LaszloLango:
We should eliminate all of the python warnings reported by pylint tool. After that we can validate our tool scripts with pylint on travis CI to improve the review process of them. There too many warnings in the current scripts. so my plan is to do it in several PRs.

PR #1644, opened 2017-03-03
Fix several pylint warnings.

LaszloLango:
Fixed
 * 'unused-wildcard-import'
 * 'bad-whitespace'
 * 'bad-continuation'
 * 'wrong-import-order'
in 'build.py', 'run-tests.py' and 'check-license.py'.

Related issue: #1600

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -16,22 +16,22 @@
 
 import argparse
 import multiprocessing
+import os
 import shutil
 import subprocess
 import sys
-from os import makedirs, uname
-from settings import *
+import settings
 
-BUILD_DIR = path.join(PROJECT_DIR, 'build')
-DEFAULT_PORT_DIR = path.join(PROJECT_DIR, 'targets/default')
+BUILD_DIR = os.path.join(settings.PROJECT_DIR, 'build')
+DEFAULT_PORT_DIR = os.path.join(settings.PROJECT_DIR, 'targets/default')
 
-PROFILE_DIR = path.join(PROJECT_DIR, 'jerry-core/profiles')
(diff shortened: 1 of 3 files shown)


tools/build.py
```diff
@@ -16,21 +16,21 @@
 
 import argparse
 import multiprocessing
+from os import makedirs, path, uname
```
galpeter: Personally I like the `import os` variant better (just like in other python files). Maybe it would make sense to use it the same way in all python scripts.

tools/run-tests.py
```diff
@@ -49,39 +49,39 @@
...
-if path.isabs(script_args.outdir):
+if os.path.isabs(script_args.outdir):
     OUTPUT_DIR = script_args.outdir
 else:
-    OUTPUT_DIR = path.join(PROJECT_DIR, script_args.outdir)
+    OUTPUT_DIR = os.path.join(settings.PROJECT_DIR, script_args.outdir)
 
 class Options:
-    def __init__(self, name = '', build_args = None, test_args = None):
+    def __init__(self, name='', build_args=None, test_args=None):
```
galpeter: Partially related: Is it possible to have an option without `name`?

LaszloLango: Yes, it is, but I think it is not recommended.

LaszloLango:
@galpeter, I've update the PR. Please check.

tools/run-tests.py
galpeter: Maybe we can restrict it a follow up PR?

galpeter:
lgtm

PR #1658, opened 2017-03-14
Fix Date objects' YearFromTime helper for out-of-range time values

### LINKED ISSUE #1657: Hang in Date constructor
###### Jerry version:

```
Checked revision: b140158
Build: debug.linux
```
###### OS:

```
Ubuntu 16.04.2 LTS
```
###### Test case:
Running this test cases lasts unexpectedly long (killed after 2 minutes):

```javascript
new Date(-565224, 1)
```

###### Build command:
```
./tools/build.py --debug --clean
```

<sup>Found by  [Fuzzinator](http://fuzzinator.readthedocs.io/)</sup>

akosthekiss:
For negative out-of-range time values, ecma_date_year_from_time
fell into an infinite loop. (Moreover, for positive out-of-range
time values, it returned an incorrect year.) This patch fixes the
helper to return NaN in these cases (and ensures that its call
sites are prepared for NaN values).

Fixes #1657.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
@@ -119,6 +119,11 @@ ecma_date_year_from_time (ecma_number_t time) /**< time value */
   ecma_number_t year = (ecma_number_t) (1970 + 285616);
   ecma_number_t lower_year_boundary = (ecma_number_t) (1970 - 285616);
 
+  if (ecma_date_time_from_year (year) < time || ecma_date_time_from_year (lower_year_boundary) > time)
+  {
+    return ecma_number_make_nan ();
+  }
+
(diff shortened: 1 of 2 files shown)


zherczeg:
Could you add test cases for these ifs?

akosthekiss:
@zherczeg I'll be able to add the reported issue at least to exercise the year-related `if` (I'll have to check whether the other code paths cannbe triggered easily). However, will we be able to test it properly? I mean do we have testing infra for non-termination or timeout?

zherczeg:
As far as I remember the python testing system detects timeout, crash, etc.

LaszloLango:
Please add the test, that was mentioned in the issue you fixed.

akosthekiss:
I've added a test as requested and updated the patch.

akosthekiss:
@zherczeg according to my best understanding, they are

PR #1663, opened 2017-03-17
Fix minor inconsistencies in the API documentation

galpeter:
In the jerry_set_prototype part the method was described
as jerry_get_prototype. Minor typo fix.

The usage of jerry_set_property was incorrect. Each
call to the jerry_set_property returns a value which must be freed.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -2062,7 +2062,8 @@ handler (const jerry_value_t function_obj,
 
   // after this, script can invoke the native handler through "handler_field (1, 2, 3);"
   jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) "handler_field");
-  jerry_set_property (glob_obj, prop_name, func_val);
+  // set property and release the return value without any check
+  jerry_release_value (jerry_set_property (glob_obj, prop_name, func_val));
   jerry_release_value (prop_name);
 
   jerry_release_value (func_val);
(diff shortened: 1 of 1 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -1998,7 +1998,8 @@ handler (const jerry_value_t function_obj,
 
   // after this, script can invoke the native handler through "handler_field (1, 2, 3);"
   jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) "handler_field");
-  jerry_set_property (glob_obj, prop_name, func_val);
+  jerry_value_t set_result = jerry_set_property (glob_obj, prop_name, func_val);
```
zherczeg: This works as well if result is ignored:

    jerry_release_value (jerry_set_property (glob_obj, prop_name, func_val));

galpeter: @zherczeg That is true, however I wanted to emphasise that there is a returned value there. Also wanted to avoid multiple jerry api calls in one line to make the whole thing clearer. After all this is an API doc :)

galpeter: Should I change it to your format?

zherczeg: Add a comment for the short form (tell that if the result is not checked, it could be done in the following way).
LGTM after that.

galpeter:
@zherczeg I've updated and rebased the PR

PR #1664, opened 2017-03-17
Compact property pair list when too many deleted slots.

zherczeg:


zherczeg:
I realized it is dangerous to change the property chain during gc, so I limited the scope of the compact function for now.

LaszloLango:
@zherczeg dump test was removed. Please rebase the PR.

zherczeg:
I need to figure out what to do with this patch first.

PR #1666, opened 2017-03-19
Implement `Object.setPrototypeOf` from ES2015 specification

akosthekiss:
`Object.prototype.__proto__` has been implemented by most JS
engines to give R/W access to prototype chains, well before it made
it into the standard. JerryScript has decided not to implement it,
exactly because it was not part of ES 5.1. The only fully
ES5.1-compatible way of accessing the prototype chain is
`Object.getPrototypeOf` for reading.

However, ES2015 defines `Object.setPrototypeOf` for rewriting the
prototype chain, and JerryScript has now an ES2015 subset profile.
So, this commit adds its implementation to JerryScript.

Note, this commit does _not_ add `Object.prototype.__proto__`,
since that is in the Annex B of ES2015 specification, which is
optional for non-web-browser hosts.


jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -135,6 +135,141 @@ ecma_builtin_object_object_get_prototype_of (ecma_value_t this_arg, /**< 'this'
...
+
+      /* b. */
+      ecma_value_t p_value = ecma_make_object_value (p_p);
+      if (ecma_op_same_value (p_value, o_value))
+      {
+        return false;
+      }
+
+      /* c.i. TODO: es2015-subset profile does not support having a different
+       * [[GetPrototypeOf]] internal method */
```
jiangzidong: I think the spec means "if the object has its own specific [[getPrototypeOf]] method, we can break the loop", rather than "does not support "

akosthekiss: but, I mean, does jerry support (as of now) to have a non-standard (i.e., an object with its own specific) [[getPrototypeOf]] implementation?

jiangzidong: you're right, for now, jerry-es2015-subset doesnt support specific getprototypeof method.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -135,6 +135,141 @@ ecma_builtin_object_object_get_prototype_of (ecma_value_t this_arg, /**< 'this'
...
+    {
+      ret_value = ecma_copy_value (obj_value);
+    }
+    else
+    {
+      /* 5. */
+      bool status = ecma_set_prototype_of (obj_value, arg2);
+
+      /* 6. TODO: es2015-subset profile does not support having a different
+       * [[SetPrototypeOf]] internal method */
```
jiangzidong: the comments should be [[GetPrototypeOf]], right?

jiangzidong:
LGTM

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
akosthekiss: no. the spec of `Object.setPrototypeOf ( O, proto )` at ES2015 19.1.2.18 (http://www.ecma-international.org/ecma-262/6.0/#sec-object.setprototypeof) prescribes the following steps:

5. Let `status` be `O.[[SetPrototypeOf]](proto)`.
6. `ReturnIfAbrupt(status)`.

but we only have the non-specific `[[SetPrototypeOf]]` implementation right now, which cannot return anything but true/false (i.e., it cannot signal any abortion, cannot throw). thus, step 6 is a no-operation.

as soon as there is a more complex logic in step 5 and we can call something different than `ecma_set_prototype_of`, we will have to flesh out step 6

jiangzidong: I see, thanks for your explaination

zherczeg:
What is the problem with the current setprototypeof?

akosthekiss:
@zherczeg Well, the foremost problem is that it does not exist. There are only two C functions that set an object's prototype: `ecma_create_object` and `jerry_set_prototype`. The first is supporting object creation in the engine with a specified proto object, while the second is an engine API function which allows roughly anything for the API user (i.e., whoever is embedding JerryScript).

However, `jerry_set_prototype` is not and should not be used in the implementation of the JS function `Object.setPrototypeOf`. The reasons are two-fold: 1: API functions should use internal engine functions not the way around, and 2: `jerry_set_prototype` is in no way ES2015 standard compliant. 

(And 3: I would not suggest making it ES2015 compliant, at least not necessarily. It is a tool that the API user may use for whatever tweaks s/he finds necessary without any standard constraints.)

zherczeg:
Strange, I thought it is so basic that it has already there.

I would not make it es2015 specific, it is trivial and so basic.

akosthekiss:
We haven't allowed anything non-standard into the ES5.1 profile until now, would not be good to start it now.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -135,6 +135,141 @@ ecma_builtin_object_object_get_prototype_of (ecma_value_t this_arg, /**< 'this'
...
+  JERRY_ASSERT (ecma_is_value_object (v_value) || ecma_is_value_null (v_value));
+
+  ecma_object_t *o_p = ecma_get_object_from_value (o_value);
+
+  /* 3. */
+  ecma_object_t *current_p = ecma_get_object_prototype (o_p);
+  ecma_value_t current_value = ecma_make_object_value (current_p);
+
+  /* 4. */
+  if (ecma_op_same_value (v_value, current_value))
```
zherczeg: The ecma_op_same_value is a very slow function. In this case a simple == would be enough.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -135,6 +135,127 @@ ecma_builtin_object_object_get_prototype_of (ecma_value_t this_arg, /**< 'this'
   return ret_value;
 } /* ecma_builtin_object_object_get_prototype_of */
 
+#ifndef CONFIG_DISABLE_ES2015_BUILTIN
```
LaszloLango: IMHO, this is too general name. It would be better to use something like `CONFIG_DISABLE_ES2015_BUILTIN_COMMON` or `CONFIG_DISABLE_ES2015_BUILTIN_OBJECT`

tests/jerry-test-suite/es2015/19/19.01/19.01.02/19.01.02-001.js
```diff
@@ -0,0 +1,41 @@
...
+  catch (e)
+  {
+    name = e.name;
+  }
+
+  assert(name === "TypeError");
+
+  if (msg)
+  {
+    print(msg + " PASS (XFAIL)");
```
LaszloLango: Please remove the print function calls from the tests.

akosthekiss: I'd very much stick with these. They don't cause problems nor do they add complexity but give a very fast feedback if one of the tests fail.

(Note that these prints are swallowed by the testing infrastructure if the whole test passes so the output is not cluttered. But if some part of the test fails, it gives an immediate feedback on how far the test got, what parts succeeded and what was the first part to crash or assert out.)

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
akosthekiss: Motivation for this name came from `CONFIG_DISABLE_ANNEXB_BUILTIN`, which is a catch-all guard for miscellaneous extensions to the ES5.1 standard. They are not specific to a single object but add things "here-and-there". This guard `CONFIG_DISABLE_ES2015_BUILTIN` could do the same.

In some discussions, there was an idea that we could keep this guard as is and
1. change `CONFIG_DISABLE_ARRAYBUFFER_BUILTIN` to `CONFIG_DISABLE_ES2015_ARRAYBUFFER_BUILTIN` and `CONFIG_DISABLE_TYPEDARRAY_BUILTIN` to `CONFIG_DISABLE_ES2015_TYPEDARRAY_BUILTIN`, and
2. make all `CONFIG_DISABLE_ES2015_xxx_BUILTIN` guards dependent on this `CONFIG_DISABLE_ES2015_BUILTIN` guard (so, if ES2015 is disabled in general then there will be no need for disabling all ES2015-specific builtins one by one).

(But this would happen in a follow-up commit.)
Would that work?

LaszloLango:
LGTM, let's do the further work in followup patches.

PR #1668, opened 2017-03-20
Style fix: return value comments

polaroi8d:


robertsipka:
LGTM (informally)

PR #1669, opened 2017-03-21
Build fix for 32-bit float ecma-number storage

robertsipka:


jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
```diff
@@ -333,7 +333,7 @@ ecma_date_daylight_saving_ta (jerry_time_zone_t *tz, /**< time zone information
    * TODO: Fix daylight saving calculation.
    *       https://github.com/jerryscript-project/jerryscript/issues/1661
    */
-  return tz->daylight_saving_time * ECMA_DATE_MS_PER_HOUR;
+  return (ecma_number_t) tz->daylight_saving_time * ECMA_DATE_MS_PER_HOUR;
```
zherczeg: Perhaps it would no be allowed to use the date with float32 since even the current date can not be represented.

zherczeg: My point is:

- There is not much point to use a date which works dates only between 1960-1980.
- People will be unaware of this fact.

So it would be better to add a # error that date does not support float32 at the moment. If we would rewrite date to use uint64 that would fix the problem.

LaszloLango: I agree with @zherczeg

robertsipka:
Thanks, I've updated the PR.

PR #1672, opened 2017-03-21
Unify timeout mechanism in test runners and ensure that they work on OSX

akosthekiss:
On OSX, the traditional GNU `timeout` command is available as
`gtimeout`. So, this commit adds some logic to the test runners to
try and find the right command name on the host.

Moreover, the commit also unifies the timeout mechanism across all
test runners by rewriting occurrences of `ulimit`-based timeouting
to use the `[g]timeout` command.

### DIFF
--- a/tools/runners/run-test-suite-test262.sh
+++ b/tools/runners/run-test-suite-test262.sh
@@ -19,6 +19,11 @@ PATH_TO_TEST262="$2"
 OUTPUT_DIR=`dirname $ENGINE`
 REPORT_PATH="${OUTPUT_DIR}/test262.report"
 TIMEOUT="90s"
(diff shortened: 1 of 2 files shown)


tools/runners/run-test-suite-test262.sh
```diff
@@ -19,6 +19,11 @@ PATH_TO_TEST262="$2"
 OUTPUT_DIR=`dirname $ENGINE`
 REPORT_PATH="${OUTPUT_DIR}/test262.report"
 TIMEOUT="90s"
+TO=`which timeout`
```
LaszloLango: I'd prefer `TO_CMD` or `TIMEOUT_CMD`

akosthekiss:
I've updated the PR according to the review

PR #1678, opened 2017-03-22
Factor out common macro un/definitions used for built-in descriptions

akosthekiss:
The default definitions and undefinitions of macros used for built-in
descriptions (SIMPLE_VALUE, NUMBER_VALUE, STRING_VALUE, OBJECT_VALUE,
ROUTINE, ACCESSOR_READ_WRITE, ACCESSOR_READ_ONLY) are heavily cloned
all over the builtin-objects directory. This commit factors them out
into two header files, which are then included in the place of the
clones. This way, maintenance becomes a lot easier: e.g., if a new
macro gets introduced, default definition and undefinition don't have
to be added to all description files (of which there are 56 right now,
and their number will most probably just grow).

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-macro-defines.inc.h
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-macro-defines.inc.h
@@ -1,0 +1,42 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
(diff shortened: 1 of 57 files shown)


akosthekiss:
Commit metrics reveal that we had hundreds of lines of copy-paste code (clones) because of these preprocessing directives.

LaszloLango:
It is unusual to use `#include` at the end of files, but I like it. :) 1K lines removal is pretty impressive.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-macro-defines.inc.h
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef SIMPLE_VALUE
+# define SIMPLE_VALUE(name, simple_value, prop_attributes)
```
LaszloLango: Old code style. Please do not indent.

jiangzidong:
LGTM

akosthekiss:
I've updated the commit according to the review.

(As for the `#include`s at unusual places: we already use such approaches, actually the builtin descriptions are all examples for that. To highlight that these are unorthodox inclusions, these files have all `.inc.h` extensions instead of the solo `.h` of regular header files.)

PR #1688, opened 2017-03-24
Add C-API for the debugger.

zherczeg:


martijnthe:
There is no locking around the debugger state, so I'm assuming that the C API is supposed to be called only from the same thread that runs the engine. Is this assumption correct? If so, in other words, if the engine is stopped at a breakpoint (it's waiting in the `while (!jerry_debugger_receive ()) {}` loop for the client to send commands), the C API cannot be used to continue the execution (unless it's permitted to use it from another thread, which would require a mutex around the debugger state). I'm trying to understand the use case for this API.

I also noticed there is some duplication between `jerry-core/debugger/jerry-debugger.c` and the newly added `jerry-core/jerry-debugger[-ws].c`. Does it make sense to call the function in the new file from the old ones?

zherczeg:
Yes, the C api cannot influence the status when stopped at a breakpoint. The point is hiding certain JS runs from the debugger, not taking over the debugging.

I don't see any duplications.

martijnthe:
> Yes, the C api cannot influence the status when stopped at a breakpoint. The point is hiding certain JS runs from the debugger, not taking over the debugging.

OK, I see.

> I don't see any duplications.

Example:
https://github.com/zherczeg/jerryscript/blob/debugger_c_api/jerry-core/debugger/jerry-debugger.c#L351
vs
https://github.com/zherczeg/jerryscript/blob/debugger_c_api/jerry-core/debugger/jerry-debugger.c#L350
Doesn't it make sense to go through the debugger C API there?

zherczeg:
No. The function call would be a nop since `!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_BREAKPOINT_MODE)` is not satisfied.

jerry-core/jerryscript-debugger.h
```diff
@@ -0,0 +1,45 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRYSCRIPT_DEBUGGER_H
+#define JERRYSCRIPT_DEBUGGER_H
+
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
```
galpeter: Do we really need these includes here? Can't we have them only in the `.c` file?

jerry-core/jerryscript-debugger.c
```diff
@@ -0,0 +1,94 @@
...
+    JERRY_CONTEXT (debugger_flags) = (uint8_t) (JERRY_CONTEXT (debugger_flags) | JERRY_DEBUGGER_VM_IGNORE);
+  }
+#endif /* JERRY_DEBUGGER */
+} /* jerry_debugger_disable_stop_at_breakpoint */
+
+/**
+ * The engine is allowed to stop at breakpoints.
+ */
+void
+jerry_debugger_enable_stop_at_breakpoint (void)
```
galpeter: I was wondering if it would make sense to merge the `jerry_debugger_disable_stop_at_breakpoint` and `jerry_debugger_enable_stop_at_breakpoint` methods into one with a single enum argument to control the stop behaviour. What do you think? @LaszloLango?

LaszloLango: I can deal with both solution, but IMHO it is unnecessary to merge in this case, because we (probably) won't extend the enum. If we merge it than I'd use a bool argument rather than enum.

`void jerry_debugger_set_enable_stop_at_breakpoint (bool is_enable)`

zherczeg: I don't like bool parameters because they are generally unreadable. However in this case it might not that bad:

`jerry_debugger_stop_at_breakpoint (false)`

If we choose this, the other function should use a bool parameter as well.

zherczeg: However the other function pair `jerry_debugger_stop/jerry_debugger_continue` would look ugly with a boolean argument. They would only work with an enum.

We could also compact all into one function, but if an application only uses 1-2 enum, that would be a waste of binary size.

Actually these functions are small. I don't expect to become bigger later.

zherczeg: What about using a bool for `jerry_debugger_stop_at_breakpoint` and leave the other as is?

LaszloLango: good to me

galpeter:
lgtm

PR #1690, opened 2017-03-25
Generate lit-magic-strings.inc.h

akosthekiss:
Currently, `lit-magic-strings.inc.h` is manually maintained. This
has several drawbacks:

- keeping the list of magic strings sorted first by length then
  alphabetically is error prone,
- it is easy to leave unused magic strings in the list by accident
  (e.g., `LIT_MAGIC_STRING_JERRY_UL` is defined as a magic string
  but not used anywhere in the code) and,
- it is very hard to add `CONFIG_DISABLE_*_BUILTIN` guards to the
  list, even though there are several magic strings, which are used
  in some of the configurations only (e.g, "setPrototypeOf" is used
  in ES2015 only).

To ease the maintenance of magic strings, this commit moves the
definition of magic strings to a config file
(`lit-magic-strings.ini`), and adds `tools/gen-magic-strings.py` to
generate the `.inc.h` file from this config file and from the use
cases of the strings in the code.

- The magic strings in the config file can appear in any order, the
  generator will ensure that they are correctly sorted.
- The generator skips those definitions that are not used anywhere
  (and emits a warning to signal that such definitions can be
  removed).
- The generator applies the same guards to the definitions in the
  `.inc.h` file as found in the code around the use of the strings
  to optimize for size.

The commit also changes some builtin-related `.inc.h` files by
adding guards that don't affect functionality but improve the
results of the generator.


akosthekiss:
The effects of the change have been evaluated by building both master and this PR for ARM/Linux by
the following commands

```bash
./tools/build.py --clean --toolchain cmake/toolchain_linux_armv7l.cmake --profile minimal
./tools/build.py --clean --toolchain cmake/toolchain_linux_armv7l.cmake --profile es5.1
```

and then looking at the size of the binaries with `arm-linux-gnueabi-size`:

```
text	   data	    bss	    dec	    hex	filename
76640	      0	1575032	1651672	 1933d8	build/master-minimal/bin/jerry
74400	      0	1575032	1649432	 192b18	build/patch-minimal/bin/jerry
141668	     16	1575168	1716852	 1a3274	build/master-es51/bin/jerry
141324	     16	1575168	1716508	 1a311c	build/patch-es51/bin/jerry
```

This shows 2,240 bytes gain for the minimal profile and 344 bytes gain for the ES5.1 profile.

I consider the patch still as WIP, as the generator tool has only been tested with python 3
and is not made part of the build system but has to be executed manually every time the ini
file changes. Before moving further, I'd like to hear the opinion of others whether it is worth
finishing it up.

jerry-core/ecma/builtin-objects/ecma-builtin-array.inc.h
```diff
@@ -19,6 +19,8 @@
 
 #include "ecma-builtin-helpers-macro-defines.inc.h"
 
+#ifndef CONFIG_DISABLE_ARRAY_BUILTIN
+
```
zherczeg: If I understand correctly this files are only included if CONFIG_DISABLE_ARRAY_BUILTIN is not present. So adding it here seems pointless. Perhaps you can add an # error if present.

akosthekiss: These `#ifndef` guards are what the commit message is referring to by saying:

> The commit also changes some builtin-related .inc.h files by
> adding guards that don't affect functionality but improve the
> results of the generator.

So, yes indeed, they are "pointless". But for the script that generates the `lit-magic-strings.inc.h` file, they are of big help, because that script analyses each `.c` and `.h` file in the project separately, which means that it cannot infer the guard context that is on the including side of a header. Thus, if these guards are not added then the generator script sees all magic string references in these `.inc.h` files as unguarded and decides that their definition must always be present not to break a build.

Right now, I see no other approach that could work, other than writing a context-sensitive (include-tracking) analyser -- but writing something like that just for this task would be an overkill IMHO.

Repeating these guards in these `.inc.h` files is a bit of duplication, I agree, but they are not completely illogical either, and I'd say they don't seem to be too hard to keep in sync with the guards of the including source files.

LaszloLango:
Please measure the perf. and memory impact of this patch on RPi2.

jerry-core/ecma/builtin-objects/ecma-builtin-array.inc.h
zherczeg: So the generator manually check these .inc files? They are included only once, so we could detect these macros from the source which includes them, couldn't we?

akosthekiss:
Please, see all the results below:

Benchmark | Heap (bytes) | Stack (bytes) | RSS (bytes) | Perf (sec) |
----: | ----: | ----: | ----: | ----: | 
3d-cube.js | 16376 -> 16376 : 0.000% | 15120 -> 15120 : 0.000% | 53248 -> 53248 : 0.000% | 0.860 -> 0.856 : +0.394% | 
3d-raytrace.js | 115264 -> 115264 : 0.000% | 2612 -> 2612 : 0.000% | 140434 -> 132827 : +5.417% | 1.060 -> 1.055 : +0.452% | 
access-binary-trees.js | 32760 -> 32760 : 0.000% | 3192 -> 3192 : 0.000% | 53248 -> 53248 : 0.000% | 0.580 -> 0.576 : +0.723% | 
access-fannkuch.js | 5328 -> 5328 : 0.000% | 1440 -> 1440 : 0.000% | 20480 -> 24576 : -20.000% | 2.189 -> 2.188 : +0.008% | 
access-nbody.js | 6440 -> 6440 : 0.000% | 1408 -> 1408 : 0.000% | 28672 -> 28672 : 0.000% | 1.094 -> 1.098 : -0.415% | 
bitops-3bit-bits-in-byte.js | 840 -> 840 : 0.000% | 1408 -> 1408 : 0.000% | 16384 -> 20480 : -25.000% | 0.580 -> 0.581 : -0.216% | 
bitops-bits-in-byte.js | 792 -> 792 : 0.000% | 1408 -> 1408 : 0.000% | 16384 -> 20480 : -25.000% | 0.862 -> 0.868 : -0.691% | 
bitops-bitwise-and.js | 480 -> 480 : 0.000% | 1184 -> 1184 : 0.000% | 16384 -> 16384 : 0.000% | 1.000 -> 0.991 : +0.900% | 
bitops-nsieve-bits.js | 98464 -> 98464 : 0.000% | 1408 -> 1408 : 0.000% | 114688 -> 114688 : 0.000% | 1.366 -> 1.375 : -0.638% | 
controlflow-recursive.js | 824 -> 824 : 0.000% | 67720 -> 67720 : 0.000% | 81920 -> 86016 : -5.000% | 0.373 -> 0.377 : -0.865% | 
crypto-aes.js | 34280 -> 40952 : -19.463% | 2412 -> 2388 : +0.995% | 72557 -> 72557 : 0.000% | 0.963 -> 0.962 : +0.098% | 
crypto-md5.js | 106488 -> 106488 : 0.000% | 1804 -> 1804 : 0.000% | 131072 -> 135168 : -3.125% | 0.659 -> 0.660 : -0.119% | 
crypto-sha1.js | 66448 -> 66448 : 0.000% | 1480 -> 1480 : 0.000% | 86016 -> 87186 : -1.360% | 0.647 -> 0.644 : +0.451% | 
date-format-tofte.js | 16376 -> 16376 : 0.000% | 1968 -> 1968 : 0.000% | 40960 -> 40960 : 0.000% | 0.747 -> 0.748 : -0.090% | 
date-format-xparb.js | 16376 -> 16376 : 0.000% | 2392 -> 2392 : 0.000% | 40960 -> 45056 : -10.000% | 0.410 -> 0.411 : -0.087% | 
math-cordic.js | 1960 -> 1960 : 0.000% | 1408 -> 1408 : 0.000% | 16384 -> 24576 : -50.000% | 1.274 -> 1.270 : +0.286% | 
math-partial-sums.js | 1360 -> 1360 : 0.000% | 1352 -> 1352 : 0.000% | 16384 -> 20480 : -25.000% | 0.745 -> 0.756 : -1.460% | 
math-spectral-norm.js | 6248 -> 6248 : 0.000% | 1476 -> 1476 : 0.000% | 24576 -> 24576 : 0.000% | 0.579 -> 0.580 : -0.157% | 
string-base64.js | 90928 -> 90928 : 0.000% | 1408 -> 1408 : 0.000% | 143360 -> 143360 : 0.000% | 1.607 -> 1.598 : +0.523% | 
string-fasta.js | 16376 -> 16376 : 0.000% | 1384 -> 1384 : 0.000% | 32768 -> 32768 : 0.000% | 1.225 -> 1.215 : +0.828% | 
Geometric mean: | -0.893% | +0.050% | -7.197% | -0.002% | 

Binary sizes (bytes)
faa9655981:140256
ba7647e0f6:137684

I'm still about to find out what might have caused the increases in RSS and/or peak memory. As for RSS, my best guess is that they are just sampling differences because peak memory shows no change (e.g., see math-cordic.js). The only item I cannot explain right now is crypto-aes.js)

As for the binary size, the 2.5K drop is misleading, as it can be attributed mostly to an alignment in the linker script -- the real gain coming from the commit should still be around 300 bytes.

jerry-core/ecma/builtin-objects/ecma-builtin-array.inc.h
akosthekiss: The generator checks all `.c` and `.h` files in the project, `.inc.h` included.

And in theory, yes, we could detect such things by replicating a lot of the behaviour of a standard preprocessor. In practice, however, I don't intend to write a new preprocessor. The generator does not deal with (analyse, track) interfile dependencies (who includes whom, what are system includes, what are project includes, what are the include paths, etc. -- this would be just way too much). As mentioned above:

> writing a context-sensitive (include-tracking) analyser -- but writing something like that just for this task would be an overkill IMHO.

Actually, I did take a look at the preprocessor of gcc whether it could output something useful for our purpose (which lines are guarded by which preprocessor conditions) -- as preprocessors do/can output quite some info in addition to the preprocessed source --, but it turned out that nothing similar is available. (Most extra dumps are for helping makefile dependency tracking.)

zherczeg: There is no need for anything special. When you check a .inc file, you can grep the file which includes it, and check the first condition in the beginning of that file. Btw you can throw an error if multiple file includes an built-in inc file.

tools/gen-magic-strings.py
```diff
@@ -0,0 +1,240 @@
...
+
+            if str_ref not in results:
+                results[str_ref] = dict()
+            str_guards = results[str_ref]
+
+            if guard_tuple not in str_guards:
+                str_guards[guard_tuple] = list()
+            file_list = str_guards[guard_tuple]
+
+            file_list.append((fname, lnum))
```
akosthekiss: Right now, we aren't. This script started off as a checker, not as a generator, and in that phase it was mandatory/useful to be able to print out locations of violations (not just file names but line numbers as well). Then it turned out that checking/proving correctness is harder that generating the right thing, that's when this script became a generator. Now, line numbers are not that needed, but are still useful for debugging (the script itself), so I've kept them.

tools/gen-magic-strings.py
```diff
@@ -0,0 +1,240 @@
...
+            for str_ref, str_value in ini_parser.items('LIT_MAGIC_STRINGS')]
+    defs = sorted(sorted(defs, key=lambda ref_value: ref_value[1]),
+                  key=lambda ref_value: len(ref_value[1]))
+    return defs
+
+
+def extract_magic_string_references():
+    results = dict()
+
+    def process_line(fname, lnum, line, guard_stack):
```
akosthekiss: I've kept these. They helped me structure the code better without exposing local code to the global scope.

tools/gen-magic-strings.py
```diff
@@ -0,0 +1,240 @@
...
+ * limitations under the License.
+ */
+
+/*
+ * List of ECMA magic strings
+ *
+ * Generated by tools/gen-magic-strings.py
+ * DO NOT EDIT!!!
+ */"""
+    print(header, file=f)
```
akosthekiss: It was not trivial, but now the script has the same behaviour on py2.7 and py3.6

jerry-core/ecma/builtin-objects/ecma-builtin-array.inc.h
akosthekiss: You are oversimplifying here. The script is processing all files only once and does not start "grepping" other files all over the project directory structure to find out whether the current file is included elsewhere. And checking the first condition in the including file can be either a dangerous heuristic (in case some extra `#if`s sneak in the source file) or insufficient (we have at least one source where we have two guards in the source file (ecma-builtin-float64array-prototype.c)

```c
#ifndef CONFIG_DISABLE_ES2015_TYPEDARRAY_BUILTIN
#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
```

zherczeg: To tell the truth I am not fond of modifying a lot of files just to make a script work. I always work in the opposite way. But since these are just ifdefs, probably would not hurt.

akosthekiss:
After rebasing updated commit onto master, I've re-evaluated it on RPi2:

Benchmark | Heap (bytes) | Stack (bytes) | RSS (bytes) | Perf (sec) |
----: | ----: | ----: | ----: | ----: | 
3d-cube.js | 16376 -> 16376 : 0.000% | 15120 -> 15120 : 0.000% | 53248 -> 53248 : 0.000% | 0.860 -> 0.856 : +0.496% | 
3d-raytrace.js | 115264 -> 115264 : 0.000% | 2612 -> 2612 : 0.000% | 122880 -> 130486 : -6.190% | 1.056 -> 1.059 : -0.249% | 
access-binary-trees.js | 32760 -> 32760 : 0.000% | 3192 -> 3192 : 0.000% | 53248 -> 53248 : 0.000% | 0.580 -> 0.575 : +0.845% | 
access-fannkuch.js | 5328 -> 5328 : 0.000% | 1440 -> 1440 : 0.000% | 20480 -> 24576 : -20.000% | 2.181 -> 2.186 : -0.233% | 
access-nbody.js | 6440 -> 6440 : 0.000% | 1408 -> 1408 : 0.000% | 28672 -> 28672 : 0.000% | 1.090 -> 1.094 : -0.393% | 
bitops-3bit-bits-in-byte.js | 840 -> 840 : 0.000% | 1408 -> 1408 : 0.000% | 16384 -> 20480 : -25.000% | 0.580 -> 0.581 : -0.322% | 
bitops-bits-in-byte.js | 792 -> 792 : 0.000% | 1408 -> 1408 : 0.000% | 16384 -> 20480 : -25.000% | 0.862 -> 0.866 : -0.408% | 
bitops-bitwise-and.js | 480 -> 480 : 0.000% | 1184 -> 1184 : 0.000% | 16384 -> 16384 : 0.000% | 0.996 -> 0.999 : -0.333% | 
bitops-nsieve-bits.js | 98464 -> 98464 : 0.000% | 1408 -> 1408 : 0.000% | 114688 -> 114688 : 0.000% | 1.367 -> 1.374 : -0.577% | 
controlflow-recursive.js | 824 -> 824 : 0.000% | 67720 -> 67720 : 0.000% | 81920 -> 86016 : -5.000% | 0.374 -> 0.378 : -1.128% | 
crypto-aes.js | 40944 -> 40952 : -0.020% | 2388 -> 2412 : -1.005% | 70217 -> 73728 : -5.000% | 0.960 -> 0.965 : -0.537% | 
crypto-md5.js | 106488 -> 106488 : 0.000% | 1804 -> 1804 : 0.000% | 131072 -> 135168 : -3.125% | 0.658 -> 0.660 : -0.244% | 
crypto-sha1.js | 66448 -> 66448 : 0.000% | 1480 -> 1480 : 0.000% | 86016 -> 86016 : 0.000% | 0.645 -> 0.643 : +0.256% | 
date-format-tofte.js | 16376 -> 16376 : 0.000% | 1968 -> 1968 : 0.000% | 40960 -> 40960 : 0.000% | 0.745 -> 0.746 : -0.081% | 
date-format-xparb.js | 16376 -> 16376 : 0.000% | 2392 -> 2392 : 0.000% | 40960 -> 45056 : -10.000% | 0.410 -> 0.409 : +0.322% | 
math-cordic.js | 1960 -> 1960 : 0.000% | 1408 -> 1408 : 0.000% | 16384 -> 24576 : -50.000% | 1.270 -> 1.274 : -0.341% | 
math-partial-sums.js | 1360 -> 1360 : 0.000% | 1352 -> 1352 : 0.000% | 16384 -> 20480 : -25.000% | 0.747 -> 0.757 : -1.316% | 
math-spectral-norm.js | 6248 -> 6248 : 0.000% | 1476 -> 1476 : 0.000% | 24576 -> 24576 : 0.000% | 0.579 -> 0.579 : +0.062% | 
string-base64.js | 90928 -> 90928 : 0.000% | 1408 -> 1408 : 0.000% | 143360 -> 143360 : 0.000% | 1.594 -> 1.598 : -0.211% | 
string-fasta.js | 16376 -> 16376 : 0.000% | 1384 -> 1384 : 0.000% | 32768 -> 32768 : 0.000% | 1.219 -> 1.226 : -0.598% | 
Geometric mean: | -0.001% | -0.050% | -8.009% | -0.248% | 

Binary sizes (bytes)
da07252c45:140256
d174926fc7:137684

Interestingly, the crypto-aes peak memory increase has vanished. (Actually, I've experienced previously that crypto-aes has some memory usage stability issues but couldn't figure out the reason, yet.) The RSS increase remains.

jerry-core/ecma/builtin-objects/ecma-builtin-array.inc.h
akosthekiss: Most of the time, I agree with you. Therefore, I've made an experiment to build up the inclusion graph of all processed files and figure out how the inclusions are guarded. After getting the work done, I found that somehow the tool has not added the builtin guards for the `.inc.h` files, contrary to expectations.

It turns out that the reason is that the builtin `.inc.h`s are transitively and indirectly included. The source files define `BUILTIN_INC_HEADER_NAME` and the include `ecma-builtin-internal-routines-template.inc.h`, which in turn includes `BUILTIN_INC_HEADER_NAME` (which gets expanded by the preprocessor). I couldn't find a way to track such indirect inclusions in an included file.

I've reverted that experiment.

akosthekiss:
@zherczeg Should it be run manually or automatically, as part of the build process? If manually, where should the documentation go? In the generator script, in the generated output, or somewhere else?

zherczeg:
It could run during the build if the impact is minimal. If not we should create a new doc: `docs/08-MAINTENANCE.md` and goes there.

LaszloLango:
I think the generation should be done automatically during every build. I have mixed feeling about this PR. I see the advantages of it.

LaszloLango:
The idea of this PR is good to me. I think there were no objections, so please remove the "WIP" prefix when the PR is ready to review or close if you don't want to merge it.

akosthekiss:
I've rebased the patch to current master (and resolved conflicts). I've also aligned the generator script with other generators of `lit/*.inc.h` files (e.g., used the same "automatically generated by ... Do not edit!" comment).

Not sure, however, what documentation to add. Would a comment in `lit-magic-strings.ini` work? E.g., "After any change to this file, tools/gen-magic-strings.py must be executed!"

BTW, I haven't added the generator script to the cmake-based build system. The reason for this is that 1) there are no other cmake-executed generators in the project (except for `jerry-all-in.c`), and 2) some of the targets in the project don't use cmake at all, so they would break if we'd rely on a generated source.

akosthekiss:
As an experiment, I've added a commit on top of the original one. It does not add the generator script to the build system but adds an automatic check to the test system. So, if it turns out that something has changed in either the ini file or in the sources that would need the re-execution of the generator, the user will be warned (at least, by Travis, if the user forgets to run checks locally).

tools/check-generate.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
LaszloLango: Imho `check-magic-strings.sh` would be a better name

LaszloLango:
Please fix the pylint warnings

akosthekiss:
I've updated the PR according to the latest comments (file name change, pylint issues), squashed the commits, and rebased to current master.

akosthekiss:
OK, will add some extra notes ASAP. Will ask for a quick re-approve afterwards.

akosthekiss:
I've added a short note to `docs/04.INTERNALS.md` where magic strings are defined, and extended `tools/check-magic-strings.sh` to explicitly print what to do if it fails. Please, re-check.

LaszloLango:
still lgtm

zherczeg:
Still lgtm

PR #1692, opened 2017-03-27
Add missing tests for debugger

polaroi8d:
Add missing command tests for the debugger. Deleted the dump test, until we found a workround, because the compressed pointer and the associated offset is always different. 

- [x] backtrace
- [x] break
- [x] continue
- [ ] dump
- [x] delete
- [x] delete "all"
- [x] eval
- [x] help
- [x] list
- [x] next
- [x] quit
- [x] src
- [x] step


LaszloLango:
IMHO, the dump test case was intended to test the number and position of breakpoints. It would be enough for me if you could add a test that put a break on every possible breakpoint and checks the number of the inserted breakpoints.

PR #1459, opened 2016-11-29
Add the JerryScript Coding Standards

zherczeg:

PR #1695, opened 2017-03-29
Implement the basic Promise

jiangzidong:
* Implement the Promise Constructor and routine: 'then'
* add some promise test cases
* add queue_init and queue_run in `main-unix` for both file mode and repl mode

In the spec of Promise, there are  [PromiseCapability Record](http://www.ecma-international.org/ecma-262/6.0/#sec-promisecapability-records) and [PromiseReaction Record](http://www.ecma-international.org/ecma-262/6.0/#sec-promisereaction-records). I use ecma_object_t to represent them cause it can reuse the current GC system

In the record object, I use UINT32 type string to represent each property names. I choose uint32 string for saving space. **Or should I use `LIT_INTERNAL_MAGIC_STRING_XXX` instead?**

### Promise

In Spec: 

Internal Slot| Description
--- | ---
[[PromiseState]]| A String value that governs how a promise will react to incoming calls to its then method. The possible values are: "pending", "fulfilled", and "rejected".
[[PromiseResult]]| The value with which the promise has been fulfilled or rejected, if any. Only meaningful if [[PromiseState]] is not "pending".
[[PromiseFulfillReactions]]| A List of PromiseReaction records to be processed when/if the promise transitions from the "pending" state to the "fulfilled" state.
[[PromiseRejectReactions]]| A List of PromiseReaction records to be processed when/if the promise transitions from the "pending" state to the "rejected" state.
 
In Jerry
```
typedef struct
{
  ecma_extended_object_t ecma_extended_object_t; /**< extended object part */
  uint8_t state; /**< promise state, see ecma_promise_state_t */
  ecma_collection_header_t *fulfill_reactions; /**< list of PromiseFullfillReactions */
  ecma_collection_header_t *reject_reactions; /**< list of PromiseRejectReactions */
} ecma_promise_object_t;
```
And it is a CLASS type object, and its `ext_object_p->u.class_prop.u.value` stores the promiseResult
Each elements in `fulfill_reactions` and `reject_reactions` are also ecma_object.
Those internal property are not accessible by js code. so js users will not read/write those values via getter/setter of promise object

### PromiseReaction Record

It is the elements of promise.[[PromiseFulfillReactions]] and [[PromiseRejectReactions]]

In Spec:

Field Name|Value|Meaning
--- | --- | ---
[[Capabilities]]|	A PromiseCapability record	|The capabilities of the promise for which this record provides a reaction handler.
[[Handler]]|	A function object or a String|	The function that should be applied to the incoming value, and whose return value will govern what happens to the derived promise. If [[Handler]] is "Identity" it is equivalent to a function that simply returns its first argument. If [[Handler]] is "Thrower" it is equivalent to a function that throws its first argument as an exception.

In Jerry:

an GENERAL_TYPE object with following properties:
"0": the PromiseCapability object
"1": handler, could be function or boolean value. If it is boolean, true means "identity" and false means "thrower"

### PromiseCapability Records

In spec:

Field Name|Value|Meaning
--- | --- | ---
[[Promise]]	|An object	|An object that is usable as a promise.
[[Resolve]]	|A function object	|The function that is used to resolve the given promise object.
[[Reject]]	|A function object	|The function that is used to reject the given promise object.

In Jerry:

Similar with PromiseReaction , a general object to represent, with uint32_string properties

"0": the promise object
"1": resolve function
"2": reject function


jerry-core/ecma/builtin-objects/ecma-builtin-promise-prototype.inc.h
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef OBJECT_VALUE
+# define OBJECT_VALUE(name, obj_builtin_id, prop_attributes)
+#endif /* !OBJECT_VALUE */
```
zherczeg: The new model is calling this include pair:

    #include "ecma-builtin-helpers-macro-defines.inc.h"
    #include "ecma-builtin-helpers-macro-undefs.inc.h"

jerry-core/ecma/builtin-objects/ecma-builtin-promise-prototype.c
```diff
@@ -0,0 +1,85 @@
...
+#include "ecma-builtin-helpers.h"
+#include "ecma-builtins.h"
+#include "ecma-exceptions.h"
+#include "ecma-globals.h"
+#include "ecma-helpers.h"
+#include "ecma-objects.h"
+#include "ecma-promise-object.h"
+#include "ecma-try-catch-macro.h"
+#include "jrt.h"
+#include "jrt-libc-includes.h"
```
zherczeg: Do we need all of these includes?

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -0,0 +1,82 @@
...
+ */
+
+#include "ecma-builtins.h"
+#include "ecma-exceptions.h"
+#include "ecma-gc.h"
+#include "ecma-globals.h"
+#include "ecma-helpers.h"
+#include "ecma-promise-object.h"
+#include "ecma-try-catch-macro.h"
+#include "jrt.h"
```
zherczeg: Please drop unnecessary includes in all files.

jerry-core/ecma/operations/ecma-jobqueue.c
```diff
@@ -28,26 +34,221 @@
  */
 
 /**
+ * Create a PromiseReactionJob
+ *
+ * @return pointer to the PromiseReactionJob
+ */
+static ecma_job_promise_reaction_t *
+create_promise_reaction_job (ecma_value_t reaction, /**< PromiseReaction */
```
zherczeg: These functions do not have ecma_ or ecma_job_ prefix. Even if they are static functions.

jerry-core/ecma/operations/ecma-jobqueue.c
```diff
@@ -28,26 +34,221 @@
...
+ */
+static ecma_value_t
+process_promise_reaction_job (void *obj_p) /**< the job to be operated */
+{
+  ecma_job_promise_reaction_t *job_p = (ecma_job_promise_reaction_t *) obj_p;
+  ecma_object_t *reaction_p = ecma_get_object_from_value (job_p->reaction);
+
+  ecma_string_t *str_0 = ecma_new_ecma_string_from_uint32 (0);
+  ecma_string_t *str_1 = ecma_new_ecma_string_from_uint32 (1);
+  ecma_string_t *str_2 = ecma_new_ecma_string_from_uint32 (2);
```
zherczeg: Why do you use indices instead of names? Do they come from the standard?

jerry-core/ecma/operations/ecma-jobqueue.c
```diff
@@ -28,26 +34,221 @@
...
+
+  ecma_value_t handler_result;
+
+  if (ecma_is_value_boolean (handler))
+  {
+    /* 4-5. True indicates "identity" and false indicates "thrower" */
+    handler_result = ecma_copy_value (job_p->argument);
+  }
+
+  else
```
zherczeg: No newline before else

jerry-core/ecma/operations/ecma-jobqueue.c
```diff
@@ -28,26 +34,221 @@
...
+ */
+static ecma_value_t
+process_promise_resolve_thenable_job (void *obj_p) /**< the job to be operated */
+{
+  ecma_job_promise_resolve_thenable_t *job_p = (ecma_job_promise_resolve_thenable_t *) obj_p;
+  ecma_object_t *promise_p = ecma_get_object_from_value (job_p->promise);
+  ecma_promise_resolving_functions_t *funcs;
+  funcs = ecma_promise_create_resolving_functions (promise_p);
+
+  ecma_value_t argv[] = {funcs->resolve, funcs->reject};
```
zherczeg: Incorrect style.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+
+/** \addtogroup ecma ECMA
+ * @{
+ *
+ * \addtogroup ecmapromiseobject ECMA Promise object related routines
+ * @{
+ */
+
+/**
+ * Check if a object is promise
```
zherczeg: if an object

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+ *
+ * @return bool
+ */
+inline bool
+ecma_is_promise (ecma_object_t *obj_p) /**< points to object */
+{
+  lit_magic_string_id_t class_id = ecma_object_get_class_name (obj_p);
+
+  return class_id == LIT_MAGIC_STRING_PROMISE_UL;
```
zherczeg: `ecma_object_get_class_name` is a slow function. I would recommend simply check if it is an object with class.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+/**
+ * Set the PromiseResult of promise
+ */
+inline void
+ecma_promise_set_result (ecma_object_t *obj_p, /**< points to promise object */
+                         ecma_value_t result) /**< the result value */
+{
+  ecma_extended_object_t *ext_object_p = (ecma_extended_object_t *) obj_p;
+
+  ext_object_p->u.class_prop.u.value = ecma_copy_value_if_not_object (result);
```
zherczeg: What happens with the previous value? If no previous value is possible, add an assert please.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+
+/**
+ * Get the result of the promise
+ *
+ * @return ecma value of the result
+ */
+inline ecma_value_t
+ecma_promise_get_result (ecma_object_t *obj_p) /**< points to promise object */
+{
+  ecma_extended_object_t *ext_object_p = (ecma_extended_object_t *) obj_p;
```
zherczeg: Add `JERRY_ASSERT (ecma_is_promise (obj_p))` to the following functions. Never hurts.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+ecma_promise_set_state (ecma_object_t *obj_p, /**< points to promise object */
+                        uint8_t state) /**< the state */
+{
+  ecma_promise_object_t *promise_object_p = (ecma_promise_object_t *) obj_p;
+
+  promise_object_p->state = state;
+} /* ecma_promise_set_state */
+
+/**
+ * Check if the alreadyResolved is true
```
zherczeg: if alreadyResolved is true

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+
+/**
+ * Takes a collection of Reactions and enqueues a new PromiseReactionJob for each Reaction
+ *
+ * See also: ES2015 25.4.1.8
+ *
+ * @return ecma value
+ */
+static ecma_value_t
+trigger_promise_reactions (ecma_collection_header_t *reactions, /**< lists of reactions */
```
zherczeg: ecma_ prefixes are missing

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+  ecma_object_t *obj_p = ecma_get_object_from_value (promise);
+
+  JERRY_ASSERT (ecma_promise_get_state (obj_p) == ECMA_PROMISE_STATE_PENDING);
+
+  ecma_promise_set_result (obj_p, reason);
+
+  ecma_promise_object_t *promise_p = (ecma_promise_object_t *) obj_p;
+  ecma_value_t ret = trigger_promise_reactions (promise_p->reject_reactions, reason);
+  promise_p->reject_reactions = ecma_new_values_collection (NULL, 0, false);
+  /* Free all fullfill_reactions */
```
zherczeg: Dot after sentence.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+  ecma_free_values_collection (promise_p->reject_reactions, false);
+  promise_p->reject_reactions = ecma_new_values_collection (NULL, 0, false);
+
+  ecma_promise_set_state (obj_p, ECMA_PROMISE_STATE_FULFILLED);
+
+  return ret;
+} /* fulfill_promise */
+
+
+/**
```
zherczeg: Two newlines.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+ * Native handler for Promise Resolve Function
+ *
+ * See also: ES2015 25.4.1.3.2
+ *
+ * @return ecma value
+ */
+static ecma_value_t
+resolve_function_handler (const ecma_value_t function,
+                          const ecma_value_t this __attribute__((unused)),
```
zherczeg: We don't use GCC specific `__attribute__ ((unused))`. Please use JERRY_UNUSED

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,707 @@
...
+  jmem_heap_free_block (funcs, sizeof (ecma_promise_resolving_functions_t));
+} /* ecma_promise_free_resolving_functions */
+
+/**
+ * Create a promise object
+ *
+ * See also: ES2015 25.4.3.1
+ *
+ * @return ecma value of the new promise object
+ *         Returned value must be freed with ecma_free_value
```
zherczeg: returned instead of Returned

jerry-core/ecma/operations/ecma-jobqueue.c
jiangzidong: In spec, they are "record" instead of object. I use object to represent them so that I can use of the current gc system instead of creating my own.
The relationship between spec definition and my design are list in the first post https://github.com/jerryscript-project/jerryscript/pull/1695#issue-217750183

Only internal code will use those properties. I choose indices to save space and time (create string from c char is less effecient) But I realized maybe I can use **LIT_INTERNAL_MAGIC_STRING_XXX** to represent the name. What do you think about it?

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: no previous value is possible, I will add an assert.

jerry-core/ecma/operations/ecma-jobqueue.c
jiangzidong: Sorry, what is the suitable style for the array? I didn't find examples in #1459

zherczeg: https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/ecma/builtin-objects/ecma-builtin-json.c#L1365

jerry-core/ecma/operations/ecma-promise-object.c
zherczeg: My bad, this should not be changed.

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: I plan to remove the `ecma_copy_value_if_not_object` in the body. caller should do this before call the ecma_promise_set_result. It is similar with ecma_op_object_put.

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: what about `return ecma_object_class_is (obj_p, LIT_MAGIC_STRING_PROMISE_UL)`

jiangzidong:
@akosthekiss Hi, the patch can pass all check in ubuntu, but failed in macOS.
Could you give my some clues that which behaviors may be different in Mac and Ubuntu? I don't have mac in my hands, and I can't debug on that, but the clues may reminds me of the potential issue. Thanks

jiangzidong:
@akosthekiss I found the reason. I set no_pure functions as attr__pure by mistake. e.g.
```
/**
 * Set the PromiseState of promise.
 */
inline void __attr_pure___ __attr_always_inline___
ecma_promise_set_state (ecma_object_t *obj_p, /**< points to promise object */
                        uint8_t state) /**< the state */
{
  JERRY_ASSERT (ecma_is_promise (obj_p));

  ((ecma_promise_object_t *) obj_p)->state = state;
} /* ecma_promise_set_state */
```
it actually changes some values in the heap.

In Ubuntu, the issue is not revealed. The GCC might do different optimizations for those 2 OS.

akosthekiss:
@jiangzidong It's great that you've found the cause. Although I'm a bit confused as that function should count as pure (at least in release builds, when asserts are disabled), or do I get something wrong?

BTW, on macOS, gcc is not gcc but a wrapper around clang/llvm. So they may easily do things differently.

jerry-core/ecma/operations/ecma-jobqueue.h
```diff
@@ -25,6 +25,22 @@
  * @{
  */
 
+typedef struct
+{
+  ecma_value_t reaction;
+  ecma_value_t argument;
+} ecma_job_promise_reaction_t;
```
LaszloLango: missing comments

jerry-core/ecma/operations/ecma-jobqueue.h
```diff
@@ -25,6 +25,22 @@
...
+
+typedef struct
+{
+  ecma_value_t promise;
+  ecma_value_t thenable;
+  ecma_value_t then;
+} ecma_job_promise_resolve_thenable_t;
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+{
+  JERRY_ASSERT (ecma_is_promise (obj_p));
+
+  ((ecma_promise_object_t *) obj_p)->state = state;
+} /* ecma_promise_set_state */
+
+/**
+ * Check if alreadyResolved is true.
+ *
+ * @return bool
```
LaszloLango: Please add a better description to the return value.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+
+  ext_object_p->u.class_prop.u.value = ecma_make_boolean_value (value);
+} /* ecma_set_already_resolved_value */
+
+/**
+ * Take a collection of Reactions and enqueue a new PromiseReactionJob for each Reaction.
+ *
+ * See also: ES2015 25.4.1.8
+ *
+ * @return ecma value
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+
+  return ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
+} /* ecma_promise_trigger_reactions */
+
+/**
+ * Reject a Promise with a reason.
+ *
+ * See also: ES2015 25.4.1.7
+ *
+ * @return ecma value
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+
+  return ret;
+} /* reject_promise */
+
+/**
+ * Fulfill a Promise with a value.
+ *
+ * See also: ES2015 25.4.1.4
+ *
+ * @return ecma value
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+
+  return ret;
+} /* fulfill_promise */
+
+/**
+ * Native handler for Promise Reject Function.
+ *
+ * See also: ES2015 25.4.1.3.1
+ *
+ * @return ecma value
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+ */
+static ecma_value_t
+ecma_promise_reject_handler (const ecma_value_t function,
+                             const ecma_value_t this,
+                             const ecma_value_t argv[],
+                             const ecma_length_t argc)
```
LaszloLango: missing comments

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+ *
+ * See also: ES2015 25.4.1.3.2
+ *
+ * @return ecma value
+ */
+static ecma_value_t
+ecma_promise_resolve_handler (const ecma_value_t function,
+                              const ecma_value_t this,
+                              const ecma_value_t argv[],
+                              const ecma_length_t argc)
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+ *
+ * See also: ES2015 25.4.1.5.1
+ *
+ * @return ecma value
+ */
+static ecma_value_t
+ecma_promise_executor_handler (const ecma_value_t function,
+                               const ecma_value_t this,
+                               const ecma_value_t argv[],
+                               const ecma_length_t argc)
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,719 @@
...
+  ecma_value_t reject = ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
+  /* 2. String '0' indicates [[Capability]] of the function object. */
+  ecma_value_t capability = ecma_op_object_get (ecma_get_object_from_value (function), str_0);
+  /* 3. String '1' indicates [[Resolve]] of the capability. */
+  ecma_value_t resolve = ecma_op_object_get (ecma_get_object_from_value (capability), str_1);
+
+  if (resolve != ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED))
+  {
+    ret = ecma_raise_type_error (ECMA_ERR_MSG ("'resolve' function should be undefined."));
+    goto end_of_executor_function;
```
LaszloLango: I'm not completely against `goto`  statements, but I think it can be avoided here.

jiangzidong:
@akosthekiss I think the pure function means it has no side-effect. In the ecma_promise_set_state , it write the value in heap `((ecma_promise_object_t *) obj_p)->state` (in jerry-heap, and it is actually the bss section of the program). So it is not a pure function.

jerry-core/ecma/operations/ecma-promise-object.h
```diff
@@ -0,0 +1,86 @@
...
+typedef struct
+{
+  ecma_extended_object_t ecma_extended_object_t; /**< extended object part */
+  uint8_t state; /**< promise state, see ecma_promise_state_t */
+  ecma_collection_header_t *fulfill_reactions; /**< list of PromiseFullfillReactions */
+  ecma_collection_header_t *reject_reactions; /**< list of PromiseRejectReactions */
+} ecma_promise_object_t;
+
+
+
```
LaszloLango: Too many empty lines. One is enough :)

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: yes, I can remove this `goto`.
What about the `goto` in `ecma_promise_resolve_handler `? I think that one is needed, otherwise the code will be quite 'dirty'

LaszloLango: I saw that too, but I agree. The code would be much less readable if you eliminate that one.

jiangzidong:
@akosthekiss Sorry that I deleted my last comment by accident.
```
inline void __attr_pure___ __attr_always_inline___
ecma_promise_set_state (ecma_object_t *obj_p, 
                        uint8_t state) 
{
  JERRY_ASSERT (ecma_is_promise (obj_p));

  ((ecma_promise_object_t *) obj_p)->state = state; //!!!HERE, change the value of memory heap
}
```
I don't  think the above function is a pure function. It has side affect, `((ecma_promise_object_t *) obj_p)->state = state` this statement will change the outer context's value (it actually changes the global variable's value)

jerry-core/ecma/operations/ecma-jobqueue.c
zherczeg: Hm, numbers are ok for me. But you can use symbolic strings if you want.

jerry-core/ecma/operations/ecma-jobqueue.h
```diff
@@ -26,6 +26,28 @@
  */
 
 /**
+ * Description of the PromiseReactionJob
+ */
+typedef struct
+{
+  ecma_value_t reaction; /**< the PromiseReaction */
+  ecma_value_t argument; /**< argument for the reaction */
+} ecma_job_promise_reaction_t;
```
zherczeg: Do these needs to be public structures?

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+ *
+ * \addtogroup ecmapromiseobject ECMA Promise object related routines
+ * @{
+ */
+
+/**
+ * Check if an object is promise.
+ *
+ * @return true - if it is a promise object.
+ *         false - if it is not a promise object.
```
zherczeg: true - if the object is a promise
false - otherwise

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+ecma_get_already_resolved_bool_value (ecma_value_t already_resolved) /**< the alreadyResolved */
+{
+  JERRY_ASSERT (ecma_is_value_object (already_resolved));
+
+  ecma_object_t *already_resolved_p = ecma_get_object_from_value (already_resolved);
+  ecma_extended_object_t *ext_object_p = (ecma_extended_object_t *) already_resolved_p;
+
+  JERRY_ASSERT (ext_object_p->u.class_prop.class_id == LIT_MAGIC_STRING_BOOLEAN_UL);
+
+  return ext_object_p->u.class_prop.u.value == ecma_make_boolean_value (true);
```
zherczeg: Why this needs to be a boolean object, not a boolean value? Objects consumes much more memory.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+  ecma_free_values_collection (reactions, false);
+} /* ecma_promise_trigger_reactions */
+
+/**
+ * Reject a Promise with a reason.
+ *
+ * See also: ES2015 25.4.1.7
+ */
+static void
+reject_promise (ecma_value_t promise, /**< promise */
```
zherczeg: ecma_reject_primise

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+{
+  ecma_object_t *obj_p = ecma_get_object_from_value (promise);
+
+  JERRY_ASSERT (ecma_promise_get_state (obj_p) == ECMA_PROMISE_STATE_PENDING);
+
+  ecma_promise_set_result (obj_p, ecma_copy_value_if_not_object (reason));
+
+  ecma_promise_object_t *promise_p = (ecma_promise_object_t *) obj_p;
+  ecma_promise_trigger_reactions (promise_p->reject_reactions, reason);
+  promise_p->reject_reactions = ecma_new_values_collection (NULL, 0, false);
```
zherczeg: Can't this be a NULL value?

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+  /* Free all fullfill_reactions. */
+  ecma_free_values_collection (promise_p->fulfill_reactions, false);
+  promise_p->fulfill_reactions = ecma_new_values_collection (NULL, 0, false);
+
+  ecma_promise_set_state (obj_p, ECMA_PROMISE_STATE_REJECTED);
```
zherczeg: Qestion: is it possible that a callback triggers the same promise again? If so, this state set must occur after the assertion.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+  ecma_promise_set_result (obj_p, ecma_copy_value_if_not_object (value));
+
+  ecma_promise_object_t *promise_p = (ecma_promise_object_t *) obj_p;
+  ecma_promise_trigger_reactions (promise_p->fulfill_reactions, value);
+  promise_p->fulfill_reactions = ecma_new_values_collection (NULL, 0, false);
+  /* Free all reject_reactions. */
+  ecma_free_values_collection (promise_p->reject_reactions, false);
+  promise_p->reject_reactions = ecma_new_values_collection (NULL, 0, false);
+
+  ecma_promise_set_state (obj_p, ECMA_PROMISE_STATE_FULFILLED);
```
zherczeg: Same comments as above.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+
+  ecma_string_t *str_0 = ecma_new_ecma_string_from_uint32 (0);
+  ecma_string_t *str_1 = ecma_new_ecma_string_from_uint32 (1);
+  ecma_object_t *function_p = ecma_get_object_from_value (function);
+  /* 2. */
+  ecma_value_t promise = ecma_op_object_get (function_p, str_0);
+  /* 1. */
+  JERRY_ASSERT (ecma_is_promise (ecma_get_object_from_value (promise)));
+  /* 3. */
+  ecma_value_t already_resolved = ecma_op_object_get (function_p, str_1);
```
zherczeg: I don't really understand this part. The function_p is a public function where we define a "1" property? What ensures that we can define this property?

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+ecma_promise_create_resolving_functions (ecma_object_t *object_p) /**< the promise object */
+{
+  /* 1. */
+  ecma_value_t already_resolved = ecma_op_create_boolean_object (ecma_make_boolean_value (false));
+
+  ecma_string_t *str_0 = ecma_new_ecma_string_from_uint32 (0);
+  ecma_string_t *str_1 = ecma_new_ecma_string_from_uint32 (1);
+  /* 2. */
+  ecma_object_t *resolve_p;
+  resolve_p = ecma_op_create_external_function_object ((ecma_external_pointer_t) ecma_promise_resolve_handler);
```
zherczeg: This is again seems to be a case where we have fixed internal handlers, there is no point creating JS functions for them, just call them directly whenever they needed. Creating JS functions is costly, and calling them is slower than natively calling them.

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: because both promise's reject_function and resolve_function share the same "alreadyResolved" value. (the boolean object). so that the reject or resolve function will know whether the promise is processed by checking the "alreadyResolved".

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: Yes, it can. But if so, when we add reactions to the list, we have to create a collection when the value is NULL

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: No. But I will put the set after the assertion to reduce misunderstanding.

jerry-core/ecma/operations/ecma-jobqueue.h
jiangzidong: It is only used in ecma-jobquue.c, and i will put the definition into the source file.

jerry-core/ecma/operations/ecma-jobqueue.c
```diff
@@ -28,26 +31,220 @@
...
+                                            &(job_p->argument),
+                                            1);
+  }
+
+  ecma_value_t status;
+  /* 7. */
+  if (ecma_is_value_false (handler) || ECMA_IS_VALUE_ERROR (handler_result))
+  {
+    /* str '2' indicates [[Reject]] of Capability */
+    handler_result = handler_result & ~ECMA_VALUE_ERROR_FLAG;
```
jiangzidong: Thanks! I didn't noticed this api

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: 1). resolve_handler and reject_handler are not only used internally, in jobqueue's `ecma_process_promise_resolve_thenable_job`. it will be used as args of thenable's then function. And the then function there, could be implemented by developer's binding code or js code. So I think it is needed to create JS function for `ecma_promise_resolve_handler` and `ecma_promise_reject_handler`.

2). you're right in your [last comment](https://github.com/jerryscript-project/jerryscript/pull/1695#discussion_r109372202), because reject and resolve will be passed to the user's custom `then` function,  we cannot define a normal string property for `function_p`. I will use LIT_INTERNAL_MAGIC_STRING_XXX as the property name, so that outer user can not get access to the property. 

3).While the creation of `ecma_promise_executor_handler` is not necessary, and I will change it.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,711 @@
...
+ecma_value_t
+ecma_promise_new_capability (void)
+{
+  /* 3. */
+  ecma_object_t *capability_p = ecma_op_create_object_object_noarg ();
+  ecma_string_t *str_0 = ecma_new_ecma_string_from_uint32 (0);
+
+  /* 4. */
+  ecma_object_t *executor_p;
+  executor_p = ecma_op_create_external_function_object ((ecma_external_pointer_t) ecma_promise_executor_handler);
```
jiangzidong: Yes, I should directly call the executor instead of creating js function.

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: I should use LIT_INTERNAL_MAGIC_STRING_XXX  here to prevent outer user from getting access to the internal property.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -0,0 +1,718 @@
...
+  /* 1. */
+  ecma_value_t already_resolved = ecma_op_create_boolean_object (ecma_make_boolean_value (false));
+
+  ecma_string_t str_promise;
+  ecma_string_t str_already_resolved;
+  ecma_init_ecma_magic_string (&str_promise, LIT_INTERNAL_MAGIC_STRING_PROMISE);
+  ecma_init_ecma_magic_string (&str_already_resolved, LIT_INTERNAL_MAGIC_STRING_ALREADY_RESOLVED);
+  /* 2. */
+  ecma_object_t *resolve_p;
+  resolve_p = ecma_op_create_external_function_object ((ecma_external_pointer_t) ecma_promise_resolve_handler);
```
zherczeg: Are these functions can be called by users? They seem internal based on their name. I would prefer just directly call them.

jiangzidong:
>  Are these functions user supplied?
>  Are these functions can be called by users? They seem internal based on their name. I would prefer just directly call them.

This pair of functions (resolve function and reject function) are not supplied by user, but will pass to user as arguments. For JS users, they are standard js functions. for example
```
var a = new Promise(function(f, r)  //here f is the resolve function, and r is the reject funtion
{
// user totally control how to use these 2 functions
......
if (...)
{
  f (x); // if some condition meets, resolve the promise
} else {
  r (y); // otherwise, reject it
}
})
```

jerry-core/ecma/operations/ecma-promise-object.c
jiangzidong: Please see my comments below

jiangzidong:
Updated for the 2nd review

jerry-core/ecma/operations/ecma-jobqueue.c
```diff
@@ -28,26 +31,243 @@
...
+
+  return job_p;
+} /* ecma_create_promise_resolve_thenable_job */
+
+/**
+ * Free the heap and the member of the PromiseResolveThenableJob.
+ */
+static void
+ecma_free_promise_resolve_thenable_job (ecma_job_promise_resolve_thenable_t *job_p) /**< points to the
+                                                                                 *  PromiseResolveThenableJob */
```
LaszloLango: wrong indentation

jerry-core/jerry-port.h
```diff
@@ -142,6 +142,8 @@ double jerry_port_get_current_time (void);
 
 #ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
 
+#define JERRY_PORT_ENABLE_JOBQUEUE
```
zherczeg: Do we still need this define?

jiangzidong: The marco is for main-unix.c, since I added the jobqueue related code in it.  Also the jerry-port-default-jobqueue.c used the macro. 

They are not in the jerry-core, so the profile config doesn't work on them

jiangzidong: Does it make sense, can I merge the patch? @zherczeg

zherczeg: Ok, lets merge it, and we can address such things later.

PR #1699, opened 2017-03-30
Combine ARRAYBUFFER and TYPEDARRAY macros together

jiangzidong:
updated according to the 2nd idea of https://github.com/jerryscript-project/jerryscript/pull/1699#issuecomment-290325703

Since arraybuffer and typedarray will not exist separately, I plan to change ~~`CONFIG_DISABLE_ES2015_TYPEDARRAY_BUILTIN` to `CONFIG_DISABLE_ES2015_ARRAYBUFFER_BUILTIN`.~~
`CONFIG_DISABLE_ES2015_ARRAYBUFFER_BUILTIN` to `CONFIG_DISABLE_ES2015_TYPEDARRAY_BUILTIN`

~~In this way, `ArrayBuffer`, `TypedArray` and future `DataView` should be under the same macro, and be disabled/enabled simultaneously.~~


zherczeg:
Just a question: why do we choose arraybuffer instead of typed array which is a more popular name. I think many people are not aware that typed arrays are backed by array buffers.

jiangzidong:
Because both typedarray and dataview will use arraybuffer, so I plan to put those 3 features under arraybuffer macro.

But another solution is:
Now we use typedarray macro to enable both typedarray and arraybuffer
Later (when we implement dataview), we use dataview macro to enable dataview, and use `dataview_macro | typedarray_macro` to enable arraybuffer.
In this way, typedarray and dataview can be enabled separatedly.

zherczeg:
I wouldn't mind to have a single macro for the whole feature set. Anyway I like your second idea. There is no point to have arraybuffer only.

jiangzidong:
@zherczeg @LaszloLango updated.

LaszloLango:
still LGTM

PR #1700, opened 2017-03-30
Add coverity scan

robertsipka:


robertsipka:
@daeyeon Could you please check it?

.travis.yml
```diff
@@ -5,6 +5,7 @@ dist: trusty
 sudo: required
 
 env:
+  matrix:
```
LaszloLango: Is this label necessary?

robertsipka: Yes, because I've added the global key. It's required to take differentiate between those two cases.

README.md
```diff
@@ -2,6 +2,10 @@
 # JerryScript: JavaScript engine for the Internet of Things
 [![License](https://img.shields.io/badge/licence-Apache%202.0-brightgreen.svg?style=flat)](LICENSE)
 [![Build Status](https://travis-ci.org/jerryscript-project/jerryscript.svg?branch=master)](https://travis-ci.org/jerryscript-project/jerryscript)
+<a href="https://scan.coverity.com/projects/jerryscript-project">
```
daeyeon: For consistency, I think you could change this as follows:
``[Coverity Scan Build Status](https://scan.coverity.com/projects/jerryscript-project)`` :-)

robertsipka: @daeyeon : Thanks! I've updated it.

.travis.yml
LaszloLango: In that case please increase the indentation of "OPTS" lines below.

tools/run-tests.py
```diff
@@ -323,6 +324,12 @@ def run_buildoption_test(options):
 
     return ret
 
+def run_coverity_scan(options):
```
LaszloLango: Why is this change needed? Why `build_command: "tools/build.py --clean"` in travis.yml at line 37 isn't enough?

robertsipka: That would be simplier, yes. I think about it, but I'm not so familiar with the travis config, and it's needed by the `script` command in travis.yml at line 71. Do you have any suggestion how to avoid it?

robertsipka: @LaszloLango : I've updated this patch based on what we discussed offline.

PR #1704, opened 2017-03-30
Fix several pylint warning in debugger python client

polaroi8d:

### DIFF
--- a/jerry-debugger/jerry-client-ws.py
+++ b/jerry-debugger/jerry-client-ws.py
@@ -69,9 +69,12 @@ WEBSOCKET_FIN_BIT = 0x80
 def arguments_parse():
     parser = argparse.ArgumentParser(description="JerryScript debugger client")
 
-    parser.add_argument("address", action="store", nargs="?", default="localhost:5001", help="specify a unique network address for connection (default: %(default)s)")
-    parser.add_argument("-v", "--verbose", action="store_true", default=False, help="increase verbosity (default: %(default)s)")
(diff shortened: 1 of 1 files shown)


jerry-debugger/jerry-client-ws.py
```diff
@@ -169,6 +172,13 @@ def insert_breakpoint(self, args):
         else:
             set_breakpoint(self.debugger, args)
 
+    def disable_args(self, args):
+        if args != "":
+            print("Error: No argument expected")
```
robertsipka: Simple use `if args:`

jerry-debugger/jerry-client-ws.py
```diff
@@ -169,6 +172,13 @@ def insert_breakpoint(self, args):
...
+            return True
+        else:
```
robertsipka: The else statement is unnecessary, you can just return with false at the end of the function.

jerry-debugger/jerry-client-ws.py
```diff
@@ -469,7 +482,7 @@ def __init__(self, address):
 
         self.idx_format = "I"
 
-        logging.debug("Compressed pointer size: %d" % (self.cp_size))
+        logging.debug("Compressed pointer size: %d", self.cp_size)
```
zherczeg: What is the problem with this?

polaroi8d: There is no problem, but this is the PEP8 logging standard.

zherczeg: OK.

PR #1710, opened 2017-03-31
Remove unused functions

bzsolt:


jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -563,25 +563,6 @@ ecma_get_number_from_value (ecma_value_t value) /**< ecma value */
 } /* ecma_get_number_from_value */
 
 /**
- * Get uint32 value from an ecma value
- *
- * @return floating point value
- */
-uint32_t __attr_pure___
-ecma_get_uint32_from_value (ecma_value_t value) /**< ecma value */
```
zherczeg: Please check why this and the following function introduced with git annotate. I am surprised that they are not used.

zherczeg: I realized it was used for setting length property for arrays. That is not needed anymore and will never be needed again.

Issue #1681, opened 2017-03-23
[Proposal] Add `type` argument to native handle API


martijnthe:
So far, all projects I've looked at that use JerryScript's [native handle APIs], fail to implement proper validation on obtained native handle values. Zephyr.js, IoT.js, mbed.js, etc. are all affected by this problem.

Let me explain the problem by using an example from the mbedos5 port. The port defines a number of bindings that define classes of JavaScript objects which, on the native side, are backed by different C++ classes (in case of a C project, they would be backed by different types of structs).

For example, there is a [`AnalogIn` C++ class] for which a [`AnalogIn` JavaScript constructor and method bindings] are defined. Below a snippet for the `read()` method. There, `jerry_get_object_native_handle()` is used to get a pointer to the `AnalogIn` C++ instance that is (presumably) associated with the JS `AnalogIn` object on which `read()` is called:

```
// AnalogIn.prototype.read(num):
DECLARE_CLASS_FUNCTION(AnalogIn, read) {
    CHECK_ARGUMENT_COUNT(AnalogIn, read, (args_count == 0));

    // Extract native AnalogIn pointer
    uintptr_t ptr_val;
    jerry_get_object_native_handle(this_obj, &ptr_val);

    AnalogIn* native_ptr = reinterpret_cast<AnalogIn*>(ptr_val);

    float result = native_ptr->read();
    return jerry_create_number(result);
}
```

The problem arises when `this` gets bound to a JS object with no native handle, or even worse, a native handle containing a pointer to a *different* C++ class instance / C struct type. For example:

```
// d's native handle is a pointer to C++ instance of class `DigitalOut`:
var d = new DigitalOut(1);

// i's native handle is a pointer to C++ instance of class `AnalogIn`:
var i = new AnalogIn(2);

// Whoops.. the read() binding will interpret the native handle as a pointer to an
// `AnalogIn` instance, while it's actually a `DigitalOut`... Unsafe/invalid cast! BAD!
i.read.call(d);
```

Note this is not only applies to getting native handles for `this`, but also for arguments to a function. Also note that checking `this.constructor` isn't bullet proof, because it can be changed from the JS code.

## TL;DR

So, in short, one would *always* have to add type checks to make sure that the native handle is actually of the expected type, before casting and dereferencing the handle's value.

Because one would always have to deal when using the native handle API and because apparently the issue is overlooked by many dependent projects, I would argue that this concern should be addressed by JerryScript's API, by adding a type field to the API.

[...]

PR #1711, opened 2017-04-01
Add 'type' argument to set/get native handle API

jiangzidong:
ecma-object have native handle type inside, so that binding code could use the type info to validate native handle's type.

#1681 talks about the purpose of this patch,

~~**Note:** This patch is not backward compatible, old binding code might cannot work after this patch. The easiest way is change  `jerry_get_object_native_handle` to `jerry_get_object_native_handle_deprecated`, and change `jerry_set_object_native_handle` to `jerry_set_object_native_handle_deprecated`.~~

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -167,12 +167,42 @@ typedef jerry_value_t (*jerry_external_handler_t) (const jerry_value_t function_
 
 **Summary**
 
-Native free callback of an object
+**Deprecated: Please use jerry_object_native_free_callback_t instead.**
+
+Native free callback of an object.
 
 **Prototype**
 
 ```c
 typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
+```
+
+## jerry_object_native_free_callback_t
+
+**Summary**
+
+Native free callback of an object. It is used in jerry_object_native_info_t.
+
+**Prototype**
+
+```c
+typedef void (*jerry_object_native_free_callback_t) (void *native_p);
+```
+
+## jerry_object_native_info_t
+
+**Summary**
+
+The type infomation of the native pointer.
+It includes the free callback that will be called when associated JavaScript object is garbage collected. It can be left NULL in case it is not needed.
+
+**Prototype**
+
+```c
+typedef struct
+{
+  jerry_object_native_free_callback_t free_cb;
+} jerry_object_native_info_t;
 ```
 
 ## jerry_object_property_foreach_t
@@ -3173,6 +3203,8 @@ jerry_set_prototype (const jerry_value_t obj_val,
 
 **Summary**
 
+**Deprecated: Please use jerry_get_object_native_pointer instead.**
+
 Get native handle, previously associated with specified object.
 
 **Prototype**
@@ -3211,12 +3243,16 @@ jerry_get_object_native_handle (const jerry_value_t obj_val,
 
 - [jerry_create_object](#jerry_create_object)
 - [jerry_set_object_native_handle](#jerry_set_object_native_handle)
+- [jerry_get_object_native_pointer](#jerry_get_object_native_pointer)
+
 
 ## jerry_set_object_native_handle
 
 **Summary**
 
-Set native handle and an optional free callback for the specified object
+**Deprecated: Please use jerry_set_object_native_pointer instead.**
+
+Set native handle and an optional free callback for the specified object.
 
 *Note*: If native handle was already set for the object, its value is updated.
 
@@ -3260,6 +3296,146 @@ jerry_set_object_native_handle (const jerry_value_t obj_val,
 
 - [jerry_create_object](#jerry_create_object)
 - [jerry_get_object_native_handle](#jerry_get_object_native_handle)
+- [jerry_set_object_native_pointer](#jerry_set_object_native_pointer)
+
+
+## jerry_get_object_native_pointer
+
+**Summary**
+
+Get native pointer and its type information.
+The pointer and the type information are previously associated with the object by jerry_set_object_native_pointer.
+Users can check the pointer's type before processing it.
+
+**Prototype**
+
+```c
+bool
+jerry_get_object_native_pointer (const jerry_value_t obj_val,
+                                 void **out_native_p,
+                                 const jerry_object_native_info_t **out_info_p)
+```
+
+- `obj_val` - object value to get native pointer from.
+- `out_native_p` - native pointer (output parameter).
+- `out_info_p` - native pointer's type infomation (output parameter).
+- return value
+  - true, if there is native pointer associated with the object
+  - false, otherwise
+
+**Example**
+
+```c
+static void native_freecb (uintptr_t native_p)
+{
+  ... // free the native pointer
+}
+
+static const jerry_object_native_info_t type_info =
+{
+  .free_cb = native_freecb
+};
+
+{
+  jerry_value_t object;
+  uintptr_t native_set;
+
+  ... // receive or construct object and native_set value
+
+  jerry_set_object_native_pointer (object, native_set, &type_info);
+
+  ...
+
+  uintptr_t native_get;
+  const jerry_object_native_info_t *type_get_p;
+  bool is_there_associated_native = jerry_get_object_native_pointer (object, &native_get, &type_get_p);
+
+  if (is_there_associated_native)
+  {
+    if (out_info_p == &type_info)
+    {
+      ... // the type of object's native pointer is expected, and then process the native pointer.
+    }
+  }
+}
+```
+
+**See also**
+
+- [jerry_create_object](#jerry_create_object)
+- [jerry_set_object_native_pointer](#jerry_set_object_native_pointer)
+- [jerry_object_native_info_t](#jerry_object_native_info_t)
+
+
+## jerry_set_object_native_pointer
+
+**Summary**
+
+Set native pointer and an optional type information for the specified object.
+You can get them by calling jerry_get_object_native_pointer later.
+
+*Note*: If native pointer was already set for the object, its value is updated.
+
+*Note*: If a non-NULL free callback is specified in the native type information,
+        it will be called by the garbage collector when the object is freed.
+        The type info is always overwrites the previous value, so passing
+        a NULL value deletes the current type info.
+
+**Prototype**
+
+```c
+bool
+jerry_set_object_native_pointer (const jerry_value_t obj_val,
+                                 void *native_p,
+                                 const jerry_object_native_info_t *info_p)
+```
+
+- `obj_val` - object to set native pointer in.
+- `native_p` - native pointer.
+- `info_p` - native pointer's type infomation or NULL.
+
+**Example**
+
+```c
+static void native_freecb (uintptr_t native_p)
+{
+  ... // free the native pointer
+}
+
+static const jerry_object_native_info_t type_info =
+{
+  .free_cb = native_freecb
+};
+
+{
+  jerry_value_t object;
+  uintptr_t native_set;
+
+  ... // receive or construct object and native_set value
+
+  jerry_set_object_native_pointer (object, native_set, &type_info);
+
+  ...
+
+  uintptr_t native_get;
+  const jerry_object_native_info_t *type_get_p;
+  bool is_there_associated_native = jerry_get_object_native_pointer (object, &native_get, &type_get_p);
+
+  if (is_there_associated_native)
+  {
+    if (out_info_p == &type_info)
+    {
+      ... // the type of object's native pointer is expected, and then process the native pointer.
+    }
+  }
+}
+```
+
+**See also**
+
+- [jerry_create_object](#jerry_create_object)
+- [jerry_get_object_native_pointer](#jerry_get_object_native_pointer)
+- [jerry_object_native_info_t](#jerry_object_native_info_t)
 
 
 ## jerry_foreach_object_property
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -347,6 +347,47 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
 } /* ecma_gc_mark */
 
 /**
+ * Free the native handle/pointer by calling its free callback
+ */
+static void
+ecma_gc_free_native_pointer (ecma_property_t *property_p, /**< property */
+                             lit_magic_string_id_t id) /**< identifier of internal property */
+{
+  JERRY_ASSERT (property_p != NULL);
+
+  JERRY_ASSERT (id == LIT_INTERNAL_MAGIC_STRING_NATIVE_HANDLE
+                || id == LIT_INTERNAL_MAGIC_STRING_NATIVE_POINTER);
+
+  ecma_property_value_t *value_p = ECMA_PROPERTY_VALUE_PTR (property_p);
+  ecma_external_pointer_t native_p;
+  ecma_external_pointer_t free_cb;
+  void *package_p;
+
+  package_p = ECMA_GET_INTERNAL_VALUE_POINTER (void, value_p->value);
+
+  if (id == LIT_INTERNAL_MAGIC_STRING_NATIVE_HANDLE)
+  {
+    native_p = ((ecma_native_handle_package_t *) package_p)->handle_p;
+    free_cb = ((ecma_native_handle_package_t *) package_p)->free_cb;
+
+    if ((jerry_object_free_callback_t) free_cb != NULL)
+    {
+      ((jerry_object_free_callback_t) free_cb) ((uintptr_t) native_p);
+    }
+  }
+  else
+  {
+    native_p = ((ecma_native_pointer_package_t *) package_p)->native_p;
+    free_cb = *(ecma_external_pointer_t *) (((ecma_native_pointer_package_t *) package_p)->info_p);
+
+    if ((jerry_object_native_free_callback_t) free_cb != NULL)
+    {
+      ((jerry_object_native_free_callback_t) free_cb) ((void *) native_p);
+    }
+  }
+} /* ecma_gc_free_native_pointer */
+
+/**
  * Free specified object
  */
 void
@@ -357,28 +398,6 @@ ecma_gc_sweep (ecma_object_t *object_p) /**< object to free */
                 && object_p->type_flags_refs < ECMA_OBJECT_REF_ONE);
 
   bool obj_is_not_lex_env = !ecma_is_lexical_environment (object_p);
-
-  if (obj_is_not_lex_env)
-  {
-    /* if the object provides free callback, invoke it with handle stored in the object */
-
-    ecma_external_pointer_t freecb_p;
-    ecma_external_pointer_t native_p;
-
-    bool is_retrieved = ecma_get_external_pointer_value (object_p,
-                                                         LIT_INTERNAL_MAGIC_STRING_FREE_CALLBACK,
-                                                         &freecb_p);
-
-    if (is_retrieved && ((jerry_object_free_callback_t) freecb_p) != NULL)
-    {
-      is_retrieved = ecma_get_external_pointer_value (object_p,
-                                                      LIT_INTERNAL_MAGIC_STRING_NATIVE_HANDLE,
-                                                      &native_p);
-      JERRY_ASSERT (is_retrieved);
-
-      jerry_dispatch_object_free_callback (freecb_p, native_p);
-    }
-  }
 
   if (obj_is_not_lex_env
       || ecma_get_lex_env_type (object_p) == ECMA_LEXICAL_ENVIRONMENT_DECLARATIVE)
@@ -403,9 +422,20 @@ ecma_gc_sweep (ecma_object_t *object_p) /**< object to free */
 
       for (int i = 0; i < ECMA_PROPERTY_PAIR_ITEM_COUNT; i++)
       {
+        ecma_property_t *property_p = (ecma_property_t *) (prop_iter_p->types + i);
+        jmem_cpointer_t name_cp = prop_pair_p->names_cp[i];
+
+        /* Call the native's free callback. */
+        if (ECMA_PROPERTY_GET_NAME_TYPE (*property_p) == ECMA_STRING_CONTAINER_MAGIC_STRING
+            && (name_cp == LIT_INTERNAL_MAGIC_STRING_NATIVE_HANDLE
+                || name_cp == LIT_INTERNAL_MAGIC_STRING_NATIVE_POINTER))
+        {
+          ecma_gc_free_native_pointer (property_p, (lit_magic_string_id_t) name_cp);
+        }
+
         if (prop_iter_p->types[i] != ECMA_PROPERTY_TYPE_DELETED)
         {
-          ecma_free_property (object_p, prop_pair_p->names_cp[i], prop_iter_p->types + i);
+          ecma_free_property (object_p, name_cp, property_p);
         }
       }
 
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -196,6 +196,28 @@ typedef int32_t ecma_integer_value_t;
 typedef uintptr_t ecma_external_pointer_t;
 
 /**
+ * Representation for native handle package.
+ *
+ * Note: It is for the deprecated api:
+ *       jerry_get_object_native_handle and jerry_set_object_native_handle
+ */
+
+typedef struct
+{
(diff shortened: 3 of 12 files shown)


jerry-core/ecma/base/ecma-globals.h
```diff
@@ -196,6 +196,25 @@ typedef int32_t ecma_integer_value_t;
...
+} ecma_native_handle_package_t;
+
+/**
+ * Representation of the native handle package for deprecated api.
+ */
+typedef struct
+{
+  ecma_native_handle_package_t base; /**< the package's info_p is NULL */
+  ecma_external_pointer_t free_cb; /**< free callback of the native handle */
+} ecma_native_handle_package_old_t;
```
zherczeg: The description says `This patch is not backward compatible` but this seems a backward compatible part.

jerry-core/jerryscript.h
```diff
@@ -166,6 +166,29 @@ typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
...
+#define JERRY_NATIVE_HANDLE_INFO_FOR_CTYPE(c_type) _jerry_native_handle_info_##c_type
+
+/**
+ * Define a native handle typeinfo based on the C type and free callback.
+ */
+#define JERRY_DEFINE_NATIVE_HANDLE_INFO(c_type, native_free_cb) \
+  static const jerry_native_handle_info_t JERRY_NATIVE_HANDLE_INFO_FOR_CTYPE (c_type) = \
+  { \
+    .free_cb = (jerry_object_free_callback_t) native_free_cb \
+  }
```
zherczeg: I don't really get this part. What is the purpose of c_type here? How can I extend jerry_native_handle_info_t with other read-only members?

jerry-core/jerryscript.h
```diff
@@ -322,9 +345,19 @@ jerry_value_t jerry_get_object_keys (const jerry_value_t obj_val);
...
-                                     jerry_object_free_callback_t freecb_p);
+__attribute__((deprecated))
+bool jerry_get_object_native_handle_deprecated (const jerry_value_t obj_val, uintptr_t *out_handle_p);
+__attribute__((deprecated))
+void jerry_set_object_native_handle_deprecated (const jerry_value_t obj_val, uintptr_t handle_p,
+                                                jerry_object_free_callback_t freecb_p);
+
+bool jerry_get_object_native_handle (const jerry_value_t obj_val,
+                                     uintptr_t *out_handle_p,
+                                     jerry_native_handle_info_t **out_info_p);
```
martijnthe: `const jerry_native_handle_info_t **out_info_p` ?

jerry-core/jerryscript.h
```diff
@@ -322,9 +345,19 @@ jerry_value_t jerry_get_object_keys (const jerry_value_t obj_val);
 jerry_value_t jerry_get_prototype (const jerry_value_t obj_val);
 jerry_value_t jerry_set_prototype (const jerry_value_t obj_val, const jerry_value_t proto_obj_val);
 
-bool jerry_get_object_native_handle (const jerry_value_t obj_val, uintptr_t *out_handle_p);
-void jerry_set_object_native_handle (const jerry_value_t obj_val, uintptr_t handle_p,
-                                     jerry_object_free_callback_t freecb_p);
+__attribute__((deprecated))
```
martijnthe: `__attribute__((deprecated))` is a gnu/gcc thing. Does it make sense to have a `JERRY_DEPRECATED` define like so:

```
#ifdef __GNUC__
#  define JERRY_DEPRECATED __attribute__((deprecated))
#else
#  error "You need to define JERRY_DEPRECATED for your compiler."
#endif

File: jerry-core/ecma/base/ecma-globals.h (hunk 1 above)
Comment (martijnthe) [inline jerry-core/ecma/base/ecma-globals.h]: The old API is now called `jerry_[g|s]et_object_native_handle_deprecated()` instead. Therefore, it's technically not compatible because any code that relies on the old API would have to rename the calls (add `..._deprecated`).

File: jerry-core/ecma/base/ecma-helpers-external-pointers.c [hunk 5]
```diff
@@ -25,27 +25,35 @@
...
-                                       ecma_external_pointer_t ptr_value) /**< value to store in the property */
+ecma_create_native_handle_property (ecma_object_t *obj_p, /**< object to create property in */
+                                    ecma_external_pointer_t handle_p, /**< native handle */
+                                    ecma_external_pointer_t info_p, /**< native handle's type info */
+                                    ecma_external_pointer_t free_cb) /**< native handle's free callback*/
 {
-  JERRY_ASSERT (id == LIT_INTERNAL_MAGIC_STRING_NATIVE_HANDLE
-                || id == LIT_INTERNAL_MAGIC_STRING_FREE_CALLBACK);
+  bool is_deprecate = (info_p == (ecma_external_pointer_t) NULL);
+  void *package_p;
```
martijnthe: Can't this be `ecma_native_handle_package_t *package_p` and save the casting in lines 52-53?

jerry-core/jerryscript.h
martijnthe: My idea was to provide a macro to make it easy to create handle infos per C (struct) type:

```
// my_binding.c

typedef struct {
  int native_obj_state;
  ...
} MyNativeObj;

static void free_my_native_obj(uintptr_t handle);

JERRY_DEFINE_NATIVE_HANDLE_INFO(MyNativeObj, free_my_native_obj);
// ^-- creates `static const jerry_native_handle_info_t _jerry_native_handle_info_MyNativeObj`

... {
  MyNativeObj *native_obj = malloc(...);
  jerry_set_object_native_handle(jerry_value, (uintptr_t)native_obj,
                       &JERRY_NATIVE_HANDLE_INFO_FOR_CTYPE(MyNativeObj));
}

```

I think we need to add docstrings for the arguments and an example as well (probably a simple one in the docstrings?).

> How can I extend jerry_native_handle_info_t with other read-only members?

I did not take additional, custom struct members into account. Do you think this is so common that it's worth complicating the macro further to accommodate for this?
You can create a new struct type that wraps `jerry_native_handle_info_t` and add your fields there, i.e.:

martijnthe:
Various files in the mbedos5 target use the old API.
Probably a good idea to update those to use the new API.

cc @janjongboom

janjongboom:
@jiangzidong @martijnthe I'm opposed to this change. Literally every package that someone has created for the mbedos5 target depends on `jerry_get_object_native_handle`, thus all packages will break when this lands.

martijnthe:
@janjongboom What do you propose to move forward?

IMO, I think it's better to deprecate the current API (sooner or later) because it's just too easy to use it in an unsafe way (see explanation here: https://github.com/jerryscript-project/jerryscript/issues/1681).

We talked about adding a `..._ext` post-fix to the API names for the new API and keep the old functions and their names for now (but add the `deprecated` compiler attribute).

But then when removing the old functions, we're still stuck with the `..._ext`. I guess at that point we could make both names go through the new API.

janjongboom:
@martijnthe I'm happy marking it as deprecated in the docs (or even add a `#warn` directive), but we should not change the function signature of such a public API.

martijnthe:
@janjongboom OK, sounds good.
@jiangzidong Let's add `..._ext` to the new APIs?

jerry-core/jerryscript.h
jiangzidong: Ok, I will refer to the codes in `jrt.h`

jiangzidong:
Hi all, I will make it backward compatible, and add `_ext` to the new api. These two days are public holidays in China, and I will update the PR on Wednesday. Sorry for the delay.

jerry-core/jerryscript.h
zherczeg: Check Linux kernel, that is very common there.

Personally I don't like macros in public headers which hides basic operations. So hard to understand the code later. I want to see `static const jerry_native_handle_info_t` because its purpose is clear rather than `JERRY_DEFINE_NATIVE_HANDLE_INFO()`, which forces you to do extra searches in the code.

martijnthe: OK, fair enough, let's remove the macros then.

zherczeg: Thank you. Of course in the documentation we need to emphasize how people should use it. We need good examples as well.

jiangzidong:
Updated, please review it again. Thanks

jiangzidong:
update the API document

docs/02.API-REFERENCE.md
```diff
@@ -3260,6 +3281,144 @@ jerry_set_object_native_handle (const jerry_value_t obj_val,
...
+
+**Example**
+
+```c
+static void handle_freecb (uintptr_t native_p)
+{
+  ... // free the native handle
+}
+
+const jerry_native_handle_info_t type_info =
```
zherczeg: static const

jerry-core/ecma/base/ecma-globals.h
zherczeg: I have been thinking about this, since it increases the memory consumption of the backward compatible variant. What about introducing two internal string types, one for the old, and another for the new API? In this case both variants would only use two pointers (8 byte on 32 bit machines), and the memory consumption of old applications would not increase. Furthermore it would be enough to use a single ecma_native_handle_package_t.

This solution would allow applications to use independent external pointers for each object, but I don't think it is a problem. Rather it is just a bad design.

jerry-core/jerryscript.h
```diff
@@ -322,9 +329,19 @@ jerry_value_t jerry_get_object_keys (const jerry_value_t obj_val);
 jerry_value_t jerry_get_prototype (const jerry_value_t obj_val);
 jerry_value_t jerry_set_prototype (const jerry_value_t obj_val, const jerry_value_t proto_obj_val);
 
+__attribute__((deprecated))
 bool jerry_get_object_native_handle (const jerry_value_t obj_val, uintptr_t *out_handle_p);
+__attribute__((deprecated))
 void jerry_set_object_native_handle (const jerry_value_t obj_val, uintptr_t handle_p,
                                      jerry_object_free_callback_t freecb_p);
+
+bool jerry_get_object_native_handle_ext (const jerry_value_t obj_val,
```
zherczeg: Actually I never liked the "native handle" name, what about changing it to something else? Handle is usually an integer number which is connected to a resource (a variant of reference), but we expect direct pointers to the resource here. I would prefer `jerry_get_object_pointer` / `jerry_set_object_pointer`. Shorter, and emphasize the purpose more. But I am open to other suggestions.

jiangzidong: What about jerry_set_object_native_pointer? I think the word 'native' is needed here

zherczeg: Ok.

jiangzidong:
>  What about introducing two internal string types, one for the old, and another for the new API?

It may hurt the performance. 
* when an object get native handle, it doesn't know which property is valid.
* when object sweep, gc doesn't know which property contains the free_cb. It may need to get property twice.

zherczeg:
Each property getter checks only their corresponding type, so native_handle and native) pointer access different types. No double property reads.

With gc we could invent a clever thing, e.g. move the free callback call after the property free loop. Properties should not be accessed in the free callback anyway.

jerry-core/jerryscript.h
martijnthe: I agree with @jiangzidong that with the word "native" it may be a bit clearer as to what it does.
So I'd +1 `jerry_set_object_native_pointer` and `jerry_get_object_native_pointer`

zherczeg:
What I realized now that, it actually makes things easier, and the resulting code is nicer, and even faster. So why don't do it?

zherczeg:
I can do the rework after this patch is landed if you wish to land this without changes.

jiangzidong:
> Each property getter checks only their corresponding type, so native_handle and native) pointer access different types. No double property reads.

What about people call `jerry_set_object_native_pointer` *new api* and `jerry_set_native_handle` *old api* to the same js object, with 2 different native handle? Is it allowed that one js object can contains 2 different native handle?

zherczeg:
Yes. Preventing people to have two native pointers has extra costs, and it is not worth it. It is a bad design after all. If people assign two pointers and two callbacks to one object, and that cause issues, it is their problem. So I would completely separate the new and old api, and do not recommend to use both, but I would do nothing to prevent it.

jiangzidong:
I know what you mean, and in gc, we don't need to change the time of calling free_cb, just do
* 1, if there is a free_cb in old package property, call it
* 2, if there is a free_cb in new package property, call it
* 3, continue gc object ...

zherczeg:
We don't need to change but searching properties takes time, and I think we can optimize this code. There is a loop right after the search, we could move this logic there. Simply if we encounter a package property, we call the free block. Just add a check before the `ecma_free_property` call.

This is safe, since if there would be a reference to the object, it would not be freed. And the free callback dose not get a reference to the object. So doing anything with this object is impossible now. Therefore it is not relevant when we call the callback (during property free or before).

Perhaps to avoid too deep indentation we could add a static function, which will be inlined by the compiler.

jiangzidong:
@zherczeg the latest update:
* change api name to jerry_object_get/set_native_pointer
* totally separate the old api and new api. They owns 2 different internal property respectively

docs/02.API-REFERENCE.md
```diff
@@ -3260,6 +3281,144 @@ jerry_set_object_native_handle (const jerry_value_t obj_val,
...
+- [jerry_create_object](#jerry_create_object)
+- [jerry_set_object_native_pointer](#jerry_set_object_native_pointer)
+- [jerry_object_native_info_t](#jerry_object_native_info_t)
+
+
+## jerry_set_object_native_pointer
+
+**Summary**
+
+Set native pointer and an optional type information for the specified object
```
zherczeg: Dot after the sentence.

docs/02.API-REFERENCE.md
```diff
@@ -3260,6 +3281,144 @@ jerry_set_object_native_handle (const jerry_value_t obj_val,
...
+
+*Note*: If native pointer was already set for the object, its value is updated.
+
+*Note*: If a non-NULL free callback is specified in the native type information,
+        it will be called by the garbage collector when the object is freed.
+        The type info is always overwrites the previous value, so passing
+        a NULL value deletes the current type info.
```
zherczeg: I would add: If the type info must be preserved it is recommended to get it by `jerry_set_object_native_pointer` first.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -347,6 +347,46 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
 } /* ecma_gc_mark */
 
 /**
+ * Free the native handle/pointer by calling its free callback
+ */
+static void
+ecma_gc_free_native (ecma_property_t *property_p, /**< property */
```
zherczeg: ecma_gc_free_native_pointer

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -347,6 +347,46 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
...
+  ecma_property_value_t *value_p = ECMA_PROPERTY_VALUE_PTR (property_p);
+  ecma_external_pointer_t native_p;
+  ecma_external_pointer_t free_cb;
+  void *package_p;
+
+#ifdef ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY
+  package_p = value_p->value;
+#else /* !ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY */
+  package_p = ECMA_GET_NON_NULL_POINTER (void, value_p->value);
+#endif /* ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY */
```
zherczeg: There are two macros for this: ECMA_GET/SET_INTERNAL_VALUE_POINTER, please use them.

jerry-core/ecma/base/ecma-helpers-external-pointers.c
```diff
@@ -25,24 +25,24 @@
  */
 
 /**
- * Create internal property with specified identifier and store external pointer in the property.
+ * Create internal property with specified identifier and store native handle/pointer of th object.
```
zherczeg: th -> the

jerry-core/jerryscript.h
```diff
@@ -322,9 +329,19 @@ jerry_value_t jerry_get_object_keys (const jerry_value_t obj_val);
 jerry_value_t jerry_get_prototype (const jerry_value_t obj_val);
 jerry_value_t jerry_set_prototype (const jerry_value_t obj_val, const jerry_value_t proto_obj_val);
 
+__attribute__((deprecated))
```
zherczeg: I am not sure this is compatible with non-GCC compilers. I would not add this to a public header.

jerry-core/jrt/jrt.h
```diff
@@ -39,6 +39,9 @@
 #ifndef __attr_pure___
 # define __attr_pure___ __attribute__((pure))
 #endif /* !__attr_pure___ */
+#ifndef __attr_deprecated___
+# define __attr_deprecated___ __attribute__((deprecated))
+#endif /* !__attr_deprecated___ */
```
zherczeg: Is this used anywhere?

zherczeg:
Personally I like the idea of having the two API-s coexist, but if you really don't like it perhaps we could use some undefined bit (e.g. the writable bit of the data property) to represent which API is used. I can do this.

Please also squash the commits.

jerry-core/jrt/jrt.h
jiangzidong: no, i should remove it

docs/02.API-REFERENCE.md
zherczeg: I am curious whether chinese writing put dots (full stops) after sentences? :) It is frequently missing and perhaps this is a cultural difference.

jerry-core/jerryscript.h
jiangzidong: What's your suggestion about the "deprecated" attribute?

I refer the jrt.h, and the __attr_xxx there are derectly defined.

zherczeg: I have no good suggestion. Imagine a situation where a GCC compiled JerryScript library is linked by a TI compiler. This is possible because of the common binary object format. The jerryscript.h is not depend on jrt.h, so the project using the TI compiler can use it. However, this use case would not be possible anymore.

I see two options: remove it, or add a define somewhere at the beginning, which defines a macro depending on the compiler. Perhaps having this define would be useful later when more functions will become deprecated.

docs/02.API-REFERENCE.md
jiangzidong: Sorry, my bad. I know that each comments should be ended with dot, except for the comments for function arguments, right?

zherczeg: Yes. But if your language differs too much from latin languages, it might be easy to forget these rules. I suspect capital letters do not exists in chinese. I am just curious about the full stops :)

jiangzidong: If it is a whole sentence or in the article, we need the dot (it looks like `。`)
But usually I will treat the code comment as short messages, and forgot to add the dot.

BTW, many old comments in Jerry are not ended with dot.

docs/02.API-REFERENCE.md
jiangzidong: I change the summary part:
```
## jerry_set_object_native_pointer

**Summary**

Set native pointer and an optional type information for the specified object.
You can get them by calling jerry_get_object_native_pointer later.
```

docs/02.API-REFERENCE.md
zherczeg: It is true. We could spend time fixing them, but I think it is not worth it. However we should not introduce more. Anyway, this is a documentation, where complete sentences are truly mandatory.

docs/02.API-REFERENCE.md
```diff
@@ -175,6 +175,21 @@ Native free callback of an object
 typedef void (*jerry_object_free_callback_t) (const uintptr_t native_p);
 ```
 
+## jerry_object_native_info_t
+
+**Summary**
+
+The type infomation of the native handle. It includes the free callback of the native handle.
```
martijnthe: native handle => native pointer

"It includes the free callback that will be called when associated JavaScript object is garbage collected. It can be left `NULL` in case it is not needed."

jerry-core/jerryscript.h
```diff
@@ -322,9 +335,19 @@ jerry_value_t jerry_get_object_keys (const jerry_value_t obj_val);
...
 bool jerry_get_object_native_handle (const jerry_value_t obj_val, uintptr_t *out_handle_p);
+__attr_deprecated___
 void jerry_set_object_native_handle (const jerry_value_t obj_val, uintptr_t handle_p,
                                      jerry_object_free_callback_t freecb_p);
+
+bool jerry_get_object_native_pointer (const jerry_value_t obj_val,
+                                      uintptr_t *out_native_p,
+                                      const jerry_object_native_info_t **out_info_p);
+void jerry_set_object_native_pointer (const jerry_value_t obj_val,
+                                      uintptr_t native_p,
```
martijnthe: Should the type for `native_p` be `void *` now instead?
If yes, should we add a new type for the `free_cb` that passes a `void *` instead of a `uintptr_t`?

jerry-core/jerryscript.h
```diff
@@ -25,6 +25,12 @@ extern "C"
 {
 #endif /* __cplusplus */
 
+#ifdef __GNUC__
+#define __attr_deprecated___ __attribute__((deprecated))
+#else /* !__GNUC__ */
+/* TODO: for other compilers */
+#define __attr_deprecated___
```
martijnthe: Does it make sense to "namespace" this, for example `JERRY_ATTR_DEPRECATED` or `__jerry_attr_deprecated__`?
I can imagine other libraries also defining this and causing conflicts in projects.
At the very least, I would add an `#ifndef` around it.

tests/unit/test-api.c
```diff
@@ -115,6 +115,24 @@ handler_construct_freecb (uintptr_t native_p)
...
+ */
+#define JERRY_DEFINE_NATIVE_HANDLE_INFO(c_type, native_free_cb) \
+  static const jerry_object_native_info_t JERRY_NATIVE_HANDLE_INFO_FOR_CTYPE (c_type) = \
+  { \
+    .free_cb = (jerry_object_free_callback_t) native_free_cb \
+  }
+
+JERRY_DEFINE_NATIVE_HANDLE_INFO (bind1, handler_construct_freecb);
+JERRY_DEFINE_NATIVE_HANDLE_INFO (bind2, handler_construct_freecb);
+
```
martijnthe: 👍

jerry-core/jerryscript.h
jiangzidong: Yes, thanks

jerry-core/jerryscript.h
jiangzidong: the type of arguments in these new api should be `void *`
But I suggest we remains the type of the free_cb, the reasons are
* in order to backward compatible , we need to define another new function type
* then we need to provide another `jerry_dispatch_object_free_callback`

What do you think?

martijnthe: > in order to backward compatible , we need to define another new function type

Yeah, I think that's OK, no?

> then we need to provide another jerry_dispatch_object_free_callback

Hmm.. internally, the pointer to the callback itself is a `ecma_external_pointer_t` (aka `uintptr_t` and a cast to `jerry_object_free_callback_t` happens in `jerry_dispatch_object_free_callback` itself. I guess as long as `sizeof(uintptr_t) == sizeof(void *)` is true, it will work OK. AFAIK, `uintptr_t` and `void *` are meant to be of equal size. So, I think casting between the callback function pointer types should be fine in that case? I guess it's a bit dangerous if the signature of one would ever change. I'm not aware of a `_Static_assert` that would check function pointer signature "compatibility".

Alternatively, if you provide another `jerry_dispatch_object_free_callback`, is the compiler smart enough to merge the two together?

@zherczeg thoughts on this?

jiangzidong: Maybe we can get rid of the jerry_dispatch_object_free_callback, and directly call the free_cb in the `ecma_gc_free_native_pointer`. Because in ecma-gc can get access to jerryscript.h

martijnthe: 👍 
Yes, that's a good idea. There was only one call site to begin with.

martijnthe:
@jiangzidong re-squash?

@zherczeg still looking OK?

jiangzidong:
@martijnthe ，I will squash commits later so that @zherczeg can see what changes clearly.

zherczeg:
Still ok for me.

PR #1714, opened 2017-04-04
Update unicode ranges to the latest version of Unicode Standard

robertsipka:
binary size : 141300 -> 141916 bytes


tools/unicode_ranges.py
```diff
@@ -207,11 +207,14 @@ def read_categories(unicode_data_file):
     # These separator chars are not in UnicodeData-3.0.0.txt or not in Zs category
```
zherczeg: UnicodeData-3.0.0.txt -> 9

PR #1715, opened 2017-04-04
Refactoring several array buffer and typed array functions.

zherczeg:
Binary size unchanged (148056 bytes) but .text is reduced from 133088 -> 132936 = 152 bytes.

### DIFF
--- a/jerry-core/ecma/operations/ecma-typedarray-object.c
+++ b/jerry-core/ecma/operations/ecma-typedarray-object.c
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+#include <math.h>
+
 #include "ecma-typedarray-object.h"
 #include "ecma-arraybuffer-object.h"
 #include "ecma-function-object.h"
@@ -39,17 +41,8 @@
  * cast it from type to ecma_number_t
  * then assign to variale number
  */
-#define GET_ELEMENT(type, location, number) \
-  type v = *((type *) location); \
-  number = (ecma_number_t) v;
-
-/**
- * Cast the number from ecma_numbet_t into type,
- * and store the number in the location
- */
-#define SET_ELEMENT(type, location, number) \
-  int64_t v = (int64_t) number; \
-  *((type *) location) = (type) v;
+#define GET_ELEMENT(type, location) \
+  ((ecma_number_t) (*((type *) location)));
 
 /**
  * set typedarray's element value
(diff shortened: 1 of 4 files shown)


zherczeg:
@jiangzidong

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -178,7 +168,7 @@ set_typedarray_element (lit_utf8_byte_t *dst, /**< the location in the internal
     }
     case LIT_MAGIC_STRING_UINT32_ARRAY_UL:
     {
-      SET_ELEMENT (uint32_t, dst, value);
+      *((uint32_t *) dst) = (uint32_t) value;
```
jiangzidong: Why uint32_t is a special case? If the value is negative float, the conversion is undefined http://stackoverflow.com/questions/2490600/iphone-floats-cast-to-unsigned-ints-get-set-to-0-if-they-are-negative

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -60,62 +51,61 @@ static ecma_number_t
...
 
+#undef GET_ELEMENT
+
+/**
+ * Cast the number from ecma_numbet_t into type,
+ * and store the number in the location
+ */
+#define SET_ELEMENT(type, location, number) \
+  int32_t v = (int32_t) number; \
+  *((type *) location) = (type) v;
```
jiangzidong: castiong to int32_t is not sufficient. See the test case 22.02.01-015, 017, 018, 019.

we are casting a float number (the ecma_number_t) to interger, we should make sure the integral part of float number can be represented by the integer type. So I choose int64_t to make sure that.
```
When a finite value of real floating type is converted to an integer type other than _Bool, the fractional part is discarded (i.e., the value is truncated toward zero). If the value of the integral part cannot be represented by the integer type, the behavior is undefined.
```

http://c0x.coding-guidelines.com/6.3.1.4.html

zherczeg: After some thinking I realized this myself. Probably int64 cast is not enough as well, we need to follow the spec, but I want to add a fast case for it. The int64 conversion is very costly on ARM-32 CPUs.

jiangzidong: Could you tell me why int64 is not enough?

zherczeg: Consider an int32 array:

https://www.ecma-international.org/ecma-262/6.0/#table-49
https://www.ecma-international.org/ecma-262/6.0/#sec-toint32

    int = floor(abs(number))
    int32bit = int mod 2^32

So if you have a double value, which is greater than 2^63, it can still have bits which affect the 32 bit part, but the conversion results with 0x80...0, so it is converted to 0.

zherczeg: The trick is that `int` is still a double value, not an integer number.

zherczeg: I found the following:

 687 If the value of the integral part cannot be represented by the integer type, the behavior is undefined.50) 

http://c0x.coding-guidelines.com/6.3.1.4.html

I just realized that int32 and int64 conversions works differently on x86, and the standard reveals that most of these conversions are undefined, so we cannot rely on a specific behaviour.

PR #1718, opened 2017-04-06
Remove duplicated profile-check from build.py

bzsolt:
This check is performed in cmake-side too.
The README.md about the profiles is also updated.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -27,7 +27,6 @@ import settings
 BUILD_DIR = os.path.join(settings.PROJECT_DIR, 'build')
 DEFAULT_PORT_DIR = os.path.join(settings.PROJECT_DIR, 'targets/default')
 
-PROFILE_DIR = os.path.join(settings.PROJECT_DIR, 'jerry-core/profiles')
 DEFAULT_PROFILE = 'es5.1'
 
 def default_toolchain():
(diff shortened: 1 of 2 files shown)


tools/build.py
```diff
@@ -155,13 +154,7 @@ def generate_build_options(arguments):
...
     build_options.append('-DPORT_DIR=%s' % arguments.port_dir)
 
-    if os.path.isabs(arguments.profile):
-        profile = arguments.profile
-    else:
-        profile = os.path.join(PROFILE_DIR, arguments.profile + '.profile')
-
-    build_options.append('-DFEATURE_PROFILE=%s' % profile)
-
+    build_options.append('-DFEATURE_PROFILE=%s' % arguments.profile)
```
zherczeg: Isn't this a feature removal?

bzsolt: No, because this check is performed [here](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/CMakeLists.txt#L195).
```cmake
if (NOT IS_ABSOLUTE ${FEATURE_PROFILE})
  set(FEATURE_PROFILE "${CMAKE_CURRENT_SOURCE_DIR}/profiles/${FEATURE_PROFILE}.profile")
endif()
```

PR #1721, opened 2017-04-06
Fix .gitignore syntax

d3zd3z:
Change instances of **.ext in the .gitignore to just *.ext.  I think
the intent was to match these patterns in any directory, which is what
will happen with simple patterns.  The '**' construct is meant to
match zero or more full paths, but it must be the sole component of
the path.

Discovered by a warning from ripgrep.

JerryScript-DCO-1.0-Signed-off-by: David Brown <david.brown@linaro.org>

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,8 +4,8 @@ build/*
 
 # IDE related files
 nbproject
-**.sublime-project
-**.sublime-workspace
+*.sublime-project
+*.sublime-workspace
 .idea
 
 # Random Trash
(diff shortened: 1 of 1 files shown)


polaroi8d:
@d3zd3z the Travis failed on your DCO message. In my opinion you dont need the '<>' bracelets between your email.

d3zd3z:
Hmm, interesting choice, but ok, I'll fix that. Any idea why we chose a non-standard format for the tag? RFC-2822 is pretty well established, and the regular SoB uses it?

akosthekiss:
@d3zd3z Frankly, the root cause of deviation from the RFC is hidden in the mists of the past... That's just how it started when the project went open. The [wiki page](https://github.com/jerryscript-project/jerryscript/wiki/JerryScript-Developer's-Certificate-of-Origin) on DCO/SoB is almost two years old now. (Which does not mean it cannot be questioned, of course.)

Issue #1717, opened 2017-04-06
[Discussion] API to associate binding state (pointers) with the JerryScript context


martijnthe:
A while ago, `JERRY_CONTEXT` define and `jerry_context_t` had been added.
Before that change, all the fields in `jerry_context_t` were `static`s that were sprinkled throughout the JerryScript code base. This wasn't great for projects where multiple instances of JerryScript are needed, because `static`s are by definition "singleton".

Luckily, the addition of the `JERRY_CONTEXT` define made it easier to create multiple instances of JerryScript. Typically, to support multiple instances, one would provide a custom definition of `JERRY_CONTEXT`, for example to look up the current `jerry_context_t` in thread local storage.

However, often, binding code also needs to keep state regarding the native resources that it uses. In a multi-context scenario, each instance of JerryScript needs to have its own "instance" of this binding state as well. Therefore, using `static` variables cannot be used. Again, because `static` equals "singleton".

Using `__thread` or a platform-specific thread local storage (TLS) API could be used of course, but this would make the binding code less portable / re-usable.

Therefore, I think it make sense to try to come up with an API to associate binding state with a JerryScript context in an easy way that promotes re-use and portability of the binding code.

Some design goals that I have in mind for this an API to associate binding state with a JerryScript context:

- enable multiple bindings to each associate their own state with a JerryScript context.
- making binding code "portable" / re-usable: avoid `__thread` or a platform specific TLS API.
- should not require having to change JerryScript code (incl. `jcontext.[h|c]`).
- "scales down": no bloat when using a single-context (same speed/RAM/code size as today).
- easy to use from a developer point of view (probably not as easy as `static`, but hopefully close?)

I imagining something like a `void *user_binding_state[]` array in `JERRY_CONTEXT`, plus a couple of `jerry_...` functions to register/allocate/get/free the state for a particular binding. But I'm not quite sure how such a solution would "scale down".

Thoughts?

martijnthe:
cc @jiangzidong @zherczeg ideas?

zherczeg:
What about adding two functions: one for setting a global user pointer and another to getting this pointer?

martijnthe:
> What about adding two functions: one for setting a global user pointer and another to getting this pointer?

That would mean you'd have to put all the pieces of state from various bindings into one big struct: not great for keeping the various bindings modular.
[...]

Issue #1722, opened 2017-04-09
Feature request: Support __attribute__((constructor)) for easy module loading


gabrielschulhof:
Related to #1717 would it be possible to implement support for constructors (`__attribute__((constructor))`) in jerry's C library? I think that's where it needs to be implemented, in the `jerry-asm.S` file.

Having support for constructors makes it easy to register modules independently of the main program, since all you have to do is compile in a module and it will be available at runtime without any additional code.

akosthekiss:
@gabrielschulhof What is your use case (hardware, OS, toolchain, program)? Jerry-libc is a very limited/restricted libc variant. Are you really using it instead of the compiler's default libc?  (I'm not saying you shouldn't, but you have to know/accept that jerry-libc is really just the bare minimum, at least for now.)

gabrielschulhof:
The use case is simply to make mixing and matching different modules easy. That is, adding a new module should be as simple as adding a few source files to your project, without having to have those source files exports some well known entry point, and without having to modify your `main.c` (or whatever) to call those entry points whenever the module mix changes for your project.

Consider the following example. You make one small change to your Makefile:
```diff
...
diff a/Makefile b/Makefile
+obj-$(MY_COOL_MODULE) = my_cool_module.c
...
```

and now, without any further code changes, your project supports
```JS
var my_cool_module = require( "my_cool_module" );
```

Without this feature, you would need
```diff
...
diff a/Makefile b/Makefile
+obj-$(MY_COOL_MODULE) = my_cool_module.c
...
diff a/main.c b/main.c
+#ifdef BUILD_WITH_MY_COOL_MODULE
+#include "my_cool_module.h"
+#endif /* def BUILD_WITH_MY_COOL_MODULE */
...
+#ifdef BUILD_WITH_MY_COOL_MODULE
+my_cool_module_init();
+#endif /* def BUILD_WITH_MY_COOL_MODULE */
...
```
which is a lot harder to maintain.

Also, Node.js uses constructors to start the process of loading modules, so this would increase interoperability with Node.js native addons - a project I'm currently [working](https://github.com/gabrielschulhof/jerryscript/compare/master...n-api) [on](https://github.com/gabrielschulhof/zephyr.js/compare/master...node-api-wip) 😀

I'm fairly certain that I'm using jerry-libc when the constructor doesn't work, because if I turn off `-static` and `-nostdlib`, the constructor works and it links to `/lib64/libc.so.6`.

[...]

Issue #1716, opened 2017-04-05
[Proposal] Utilities for argument validation in external functions


martijnthe:
When implementing external functions, you might have written something like this:

```
// Native implementation of Xyz.prototype.doSomething(a, b)
// First arg is expected to be a string.
// Second (optional) arg is expected to be a boolean.
// `this` is expected to be bound to an instance of Xyz.
static jerry_value_t native_method_impl(const jerry_value_t func_val __attribute__((unused)),
                                        const jerry_value_t this_val __attribute__((unused)),
                                        const jerry_value_t *args_p __attribute__((unused)),
                                        const jerry_length_t args_cnt __attribute__((unused))) {
    if (args_cnt < 2) {
        return jerry_create_error(...);
    }

    // Validate and "transform" (copy to C string) argument 0:
    if (!jerry_value_is_string(args_p[0])) {
        return jerry_create_error(...);
    }
    char arg0[32];
    if (0 == jerry_string_to_utf8_char_buffer(arg[0], (jerry_char_t *)arg0, sizeof(arg0))) {
        // arg0 buffer too small
        return jerry_create_error(...);
    }

    // Validate and "transform" (copy to C bool) argument 1:
    bool arg1 = true; /* default value is true */
    if (args_cnt >= 1) {
        // Optional 2nd arg of type boolean
        if (!jerry_value_is_boolean(args_p[1])) {
            return jerry_create_error(...);
        }
        arg1 = jerry_get_boolean_value(args_p[1]);
    }

    // Validate `this`:
    struct native_obj_t *this_native_obj;
    if (!jerry_get_object_native_handle(this_val, (uintptr_t *)&this_native_obj)) {
        // Whoops, no native handle at all! Caller probably re-bound the method.
        return jerry_create_error(...);
    }
    if (!check_is_native_obj(native_obj)) {
        // Whoops, some other kind of struct! Caller probably re-bound the method.
        return jerry_create_error(...);
    }

    // yay, arguments validated (finally!)
    // Now, let's use this_native_obj, arg0 and arg1 to do something interesting!
    // ... method implementation here ...

    return jerry_create_undefined();
}

void xyz_init(void)
{
    // ...
    jerry_value_t do_something = jerry_create_external_function(native_method_impl);
    jerry_value_t rv = jerry_set_property (xyz_prototype, do_something_name, do_something);
    // ...
}

```

In short, when implementing a native external function, one may have to:
1. Check that mandatory arguments...
[...]

PR #1723, opened 2017-04-10
PoC: Implement basic per-context module data

gabrielschulhof:
Given an expected number of modules (JERRY_CONTEXT_MODULE_COUNT) which
can be defined at compile time, reserve two pointers in each context for
module-specific data: one void * for the data, and a function pointer
for the deleter to be called when the context is discarded.

The modules receive an index into the data array when they register.
They can later use the index to store data in the context.

Re #1717

@martijnthe is this like what you had in mind?
@zherczeg note that when JERRY_CONTEXT_MODULE_COUNT is defined as 1 then this is exactly the single pointer case.


gabrielschulhof:
... and calling `jerry_register_module()` from a library constructor would make module registration complete, very clean, and in line with what Node.js does (#1722).

jerry-core/jerry.c
```diff
@@ -2087,3 +2087,46 @@ jerry_dispatch_object_free_callback (ecma_external_pointer_t freecb_p, /**< poin
...
 } /* jerry_dispatch_object_free_callback */
+
+/**
+ * Reserve a slot in the context for a module
+ *
+ * Returns: 0 or positive on success, -1 on failure.
+ */
+
+/* This value is context-independent */
+static int module_index = -1;
```
jiangzidong: does "context-independent" means all context share the same index? if so, Why?

IMO, the index should be in jerry_context, and each context can have different modules.

gabrielschulhof: Because the index counts the modules that were registered in the process not the instances of the modules that were started in a context.

I originally had the index in the context but I realized that `jerry_register_module()` will only be called once for each module, no matter how many contexts are later created.

gabrielschulhof: So, like, slot 0 belongs to `canvas` and slot 1 belongs to `sqlite`, etc. When you start a new context, then slot 0 in that context will belong to the instance of `canvas` loaded into that context, and slot ` in that context will belong to `sqlite` loaded into that context, and the same for every subsequent context.

gabrielschulhof: I guess I should add a test with two trivial modules to test the data handholding.

jerry-core/jerryscript.h
```diff
@@ -344,6 +344,13 @@ size_t jerry_parse_and_save_literals (const jerry_char_t *source_p, size_t sourc
                                       uint32_t *buffer_p, size_t buffer_size, bool is_c_format);
 
 /**
+ * Module management
+ */
+int jerry_register_module(void);
+void *jerry_get_module_data(int index);
+void jerry_set_module_data(int index, void *data, void (*deleter)(void *));
```
zherczeg: How do you know the module index in a module?

gabrielschulhof: global static, assigned only once when the constructor gets called (implemented ideally via `__attribute__((constructor))` but an exported function that is called from `main()` will work too).

gabrielschulhof:
Man `vera` is hard to find - not part of Fedora, at any rate ...

jerry-core/jcontext/jcontext.h
```diff
@@ -103,6 +113,7 @@ typedef struct
   uint8_t valgrind_freya_mempool_request; /**< Tells whether a pool manager
                                            *   allocator request is in progress */
 #endif /* JERRY_VALGRIND_FREYA */
+  jerry_module_data_t module_data[JERRY_CONTEXT_MODULE_COUNT];
```
martijnthe: My guess is we'll want to have a "feature-flag" (#define / #ifdef) around all of this so that people still have the option to go for the most bare bones config possible.

jerry-core/jerry.c
martijnthe: Yeah, in the proposal that arose from #1717, module indices are global (and assigned once per boot-up).
That means that if multiple contexts use different sets of modules, there will be unused slots in the `module_data` array. I think this is probably acceptable.

The alternative is to add another per-context module registration step and another level of indirection. That would also add extra memory, so I'm not convinced that at the bottom line this would be better compared to having a couple unused slots.

gabrielschulhof:
OK, I dunno how to fix the rest of the linting errors 🤷

zherczeg:
I think native module management should not be part of JerryScript, which is an ECMAScript engine. This should be done in the utility library, where Standard Modules could use a nice generic interface. This could be one implementation of the interface. What do you think?

jerry-core/jerry.c
jiangzidong: I see, it makes sense.

zherczeg:
@gabrielschulhof please check the discussion in #1717

gabrielschulhof:
@zherczeg so ... move to a new `jerry-utilities/module`?

zherczeg:
In #1716 @martijnthe proposed 3 places for such utilities. Perhaps you could also vote.

LaszloLango:
Is this PR still valid?

PR #1724, opened 2017-04-10
Small improvements for `type` property support.

zherczeg:


zherczeg:
@jiangzidong

jerry-core/jerry.c
```diff
@@ -1979,14 +1954,54 @@ jerry_set_object_native_handle (const jerry_value_t obj_val, /**< object to set
...
+  if (ecma_is_value_object (obj_val))
+  {
+    ecma_object_t *object_p = ecma_get_object_from_value (obj_val);
 
-  ecma_create_native_handle_property (object_p,
-                                      handle_p,
-                                      (ecma_external_pointer_t) freecb_p);
+    ecma_create_native_handle_property (object_p,
+                                        (void *) handle_p,
+                                        (void *) (ecma_external_pointer_t) freecb_p);
```
jiangzidong: `(void *) freecb_p` should be enough.

zherczeg: Unfortunately not. The compiler complains that function ptr cannot be casted to void* according to ISO C99. You can convert it to an unsigned number, that is valid. This is really illogical, and I really hate that error message. Even PowerPC 64 old ABI has a workaround for this (each function pointer is a 3*64 bit record for the function address, r2 and r11 registers).

jiangzidong: Wow, new programming tips. Thanks. LGTM

PR #1725, opened 2017-04-10
Add support for init/fini arrays to libc

akosthekiss:
The arrays contain the addresses of functions annotated with
constructor or destructor attributes. The support is optional,
requires FEATURE_INIT_FINI cmake option to be set. As of now, the
option is _not_ available in tools/build.py directly, only via
`--cmake-param="-DFEATURE_INIT_FINI=ON"`.

### DIFF
--- a/jerry-libc/jerry-libc-init.c
+++ b/jerry-libc/jerry-libc-init.c
@@ -1,0 +1,98 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ * This file is based on work under the following copyright and permission
+ * notice:
+ *
+ *    Copyright (C) 2004 CodeSourcery, LLC
+ *
+ *    Permission to use, copy, modify, and distribute this file
+ *    for any purpose is hereby granted without fee, provided that
+ *    the above copyright notice and this notice appears in all
+ *    copies.
+ *
+ *    This file is distributed WITHOUT ANY WARRANTY; without even the implied
+ *    warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+ */
+
+#include "jerry-libc-defs.h"
+
+#ifdef ENABLE_INIT_FINI
+
+/* These magic symbols are provided by the linker.  */
+typedef void (*libc_init_fn_t) (void);
+
+extern libc_init_fn_t __preinit_array_start[] __attr_weak___;
+extern libc_init_fn_t __preinit_array_end[] __attr_weak___;
+extern libc_init_fn_t __init_array_start[] __attr_weak___;
+extern libc_init_fn_t __init_array_end[] __attr_weak___;
+extern libc_init_fn_t __fini_array_start[] __attr_weak___;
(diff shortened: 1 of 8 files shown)


akosthekiss:
This patch is a proposed solution for #1722.

The constructor/destructor behavior can be tested, e.g., by adding the below two functions anywhere in the compiled sources. (For my experiments, I've copied them into `jerry-main/main-unix.c`.)

```c
static void __attribute__ ((constructor))
start_jerry (void)
{
  printf ("JERRY STARTED\n");
}

static void __attribute__ ((destructor))
finish_jerry (void)
{
  printf ("JERRY FINISHED\n");
}
```

If configured properly, `start_jerry` and `finish_jerry` will be invoked even though they are not called anywhere explicitly. First example on ARM32.

```bash
$ ./tools/build.py --clean --cmake-param="-DFEATURE_INIT_FINI=ON" --toolchain cmake/toolchain_linux_armv7l.cmake
$ file ./build/bin/jerry
./build/bin/jerry: ELF 32-bit LSB  executable, ARM, EABI5 version 1 (SYSV), statically linked, BuildID[sha1]=5863cc6fdcdcf956d6910a266f7ce583fd20c995, stripped
$ qemu-arm-static ./build/bin/jerry ./tests/hello.js
JERRY STARTED
Hello JerryScript!
JERRY FINISHED
```

It also works on x86-64:

```bash
$ ./tools/build.py --clean --cmake-param="-DFEATURE_INIT_FINI=ON"
$ file ./build/bin/jerry 
./build/bin/jerry: ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), statically linked, BuildID[sha1]=73dba6c37fed7aa9aea19b533e76f1d2bfcf28e2, stripped
$ ./build/bin/jerry ./tests/hello.js 
JERRY STARTED
Hello JerryScript!
JERRY FINISHED
```

Sanity check: if the feature is not enabled in jerry-libc, the constructors/destructors don't run.

```bash
$ ./tools/build.py --clean
$ ./build/bin/jerry ./tests/hello.js
Hello JerryScript!
```

Note 1: Haven't tested on x86.

Note 2: Had to modify cppcheck suppression list, as cppcheck incorrectly reported `void *` arithmetics. Actually, arithmetics is performed on function pointers for which the return type is void, but that should be a completely different story, IMHO.

```
jerry-libc/jerry-libc-init.c:62: portability(arithOperationsOnVoidPointer): '__preinit_array_end' is of type 'void *'. When using void pointers in calculations, the behaviour is undefined.
```

akosthekiss:
An additional sanity check: the enhanced jerry-libc works just like the system libc (which needs no extra feature switches, of course).

```bash
$ ./tools/build.py --clean --jerry-libc=off --jerry-libm=off --static-link=off --link-lib="-lm"
$ file ./build/bin/jerry 
./build/bin/jerry: ELF 64-bit LSB  executable, x86-64, version 1 (SYSV), dynamically linked (uses shared libs), for GNU/Linux 2.6.24, BuildID[sha1]=138ef44c61b06bb6fe7f5b0b36be1994b6b4d2c0, stripped
$ ./build/bin/jerry ./tests/hello.js 
JERRY STARTED
Hello JerryScript!
JERRY FINISHED
```

gabrielschulhof:
Super-awesome 🎉

jerry-libc/jerry-libc-init.c
```diff
@@ -0,0 +1,90 @@
...
+extern libc_init_fn_t __preinit_array_end[] __attr_weak___;
+extern libc_init_fn_t __init_array_start[] __attr_weak___;
+extern libc_init_fn_t __init_array_end[] __attr_weak___;
+extern libc_init_fn_t __fini_array_start[] __attr_weak___;
+extern libc_init_fn_t __fini_array_end[] __attr_weak___;
+extern void _init (void);
+extern void _fini (void);
+
+
+/** No-op default _init. */
```
zherczeg: The function header comment is defferent for libc functions?

akosthekiss: I might have been sloppy. Will get it right asap.

akosthekiss:
I've fixed the function doc comments in the new file and updated the commit.

PR #1726, opened 2017-04-11
Fix the stlink binary path

polaroi8d:
Fix the Stlink path for the binary file. Add to checkout for a single commit hash, and we don't need to update the README all the time when something is changes.

### DIFF
--- a/targets/nuttx-stm32f4/README.md
+++ b/targets/nuttx-stm32f4/README.md
@@ -85,12 +85,12 @@ make
 
 Connect Mini-USB for power supply and connect Micro-USB for `NSH` console.
 
-To configure `stlink` utility for flashing, follow the instructions [here](https://github.com/texane/stlink#build-from-sources).
+To configure `stlink` utility for flashing, follow the instructions in the official [Stlink repository](https://github.com/texane/stlink).
 
 To flash,
 ```
 # assuming you are in nuttx folder
-sudo ../stlink/build/st-flash write nuttx.bin 0x8000000
+sudo ../stlink/build/Release/st-flash write nuttx.bin 0x8000000
 ```
 
 ### Running JerryScript


targets/nuttx-stm32f4/README.md
```diff
@@ -16,6 +16,8 @@ git clone https://github.com/jerryscript-project/jerryscript.git
 git clone https://bitbucket.org/nuttx/nuttx.git
 git clone https://bitbucket.org/nuttx/apps.git
 git clone https://github.com/texane/stlink.git
+cd stlink
+git checkout 3d24377c2878fc328fb09b67cdfbda529d209417
```
akosthekiss: I'm not sure I agree with this. Why do we have to fix stlink to a given hash? In the commit message you've mentioned that this saved us from the need for frequent updates to the readme. But what is changing so frequently in stlink that triggers updates here?

If we specify the hash here, we will have a "last known good version", for sure. But we lose any potential benefit from the development that happens on stlink master. And if we wanted to follow up on that then we had to update readme with new "last known good version" hashes over and over again.

polaroi8d: @akosthekiss Okey, I agree with that. What do you want, just update the changed path for the binary, and that's all?

akosthekiss: I indeed _suggest_ to limit the changes to section 5. The file path to the stlink  binary is one thing. The other is the flashing instructions URL, where I'd suggest to link to the repository only, without the `#build-from-sources` (which does not exist).

polaroi8d: Updated following your opinion.

targets/nuttx-stm32f4/README.md
```diff
@@ -85,12 +85,12 @@ make
 
 Connect Mini-USB for power supply and connect Micro-USB for `NSH` console.
 
-To configure `stlink` utility for flashing, follow the instructions [here](https://github.com/texane/stlink#build-from-sources).
+To configure `stlink` utility for flashing, follow the instructions in the official Stlink repository [here](https://github.com/texane/stlink).
```
akosthekiss: Consider this nitpicking, but why not put the link on "Stlink repository", why is "here" needed?

PR #1727, opened 2017-04-11
Support a user context void * pointer in jerry_context_t (#1717)

### LINKED ISSUE #1717: [Discussion] API to associate binding state (pointers) with the JerryScript context

gabrielschulhof:
This modification makes it possible to initialize a context in such a
way that a `void *` pointer is stored inside the context and is made
available via a new `jerry_get_user_context()` API.

The pointer is initialized via a new `jerry_init_with_user_context()`
API, which calls the existing `jerry_init()`, after which it sets the
value of the new `user_context` element in the `jerry_context_t`
structure using the context allocation callback provided as the second
parameter to the new `jerry_init_with_user_context()` API. The location
of the cleanup function responsible for deallocating the pointer created
by the context allocation callback is provided as the third parameter.
This location is stored in the context along with the pointer itself.

When a context is discarded via `jerry_cleanup()`, the user context
cleanup function is called to dispose of the pointer stored within the
context.

The semantics behind the API are such that it is now possible to choose
for each context an agent which manages arbitrary user data keyed to the
given context. The agent must be chosen at context instantiation time
and cannot be changed afterwards, remaining in effect for the lifetime
of the context.

Fixes #1717

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -225,8 +225,10 @@ typedef bool (*jerry_object_property_foreach_t) (const jerry_value_t property_na
 
 **Summary**
 
-Initializes the JerryScript engine, making possible to run JavaScript code and perform operations on
-JavaScript values.
+Initializes the JerryScript engine, making it possible to run JavaScript code and perform operations
+on JavaScript values. See also [jerry_init_with_user_context](#jerry_init_with_user_context) if you
+wish to initialize the JerryScript engine in such a way that its context contains a custom pointer
+which you can later retrieve using [jerry_get_user_context](#jerry_get_user_context).
 
 **Prototype**
 
@@ -259,13 +261,91 @@ jerry_init (jerry_init_flag_t flags)
 **See also**
 
 - [jerry_cleanup](#jerry_cleanup)
+- [jerry_init_with_user_context](#jerry_init_with_user_context)
+
+
+## jerry_init_with_user_context
+
+**Summary**
+
+Calls [jerry_init](#jerry_init) to initialize the JerryScript engine, thereby making it possible
+to run JavaScript code and perform operations on JavaScript values. In addition to the first
+parameter this function accepts two more parameters with which it allows the caller to store a
+`void *` pointer inside the context being initialized with `jerry_init ()`. The function calls the
+callback given in its `init_cb` parameter to allocate the memory for the pointer and it stores the
+function pointer given in the `deinit_cb` parameter along with the pointer so that it may be called
+to free the stored pointer when `jerry_cleanup ()` is later called to dispose of the context.
+
+**Prototype**
+
+```c
+void
+jerry_init_with_user_context (jerry_init_flag_t flags,
+                              jerry_user_context_init_cb init_cb,
+                              jerry_user_context_deinit_cb deinit_cb);
+```
+
+`flags` - combination of various engine configuration flags:
+
+- `JERRY_INIT_EMPTY` - no flags, just initialize in default configuration.
+- `JERRY_INIT_SHOW_OPCODES` - print compiled byte-code.
+- `JERRY_INIT_SHOW_REGEXP_OPCODES` - print compiled regexp byte-code.
+- `JERRY_INIT_MEM_STATS` - dump memory statistics.
+- `JERRY_INIT_MEM_STATS_SEPARATE` - dump memory statistics and reset peak values after parse.
+- `JERRY_INIT_DEBUGGER` - enable all features required by debugging.
+
+`init_cb` - a function pointer that will be called to allocate the custom pointer.
+
+`deinit_cb` - a function pointer that will be called when the custom pointer must be freed.
+
+**Example**
+
+```c
+void *
+init_user_context (void)
+{
+  void *return_value;
+
+  /* allocate and initialize return_value */
+
+  return return_value;
+} /* init_user_context */
+
+void
+free_user_context (void *context)
+{
+
+  /* free the value allocated above */
+
+} /* free_user_context */
+
+{
+  /* init_user_context () will be called before the call below returns */
+  jerry_init_with_user_context (JERRY_INIT_SHOW_OPCODES | JERRY_INIT_SHOW_REGEXP_OPCODES,
+                                init_user_context,
+                                free_user_context);
+
+  /* ... */
+
+  /* free_user_context () will be called before the call below returns */
+  jerry_cleanup ();
+}
+```
+
+**See also**
+
+- [jerry_cleanup](#jerry_cleanup)
+- [jerry_get_user_context](#jerry_get_user_context)
 
 
 ## jerry_cleanup
 
 **Summary**
 
-Finish JavaScript engine execution, freeing memory and JavaScript values.
+Finish JavaScript engine execution, freeing memory and JavaScript values. If the context was
+initialized with `jerry_init_with_user_context ()` and a `deinit_cb` was provided, then it will
+be called to free the memory at the custom pointer which was associated with the context being
+cleaned up.
 
 *Note*: JavaScript values, received from engine, will be inaccessible after the cleanup.
 
@@ -279,6 +359,37 @@ jerry_cleanup (void);
 **See also**
 
 - [jerry_init](#jerry_init)
+- [jerry_init_with_user_context](#jerry_init_with_user_context)
+
+
+## jerry_get_user_context
+
+**Summary**
+
+Retrieve the pointer stored within the current context.
+
+**Prototype**
+
+```c
+void *
+jerry_get_user_context (void);
+```
+
+- return value: the pointer that was assigned during `jerry_init_with_user_context ()`
+
+**Example**
+
+```c
+{
+  /* ... */
+  my_context *custom_data = (my_context *) jerry_get_user_context ();
+  /* ... */
+}
+```
+
+**See also**
+- [jerry_init_with_user_context](#jerry_init_with_user_context)
+- [jerry_cleanup](#jerry_cleanup)
 
 
 ## jerry_register_magic_strings
--- a/jerry-core/jcontext/jcontext.h
+++ b/jerry-core/jcontext/jcontext.h
@@ -61,6 +61,8 @@ typedef struct
   ecma_lit_storage_item_t *number_list_first_p; /**< first item of the literal number list */
   ecma_object_t *ecma_global_lex_env_p; /**< global lexical environment */
   vm_frame_ctx_t *vm_top_context_p; /**< top (current) interpreter context */
+  void *user_context_p; /**< user-provided context-specific pointer */
+  jerry_user_context_deinit_cb user_context_deinit_cb; /**< user-provided deleter for context-specific pointer */
   size_t ecma_gc_objects_number; /**< number of currently allocated objects */
   size_t ecma_gc_new_objects; /**< number of newly allocated objects since last GC session */
   size_t jmem_heap_allocated_size; /**< size of allocated regions */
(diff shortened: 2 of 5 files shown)


jerry-core/jcontext/jcontext.h
```diff
@@ -103,6 +103,8 @@ typedef struct
   uint8_t valgrind_freya_mempool_request; /**< Tells whether a pool manager
                                            *   allocator request is in progress */
 #endif /* JERRY_VALGRIND_FREYA */
+  void *user_context;
+  jerry_user_context_deinit_cb user_context_deinit_cb;
```
zherczeg: These are mandatory pointers, I would put them after `vm_frame_ctx_t *vm_top_context_p;`

martijnthe:
Nice 💯

jerry-core/jerry.c
```diff
@@ -167,6 +167,18 @@ jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
 } /* jerry_init */
 
 /**
+ * Jerry engine initialization with custom user context
+ */
+void
+jerry_init_with_user_context (jerry_init_flag_t flags, jerry_user_context_init_cb init_cb,
+                              jerry_user_context_deinit_cb deinit_cb)
```
zherczeg: Full stop after each sentence. Please add a comment for each function. Coding style is in #1459.

jerry-core/jerry.c
```diff
@@ -185,9 +197,23 @@ jerry_cleanup (void)
...
   jerry_make_api_unavailable ();
+
+  if (JERRY_CONTEXT (user_context_deinit_cb))
+  {
+    JERRY_CONTEXT (user_context_deinit_cb) (JERRY_CONTEXT (user_context));
+  }
 } /* jerry_cleanup */
 
 /**
+ * Retrieve user context
```
zherczeg: Please add return value description.

tests/unit/test-user-context.c
```diff
@@ -0,0 +1,51 @@
...
+
+#include "config.h"
+#include "jerryscript.h"
+#include "test-common.h"
+
+static const char *identifying_string = "identifying string";
+static bool user_context_new_was_called = false;
+static bool user_context_free_was_called = false;
+
+static void *user_context_new (void)
```
zherczeg: Newline after `static void`.

gabrielschulhof:
@zherczeg I have addressed your review comments. Could you please take another look?

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,75 @@ jerry_init (jerry_init_flag_t flags)
...
+function calls the callback given in its `init_cb` parameter to allocate the memory for the pointer
+and it stores the function pointer given in the `deinit_cb` parameter along with the pointer so that
+it may be called to free the stored pointer when the context is discarded.
+
+**Prototype**
+
+```c
+void
+jerry_init_with_user_context (jerry_init_flag_t flags, jerry_user_context_init_cb init_cb,
+                              jerry_user_context_deinit_cb deinit_cb);
```
LaszloLango: missing line break:
```c
jerry_init_with_user_context (jerry_init_flag_t flags,
                              jerry_user_context_init_cb init_cb,
                              jerry_user_context_deinit_cb deinit_cb);
```

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,75 @@ jerry_init (jerry_init_flag_t flags)
...
+
+`init_cb` - a function pointer that will be called to allocate the custom pointer.
+
+`deinit_cb` - a function pointer that will be called when the custom pointer must be freed.
+
+**Example**
+
+```c
+void *
+init_user_context(void) {
```
LaszloLango:
```c
init_user_context (void)
{
```

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,75 @@ jerry_init (jerry_init_flag_t flags)
...
+  void *return_value;
+
+  /* allocate and initialize return_value */
+
+  return return_value;
+} /* init_user_context */
+
+void
+free_user_context(void *context) {
```
LaszloLango:
```c
free_user_context (void *context)
{
```

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,75 @@ jerry_init (jerry_init_flag_t flags)
...
+
+  /* free the value allocated above */
+
+} /* free_user_context */
+{
+
```
LaszloLango: move this empty line before the open brace

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,75 @@ jerry_init (jerry_init_flag_t flags)
...
+  /* init_user_context() will be called before the call below returns */
```
LaszloLango: missing space: `init_user_context ()`

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,75 @@ jerry_init (jerry_init_flag_t flags)
...
+  jerry_init_with_user_context (JERRY_INIT_SHOW_OPCODES | JERRY_INIT_SHOW_REGEXP_OPCODES,
+    init_user_context, free_user_context);
```
LaszloLango: wrong indentation

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,75 @@ jerry_init (jerry_init_flag_t flags)
...
+
+  /* ... */
+
+  /* free_user_context() will be called before the call below returns */
```
LaszloLango: missing space: `free_user_context ()`

docs/02.API-REFERENCE.md
```diff
@@ -281,6 +350,36 @@ jerry_cleanup (void);
...
+Retrieve the pointer stored within the current context.
+
+**Prototype**
+
+```c
+void *
+jerry_get_user_context (void);
+```
+
+- return value: the pointer that was assigned during `jerry_init_with_user_context()`
```
LaszloLango: missing space before `(`

jerry-core/jerry.c
```diff
@@ -167,6 +167,18 @@ jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
 } /* jerry_init */
 
 /**
+ * Initialize Jerry engine with custom user context.
+ */
+void
+jerry_init_with_user_context (jerry_init_flag_t flags, jerry_user_context_init_cb init_cb,
```
LaszloLango: missing line break

jerry-core/jerryscript.h
```diff
@@ -189,11 +199,14 @@ typedef struct
  * General engine functions
  */
 void jerry_init (jerry_init_flag_t flags);
+void jerry_init_with_user_context (jerry_init_flag_t flags, jerry_user_context_init_cb init_cb,
```
LaszloLango: ditto

gabrielschulhof:
@zherczeg @LaszloLango I have now address all review comments so far.

jerry-core/jcontext/jcontext.h
```diff
@@ -61,6 +61,8 @@ typedef struct
   ecma_lit_storage_item_t *number_list_first_p; /**< first item of the literal number list */
   ecma_object_t *ecma_global_lex_env_p; /**< global lexical environment */
   vm_frame_ctx_t *vm_top_context_p; /**< top (current) interpreter context */
+  void *user_context; /**< user-provided context-specific pointer */
```
zherczeg: user_context_p - this is a pointer

gabrielschulhof:
@LaszloLango should all be good now. Please take another look!

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,78 @@ jerry_init (jerry_init_flag_t flags)
...
+
+**Summary**
+
+Initializes the JerryScript engine, making it possible to run JavaScript code and perform operations
+on JavaScript values. In addition to the first parameter, this function accepts two more parameters
+with which it is possible to store a `void *` pointer inside the context being initialized. The
+function calls the callback given in its `init_cb` parameter to allocate the memory for the pointer
+and it stores the function pointer given in the `deinit_cb` parameter along with the pointer so that
+it may be called to free the stored pointer when the context is discarded.
+
```
jiangzidong: Could you please describe the relationship among `jerry_init_with_user_context`, `jerry_init` and `jerry_cleanup`.
e.g.
in the `##jerry_init`: Note: If a user context need to be initialized, please use `jerry_init_with_user_context` instead.
in the `##jerry_init_with_user_context`: Note: `jerry_init` will be called inside `jerry_init_with_user_context`, and `deinit_cb` will be called in the `jerry_cleanup`
in the `##jerry_cleanup`: If the engine is Initialized by `jerry_init_with_user_context` with `deinit_cb`, then the `deinit_cb` will be called during cleanup.

What do you think?

zherczeg: This is a good idea.

docs/02.API-REFERENCE.md
```diff
@@ -261,6 +261,78 @@ jerry_init (jerry_init_flag_t flags)
...
+{
+
+  /* free the value allocated above */
+
+} /* free_user_context */
+
+{
+  /* init_user_context () will be called before the call below returns */
+  jerry_init_with_user_context (JERRY_INIT_SHOW_OPCODES | JERRY_INIT_SHOW_REGEXP_OPCODES,
+                                init_user_context, free_user_context);
```
LaszloLango: missing line break. Every parameter should be on separate line.

jerry-core/jerry.c
```diff
@@ -167,6 +167,19 @@ jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
 } /* jerry_init */
 
 /**
+ * Initialize Jerry engine with custom user context.
+ */
+void
+jerry_init_with_user_context (jerry_init_flag_t flags,
+                              jerry_user_context_init_cb init_cb,
+                              jerry_user_context_deinit_cb deinit_cb)
```
LaszloLango: Missing comments of the arguments.

gabrielschulhof:
@jiangzidong, @LaszloLango how's about now?

zherczeg:
Still LGTM.

PR #1728, opened 2017-04-11
Refactor js2c.py and eliminate related pylint warnings

robertsipka:


tools/js2c.py
```diff
@@ -0,0 +1,126 @@
...
+
+def reduce_code(code):
+    code = re.sub(re.compile(r"/\*.*?\*/", re.DOTALL), "", code)  # remove all occurance streamed comments
+    code = re.sub(re.compile(r"//.*?\n"), "", code)  # remove all occurance singleline comments
+    code = re.sub('\n+', '\n', re.sub('\n +', '\n', code))  # remove white spaces
+    return code
+
+
+def js_to_native_code(path, name, build_type):
+    code = {}
```
galpeter: I don't think we need this line here. The `code` will be overwritten in line 61.

tools/js2c.py
```diff
@@ -0,0 +1,126 @@
...
+
+    with open(path, 'r') as js_source:
+        code = js_source.read() + '\0'
```
galpeter: Do we still need the `\0` here?

tools/js2c.py
```diff
@@ -0,0 +1,126 @@
...
+
+        if build_type != 'debug':
```
galpeter: This can be de-indented with one, no need to have it inside the `with` block.

tools/js2c.py
```diff
@@ -0,0 +1,126 @@
...
+            code = reduce_code(code)
+
+    data = c_source_helper.format_code(code, 1, 2)
+
+    native_code = """const static char {0}_n[] = \"{0}\";
```
galpeter: If we are using the `"""... """`  notation there is no need to escape the  `"` character(s)

tools/js2c.py
```diff
@@ -0,0 +1,126 @@
...
+const static char {0}_s[] =
+{{
+{1}
+}};
+const static int {0}_l = {2};
+""".format(name, data, str(len(code)-1))
```
galpeter: Why do we need to convert he length to string?

tools/js2c.py
```diff
@@ -0,0 +1,126 @@
...
+
+
+def extract_name(path):
+    special_chars = re.compile(r'[-\\?\'".]')
+    return special_chars.sub('_', os.path.splitext(os.path.basename(path))[0])
+
+
+def reduce_code(code):
+    code = re.sub(re.compile(r"/\*.*?\*/", re.DOTALL), "", code)  # remove all occurance streamed comments
+    code = re.sub(re.compile(r"//.*?\n"), "", code)  # remove all occurance singleline comments
```
galpeter: Using the re.compile here doesn't make any sense, just do: `re.sub(<pattern>,..., flags=re.DOTALL)`

tools/unicode_case_conversion.py
```diff
@@ -395,7 +396,7 @@ def extract_character_pair_ranges(letter_case, reverse_letter_case):
             in_range = False
 
     # Remove all founded case mapping from the conversion tables after the scanning method
-    for idx in range(len(start_points)):
+    for idx, _ in enumerate(start_points):
```
galpeter: Was this reported as a pylint error?

robertsipka: Yes, with the following msg: 
"Consider using enumerate instead of iterating with range and len (consider-using-enumerate)"

robertsipka:
@galpeter : Thanks, I've updated this patch based on your review.

tools/unicode_case_conversion.py
galpeter: I see. Then it is also possible to remove the `letter_id = ..` line and use the `letter_id` instead of the `_`.

robertsipka: You're right, I've modified the patch accordingly.

PR #1729, opened 2017-04-12
Implement other routines of Promise

jiangzidong:
add Promise.resolve, Promise.reject, Promise.race, Promise.all and Promise.prototype.catch


jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,621 @@
...
+  ecma_free_value (call_ret);
+
+  /* String '0' here indicates the [[Promise]] of Capability. */
+  ecma_string_t *str_0 = ecma_new_ecma_string_from_uint32 (0);
+  ecma_value_t promise_new = ecma_op_object_get (ecma_get_object_from_value (capability), str_0);
+  ecma_deref_ecma_string (str_0);
+  ecma_free_value (capability);
+
+
+  return promise_new;
```
zherczeg: One newline.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,621 @@
...
+} /* ecma_builtin_promise_reject_or_resolve */
+
+
+/**
```
zherczeg: One newline.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,621 @@
...
+ * @return ecma value of the new promise.
+ *         Returned value must be freed with ecma_free_value.
+ */
+inline static ecma_value_t
+ecma_builtin_promise_do_race (ecma_value_t array, /**< the array for race */
+                              ecma_value_t capability, /**< PromiseCapability record */
+                              ecma_value_t ctor) /**< the caller of Promise.race */
+{
+
+  JERRY_ASSERT (ecma_is_value_object (capability)
```
zherczeg: No newline before assert.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,621 @@
...
+  ecma_object_t *array_p = ecma_get_object_from_value (array);
+  ecma_value_t len_value = ecma_op_object_get (array_p, magic_string_length_p);
+  ecma_deref_ecma_string (magic_string_length_p);
+  ecma_length_t len = (ecma_length_t) ecma_get_integer_from_value (len_value);
+  ecma_fast_free_value (len_value);
+  ecma_string_t *str_0 = ecma_new_ecma_string_from_uint32 (0);
+  ecma_string_t *str_1 = ecma_new_ecma_string_from_uint32 (1);
+  ecma_string_t *str_2 = ecma_new_ecma_string_from_uint32 (2);
+  ecma_string_t *str_3 = ecma_new_ecma_string_from_uint32 (3);
+  ecma_string_t *str_4 = ecma_new_ecma_string_from_uint32 (4);
```
zherczeg: Perhaps we could replace these 0 .. 4 numbers by symbolic constants, something like:

    ecma_new_ecma_string_from_uint32 (ECMA_PROMISE_THIS_IS_A_GREAT_PROPERTY);

Would be much easier to follow these code paths.

jiangzidong: That's a good idea.
But here the numberic property names are reused among different objects.
e.g.
str_0 indicates the [[Promise]] in capability object, at the sametime, it indicated the [[AlreadyCalled]] in all-handler-function.
Should we separate them as two different strings? (which may consume more memory/performance)

jiangzidong: @zherczeg What do you think? Should I use macro constants which may lead to more variable/operation/memory.

zherczeg: In Jerry the number of properties increase the memory consumption not their value. You can also assign the same constant to multiple symbolic constants. Just write down the transitions clearly.

jiangzidong: Updated. You're right, since they are uint32-string, only the uint32 numbers are stored in the property, the total memory will not affected. 
Performance will be affected, because of the more `ecma_new_ecma_string_from_uint32` and `ecma_deref_ecma_string`, but only slightly effect.

jerry-core/ecma/operations/ecma-promise-object.h
```diff
@@ -58,6 +58,23 @@ typedef struct
...
+{
+  ECMA_PROMISE_PROPERTY_PROMISE, /**< [[Promise]] property */
+  ECMA_PROMISE_PROPERTY_RESOLVE, /**< [[Resolve]] property */
+  ECMA_PROMISE_PROPERTY_REJECT, /**< [[Reject]] property */
+  ECMA_PROMISE_PROPERTY_CAPABILITY, /**< [[Capability]] property */
+  ECMA_PROMISE_PROPERTY_HANDLER, /**< [[Handler]] property */
+  ECMA_PROMISE_PROPERTY_ALREADY_CALLED, /**< [[AlreadyCalled]] property */
+  ECMA_PROMISE_PROPERTY_INDEX, /**< [[Index]] property */
+  ECMA_PROMISE_PROPERTY_VALUE, /**< [[Values]] property */
+  ECMA_PROMISE_PROPERTY_REMAINING_ELEMENT /**< [[RemainingElement]] property */
```
zherczeg: You can use = operator to create the same values, so if you worry about the performance please use it.

zherczeg: Thank you for the chenge, the algorithms are much more readable now.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,620 @@
...
+  ecma_string_t *str_promise = ecma_new_ecma_string_from_uint32 (ECMA_PROMISE_PROPERTY_PROMISE);
+  ecma_string_t *str_resolve = ecma_new_ecma_string_from_uint32 (ECMA_PROMISE_PROPERTY_RESOLVE);
+  ecma_string_t *str_reject = ecma_new_ecma_string_from_uint32 (ECMA_PROMISE_PROPERTY_REJECT);
+
+  ecma_value_t resolve = ecma_op_object_get (ecma_get_object_from_value (capability),
+                                             str_resolve);
+  ecma_value_t reject = ecma_op_object_get (ecma_get_object_from_value (capability),
+                                            str_reject);
+
+  while (true)
```
LaszloLango: I think a `for` would be better here

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,620 @@
...
+
+  JERRY_ASSERT (ext_object_p->u.class_prop.class_id == LIT_MAGIC_STRING_NUMBER_UL);
+
+  JERRY_ASSERT (ecma_is_value_integer_number (ext_object_p->u.class_prop.u.value));
+
+  ecma_length_t current = (ecma_length_t) ecma_get_integer_from_value (ext_object_p->u.class_prop.u.value);
+
+  if (is_inc)
+  {
+    current ++;
```
LaszloLango: no space: `current++;`

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,620 @@
...
+  }
+  else
+  {
+    current --;
```
LaszloLango: ditto

jerry-core/ecma/operations/ecma-promise-object.h
zherczeg: What do you think about the idea of assigning the same integer number to multiple enums? That would solve the performance issue and still would be more readable than simple numbers?

jiangzidong: I don't think they can solve the performance issue, because I think the variable name of ecma_string should also be readable.

old one:
```
// use number
 ecma_string_t *str_0 = ecma_new_ecma_string_from_uint32 (0);

// 0 can represent [[Promise]] of capability
operate str_0 to handle with [[Promise]]

// 0 can also represent [[capability]] of executor
operate str_0 to handle with capability

ecma_deref_ecma_string(str_0)
```

new_api
```
 ecma_string_t *str_promise = ecma_new_ecma_string_from_uint32 (ECMA_PROMISE_PROPERTY_PROMISE);

//operate the str_promise

 ecma_string_t *str_capability = ecma_new_ecma_string_from_uint32 (ECMA_PROMISE_PROPERTY_CAPABILITY);

//operate the str_capability

ecma_deref_ecma_string(str_promise);
ecma_deref_ecma_string(str_capability)
```

jerry-core/lit/lit-magic-strings.inc.h
```diff
@@ -53,6 +53,7 @@ LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_LN2_U, "LN2")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_NAN, "NaN")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_UTC_U, "UTC")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_ABS, "abs")
+LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_ALL, "all")
```
akosthekiss: With #1690 , the definition of magic strings changed. Please, rebase the commit onto current master and add these new strings to `jerry-core/lit/lit-magic-strings.ini`. Then, by running `tools/gen-magic-strings.py`, you'll re-generate this `.inc.h` file. (Should you accidentally forget to do that, `tools/run-tests.py` will remind you anyway.)

jiangzidong:
@akosthekiss Thanks for your reminding. Rebased to the lastest master branch.

But why the magic string of Promise's routine (all, race, resolve, reject) (in ecma-builtin-promise.inc.h) didn't be guarded by marcos? I didn't dive into your py script yet.

jiangzidong:
@akosthekiss I see, I didn't guard the ecma-builtin-promise.inc.h with macro. Fixed

akosthekiss:
@jiangzidong ah, that's my fault. Apparently I've missed to add that guard when the magic string generator patch landed. (Note that not adding a guard causes no error, "only" bigger lib size for some configurations.) Thanks for fixing up my sloppyness.

jerry-core/ecma/operations/ecma-promise-object.h
zherczeg: I see. We can simply assign aliases.

ecma_string_t *str_promise = ecma_new_ecma_string_from_uint32 (ECMA_PROMISE_PROPERTY_PROMISE);
COMPILE_ASSERT(ECMA_PROMISE_PROPERTY_PROMISE == ECMA_PROMISE_PROPERTY_CAPABILITY);
str_capability = str_promise;

Or we can use a ref/deref, but I am fine with this code above.

jiangzidong: I am not sure what is "code above"? the code in your last comments or the code in the patch?

Do we have to assign the same integer number to multiple enums?

zherczeg: My code. The problem with this 0,1,2,3 is that it can be changed anytime, and very hard to track all of them in the code unless there are symbolic constants and asserts in each places. They cost nothing at runtime but heavily improves maintenance.

jiangzidong: Yes, and now the patch already used symbolic constants

zherczeg: Ok, I am confused now. You have this sentence:

I don't think they can solve the performance issue, because I think the variable name of ecma_string should also be readable.

Is still there a perf issue or not?

jiangzidong: Sorry @zherczeg , I didn't explain well. 

First of all, it is not a big perf issue, since the "prop name sharing" only appears in some routines.
And to this problem, there always be a tradeoff between perf and readability/maintainability. I dont think `str_capability = str_promise;` is a good idea, people may be confused: why str_capability can be assign with str_promise, why  str_resolve cannot? 

Since it will not hurt the perf much, I think it is best to keep the current code.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,616 @@
...
+/**
+ * Runtime Semantics: PerformPromiseAll.
+ *
+ * See also:
+ *         ES2015 25.4.4.1.1
+ *
+ * @return ecma value of the new promise.
+ *         Returned value must be freed with ecma_free_value.
+ */
+inline static ecma_value_t
```
zherczeg: Lots of large functions are inlined. Is there a reason for that?

jiangzidong: Because those functions ( `ecma_builtin_promise_do_all`, `ecma_builtin_promise_do_race` and `ecma_builtin_promise_reject_abrupt`) are only called in one certain place. I get out those code and form a function to enhance the readability.
So I think "inline" will prevent the function-call-overhead, and at the same time, it will not increase the code size.

zherczeg: As far as I know compilers inline static function which used only once. But I don't mind keeping the inline.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -38,6 +41,616 @@
...
+  uint8_t builtin_id = ecma_get_object_builtin_id (ecma_get_object_from_value (this_arg));
+
+  if (builtin_id != ECMA_BUILTIN_ID_PROMISE)
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("'this' is not the Promise constructor."));
+  }
+
+  if (is_resolve)
+  {
+    if (ecma_is_value_object (argument)
```
LaszloLango: please merge these two if statements

LaszloLango:
Great work! LGTM after minor fix.

jiangzidong:
@akosthekiss do you have further comments?

martijnthe:
🚢

PR #1731, opened 2017-04-13
Eliminate pylint warnings from generate_pins.py

robertsipka:
It also contains small refactoring steps:

* Added a `main` function - instead of just writing all the code into the
  "if name" block - to avoids creating global variables that could affect
  other functions.

* Added a `write_pins_to_files` function - which writes the generated pins
  into the output JS and C++ files - to avoid too many local variables.


targets/mbedos5/tools/generate_pins.py
```diff
@@ -53,41 +69,50 @@ def find_file(root_dir, directories, name):
...
 
 
 class TypeDeclVisitor(c_ast.NodeVisitor):
-    def __init__(self, filter_names=[]):
+    """
+    A TypeDecl visitor class that walks the ast and calls a visitor function for every node found.
+    """
+    def __init__(self, filter_names=None):
+        if filter_names is None:
+            filter_names = []
```
akosthekiss: This can be written in a shorter (arguably, more pythonic) way:
```py
filter_names = filter_names or []
```

targets/mbedos5/tools/generate_pins.py
```diff
@@ -111,29 +136,71 @@ def visit_TypeDecl(self, node):
...
+
 def enumerate_pins(c_source_file, include_dirs, definitions):
     """
     Enumerate pins specified in PinNames.h, by looking for a PinName enum
     typedef somewhere in the file.
     """
-    definitions += ['__attribute(x)__=', '__extension__(x)=', 'register=', '__IO=', 'uint32_t=unsigned int']
+    definitions += ['__attribute(x)__=',
+                    '__extension__(x)=',
+                    'register=', '__IO=',
```
akosthekiss: Shouldn't these two strings be on separate lines, too?

robertsipka:
@akosthekiss : Thanks for the review, I've updated the patch accordingly.

targets/mbedos5/tools/generate_pins.py
```diff
@@ -53,41 +69,50 @@ def find_file(root_dir, directories, name):
...
         yield root
 
 
 class TypeDeclVisitor(c_ast.NodeVisitor):
-    def __init__(self, filter_names=[]):
+    """
+    A TypeDecl visitor class that walks the ast and calls a visitor function for every node found.
+    """
+    def __init__(self, filter_names=None):
+        filter_names = filter_names or []
```
akosthekiss: this two-step assignment is unnecessary. let's just use `self.names = filter_names or []`

robertsipka:
@akosthekiss : Yes, it seemed to me a little tricky. It required to append `mbed-os` to the sys.path list, which determines the interpreter’s search path for `tools.targets` module. It depends on the `sys` import, so I do not know how to avoid this warning. Do you have any suggestion?

akosthekiss:
@robertsipka indeed, that import only works once the sys.path has been tweaked. It's not ideal, but as we cannot get rid of all the pylint warnings anyway, let's keep this as is.

PR #1732, opened 2017-04-13
Buildfix for nuttx-stm32f4 target

robertsipka:
The `jerry_port_jobqueue_enqueue` function is undefined on this target.


zherczeg:
Would be good to make this configurable.

PR #1734, opened 2017-04-13
Always define debugger-related opcodes

bzsolt:
This change makes the generated snapshot independent from buildconfig.


LaszloLango:
IMHO we should update the snapshot version too

bzsolt:
@LaszloLango I've updated the snapshot-version too.

jiangzidong:
@bzsolt Could you please explain the reason of the patch?

After the patch, the snapshot of a js code still depends on the `FEATURE_DEBUGGER`. e.g. only when FEATURE_DEBUGGER is enabled, the bytecode will insert CBC_BREAKPOINT_DISABLED at the end of each statement. am I right?

zherczeg:
The problem was that the snapshot byte code generated when the debugger is enabled (although not used) is different from when the debugger is disabled. The opposite is also true, when you enable debugger you cannot execute a snapshot which was generated by a JerryScript without a debugger.

jiangzidong:
I see, after the patch, all snapshot (no matter generated w/ or w/o debugger) can be executed in all jerry.

```
The opposite is also true, when you enable debugger you cannot execute a snapshot which was generated by a JerryScript without a debugger.
```
Is that a typo?
I think the case in your example can work without the patch. Only when you **disable** debugger, you cannot execute snapshot generated **with a debugger**.

zherczeg:
:) I don't understand your sentence, but probably we are thinking the same thing.

A: jerry with debugger
B: jerry without debugger

Combinations: [generates snapshot, executes snapshot] 

Old: [A,A] [B,B]
New: [A,A] [B,B] [A,B] [B,A]

jiangzidong:
will [B, A] fail before the patch? Let's me have a quick test

jiangzidong:
Oh, I see, the debugger related cbc codes are insert into the middle of the opcode list (not the tail). You're right. :)

PR #1736, opened 2017-04-14
Add Python wrapper for the HTML-based debugger

akosthekiss:
The patch aims at easing the startup of the HTML-based JerryScript
debugger client by adding a python script that starts the default
browser with the HTML page implementing the debugger client.
Additionally, the pyton script encodes the host:port address
information of the debug server in the file url and the HTML client
is extended to retrieve this information. Thus, both the console
and the browser-based debugger client can be started similarly.

### DIFF
--- a/jerry-debugger/jerry-client-ws-html.py
+++ b/jerry-debugger/jerry-client-ws-html.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python
+
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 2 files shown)


zherczeg:
I don't really understand the use case for this.

akosthekiss:
@zherczeg I'm a console-bound type of a coder, most of the time I'm rather typing than clicking. Starting up the console-based debugger is convenient and compact (a single line of `python jerry-client-ws.py address:port`), however, starting up the html variant is full of clicking around (new tab/window, open local file, enter the connect address:port -- I know, I should learn the keyboard shortcuts). This PR simply adds a command line wrapper around the html variant of the debugger, so that I (or anyone else) can just type the a similar command line (`python jerry-client-ws-html.py address:port`) to bring up the browser, load the debugger page, and fill in the appropriate connect command.

If it turns out that this is a use case valid for me only, then we can drop the idea.

LaszloLango:
Do we really need a script to open a browser? :) Does not make much sense if you ask me. But if it helps to you then let it be. LGTM

yichoi:
LGTM. I've found the use case of this. :)

PR #1739, opened 2017-04-15
Ensure that const and pure function attributes are used properly

akosthekiss:
Some functions were incorrectly marked as const but were pure only
(or not even pure). Some functions were marked as pure but
qualified as const. Some functions were not attributed at all but
qualified either as pure or const. Some functions had attributes
at definition but not at declaration. This commit fixes these
inconsistencies.


akosthekiss:
Interestingly, neither binary size nor runtime stats (heap peak, stack peak, max RSS, running time) changed compared to master. However, I still think that getting the attributes right is important, as it has turned out recently that an unfit attribute might cause misguided optimisation and incorrect runtime behaviour (#1695).

PR #1744, opened 2017-04-18
[nuttx-stm32f4] Use jerry-libm instead of the math library in NuttX

bzsolt:


yichoi:
Please rebase it to the master

PR #1748, opened 2017-04-19
Rename the doc of coding standard and add it to the `update-webpage.sh`

bzsolt:


tools/update-webpage.sh
```diff
@@ -29,16 +29,18 @@ INTERNALS_MD="04.INTERNALS.md"
...
 
 titles[$GETTING_STARTED_MD]="Getting Started"
 titles[$API_REFERENCE_MD]="API Reference"
 titles[$API_EXAMPLES_MD]="API Examples"
-titles[$DEBUGGER_MD]="Debugger"
 titles[$INTERNALS_MD]="Internals"
 titles[$PORT_API_MD]="Port API"
 titles[$REFERENCE_COUNTING_MD]="Reference counting"
+titles[$DEBUGGER_MD]="Debugger"
+titles[$CODING_STANDARDS_MD]="Coding standards"
```
akosthekiss: Shouldn't we use "Title Case" everywhere? (Note: if so, "Reference counting" might need to be updated, too.)

bzsolt: Yes, we should. I've fixed it, thank you.

PR #1750, opened 2017-04-19
OSX-related CMakeLists maintenance

akosthekiss:
The linker of OSX does not support `-s` for link-time stripping.
(The option does not cause an error but clutters build output with
warnings.)


akosthekiss:
Found one more OSX-related issue in CMakeLists:
No OSX-dependent LTO flags: LTO is disabled on OSX.

Updated the commit and changed the PR title accordingly.

yichoi:
LGTM

PR #1753, opened 2017-04-20
Support ECMAScript stopping in JerryScript.

zherczeg:


tools/build.py
```diff
@@ -109,6 +109,8 @@ def devhelp(helpstring):
                         help='build unittests')
     parser.add_argument('-v', '--verbose', action='store_const', const='ON', default='OFF',
                         help='increase verbosity')
+    parser.add_argument('--vm-exec-stop', metavar='X', choices=['ON', 'OFF'], default='ON', type=str.upper,
```
robertsipka: I think you also need to disable the `vm execution stopping` by default in here.

docs/02.API-REFERENCE.md
```diff
@@ -3877,3 +3879,75 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+## jerry_set_vm_exec_stop_callback
+
+**Summary**
+
+When JERRY_FEATURE_VM_EXEC_STOP is enabled a callback function can be
+specified by this function. This callback is periodically called when
+JerryScript executes an ECMASCript program.
+
+If the callback returns with undefined value the ECMAScript execution
+continues. Otherwise the result is converted to an error and thrown
```
martijnthe: Doesn't it make more sense to let the callback return an `Error` as opposed to letting the engine convert it into an error?

That way the callback can decide what `Error` (sub)type to use for this.

tools/build.py
zherczeg: Good catch. Sorry.

docs/02.API-REFERENCE.md
zherczeg: This would make the return value processing a bit more difficult since we would need to free the returned value, but would make the code more robust. If this is an acceptable overhead, I can change it.

martijnthe: Not sure if I follow.. If the return value is a Jerry string you don't need to free it while if it's an object (`Error` instance) you do?

zherczeg: It is not error case, but the normal case. If we return with undefined, we don't need to do anything (it is a primitive value). But if we allow to return with anything in the normal case, we need to prepare for complex objects, and free them.

Alternative suggestion. Return with:
- undefined (with or without error flag): do nothing
- anything else: if the value does not have an error flag, automatically add it. Throw the value.

martijnthe: Sounds good to me!

tests/unit/test-api.c
```diff
@@ -1002,6 +1017,28 @@ main (void)
...
+  if (jerry_is_feature_enabled (JERRY_FEATURE_VM_EXEC_STOP))
+  {
+    jerry_init (JERRY_INIT_EMPTY);
+    const char *inf_loop_code_src_p = "while(true) {}";
+    parsed_code_val = jerry_parse ((jerry_char_t *) inf_loop_code_src_p, strlen (inf_loop_code_src_p), false);
+    TEST_ASSERT (!jerry_value_has_error_flag (parsed_code_val));
+
+    int countdown = 6;
+
+    jerry_set_vm_exec_stop_callback (vm_exec_stop_callback, &countdown, 16);
```
jiangzidong: a question to verify:
if the JS code do contains useful loop, e.g.
```
for (i = 0; i<6*16; i++)
{
 a[i] = process(i);
}
```
or multiple loops but the total VM_OC_BACKWARD_BRANCH bytecode exceeds 6*16, the case will also be stopped
here, right?

zherczeg: I don't really understand this question. In your case the loop will probably stop at 6*16-1 index (a function start check and 6*16-1 backward loop checks).

To reduce the overhead the check only happens at backward loops and stack rollbacks (includes function entries).

jiangzidong: Sorry, it's a bad example.
I wanted to ask whether the current design only count the VM_OC_BACKWARD_BRANCH, and then if countdown to 0, the user provided stop_cb will be invoked.  And now I know the answer is YES.

docs/02.API-REFERENCE.md
```diff
@@ -3877,3 +3879,75 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+- `user_p` - user pointer passed to the `stop_cb` function
+- `frequency` - frequency of calling the `stop_cb` function
+
+**Example**
+
+```c
+static jerry_value_t
+vm_exec_stop_callback (void *user_p)
+{
+  static countdown = 10;
```
glistening: Probably, `static int countdown = 10;`

zherczeg: Yes! Thanks for catching it.

tests/unit/test-api.c
zherczeg: To be more precise it counts the backward branches, function entries and stack rollbacks, but in your case there is only a single function entry (the global code is a function). 

This code stops sooner:

    for (i = 0; i<6*16; i++)
    {
      try {
       throw 1
      } catch(e) {
      }
    }

zherczeg: Before you ask: why it counts stack rollbacks? Because we throw an exception which cause a stack rollback and an active catch block catches this exception and restart executing the bzte code. However, our termination check runs again just before the code is executed, and raise another exception. At some point the stack will fully rolled back and the actual function returns with the exception. Then stack roll back start for the callee function.

Hm, function entry checks might not be needed. This could further optimize this code. But I need to think over this.

jiangzidong: Thanks for your explanation :)

zherczeg:
Patch updated, all comments addressed.

LaszloLango:
Please rebase to the current master.

zherczeg:
Rebased.

docs/02.API-REFERENCE.md
```diff
@@ -35,7 +35,8 @@ Possible compile time enabled feature types:
  - JERRY_FEATURE_REGEXP_DUMP - regexp byte-code dumps
  - JERRY_FEATURE_SNAPSHOT_SAVE - saving snapshot files
  - JERRY_FEATURE_SNAPSHOT_EXEC - executing snapshot files
- - JERRY_FEATURE_DEBUGGER - debugging
+ - JERRY_FEATURE_DEBUGGER - debugger server is available
```
akosthekiss: Why this change? It is both unrelated and makes the typedef out of sync from the doc.

jerry-core/jerry.c
```diff
@@ -2149,6 +2152,31 @@ jerry_is_valid_cesu8_string (const jerry_char_t *cesu8_buf_p, /**< CESU-8 string
 } /* jerry_is_valid_cesu8_string */
 
 /**
+ * If JERRY_VM_EXEC_STOP is defined the periodically called callback
```
akosthekiss: incomplete sentence?

jerry-core/jerry.c
```diff
@@ -2149,6 +2152,31 @@ jerry_is_valid_cesu8_string (const jerry_char_t *cesu8_buf_p, /**< CESU-8 string
...
+ */
+void
+jerry_set_vm_exec_stop_callback (jerry_vm_exec_stop_callback_t stop_cb, /**< periodically called user function */
+                                 void *user_p, /**< pointer passed to the function */
+                                 uint32_t frequency) /**< frequency of the function call */
```
akosthekiss: what is the unit of frequency?

jerry-core/jerryscript.h
```diff
@@ -86,6 +86,7 @@ typedef enum
   JERRY_FEATURE_SNAPSHOT_SAVE, /**< saving snapshot files */
   JERRY_FEATURE_SNAPSHOT_EXEC, /**< executing snapshot files */
   JERRY_FEATURE_DEBUGGER, /**< debugging */
+  JERRY_FEATURE_VM_EXEC_STOP, /**< vm execution stopping enabled */
```
akosthekiss: this is also not the same text as the one found in API docs

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,76 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+
+  jerry_set_vm_exec_stop_callback (vm_exec_stop_callback, &countdown, 16);
+
+  // Inifinte loop.
+  const char *src_p = "while(true) {}";
+
+  jerry_value_t src = jerry_parse ((jerry_char_t *) src_p, strlen (src_p), false);
+  jerry_release_value (jerry_run (src));
+  jerry_release_value (src);
+}
```
LaszloLango: missing `jerry_cleanup ();`

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,76 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
 - [jerry_init](#jerry_init)
 - [jerry_cleanup](#jerry_cleanup)
 - [jerry_register_magic_strings](#jerry_register_magic_strings)
+
+
```
LaszloLango: Based on `jerryscript.h`: add a new top level header here: `# Miscellaneous functions.`

jerry-core/vm/vm.c
```diff
@@ -2595,6 +2620,35 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+            if (ecma_is_value_undefined (result))
+            {
+              JERRY_CONTEXT (vm_exec_stop_counter) = JERRY_CONTEXT (vm_exec_stop_frequency);
+            }
+            else
+            {
+              JERRY_CONTEXT (vm_exec_stop_counter) = 1;
+
+              left_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
+              right_value = ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
```
LaszloLango: `left_value` and `right_value` shouldn't be freed here?

zherczeg: It is freed after the `error:` label.

jerry-core/jerry.c
zherczeg: This is described in the documentation. There is no real unit. If the callback could be called `N` times, it is actually called `floor(N/frequency)` times

zherczeg:
Patch updated, please check.

jerry-core/vm/vm.c
zherczeg: Sorry I wanted to write that the free happens after `error:` label above, and when we reach this position, the variables hold garbage. Since we go back to `error:` we would encounter a double free issue.

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,79 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+# Miscellaneous functions
+
+## jerry_set_vm_exec_stop_callback
+
+**Summary**
+
+When JERRY_FEATURE_VM_EXEC_STOP is enabled a callback function can be
+specified by this function. This callback is periodically called when
+JerryScript executes an ECMASCript program.
```
akosthekiss: ECMAScript (note the 'c')

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,79 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+
+If the callback returns with undefined value the ECMAScript execution
+continues. Otherwise the result is thrown by the engine (if the error
+flag is not set for the returned value the engine automatically sets
+it). The callback function might be called again even if it thrown
```
akosthekiss: threw

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,79 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+an error. In this case the function must throw the same error again.
+
+Frequently calling a callback in the main executor loop reduces
+its performance. The `frequency` argument allows decreasing
+this overhead. If its value is `N`, the callback is called only
+every `N` times. The minimum value of `N` is `1` (`0` is converted
```
akosthekiss: I know I've asked that before but it's still not clear to me, at least from the doc, what this exactly means. "every `N` times" of what? Only after every N cycles of the VM, i.e., after the execution of every N byte code? Or after the execution of every N JS statement? Or something else? (I should not have to read the code to understand it, IMHO)

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,79 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+```c
+static jerry_value_t
+vm_exec_stop_callback (void *user_p)
+{
+  static int countdown = 10;
+
+  while (countdown > 0)
+  {
+    countdown--;
+    return jerry_create_undefined();
```
akosthekiss: add space before `(` (our examples should also follow our style guide)

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,79 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+  }
+
+  // The error flag is added automatically.
+  return jerry_create_string((const jerry_char_t *) "Abort script");
```
akosthekiss: add space before `(`

jerry-core/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ set(FEATURE_SNAPSHOT_SAVE    OFF     CACHE BOOL   "Enable saving snapshot files?
 set(FEATURE_SYSTEM_ALLOCATOR OFF     CACHE BOOL   "Enable system allocator?")
 set(FEATURE_VALGRIND         OFF     CACHE BOOL   "Enable Valgrind support?")
 set(FEATURE_VALGRIND_FREYA   OFF     CACHE BOOL   "Enable Valgrind-Freya support?")
+set(FEATURE_VM_EXEC_STOP     OFF     CACHE BOOL   "Enable vm execution stopping?")
```
akosthekiss: (nitpicking) VM (it's an abbreviation)

docs/02.API-REFERENCE.md
```diff
@@ -3896,3 +3897,79 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
...
+its performance. The `frequency` argument allows decreasing
+this overhead. If its value is `N`, the callback is called only
+every `N` times. The minimum value of `N` is `1` (`0` is converted
+to `1` automatically).
+
+**Prototype**
+
+```c
+void
+jerry_set_vm_exec_stop_callback (jerry_vm_exec_stop_callback_t stop_cb,
```
akosthekiss: Shouldn't we document the signature of `jerry_vm_exec_stop_callback_t`, too? Right now, it can only be deduced from the example.

jerry-core/vm/vm.c
```diff
@@ -847,6 +847,31 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
 
         if (opcode_data & VM_OC_BACKWARD_BRANCH)
         {
+#ifdef JERRY_VM_EXEC_STOP
+          if (JERRY_CONTEXT (vm_exec_stop_cb) != NULL)
+          {
+            if (--JERRY_CONTEXT (vm_exec_stop_counter) == 0)
```
akosthekiss: merge the two `if`s?
```c
if (JERRY_CONTEXT (vm_exec_stop_cb) != NULL && --JERRY_CONTEXT (vm_exec_stop_counter) == 0)
```

jerry-core/vm/vm.c
```diff
@@ -2595,6 +2620,35 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
           JERRY_ASSERT (VM_GET_CONTEXT_TYPE (stack_top_p[-1]) == VM_CONTEXT_FINALLY_THROW);
           stack_top_p[-2] = result;
         }
+
+#ifdef JERRY_VM_EXEC_STOP
+        if (JERRY_CONTEXT (vm_exec_stop_cb) != NULL)
+        {
+          if (--JERRY_CONTEXT (vm_exec_stop_counter) == 0)
```
akosthekiss: likewise, merge the two `if`s?

tools/build.py
```diff
@@ -108,6 +108,8 @@ def devhelp(helpstring):
                         help='build unittests')
     parser.add_argument('-v', '--verbose', action='store_const', const='ON', default='OFF',
                         help='increase verbosity')
+    parser.add_argument('--vm-exec-stop', metavar='X', choices=['ON', 'OFF'], default='OFF', type=str.upper,
+                        help='enable vm execution stopping (%(choices)s; default: %(default)s)')
```
akosthekiss: (nitpicking) VM?

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -204,6 +204,16 @@ typedef struct
   void *info_p; /**< free info or callback */
 } ecma_native_pointer_t;
 
+#ifdef JERRY_VM_EXEC_STOP
+
+/**
+ * Callback which returns with a non-undefined value
+ * when the ECMAScript execution should be stopped.
+ */
+typedef ecma_value_t (*jerry_vm_exec_stop_cb_t) (void *user_p);
```
akosthekiss: I've tried to figure out why this typedef duplication is needed (i.e., `jerryscript.h` also has `jerry_vm_exec_stop_callback_t`) but didn't succeed. We have other callback types in `jerryscript.h`, too (e.g., `jerry_external_handler_t`), but they have no duplicates internally. Couldn't we drop this also and use the single public typedef?

docs/02.API-REFERENCE.md
zherczeg: Please recommend me a good sentence then:

This is an extra layer for improving performance. Every time when the callback could be called, we decrease a counter, and only call the callback if it is 0. We also reset it back to frequency (that is why frequnce must be > 0, to avoid underflow). The dividers of CPU clock works this way. Basically if you set to to N then N-1 times the call is skipped, and only the Nth time it is executed.

jerry-core/vm/vm.c
zherczeg: I don't know how `--` works in that case. The code is clearer this way.

jerry-core/ecma/base/ecma-globals.h
zherczeg: No. jerryscript.h is not included inside jerry-core. I don't want to use external_pointer_t, because thet requires lot of type casting. This is easier to understand.

docs/02.API-REFERENCE.md
akosthekiss: I completely understand what N is for. What's undocumented is what "Every time when the callback could be called" means. When could the callback be called? After/before every byte code? After/before every JS statement?

jerry-core/vm/vm.c
akosthekiss: It is completely well-defined and standard-specified, there is no vagueness involved. `&&` in C has short circuit evaluation. Please see ISO/IEC 9899 standard, section 6.5.13, "Logical AND operator" (http://www.open-std.org/JTC1/SC22/WG14/www/docs/n1256.pdf, http://www.open-std.org/JTC1/SC22/WG14/www/standards.html#9899)

jerry-core/ecma/base/ecma-globals.h
akosthekiss: This is clearly not true. `jerry-core/jrt/jrt.h` is including `jerryscript.h` (line 22) and `jerry-core/ecma/base/ecma-globals.h` is including `jrt.h` (line 20).

zherczeg: This is very bad news, I am not aware of that. I don't think removing this dependency should be in this patch though.

jerry-core/vm/vm.c
zherczeg: I had enough problems with `--` before. It is mostly undefined.

zherczeg: I would like the current code.

jerry-core/ecma/base/ecma-globals.h
akosthekiss: This is not bad news, this is how existing callback types are already used.
- jerry_external_handler_t in ecma-function-object.c
- jerry_object_free_callback_t in ecma-gc.c
- jerry_user_context_deinit_cb in jcontext.h

They are all defined in jerryscript.h and used in the engine. There is no "dependency to be removed". It's the callback typedef duplication that's to be removed from ecma-globals.h.

jerry-core/vm/vm.c
akosthekiss: I'm really tempted to start a potentially long-running debate on "It is mostly undefined". (It is certainly not.) But I wont. Keep this as is.

jerry-core/ecma/base/ecma-globals.h
akosthekiss: I'm strongly insisting on dropping this internal typedef.

zherczeg:
Follow-up: #1791

PR #1755, opened 2017-04-20
Allow platforms/compilers to signal that user-specified settings are overridden

akosthekiss:
Currently, the Darwin platform and the TI compiler force some of
the settings to predefined values. This patch ensures that these
overrides are highlighted during build.


yichoi:
LGTM

PR #1756, opened 2017-04-20
Fix/rewrite/darwinize test runner of jerry-debugger

akosthekiss:
First issue was that `diff -u0` was not supported on OSX, thus got
replaced by `-U0`. Then, it turned out that test outcome was
determined based on diff's stdout, which was empty even though the
exit code was non-0 - this got changed, too.

### DIFF
--- a/tools/runners/run-debugger-test.sh
+++ b/tools/runners/run-debugger-test.sh
@@ -24,15 +24,19 @@ echo "$START_DEBUG_SERVER"
 eval "$START_DEBUG_SERVER"
 sleep 1s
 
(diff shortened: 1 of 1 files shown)


LaszloLango:
After the change the script won't print out the diff. I understand that it was not work on OS X, but I'd like to see the diff at least on linux systems. It is easier to track down the problem if we can see the diff in the travis logs and not only the "fail/passed" message..

LaszloLango:
My bad. It prints the diff too. :) LGTM

yichoi:
LGTM

PR #1757, opened 2017-04-20
Make string handling in debugger more pythonic

akosthekiss:
* `to_string` is unpythonic, let's use `__str__` instead.
* `\"%s\"` for strings is also unpythonic, let's use `%r`.


yichoi:
LGTM

PR #1740, opened 2017-04-17
A tool in jerry-ext: arg validator for binding

jiangzidong:
**Updated on April 23th**

the patch is updated by combining review comments:

* the folder name is jerry-ext, and all types/symbol inside are named by prefix jerryx_
* the header is under jerry-ext/include/jerrycript-ext/xxx
* the codes for args transformer are in jerry-ext/args/xxx
* the unit test codes are in tests/unit-ext/test-ext-xxx

The doc is not added yet, since we don't get consensus.

---

It provides some APIs for binding developers, so that
they can validate the type of the js argument and convert/assign them
to the native argument.

Related Issue: #1716

It is the first patch of this util, and some **Notes**  are listed below:

* Usage, the example is in `test/unit-ext/test-ext-args`
```
/**
 * The handler should have following arguments:
 *   this: Ignore.
 *   arg1: Bool.
 *   arg2: Number. It must be strict primitive number.
 *   arg3: String.
 *   arg4: function. It is an optional argument.
 *
 */
static jerry_value_t
test_validator_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
                        const jerry_value_t this_val, /**< this value */
                        const jerry_value_t args_p[], /**< arguments list */
                        const jerry_length_t args_cnt) /**< arguments length */
{

  bool arg1;
  double arg2 = 0.0;
  char arg3[5];
  jerry_value_t arg4 = jerry_create_undefined ();

  jerryx_arg_t mapping[] =
  {
    /* ignore this */
    jerryx_arg_ignore (),
    /* 1st argument should be boolean */
    jerryx_arg_boolean (&arg1, JERRY_ARG_CONVERSION_YES, JERRY_ARG_REQUIRED),
    /* 2nd argument should be strict number */
    jerryx_arg_number (&arg2, JERRY_ARG_CONVERSION_NO, JERRY_ARG_REQUIRED),
    /* 3th argument should be string */
    jerryx_arg_string (arg3, 5, JERRY_ARG_CONVERSION_YES, JERRY_ARG_REQUIRED),
    /* 4th argument should be function, and it is optional */
    jerryx_arg_function (&arg4, JERRY_ARG_OPTIONAL)
  };

  jerry_value_t is_ok = jerryx_validate_and_assign_all (this_val,
                                                        args_p,
                                                        args_cnt,
                                                        mapping,
                                                        5);
```

We use functions like `jerry_arg_string`, `jerry_arg_number` to create the jerry_arg_t item for each type.

[...]

PR #1761, opened 2017-04-21
Split unit tests into separate directories

akosthekiss:
The unit tests should follow the component structure, so this patch
moves all `jerry-core` unit tests under `tests/unit-core` and the
`jerry-libm` unit tests under `tests/unit-libm`.


akosthekiss:
This PR also aims at paving the way for `tests/unit-ext`, see #1740 .

PR #1762, opened 2017-04-21
Rename `jerry-port.h` to `jerryscript-port.h`

akosthekiss:
All public headers should follow the pattern `jerryscript[-*].h`.
The `jerry-api.h` to `jerryscript.h` renaming has already happened
a while ago, now it's time for the port API header to follow.

This patch
* renames the public header file,
* updates all includes to use the new file name (in `jerry-main`,
  in all the targets, and in the docs), and
* keeps `jerry-port.h` as a deprecated forwarding header to leave
  some time for external users to follow up with this change.

As a related change, the header of the default port implementation
is also changed to `jerryscript-port-default.h`.


yichoi:
LGTM

PR #1764, opened 2017-04-21
Fix alignment in case of enabled 32bit compressed pointers

bzsolt:

### DIFF
--- a/jerry-core/parser/js/byte-code.h
+++ b/jerry-core/parser/js/byte-code.h
@@ -637,6 +637,9 @@ typedef struct
   uint16_t ident_end;               /**< end position of the identifier group */
   uint16_t const_literal_end;       /**< end position of the const literal group */
   uint16_t literal_end;             /**< end position of the literal group */
+#ifdef JERRY_CPOINTER_32_BIT
+  uint16_t padding;                 /**< an unused value */
+#endif
 } cbc_uint16_arguments_t;
 
 /**
--- a/jerry-core/parser/js/byte-code.c
+++ b/jerry-core/parser/js/byte-code.c
@@ -15,6 +15,12 @@
 
 #include "js-parser-internal.h"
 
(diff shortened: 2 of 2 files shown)


bzsolt:
The size of `cbc_uint16_arguments_t` is 18 bytes, in case of 32bit compressed pointers this can cause misaligned reads/writes, therefore an extra field is added.

LaszloLango:
I can't see why does it depend on 32bit compressed pointers.

bzsolt:
When 32bit cpointers are enabled, the `jmem_cpointer_t` becomes an `uint32_t` which requires 4 byte alignment.

jerry-core/parser/js/byte-code.h
```diff
@@ -637,6 +637,9 @@ typedef struct
   uint16_t ident_end;               /**< end position of the identifier group */
   uint16_t const_literal_end;       /**< end position of the const literal group */
   uint16_t literal_end;             /**< end position of the literal group */
+#ifdef JERRY_CPOINTER_32_BIT
+  uint16_t padding;                 /**< an unused value */
```
akosthekiss: Does that mean that if `JERRY_CPOINTER_32_BIT` then the size of both `cbc_uint8_arguments_t` and `cbc_uint16_arguments_t` **must** be a multiple of 4 bytes? I think that would deserve a static assert. When looking at `cbc_uint16_arguments_t`, it is not that obvious that the current size is 18, because `ecma_compiled_code_t` (the type of the first `header` field of this struct) is defined "far away", i.e., in a different include in a different subdir. If that type changes for whatever reason, then the current padding will need to be changed, and perhaps even `cbc_uint8_arguments_t` will require some padding.

zherczeg: Good idea.

bzsolt: Yes, it does actually. I've added the asserts.

PR #1767, opened 2017-04-21
Merge `lit-unicode-*.inc.h`-related generator scripts

akosthekiss:
The `conversions` and `ranges` files had separate generators, which
shared some functionality through a 3rd python module, and were
quite similar to each other -- but also had some differences.
Moreover, as they were separate scripts, nothing forced to keep the
outputs in sync (to re-generate them at the same time from the same
inputs).

To fix these maintainability issues, this patch
- merges the two python files and part of the 3rd utility module
  into a single script while keeping the existing functionality,
- names the new file `gen-unicode.py` to align better with other
  generator script in `tools`, and
- adds some extra documentation how to retrieve the input files
  (as they are not part of the repository).

The refactoring affected the utility module as well, so this patch
also
- renames `c_source_helper.py` to `gen_c_source.py` (again, for
  naming consistency),
- reorganizes some of its functions not to export module-local
  helper functions, and
- adapts `js2c.py`, as it also uses this utility module.


zherczeg:
@robertsipka please check this change

robertsipka:
I strongly endorse this merge. I also think it is better to keep the outputs in sync. Previously, the version of the inputs were different, but after adapting the generator script of unicode ranges (to python code from bash script), and updating it with the latest version of Unicode Standard, it's not needed to re-generate them from different inputs and scripts anymore.
LGTM (informally)

PR #1772, opened 2017-04-22
Update paths in check tools

akosthekiss:
In #1761, not all unit test paths have been updated in check tools.
This triggers no errors as `check-vera` emits a warning only while
`check-cppcheck` signals nothing. Still, this means that some of
the code in the repository that has been quality-assured, is now
avoiding checks. This patch fixes this.

Moreover, `check-cppcheck` has had incorrect paths for a while: it
looked for source files in the root of the repository, but those
files have been in `jerry-main` for almost a year now. This patch
fixes this, too.

Fallout of the above: `main-unix-minimal.c` had to be improved to
make `check-cppcheck` pass.


yichoi:
LGTM

Issue #1746, opened 2017-04-19
[Discussion] Improved multi-instance support (incl. runtime configurable heap size)


martijnthe:
For our project, I would like to be able to instantiate (multiple) JerryScript engines, each with their own heap size.

Currently, the heap size is configured at compile-time using `CONFIG_MEM_HEAP_AREA_SIZE`, which controls these defines and structs:

```
#define JMEM_HEAP_SIZE ((size_t) (CONFIG_MEM_HEAP_AREA_SIZE))
...
#define JMEM_HEAP_AREA_SIZE (JMEM_HEAP_SIZE - JMEM_ALIGNMENT)
...
typedef struct
{
  jmem_heap_free_t first; /**< first node in free region list */
  uint8_t area[JMEM_HEAP_AREA_SIZE]; /**< heap area */
} jmem_heap_t;
...
extern jmem_heap_t jerry_global_heap;
...
// All accesses of `jerry_global_heap` should be done through this macro:
#define JERRY_HEAP_CONTEXT(field) (jerry_global_heap.field)
```

Simplest way to solve this would be to add `#ifndef JERRY_HEAP_CONTEXT` around the `#define JERRY_HEAP_CONTEXT ...` and make it possible to let the client of JerryScript define `JERRY_HEAP_CONTEXT` without having to touch `jcontext.h`. Should the `jerry_global_heap` variable declaration in `jcontext.c` also be #ifdef'd out in this case?

BTW, `JERRY_CONTEXT`, `JERRY_HASH_TABLE_CONTEXT`, etc. should probably also have `#ifndef`s around them?

Thoughts?

martijnthe:
cc @jiangzidong

zherczeg:
Hm, this is an interesting request. I think the challenge here is that the Jerry allocator does not prepared for variable sized heaps. It expects that the heap size is JMEM_HEAP_AREA_SIZE.

Yes, the definitions of the accessor macros could be improved as well. However they must be defines when jcontext.h is included. So we can add existence checkers around them, but that probably solves only half of the problem. The other side who defines them is still an interesting question. Perhaps jcontext.h should be able to include a custom header which defines these macros.

dbatyai:
[...]

PR #1778, opened 2017-04-24
Support external context, heap and lcache

jiangzidong:
JerryScript should support external context, heap and lcache,
so that it can have multiple instances and runtime configurable heap
size.

Related issue: #1746

* example codes are in `unix-main.c` (a temporary place, later I will put the example code in some unit test case), we should first build jerry as `./tool/build.py --jerry-libc=off --external-context=on`

* people should prepare the jerry_instance_t by themselves, and also provide the port function `jerry_port_get_current_instance` let the engine find the current port. In the multiple-thread cases, the function should deal with thread local storage APIs, but the `default-port` just provide a simplest implementation .

* I will add the doc after we have consensus about the framwork.

### DIFF
--- a/docs/05.PORT-API.md
+++ b/docs/05.PORT-API.md
@@ -94,6 +94,24 @@ bool jerry_port_get_time_zone (jerry_time_zone_t *);
 double jerry_port_get_current_time (void);
 ```
 
+## External instance
+
+Allow user to provide external buffer for jerry instance (which includes an isolated context and heap with other instances), so that user can config the heap size in runtime and run multiple JS apps simultaneously.
+
+```c
+/**
+ * Get the current instance, which contains the current context, heap and other infomation.
+ * Each port should provide its own implementation of this interface.
+ *
+ *Note:
+ *    This port function will be called automatically by jerry-core
+ *    wnen JERRY_ENABLE_EXTERNAL_CONTEXT is defined. If not, this function will never be called.
+ *
+ * @return the pointer to the jerry instance.
+ */
+struct jerry_instance_t *jerry_port_get_current_instance (void);
+```
+
 # How to port JerryScript
 
 This section describes a basic port implementation which was created for Unix based systems.
@@ -180,4 +198,36 @@ double jerry_port_get_current_time (void)
 
   return ((double) tv.tv_sec) * 1000.0 + ((double) tv.tv_usec) / 1000.0;
 } /* jerry_port_get_current_time */
+```
+## External instance
+
+```c
+#include "jerryscript-port.h"
+#include "jerryscript-port-default.h"
+
+/**
+ * Pointer to the current instance.
+ * Note that it is a global variable, and is not a thread safe implementation.
+ */
+static jerry_instance_t *current_instance_p = NULL;
+
+/**
+ * Set the current_instance_p as the passed pointer.
+ */
+void
+jerry_port_default_set_instance (jerry_instance_t *instance_p) /**< points to the created instance */
+{
+  current_instance_p = instance_p;
+} /* jerry_port_default_set_instance */
+
+/**
+ * Get the current instance.
+ *
+ * @return the pointer to the current instance
+ */
+jerry_instance_t *
+jerry_port_get_current_instance (void)
+{
+  return current_instance_p;
+} /* jerry_port_get_current_instance */
 ```
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -2248,6 +2248,64 @@ jerry_is_valid_cesu8_string (const jerry_char_t *cesu8_buf_p, /**< CESU-8 string
                                     (lit_utf8_size_t) buf_size);
 } /* jerry_is_valid_cesu8_string */
 
+/*
+ * Create a jerry instance for external context.
+ *
+ * @return the pointer to the instance.
+ */
+jerry_instance_t *
+jerry_create_instance (uint32_t heap_size, /**< the size of heap */
+                       jerry_instance_alloc_t alloc, /**< the alloc function */
+                       void *cb_data_p) /**< the cb_data for alloc function */
+{
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+
+  uint32_t lcache_size = 0;
+
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+  lcache_size = sizeof (jerry_hash_table_t);
+#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+
+  uint32_t alloc_size = (uint32_t) (sizeof (jerry_instance_t) + lcache_size);
+
+#ifndef JERRY_SYSTEM_ALLOCATOR
+  alloc_size = alloc_size + heap_size + JMEM_ALIGNMENT - 1;
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+
+  jerry_instance_t *instance_p = (jerry_instance_t *) alloc (alloc_size, cb_data_p);
+  memset (instance_p, 0, alloc_size);
+
+  if (instance_p == NULL)
+  {
+    return NULL;
+  }
+
+  instance_p->heap_size = heap_size;
+  instance_p->lcache_p = instance_p->buffer;
+
+#ifndef JERRY_SYSTEM_ALLOCATOR
+  uint8_t *unaligned_heap_p = instance_p->lcache_p + lcache_size;
+  uintptr_t alignment_mask = ~(uintptr_t) (JMEM_ALIGNMENT - 1);
+  instance_p->heap_p = (uint8_t *) (((uintptr_t) unaligned_heap_p + JMEM_ALIGNMENT - 1) & alignment_mask);
+
+  JERRY_ASSERT (((uintptr_t) instance_p->heap_p & (uintptr_t) (JMEM_ALIGNMENT - 1)) == 0);
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+
+  return instance_p;
+
+#else /* !JERRY_ENABLE_EXTERNAL_CONTEXT */
+
+  JERRY_UNUSED (heap_size);
+  JERRY_UNUSED (alloc);
+  JERRY_UNUSED (cb_data_p);
+
+  return NULL;
+
+#endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
+} /* jerry_create_instance */
+
+
+
 /**
  * If JERRY_VM_EXEC_STOP is defined the callback passed to this function is
  * periodically called with the user_p argument. If frequency is greater
--- a/jerry-core/ecma/base/ecma-lcache.c
+++ b/jerry-core/ecma/base/ecma-lcache.c
@@ -98,7 +98,7 @@ ecma_lcache_insert (ecma_object_t *object_p, /**< object */
   size_t row_index = ecma_lcache_row_index (object_cp, name_hash);
   ecma_lcache_hash_entry_t *entries_p = JERRY_HASH_TABLE_CONTEXT (table)[row_index];
 
-  int32_t entry_index;
+  uint32_t entry_index;
   for (entry_index = 0; entry_index < ECMA_LCACHE_HASH_ROW_LENGTH; entry_index++)
   {
     if (entries_p[entry_index].object_cp == ECMA_NULL_POINTER)
--- a/jerry-core/include/jerryscript-core.h
+++ b/jerry-core/include/jerryscript-core.h
@@ -206,12 +206,22 @@ typedef void *(*jerry_user_context_init_t) (void);
 typedef void (*jerry_user_context_deinit_t) (void *user_context_p);
 
 /**
+ * Function type for allocating buffer for JerryScript instance.
+ */
+typedef void *(*jerry_instance_alloc_t) (size_t size, void *cb_data_p);
+
+/**
  * Type information of a native pointer.
  */
 typedef struct
 {
   jerry_object_native_free_callback_t free_cb; /**< the free callback of the native pointer */
 } jerry_object_native_info_t;
+
+/**
+ * A forward declaration of the JerryScript instance structure.
+ */
+typedef struct jerry_instance_t jerry_instance_t;
 
 /**
  * General engine functions.
@@ -402,6 +412,11 @@ jerry_value_t jerry_resolve_or_reject_promise (jerry_value_t promise, jerry_valu
  */
 bool jerry_is_valid_utf8_string (const jerry_char_t *utf8_buf_p, jerry_size_t buf_size);
 bool jerry_is_valid_cesu8_string (const jerry_char_t *cesu8_buf_p, jerry_size_t buf_size);
+
+/*
+ * External context functions.
+ */
+jerry_instance_t *jerry_create_instance (uint32_t heap_size, jerry_instance_alloc_t alloc, void *cb_data_p);
 
 /**
  * Miscellaneous functions.
--- a/jerry-core/jcontext/jcontext.c
+++ b/jerry-core/jcontext/jcontext.c
@@ -19,6 +19,7 @@
  * @{
  */
 
+#ifndef JERRY_ENABLE_EXTERNAL_CONTEXT
 /**
  * Global context.
  */
@@ -48,6 +49,7 @@ jmem_heap_t jerry_global_heap __attribute__ ((aligned (JMEM_ALIGNMENT))) JERRY_G
 jerry_hash_table_t jerry_global_hash_table;
 
 #endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+#endif /* !JERRY_ENABLE_EXTERNAL_CONTEXT */
 
 /**
  * @}
--- a/jerry-core/jcontext/jcontext.h
+++ b/jerry-core/jcontext/jcontext.h
@@ -21,6 +21,7 @@
 
 #include "ecma-builtins.h"
 #include "ecma-jobqueue.h"
+#include "jerryscript-port.h"
 #include "jerry-debugger.h"
 #include "jmem.h"
 #include "re-bytecode.h"
@@ -121,6 +122,84 @@ typedef struct
 } jerry_context_t;
 
 /**
+ * Description of jerry instance, which contains the meta data and buffer.
+ */
+struct jerry_instance_t
+{
+  uint32_t heap_size; /**< size of the heap */
+#ifndef JERRY_SYSTEM_ALLOCATOR
+  uint8_t *heap_p; /**< point to the entrance of the heap in buffer */
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+  uint8_t *lcache_p; /**< point to the entrance of the lcache in buffer */
+#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+  jerry_context_t context; /**< the context of the instance */
+  uint8_t buffer[]; /**< the flexible array for storing context, heap and lcache */
+};
+
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+/**
+ * Hash table for caching the last access of properties.
+ */
+typedef struct
+{
+  ecma_lcache_hash_entry_t table[ECMA_LCACHE_HASH_ROWS_COUNT][ECMA_LCACHE_HASH_ROW_LENGTH];
+} jerry_hash_table_t;
+#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+
+/**
+ * This part is for Jerry which enable external context.
+ */
+
+typedef struct
+{
+  jmem_heap_free_t first; /**< first node in free region list */
+  uint8_t area[]; /**< heap area */
+} jmem_heap_t;
+
+#define JERRY_CONTEXT(field) (jerry_port_get_current_instance ()->context.field)
+
+#ifndef JERRY_SYSTEM_ALLOCATOR
+
+static inline jmem_heap_t * __attr_always_inline___
+jerry_context_get_current_heap (void)
+{
+  return (jmem_heap_t *) (jerry_port_get_current_instance ()->heap_p);
+} /* jerry_context_get_current_heap */
+
+#define JERRY_HEAP_CONTEXT(field) (jerry_context_get_current_heap ()->field)
+
+#ifdef JMEM_HEAP_SIZE
+#error "JMEM_HEAP_SIZE must not be defined if JERRY_ENABLE_EXTERNAL_CONTEXT is defined"
+#endif /* JMEM_HEAP_SIZE */
+
+#define JMEM_HEAP_SIZE (jerry_port_get_current_instance ()->heap_size)
+
+#define JMEM_HEAP_AREA_SIZE (jerry_port_get_current_instance ()->heap_size - JMEM_ALIGNMENT)
+
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+
+static inline jerry_hash_table_t * __attr_always_inline___
+jerry_context_get_current_lcache (void)
+{
+  return (jerry_hash_table_t *) (jerry_port_get_current_instance ()->lcache_p);
+} /* jerry_context_get_current_lcache */
+
+#define JERRY_HASH_TABLE_CONTEXT(field) (jerry_context_get_current_lcache ()->field)
+
+#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+
+#else /* !JERRY_ENABLE_EXTERNAL_CONTEXT */
+
+/**
+ * This part is for Jerry which use default context.
+ */
+
+/**
  * Calculate heap area size, leaving space for a pointer to the free list
  */
 #define JMEM_HEAP_AREA_SIZE (JMEM_HEAP_SIZE - JMEM_ALIGNMENT)
@@ -142,21 +221,6 @@ typedef struct
   jmem_heap_free_t first; /**< first node in free region list */
   uint8_t area[JMEM_HEAP_AREA_SIZE]; /**< heap area */
 } jmem_heap_t;
-
-#ifndef CONFIG_ECMA_LCACHE_DISABLE
-
-/**
- * JerryScript global hash table for caching the last access of properties.
- */
-typedef struct
-{
-  /**
-   * Hash table
-   */
-  ecma_lcache_hash_entry_t table[ECMA_LCACHE_HASH_ROWS_COUNT][ECMA_LCACHE_HASH_ROW_LENGTH];
-} jerry_hash_table_t;
-
-#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
 
 /**
  * Global context.
@@ -189,6 +253,7 @@ extern jerry_hash_table_t jerry_global_hash_table;
  * Provides a reference to the area field of the heap.
  */
 #define JERRY_HEAP_CONTEXT(field) (jerry_global_heap.field)
+
 #endif /* !JERRY_SYSTEM_ALLOCATOR */
 
 #ifndef CONFIG_ECMA_LCACHE_DISABLE
@@ -199,6 +264,8 @@ extern jerry_hash_table_t jerry_global_hash_table;
 #define JERRY_HASH_TABLE_CONTEXT(field) (jerry_global_hash_table.field)
 
 #endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+
+#endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
 
 /**
  * @}
--- a/jerry-core/jmem/jmem-heap.c
+++ b/jerry-core/jmem/jmem-heap.c
@@ -103,11 +103,13 @@ jmem_heap_get_region_end (jmem_heap_free_t *curr_p) /**< current region */
 } /* jmem_heap_get_region_end */
 #endif /* !JERRY_SYSTEM_ALLOCATOR */
 
+#ifndef JERRY_ENABLE_EXTERNAL_CONTEXT
 /**
  * Check size of heap is corresponding to configuration
  */
 JERRY_STATIC_ASSERT (sizeof (jmem_heap_t) <= JMEM_HEAP_SIZE,
                      size_of_mem_heap_must_be_less_than_or_equal_to_MEM_HEAP_SIZE);
+#endif /* !JERRY_ENABLE_EXTERNAL_CONTEXT */
 
 #ifdef JMEM_STATS
 
@@ -148,8 +150,8 @@ void
 jmem_heap_init (void)
 {
 #ifndef JERRY_CPOINTER_32_BIT
-  JERRY_STATIC_ASSERT (((UINT16_MAX + 1) << JMEM_ALIGNMENT_LOG) >= JMEM_HEAP_SIZE,
-                       maximum_heap_size_for_16_bit_compressed_pointers_is_512K);
+  /* the maximum heap size for 16bit compressed pointers should be 512K */
+  JERRY_ASSERT (((UINT16_MAX + 1) << JMEM_ALIGNMENT_LOG) >= JMEM_HEAP_SIZE);
 #endif /* !JERRY_CPOINTER_32_BIT */
 
 #ifndef JERRY_SYSTEM_ALLOCATOR
@@ -181,7 +183,7 @@ jmem_heap_finalize (void)
 {
   JERRY_ASSERT (JERRY_CONTEXT (jmem_heap_allocated_size) == 0);
 #ifndef JERRY_SYSTEM_ALLOCATOR
-  VALGRIND_NOACCESS_SPACE (&JERRY_HEAP_CONTEXT (first), sizeof (jmem_heap_t));
+  VALGRIND_NOACCESS_SPACE (&JERRY_HEAP_CONTEXT (first), JMEM_HEAP_SIZE);
 #endif /* !JERRY_SYSTEM_ALLOCATOR */
 } /* jmem_heap_finalize */
 
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -390,6 +390,20 @@ check_usage (bool condition, /**< the condition that must hold */
   }
 } /* check_usage */
 
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+
+/**
+ * The alloc function passed to jerry_create_instance
+ */
+static void *
+instance_alloc (size_t size,
+                void *cb_data_p __attribute__((unused)))
+{
+  return malloc (size);
+} /* instance_alloc */
+
+#endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
+
 int
 main (int argc,
       char **argv)
@@ -559,6 +573,13 @@ main (int argc,
   {
     is_repl_mode = true;
   }
+
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+
+  jerry_instance_t *instance_p = jerry_create_instance (512*1024, instance_alloc, NULL);
+  jerry_port_default_set_instance (instance_p);
+
+#endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
 
   jerry_init (flags);
 
@@ -764,6 +785,8 @@ main (int argc,
   }
   jerry_release_value (ret_value);
   jerry_cleanup ();
-
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+  free (instance_p);
+#endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
   return ret_code;
 } /* main */
--- a/jerry-port/default/default-external-context.c
+++ b/jerry-port/default/default-external-context.c
@@ -1,0 +1,43 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript-port.h"
+#include "jerryscript-port-default.h"
+
+/**
+ * Pointer to the current instance.
+ * Note that it is a global variable, and is not a thread safe implementation.
+ */
+static jerry_instance_t *current_instance_p = NULL;
+
+/**
+ * Set the current_instance_p as the passed pointer.
+ */
+void
+jerry_port_default_set_instance (jerry_instance_t *instance_p) /**< points to the created instance */
+{
+  current_instance_p = instance_p;
+} /* jerry_port_default_set_instance */
+
+/**
+ * Get the current instance.
+ *
+ * @return the pointer to the current instance
+ */
+jerry_instance_t *
+jerry_port_get_current_instance (void)
+{
+  return current_instance_p;
+} /* jerry_port_get_current_instance */
--- a/jerry-port/default/include/jerryscript-port-default.h
+++ b/jerry-port/default/include/jerryscript-port-default.h
@@ -37,6 +37,8 @@ bool jerry_port_default_is_abort_on_fail (void);
 jerry_log_level_t jerry_port_default_get_log_level (void);
 void jerry_port_default_set_log_level (jerry_log_level_t level);
 
+void jerry_port_default_set_instance (jerry_instance_t *instance_p);
+
 /**
  * @}
  */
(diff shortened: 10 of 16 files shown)


jerry-main/main-unix.c
```diff
@@ -50,6 +50,25 @@
 
 static uint8_t buffer[ JERRY_BUFFER_SIZE ];
 
+#ifdef JERRY_PORT_ENABLE_EXTERNAL_CONTEXT
```
zherczeg: What is the point of having an external context in main-unix?

jiangzidong: I think main-unix is more like a demo for the default target, right? I put here to show the ability/ feasibility of the feature.
The writer of main-unix can decide the size of the jerry-heap during runtime, instead of during the jerry-core compiling time.

I also realize it may be not a proper place. I should put it inside a unit test case.

jerry-core/jcontext/jcontext.h
```diff
@@ -187,6 +190,36 @@ extern jerry_hash_table_t jerry_global_hash_table;
...
+#define JMEM_HEAP_AREA_SIZE (jerry_port_get_current_instance ()->heap_size - JMEM_ALIGNMENT)
+
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+
+#define JERRY_HASH_TABLE_CONTEXT(field) (((jerry_hash_table_t *) jerry_port_get_current_instance ()->heap_p)->field)
+
+#undef ECMA_LCACHE_HASH_ROWS_COUNT
+#define ECMA_LCACHE_HASH_ROWS_COUNT (jerry_port_get_current_instance ()->lcache_row_count)
```
jiangzidong: My concern 1:
in this way, the JMEM_HEAP_SIZE, JMEM_HEAP_AREA_SIZE, ECMA_LCACHE_HASH_ROWS_COUNT and ECMA_LCACHE_HASH_ROW_LENGTH  are not always constant values.

jerry-core/jmem/jmem-heap.c
```diff
@@ -148,8 +150,8 @@ void
 jmem_heap_init (void)
 {
 #ifndef JERRY_CPOINTER_32_BIT
-  JERRY_STATIC_ASSERT (((UINT16_MAX + 1) << JMEM_ALIGNMENT_LOG) >= JMEM_HEAP_SIZE,
-                       maximum_heap_size_for_16_bit_compressed_pointers_is_512K);
+  /* maximum_heap_size_for_16_bit_compressed_pointers_is_512K */
+  JERRY_ASSERT (((UINT16_MAX + 1) << JMEM_ALIGNMENT_LOG) >= JMEM_HEAP_SIZE);
```
jiangzidong: because JMEM_HEAP_SIZE maybe not constant, so I use JERRY_ASSERT instead of JERRY_STATIC_ASSERT

jerry-core/jmem/jmem-heap.c
```diff
@@ -181,7 +183,7 @@ jmem_heap_finalize (void)
 {
   JERRY_ASSERT (JERRY_CONTEXT (jmem_heap_allocated_size) == 0);
 #ifndef JERRY_SYSTEM_ALLOCATOR
-  VALGRIND_NOACCESS_SPACE (&JERRY_HEAP_CONTEXT (first), sizeof (jmem_heap_t));
+  VALGRIND_NOACCESS_SPACE (&JERRY_HEAP_CONTEXT (first), JMEM_HEAP_SIZE);
```
jiangzidong: My concern 2:
Can I use `JMEM_HEAP_SIZE` to replace the `sizeof (jmem_heap_t)`, are they always same?

jerry-core/jmem/jmem-heap.c
```diff
@@ -103,11 +103,13 @@ jmem_heap_get_region_end (jmem_heap_free_t *curr_p) /**< current region */
 } /* jmem_heap_get_region_end */
 #endif /* !JERRY_SYSTEM_ALLOCATOR */
 
+#ifndef JERRY_ENABLE_EXTERNAL_CONTEXT
 /**
  * Check size of heap is corresponding to configuration
  */
 JERRY_STATIC_ASSERT (sizeof (jmem_heap_t) <= JMEM_HEAP_SIZE,
                      size_of_mem_heap_must_be_less_than_or_equal_to_MEM_HEAP_SIZE);
+#endif /* !JERRY_ENABLE_EXTERNAL_CONTEXT */
```
jiangzidong: related to  concern 2: When will sizeof (jmem_heap_t)  **less than** JMEM_HEAP_SIZE

jerry-main/main-unix.c
zherczeg: Yes, main unix is a demo not a unit test. I don't think anybody wants to compile main-unix with different options, this is not its purpose.

jiangzidong: Can I update it after receiving enough comments on the external context api/implementation?
I guess there should be lots of review comments on that.

jerry-core/jcontext/jcontext.h
```diff
@@ -187,6 +190,36 @@ extern jerry_hash_table_t jerry_global_hash_table;
 
 #endif /* !CONFIG_ECMA_LCACHE_DISABLE */
 
+#else /* JERRY_ENABLE_EXTERNAL_CONTEXT */
+
+#define JERRY_CONTEXT(field) (((jerry_context_t *) jerry_port_get_current_instance()->context_p)->field)
```
martijnthe: Hmm, this macro is used a ton, so I think we should optimize this for code size.
I would add a separate `jerry_port_get_current_context()`, to get rid of the extra dereferencing / level of indirection.

@zherczeg what do you think?

jerry-core/jcontext/jcontext.h
```diff
@@ -187,6 +190,36 @@ extern jerry_hash_table_t jerry_global_hash_table;
...
+#ifndef JERRY_SYSTEM_ALLOCATOR
+
+#define JERRY_HEAP_CONTEXT(field) (((jmem_heap_t *) jerry_port_get_current_instance ()->lcache_p)->field)
+
+#undef JMEM_HEAP_SIZE
+#define JMEM_HEAP_SIZE (jerry_port_get_current_instance ()->heap_size)
+
+#undef JMEM_HEAP_AREA_SIZE
+#define JMEM_HEAP_AREA_SIZE (jerry_port_get_current_instance ()->heap_size - JMEM_ALIGNMENT)
+
```
martijnthe: The other macros are used much less compared to `JERRY_CONTEXT`, but if you agree it makes sense to have a separate `jerry_port_get_current_context()`, maybe it makes sense to create separate getters for heap, hash table and lcache, just to make the API more "regular".

What do you think?

jerry-core/jcontext/jcontext.h
zherczeg: I don't understand. Please show me an example.

jerry-core/jcontext/jcontext.h
zherczeg: Not sure I follow, but we must not sacrifice the advantage of absolute addresses on certain (these are the lowest-end) machines.

jerry-core/jmem/jmem-heap.c
zherczeg: Likely.

jerry-core/jmem/jmem-heap.c
zherczeg: Ok for me.

jerry-core/jcontext/jcontext.h
zherczeg: Any suggestion?

jerry-core/jcontext/jcontext.h
martijnthe: I'm suggesting to changing this to:

```
#define JERRY_CONTEXT(field) (((jerry_context_t *)jerry_port_get_current_context())->field)
```

For every place `JERRY_CONTEXT` is used (~280 times, see below), this alternative would probably result in smaller binary code size, because there's no need for the `->context_p` dereferencing (so no load instruction or whatever it compiles to).

```
$ find ./jerry-core -iname *.c | xargs cat | grep -e JERRY_CONTEXT | wc -l
     282
```

zherczeg: Yes, less indirection would help, that is why we have several macros rather than a single one (which would be enough theoretically, but in practice it is not efficient).

As for me I would prefer some kind of strong inlining here, that was the purpose of the macro originally. When you write:

    JERRY_CONTEXT (a) = 5;
    JERRY_CONTEXT (b) = 5;
    JERRY_CONTEXT (c) = 5;
    JERRY_CONTEXT (d) = 5;

The compiler should be able to recognize that the base is the same. Somehow tell the compiler that the base is non-volatile.

jerry-core/jerryscript-port.h
```diff
@@ -150,6 +150,25 @@ void jerry_port_jobqueue_enqueue (jerry_job_handler_t handler, void *job_p);
...
+ */
+typedef struct
+{
+  void *context_p; /**< points to the external context */
+  void *heap_p; /**< points to the external heap */
+  void *lcache_p; /**< points to the external lcache */
+  uint32_t heap_size; /**< the total size of heap */
+  uint32_t lcache_row_count; /**< the row count of lcache hash table */
+  uint32_t lcache_row_length; /**< the row length of lcache hash table */
+} jerry_instance_t;
```
martijnthe: I wonder if it's desirable to "expose" that internally there is a separate struct for the context, the heap and the lcache.

What if in the future we'd want to consolidate these? You'd have to change this API?
What if in the future we'd want to split `jerry_context_t`? You'd have to change this API?

Would it make sense to hide this from the API, by having one `void *` and let the project allocate one big slab of memory instead of 3 separate ones?

zherczeg: Since this is a critical part of the engine, yes, we want to change the API if something changes. And yes, port maintainers must follow those changes. These changes will be rare though.

zherczeg: (The point is not giving a lot of extra work for port maintainers, but if something has a good reason to change, we need to change it. Similar to the 'print' story.)

jerry-core/jcontext/jcontext.h
martijnthe: > Yes, less indirection would help, that is why we have several macros rather than a single one (which would be enough theoretically, but in practice it is not efficient).

Right right. Sorry if I was unclear.
I'm basically saying we should have have separate `jerry_port_...` getters, one for each of these macros.

jerry-core/jcontext/jcontext.h
martijnthe: Note that if not using `JERRY_SYSTEM_ALLOCATOR`, the existing macros are used, like `#define JERRY_CONTEXT(field) (jerry_global_context.field)`

jerry-core/jerryscript-port.h
martijnthe: OK, fair enough, probably quite rare indeed.
I guess the advantage of this split is that some ports may want to control where to allocate the various blocks.

jerry-core/jcontext/jcontext.h
jiangzidong: In this way, the users should provide setter function for `heap`, `context`, `lcache` and `instance (for get the size, row-count, row-length)`. Will it be too complicated for users?

jerry-core/jerryscript-port.h
jiangzidong: I need to find a way to make sure in the big slab memory, the heap start with the address which align with 8 bytes

jerry-core/jcontext/jcontext.h
martijnthe: Hmm.. yeah..

I guess another advantage of the "one big slab of mem" approach is that these getters would be part of jerry-core and the user would only have to provide a getter for that one big buffer.

zherczeg: Yes, this can be an option for this approach (those who select this mode at compile time have to deal with a single pointer).

jiangzidong: On second thoughts, I don't think "one big slab of mem" is better than current jerry_instance_t (with 3 buffers)

1. > I guess another advantage of the "one big slab of mem" approach is that these getters would be part of jerry-core and the user would only have to provide a getter for that one big buffer.

Current jerry_instance_y can also support this. User just provide the getter for the instance, and other getters are in jerry-core.

2. > I wonder if it's desirable to "expose" that internally there is a separate struct for the context, the heap and the lcache.

Even if we use the one big buffer, we still needs the information of "heap size", "lcache size" or "offset of these 3 parts"  in the jerry_instance_p so that the helper function can find the context, heap, lcache. in other words, we have to expose the exists of these 3 structs.

3. current jerry_instance_t is more flexible, because the 3 buffers can from diff places, e.g. one is in heap, one is mapped from file, one is global array, ...

martijnthe: As per the phone call I just had with @jiangzidong, let's go for the 3 separate allocations design.

I think there's merit in letting enabling the port to control where the different pieces are allocated. Zoltan also has a good point that this is not likely to change very often.

jiangzidong:
@zherczeg I still want to put the feature in the unix-main, because I need to test this feature by running jerry-tests and jerry-test-suite. What do you think?

martijnthe:
Can you add a TODO to add multiple instances to the `jerry-context-test.c` test in case https://github.com/jerryscript-project/jerryscript/pull/1784 lands before this one?

jiangzidong:
@martijnthe  I need to find a way to put the external context into the unittest framwork.

martijnthe:
> I need to find a way to put the external context into the unit-core framework, because it needs the build option : --jerry-libc=off --external-context=on

Yes, makes sense. I haven't looked into the unit test set up yet. Are (certain) tests being run with different build configs?

jiangzidong:
@martijnthe I don't think so. They build configs are mainly works on the jerry-core, and the unittest just link the libjerry-core.a

jerry-core/ecma/base/ecma-lcache.c
```diff
@@ -96,9 +89,9 @@ ecma_lcache_insert (ecma_object_t *object_p, /**< object */
 
   lit_string_hash_t name_hash = ecma_string_get_property_name_hash (*prop_p, name_cp);
   size_t row_index = ecma_lcache_row_index (object_cp, name_hash);
-  ecma_lcache_hash_entry_t *entries_p = JERRY_HASH_TABLE_CONTEXT (table)[row_index];
+  ecma_lcache_hash_entry_t *entries_p = &JERRY_HASH_TABLE_CONTEXT (table)[row_index * ECMA_LCACHE_HASH_ROW_LENGTH];
```
zherczeg: JERRY_HASH_TABLE_CONTEXT (table) + (row_index * ECMA_LCACHE_HASH_ROW_LENGTH)

jerry-core/jcontext/jcontext.c
```diff
@@ -19,6 +19,7 @@
  * @{
  */
 
+#ifndef JERRY_ENABLE_EXTERNAL_CONTEXT
```
zherczeg: I would like special cases first, and default case last. When we have multiple context types, that will look better.

jerry-core/jcontext/jcontext.h
```diff
@@ -187,6 +190,46 @@ extern jerry_hash_table_t jerry_global_hash_table;
...
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+
+jerry_hash_table_t *jerry_context_get_current_lcache (void);
+
+#define JERRY_HASH_TABLE_CONTEXT(field) (jerry_context_get_current_lcache ()->field)
+
+#undef ECMA_LCACHE_HASH_ROWS_COUNT
+#define ECMA_LCACHE_HASH_ROWS_COUNT (jerry_port_get_current_instance ()->lcache_row_count)
```
zherczeg: Undefing ECMA_LCACHE_HASH_ROWS_COUNT after the definition above:

    ecma_lcache_hash_entry_t table[ECMA_LCACHE_HASH_ROWS_COUNT * ECMA_LCACHE_HASH_ROW_LENGTH];

This will surely not work.

jerry-core/ecma/base/ecma-lcache.c
```diff
@@ -72,7 +65,7 @@ ecma_lcache_row_index (jmem_cpointer_t object_cp, /**< compressed pointer to obj
 {
   /* Randomize the hash of the property name with the object pointer using a xor operation,
    * so properties of different objects with the same name can be cached effectively. */
-  return (size_t) ((name_hash ^ object_cp) & ECMA_LCACHE_HASH_MASK);
+  return (size_t) ((name_hash ^ object_cp) & (ECMA_LCACHE_HASH_ROWS_COUNT - 1));
```
zherczeg: Is there an assert somewhere which checks that ECMA_LCACHE_HASH_ROWS_COUNT is a power of 2?

jerry-core/jerryscript.h
```diff
@@ -389,6 +389,11 @@ size_t jerry_parse_and_save_literals (const jerry_char_t *source_p, size_t sourc
                                       uint32_t *buffer_p, size_t buffer_size, bool is_c_format);
 
 /**
+ * External context functions.
+ */
+size_t jerry_get_context_size (void);
+size_t jerry_get_lcache_size (size_t row_count, size_t row_length);
```
zherczeg: This will be public API? I don't expect a user to understand what is lcache, and I would like to reserve the right to change such internal stuff (e.g. replace lcache with something else). So we definitely need something better here.

jerry-core/jmem/jmem-heap.c
zherczeg: Change the sentence to normal English sentence.

jerry-core/ecma/base/ecma-lcache.c
jiangzidong: No, and I can put the assert inside `ecma_lcache_init`

jerry-core/jcontext/jcontext.h
jiangzidong: Sorry that I don't understand this comment.
the #undef is after the ecma_lcache_hash_entry_t table[ECMA_LCACHE_HASH_ROWS_COUNT * ECMA_LCACHE_HASH_ROW_LENGTH]; and it can work well.

jerry-core/jerryscript.h
martijnthe: I think that's a consequence of going for 3 separate allocations for heap, lcache and context.
See discussion here https://github.com/jerryscript-project/jerryscript/pull/1778/files#r113155225
(Somehow I interpreted your answers there as being in favor of the making the user allocate 3 separate memory blocks.)

Suggestions on how to simplify the API?

I had suggested something like this [before]:

To create a context, we could also have an API like:
```
typedef struct {
  // user's allocator function, will be called multiple times during jerry_instance_create().
  // semantics are the same as malloc.
  void *(*alloc)(size_t size_in_bytes);

  // JS heap size. Leave zero to pick default size.
  size_t heap_size;

  // LCache sizes. Leave any zero to pick the default size.
  struct {
   size_t row_count;
   size_t row_length;
  } lcache;
} jerry_instance_config_t;

jerry_instance_t *jerry_instance_create(const jerry_instance_config_t *config);
```
`jerry_instance_t *` can be a fwd declared struct.
This way we're spilling less of the internals.

@zherczeg what do you think about that?

[before]: https://github.com/jerryscript-project/jerryscript/issues/1746#issuecomment-295645286

jerry-core/jcontext/jcontext.h
zherczeg: You redefine the constant, so the size of the hash table will be different than the size defined in the header (which uses a constant, not a function call).

jiangzidong: Yes, and that's the reason I did't use the sizeof (table) in the source code.

What about defining another the `jerry_hash_table_t` and `jmem_heap_t` under JERRY_ENABLE_EXTERNAL_CONTEXT, like
```
typedef struct
{
  ecma_lcache_hash_entry_t *table;
} jerry_hash_table_t;

typedef struct
{
  jmem_heap_free_t first; /**< first node in free region list */
  uint8_t *area; /**< heap area */
} jmem_heap_t;

```

jerry-core/jerryscript.h
jiangzidong: My proposal:
* put the `table` of `jerry_hash_table_t` into `jerry_context_t`, and remove the jerry_global_hash_table. In other words, lcache become the member of jerry_context

* since the `ECMA_LCACHE_HASH_ROWS_COUNT` and `ECMA_LCACHE_HASH_ROW_LENGTH` is defined in `ecma-globals.h`, and cannot be changed via public api or config option. So in this patch, we still don't expose the size/existence of the lcache to public.

zherczeg: My problem is we export internals to users, and these internals can change, while public API should not. Furthermore if I understand correctly, the problem with jerry_instance_create that it cannot allocate memory (since jerry does not depend on malloc).

Perhaps I would create a function, which has a simple configuration parameter (this is just a scratch, better names are needed):

    enum {
      type_maximum_performance,
      type_minimal_memory,
    } type;

    size_t jerry_get_instance_size(type, heap_size, index);

The heap_size is a generic parameter for the recommended heap size, but the function can change (e.g. align memory) it.

If you pass 0 as index, it return the number of spaces (e.g. 3 in the current implementation). Then you call it 3 times with index 1, 2 and 3 (type, heap_size must not be changed). You allocate memory for each size returned. And you call another function passing these pointers as an array of pointers.

jiangzidong: Can we decide that there only will be 2 "globals" for each instance, one is heap, and the other is context. In the future, if we want to change/extend the "globals", we just change/add the items in context.

zherczeg: You can decide this alone for your extension. Internally, we will have 3 items, but you exports only two in the public interface. Internally you add the size of the context and lcache together, and provide it as a single item.

Why 3 items: CPU instructions has a limited immediate range for reading memory locations (especially ARM, I know intel has 32 bit offset). So if we put everything together we might frequently for the cpu to construct a read from 2 instructions (increase binary size). So I would not want the basic context grow too big.

martijnthe: > My problem is we export internals to users, and these internals can change, while public API should not.

Hmm this is kind of the opposite of what you suggested [earlier]... ;P

I think using a [config struct] + making `jerry_instance_t` a fwd declaration helps with future API compatibility (at least semantically / "do projects still compile"):

If the internals would be removed -- say, just for the sake of discussion: LCache would be removed (unlikely scenario, I know): the `lcache` struct field would be marked deprecated, but left in place (so projects still compile, but with a warning) and its values would be ignored at run time.

If the internals would be extended -- say some fancy new regex cache is added: new config fields are added to the `jerry_instance_config_t` struct. For existing projects using the API, the values would be zero and thus defaults would be used.

> Furthermore if I understand correctly, the problem with jerry_instance_create that it cannot allocate memory (since jerry does not depend on malloc).

No, you don't need malloc to implement that `alloc` callback. The simplest callback would select a statically allocated buffer and return it.

Re. `jerry_get_instance_size`: you would still have to expose the internals of `jerry_instance_t`, so if the internals change, you'd have to change that struct and break existing projects. In [my proposal], that struct is internal / hidden / forward declared.

[earlier]: https://github.com/jerryscript-project/jerryscript/pull/1778/files#r113156069
[my proposal]: https://github.com/jerryscript-project/jerryscript/pull/1778#discussion_r114267745
[config struct]: https://github.com/jerryscript-project/jerryscript/pull/1778#discussion_r114267745

jiangzidong:
@martijnthe @zherczeg I just refactor the patch. 
* the API : I took the proposal from @martijnthe , but I don't use the `jerry_instance_config_t`, since when the internal of `jerry_instance_t` changed, people still need to set/change the item of `jerry_instance_config_t`. 

the API is 
```
jerry_instance_t *jerry_create_instance (uint32_t heap_size, jerry_instance_alloc_t alloc);
```

* LCACHE_ROW_LENGTH and COUNT is internal parameter, and user will not change/access them.

jerry-port/default/default-external-context.c
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript-port.h"
+#include "jerryscript-port-default.h"
+
+jerry_instance_t *current_instance_p = NULL;
```
akosthekiss: is it necessary to expose? keep it static?

jerry-port/default/include/jerryscript-port-default.h
```diff
@@ -40,6 +40,8 @@ void jerry_port_default_set_log_level (jerry_log_level_t level);
 void jerry_port_default_jobqueue_init (void);
 jerry_value_t jerry_port_default_jobqueue_run (void);
 
+void jerry_port_set_instance (jerry_instance_t *instance_p);
```
akosthekiss: this is not a generic jerry port API function but an addition by the default port implementation. all such functions should be prefixed by jerry_port_default_ (not by jerry_port_) to highlight that they are not available in all ports

jerry-core/jerryscript-port.h
```diff
@@ -165,6 +166,8 @@ typedef uint32_t (*jerry_job_handler_t) (void *);
  */
 void jerry_port_jobqueue_enqueue (jerry_job_handler_t handler, void *job_p);
 
+jerry_instance_t *jerry_port_get_current_instance (void);
```
akosthekiss: doc comment is definitely needed

jerry-port/default/default-external-context.c
jiangzidong: Oh, sorry that I removed the `static` for easy-debug but forgot to re-add it.

jerry-core/jerryscript.h
```diff
@@ -213,6 +218,21 @@ typedef struct
...
+{
+  struct
+  {
+    uint32_t heap_size; /**< the heap size */
+    uint8_t *context_p; /**< point to the entrance of the context in buffer */
+    uint8_t *heap_p; /**< point to the entrance of the heap in buffer */
+    uint8_t *lcache_p; /**< point to the entrance of the lcache in buffer */
+  } head;
+  uint8_t buffer[]; /**< the flexible array for storing context, heap and lcache */
+} jerry_instance_t;
```
zherczeg: You can create a forward definition to hide this structure.

jerry-core/jerryscript.h
```diff
@@ -402,6 +422,11 @@ jerry_value_t jerry_exec_snapshot (const uint32_t *snapshot_p, size_t snapshot_s
 size_t jerry_parse_and_save_literals (const jerry_char_t *source_p, size_t source_size, bool is_strict,
                                       uint32_t *buffer_p, size_t buffer_size, bool is_c_format);
 
+/*
+ * External context functions.
+ */
+jerry_instance_t *jerry_create_instance (uint32_t heap_size, jerry_instance_alloc_t alloc);
```
zherczeg: I am not a fond of passing callbacks to everything. Makes the code so complicated. Anyway at least add a user pointer, which is passed to the alloc function.

jiangzidong: If we don't pass the alloc function, then we may need 2 api to replace `jerry_create_instance`
``` c
jerry_get_instance_size(size_t heap_size);
jerry_init_instance(jerry_instance_t *instance_p, size_t heap_size)
```
In this way, users should first get the size, then alloc the instance on their own, then init the instance. And users will not know the internal details of the jerry_instance.

What do you think? @martijnthe @zherczeg

jerry-core/jerryscript.h
martijnthe: I agree.

jerry-core/jcontext/jcontext.c
```diff
@@ -48,7 +92,7 @@ jmem_heap_t jerry_global_heap __attribute__ ((aligned (JMEM_ALIGNMENT))) JERRY_G
 jerry_hash_table_t jerry_global_hash_table;
 
 #endif /* !CONFIG_ECMA_LCACHE_DISABLE */
-
+#endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
```
zherczeg: Add a newline after this line.

jerry-core/jcontext/jcontext.h
```diff
@@ -116,6 +117,77 @@ typedef struct
 } jerry_context_t;
 
 /**
+ * Description of jerry instance, which contains the meta data and buffer.
+ */
+struct jerry_instance_t
+{
+  struct
+  {
+    uint32_t heap_size; /**< the heap size */
```
zherczeg: size of the heap

jerry-core/jcontext/jcontext.h
```diff
@@ -116,6 +117,77 @@ typedef struct
...
+
+#define JERRY_CONTEXT(field) (jerry_context_get_current_context ()->field)
+
+#ifndef JERRY_SYSTEM_ALLOCATOR
+
+jmem_heap_t *jerry_context_get_current_heap (void);
+
+#define JERRY_HEAP_CONTEXT(field) (jerry_context_get_current_heap ()->field)
+
+#undef JMEM_HEAP_SIZE
```
zherczeg: Undefing a symbol is dangerous since it only take effect if this header file is included (a hidden dependency). I would prefer that if this context is selected, JMEM_HEAP_SIZE must not be defined. So replace this undef to:

    #ifdef JMEM_HEAP_SIZE
    #error "JMEM_HEAP_SIZE must not be defined if JERRY_ENABLE_EXTERNAL_CONTEXT is defined"
    #endif /* JMEM_HEAP_SIZE */

jerry-port/default/default-external-context.c
```diff
@@ -0,0 +1,39 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript-port.h"
+#include "jerryscript-port-default.h"
+
+static jerry_instance_t *current_instance_p = NULL;
```
zherczeg: There is no comment for this variable. We must mention that this is not thread safe somewhere.

jerry-main/main-unix.c
```diff
@@ -418,6 +418,20 @@ check_usage (bool condition, /**< the condition that must hold */
...
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+
+/**
+ * The alloc function passed to jerry_create_instance
+ */
+static void *
+instance_alloc (size_t size,
+                void *cb_data_p __attribute__((unused)))
+{
+  return malloc (size);
```
zherczeg: Malloc is not necessary available in main-unix.c. Is there a check that system libc must be enabled?

zherczeg:
The 05.PORT-API.md needs to be updated as well.

jerry-core/jerryscript-port.h
```diff
@@ -19,6 +19,7 @@
 #include <stdbool.h>
 #include <stdint.h>
 #include <stdio.h>
+#include "jerryscript.h"
```
zherczeg: A forward definition to jerry_instance_t could replace this include.

jerry-main/main-unix.c
jiangzidong: No. But if the jerry-libc is on, it will throw error when building, just like the case of `--jerry-libc=off --system-allocator=on`.

I can add a check in `main-unix.c`

jiangzidong:
@zherczeg, updated the docs and patches according to the latest comments

docs/05.PORT-API.md
```diff
@@ -94,6 +94,24 @@ bool jerry_port_get_time_zone (jerry_time_zone_t *);
 double jerry_port_get_current_time (void);
 ```
 
+## External instance
+
+Allow user to provide external buffer for jerry instance (which incluldes an isolated context and heap with other instances), so that user can config the heap size in runtime and run multiple JS apps simultaneously.
```
zherczeg: includes

docs/05.PORT-API.md
```diff
@@ -94,6 +94,24 @@ bool jerry_port_get_time_zone (jerry_time_zone_t *);
...
+
+```c
+/**
+ * Get the current instance, which contains the current context, heap and other infomations.
```
zherczeg: information has no plural

jerry-core/api/jerry.c
```diff
@@ -2242,6 +2242,51 @@ jerry_is_valid_cesu8_string (const jerry_char_t *cesu8_buf_p, /**< CESU-8 string
...
+
+#ifndef JERRY_SYSTEM_ALLOCATOR
+  uint8_t *unaligned_heap_p = instance_p->head.lcache_p + lcache_size;
+  uintptr_t alignment_mask = ~(uintptr_t) (JMEM_ALIGNMENT - 1);
+  instance_p->head.heap_p = (uint8_t *) (((uintptr_t) unaligned_heap_p + JMEM_ALIGNMENT - 1) & alignment_mask);
+
+  JERRY_ASSERT (((uintptr_t) instance_p->head.heap_p & (uintptr_t) (JMEM_ALIGNMENT - 1)) == 0);
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+
+  return instance_p;
```
zherczeg: Isn't this function should be enclosed by JERRY_ENABLE_EXTERNAL_CONTEXT ?

jerry-core/include/jerryscript-core.h
```diff
@@ -206,6 +206,11 @@ typedef void *(*jerry_user_context_init_t) (void);
 typedef void (*jerry_user_context_deinit_t) (void *user_context_p);
 
 /**
+ * Function type for allocating buffer in jerry instance.
```
zherczeg: for JerryScript instance

jerry-core/jcontext/jcontext.c
```diff
@@ -19,6 +19,50 @@
...
 
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+
+/**
+ * Get the current context from the current instance.
+ *
+ * @return the pointer to the context
+ */
+jerry_context_t *
+jerry_context_get_current_context (void)
```
zherczeg: should be always inline

jerry-core/jcontext/jcontext.h
```diff
@@ -116,6 +117,80 @@ typedef struct
...
+ * Description of jerry instance, which contains the meta data and buffer.
+ */
+struct jerry_instance_t
+{
+  struct
+  {
+    uint32_t heap_size; /**< size of the heap */
+    uint8_t *context_p; /**< point to the entrance of the context in buffer */
+    uint8_t *heap_p; /**< point to the entrance of the heap in buffer */
+    uint8_t *lcache_p; /**< point to the entrance of the lcache in buffer */
```
zherczeg: CONFIG_ECMA_LCACHE_DISABLE / JERRY_SYSTEM_ALLOCATOR should be checked here as well.

jerry-core/jcontext/jcontext.c
martijnthe: @jiangzidong can you check what the code size difference is between inlining vs not inlining?

jiangzidong: The .text section size: "current": 775134 ==> "inline": 775230 (+100) 

And I also measure the function size by `nm build/bin/jerry -S | grep jerry_get_user_context `
```
void *
jerry_get_user_context (void)
{
  return JERRY_CONTEXT (user_context_p);
} /* jerry_get_user_context */
```
because it's simple and only contain the JERRY_CONTEXT 

"current": 18byte
"inline": 22byte

So in the tradeoff between perf and binary size, `inline` is  prone to perf. 
What should we choose?

jiangzidong: I think in most embedded cases, binary/memory size is more important. Usually the heavy computation is handled by native binding, and JS is more like a task dispatcher, so its performance affects the global performance a little.

martijnthe: What architecture did you build for? 32-bit x86?

jiangzidong: x64

jiangzidong: 32-bit x86: --compile-flag=-m32
size of `jerry_get_user_context`
"current": 19 byte
"inline": 22 byte

ARMv7l: --toolchain=cmake/toolchain_linux_armv7l.cmake
"current": 18 byte
"inline": 22 byte

zherczeg: Probably the ->head.context_p increase the size. Ok, leave it this way until we figure out something better.

jerry-core/api/jerry.c
zherczeg: You may misunderstand my comment. API functions must exists regardless of defines, but these should no nothing if the appropriate define is not available. So move the guards into the function, and make other arguments unused and return with NULL if the define is not available.

jerry-core/jcontext/jcontext.c
zherczeg: "its performance affects the global performance a little." - I am not fully agree with this. People do things on JS level many times.

jerry-core/jcontext/jcontext.h
```diff
@@ -116,6 +117,84 @@ typedef struct
...
+    uint32_t heap_size; /**< size of the heap */
+    uint8_t *context_p; /**< point to the entrance of the context in buffer */
+#ifndef JERRY_SYSTEM_ALLOCATOR
+    uint8_t *heap_p; /**< point to the entrance of the heap in buffer */
+#endif /* !JERRY_SYSTEM_ALLOCATOR */
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+    uint8_t *lcache_p; /**< point to the entrance of the lcache in buffer */
+#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+  } head;
+  uint8_t buffer[]; /**< the flexible array for storing context, heap and lcache */
```
zherczeg: Hm an idea: if context_p is the first member of buffer, it could be accessed without context_p. That would be one less memory read.

    in jerry_context_get_current_context():
    return (jerry_context_t *) (jerry_port_get_current_instance () + 1);

Is buffer[] supported by C99? And do we need this member at all (`jerry_port_get_current_instance () + 1` is the start for the data)? This way we would not need the "head" nested structure as well.

Furthermore inlining this could be good for optimization, since the compiler is clever enough to change `jerry_context_get_current_context()->something` to a `jerry_port_get_current_instance ()+offset` read (no size increase, better performance).

Another alternative: we could make heap_size, heap_p, lcache_p part of the jerry_context_t.

martijnthe: > Hm an idea: if context_p is the first member of buffer, it could be accessed without context_p. That would be one less memory read.

Good idea indeed. It is already the first member in the buffer, plus it's size is known at compile time:

```
...
instance_p->head.context_p = instance_p->buffer;
instance_p->head.lcache_p = instance_p->head.context_p + sizeof (jerry_context_t);
...
```

So, I think we can change `jerry_instance_t` to:

```
struct jerry_instance_t
{
  struct
  {
    uint32_t heap_size; /**< size of the heap */
#ifndef JERRY_SYSTEM_ALLOCATOR
    uint8_t *heap_p; /**< point to the entrance of the heap in buffer */
#endif /* !JERRY_SYSTEM_ALLOCATOR */
#ifndef CONFIG_ECMA_LCACHE_DISABLE
    uint8_t *lcache_p; /**< point to the entrance of the lcache in buffer */
#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
    jerry_context_t context;
  } head;
  uint8_t buffer[]; /**< the flexible array for storing heap and lcache */
}
```

Personally, I do like the `buffer[]` member, it's more self-documenting than without it.

jiangzidong: `buffer[]` is c99 syntax. https://en.wikipedia.org/wiki/Flexible_array_member

@martijnthe I will try your proposed `jerry_instance_t`. I think we can also remove the `head`, and let `heap_size`, `heap_p` ... be the first level member of jerry_instance

martijnthe: > be the first level member of jerry_instance

Yeah, makes sense.

jiangzidong:
@martijnthe @zherczeg I use `#define JERRY_CONTEXT(field) (jerry_port_get_current_instance ()->context.field)` now. 
and in x64, the size of jerry_get_user_context is 18byte

jerry-core/jcontext/jcontext.h
```diff
@@ -121,6 +122,88 @@ typedef struct
 } jerry_context_t;
 
 /**
+ * Description of jerry instance, which contains the meta data and buffer.
+ */
+struct jerry_instance_t
+{
+
+  uint32_t heap_size; /**< size of the heap */
```
zherczeg: Remove newline before.

jerry-core/jcontext/jcontext.h
```diff
@@ -121,6 +122,88 @@ typedef struct
...
+ * This part is for Jerry which enable external context.
+ */
+
+typedef struct
+{
+  jmem_heap_free_t first; /**< first node in free region list */
+  uint8_t area[]; /**< heap area */
+} jmem_heap_t;
+
+#define JERRY_CONTEXT(field) (jerry_port_get_current_instance ()->context.field)
```
zherczeg: Excellent!

jerry-core/jcontext/jcontext.h
```diff
@@ -121,6 +122,87 @@ typedef struct
...
+};
+
+#ifndef CONFIG_ECMA_LCACHE_DISABLE
+/**
+ * Hash table for caching the last access of properties.
+ */
+typedef struct
+{
+  /**
+   * Hash table
```
LaszloLango: I think this comment is an unnecessary duplicate. Please remove it.

PR #1780, opened 2017-04-24
Fix the exception handler bug

polaroi8d:


LaszloLango:
Is it possible to write a test for this? It would great to add more description. "THE bug" is not obvious for me. :) What was the original problem?

polaroi8d:
The original problem, if I want to evaluate a variable out of the scope in eval mode, the engine was not working. I don't think it need a new tests.

zherczeg:
The strange thing is that existing tests don't catch this. Do we have a test which tests syntax error in eval?

polaroi8d:
No.

LaszloLango:
Please write a test and update the commit message.

PR #1782, opened 2017-04-25
Remove `jerry-internal.h`

akosthekiss:
The internal header was a legacy, declared two functions only. One
of them was not used anymore at all, the other was used at a single
call site. Moreover, their documentation was outdated. This patch
removes the header and the related function implementations -- as
well as the temptation to use such internal functions from outside
the library.


yichoi:
LGTM

PR #1783, opened 2017-04-26
Remove unnecessary subfolder in `tests/jerry/fail/`

robertsipka:
There is no need to create error code named subfolders since every failing tests return with the same code.


tools/runners/run-test-suite.sh
```diff
@@ -118,13 +118,13 @@ ENGINE_TEMP=`mktemp engine-out.XXXXXXXXXX`
 
 for test in `cat $TEST_FILES`
 do
-    error_code=`echo $test | grep -e "^.\/fail\/[0-9]*\/" -o | cut -d / -f 3`
-    if [ "$error_code" = "" ]
+    if [[ $test =~ ^\.\/fail\/ ]];
```
rtakacs: Semicolon is not necessary at the end of the line.

rtakacs:
@LaszloLango I think if there are no different exit codes, we should remove the sub-folder. If someone will support different codes in the near (or far) future, the folder structure can be restored.

zherczeg:
I think there are different error codes when out-of-memory happens, but I am not 100% sure. Could you check it? But for JS error there should only be one error code.

robertsipka:
@zherczeg : Yes, there are different error codes (fatal codes):
- ERR_OUT_OF_MEMORY (10),
- ERR_SYSCALL (11),
- ERR_REF_COUNT_LIMIT (12),
- ERR_FAILED_INTERNAL_ASSERTION (120)

For now, if one of the test cases returns with a fatal error, the test system will catch it and report that with the throwed error code.
IMHO we shouldn't add test cases for fatal failures.

LaszloLango:
@robertsipka you are right. We should not do it. I think you may land this patch, because there were no objections.

PR #1784, opened 2017-04-26
Two extensions: An extended context with a compile-time-defined number of pointers, and modules

gabrielschulhof:
Much like #1740 this also introduces infrastructure for multiple extensions, taking into account my [comment](https://github.com/jerryscript-project/jerryscript/pull/1740#issuecomment-297245510). It also moves the user context deinit to the top of `jerry_cleanup ()` because the context has to be functional when deinit the user context, since the user context may contain `jerry_value_t` variables and may need to make API calls.

### DIFF
--- a/jerry-ext/context/include/jerry-context.h
+++ b/jerry-ext/context/include/jerry-context.h
@@ -1,0 +1,52 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRYX_CONTEXT_H
+#define JERRYX_CONTEXT_H
+
+#include "jerryscript.h"
+
+/**
+ * Function to pass to jerry_init_with_user_context() as its second parameter
+ */
+void *jerryx_context_init (void);
+
+/**
+ * Function to pass to jerry_init_with_user_context() as its third parameter
+ */
+void jerryx_context_deinit (void *user_context_p);
+
+typedef struct
+{
+  jerry_user_context_init_cb init_cb;
+  jerry_user_context_deinit_cb deinit_cb;
+  int index;
+} jerryx_context_slot_t;
+
+#define JERRYX_CONTEXT_DEFINE_SLOT(slot_name, init_cb_ptr, deinit_cb_ptr)         \
+static jerryx_context_slot_t __jerryx_context_slot_ ## slot_name                  \
+__attribute__((used, section("jerryx_context_slots"), aligned(sizeof(void *)))) = \
+{                                                                                 \
+  .init_cb = (init_cb_ptr),                                                       \
+  .deinit_cb = (deinit_cb_ptr),                                                   \
+  .index = -1                                                                     \
+};
+
+void *jerryx_context_get (int slot);
+
+#define JERRYX_CONTEXT_SLOT(slot_name) \
+  (jerryx_context_get (__jerryx_context_slot_ ## slot_name.index))
+
+#endif /* !JERRYX_CONTEXT_H */
--- a/jerry-ext/context/jerry-context.c
+++ b/jerry-ext/context/jerry-context.c
@@ -1,0 +1,120 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string.h>
+#include "jerry-context.h"
+#include "jmem.h"
+
+static int jerryx_context_slot_count = -1;
+extern jerryx_context_slot_t __start_jerryx_context_slots[] __attribute__((weak));
+extern jerryx_context_slot_t __stop_jerryx_context_slots[] __attribute__((weak));
+#define JERRYX_CONTEXT_ITERATE_OVER_SLOTS(index, slot_p)              \
+(index = 0, slot_p = &__start_jerryx_context_slots[index];            \
+  &__start_jerryx_context_slots[index] < __stop_jerryx_context_slots; \
+  index++, slot_p = &__start_jerryx_context_slots[index])
+
+static void jerryx_context_static_init (void);
+static void
+jerryx_context_static_init (void)
+{
+  jerryx_context_slot_t *slot_p;
+  int index;
+
+  if (jerryx_context_slot_count >= 0)
+  {
+    jerry_port_fatal (ERR_FAILED_INTERNAL_ASSERTION);
+  }
+
+  for JERRYX_CONTEXT_ITERATE_OVER_SLOTS (index, slot_p)
+  {
+    slot_p->index = index;
+  }
+
+  jerryx_context_slot_count = index;
+} /* jerryx_context_static_init */
+
+/* Context-free data */
+
+/**
+ * Create a new context. This function can be passed to jerry_init_with_user_context () as its second parameter.
+ *
+ * @return the pointer to store as the user context.
+ */
+void *
+jerryx_context_init (void)
+{
+  void **ret_pp;
+  int index;
+  size_t context_size;
+  jerryx_context_slot_t *slot_p = NULL;
+
+  if (unlikely (jerryx_context_slot_count < 0))
+  {
+    jerryx_context_static_init ();
+  }
+
+  context_size = ((size_t) jerryx_context_slot_count) * sizeof (void *);
+  ret_pp = (void **) jmem_heap_alloc_block (context_size);
+  memset (ret_pp, 0, context_size);
+
+  for JERRYX_CONTEXT_ITERATE_OVER_SLOTS (index, slot_p)
+  {
+    ret_pp[slot_p->index] = slot_p->init_cb ();
+  }
+
+  return ((void *) ret_pp);
+} /* jerryx_context_init */
+
+/**
+ * Free a context. This function can be passed to jerry_init_with_user_context () as its third parameter.
+ */
+void
+jerryx_context_deinit (void *user_context_p) /**< user context to deinit */
+{
+  int index;
+  jerryx_context_slot_t *slot_p = NULL;
+  void **slots_p = NULL;
+
+  if (user_context_p)
+  {
+    slots_p = (void **) user_context_p;
+    for JERRYX_CONTEXT_ITERATE_OVER_SLOTS (index, slot_p)
+    {
+      if (slot_p->deinit_cb)
+      {
+        slot_p->deinit_cb (slots_p[slot_p->index]);
+      }
+    }
+    jmem_heap_free_block (slots_p, ((size_t) jerryx_context_slot_count) * sizeof (void *));
+  }
+} /* jerryx_context_deinit */
+
+/**
+ * Request the slot at the given index.
+ *
+ * @return the user data that was created in the given slot when the context was initialized.
+ */
+void *
+jerryx_context_get (int slot) /**< the slot from which to retrieve the data */
+{
+  void **slots_p = (void **) jerry_get_user_context ();
+
+  if (jerryx_context_slot_count < 0 || !slots_p || slot < 0 || slot >= jerryx_context_slot_count)
+  {
+    jerry_port_fatal (ERR_FAILED_INTERNAL_ASSERTION);
+  }
+
+  return slots_p[slot];
+} /* jerryx_context_get */
--- a/jerry-ext/module/include/jerry-module.h
+++ b/jerry-ext/module/include/jerry-module.h
@@ -1,0 +1,56 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRYX_MODULE_H
+#define JERRYX_MODULE_H
+
+#include "jerryscript.h"
+
+typedef struct
+{
+  jerry_char_t *name;
+  jerry_value_t (*on_resolve)(void);
+} jerryx_module_t;
+
+/**
+ * Function to create a module manager context
+ */
+void *jerryx_module_manager_init (void);
+
+/**
+ * Function to clean up a given module manager context
+ */
+void jerryx_module_manager_deinit (void *context_p);
+
+/**
+ * Function to load a copy of a module into the current context
+ */
+jerry_value_t jerryx_module_resolve (const jerry_char_t *name);
+
+#define JERRYX_MODULE(module_name, on_resolve_cb)                      \
+static jerryx_module_t _module \
+__attribute__((used, section("jerryx_modules"), aligned(sizeof(void *)))) = \
+{ \
+  .name = ((jerry_char_t *)#module_name), \
+  .on_resolve = (on_resolve_cb) \
+};
+
+#define JERRYX_MODULE_RESOLVER(cb_name)                                              \
+static jerry_value_t cb_name (const jerry_char_t *name);                             \
+static jerry_value_t (*__static_global_ ## cb_name) (const jerry_char_t *)           \
+__attribute__((used)) __attribute__((section("jerryx_module_resolvers"))) = cb_name; \
+static jerry_value_t cb_name (const jerry_char_t *name)
+
+#endif /* !JERRYX_MODULE_H */
--- a/jerry-ext/module/jerry-module.c
+++ b/jerry-ext/module/jerry-module.c
@@ -1,0 +1,198 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jmem.h"
+#include "jerryscript.h"
+#include "jerry-module.h"
+
+typedef struct jerryx_module_header
+{
+  jerry_char_t *name;
+  struct jerryx_module_header *next_p;
+} jerryx_module_header_t;
+
+#define JERRYX_MODULE_HEADER_UNLINK(header) \
+  ((jerryx_module_header_t *) (header))->next_p = NULL
+
+#define JERRYX_MODULE_HEADER_RUNTIME_INIT(header, module_name) \
+  ((jerryx_module_header_t *) (header))->name = (module_name); \
+  JERRYX_MODULE_HEADER_UNLINK((header))
+
+static void
+jerryx_module_header_insert (jerryx_module_header_t *header_p,
+                             jerryx_module_header_t **root_pp)
+{
+  if (*root_pp)
+  {
+    header_p->next_p = (*root_pp);
+  }
+  (*root_pp) = header_p;
+} /* jerryx_module_header_insert */
+
+static jerryx_module_header_t *
+jerryx_module_header_find (jerryx_module_header_t *root_p,
+                           const jerry_char_t *name,
+                         jerryx_module_header_t **previous_pp)
+{
+  for (;root_p && strcmp ((char *) name, ((char *) root_p->name)); root_p = root_p->next_p)
+  {
+    if (previous_pp)
+    {
+      (*previous_pp) = root_p;
+    }
+  }
+
+  return root_p;
+} /* jerryx_module_header_find */
+
+static void
+jerryx_module_header_free (jerryx_module_header_t *root_p,
+                           void (*free_link) (jerryx_module_header_t *))
+{
+  jerryx_module_header_t *next_p = NULL;
+  while (root_p)
+  {
+    next_p = root_p->next_p;
+    free_link (root_p);
+    root_p = next_p;
+  }
+} /* jerryx_module_header_free */
+
+typedef struct
+{
+  jerryx_module_header_t header;
+  jerry_value_t export;
+} jerryx_module_instance_t;
+
+static void
+jerryx_module_free_instance (jerryx_module_header_t *header_p)
+{
+  jerryx_module_instance_t *instance_p = (jerryx_module_instance_t *) header_p;
+  jerry_release_value (instance_p->export);
+  jmem_heap_free_block (instance_p, sizeof (jerryx_module_instance_t));
+} /* jerryx_module_free_instance */
+
+void *
+jerryx_module_manager_init (void)
+{
+  void *ret = jmem_heap_alloc_block (sizeof (jerryx_module_instance_t *));
+  memset (ret, 0, sizeof (jerryx_module_instance_t *));
+  return ret;
+} /* jerryx_module_manager_init */
+
+void
+jerryx_module_manager_deinit (void *context_p)
+{
+  if (context_p)
+  {
+    jerryx_module_header_free (*((jerryx_module_header_t **) (context_p)), jerryx_module_free_instance);
+    jmem_heap_free_block (context_p, sizeof (jerryx_module_header_t *));
+  }
+} /* jerryx_module_manager_deinit */
+
+#ifdef JERRYX_MODULE_HAVE_CONTEXT
+#include "jerry-context.h"
+JERRYX_CONTEXT_DEFINE_SLOT (jerryx_module_slot, jerryx_module_manager_init, jerryx_module_manager_deinit)
+#define JERRYX_MODULE_CONTEXT \
+  JERRYX_CONTEXT_SLOT(jerryx_module_slot)
+#else /* !JERRYX_MODULE_HAVE_CONTEXT */
+#define JERRYX_MODULE_CONTEXT \
+  (jerry_get_user_context ())
+#endif /* JERRYX_MODULE_HAVE_CONTEXT */
+
+/*
+ * These symbols are made available by the linker
+ */
+extern jerryx_module_t __start_jerryx_modules[] __attribute__((weak));
+extern jerryx_module_t __stop_jerryx_modules[] __attribute__((weak));
+#define JERRYX_MODULE_ITERATE_OVER_MODULES(index, module_p) \
+(index = 0, module_p = &__start_jerryx_modules[index];     \
+  &__start_jerryx_modules[index] < __stop_jerryx_modules;  \
+  index++, module_p = &__start_jerryx_modules[index])
+
+JERRYX_MODULE_RESOLVER (native_resolver)
+{
+  int index;
+  jerry_value_t ret = 0;
+  jerryx_module_t *module_p = NULL;
+  jerryx_module_instance_t **instances_pp = JERRYX_MODULE_CONTEXT;
+  jerryx_module_instance_t *instance_p = NULL;
+
+  instance_p =
+  (jerryx_module_instance_t *) jerryx_module_header_find (*((jerryx_module_header_t **) instances_pp), name, NULL);
+
+  if (instance_p)
+  {
+    return instance_p->export;
+  }
+
+  for JERRYX_MODULE_ITERATE_OVER_MODULES (index, module_p)
+  {
+    if (!strcmp ((char *) module_p->name, (char *) name))
+    {
+      ret = module_p->on_resolve ();
+      instance_p = (jerryx_module_instance_t *) jmem_heap_alloc_block (sizeof (jerryx_module_instance_t));
+      JERRYX_MODULE_HEADER_RUNTIME_INIT (instance_p, module_p->name);
+      instance_p->export = jerry_acquire_value (ret);
+      jerryx_module_header_insert ((jerryx_module_header_t *) instance_p, (jerryx_module_header_t **) instances_pp);
+      return ret;
+    }
+  }
+
+  return 0;
+} /* JERRYX_MODULE_RESOLVER */
+
+static const char not_found_prologue[] = "Module '";
+static const char not_found_epilogue[] = "' not found";
+
+/*
+ * These symbols are made available by the linker
+ */
+typedef jerry_value_t (*jerryx_module_resolver_t)(const jerry_char_t *);
+extern jerryx_module_resolver_t __start_jerryx_module_resolvers[] __attribute__((weak));
+extern jerryx_module_resolver_t __stop_jerryx_module_resolvers[] __attribute__((weak));
+
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name)
+{
+  jerry_value_t ret;
+  char *error_message = NULL;
+  size_t error_message_size = 0, name_length = strlen ((char *) name);
+  size_t index, resolver_count = (size_t) (__stop_jerryx_module_resolvers - __start_jerryx_module_resolvers);
+
+  for (index = 0; index < resolver_count; index++)
+  {
+    ret = __start_jerryx_module_resolvers[index] (name);
+    if (ret)
+    {
+      return ret;
+    }
+  }
+
+  error_message_size = name_length + sizeof (not_found_prologue) + sizeof (not_found_epilogue) - 1;
+  error_message = jmem_heap_alloc_block_null_on_error (error_message_size);
+  if (error_message)
+  {
+    memcpy (error_message, not_found_prologue, sizeof (not_found_prologue) - 1);
+    memcpy (&error_message[sizeof (not_found_prologue) - 1], ((char *) name), name_length);
+    memcpy (&error_message[sizeof (not_found_prologue) - 1 + name_length], not_found_epilogue,
+            sizeof (not_found_epilogue) - 1);
+    error_message[error_message_size - 1] = 0;
+    ret = jerry_create_error (JERRY_ERROR_COMMON, (jerry_char_t *) error_message);
+    jmem_heap_free_block (error_message, error_message_size);
+  }
+
+  return ret;
+} /* jerryx_module_resolve */
--- a/tests/unit-core/jerry-ext/module/my-custom-module.c
+++ b/tests/unit-core/jerry-ext/module/my-custom-module.c
@@ -1,0 +1,25 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript.h"
+#include "jerry-module.h"
+
+static jerry_value_t
+my_custom_module_init (void)
+{
+  return jerry_create_number (42);
+} /* my_custom_module_init */
+
+JERRYX_MODULE (my_custom_module, my_custom_module_init)
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ env:
     - OPTS="--unittests"
     - OPTS="--test262"
     - OPTS="--check-pylint"
+    - OPTS="--unittests --buildoptions=--cmake-param=-DJERRYX_CONTEXT=ON"
   global:
     # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
     #   via the "travis encrypt" command using the project repo's public key
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,8 @@ env:
     - OPTS="--test262"
     - OPTS="--check-pylint"
     - OPTS="--unittests --buildoptions=--cmake-param=-DJERRYX_CONTEXT=ON"
+    - OPTS="--unittests --buildoptions=--cmake-param=-DJERRYX_MODULE=ON,--cmake-param=-DJERRYX_CONTEXT=ON"
+    - OPTS="--unittests --buildoptions=--cmake-param=-DJERRYX_MODULE=ON"
   global:
     # The next declaration is the encrypted COVERITY_SCAN_TOKEN, created
     #   via the "travis encrypt" command using the project repo's public key
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -250,3 +250,5 @@ if(UNITTESTS)
     add_subdirectory(tests/unit-libm)
   endif()
 endif()
+
+add_subdirectory(jerry-ext)
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -187,6 +187,11 @@ jerry_cleanup (void)
 {
   jerry_assert_api_available ();
 
+  if (JERRY_CONTEXT (user_context_deinit_cb))
+  {
+    JERRY_CONTEXT (user_context_deinit_cb) (JERRY_CONTEXT (user_context_p));
+  }
+
   ecma_finalize ();
 
 #ifdef JERRY_DEBUGGER
@@ -198,11 +203,6 @@ jerry_cleanup (void)
 
   jmem_finalize ();
   jerry_make_api_unavailable ();
-
-  if (JERRY_CONTEXT (user_context_deinit_cb))
-  {
-    JERRY_CONTEXT (user_context_deinit_cb) (JERRY_CONTEXT (user_context_p));
-  }
 } /* jerry_cleanup */
 
 /**
--- a/jerry-ext/CMakeLists.txt
+++ b/jerry-ext/CMakeLists.txt
@@ -1,0 +1,26 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set(JERRYX_CONTEXT OFF CACHE BOOL "Build context support?")
+
+if("${PLATFORM}" STREQUAL "DARWIN")
+  set(JERRYX_CONTEXT "OFF")
+endif()
+
+# Status messages
+message(STATUS "JERRYX_CONTEXT            " ${JERRYX_CONTEXT})
+
+if(JERRYX_CONTEXT)
+  add_subdirectory(context)
+endif()
--- a/jerry-ext/CMakeLists.txt
+++ b/jerry-ext/CMakeLists.txt
@@ -13,14 +13,21 @@
 # limitations under the License.
 
 set(JERRYX_CONTEXT OFF CACHE BOOL "Build context support?")
+set(JERRYX_MODULE  OFF CACHE BOOL "Build module support?")
 
 if("${PLATFORM}" STREQUAL "DARWIN")
   set(JERRYX_CONTEXT "OFF")
+  set(JERRYX_MODULE "OFF")
 endif()
 
 # Status messages
 message(STATUS "JERRYX_CONTEXT            " ${JERRYX_CONTEXT})
+message(STATUS "JERRYX_MODULE             " ${JERRYX_MODULE})
 
 if(JERRYX_CONTEXT)
   add_subdirectory(context)
+endif()
+
+if(JERRYX_MODULE)
+  add_subdirectory(module)
 endif()
--- a/jerry-ext/context/CMakeLists.txt
+++ b/jerry-ext/context/CMakeLists.txt
@@ -1,0 +1,37 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required (VERSION 2.8.12)
+set(JERRYX_CONTEXT_NAME jerryx-context)
+project (${JERRYX_CONTEXT_NAME} C)
+
+# Include directories
+set(INCLUDE_CONTEXT "${CMAKE_CURRENT_SOURCE_DIR}/include")
+set(DEFINES_CONTEXT "")
+
+# Source directories
+file(GLOB SOURCE_CONTEXT *.c)
+
+add_library(${JERRYX_CONTEXT_NAME} STATIC ${SOURCE_CONTEXT})
+set_property(TARGET ${JERRYX_CONTEXT_NAME}
+             PROPERTY COMPILE_FLAGS "${DEFINES_CONTEXT}")
+
+target_include_directories(${JERRYX_CONTEXT_NAME} PRIVATE ${INCLUDE_CONTEXT})
+target_include_directories(${JERRYX_CONTEXT_NAME} INTERFACE ${INCLUDE_CONTEXT})
+target_include_directories(${JERRYX_CONTEXT_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-core")
+target_include_directories(${JERRYX_CONTEXT_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-core/jmem")
+target_include_directories(${JERRYX_CONTEXT_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-core/jrt")
+
+install(TARGETS ${JERRYX_CONTEXT_NAME} DESTINATION lib)
+install(DIRECTORY ${INCLUDE_CONTEXT}/ DESTINATION include/jerry-ext)
--- a/jerry-ext/module/CMakeLists.txt
+++ b/jerry-ext/module/CMakeLists.txt
@@ -1,0 +1,43 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required (VERSION 2.8.12)
+set(JERRYX_MODULE_NAME jerryx-module)
+project (${JERRYX_MODULE_NAME} C)
+
+# Include directories
+set(INCLUDE_MODULE "${CMAKE_CURRENT_SOURCE_DIR}/include")
+
+if(JERRYX_CONTEXT)
+  set(DEFINES_MODULE "-DJERRYX_MODULE_HAVE_CONTEXT")
+endif()
+
+# Source directories
+file(GLOB SOURCE_MODULE *.c)
+
+add_library(${JERRYX_MODULE_NAME} STATIC ${SOURCE_MODULE})
+if(JERRYX_CONTEXT)
+  target_include_directories(${JERRYX_MODULE_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-ext/context/include")
+endif()
+target_include_directories(${JERRYX_MODULE_NAME} PRIVATE ${INCLUDE_MODULE})
+target_include_directories(${JERRYX_MODULE_NAME} INTERFACE ${INCLUDE_MODULE})
+target_include_directories(${JERRYX_MODULE_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-core")
+target_include_directories(${JERRYX_MODULE_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-core/jmem")
+target_include_directories(${JERRYX_MODULE_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-core/jrt")
+if(NOT "${DEFINES_MODULE}" STREQUAL "")
+  target_compile_definitions(${JERRYX_MODULE_NAME} PRIVATE "${DEFINES_MODULE}")
+endif()
+
+install(TARGETS ${JERRYX_MODULE_NAME} DESTINATION lib)
+install(DIRECTORY ${INCLUDE_MODULE}/ DESTINATION include/jerry-ext)
--- a/tests/unit-core/CMakeLists.txt
+++ b/tests/unit-core/CMakeLists.txt
@@ -14,6 +14,7 @@
 
 cmake_minimum_required (VERSION 2.8.12)
 project (unit-core C)
+set(INCLUDE_UNITTESTS ${CMAKE_CURRENT_LIST_DIR})
 
 if (NOT IS_ABSOLUTE ${FEATURE_PROFILE})
   set(FEATURE_PROFILE "${CMAKE_SOURCE_DIR}/jerry-core/profiles/${FEATURE_PROFILE}.profile")
@@ -43,3 +44,5 @@ foreach(SOURCE_UNIT_TEST_MAIN ${SOURCE_UNIT_TEST_MAIN_MODULES})
 
   add_dependencies(unittests-core ${TARGET_NAME})
 endforeach()
+
+add_subdirectory(jerry-ext)
--- a/tests/unit-core/jerry-ext/CMakeLists.txt
+++ b/tests/unit-core/jerry-ext/CMakeLists.txt
@@ -15,3 +15,7 @@
 if(JERRYX_CONTEXT)
   add_subdirectory(context)
 endif()
+
+if(JERRYX_MODULE)
+  add_subdirectory(module)
+endif()
--- a/tests/unit-core/jerry-ext/CMakeLists.txt
+++ b/tests/unit-core/jerry-ext/CMakeLists.txt
@@ -1,0 +1,17 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if(JERRYX_CONTEXT)
+  add_subdirectory(context)
+endif()
--- a/tests/unit-core/jerry-ext/context/CMakeLists.txt
+++ b/tests/unit-core/jerry-ext/context/CMakeLists.txt
@@ -1,0 +1,26 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+cmake_minimum_required (VERSION 2.8.12)
+set(JERRYX_CONTEXT_UNITTEST_NAME unit-test-jerry-context)
+
+project (${JERRYX_CONTEXT_UNITTEST_NAME} C)
+
+file(GLOB JERRYX_CONTEXT_UNIT_TEST_SOURCES *.c)
+
+add_executable(${JERRYX_CONTEXT_UNITTEST_NAME} ${JERRYX_CONTEXT_UNIT_TEST_SOURCES})
+set_property(TARGET ${JERRYX_CONTEXT_UNITTEST_NAME} PROPERTY LINK_FLAGS "${LINKER_FLAGS_COMMON}")
+target_link_libraries(${JERRYX_CONTEXT_UNITTEST_NAME} jerryx-context jerry-core jerry-port-default-minimal)
+target_include_directories(${JERRYX_CONTEXT_UNITTEST_NAME} PRIVATE ${INCLUDE_UNITTESTS})
+target_include_directories(${JERRYX_CONTEXT_UNITTEST_NAME} PRIVATE "${CMAKE_SOURCE_DIR}/jerry-ext/context/include")
--- a/tests/unit-core/jerry-ext/context/jerry-context-test.c
+++ b/tests/unit-core/jerry-ext/context/jerry-context-test.c
@@ -1,0 +1,70 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript.h"
+#include "test-common.h"
+#include "jerry-context.h"
+
+static char *static_slot1 = "static slot 1";
+static char *static_slot2 = "static slot 2";
+
+static int deinit_called_count = 0;
+
+static void *
+init_slot1 (void)
+{
+  return static_slot1;
+} /* init_slot1 */
+
+static void
+deinit_slot1 (void *slot)
+{
+  TEST_ASSERT (slot == static_slot1);
+  deinit_called_count++;
+} /* deinit_slot1 */
+
+JERRYX_CONTEXT_DEFINE_SLOT (slot1, init_slot1, deinit_slot1)
+
+static void *
+init_slot2 (void)
+{
+  return static_slot2;
+} /* init_slot2 */
+
+static void
+deinit_slot2 (void *slot)
+{
+  TEST_ASSERT (slot == static_slot2);
+  deinit_called_count++;
+} /* deinit_slot2 */
+
+JERRYX_CONTEXT_DEFINE_SLOT (slot2, init_slot2, deinit_slot2)
+
+int
+main (int argc, char **argv)
+{
+  (void) argc;
+  (void) argv;
+  TEST_INIT ();
+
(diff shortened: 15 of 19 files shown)


zherczeg:
CC @jiangzidong 

He plans to implement import and this is definitely related.

gabrielschulhof:
`jerry_module_load ()` could definitely be passed into whatever ends up handling `import` as a hook to external native modules, in a fashion similar to the function pointers passed into `jerry_init_with_user_context ()`.

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+  jerry_user_context_deinit_cb deinit_cb;
+} jerryx_context_slot_info;
+
+/* Context-free data */
+
+/* We do not allocate slots after a context has already been created. */
+bool before_first_context = true;
+static int next_index = -1;
+static jerryx_context_slot_info slot_info[JERRYX_CONTEXT_SLOTS] =
+{
```
jiangzidong: Maybe we should avoid using global variables directly, it will make jerry hard to port to multiple-thread cases. 
Those values are context/instance-specific, right?

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+/**
+ * Create a new context. This function can be passed to jerry_init_with_user_context () as its second parameter.
+ *
+ * @return the pointer to store as the user context.
+ */
+void *
+jerryx_context_init (void)
+{
+  size_t index = 0;
+  void **return_value = NULL;
```
jiangzidong: I will suggest we use `ret` like most source code in this project did.

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+
+  before_first_context = false;
+
+  return_value = (void **) jmem_heap_alloc_block_null_on_error (HEAP_BLOCK_SIZE);
+
+  if (return_value)
+  {
+    memset (return_value, 0, HEAP_BLOCK_SIZE);
+    for (index = 0; index < JERRYX_CONTEXT_SLOTS; index++)
```
jiangzidong: should assert an empty line before and after block code like `for`, `if`, `while`

tests/unit-core/jerry-ext/jerry-context/jerry-context-test.c
```diff
@@ -0,0 +1,68 @@
...
+static void *
+init_slot1 (void)
+{
+  return static_slot1;
+} /* init_slot1 */
+
+static void
+deinit_slot1 (void *slot)
+{
+  TEST_ASSERT (slot == static_slot1);
```
jiangzidong: We should also test whether the deinit function is executed.
```
static int deinit_called_count = 0;

in deinit_slot1 : deinit_called_count ++;
in deinit_slot2 : deinit_called_count ++;

after jerry_cleanup:  TEST_ASSERT (deinit_called_count == 2);
```

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+ * Request an index for per-context data.
+ *
+ * @return the slot in which the user data will be stored, or -1 on failure. The value can later be passed to
+ * jerryx_context_get_slot () to retrieve the per-context user data.
+ */
+int
+jerryx_context_request_slot (jerry_user_context_init_cb init_cb, /**< callback for creating a new context */
+                             jerry_user_context_deinit_cb deinit_cb) /**< callback for freeing context */
+{
+  int return_value = -1;
```
jiangzidong: What about changing "return_value" to "slot_index" here?

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,198 @@
...
+  {
+    link->next = (*root);
+  }
+  (*root) = link;
+} /* jerryx_module_link_insert */
+
+static jerryx_module_link_t *
+jerryx_module_link_find (jerryx_module_link_t *root,
+                         const jerry_char_t *name,
+                         jerryx_module_link_t **previous)
```
jiangzidong: I would set `jerryx_module_link_t` as double linked lists instead of using `previous` in function

LaszloLango:
> Maybe we should compile one jerry-ext lib instead of small ones. so that users only need to link one lib.

I highly agree. Creating multiple jerry-ext libs is a bad idea. It would be also hard to maintain and test the different combinations.

jerry-ext/jerry-context/jerry-context.c
gabrielschulhof: No. There's a comment above these lines saying these are context-independent. In fact, that's what `before_first_context` is for. If you create a context, you can no longer assign slots. This is a list of functions to call when you create a context, so the list must be constructed before you create your first context.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,198 @@
+#include "jmem.h"
```
gabrielschulhof: Yeah, I'm not sure about that either. How does one allocate memory?

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: Yeah, but that's one more pointer taking up space.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,198 @@
...
+      {
+        JERRYX_MODULE_LINK_RUNTIME_INIT (instance_p, module->link.name);
+        instance_p->result = jerry_acquire_value (return_value);
+        jerryx_module_link_insert ((jerryx_module_link_t *) instance_p, (jerryx_module_link_t **) instances_pp);
+      }
+    }
+    return return_value;
+  }
+
+  error_message_size = 1 + name_length + sizeof (not_found_prologue) - 1 + sizeof (not_found_epilogue) - 1;
```
gabrielschulhof: No, because `sizeof (not_found_prologue)` is not the same as `strlen (not_found_prologue)`. `sizeof` includes the `\0`:
```C
#include <stdio.h>
int main(int argc, char **argv) {
  char abc[] = "abc";
  printf("%d\n", sizeof(abc));
  return 0;
}
```
This will print 4, not 3.

gabrielschulhof:
I don't think there's a difference between one jerry-ext lib or multiple, because the different parts are controlled by flags anyway, so you have to test just as many combinations either way. It's only a matter of how many .a files will you have. Everything's linked statically anyway, so if you start with one .a file or twelve, what difference does that make?

gabrielschulhof:
Wait, I see a benefit. You don't have to have twelve arguments in your `target_link_libraries`. OK. I'll change it to one library controlled by multiple flags. Still need to test the combinations though.

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: I should also change this to not being recursive, because that's more stack frames.

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+#include "jerry-context.h"
+#include "jmem.h"
+
+#define HEAP_BLOCK_SIZE (JERRYX_CONTEXT_SLOTS * sizeof (void *))
+
+typedef struct
+{
+  jerry_user_context_init_cb init_cb;
+  jerry_user_context_deinit_cb deinit_cb;
+} jerryx_context_slot_info;
```
martijnthe: `jerryx_context_slot_info_t` (missing `_t`)

martijnthe: Hmm I just see now that we're missing the `_t` suffix on `jerry_user_context_init_cb` and `jerry_user_context_deinit_cb` as well (which is already in `master`).

Should probably be fixed (separately)?

jerry-ext/jerry-module/jerry-module.c
jiangzidong: I mean to compute `1 + 1 - 1 = -1` ahead

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string.h>
+#include "jerry-context.h"
+#include "jmem.h"
```
martijnthe: I don't think `jmem.h` and `jerry-context.h` are part of the public API.
My understanding is that `jerry-ext` should only use public API.

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+ */
+void *
+jerryx_context_init (void)
+{
+  size_t index = 0;
+  void **return_value = NULL;
+
+  before_first_context = false;
+
+  return_value = (void **) jmem_heap_alloc_block_null_on_error (HEAP_BLOCK_SIZE);
```
martijnthe: Does it make sense to add a `jerry_malloc()`/`jerry_free()` as public API?
And/or have a `jerry_port_malloc()`/`jerry_port_free()`?

jerry-ext/jerry-context/jerry-context.c
jiangzidong: But user may create multiple-context for different jerry instance, those context should be able to have totally different user_contexts and slots. 
In your way, all the context must have same slot, right?

tests/unit-core/jerry-ext/jerry-context/jerry-context-test.c
```diff
@@ -0,0 +1,68 @@
...
+
+  int slot1 = jerryx_context_request_slot (init_slot1, deinit_slot1);
+  int slot2 = jerryx_context_request_slot (init_slot2, deinit_slot2);
+  int slot3 = jerryx_context_request_slot (init_slot2, deinit_slot2);
+
+  TEST_ASSERT (slot1 == 0);
+  TEST_ASSERT (slot2 == 1);
+  TEST_ASSERT (slot3 == -1);
+
+  jerry_init_with_user_context (JERRY_INIT_EMPTY, jerryx_context_init, jerryx_context_deinit);
```
martijnthe: Is it possible at this moment to test this with multiple Jerry instances? When https://github.com/jerryscript-project/jerryscript/pull/1778 is merged this should be easily possible, but not sure if it is possible easily right now.

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,139 @@
...
+  if (return_value)
+  {
+    memset (return_value, 0, HEAP_BLOCK_SIZE);
+    for (index = 0; index < JERRYX_CONTEXT_SLOTS; index++)
+    {
+      if (slot_info[index].init_cb)
+      {
+        return_value[index] = slot_info[index].init_cb ();
+      }
+    }
```
martijnthe: `return_value == NULL` not being handled / failing silently?
Personally, I'd rather trip an assert in this case.

gabrielschulhof:
There seem to be two problems:

0. The extensions need to allocate/free stuff. There needs to be an API for that, otherwise we introduce all kinds of portability issues which will prevent jerryscript from being self-contained.
0. Making jerry-ext one target with multiple conditionally included parts is awkward because there is a case where the library is requested (say, a new flag `JERRY_EXT` is set to `ON`), but none of its pieces are requested (that is, both `JERRYX_MODULE` and `JERRYX_CONTEXT` are left `OFF`). What do you build then?

    There's no big difference between having one library and having multiple, and in the multiple case, the individual flags are clear. If you turn on `JERRYX_MODULE` you will have the module extension, etc. The only inconvenience is that in your target you have to say
    ```
    target_link_libraries(${MY_TARGET} jerry-core jerry-context jerry-module jerry-blah ...)
    ```
    instead of simply
    ```
    target_link_libraries(${MY_TARGET} jerry-core jerry-ext)
    ```

jerry-ext/jerry-context/include/jerry-context.h
```diff
@@ -0,0 +1,43 @@
...
+void jerryx_context_deinit (void *user_context_p);
+
+/**
+ * Function to request an index for per-context data
+ */
+int jerryx_context_request_slot (jerry_user_context_init_cb init_cb,
+                                 jerry_user_context_deinit_cb deinit_cb);
+
+
+void *jerryx_context_get_slot (int slot);
```
martijnthe: Maybe just `jerryx_context_get`? In my mind the ID/int is the "slot".

gabrielschulhof: Sure.

gabrielschulhof: Although ... `get` is so generic.

gabrielschulhof: Well, whatever. Let's go with it. People will get the idea.

jerry-ext/jerry-context/jerry-context.c
gabrielschulhof: Yes. Each context has the same number of slots. The functions responsible for maintaining each slot are given before the first slot is created.

gabrielschulhof: The structure of all the user contexts is thus defined before any contexts are created.

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: Oh, of course. D'oh :)

gabrielschulhof: OTOH it probably gets optimized out, but whatever. I'll change it.

tests/unit-core/jerry-ext/jerry-context/jerry-context-test.c
gabrielschulhof: Actually, perhaps that PR should test to make sure `jerry_init_with_user_context ()`/`jerry_get_user_context ()` works properly, right?

gabrielschulhof: This PR assumes those two work as expected.

jerry-ext/jerry-context/include/jerry-context.h
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Function to pass to jerry_init_with_user_context() as its third parameter
+ */
+void jerryx_context_deinit (void *user_context_p);
+
+/**
+ * Function to request an index for per-context data
+ */
+int jerryx_context_request_slot (jerry_user_context_init_cb init_cb,
+                                 jerry_user_context_deinit_cb deinit_cb);
```
martijnthe: As an alternative to requesting/getting "slots" at runtime, I think it's also possible to do this at compile/build time *without having a single file that defines/lists the slot IDs*:

By defining a dedicated linker section, i.e. `.jerryx_context_slots`, into which to put each `static jerryx_context_slot_info_t`. The linker script can add a terminator at the end of the section with `NULL` values. The linker script should also emit a symbols with the base address and also the size of the section, so that the "slot" indices and also the total number of slots can be calculated at build time. Probably `jerry-ext/...` should include a linker script snippet defining the section + symbols, so it can be `INCLUDE`d into the .ld script by the user of the util.

Doing this vs runtime requesting/getting would save code space and probably be a bit faster as well.

Now, admittedly, this solution is probably a bit more complicated to use, because it involves having to add a couple things to the target's linker script.

So, I think having a simple implementation (your one) has value.

Ideally, I would like to be able to swap out your implementation with the linker script based one **without having to change the code that uses the util**. But I think the current API doesn't let me do this easily.

Runtime registration implementation:
```
static int my_slot = -1;
// register at boot:
{
  my_slot = jerryx_context_request_slot (my_init, my_deinit);
}
// get context:
{
  void *ctx = jerryx_context_get_slot (my_slot);
}
```

Compile-time registration implementation:
```
static const jerryx_context_slot_info_t my_context_cbs __attribute__((section(".jerryx_context_slots"))) = {
  .init = my_init,
  .deinit = my_deinit,
};

extern jerryx_context_slot_info_t *jerryx_context_slots_first; // defined in linker script, todo: move to jerry-context.h
// note const! -- when const, the slot value in jerryx_context_get_slot() can be an immediate vs having to load it.
static const int my_slot =  &my_context_cbs - jerryx_context_slots_first; // todo: put macro in jerry-context.h

// get context -- still the same:
{
  void *ctx = jerryx_context_get_slot (my_slot);
}

// the implementation of jerryx_context_get_slot() can be simpler,
// no checking of the slot arg is needed:
inline void *jerryx_context_get_slot(int slot) {
  return ((void **) jerry_get_user_context ())[slot];
}
```

Thoughts? Ideas on how we could have one interface that enables both implementations?
@zherczeg @akosthekiss Thoughts on the linker script trickery I'm suggesting here?

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,82 @@
+#ifndef JERRYX_MODULE_H
```
martijnthe: Missing the license comment.

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,82 @@
...
+ */
+jerry_value_t jerryx_module_load (const jerry_char_t *name);
+
+#define JERRYX_MODULE(name, init)                         \
+EXTERN_C_START                                            \
+static jerryx_module_t _module =                          \
+{                                                         \
+  JERRYX_MODULE_LINK_STATIC_INIT((jerry_char_t *) #name), \
+  JERRYX_MODULE_VERSION,                                  \
+  (init)                                                  \
```
martijnthe: Add field names -- makes it easier to read the code:
```
{
  .link = ...,
  .version = ...,
...
```

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,82 @@
...
+#define JERRYX_MODULE_VERSION 1
+
+typedef struct jerryx_module_link
+{
+  jerry_char_t *name;
+  struct jerryx_module_link *next;
+} jerryx_module_header_t;
+
+#define JERRYX_MODULE_LINK_STATIC_INIT(name) \
+  {name, NULL}
```
martijnthe: Same here, use `.field_name = ...`

tests/unit-core/jerry-ext/jerry-context/jerry-context-test.c
jiangzidong: `jerry_init_with_user_context ()/jerry_get_user_context ()` works well in #1778 

I changed the unit-test-user-context, by building it with `./tools/build.py --unittest --jerry-libc=off --external-context=on`, then run `./build/bin/unit-test-user-context`, it passed

I still consider where to put the test case.
``` C
#include "config.h"
#include "jerryscript.h"
#include "test-common.h"
#include "jerryscript-port.h"
#include "jerryscript-port-default.h"


#define JERRY_EXTERNAL_HEAP_SIZE (1024)
#define JERRY_EXTERNAL_LCACHE_ROW_COUNT 4
#define JERRY_EXTERNAL_LCACHE_ROW_LENGTH 2
#define JERRY_EXTERNAL_HEAP_ALIGN 8

static jerry_instance_t *instance1_p;
static jerry_instance_t *instance2_p;

static const char *identifying_string1 = "identifying string 1";
static bool user_context1_new_was_called = false;
static bool user_context1_free_was_called = false;

static const char *identifying_string2 = "identifying string 2";
static bool user_context2_new_was_called = false;
static bool user_context2_free_was_called = false;

static void *
user_context1_new (void)
{
  user_context1_new_was_called = true;
  return (void *) identifying_string1;
} /* user_context1_new */

static void
user_context1_free (void *user_context_p)
{
  user_context1_free_was_called = true;
  TEST_ASSERT (((const char *) user_context_p) == identifying_string1);
} /* user_context1_free */

static void *
user_context2_new (void)
{
  user_context2_new_was_called = true;
  return (void *) identifying_string2;
} /* user_context2_new */

static void
user_context2_free (void *user_context_p)
{
  user_context2_free_was_called = true;
  TEST_ASSERT (((const char *) user_context_p) == identifying_string2);
} /* user_context2_free */

int
main (void)
{
  TEST_INIT ();

  size_t context_size = jerry_get_context_size ();
  size_t lcache_size = jerry_get_lcache_size (JERRY_EXTERNAL_LCACHE_ROW_COUNT, JERRY_EXTERNAL_LCACHE_ROW_LENGTH);
  uintptr_t alignment_mask = ~(uintptr_t) (JERRY_EXTERNAL_HEAP_ALIGN - 1);
  /* create instance 1 */
  void *context1_p = malloc (context_size);
  memset (context1_p, 0, context_size);
  void *lcache1_p = malloc (lcache_size);
  void *heap1_p = malloc (JERRY_EXTERNAL_HEAP_SIZE + JERRY_EXTERNAL_HEAP_ALIGN - 1);
  instance1_p = (jerry_instance_t *) malloc (sizeof (jerry_instance_t));
  instance1_p->context_p = context1_p;
  instance1_p->lcache_p = lcache1_p;
  instance1_p->heap_p = (void *) (((uintptr_t) heap1_p + JERRY_EXTERNAL_HEAP_ALIGN - 1) & alignment_mask);
  instance1_p->heap_size = JERRY_EXTERNAL_HEAP_SIZE;
  instance1_p->lcache_row_count = JERRY_EXTERNAL_LCACHE_ROW_COUNT;
  instance1_p->lcache_row_length = JERRY_EXTERNAL_LCACHE_ROW_LENGTH;
  /* create instance 2 */
  void *context2_p = malloc (context_size);
  memset (context2_p, 0, context_size);
  void *lcache2_p = malloc (lcache_size);
  void *heap2_p = malloc (JERRY_EXTERNAL_HEAP_SIZE + JERRY_EXTERNAL_HEAP_ALIGN - 1);
  instance2_p = (jerry_instance_t *) malloc (sizeof (jerry_instance_t));
  instance2_p->context_p = context2_p;
  instance2_p->lcache_p = lcache2_p;
  instance2_p->heap_p = (void *) (((uintptr_t) heap2_p + JERRY_EXTERNAL_HEAP_ALIGN - 1) & alignment_mask);
  instance2_p->heap_size = JERRY_EXTERNAL_HEAP_SIZE;
  instance2_p->lcache_row_count = JERRY_EXTERNAL_LCACHE_ROW_COUNT;
  instance2_p->lcache_row_length = JERRY_EXTERNAL_LCACHE_ROW_LENGTH;
  /* switch to instance 1 */
  jerry_port_set_instance (instance1_p);
  jerry_init_with_user_context (JERRY_INIT_EMPTY, user_context1_new, user_context1_free);
  TEST_ASSERT ((((const char *)(jerry_get_user_context ()))) == identifying_string1);
  /* switch to instance 2 */
  jerry_port_set_instance (instance2_p);
  jerry_init_with_user_context (JERRY_INIT_EMPTY, user_context2_new, user_context2_free);
  TEST_ASSERT ((((const char *)(jerry_get_user_context ()))) == identifying_string2);
  /* switch to instance 1 */
  jerry_port_set_instance (instance1_p);
  jerry_cleanup ();
  TEST_ASSERT (user_context1_new_was_called);
  TEST_ASSERT (user_context1_free_was_called);
  /* switch to instance 2 */
  jerry_port_set_instance (instance2_p);
  jerry_cleanup ();
  TEST_ASSERT (user_context2_new_was_called);
  TEST_ASSERT (user_context2_free_was_called);

  return 0;
} /* main */

```

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,192 @@
...
+jerryx_module_load (const jerry_char_t *name)
+{
+  jerry_value_t return_value = 0;
+  size_t error_message_size = 0, name_length = strlen ((char *) name);
+  char *error_message = NULL;
+  jerryx_module_t *module = NULL;
+  jerryx_module_instance_t **instances_pp = JERRYX_MODULE_CONTEXT;
+  jerryx_instance_t *instance_p = NULL;
+
+  if (instances_pp)
```
martijnthe: When would this be `NULL`?

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,140 @@
...
+void *
+jerryx_context_get (int slot) /**< the slot from which to retrieve the data */
+{
+  void *return_value = NULL;
+  void **slots = (void **) jerry_get_user_context ();
+
+  if (slots && slot >= 0 && slot < JERRYX_CONTEXT_SLOTS)
+  {
+    return_value = slots[slot];
+  }
```
martijnthe: Probably OK to `jerry_port_fatal(...)` in this case, as opposed to having to handle a `NULL` result in all the call sites?

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  error_message_size = name_length + sizeof (not_found_prologue) + sizeof (not_found_epilogue) - 1;
+  error_message = jmem_heap_alloc_block_null_on_error (error_message_size);
+  if (error_message)
+  {
+    memcpy (error_message, not_found_prologue, sizeof (not_found_prologue) - 1);
+    memcpy (&error_message[sizeof (not_found_prologue) - 1], ((char *) name), name_length);
+    memcpy (&error_message[sizeof (not_found_prologue) - 1 + name_length], not_found_epilogue,
+            sizeof (not_found_epilogue) - 1);
```
martijnthe: Maybe add `strcat` to `jerry-libc`?

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,82 @@
...
+
+#define JERRYX_MODULE_LINK_FLOAT(link) \
+  ((jerryx_module_header_t *) link)->next = NULL
```
martijnthe: `..._END`? `..._FLOAT` reminds me of more of `float` than ending a list :)

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: If, upon initialization of the context, it fails to allocate memory for the root instance pointer. Then the module cache doesn't work, and the `jerry_value_t (*init)(void)` callback for the module has to be called every time.

jerry-ext/jerry-context/jerry-context.c
gabrielschulhof: In the case of modules, `NULL` is not fatal. It just means that the result of the module initialization cannot be reused if somebody loads the module multiple times.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,192 @@
...
+{
+  jerryx_module_header_t link;
+  jerry_value_t result;
+} jerryx_instance_t;
+
+static const char not_found_prologue[] = "Module '";
+static const char not_found_epilogue[] = "' not found";
+
+jerry_value_t
+jerryx_module_load (const jerry_char_t *name)
```
martijnthe: `..._resolve`?

gabrielschulhof: I'm reluctant to call it that, because `_resolve` suggests that it's already there, you just need to figure out which one `name` is referring to - whereas in this case, you're allocating potentially quite a bit. You're building up a structure in memory and passing it back into JS. I'd rather call that `load`.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,192 @@
...
+                                                                 jerryx_module_context_deinit);
+} /* JERRYX_C_CTOR */
+#define JERRYX_MODULE_CONTEXT \
+  (jerryx_context_get (jerryx_module_user_context_slot))
+#else /* !JERRYX_MODULE_HAVE_CONTEXT */
+#define JERRYX_MODULE_CONTEXT \
+  (jerry_get_user_context ())
+#endif /* JERRYX_MODULE_HAVE_CONTEXT */
+
+/* Context-free list of modules */
```
martijnthe: FYI, I first misunderstood this comment: I thought this was a list of stateless ('context-free') modules that do not require a `init()`.

gabrielschulhof: OK. I'll elaborate.

jerry-ext/jerry-module/jerry-module.c
martijnthe: OK, yeah makes sense.

jerry-ext/jerry-context/jerry-context.c
```diff
@@ -0,0 +1,140 @@
...
+void *
+jerryx_context_init (void)
+{
+  size_t index = 0;
+  void **ret = NULL;
+
+  before_first_context = false;
+
+  ret = (void **) jmem_heap_alloc_block_null_on_error (HEAP_BLOCK_SIZE);
+
```
martijnthe: This allocation happens really early on and is probably pretty small.
If there's no memory for it, I would `jerry_port_fatal(...)` instead of trying to keep trucking along and as a result cause every caller of `jerryx_context_get(...)` to have to worry about handling a `NULL` return value.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,209 @@
...
+{
+  jerryx_module_instance_t *instance = (jerryx_module_instance_t *) link;
+  jerry_release_value (instance->result);
+  jmem_heap_free_block (instance, sizeof (jerryx_module_instance_t));
+} /* jerryx_module_free_instance */
+
+void *
+jerryx_module_context_init (void)
+{
+  void *return_value = jmem_heap_alloc_block_null_on_error (sizeof (jerryx_module_instance_t *));
```
martijnthe: Check for `NULL`? Again, I'd `jerry_port_fatal()` here in case it couldn't allocate these couple bytes.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,209 @@
...
+  memset (return_value, 0, sizeof (jerryx_module_instance_t *));
+  return return_value;
+} /* jerryx_module_context_init */
+
+void
+jerryx_module_context_deinit (void *context_p)
+{
+  if (context_p)
+  {
```
martijnthe: Shouldn't you also traverse the linked list and clean up each link?

martijnthe: Nevermind, I see now that's done already.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,209 @@
...
+    next = root->next;
+    free_link (root);
+    root = next;
+  }
+} /* jerryx_module_link_free */
+
+typedef struct
+{
+  jerryx_module_header_t link;
+  jerry_value_t result;
```
martijnthe: `result` => `export`?

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,209 @@
...
+void
+jerryx_module_unregister (jerryx_module_t *module_p)
+{
+  jerryx_module_link_remove (module_p->link.name, ((jerryx_module_header_t **) (&modules)));
+} /* jerryx_module_unregister */
+
+typedef struct
+{
+  jerryx_module_header_t link;
+  jerry_value_t result;
```
martijnthe: `result` => `export`?

martijnthe: I'm a bit confused about the difference between `jerryx_instance_t` vs `jerryx_module_instance_t`.

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,209 @@
...
+
+  if (module)
+  {
+    return_value = module->init ();
+    if (instances_pp)
+    {
+      instance_p = (jerryx_instance_t *) jmem_heap_alloc_block_null_on_error (sizeof (jerryx_instance_t));
+      if (instance_p)
+      {
+        JERRYX_MODULE_HEADER_RUNTIME_INIT (instance_p, module->link.name);
```
martijnthe: Personally, I'm not a huge fan of this macro.
It doesn't do all that much and the name also isn't that great.
I'd rather just put the C code here instead.

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+static void fn(void) __attribute__((constructor)); \
+static void fn(void)
+
+#define JERRYX_MODULE_VERSION 1
+
+typedef struct jerryx_module_link
+{
+  jerry_char_t *name;
+  struct jerryx_module_link *next;
+} jerryx_module_header_t;
```
martijnthe: I'd try to move the link list implementation details out of the .h and into the .c, to make it easier to add alternative implementations in the future.

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+jerry_value_t jerryx_module_load (const jerry_char_t *name);
+
+#define JERRYX_MODULE(name, init_cb)                                 \
+EXTERN_C_START                                                       \
+static jerryx_module_t _module =                                     \
+{                                                                    \
+  .link = JERRYX_MODULE_HEADER_STATIC_INIT ((jerry_char_t *) #name), \
+  .version = JERRYX_MODULE_VERSION,                                  \
+  .init = (init_cb)                                                  \
```
martijnthe: No `deinit`?

jerry-ext/jerry-module/jerry-module.c
```diff
@@ -0,0 +1,209 @@
...
+{
+  jerryx_module_header_t link;
+  jerry_value_t result;
+} jerryx_module_instance_t;
+
+static void
+jerryx_module_free_instance (jerryx_module_header_t *link)
+{
+  jerryx_module_instance_t *instance = (jerryx_module_instance_t *) link;
+  jerry_release_value (instance->result);
```
martijnthe: I think we should also call a module `deinit` here.

jerry-ext/jerry-module/jerry-module.c
martijnthe: >  _resolve suggests that it's already there

Actually, I'm not sure if I agree. Or at least, `resolve` also seems to imply any necessary creation in these descriptions:

https://nodejs.org/api/modules.html#modules_all_together
http://www.ecma-international.org/ecma-262/6.0/#sec-resolveexport
http://www.ecma-international.org/ecma-262/6.0/#sec-hostresolveimportedmodule

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Function to unregister a module
+ */
+void jerryx_module_unregister (jerryx_module_t *module);
+
+/**
+ * Function to load a copy of a module into the current context
+ */
+jerry_value_t jerryx_module_load (const jerry_char_t *name);
```
martijnthe: What are your thoughts on how you would extend this to be able to implement `require('some_file.js')`?
Ideally, you'd piggy-back on the module book-keeping that you have built, to track module instances created in other ways (i.e. loading a module from source). Maybe have a user callback inside `jerryx_module_load`..?

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: I do, in the line below this brace :)

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: The reason I sought to encapsulate the initialization of these linked list items is for the case that, later on, we switch to a different data structure, like a tree. Then we don't have to scour the code to make sure all the data structure -related data is up-to-date.

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: I thought about that too, but if we want to be able to link global static structures like the module definitions into a data structure like a list or a tree without performing additional runtime allocation we need to expose the data structure's metadata.

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: Hmmm ... that's a good point, however, the entire result of `init_cb()` is a single `jerry_value_t`. If the module needs per-context info, it can use jerry-context and request a slot.

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: ... but what to pass it? The module has no place to store per-context info except in jerry-context, but that already has an API for deinit. I don't want to duplicate that API.

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: That's something to think about.

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: Hmmm, according to the last link not having a cache is not an option because, without a cache, you cannot have idempotency.

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: There is no difference. I accidentally redefined it.

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: Fixed.

tests/unit-core/jerry-ext/jerry-context/jerry-context-test.c
gabrielschulhof: Fixed.

jerry-ext/jerry-context/include/jerry-context.h
gabrielschulhof: The standard linker script used by vanilla gcc [does indeed declare symbols](http://stackoverflow.com/questions/16552710/how-do-you-get-the-start-and-end-addresses-of-a-custom-elf-section-in-c-gcc) `_start_<CUSTOM_SECTION_NAME>` and `_stop_<CUSTOM_SECTION_NAME>` when it encounters `__attribute__((section("<CUSTOM_SECTION_NAME>")))`. It'd be nice to have that feature in jerryscript as well.

gabrielschulhof: So, for example, the following works as expected with vanilla gcc both in one file, as shown for simplicity below, or in multiple files, as indicated by the comments:
```C
#include <stdio.h>

/* module.h */
typedef void (*module_init)(void);

/* module1.c */
static void module1_init(void) {
  printf("module1_init\n");
}
module_init __module1_init __attribute__((section("__module_init"))) = module1_init;

/* module2.c */
static void module2_init(void) {
  printf("module2_init\n");
}
module_init __module2_init __attribute__((section("__module_init"))) = module2_init;

/* main.c */
extern module_init __start___module_init;
extern module_init __stop___module_init;

int main(int argc, char **argv) {
  module_init *iter;

  for (iter = &__start___module_init; iter < &__stop___module_init; ++iter) {
    (*iter)();
  }
  return 0;
}
```
Unfortunately, when building this code within jerryscript, the two symbols `__start___module_init` and `__stop___module_init` are missing - so that's what we'd need our linker script to produce.

gabrielschulhof: @akosthekiss you did this for constructors, for which the name of the section is fixed. Is it possible to add these `_start_*` and `_stop_*` symbols for arbitrary section names as well?

martijnthe: Maybe I'm missing something, but you can `PROVIDE` these start/stop symbols in the linker script of your project?
https://sourceware.org/binutils/docs/ld/PROVIDE.html#PROVIDE
That's what I've been doing in the past at least.

jerry-ext/jerry-module/include/jerry-module.h
martijnthe: Ah, I see, ok fair enough!

jerry-ext/jerry-module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+#define EXTERN_C_END
+#endif
+
+#define JERRYX_C_CTOR(fn)                          \
+static void fn(void) __attribute__((constructor)); \
+static void fn(void)
+
+#define JERRYX_MODULE_VERSION 1
+
+typedef struct jerryx_module_link
```
martijnthe: Why `jerryx_module_link` for the tag and `jerryx_module_header_t` for the type?
I think the style of this project for them to be the same.

gabrielschulhof: Ooops ... forgot to rename it ...

jerry-ext/jerry-module/include/jerry-module.h
martijnthe: Oh, right, the malloc/free of the native context info is taken care of by the `jerryx_context_...` layer.

This callback is called as result of `jerryx_module_resolve()`. Maybe name the callback `onresolve` or something, to avoid confusion with the context info allocation/freeing concern?

I just realized something else: you can get a weird context deinit ordering, because there's little control over ordering when using `__attribute__((constructor))`. I.e. what if a module calls  `jerryx_context_request_slot()` before `jerryx_module.c` calls calls  `jerryx_context_request_slot()`? Then the module's deinit would be called before the `jerryx_module_context_deinit` gets called.

Does that matter?

jerry-ext/jerry-module/jerry-module.c
martijnthe: Yes you're right.
Already discussed here: https://github.com/jerryscript-project/jerryscript/pull/1784#pullrequestreview-35334327

jerry-ext/jerry-module/jerry-module.c
martijnthe: Why is having a cache a problem?

jerry-ext/jerry-context/include/jerry-context.h
martijnthe: Some more info on the "magic" `__start_...` and `__stop_...` symbols that gnu ld emits:
http://www.airs.com/blog/archives/56

There are 2 underscores before start/stop and **only 1 after it**.
Maybe that's why your code didn't link properly.

gabrielschulhof: Ideally the extension would provide this functionality without asking the project maintainer to modify their linker script. Heck, I didn't even know about linker scripts until this came up :)

I'd take the same approach as with constructors, where I'd provide a macro that'd let you write, like
```C
JERRYX_CONTEXT_SLOT(my_module_context_init, my_module_context_deinit);
```
That would declare the variables into the appropriate section - and it would Just Work™.

As for the underscores, I named my section `__module_init` so the resulting symbols should've been `__start_`+`__module_init` etc. I think the underscores are right but the symbols are not inserted by default.

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: So, you mean that one or more of the module's functions that was exposed via `onresolve()` may be called after the module's context destructor was already called, right?

I believe it's possible for this to happen only if one module gets a hold of a `jerry_value_t` that refers to a native function in a second module and uses `jerry_call_function` to call that function during its own deinit. This is case is pretty corner though.

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: This is another thing that could be solved using custom sections. Functions that take a string and return a `jerry_value_t`.

jerry-ext/jerry-context/include/jerry-context.h
zherczeg: I don't think it is a good idea that people need specialized linker scripts for using this extension.

jerry-ext/jerry-context/CMakeLists.txt
```diff
@@ -0,0 +1,46 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
```
zherczeg: Isn't the file name should be  jerry-ext/jerryx-context/CMakeLists.txt ? Or simply jerry-ext/context/CMakeLists.txt ?

jerry-ext/jerry-context/include/jerry-context.h
zherczeg: I remember the Linux kernel has hardware lists, but I don't remember how they work. Perhaps we could use a similar idea.

jerry-ext/jerry-context/jerry-context.c
zherczeg: No. That would be too much restriction. To me extensions are core part of the project, and those who write extensions understand jerryscript internals (and fix their extensions after breaking changes). The dependency is one way, so extensions can use core functions, but core functions cannot use extensions.

User applications should only use public jerry and jerry extension api.

tests/unit-core/jerry-ext/jerry-module/jerry-module-test.c
```diff
@@ -0,0 +1,80 @@
...
+  (void) argv;
+  jerry_value_t js_global = 0, js_function = 0, js_property_name = 0, js_eval_result = 0;
+
+  TEST_INIT ();
+
+#ifdef JERRYX_MODULE_HAVE_CONTEXT
+  jerry_init_with_user_context (JERRY_INIT_EMPTY, jerryx_context_init, jerryx_context_deinit);
+#else /* !JERRYX_MODULE_HAVE_CONTEXT */
+  jerry_init_with_user_context (JERRY_INIT_EMPTY, jerryx_module_context_init, jerryx_module_context_deinit);
+#endif /* JERRYX_MODULE_HAVE_CONTEXT */
```
zherczeg: This is a bit confusing for me. If you use modules, you implicitly use context pointers?

jerry-ext/jerry-context/include/jerry-context.h
martijnthe: I agree it's not user friendly.
All I'm asking for is to design the API such that build-time / static registration of the contexts/modules is supported too.
I think @gabrielschulhof 's idea of using user-definable macros could work to accomplish this.

jerry-ext/jerry-context/CMakeLists.txt
gabrielschulhof: OK. I'll move the extensions to `jerry-ext/context` and `jerry-ext/module`.

tests/unit-core/jerry-ext/jerry-module/jerry-module-test.c
gabrielschulhof: Not necessarily. The module extension only uses a single slot, because it stores the `jerry_value_t` results of loading modules in a linked list for which the root is a pointer stored in the user context.

If modules are the only extension used by the project, there's only one slot in use so there's no need for the context extension, and the modules extension can be used directly with `jerry_init_with_user_context()/jerry_get_user_context()`.

However, if the project wants modules but needs the context for other things as well, it needs more slots, so it needs to also add the context extension, and the module extension needs to integrate with the context extension by requesting a slot at boot time, instead of expecting the project to pass `jerry_module_context_init ()` and `jerry_module_context_deinit ()` to `jerry_init_with_user_context ()`.

This unit test ensures both scenarios are working correctly. It should work both if you build it with `-DJERRYX_CONTEXT=ON` and if you build it with `-DJERRYX_CONTEXT=OFF` (the default).

jerry-ext/jerry-module/include/jerry-module.h
martijnthe: Not sure if I follow. Can you elaborate?

jerry-ext/jerry-context/jerry-context.c
martijnthe: OK, understood. Works for me.

jerry-ext/jerry-context/include/jerry-context.h
martijnthe: @zherczeg could you try to find info about / examples w.r.t. the Linux hardware lists? I googled a bit but don't think I found what you are talking about. I'd like to learn more... thanks!

jerry-ext/jerry-module/jerry-module.c
gabrielschulhof: It's not a problem. Quite the contrary: The requirement to have a cache (because otherwise you cannot have idempotency) supports your argument to `jerry_port_fatal ()` if we fail to allocate the memory for the module instance. That's why I switched to `jmem_heap_alloc_block ()`.

jerry-ext/module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+ * Function to unregister a module
+ */
+void jerryx_module_unregister (jerryx_module_t *module_p);
+
+/**
+ * Function to load a copy of a module into the current context
+ */
+jerry_value_t jerryx_module_resolve (const jerry_char_t *name);
+
+#define JERRYX_MODULE(name, init_cb)                                 \
```
martijnthe: `init_cb` => `onresolve_cb`

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof:
```C
#define JERRYX_MODULE_RESOLVER(name) \
static jerry_value_t name(const char *module_name) __attribute__((section("__module_resolvers")));
static jerry_value_t name(const char *module_name)
```
and then, in your project, or in some extension, you can do
```C
JERRYX_MODULE_RESOLVER(load_js_file_from_disk) {
  jerry_value_t result = 0;
  FILE *fp = fopen(module_name, "r");
  ...
  return result;
}
```
and in `jerry-ext/module/jerry-module.c` we can add to `jerryx_module_resolve ()`:
```C
typedef jerry_value_t (*module_resolver)(const char *);
...
  module_resolver *iter;
  for (iter = (module_resolver *)__start___module_resolvers; iter < (module_resolver *)__stop___module_resolvers; iter++) {
    result = iter(name);
    if (result) {
      /* store result in the linked list and return it */
    }
  }
```

martijnthe: Excellent, works for me.

I'm guessing @zherczeg would like an easier-to-use (runtime) registration API alongside this.

jerry-ext/jerry-context/include/jerry-context.h
zherczeg: A random driver:

https://github.com/torvalds/linux/blob/master/drivers/dma/coh901318.c#L2781

    subsys_initcall(coh901318_init);
    module_exit(coh901318_exit);

I think there is a module_init as well. I don't know how these macros work, but they make it very easy to register fixed hardware modules (e.g. for SOC CPUs).

zherczeg: Migh help:

http://lxr.free-electrons.com/source/include/linux/module.h#L129

martijnthe: Thanks @zherczeg .
They also use special linker sections:

`subsys_initcall` in `subsys_initcall(coh901318_init);` maps to this define:
http://lxr.free-electrons.com/source/include/linux/init.h?v=3.14#L176

martijnthe: Here's the code where the special section is actually traversed at runtime:
http://lxr.free-electrons.com/source/init/main.c#L885

zherczeg: Ok, they use a similar technique. We can assume this is the best choice then. Writing linker scripts is not trivial, so this should be an optional feature, but we can go this way.

martijnthe: I noticed Linux used the C preprocessor to process linker scripts, so they can use macros / templates to "import" Linux-specific/common pieces that define certain sections and also to avoid duplication between ld scripts for different targets.

FWIW, GNU ld also has an `INCLUDE` directive: https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/4/html/Using_ld_the_GNU_Linker/simple-commands.html
which might be sufficient for our needs (I don't think we need macros).

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: Yeah, but we need support for these `__start_<section-name>` and `__stop_<secion-name>` variables. Do we have some linker script somewhere or do we simply tell ld not to use the linker script it normally uses?

jerry-ext/jerry-context/include/jerry-context.h
gabrielschulhof: There's one thing I don't understand. We also use GNU ld, just like the kernel. Do we have a LDFLAG where we're telling it *not* to add the `__start_` and `__stop_` symbols?

gabrielschulhof: I ran
```sh
./tools/build.py --unittests \
  --cmake-param=-DJERRYX_CONTEXT=ON \
  --cmake-param=-DJERRYX_CONTEXT_SLOTS=2 \
  --cmake-param=-DJERRYX_MODULE=ON \
  --cmake-param=-DFEATURE_INIT_FINI=ON \
  --debug -v -j 1 \
  --linker-flag=--verbose \
  --compile-flag=-Wl,-verbose
```
and this outputs the linker script. It does not differ from the plain LD linker script. So, I think one of our flags might be stripping these symbols.

jerry-ext/jerry-module/include/jerry-module.h
martijnthe: @gabrielschulhof what platform + compiler are you using?

I just did a quick test on Linux (Ubuntu 14.0.1) and gcc 4.8.4 (both "very" old).

In `main-unix.c` I added in `read_file` (but it doesn't really matter to what function you add it):

```
static uint8_t my_section_test[16] __attribute__((section("my_section"))) __attribute__((used));
extern int __start_my_section;
extern int __stop_my_section;
printf("my_section starts at: %p", &__start_my_section);
printf("my_section ends at: %p", &__stop_my_section);
```

Then build and check `nm`:
```
$./tools/build.py --debug
....
$ nm build/bin/jerry | grep my_section
00000000006d0020 d my_section_test.1742.2506.6498
00000000006d0020 D __start_my_section
00000000006d0030 D __stop_my_section
```
So, for me the "automagic" `__start_...` and `__stop_...` pseudo-symbols are added by GNU ld.
I did not have to add a custom linker script for Linux.

I also tried on macOS (10.12.4) with **gcc 6.3.0.1** and also Apple-clang (the one that ships with Xcode, 8.1.0 (clang-802.0.42)). The section attribute seems to be supported, but I had to write `...((section("__text,my_section")))`. But no "automagic" start/stop symbols unfortunately.
It seems custom linker scripts aren't supported on macOS.
Not a show-stopper in my opinion: when building for macOS, you probably don't care about code size :D and would pick the dynamic/runtime registration alternative.

Bottom line for me: if it works on the Linux that JerryScript CI uses, it's probably good enough to be able to add it to the project. That way we can make sure it doesn't break over time. Non-Linux (embedded) targets potentially need custom linker script foo, but IMO, we can leave that up to the user. And for those targets one can take the route of runtime/dynamic registration as well.

jerry-ext/jerry-context/include/jerry-context.h
martijnthe: See my comment here: https://github.com/jerryscript-project/jerryscript/pull/1784#discussion_r114297960

jerry-ext/jerry-module/include/jerry-module.h
gabrielschulhof: I have
```
gcc (GCC) 6.3.1 20161221 (Red Hat 6.3.1-1)
Copyright (C) 2016 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```

gabrielschulhof: If I add support for both build-time and for compile-time registration, I need to be able to distinguish between the two. What tells me that the linker automagically adds the symbols?

jerry-ext/context/include/jerry-context.h
```diff
@@ -0,0 +1,43 @@
...
+#include "jerryscript.h"
+
+#ifndef JERRYX_CONTEXT_SLOTS
+#define JERRYX_CONTEXT_SLOTS 1
+#endif /* !JERRYX_CONTEXT_SLOTS */
+
+/**
+ * Function to pass to jerry_init_with_user_context() as its second parameter
+ */
+void *jerryx_context_init (void);
```
jiangzidong: you used the prefix  `jerryx_context_` in this ext. But actually, this ext is about mgmt of multiple items (slots) of the user_context of the context, but not the jerry context.

IMO, the name 'user_context' is also confusing, because it is not a context but a user data pointer in the context.

Similarly `context_slot` is also confusing, it is like we have multiple contexts.

My suggestion (draft)
* change `user_context` to `user_data`
* the prefix `jerryx_user_data_slot_`

Make sense?

jerry-ext/context/jerry-context.c
```diff
@@ -0,0 +1,136 @@
...
+typedef struct
+{
+  jerry_user_context_init_cb init_cb;
+  jerry_user_context_deinit_cb deinit_cb;
+} jerryx_context_slot_info_t;
+
+/* Context-free data */
+
+/* We do not allocate slots after a context has already been created. */
+bool before_first_context = true;
```
jiangzidong: static bool

jerry-ext/context/jerry-context.c
```diff
@@ -0,0 +1,136 @@
...
+static int next_index = -1;
```
jiangzidong: what about `static int available_index = 0;` and in the` jerryx_context_request_slot ` we use `slot_index = (available_index ++);`

jerry-ext/context/jerry-context.c
jiangzidong: what about name `static bool allow_request_slot = true`

jerry-ext/module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+#define JERRYX_MODULE_HEADER_UNLINK(link) \
+  ((jerryx_module_header_t *) link)->next = NULL
+
+#define JERRYX_MODULE_HEADER_RUNTIME_INIT(link, module_name) \
+  ((jerryx_module_header_t *) link)->name = (module_name); \
+  JERRYX_MODULE_HEADER_UNLINK((link))
+
+typedef struct
+{
+  jerryx_module_header_t link;
```
jiangzidong: `link` -> `head` or `header` ?

jerry-ext/module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+  int version;
+  jerry_value_t (*onresolve)(void);
```
jiangzidong: on_resolve

jerry-ext/module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+  struct jerryx_module_header *next;
+} jerryx_module_header_t;
+
+#define JERRYX_MODULE_HEADER_STATIC_INIT(module_name) \
+  {                                                 \
+    .name = (module_name),                          \
+    .next = NULL                                    \
+  }
+
+#define JERRYX_MODULE_HEADER_UNLINK(link) \
```
jiangzidong: `#define JERRYX_MODULE_HEADER_NEXT_NULL(head)`

jerry-ext/module/include/jerry-module.h
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Function to create a module manager context
+ */
+void *jerryx_module_context_init (void);
+
+/**
+ * Function to clean up a given module manager context
+ */
+void jerryx_module_context_deinit (void *context_p);
```
jiangzidong: `module_context` =>`module_manager`

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+static const char not_found_prologue[] = "Module '";
+static const char not_found_epilogue[] = "' not found";
+
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name)
+{
+  jerry_value_t return_value = 0;
+  size_t error_message_size = 0, name_length = strlen ((char *) name);
+  char *error_message = NULL;
+  jerryx_module_t *module = NULL;
```
jiangzidong: module_p

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+  (jerryx_context_get (jerryx_module_user_context_slot))
+#else /* !JERRYX_MODULE_HAVE_CONTEXT */
+#define JERRYX_MODULE_CONTEXT \
+  (jerry_get_user_context ())
+#endif /* JERRYX_MODULE_HAVE_CONTEXT */
+
+/* This list of module definitions is constructed by calls to jerry_module_register (). jerry_module_register () does
+ * not require a user context. The list is simply a collection of global static structures defined in each module,
+ * arranged into a linked list. */
+static jerryx_module_t *modules = NULL;
```
jiangzidong: static jerryx_module_t *module_list_head = NULL;

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+ * limitations under the License.
+ */
+
+#include "jmem.h"
+#include "jerryscript.h"
+#include "jerry-module.h"
+
+static void
+jerryx_module_header_insert (jerryx_module_header_t *link,
+                            jerryx_module_header_t **root)
```
jiangzidong: Based on my understanding, `root` is usually for tree, and `head`/`tail` is for linked list.
What about the signature
```c
static void
jerryx_module_list_insert_head (jerryx_module_header_t *item_p,
                                jerryx_module_header_t **head_pp)
```

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+    link->next = (*root);
+  }
+  (*root) = link;
+} /* jerryx_module_header_insert */
+
+static jerryx_module_header_t *
+jerryx_module_header_find (jerryx_module_header_t *root,
+                           const jerry_char_t *name,
+                         jerryx_module_header_t **previous)
+{
```
jiangzidong:
```
jerryx_module_list_find (jerryx_module_header_t *head_p,
                            const jerry_char_t *name,
                          jerryx_module_header_t **last_item_pp)
```

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+    }
+  }
+
+  return root;
+} /* jerryx_module_header_find */
+
+static jerryx_module_header_t *
+jerryx_module_header_remove (const jerry_char_t *name,
+                             jerryx_module_header_t **root)
+{
```
jiangzidong:
```c
jerryx_module_list_remove (const jerry_char_t *name,
                             jerryx_module_header_t **head_pp)
```

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+    JERRYX_MODULE_HEADER_UNLINK (return_value);
+  }
+
+  return return_value;
+} /* jerryx_module_header_remove */
+
+static void
+jerryx_module_header_free (jerryx_module_header_t *root,
+                           void (*free_link) (jerryx_module_header_t *link))
+{
```
jiangzidong:
```
static void
jerryx_module_list_free (jerryx_module_header_t *head_p,
                            void (*free_item) (jerryx_module_header_t *item))
```

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+  jerryx_module_header_t *next;
```
jiangzidong: next_p

jerry-ext/module/jerry-module.c
```diff
@@ -0,0 +1,194 @@
...
+  {
+    next = root->next;
+    free_link (root);
+    root = next;
+  }
+} /* jerryx_module_header_free */
+
+typedef struct
+{
+  jerryx_module_header_t link;
```
jiangzidong: `link` => `head` or `header `

jerry-ext/module/include/jerry-module.h
gabrielschulhof: I'm reluctant to rename this function because the idea is that `jerry_module_context_init ()` and `jerry_module_context_deinit ()` can be passed directly to `jerry_init_with_user_context ()`. Now, if we choose to rename the user context in core to, like, `jerry_init_with_user_data ()` and `jerry_get_user_data ()`, then calling these functions `jerryx_module_user_data_init ()` and `jereryx_module_user_data_deinit ()` will start to make sense.

jiangzidong: I think it makes sense to pass `jerryx_module_manager_init` to `jerry_init_with_user_data`, becasue in this case, the `user_data` is  exactly the "module manager". We create the manager by `jerryx_module_manager_init` and then assign it to the context's user_data member. What do you think?

tests/unit-core/jerry-ext/module/my-custom-module.c
```diff
@@ -0,0 +1,25 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript.h"
+#include "jerry-module.h"
+
+static jerry_value_t
+my_custom_module_init (void)
```
martijnthe: `init` => `onresolve`

jerry-ext/context/include/jerry-context.h
gabrielschulhof: How about this:
In core:
`jerry_init_with_user_data ()` and `jerry_get_user_data ()`
and we rename `jerry-ext/context` to `jerry-ext/user-data` and the `jerry-ext/user-data` API then consists of `jerryx_user_data_init ()`, `jerryx_user_data_deinit ()`, `jerryx_user_data_request_slot ()`, and `jerryx_user_data_get ()`.

jerry-ext/module/include/jerry-module.h
gabrielschulhof: You're right. I will make the change.

jerry-ext/context/include/jerry-context.h
jiangzidong: It is much better to me (non-native-speaker).

Do you know any other words to replace `user_data`? I'm not sure, I am just afraid that people will treat `user_data` as sth like multiple-user-management, will they?

martijnthe: The current core API is called `jerry_init_with_user_context`. IMHO, this name fits with its purpose. User refers to the user of JerryScript and context refers to the context of the current JerryScript instance (`jerry_context_t`). At that level, those words fit well I think.

To me the key thing of the layer currently called `jerry-context` is the multiplexing of the single user context pointer that core provides. IMO, the current name and also the suggestions you are proposing don't really cover that aspect.

Maybe `multi-context`? `mux_ctx`? `multi_user_ctx`?

I don't know.. naming is hard!

gabrielschulhof:
Actually, we already have that. The module *definition* is stored in a global static structure in each module which is passed to `jerry_module_register ()` from a constructor. `jerry_module_register ()` connects these structures into a linked list. This is part of the executable structure, so it's done without any allocation and, if you wish to make a module available to *any* context, then you have to link it in. You can't save space that way.

Now, per-context, a module *instance* is stored on the context in a separate linked list. An instance is not created unless the module is requested in that context. So, the size of the linked list in any given context depends on how many modules were requested.

When the context extension is present, the module manager takes up only a single slot in each context, because it does its own allocation when a module *instance* is created.

The reason for `jerry-ext/context` is not so each module can have a slot, but so the module manager can have *one* slot while possibly retaining a free slot for *something other than the module manager*.

I'll make a Google Drawing of the memory layout so we have something clear to look at.

gabrielschulhof:
... and I'm actually thinking about doing away with `jerry_module_register ()` and using a special section for module definitions, just like I'll be using for module resolvers and context slots. That way, *everything* becomes part of the structure of the executable, and only module instances need to be cached *on demand* and *per-context*.

martijnthe:
> An instance is not created unless the module is requested in that context.

At the module layer, I guess my project could have some access control logic before it calls `jerryx_module_resolve`. So that's a way to solve that part.

However, if you have bindings that do not use the module system (and only the context muxing layer), the `jerry_user_context_init_cb`s of those bindings will be called unconditionally. There's currently no easy way to control that on a per-Jerry-instance basis (w/o modifying the util code).

See my point?

gabrielschulhof:
@martijnthe OK. I take this to mean that the problem is restricted to `jerry-ext/context` and that what we need is per-context variability in the amount of memory allocated by `jerry-ext/context` and probably the indices of the various slots.

I designed `jerry-ext/context` with the assumption that there won't be a need for too many slots, because, pretty much it's one slot per framework, because, just like modules, a framework uses one slot and then manages its own memory within that slot.

So, if we want per-context length and index variability, then we cannot go with indices stored in global static integers and assigned at boot time. We could use a string-keyed data structure like a linked list, the same way as modules use it but I suspect that's a lot of room wasted on meta-data, because there won't be a need for that many slots (again, my assumption).

I'm by no means an expert on data structures but I can't help thinking that, if JerryScript can accommodate JavaScript objects, then it must contain an efficient mechanism for creating string-keyed variable-length associative arrays. Can we not factor that mechanism out and have a string-keyed, pointer-valued associative array made available to extensions? (Disclosure: We may actually already have that, just like we have jmem, but I haven't checked :))

gabrielschulhof:
Oh, and my other assumption was that there won't be too many concurrent contexts, so the number of unused slots will not be multiplied by a large number of contexts.

gabrielschulhof:
In terms of market research, since this functionality has never before been available, and since we're not fully committed to semver, I think we might benefit from releasing `jerry-ext/context` with a fixed number of slots and if our users complain that it wastes too much memory, we can later improve it.

Of course, I'm also happy to wrangle the problem of per-context variation in length and indexing *a priori* :)

martijnthe:
OK, I guess one solution to my problem: if you want per-Jerry-instance control over what bindings are usable & may take up resource, build everything on top of the module system, not the user context muxing layer.

Re. index variation: I think it's fine to keep the indices static and some have unused slots if the initializer wasn't called because it got filtered out.

Anyway, I'll just build everything on top of the module system instead of the user context muxing layer, that works too.

jerry-ext/context/include/jerry-context.h
jiangzidong: The confusing part is `use_context` is not a context, it is a user specific data space inside of a context.

`multiple-context` is also inappropriate. We have PR #1778 about multiple-context, which means jerry can switch whole context to another. But here, all the slots are in one context.

I think `user_data` is better than `user_context`.

Any ideas? @martijnthe

gabrielschulhof:
Yeah, that was my thinking too. The context muxing is there to answer the question *How many different reasons do you have for attaching stuff to a context?* for the case where you have more than one reason. Modules are *one* such reason, and modules alone **already** offer a lot of power.

gabrielschulhof:
Another thing to consider is that extensions, whether modules or not, can probably use `jerry_set_object_native_handle ()` for per-context data.

jerry-ext/context/include/jerry-context.h
gabrielschulhof: Actually, I think `jerryx_context` is kinda appropriate as an abbreviation of "JerryScript extended context", although as such it is a pun on "JerryScript extension for context management".

jiangzidong: That's the problem, this PR seems not an extended context, but an extended user_context (a member of the jerry context), right?

martijnthe: Maybe @akosthekiss or @zherczeg have some ideas on how to name this layer?

martijnthe:
Would it be an idea to split off the muxing layer into a separate PR? Just to keep the git history cleaner and have more concise commit messages with each piece.

akosthekiss:
@gabrielschulhof Please, update the PR to latest master. We already have jerry-ext in the tree, and a rebase would help (at least to me) to see better what's the addition of this PR and what was needed to set up jerry-ext until it hasn't landed.

gabrielschulhof:
@akosthekiss I'll do that ASAP.

martijnthe:
@gabrielschulhof will you be putting up a PR for the module parts (on top of the context-items-link-list, #1833)?

gabrielschulhof:
@martijnthe Real Soon Now™

PR #1786, opened 2017-04-26
target: tizenrt-artik05x: Add missing jerry_port.c

glistening:
Add missing jerry_port.c in previous initial port.


targets/tizenrt-artik05x/apps/jerryscript/jerry_port.c
```diff
@@ -0,0 +1,121 @@
...
+ */
+
+#include <stdio.h>
+#include <string.h>
+#include <stdlib.h>
+#include <setjmp.h>
+#include <stdarg.h>
+
+#include "jerryscript.h"
+#include "jmem.h"
```
akosthekiss: This is an internal header of jerry-core. It should not be used outside of the library (an does not seem to be used in this file anyway).

targets/tizenrt-artik05x/apps/jerryscript/jerry_port.c
```diff
@@ -0,0 +1,121 @@
...
+
+
+/**
+ * JerryScript log level
+ */
+static jerry_log_level_t jerry_log_level = JERRY_LOG_LEVEL_ERROR;
```
akosthekiss: Is this static global needed? There is no support to set its value anywhere, so it might be easier to just use JERRY_LOG_LEVEL_ERROR directly in jerry_port_log.

targets/tizenrt-artik05x/apps/jerryscript/jerry_port.c
glistening: It is necessary for type `jerry_log_level`. I can remove it after #1749 is landed. But I found "jerryscript-port.h" is better header for 'jerry_log_level'.

I will remove 'jmem.h' immediately. It seems leftover.

targets/tizenrt-artik05x/apps/jerryscript/jerry_port.c
glistening: I want to enable the setting jerry_log_level. But there is no `jerry_log_level` setting api in jerryscript-port.h. In fact, I defined the setting function internally and used it. But recently many changes happen in especially in porting and external functions. So I tried to put the minimals. Anyway, I will follow your suggestion temporarily.

targets/tizenrt-artik05x/apps/jerryscript/jerry_port.c
akosthekiss: Exactly, please use "jerryscript-port.h"

targets/tizenrt-artik05x/apps/jerryscript/jerry_port.c
akosthekiss: There is deliberately no common port API function for setting the log level, as jerry does not call anything like that. The port API contains such functions only, which are called from jerry-core. If a port needs log level setting, it has to define any variables needed for storing that value and any functions for accessing (R and/or W) that value.

glistening: Before #1777 landed, jerry_port.c was linked in jerry library so application cannot access the jerry_log_level, defined in jerry library, without external API or making jerry_log_level public. But since for now, at least tizenrt port can follow your suggestion. For your information, currently `jerryscript-port-default` defines `void jerry_port_default_set_log_level (jerry_log_level_t level);`

yichoi:
LGTM

PR #1788, opened 2017-04-27
Ensure that jerry-port is license-checked

akosthekiss:


yichoi:
It would be better to list up in alphabet order

akosthekiss:
Sure. Updated.

PR #1792, opened 2017-04-28
`config.h` maintenance

akosthekiss:
The config header was full of legacy, no-longer-in-use defines. It
was also included from multiple files where it was not necessary.
This patch removes the unused defines, and cleans up headers so
that only `ecma-globals.h` includes `config.h`.

### DIFF
--- a/jerry-core/config.h
+++ b/jerry-core/config.h
@@ -27,13 +27,6 @@
 #define CONFIG_MEM_STACK_LIMIT (4096)
 
 /**
- * Size of pool chunk
- *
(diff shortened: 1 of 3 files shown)


akosthekiss:
I was surprised to see at least some of the defines unused (CONFIG_MEM_POOL_CHUNK_SIZE, CONFIG_LITERAL_HASH_TABLE_KEY_BITS, CONFIG_ECMA_REFERENCE_COUNTER_LIMIT, CONFIG_ECMA_CHAR_ENCODING), but it may be the case that as the project evolved, the concepts represented by these defines became fixed/unconfigurable.

Issue #1794, opened 2017-04-28
Add C API for Promise


martijnthe:
Now we've got ES6 `Promise` support from within .js, I think it's a good time to add the C API as well :)

martijnthe:
cc @jiangzidong @jprestwo @grgustaf

jiangzidong:
sure, like `jerry_create_promise` and `jerry_value_is_promise`, right?

martijnthe:
Yeah and API to reject / fullfill, for example:
https://github.com/01org/zephyr.js/blob/master/src/zjs_promise.h

@jprestwo @grgustaf what else?

jprestwo:
@martijnthe,
`create`, `fulfill`, and `reject` are all we use/need in ZJS. `is_promise` could be useful as well.

martijnthe:
@jprestwo, can you help me understand why you have the `zjs_post_promise_func post` argument in `zjs_make_promise`? Is this needed here too you think? Can't you use the native handle/pointer API for this?

jprestwo:
@martijnthe, no that is not needed. I have a pending PR for ZJS that uses JerryScript promises and I took out that parameter:
https://github.com/01org/zephyr.js/pull/1058

Of course this PR is irrelevant once JerryScript has C API's.

jiangzidong:
On second thought, I don't think the promise api in https://github.com/01org/zephyr.js/pull/1058 is suitable for jerry C api. Because it saves the resolve/reject handler inside the native pointer of promise object, however, the ideal `create_promise` should return a promise object which has empty native pointer, so that users can set it depends on their own scenario.

I suggest the the signature should be 
```
jerry_value_t jerry_create_promise (jerry_value_t *resolve_p, jerry_value_t *reject_p)
```
the resolve_p and reject_p are function object, and are OUT arguments, so that users get the handler, and should maintain them (including free them). When they want to resolve a promise, they just jerry_call_function(*resolve_p, ......). And we don't need the explicit api like fullfill/resolve/reject.

What do you think? @martijnthe @jprestwo

martijnthe:
@jiangzidong I'm not sure if I like that signature.
It is closest to how the JS Promise constructor works, but in reality I think it will be very cumbersome to use:

1. Many async C APIs take a `void *cb_data` as parameter. With your proposal, you have to allocate a helper containing the 2 `resolve` and `reject` jerry values and pass a pointer to the helper as `cb_data`.
[...]

PR #1796, opened 2017-05-02
Add promise C API

jiangzidong:
related issue: 1794

* add api: 
```
bool jerry_value_is_promise (const jerry_value_t value);
jerry_value_t jerry_create_promise (void);
jerry_value_t jerry_resolve_promise (jerry_value_t promise, jerry_value_t argument);
jerry_value_t jerry_reject_promise (jerry_value_t promise, jerry_value_t reason);
```

* unittest support es2015-subset profile

* doc will be added later when we reach the consensus

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -1072,6 +1072,47 @@ jerry_value_is_object (const jerry_value_t value)
 - [jerry_release_value](#jerry_release_value)
 
 
+## jerry_value_is_promise
+
+**Summary**
+
+Returns whether the given `jerry_value_t` is a promise value.
+
+*Note*: This API depends on the ES2015-subset profile.
+
+**Prototype**
+
+```c
+bool
+jerry_value_is_promise (const jerry_value_t value)
+```
+
+- `value` - api value
+- return value
+  - true, if the given `jerry_value_t` is a promise
+  - false, otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t value;
+  ... // create or acquire value
+
+  if (jerry_value_is_promise (value))
+  {
+    ...
+  }
+
+  jerry_release_value (value);
+}
+```
+
+**See also**
+
+- [jerry_release_value](#jerry_release_value)
+
+
 ## jerry_value_is_string
 
 **Summary**
@@ -1980,6 +2021,64 @@ jerry_value_to_string (const jerry_value_t value);
 - [jerry_value_to_primitive](#jerry_value_to_primitive)
 
 
+# Functions for promise objects
+
+These APIs all depends on the ES2015-subset profile.
+
+## jerry_resolve_or_reject_promise
+
+**Summary**
+
+Resolve or reject the promise with an argument.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_resolve_or_reject_promise (jerry_value_t promise,
+                                 jerry_value_t argument,
+                                 bool is_resolve)
+```
+
+- `promise` - the promise value
+- `argument` - the argument for resolve or reject
+- `is_resolve` - whether the promise should be resolved or rejected
+- return value
+  - undefined jerry value - resolve or reject successed
+  - jerry value with error flag - otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t promise = ... // acquire/create a promise object.
+
+  ...
+
+  bool is_resolve = ... // whether the promise should be resolved or rejected
+  jerry_value_t argument = ... // prepare the argumnent for the resolve or reject.
+
+  jerry_value_t is_ok = jerry_resolve_or_reject_promise (promise,
+                                                         argument,
+                                                         is_resolve);
+
+  if (jerry_value_has_error_flag (is_ok))
+  {
+    // handle the error.
+  }
+
+  jerry_release_value (is_ok);
+  jerry_release_value (argument);
+  jerry_release_value (promise);
+}
+```
+
+**See also**
+
+- [jerry_release_value](#jerry_release_value)
+- [jerry_value_has_error_flag](#jerry_value_has_error_flag)
+
+
 # Acquire and release API values
 
 ## jerry_acquire_value
@@ -2422,6 +2521,41 @@ jerry_create_object (void);
 
 **See also**
 
+- [jerry_release_value](#jerry_release_value)
+
+
+## jerry_create_promise
+
+**Summary**
+
+Create an empty promise object which can be resolved or rejected later
+by calling jerry_resolve_or_reject_promise.
+
+*Note*: This API depends on the ES2015-subset profile.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_promise (void)
+```
+
+- return value - value of the newly created promise
+
+**Example**
+
+```c
+{
+  jerry_value_t p = jerry_create_promise ();
+
+  ...// usage of the promise
+
+  jerry_release_value (p);
+}
+
+**See also**
+
+- [jerry_resolve_or_reject_promise](#jerry_resolve_or_reject_promise)
 - [jerry_release_value](#jerry_release_value)
 
 
@@ -3528,7 +3662,7 @@ static const jerry_object_native_info_t native_obj_type_info =
     {
       // The type of this's native pointer matches what is expected.
       // Only now is it safe to cast to native_obj_t * and dereference the
-      // pointer: 
+      // pointer:
       native_obj_t *native_obj = native_p;
       native_obj->bar = ...; // Safe to access now!
     }
@@ -3582,7 +3716,7 @@ jerry_set_object_native_pointer (const jerry_value_t obj_val,
 **Example**
 
 See [jerry_get_object_native_pointer](#jerry_get_object_native_pointer) for a
-best-practice example. 
+best-practice example.
 
 **See also**
 
--- a/jerry-core/ecma/operations/ecma-promise-object.h
+++ b/jerry-core/ecma/operations/ecma-promise-object.h
@@ -38,6 +38,16 @@ typedef enum
 } ecma_promise_state_t;
 
 /**
+ * Indicates the type of the executor in promise construct.
+ */
+typedef enum
+{
+  ECMA_PROMISE_EXECUTOR_FUNCTION, /**< the executor is a function, it is for the usual constructor */
+  ECMA_PROMISE_EXECUTOR_OBJECT, /**< the executor is an object, it is for the `then` routine */
+  ECMA_PROMISE_EXECUTOR_EMPTY /**< the executor is empty, it is for external C API */
+} ecma_promise_executor_type_t;
+
+/**
  * Description of the promise resolving functions.
  */
 typedef struct
@@ -81,10 +91,7 @@ void ecma_promise_set_result (ecma_object_t *obj_p, ecma_value_t result);
 uint8_t ecma_promise_get_state (ecma_object_t *obj_p);
 void ecma_promise_set_state (ecma_object_t *obj_p, uint8_t state);
 ecma_value_t
-ecma_op_create_promise_object (ecma_value_t executor, bool is_func);
-ecma_promise_resolving_functions_t *
-ecma_promise_create_resolving_functions (ecma_object_t *obj_p);
-void ecma_promise_free_resolving_functions (ecma_promise_resolving_functions_t *funcs);
+ecma_op_create_promise_object (ecma_value_t executor, ecma_promise_executor_type_t type);
 ecma_value_t ecma_promise_new_capability (void);
 ecma_value_t
 ecma_promise_then (ecma_value_t promise,
--- a/jerry-core/jerry.c
+++ b/jerry-core/jerry.c
@@ -29,6 +29,7 @@
 #include "ecma-literal-storage.h"
 #include "ecma-objects.h"
 #include "ecma-objects-general.h"
+#include "ecma-promise-object.h"
 #include "jcontext.h"
 #include "jerryscript.h"
 #include "jerry-debugger.h"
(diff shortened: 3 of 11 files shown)


jerry-core/jerry.c
```diff
@@ -553,6 +554,26 @@ jerry_value_is_object (const jerry_value_t value) /**< api value */
...
+ * Check if the specified value is promise.
+ *
+ * @return true  - if the specified value is promise,
+ *         false - otherwise
+ */
+bool
+jerry_value_is_promise (const jerry_value_t value) /**< api value */
+{
+  jerry_assert_api_available ();
+#ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
```
zherczeg: newline before ifdef

tests/unit-core/test-promise.c
```diff
@@ -0,0 +1,161 @@
...
+
+static int count_in_assert = 0;
+static jerry_value_t my_promise1;
+static jerry_value_t my_promise2;
+const jerry_char_t s1[] = "resolved";
+const jerry_char_t s2[] = "rejected";
+
+static jerry_value_t
+create_promise1_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
+                        const jerry_value_t this_val __attribute__((unused)), /**< this value */
```
zherczeg: One more space needed.

tests/unit-core/test-promise.c
```diff
@@ -0,0 +1,161 @@
...
+{
+  jerry_value_t ret =  jerry_create_promise ();
+  my_promise1 = jerry_acquire_value (ret);
+
+  return ret;
+} /* create_promise1_handler */
+
+static jerry_value_t
+create_promise2_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
+                        const jerry_value_t this_val __attribute__((unused)), /**< this value */
```
zherczeg: One more space needed.

jerry-core/jerry.c
zherczeg: Please add newlines before other ifdefs in this file.

jerry-core/jerry.c
```diff
@@ -1019,6 +1040,26 @@ jerry_create_object (void)
...
+ * @return value of the created object
+ */
+jerry_value_t
+jerry_create_promise (void)
+{
+  jerry_assert_api_available ();
+#ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
+  return ecma_op_create_promise_object (ecma_make_simple_value (ECMA_SIMPLE_VALUE_EMPTY), ECMA_PROMISE_EXECUTOR_EMPTY);
+#else /* CONFIG_DISABLE_ES2015_PROMISE_BUILTIN */
+  return ecma_raise_syntax_error (ECMA_ERR_MSG ("Promise profile has been disabled."));
```
martijnthe: Maybe better to exclude the implementation so you get a linker error?
I don't know... returning an error adds code space.

jerry-core/jerry.c
```diff
@@ -2121,6 +2162,85 @@ jerry_foreach_object_property (const jerry_value_t obj_val, /**< object value */
...
 /**
+ * Common part for jerry_resolve_promise and jerry_reject_promise.
+ *
+ * @return undefined value - if success
+ *         value marked with error flag - otherwise
+ */
+static jerry_value_t
+jerry_resolve_or_reject_promise (jerry_value_t promise,
+                                 jerry_value_t argument,
+                                 bool is_resolve)
```
martijnthe: I think it makes sense to expose something like this in the public API.

I think people will be writing code like in [this example] a lot, so you naturally have a `bool` to make the decision whether `resolve` or `reject` should be called.

In an [earlier sketch], I used an `enum`, I think that has a bit better readability vs `bool`, but I'm not feeling strongly about it.

[this example]: https://github.com/jerryscript-project/jerryscript/issues/1794#issuecomment-298308945
[earlier sketch]: https://github.com/jerryscript-project/jerryscript/issues/1794#issuecomment-298319724

martijnthe:
@jprestwo any feedback on this API?

jerry-core/jerry.c
jiangzidong: My thought is the C api should be similar with js api, so it will be easy to use. In JS world, usually the code is like
``` javascript
var p = new Promise(function(resolve, reject)
{
  ....
  if (some condition)
  {
    var reason = .....//prepare the reason
     reject (reason)
  }
  
  var argument = .... //prepare the argument
   resolve(argument)
})
```

in your example, 
``` c
static void my_get_batt_info_cb (const batt_info_t *info, void *cb_data)
{
        jerry_value_t promise = (jerry_value_t)(uintptr_t) cb_data; 
        
        // do stuff with *info and assign result + should_call_resolve:
        jerry_value_t result = ...;
        bool should_call_resolve = ...;
        
        jerry_call_resolve_reject(promise, result, should_call_resolve);
        jerry_value_release(promise);
}
```
there should be some if-else or branch code in `result = ` and `should_call_resolve =` , e.g.
``` c
if (some condition)
 { result = some_error_reason; should_call_resolve = false; }
else
{result = some_argument; should_call_resolve = true;}
```
why just directly call the resolve/reject in the branch?

martijnthe: > why just directly call the resolve/reject in the branch?

Code space savings. You'd have 2x function calls vs 1x.
Often you already have the `bool should_call_resolve` one way or another, for example in your snippet, there is `some_condition`. In this case, `some_condition == !should_call_resolve`.

jiangzidong: Ok, it makes sense :)

docs/02.API-REFERENCE.md
```diff
@@ -1980,6 +2021,139 @@ jerry_value_to_string (const jerry_value_t value);
...
+Get the resolve function of the promise object.
+
+*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
+is no longer needed.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_promise_get_resolve (jerry_value_t promise)
```
zherczeg: What I don't understand is that we have one function to resolve/reject, but two functions for getting functions. I think it would be better to have a boolean parameter here as well.

martijnthe: My gut says that if your project needs access to the functions, you probably are going to need both.

Would it make sense to have:

```
bool jerry_promise_get_resolve (
     jerry_value_t promise, jerry_value_t *resolve_out_p, jerry_value_t *reject_out_p)
```

martijnthe: Hmm I wonder if we're adding the resolve/reject getter prematurely...
Maybe just remove it and get on with life until someone comes along that has a clear use for it?

jiangzidong: agree

docs/02.API-REFERENCE.md
```diff
@@ -1980,6 +2021,64 @@ jerry_value_to_string (const jerry_value_t value);
 - [jerry_value_to_primitive](#jerry_value_to_primitive)
 
 
+# Functions for promise value
```
zherczeg: Functions for promise objects.

docs/02.API-REFERENCE.md
```diff
@@ -1980,6 +2021,64 @@ jerry_value_to_string (const jerry_value_t value);
...
+
+**Example**
+
+```c
+{
+  jerry_value_t promise = ... // acquire/create a promise object.
+
+  ...
+
+  bool is_resolve = ... // whether the promise should be resolve or reject
```
zherczeg: resolved or rejected

docs/02.API-REFERENCE.md
```diff
@@ -2425,6 +2524,41 @@ jerry_create_object (void);
 - [jerry_release_value](#jerry_release_value)
 
 
+## jerry_create_promise
+
+**Summary**
+
+Create an empty Promise object which can be resolve/reject later
```
zherczeg: resolved/rejected

docs/02.API-REFERENCE.md
```diff
@@ -2425,6 +2524,41 @@ jerry_create_object (void);
...
+*Note*: This API depends on the ES2015-subset profile.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_promise (void)
+```
+
+- return value - value of the created promise
```
zherczeg: newly created promise

jerry-core/ecma/operations/ecma-promise-object.h
```diff
@@ -38,6 +38,16 @@ typedef enum
 } ecma_promise_state_t;
 
 /**
+ * Indicates the type of the executor in promise construct.
+ */
+typedef enum
+{
+  ECMA_PROMISE_EXECUTOR_FUNCTION, /**< the executor is a function, it is for the usual constructor */
+  ECMA_PROMISE_EXECUTOR_OBJECT, /**< the executor is a object, it is for the `then` routine */
```
zherczeg: the executor is an object

jerry-core/jerry.c
```diff
@@ -2175,6 +2176,84 @@ jerry_set_vm_exec_stop_callback (jerry_vm_exec_stop_callback_t stop_cb, /**< per
...
+jerry_value_is_promise (const jerry_value_t value) /**< api value */
+{
+  jerry_assert_api_available ();
+
+  return (ecma_is_value_object (value)
+          && ecma_is_promise (ecma_get_object_from_value (value)));
+} /* jerry_value_is_promise */
+
+
+/**
```
zherczeg: Only one newline.

jerry-core/jerry.c
```diff
@@ -2175,6 +2176,83 @@ jerry_set_vm_exec_stop_callback (jerry_vm_exec_stop_callback_t stop_cb, /**< per
...
+#ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
+
+/**
+ * Check if the specified value is promise.
+ *
+ * @return true  - if the specified value is promise,
+ *         false - otherwise
+ */
+bool
+jerry_value_is_promise (const jerry_value_t value) /**< api value */
```
LaszloLango: Please use the same order of functions as 'jerryscript.h'.

tests/unit-core/test-promise.c
```diff
@@ -0,0 +1,162 @@
...
+                           );
+
+static int count_in_assert = 0;
+static jerry_value_t my_promise1;
+static jerry_value_t my_promise2;
+const jerry_char_t s1[] = "resolved";
+const jerry_char_t s2[] = "rejected";
+
+static jerry_value_t
+create_promise1_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
```
LaszloLango: `JERRY_UNUSED` must be available here since test-common.h includes jrt.h
```c
static jerry_value_t
create_promise1_handler (const jerry_value_t func_obj_val , /**< function object */
                         const jerry_value_t this_val, /**< this value */
                         const jerry_value_t args_p[], /**< arguments list */
                         const jerry_length_t args_cnt) /**< arguments length */
{
JERRY_UNUSED (func_obj_val);
JERRY_UNUSED (this_val);
JERRY_UNUSED (args_p);
JERRY_UNUSED (args_cnt);

...
```

tests/unit-core/test-promise.c
```diff
@@ -0,0 +1,162 @@
...
+                         const jerry_length_t args_cnt __attribute__((unused))) /**< arguments length */
+{
+  jerry_value_t ret =  jerry_create_promise ();
+  my_promise1 = jerry_acquire_value (ret);
+
+  return ret;
+} /* create_promise1_handler */
+
+static jerry_value_t
+create_promise2_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
```
LaszloLango: ditto

tests/unit-core/test-promise.c
```diff
@@ -0,0 +1,162 @@
...
+                         const jerry_length_t args_cnt __attribute__((unused))) /**< arguments length */
+{
+  jerry_value_t ret =  jerry_create_promise ();
+  my_promise2 = jerry_acquire_value (ret);
+
+  return ret;
+} /* create_promise2_handler */
+
+static jerry_value_t
+assert_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
```
LaszloLango: ditto

tests/unit-core/CMakeLists.txt
```diff
@@ -19,8 +19,8 @@ if (NOT IS_ABSOLUTE ${FEATURE_PROFILE})
   set(FEATURE_PROFILE "${CMAKE_SOURCE_DIR}/jerry-core/profiles/${FEATURE_PROFILE}.profile")
 endif()
 
-if(NOT ${FEATURE_PROFILE} STREQUAL "${CMAKE_SOURCE_DIR}/jerry-core/profiles/es5.1.profile")
-  message(FATAL_ERROR "FEATURE_PROFILE='${FEATURE_PROFILE}' isn't supported with UNITTESTS=ON")
+if(${FEATURE_PROFILE} STREQUAL "${CMAKE_SOURCE_DIR}/jerry-core/profiles/minimal.profile")
+  message(FATAL_ERROR "minimal profile isn't supported in unit-core")
```
LaszloLango: Is this change related to this this PR? It looks to me if promise is not available than the test-promise returns zero, which is good. Am I missing something?

jiangzidong: Before this change, I cannot `./tools/build.py --unittest --profile=es2015-subset`, because it's profile is not es5.1.

jiangzidong: And also I changed the run-test.py, now es2015-subset profile is passed to unittest

```
# Test options for unittests
JERRY_UNITTESTS_OPTIONS = [
    Options('unittests',
            ['--unittests', '--error-messages=on', '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset']),
    Options('unittests-debug',
            ['--unittests', '--debug', '--error-messages=on', '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset'])
]
```

LaszloLango: I see. Thanks for the explanation.

martijnthe:
💯

Issue #1795, opened 2017-04-28
Prefixing "internal global" symbols of jerry-core


akosthekiss:
# Problem / Motivation

I've looked around the jerry-core library a bit to see what we expose of ourselves. I was especially interested in the global symbols available from the static `libjerry-core.a` archive. We have two types of global symbols, at least conceptually: there are those which are part of the public API (e.g., `jerry_release_value` -- let's call these "public globals"), and there are those which should probably not be available to the outside world but cannot be source-local `static`s because of modularization (e.g., `vm_run` or `ecma_op_to_string` -- let's call these "internal globals"). We have had problems with "internal globals" before: internal memory management routines used to have a `mem_` prefix but that conflicted with other projects that wanted to embed the library. We ended up renaming the memory management module to `jmem` and re-prefixing all routines with `jmem_`. However, I don't think that the threat of conflicting symbols has been completely lifted.

In the most basic build variant, on OSX, I get the following results when looking for global symbols that _don't_ have the `jerry_` prefix:

```bash
$ ./tools/build.py --clean
$ nm ./build/lib/libjerry-core.a -g | grep -v ' U ' | grep -v 'jerry_' | wc -l
     611
```

This means that there are 600+ global symbols that are only for the internal working of jerry-core but may potentially conflict with code we have no influence on. What's more frightening (to me, at least) is that these "internal globals" follow quite diverse naming schemes:

```bash
$ nm ./build/lib/libjerry-core.a -g | grep -v ' U ' | grep -v 'jerry_' | cut -d ' ' -f 2,3 | cut -d '_' -f 1,2 | sort | uniq -c
   1
   1 ./build/lib/libjerry-core.a(jerry-all-in.c.o):
   2 S _cbc
  35 S _ecma
   2 T _do
 399 T _ecma
  16 T _jmem
  10 T _lexer
  56 T _lit
  15 T _opfunc
  44 T _parser
  19 T _re
   1 T _util
  10 T _vm
```

I think it's quite hard to prepare our users: it's very hard to tell them what identifiers they _must_ avoid if they don't want to run into cryptic linker/runtime(?) issues. (E.g., a `do_`, `util_`, or `re_` prefix is not unimaginable to be used elsewhere.)

_Note_: I'm not concerned about "claiming" the `jerry_` prefix. That should not conflict. If someone chooses this project, s/he should live with the fact that this prefix is reserved for the JS engine.

# Suggestion / Options

**Option 1**: The easy route

[...]

PR #1801, opened 2017-05-06
Rename internal debugger identifiers

akosthekiss:
We should keep the `jerry_` prefix for public API which is either
declared in jerry-core/include or implemented in jerry-core/api.
Moreover, we should also keep the convention to use short
identifiers for componentization within jerry-core, and use these
identifiers as directory names, file name prefixes, and identifier
prefixes.

Therefore, the jerry-core/debugger directory is renamed as
jerry-core/dbg, the files in that directory are renamed to start
with dbg- instead of jerry-debugger-, and all internal identifiers
that started with `jerry_debugger_` got renamed as `dbg_`.

Note: There is a single enum type with all its labels that did not
get renamed: `jerry_debugger_header_type_t`. The very same
`JERRY_DEBUGGER_` identifiers and associated values are used at
other parts of the project, too: in the two debugger client
implementations (HTML/JS, and Python), to make communication
possible. Not renaming those IDs in jerry-core helps keeping these
in sync.

Also, the tools/gen-unicode.py-generated arrays in
jerry-core/lit are renamed to use `lit_` prefix instead of `jerry_`.

### DIFF
--- a/jerry-core/lit/lit-char-helpers.c
+++ b/jerry-core/lit/lit-char-helpers.c
@@ -136,11 +136,11 @@ lit_char_is_white_space (ecma_char_t c) /**< code unit */
   {
     return (c == LIT_CHAR_NBSP
             || c == LIT_CHAR_BOM
-            || (c >= jerry_unicode_separator_char_interval_sps[0]
-                && c <= jerry_unicode_separator_char_interval_sps[0] + jerry_unicode_separator_char_interval_lengths[0])
+            || (c >= lit_unicode_separator_char_interval_sps[0]
+                && c <= lit_unicode_separator_char_interval_sps[0] + lit_unicode_separator_char_interval_lengths[0])
             || search_char_in_char_array (c,
-                                          jerry_unicode_separator_chars,
-                                          NUM_OF_ELEMENTS (jerry_unicode_separator_chars)));
+                                          lit_unicode_separator_chars,
+                                          NUM_OF_ELEMENTS (lit_unicode_separator_chars)));
   }
 } /* lit_char_is_white_space */
 
@@ -181,10 +181,10 @@ static bool
 lit_char_is_unicode_letter (ecma_char_t c) /**< code unit */
 {
   return (search_char_in_interval_array (c,
-                                         jerry_unicode_letter_interval_sps,
-                                         jerry_unicode_letter_interval_lengths,
-                                         NUM_OF_ELEMENTS (jerry_unicode_letter_interval_sps))
-          || search_char_in_char_array (c, jerry_unicode_letter_chars, NUM_OF_ELEMENTS (jerry_unicode_letter_chars)));
+                                         lit_unicode_letter_interval_sps,
+                                         lit_unicode_letter_interval_lengths,
+                                         NUM_OF_ELEMENTS (lit_unicode_letter_interval_sps))
+          || search_char_in_char_array (c, lit_unicode_letter_chars, NUM_OF_ELEMENTS (lit_unicode_letter_chars)));
 } /* lit_char_is_unicode_letter */
 
 /**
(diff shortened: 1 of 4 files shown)


akosthekiss:
Note: This PR is only tangentially related to issue #1795. I know that the issue is still open without conclusion, and may possibly remain so for long (or even close with a conclusion not to change anything). The key question of that issue is whether we should add a common prefix to all "internal global" symbols to avoid name clashes with external projects. This PR aims at fixing an existing inconsistency _within_ jerry-core: the `jerry_` prefix is used _mostly_ for public API symbols (types, function names), but not always, there are deviations (library-internal symbols, be they local or global, being prefixed with `jerry_`). I argue that `jerry_` prefix should be used for API symbols _only_ - this PR is the first attempt at that.

(As the reception of #1795 was ambivalent, this PR may also face some opposition. I'm happy/ready to discuss.)

jerry-core/CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ message(STATUS "MEM_HEAP_SIZE_KB          " ${MEM_HEAP_SIZE_KB})
 set(INCLUDE_CORE
     "${CMAKE_CURRENT_SOURCE_DIR}"
     "${CMAKE_CURRENT_SOURCE_DIR}/api"
-    "${CMAKE_CURRENT_SOURCE_DIR}/debugger"
+    "${CMAKE_CURRENT_SOURCE_DIR}/dbg"
```
zherczeg: I don't really like this abbreviation. Actually I don't like abbreviations, they make so hard to understand the code.

akosthekiss: We already have api, ecma, (j)mem, jrt, lit, js, re, vm. Actually, even jerry is an abbreviation. So, based on established precedent, I see nothing bad in abbreviations. Moreover, with the line length rules we have in place, "jerry_debugger" as a prefix takes more than 10% of the allowed spaced (compare it to "dbg"). Finally, it is very subjective, but DBG is quite a clear abbreviation for DeBuG or DeBuGging to me.

zherczeg: Let's see what others think.

akosthekiss: ping others?

LaszloLango: I understand and I can accept the motivation of this PR, but I'd prefer `jdebugger_` or `jdbg_` prefixes.

akosthekiss: why that `j` all over? (will we also have `jvm`?)

akosthekiss: (no question: if I ever had the chance, I'd get rid of them from `jcontext` and `jmem`, too. as explained at #1795. fortunately, it's not up to me, only :) )

LaszloLango: dbg_ prefix is too general. IMHO might conflict with other symbols

akosthekiss: I thought that with prefix conflicts we go with the "cross that bridge when we come to it" approach?
> do nothing until this issue got hit by someone

zherczeg: Actually I would be ok with this if the prefix would be longer rather than shorter.

LaszloLango: Sure but don't increase the chance of it's possible. It's not the same.

LaszloLango: You are planning to change an almost never conflicting name with a new that will possibly conflicting.

akosthekiss: do we agree at least that jerry_ prefix should not be used for non-public symbols?

LaszloLango: yes

zherczeg: Personally I would prefer using less namespaces, and use jerry_ prefix everywhere. However this would be too much change, so probably ecma_ namespace could be kept as well. On the longer run I would prefer removing all other namespaces at the first time when they cause any problems (e.g. if somebody complains about jmem, I would rename it to jerry_alloc_). This reduces namespace conflicts.

Regarding this change even jerry_dbg_ is better than debugger_ for me. Although longer identifiers are not a problem as long as the code is well designed. In the debugger the jerry_debugger_ never caused any problems, and the chance of symbol clash is very low. So making this shorter just for the sake of making it shorter is not exactly a good reason for me.

akosthekiss: Please, share your thoughts at #1795, too.

It seems that there will be no easy consensus on this topic, so let's drop the debugger rewrite for now (the patch already became bitrotten anyway). But what abut the `lit`-related part (the topmost commit)?

zherczeg:
I am neutral about renaming the lit part (i.e. you can land it if you want).

akosthekiss:
PR updated (debugger-related commit dropped, only lit-changes remained)

glistening:
Here is my two cents:
[uv](libuv/libuv) uses the convention using `uv_` for public global, and `uv__` for internal global. We may use this convention. (i.e., `jerry_` for public APIs, `jerry__` for global internal). It will avoid name clash and all jerry's globals (either public or internal global) starts with `jerry_`.

zherczeg:
jerry__ is ok for me. Since we have jerryx_, perhaps we could consider jerryg_ as well. Anyway I would not do anything until somebody has a symbol clash.

PR #1803, opened 2017-05-08
Remove include jerryscript.h from jrt.h.

zherczeg:


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -26,6 +26,7 @@
 #include "ecma-objects-arguments.h"
 #include "ecma-try-catch-macro.h"
 #include "jcontext.h"
+#include "jerryscript-core.h"
```
LaszloLango: Why is this header needed here?

zherczeg: Because it calls the user callbacks.

akosthekiss: but the whole point of the rewrite is to completely get rid of the jerry_ types in internals of the library, isn't it?

zherczeg: Valid point. Ok I remove it, it is easy.

zherczeg: I just realized this is very good for the project since we can do all external_pointer_t casts and let the compiler to do its checks. The patch is a bit bigger but well worth it.

zherczeg:
Patch updated.

Issue #1799, opened 2017-05-04
Is the JobQueue Port API a good abstraction?


akosthekiss:
TL;DR: Is there any other way to implement a Promise job queue than what's implemented in the default port? If not, should it not be merged into jerry-core?

Source of my confusion:
- If Promises are available/enabled then those JS functions that are in `then`s or `catch`es are to be executed after the main script completed. Until they get their turn, they are enqueued by `jerry_port_jobqueue_enqueue`.
- Enqueued jobs _must_ be executed, otherwise they keep references to their environment and memory will not be freed. (Just try to comment out `jerry_port_default_jobqueue_run` in main-unix.c, the debug-built jerry tool will assert out at cleanup when running a script with Promises, because heap does not get emptied.)
- However, there is no general API to execute the enqueued jobs. The default port provides `jerry_port_default_jobqueue_run`, but it is specific to that port.
- What's more, the default port implementation uses the engine's internal memory handling routines (`jmem_heap_alloc_block`, `jmem_heap_free_block`), which is quite a no-no IMHO.

I've taken a quick look at another VM (JavaScriptCore) and command line tool (jsc), how it handles Promises and enqueued jobs. Without digging too deep, it seems to me that there the VM has an API call to execute such jobs (called microtasks, if I'm not mistaken).
(`vm.drainMicrotasks();` at https://trac.webkit.org/browser/webkit/trunk/Source/JavaScriptCore/jsc.cpp#L3522)

Wouldn't it be better for us, too, to integrate Promise job handling into the engine. E.g., something like
- `jerry_run_enqueued` to run one enqueued job,
- `jerry_run_all_enqueued` to run until the queue gets empty (replacement for `jerry_port_default_jobqueue_run` and analogous to JSC's `vm.drainMicrotasks`, I think).

Advantages:
- the whole API needed to run a Promise-enabled engine is port-independent (in addition to the above discussed, `jerry_init` can do the queue initialization instead of  `jerry_port_default_jobqueue_init`),
- no API layering violation,
- less burden on port implementers.

However, I'm not sure whether there are any use cases for which this approach would be inappropriate.

Feedbacks are welcome.

akosthekiss:
cc @jiangzidong @martijnthe

martijnthe:
Prior discussion for the reasoning behind putting the queue implementation in the port:
https://github.com/jerryscript-project/jerryscript/pull/1685#issuecomment-289285036

[...]

PR #1804, opened 2017-05-08
Move the job queue from the ports into jerry-core

akosthekiss:
Related issue: #1799


martijnthe:
@jiangzidong is on vacation this week. I think he should also take a look at this.
Would you mind waiting until coming week?

akosthekiss:
no worries, doesn't look like fast changing code

jiangzidong:
@akosthekiss sorry for the late response. good patch to me, except that api `jerry_run_all_enqueued`. It seems only used in the unix-main case (the simplest queue), and I am not sure whether other ports will use it. Should we remove that API, and implement the simple queue in `unix-main` instead of in `jerry-core` ?

akosthekiss:
@jiangzidong thanks for the feedback. I've added a commit that implements your suggestion.

zherczeg:
What is the motivation of this? The point of the design was that most systems has a built-in processing mechanism for this. This could be an extension at best, but never core.

akosthekiss:
@zherczeg Please see issue referenced in the very first comment of this PR (#1799). There is a 2000 characters long analysis on the issues with the current job API (+ discussion with others).

jiangzidong:
@zherczeg, the original implementation needs users to write the port-specific code for jobqueue, but the new patch doesn't. Though the target platform may have own event-queue, it is better to seperate the outer event queue and internal job queue. 

@akosthekiss , I am sorry I want to change my last review comments. After reading the code of V8 and nodejs (they also use microtask, like jsc did), I think we should 
 * keep the `jerry_run_all_enqueued` : we can change the name to jerry_run_enqueued_jobs
 * remove the `jerry_run_enqueued` and `jerry_port_job_enqueued`

After users call `jerry_run`, or call the callbacks in the binding layer, users also need call `jerry_run_all_enqueued` to let jerry execute **all** **current** jobs in queue. If the new jobs are enqueued during `jerry_run_all_enqueued` , jerry will execute it soon (without any block)

If the promise will be resolved/rejected later (which covers the 99%+ cases of promise), for example, 
```javascript
var p = new Promise(function(resolve, reject)
{
  setTimeout(function(){resolve(...)}, 1000);// resolve the promise after 1s
})
p.then(...)
```
We don't need `jerry_port_job_enqueued` to inform the platform, `setTimeout` will add the corresponding timer event into the event-queue, and user should not cleanup the jerry if the event-queue is not empty. When the timer hit, users call the callback (`function(){resolve(...)}`) **and then call the `jerry_run_all_enqueued`**. `resolve()` will enqueue a job, and then the job will be executed. All above things are operated by `setTimeout` binding code and event-queue code, jerry itself just provide the api `jerry_run_all_enqueued` to let users call.

I cant find the case where we need `jerry_run_enqueued`. We should always finish all available jobs in any ripe situation.

Does it make sense?

zherczeg:
Promise is a new feature and has not fully completed yet. I worry about defining core API for it. What about making this as an extension?

jiangzidong:
@zherczeg I think promise is fully completed (except for some unknown bugs)

akosthekiss:
@zherczeg Promises cannot be implemented outside of core. Even if the API is experimental it must be in core, not in an extension.

@jiangzidong works with me, but I think an "is the job queue empty?" query may still be useful in addition to the "run all enqueued jobs" function.

jiangzidong:
@akosthekiss "is_empty" will be called inside "run_all_jobs". Do you have any ideas about the cases where users want to know the status of the jobqueue but don't want to execute the jobs?

akosthekiss:
@jiangzidong not really, so I'll drop that and keep "run all jobs" only.

akosthekiss:
I've updated the PR and rebased onto latest master. Now, the only added core API function is `jerry_run_all_enqueued_jobs` and ports don't have to implement anything anymore.

glistening:
@akosthekiss I am almost ready to start porting promise on tizenrt-artik05x since minimal porting is almost done. I can check that current structure and api are reusable for other platform. Could you please wait before landing this?

akosthekiss:
@glistening yes, can do (the nice thing is, however, that with this PR, there is no porting effort needed)

docs/02.API-REFERENCE.md
```diff
@@ -757,6 +757,42 @@ jerry_eval (const jerry_char_t *source_p,
 - [jerry_create_external_function](#jerry_create_external_function)
 - [jerry_external_handler_t](#jerry_external_handler_t)
 
+## jerry_run_all_enqueued_jobs
+
+**Summary**
+
+Run enqueued Promise jobs until the first thrown error or until all get executed.
+
```
LaszloLango: Please add a note too:
` *Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
 is no longer needed.`

akosthekiss: will do (unfortunately, we're not consistent about this in the docs. `jerry_run` and `jerry_eval` have this note in the sources, too, but not in the docs. that's why I've skipped adding it for `jerry_run_all_enqueued_jobs`)

LaszloLango: That should be fixed in the docs than.

jerry-core/api/jerry.c
```diff
@@ -447,6 +447,26 @@ jerry_eval (const jerry_char_t *source_p, /**< source code */
...
 /**
+ * Run enqueued Promise jobs until the first thrown error or until all get executed.
+ *
+ * Note:
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return result of last executed job, may be error value.
+ */
+jerry_value_t
+jerry_run_all_enqueued_jobs (void)
```
zherczeg: I would prefer a `jerry_run_enqueued_jobs` with some parameter which provides some control. The problem is that a job can create other jobs, and this could lead to an infinite loop, which might not be desired by the platform. So it would be good to have a termination condition (e.g. max number of executed jobs, -1 for any).

docs/02.API-REFERENCE.md
akosthekiss: in this PR or later?

jerry-core/api/jerry.c
akosthekiss: But those jobs are only executed in the same `jerry_run_enqueued_jobs` call if the promise can be resolved immediately, don't they? Which is not the classic use of promises. And even that only causes problems if a job always enqueues itself (or something like that) immediately. Which is really an infinite loop, analogous to `while (true) {}`, but that's a programmer error. We have no way of preventing while-true either. (Other than the recently added stop-the-vm callback, but that'd work here, too.)

jiangzidong: Yes, it just like the following code in nodejs, which will lead to infinite loop, but it is not engine/runtime's fault.
```
function foo()
{
  console.log("1");
  process.nextTick(foo);
}
foo();
```

martijnthe:
> We don't need jerry_port_job_enqueued to inform the platform, setTimeout will add the corresponding timer event into the event-queue

But what if a promise is created that does not involve any native bindings / add any event to the project's event queue? For example:
```
Promise.resolve.('xyz').then(function(v) {
 // should be run async.
 // v == 'xyz'
});
```

I also wonder if `jerry_run_enqueued_jobs` is going to be a source of re-ordering problems.
Imagine:
```
while (e = get_next_project_event_or_block()) {
  e->run_handler(e);
  jerry_run_enqueued_jobs();
}
```

Imagine a job that 1) calls a native binding that puts an event on the project's event queue 2) adds another job to the job queue. Now the job of 2) will run before the event handler for 1) will run. I think it's much more logical and easier to wrap your brain around if jobs and platform events are handled in the same order as which they were enqueued.

martijnthe:
Just had an offline discussion about the re-ordering issue.
It's a bit of an hypothetical case, so maybe just run with what's here and see if it's really an issue when using this simplified API?

glistening:
@akosthekiss, @zherczeg , @LaszloLango  I changed my plan. I will port promise to **Linux x86_64 IoT.js** first (since it is easier and faster than tizenrt-artik05x development environment.). As a baby step, I just enabled promise without any change (presumably it used default-port), and it showed assertion failure for very simple example (Samsung/iotjs#845). I will take a closer look for Promise the default-port and APIs, and fix the bug. Since I am not yet familiar with jerry-core internal, any suggestion or clues will help me much.

akosthekiss:
@glistening Please, see my comment there. If promises are enabled, you have to drain the job queue before shutting down the engine.

glistening:
@akosthekiss I succeed to run the simple promise example in iotjs. And I think it will work for tizenrt-artik05x, too.

docs/02.API-REFERENCE.md
LaszloLango: Sorry, I just saw your last comment. The note for this should be added in this PR. The missing notes from other API functions can go in a new PR, because they are not related to this patch.

yichoi:
LGTM

glistening:
@akosthekiss To make my point clear, I am ok for landing this patch now.

PR #1807, opened 2017-05-08
Fix jobqueue's return value handling in jerry-main's repl mode

akosthekiss:


yichoi:
LGTM

PR #1809, opened 2017-05-09
Add lightweight command line option processor to jerry-main

akosthekiss:
Eases command line option definition, usage summary and detailed
help message printing, and option processing.

### DIFF
--- a/jerry-main/cli.c
+++ b/jerry-main/cli.c
@@ -1,0 +1,429 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <stdio.h>
+#include <string.h>
+
+#include "cli.h"
+
+/*
+ * Fixed layout settings
+ */
+
+/**
+ * Wrap lines at:
+ */
+#define CLI_LINE_LENGTH 80
+
+/**
+ * Indent various lines with:
+ */
+#define CLI_LINE_INDENT 2
+
+/**
+ * Tab stop (for multi-column display) at:
+ */
+#define CLI_LINE_TAB 24
+
+/*
+ * Print helper functions
+ */
+
+/**
+ * Pad with spaces.
+ */
+static void
+cli_print_pad (int cnt) /**< number of spaces to print */
+{
+  for (int i = 0; i < cnt; i++)
+  {
+    printf (" ");
+  }
+} /* cli_print_pad */
+
+/**
+ * Print the prefix of a string.
+ */
+static void
+cli_print_prefix (const char *str, /**< string to print */
+                  int len) /**< length of the prefix to print */
+{
+  for (int i = 0; i < len; i++)
+  {
+    printf ("%c", *str++);
+  }
+} /* cli_print_prefix */
+
+/**
+ * Print a help message wrapped into the second column.
+ */
+static void
+cli_print_help (const char *help) /**< the help message to print */
+{
+  while (help != NULL && *help != 0)
+  {
+    int length = -1;
+    int i = 0;
+    for (; i < CLI_LINE_LENGTH - CLI_LINE_TAB && help[i] != 0; i++)
+    {
+      if (help[i] == ' ')
+      {
+        length = i;
+      }
+    }
+    if (length < 0 || i < CLI_LINE_LENGTH - CLI_LINE_TAB)
+    {
+      length = i;
+    }
+
+    cli_print_prefix (help, length);
+
+    help += length;
+    while (*help == ' ' && *help != 0)
+    {
+      help++;
+    }
+
+    if (*help != 0)
+    {
+      printf ("\n");
+      cli_print_pad (CLI_LINE_TAB);
+    }
+  }
+} /* cli_print_help */
+
+/*
+ * Command line option handling
+ */
+
+/**
+ * Initialize a command line option processor.
+ *
+ * @return the state that should be passed iteratively to cli_opt_process.
+ */
+cli_opt_state_t
+cli_opt_init (const cli_opt_t *opts, /**< array of option definitions, terminated by CLI_OPT_END */
+              int argc, /**< number of command line arguments */
+              char **argv) /**< array of command line arguments */
+{
+  return (cli_opt_state_t)
+  {
+    .opts = opts,
+    .argc = argc,
+    .argv = argv,
+    .opt = NULL,
+    .arg = NULL
+  };
+} /* cli_opt_init */
+
+/**
+ * Perform one step of the command line option processor.
+ *
+ * @return the ID of the option that was found. (The definition of the found
+ *         option (if any) is available via state->opt, while the option
+ *         string and its arguments are available via state->arg[0..].)
+ *         CLI_OPT_END signals that all command line arguments are consumed.
+ */
+int
+cli_opt_process (cli_opt_state_t *state) /**< state of the command line option processor */
+{
+  if (state->argc <= 0)
+  {
+    state->opt = NULL;
+    state->arg = NULL;
+    return CLI_OPT_END;
+  }
+
+  state->arg = state->argv;
+
+  for (const cli_opt_t *o = state->opts; o->id != CLI_OPT_END; o++)
+  {
+    state->opt = o;
+
+    if (o->id == CLI_OPT_POSITIONAL && (state->arg[0][0] != '-' || !strcmp (state->arg[0], "-")))
+    {
+      state->argc--;
+      state->argv++;
+      return CLI_OPT_POSITIONAL;
+    }
+    else if ((o->opt != NULL && !strcmp (o->opt, state->arg[0]))
+        || (o->longopt != NULL && !strcmp (o->longopt, state->arg[0])))
+    {
+      if (state->argc > o->argc)
+      {
+        state->argc -= 1 + o->argc;
+        state->argv += 1 + o->argc;
+        return o->id;
+      }
+      else
+      {
+        state->argv += state->argc;
+        state->argc = 0;
+        return CLI_OPT_INCOMPLETE;
+      }
+    }
+  }
+
+  state->opt = NULL;
+  state->argc--;
+  state->argv++;
+  return CLI_OPT_UNKNOWN;
+} /* cli_opt_process */
+
+/**
+ * Print usage summary of options.
+ */
+void
+cli_opt_usage (const char *progname, /**< program name, typically argv[0] */
+               const cli_opt_t *opts) /**< array of command line option definitions, terminated by CLI_OPT_END */
+{
+  int length = (int) strlen (progname);
+  printf ("%s", progname);
+
+  for (const cli_opt_t *o = opts; o->id != CLI_OPT_END; o++)
+  {
+    const char *opt = o->opt != NULL ? o->opt : o->longopt;
+    opt = opt != NULL ? opt : o->meta;
+
+    int opt_length = (int) strlen (opt);
+    if (o->argc > 0)
+    {
+      opt_length += o->meta != NULL ? 1 + (int) strlen (o->meta) : o->argc * 2;
+    }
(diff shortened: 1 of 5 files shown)


zherczeg:
We should wait landing #1797 first.

akosthekiss:
On the contrary, I'd like to see this landed first. The CLI of our C tools is less than professional, and the snapshot merging tool just continues that bad tradition. I'd like to see the new tool use proper command line option handling, too.

zherczeg:
The command line parsing in that patch is correct, and it would be better if this patch would contain all changes related to moving to a new method. I would rather to avoid a situation when that patch needs to rework command line parsing again.

LaszloLango:
Is this a binary size optimization?

akosthekiss:
> The command line parsing in that patch is correct

We disagree in that regard. It arbitrarily differs from the other tools we are using/writing. (git is not a right analogy for that tool, but I'll comment there, too.) I'm happy to show how to adapt to this approach and get a more flexible, easier to use, more intuitive CLI.

> Is this a binary size optimization?

No (or, at least, I don't know). As far as I know, we don't care about the size of `jerry` the tool. This turns this

```
$ build/bin/jerry --help
Usage: build/bin/jerry [OPTION]... [FILE]...

Options:
  -h, --help
  -v, --version
  --mem-stats
  --mem-stats-separate
  --parse-only
  --show-opcodes
  --show-regexp-opcodes
  --start-debug-server
  --save-snapshot-for-global FILE
  --save-snapshot-for-eval FILE
  --save-literals-list-format FILE
  --save-literals-c-format FILE
  --exec-snapshot FILE
  --log-level [0-3]
  --abort-on-fail
  --no-prompt
```
into this
```
$ build/bin/jerry --help
build/bin/jerry [-h] [-v] [--mem-stats] [--mem-stats-separate] [--parse-only]
  [--show-opcodes] [--show-regexp-opcodes] [--start-debug-server]
  [--save-snapshot-for-global FILE] [--save-snapshot-for-eval FILE]
  [--save-literals-list-format FILE] [--save-literals-c-format FILE]
  [--exec-snapshot FILE]... [--log-level NUM] [--abort-on-fail] [--no-prompt]
  [FILE]...

  -h, --help            print this help and exit
  -v, --version         print tool and library version and exit
  --mem-stats           dump memory statistics
  --mem-stats-separate  dump memory statistics and reset peak values after parse
  --parse-only          don't execute JS input
  --show-opcodes        dump parser byte-code
  --show-regexp-opcodes dump regexp byte-code
  --start-debug-server  start debug server and wait for a connecting client
  --save-snapshot-for-global FILE
  --save-snapshot-for-eval FILE
  --save-literals-list-format FILE
  --save-literals-c-format FILE
  --exec-snapshot FILE  execute input snapshot file(s)
  --log-level NUM       set log level (0-3)
  --abort-on-fail       segfault on internal failure (instead of non-zero exit code)
  --no-prompt           don't print prompt in REPL mode
  FILE                  input JS file(s)
```

LaszloLango:
I see. Good patch. Why don't have any description for `--save-*` options? Is that intentional?

akosthekiss:
> Why don't have any description for --save-* options?

Now, I have to reveal that I'm not sure what they do exactly (or rather, what the exact difference between the variants is). I'd be happy to add the help messages if someone sent them to me (PM is fine).

zherczeg:
--save-snapshot-for-global FILE
Snapshot will be evaluated at global context (even if we run it from a JS callback)
--save-snapshot-for-eval FILE
Snapshot will be evaluated at local context (variables in the scope chain are partly visible to the snapshot)
--save-literals-list-format FILE
Literals encountered during parsing exported in a list format (useful for external magic strings).
--save-literals-c-format FILE
Literals encountered during parsing exported in a C format (useful for external magic strings).

jerry-main/cli.c
```diff
@@ -0,0 +1,260 @@
...
+
+/**
+ * Pad with spaces.
+ */
+static void
+cli_pad (int cnt) /**< number of spaces to print */
+{
+  for (int i = 0; i < cnt; i++)
+  {
+    printf (" ");
```
galpeter: Just a nitpick: can't we do just this?: ` printf("%*s", cnt, " "); `

LaszloLango: I'm not sure, but I think jerry-libc does not support it.

akosthekiss: Unfortunately, no. jerry-libc doesn't support `*` width specifier

akosthekiss:
Thanks for helping me out with the help strings. PR updated

akosthekiss:
PR is updated (option help message line wrapping added + some minor light touches). current output is:
```
$ ./build/bin/jerry --help
./build/bin/jerry [-h] [-v] [--mem-stats] [--mem-stats-separate] [--parse-only]
  [--show-opcodes] [--show-regexp-opcodes] [--start-debug-server]
  [--save-snapshot-for-global FILE] [--save-snapshot-for-eval FILE]
  [--save-literals-list-format FILE] [--save-literals-c-format FILE]
  [--exec-snapshot FILE]... [--log-level NUM] [--abort-on-fail] [--no-prompt]
  [FILE]...

  -h, --help            print this help and exit
  -v, --version         print tool and library version and exit
  --mem-stats           dump memory statistics
  --mem-stats-separate  dump memory statistics and reset peak values after
                        parse
  --parse-only          don't execute JS input
  --show-opcodes        dump parser byte-code
  --show-regexp-opcodes dump regexp byte-code
  --start-debug-server  start debug server and wait for a connecting client
  --save-snapshot-for-global FILE
                        save binary snapshot of parsed JS input (for execution
                        in global context)
  --save-snapshot-for-eval FILE
                        save binary snapshot of parsed JS input (for execution
                        in local context by eval)
  --save-literals-list-format FILE
                        export literals found in parsed JS input (in list
                        format)
  --save-literals-c-format FILE
                        export literals found in parsed JS input (in C source
                        format)
  --exec-snapshot FILE  execute input snapshot file(s)
  --log-level NUM       set log level (0-3)
  --abort-on-fail       segfault on internal failure (instead of non-zero exit
                        code)
  --no-prompt           don't print prompt in REPL mode
  FILE                  input JS file(s)
```

zherczeg:
GCC says

    Usage: gcc [options] file...
    Options:
      -pass-exit-codes         Exit with highest error code from a phase
      ...

I think a similar approach is enough for us.

akosthekiss:
@zherczeg but compare to our other tools `tools/build.py --help`, `tools/run-tests.py --help`, etc.

jerry-main/cli.c
```diff
@@ -0,0 +1,429 @@
...
+ * Print a help message wrapped into the second column.
+ */
+static void
+cli_print_help (const char *help) /**< the help message to print */
+{
+  for (; help != NULL && *help != 0;)
+  {
+    int length = -1;
+    int i = 0;
+    for (; i < CLI_LINE_LENGTH - CLI_LINE_TAB && help[i] != 0; i++)
```
zherczeg: for (int i = 0;

jerry-main/cli.c
```diff
@@ -0,0 +1,429 @@
...
+ * Initialize a command line option processor.
+ *
+ * @return the state that should be passed iteratively to cli_opt_process.
+ */
+cli_opt_state_t
+cli_opt_init (const cli_opt_t *opts, /**< array of option definitions, terminated by CLI_OPT_END */
+              int argc, /**< number of command line arguments */
+              char **argv) /**< array of command line arguments */
+{
+  return (cli_opt_state_t)
```
zherczeg: Returning with structures is always costly. It is better to pass a pointer.

jerry-main/cli.c
```diff
@@ -0,0 +1,429 @@
...
+  printf ("\n");
+} /* cli_opt_usage */
+
+/**
+ * Print detailed help for options.
+ */
+void
+cli_opt_help (const cli_opt_t *opts) /**< array of command line option definitions, terminated by CLI_OPT_END */
+{
+  for (; opts->id != CLI_OPT_END; opts++)
```
zherczeg: This should be a while loop.

jerry-main/cli.h
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Quantifiers of command line options
+ */
+typedef enum
+{
+  CLI_QUANT_Q, /**< ? (Question mark: optional, zero or one) */
+  CLI_QUANT_A, /**< * (Asterisk, star: zero or one or more) */
+  CLI_QUANT_P, /**< + (Plus sign: one or more) */
+  CLI_QUANT_1  /**< 1 (One: one) */
```
zherczeg: What about two argument.

akosthekiss: No need for that, as there is no use case other than these, which could be supported easily. Note that the quantifiers are for display only, not enforced in any other way.
* `?` gets displayed as `[-x X]` or `[X]`
* `*` gets displayed as `[-x X]...` or `[X]...`
* `+` gets displayed as `-x X...` or `X...`
* `1` gets displayed as `-x X` or `X`

There is no easy or conventional notation for more general quantifiers (like min 3, max infinity) for command line arguments, and I'm not sure we'd need them. (This is not a general purpose CLI library, it's for the needs of jerry-main only.)

jerry-main/cli.c
akosthekiss: cannot do that, `i` is used after the loop

jerry-main/cli.c
akosthekiss: On many platforms, this is exactly what happens when returning structs. AFAIK both on ARM32 and X86, struct returns are done by passing a hidden pointer argument to the function, where it has to write the results. So, `x = f();` and `f(&x);` are exactly the same. (The former might work better with optimizations, as the ABI-enforced extra pointer argument might be easier to track in pointer aliasing algorithms than completely generic pointer arguments, but that can be use case specific, whether there is a real difference).

(BTW, this is not performance-critical code (even if there would be any performance penalty). This function is typically called only once in an application. And this is for jerry-main only, not for anyone to use.)

jerry-main/cli.h
zherczeg: (Just don't forget that the snapshot merging patch requires an optional argument with two items, a name and an index)

akosthekiss: That's not the same, that's what `.argc` is for. `.quant` describes whether and option (together with its argument(s)) is optional or not (or how many times it should appear on the command line). E.g., `--exec-snapshot snapfile1 --exec-snapshot snapfile2 --exec-snapshot snapfile3` is a valid use, that's why we have `[--exec-snapshot FILE]...` in the usage. For the snapshot merging patch, another option is to be added, used as `--exec-snapshot-func snapfile1 index1` IIRC, which means that it has two arguments, but the option together with its two arguments may also appear arbitrarily many times on the command line (`.argc = 2, .meta = "FILE NUM", .quant = CLI_QUANT_A`), displayed in usage as `[--exec-snapshot-func FILE NUM]...`.

akosthekiss:
I've updated the PR and rebased it to latest master.

tools/cppcheck/suppressions-list
```diff
@@ -2,4 +2,4 @@ wrongmathcall:tests/unit-libm/test-libm.inc.h
 variableScope:jerry-libm/*.c
 invalidPointerCast:jerry-libm/*.c
 arithOperationsOnVoidPointer:jerry-libc/*.c
-commaSeparatedReturn:jerry-ext/include/jerryscript-ext/arg.impl.h
+commaSeparatedReturn:*
```
LaszloLango: Could you explain this change? I'm not sure suppressing a warning in the whole code base is a good idea.

jerry-main/main-unix.c
```diff
@@ -325,6 +292,71 @@ register_js_function (const char *name_p, /**< name of the function */
...
+  OPT_DEBUG_SERVER,
+  OPT_SAVE_SNAP_GLOBAL,
+  OPT_SAVE_SNAP_EVAL,
+  OPT_SAVE_LIT_LIST,
+  OPT_SAVE_LIT_C,
+  OPT_EXEC_SNAP,
+  OPT_LOG_LEVEL,
+  OPT_ABORT_ON_FAIL,
+  OPT_NO_PROMPT
+} jerry_opt_id_t;
```
LaszloLango: Based on the previous conversation on the naming conventions, I'd rename this to `cli_opt_id_t`. `jerry_` prefix should be reserved for API functions and types.

tools/cppcheck/suppressions-list
akosthekiss: Cppcheck warns on every struct literal in returns, because it misinterprets them as comma-separated expressions. This is quite bad IMO and obviously a shortcoming of cppcheck, because struct literals are completely valid C constructs.

jerry-main/main-unix.c
akosthekiss: what about main_opt_id_t ? (I'd keep cli_ for cli.{c,h})

akosthekiss:
I've rebased the PR to latest master, resolved conflicts, and renamed variables and types in main-unix.c as requested

PR #1811, opened 2017-05-10
Add docs about extensions to webpage-generator script

bzsolt:

### DIFF
--- a/tools/update-webpage.sh
+++ b/tools/update-webpage.sh
@@ -30,6 +30,8 @@ PORT_API_MD="05.PORT-API.md"
 REFERENCE_COUNTING_MD="06.REFERENCE-COUNTING.md"
 DEBUGGER_MD="07.DEBUGGER.md"
 CODING_STANDARDS_MD="08.CODING-STANDARDS.md"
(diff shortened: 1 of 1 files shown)


tools/update-webpage.sh
```diff
@@ -41,6 +43,8 @@ titles[$PORT_API_MD]="Port API"
 titles[$REFERENCE_COUNTING_MD]="Reference Counting"
 titles[$DEBUGGER_MD]="Debugger"
 titles[$CODING_STANDARDS_MD]="Coding Standards"
+titles[$EXT_REFERENCE_ARG_MD]="External Argument Reference"
+titles[$EXT_REFERENCE_HANDLER_MD]="External Handler Reference"
```
bzsolt: Better names are welcome.

akosthekiss: My variants:
Extension API Reference: Argument Validation
Extension API Reference: External Function Handlers

bzsolt: I've omitted the 'Reference' in order to reduce the width of the options and the titles with `:` are enclosed with quote-marks to avoid interpret them as a key-value pair.

PR #1812, opened 2017-05-10
Rework memory statistics to provide useful user information.

zherczeg:


jerry-core/ecma/base/ecma-alloc.c
```diff
@@ -80,12 +80,37 @@ JERRY_STATIC_ASSERT (sizeof (ecma_extended_object_t) - sizeof (ecma_object_t) <=
...
-DECLARE_ROUTINES_FOR (string)
-DECLARE_ROUTINES_FOR (getter_setter_pointers)
+
+/**
+ * Allocate memory for ecma-object
+ *
+ * @return pointer to allocated memory
+ */
+inline ecma_object_t * __attr_always_inline___
+ecma_alloc_object (void)
```
akosthekiss: to reduce copy-paste, couldn't we have templates for stat-collecting alloc/free pairs?

```c
#ifdef JMEM_STATS
#define STAT_ALLOCATE_BYTES(ecma_type) \
  jmem_stats_allocate_ ## ecma_type ## _bytes (sizeof (ecma_ ## ecma_type ## _t));
#define STAT_FREE_BYTES(ecma_type) \
  jmem_stats_free_ ## ecma_type ## _bytes (sizeof (ecma_ ## ecma_type ## _t));
#else /* !JMEM_STATS */
#define STAT_ALLOCATE_BYTES(ecma_type) 
#define STAT_FREE_BYTES(ecma_type) 
#endif /* JMEM_STATS */

/**
 * Template of an allocation routine that collects statistics
 */
#define STAT_ALLOC(ecma_type) \
  ecma_ ## ecma_type ## _t * \
  ecma_alloc_ ## ecma_type (void) \
  { \
    STAT_ALLOCATE_BYTES (ecma_type) \
    return (ecma_ ## ecma_type ## _t *) jmem_pools_alloc (sizeof (ecma_ ## ecma_type ## _t)); \
  }

/**
 * Deallocation routine template that collects statistics
 */
#define STAT_DEALLOC(ecma_type) \
  void \
  ecma_dealloc_ ## ecma_type (ecma_ ## ecma_type ## _t *ecma_type ## _p) \
  { \
    STAT_FREE_BYTES (ecma_type) \
    jmem_pools_free ((uint8_t *) ecma_type ## _p, sizeof (ecma_ ## ecma_type ## _t)); \
  }

/**
 * Declaration of alloc/free routines that collect statistics for specified ecma-type.
 */
#define DECLARE_STAT_ROUTINES_FOR(ecma_type) \
  STAT_ALLOC (ecma_type) \
  STAT_DEALLOC (ecma_type)
```

jerry-core/jmem/jmem-allocator.c
```diff
@@ -153,22 +153,99 @@ jmem_run_free_unused_memory_callbacks (jmem_free_unused_memory_severity_t severi
...
+  jmem_heap_stats_print ();
+} /* jmem_stats_print */
 
 /**
- * Print memory usage statistics
+ * Register string allocaton.
  */
 void
-jmem_stats_print (void)
+jmem_stats_allocate_string_bytes (size_t string_size)
```
akosthekiss: these routines could also be macro-templated

jerry-core/api/jerry.c
```diff
@@ -323,7 +323,6 @@ jerry_parse (const jerry_char_t *source_p, /**< script source */
   if (JERRY_CONTEXT (jerry_init_flags) & JERRY_INIT_MEM_STATS_SEPARATE)
   {
     jmem_stats_print ();
-    jmem_stats_reset_peak ();
```
zherczeg: No, but we need to keep it for compatiblity. At some point it should be dropped.

jerry-core/jmem/jmem-allocator.c
zherczeg: I don't like macros because they are very hard to debug. I would prefer to keep the current code.

zherczeg:
Byte code is also collected now.

jerry-core/api/jerry.c
akosthekiss: what compatibility is that we'd like to keep? is that semantic, API level, or command line level (or something else)? because I see several ways to deprecate this flag. e.g.,
- if we want API compatibility, we could keep the flag in the enum but mark it as deprecated in its doc and don't use it anywhere in jerry-core (perhaps only for logging a warning).
- if we don't want to keep command line compatibility, we could even remove it from the CLI of the jerry tool.

zherczeg: I have no preference. What do you choose?

akosthekiss: no preference. I was just interested what the reason was. this deprecation could even go into a separate PR, but I'd like to learn what compatibility we'd like to keep?

zherczeg: I don't know. I don't think anybody ever used this, since after the new byte code was landed the majority of its functionality had been lost. Perhaps we should keep the constant as a hole filler for now, and drop it in the next release.

jerry-core/ecma/base/ecma-alloc.c
akosthekiss: > I don't like macros because they are very hard to debug. I would prefer to keep the current code.

That can be a valid point. But here, we have a slight inconsistency now. Macro generated alloc routines have and assert on non-null allocation. Should all allocs have that (also the now-written routines) or is it not needed anymore and can be dropped from the remaining templates, too?

jerry-core/api/jerry.c
akosthekiss: works with me

zherczeg: What about the main-unix? Shall I delete the option?

jerry-core/ecma/base/ecma-alloc.c
zherczeg: I would not allow any macro generated functions in the future. So bad for debugging, we are not optimizing for source code size, and it does not affect binary size at all. The existing macro is legacy, and could be removed, but not in this patch. A low hanging fruit for somebody.

jerry-core/api/jerry.c
akosthekiss: if it has no functionality, I'd propagate this change through the whole repo

jerry-core/ecma/base/ecma-alloc.c
akosthekiss: I was not suggesting the removal of the existing `ALLOC` macro. what I pointed out was that `ALLOC` checks the result of `jmem_pools_alloc` and asserts out if it was `NULL`, while handwritten allocs don't check that. now, we have some functions with checks, some without. which one is right?

zherczeg: We have two allocator functions jmem_heap_alloc_block and jmem_heap_alloc_block_null_on_error. The former cannot return with NULL. The null check in the original code is unnecessary.

akosthekiss: we could drop that from line 62 then

zherczeg:
Patch updated, mem stats separate removed.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1424,6 +1424,10 @@ ecma_bytecode_deref (ecma_compiled_code_t *bytecode_p) /**< byte code pointer */
       return;
     }
 #endif /* JERRY_DEBUGGER */
+
+#ifdef JMEM_STATS
+    jmem_stats_free_byte_code_bytes (((size_t) bytecode_p->size) << JMEM_ALIGNMENT_LOG);
+#endif /* JMEM_STATS */
```
akosthekiss: is this stat call at the right place? the corresponding `jmem_heap_free_block` call is several lines below and not within the same `if`

jerry-core/parser/js/js-parser.c
```diff
@@ -1489,6 +1489,10 @@ parser_post_processing (parser_context_t *context_p) /**< context */
 
   compiled_code_p = (ecma_compiled_code_t *) parser_malloc (context_p, total_size);
 
+#ifdef JMEM_STATS
+  jmem_stats_allocate_byte_code_bytes (total_size);
+#endif /* JMEM_STATS */
```
akosthekiss: parser_malloc is used at other places, too, but they don't seem to be stat'd. moreover, parser_free is not stat'd anywhere. is that intentional?

zherczeg: Yes. This measures the byte code only (a typical non-temporary data).

jerry-core/ecma/base/ecma-helpers.c
zherczeg: Yes, because it measures the parser byte code only, regex byte code is different topic. I don't think regexes consumes that much memory in most cases.

PR #1814, opened 2017-05-11
Use external print handler in riot-stm32f4 target

robertsipka:
Extend example code with registering the `print` function in the global object.

### DIFF
--- a/targets/riot-stm32f4/Makefile.riot
+++ b/targets/riot-stm32f4/Makefile.riot
@@ -42,9 +42,10 @@ libjerry:
 	 -DEXTERNAL_COMPILE_FLAGS="$(EXT_CFLAGS)" \
 	 -DMEM_HEAP_SIZE_KB=$(JERRYHEAP)
 
-	make -C$(BUILD_DIR) jerry-core jerry-port-default-minimal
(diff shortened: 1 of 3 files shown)


targets/riot-stm32f4/Makefile.riot
```diff
@@ -42,9 +42,10 @@ libjerry:
 	 -DEXTERNAL_COMPILE_FLAGS="$(EXT_CFLAGS)" \
 	 -DMEM_HEAP_SIZE_KB=$(JERRYHEAP)
 
-	make -C$(BUILD_DIR) jerry-core jerry-port-default-minimal
+	make -C$(BUILD_DIR) jerry-core jerry-port-default-minimal jerry-ext
 	cp $(BUILD_DIR)/lib/libjerry-core.a $(COPYTARGET)/libjerrycore.a
 	cp $(BUILD_DIR)/lib/libjerry-port-default-minimal.a $(COPYTARGET)/libjerryport-minimal.a
+	cp $(BUILD_DIR)/lib/libjerry-ext.a $(COPYTARGET)/libjerry-ext.a
```
akosthekiss: nitpicking: it seems that for some reason, the libraries tend to get renamed by removing the first dashes during the copy. so, most probably, this should be libjerryext.a (although this is nothing functional)

PR #1815, opened 2017-05-11
Lazily create a linked list of context items

gabrielschulhof:
This approach has the benefit that it does not require any *a priori*
initialization, and that each context pointer is identified by the way
in which it was created. Additionally, retrieving the context pointer
now requires that the entity responsible for creating/destroying it
(the manager) be given. Since managers are stored in global static
const structures, they should not normally be visible across source
files, and thus there should be no danger that a context item will be
retrieved by the wrong manager and thus cast into the wrong data type.

Since the items are stored in a linked list, their number will be
limited to exactly as many as are needed for a given context, with the
caveat that storing too many on a context will cause slow retrieval.

Thanks @mhdawson for the idea!


jerry-core/api/jerry.c
```diff
@@ -207,22 +197,33 @@ jerry_cleanup (void)
...
+  jerry_user_context_item_t *item_p;
+
+  for (item_p = JERRY_CONTEXT (user_context_p); item_p; item_p = item_p->next_p)
+  {
+    if (item_p->manager_p == manager_p)
+    {
+      return item_p;
+    }
+  }
+
```
martijnthe: Hmm, this way `init_cb` is only called after the first `jerry_get_user_context()` call? That's a fundamental difference from the approach from #1802, where all `init_cb`s were called during `jerry_init`.

I like the approach of #1802 a little bit better, because it enables components built on top of it to automagically (but optionally) initialize together with JerryScript itself – just add the sources of the component to your project's build and you're done. This is useful for example if the component adds APIs to the global object. If we go with the approach in this PR, you'd manually have to call a `my_component_api_add_to_global_object()` after `jerry_init()`.

zherczeg:
Am I see right this is basically a chain list based lightweight module manager? And it is the responsibility of the user that too many managers have heavy performance costs. This approach is ok for me.

gabrielschulhof:
@zherceg yeah, and if you don't use one in a given context it doesnt take up space.

gabrielschulhof:
Oh, and let's remember that this approach is safer than the previous approach, because there's kind of a runtime type check by virtue of the fact that the `jerry_user_context_manager_t *` you pass to `jerry_get_user_context ()` is compared to the `jerry_user_context_manager_t *` stored in the context item, and bad things happen if none of the items match.

So, even if we choose not to adopt this approach, I believe we should keep the part where `jerry_get_user_context ()` accepts a parameter using which one can specify *whose* pointer to retrieve.

jerry-core/api/jerry.c
gabrielschulhof: I'm mostly concerned about type safety here. When we first designed the user context interface, we made init/deinit safe by requiring a pair of functions. That way, we could ensure that the right deinit was called every time. However, we left the get () completely separate. Thus, I realized that the following problem might happen:
```c
jerry_init_with_user_context(JERRY_INIT_EMPTY, jerry_foo_new, jerry_foo_free);
...
/* A mistake in the code - calls jerryx_bar_do_something_useful()
 * which calls jerry_get_user_context () assuming the context was
 * initialized with jerry_bar_new()/jerry_bar_free() */
jerryx_bar_do_something_useful();
```

Additionally, I also realized that jerryx_context has another shortcoming. Users of jerryx_context must call `jerryx_context_get_slot ()` instead of `jerry_get_user_context ()` to retrieve the context pointer. But how does context user `foo` know that the context was initialized with
```c
jerry_init_with_user_context (JERRY_INIT_EMPTY, jerryx_user_context_init, jerryx_user_context_deinit);
```
and not with
```c
jerry_init_with_user_context (JERRY_INIT_EMPTY, jerryx_foo_init, jerryx_foo_deinit);
```
I'd solved this latter problem with  -DJERRYX_HAVE_CONTEXT at build time and an ifdef that calls `jerry_get_user_context ()` if the flag is absent, and `jerryx_context_get_slot ()` if the flag is absent.

But this means that it does the same thing for every context.

And besides, it is only when an extension calls `jerry_get_user_context ()` that you can be certain that the extension needs the context.

docs/02.API-REFERENCE.md
```diff
@@ -101,6 +101,39 @@ created by API functions has the error flag set.
...
+Structure that holds a single user context item, identified by its manager. It is intended to serve as the base for
+larger structures that need to be allocated at runtime.
+
+**Prototype**
+
+```c
+typedef struct jerry_user_context_item
+{
+  const struct jerry_user_context_manager *manager_p; /**< manager responsible for deleting this item */
+  struct jerry_user_context_item *next_p; /**< pointer to next context item */
```
zherczeg: I would prefer next as the first item.

zherczeg:
I do like this approach. Linear search on a low-number of context managers is totally fits to embedded systems.

jiangzidong:
@gabrielschulhof good patch to me. but about the naming, I still dont think `user_context` is a proper name. (https://github.com/jerryscript-project/jerryscript/pull/1784#discussion_r114477997)

My concern is "it is not a context, but a user data field of the context". and my suggestion is:
 * in the jerry_context_t, the filed name is `user_data`
 * `jerry_user_context_manager_t ` ==> `jerry_context_user_data_manager_t`
 * `jerry_get_user_context ` ==> `jerry_context_get_user_data`

What do yo think about the naming? @zherczeg 

And I am not sure whether the change should be made in this patch.

gabrielschulhof:
@jiangzidong @zherczeg @martijnthe @akosthekiss how about `external_data`?

```
jerry_context_external_data_manager_t
jerry_context_external_data_item_t
jerry_context_get_external_data (jerry_context_external_data_manager_t *manager_p)
```

gabrielschulhof:
We already use `_external_` for native functions that will be exposed to JS.

gabrielschulhof:
I think the name change should be made in this PR, because this PR amounts to a re-write of the context external data interface.

jiangzidong:
@gabrielschulhof `external_data` is a good name :)

gabrielschulhof:
Sold! :)

zherczeg:
For me `jerry_context_external_data_item_t` sounds a bit strange, and this is actually a reference, not the data itself. So for me `pointer` or `reference` sounds better.

`jerry_context_external_pointer_t`
`jerry_context_external_reference_t`

`jerry_context_external_data_pointer_t` is too long for me. However `jerry_context_data_pointer_t` sounds good. `jerry_context_get_data_pointer` also sounds good.

zherczeg:
I would like to emphasize in the documentation that the number of context data pointers should be low (<= 5), and this is not recommended for implementing modules.

zherczeg:
I would like to emphasize in the documentation that the number of context data pointers should be low (<= 5), and this is not recommended for implementing modules.

jiangzidong:
`jerry_context_data_pointer_t` and `jerry_context_get_data_pointer` are good.

`jerry_context_external_pointer_t` is NOT good because it might be confused with external context #1778

gabrielschulhof:
I would still call it `item` not `pointer` because `_item_t *` is a pointer by virtue of the `*`, and the pointer will be cast to a larger structure whose first member is a `_item_t`.

`_pointer_t *` sounds like a pointer to a pointer, which it's not.

gabrielschulhof:
`jerry_context_data_item_t` and `jerry_context_data_manager_t` doesn't sound bad.

gabrielschulhof:
@zherczeg I wanted to avoid memory fragmentation, wasting space on storing extra pointers, and additional indirection when I defined what I wish to call `jerry_context_data_item_t` as
```C
typedef struct jerry_context_data_item
{
  struct jerry_context_data_item *next;
  jerry_context_data_manager_t *manager_p;
} jerry_context_data_item_t;
```
instead of
```C
typedef struct jerry_context_data_pointer
{
  struct jerry_context_data_pointer *next;
  jerry_context_data_manager_t *manager_p;
  void *data_p;
} jerry_context_data_pointer_t;
```
That's also why I documented that `jerry_context_data_item_t` is intended as the base of larger structures.

gabrielschulhof:
I also realized that we should have a remove(), for completeness' sake.

zherczeg:
I know that, and I prefer your solution. But the API functions returns a pointer to a user data structure which has an expected header. Data item sounds strange for me, it should be simply data or item.

I was thinking about the delete, but it can come in a later patch.

zherczeg:
`jerry_external_data_header_t` and `jerry_get_external_data` sounds good (the later is the current one).

gabrielschulhof:
@zherczeg @jiangzidong had reservations about "`external`". I will call it `jerry_context_data_header_t`.

gabrielschulhof:
@zherczeg and for the same reason `jerry_get_context_data ()`

gabrielschulhof:
Boy, naming is difficult!

zherczeg:
`jerry_context_data_header_t` and `jerry_get_context_data ()` are perfect for me. Yes, naming is difficult, there are meetings where the only topic is discussing public API names. Finding a good, short, descriptive name is a non-trivial task.

gabrielschulhof:
@zherczeg @jiangzidong @akosthekiss *bump*

docs/02.API-REFERENCE.md
```diff
@@ -101,6 +101,39 @@ created by API functions has the error flag set.
...
+larger structures that need to be allocated at runtime.
+
+**Prototype**
+
+```c
+typedef struct jerry_context_data_header
+{
+  struct jerry_context_data_header *next_p; /**< pointer to next context item */
+  const struct jerry_context_data_manager *manager_p; /**< manager responsible for deleting this item */
+} jerry_context_data_header_t;
```
zherczeg: I would emphasize here that the fields of this structure must not be written (preferrably not even read) by the application. It should consider it as an opaque structure (i.e. only its size matters, its content should be ignored).

gabrielschulhof: I just realized we could make this safe by moving the allocation into JerryScript and returning a pointer to just after `manager_p`.

zherczeg: I think the point was that these areas are allocated in one step, not two. Moving into jerry would require an extra pointer, which increases its size to 12 byte, which is 16 byte after alignment. Not a big deal, but still kind of unnecessary. I like the current approach, just we need to emphasize that no application should depend on the fileds of this structure.

gabrielschulhof: Though that would mean calling `init_cb` twice:

```c
typedef struct
{
  int something;
  int something_else;
} my_data_t;

size_t
my_data_init(void *context_data_pp)
{
  if (context_data_p)
  {
    (( my_data_t *) context_data_p)->something = 5;
    (( my_data_t *) context_data_p)->something_else = 12;
  }
  return sizeof (my_data_t);
}
```
and in the `jerry_get_context_data ()`
```c
/* First call to init_cb () determines how much it needs */
size_t size_needed = manager_p->init_cb (NULL);
jerry_context_data_header_t *data_p =
jmem_heap_alloc_block (sizeof (jerry_context_data_header_t) + size_needed);

data_p->next_p = JERRY_CONTEXT (context_data_p);
data_p->manager_p = manager_p;
JERRY_CONTEXT (context_data_p) = data_p;

void *user_data_p = &((char *) data_p)[sizeof (jerry_context_data_header_t)];

/* Second call to init_cb () has a non-NULL pointer as the parameter,
 * so that means initialize the memory at the pointer */
manager_p->init_cb (user_data_p);
...
return user_data_p;
```

gabrielschulhof: ... and then we can make the `jerry_context_data_header_t` structure internal.

gabrielschulhof: @zherczeg The memory structure would still be the same (i.e. header followed by user data), but the knowledge that the user's `void *` is preceded by a `jerry_context_data_header_t` would be internal to JerryScript.

gabrielschulhof: This would also move the alloc()/free() back into JerryScript.

gabrielschulhof: @zherczeg I've added warnings to the docs to not modify the header, but I'll send another PR as an alternative to this one which does not expose the header, but keeps the memory continguous. Let's see which one we like better.

gabrielschulhof: Also, if we add the number of bytes needed as a field of manager_t then we don't need to call init_cb twice.

PR #1816, opened 2017-05-12
Include jerry-ext/handler only in nuttx-stm32f4 target

glistening:
I included jerry-ext/handler only, rather than including all in jerry-ext.
Currently only handler is used in nuttx-stm32f4 repl.
nuttx.bin binary size is reduced from 231576 to 230456 bytes.


robertsipka:
LGTM (informally)

PR #1817, opened 2017-05-12
Increase verbosity when breakpoints not found in HTML (JavaScript) Debugger client

polaroi8d:
In the Python Client if users try to add breakpoint in a wrong index, the client warn them, "Breakpoint not found". Add this defect to the HTML ( JavaScript ) client too.


robertsipka:
LGTM (informally)

PR #1819, opened 2017-05-12
Tracking variables to reduce the memory consumption of scope chain.

zherczeg:
function f() {
      var a = [];
      for (var i = 0; i < 100; i++)
        a[i] = {};
      return function () {}
    }

    var a = [];
    for (var i = 0; i < 100; i++)
      a[i] = f();

After this patch the memory consumption of this code is reduced from 191.3Kb to 7.9Kb.

This change is far from simple, so I expecting issues caught by fuzzer, but this is the way to go if we want to support node.js like codes efficiently.

### DIFF
--- a/jerry-core/parser/js/common.h
+++ b/jerry-core/parser/js/common.h
@@ -67,10 +67,12 @@ typedef enum
 #define LEXER_FLAG_FUNCTION_NAME 0x08
 /* This local identifier is a function argument. */
 #define LEXER_FLAG_FUNCTION_ARGUMENT 0x10
+/* This local identifier is not used in the current context. */
+#define LEXER_FLAG_UNUSED_IDENT 0x20
 /* No space is allocated for this character literal. */
-#define LEXER_FLAG_SOURCE_PTR 0x20
+#define LEXER_FLAG_SOURCE_PTR 0x40
 /* Initialize this variable after the byte code is freed. */
-#define LEXER_FLAG_LATE_INIT 0x40
+#define LEXER_FLAG_LATE_INIT 0x80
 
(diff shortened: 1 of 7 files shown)


akosthekiss:
looks consistent to me (I don't dare to say more)

one question/note: I've played with the motivating example and tweaked it a bit:
```js
function f() {
  var a = [];
  for (var i = 0; i < 100; i++) {
    a[i] = {/*b: 'c'*/};
  }
  return function () { eval("print(JSON.stringify(a[99]));"); }
}

var a = [];
for (var i = 0; i < 100; i++)
  a[i] = f();

a[99]();
print(JSON.stringify(a[99]));
print("done");
```
with a `./tools/build.py --clean --error-messages=on --debug` build, I get:
```
{}
undefined
done
```
as output, but if the commented-out part of the code is enabled, then it becomes
```
Error: ERR_REF_COUNT_LIMIT
```
Master gives the same results, so there is no regression. Just wondering what the root cause for that might be (because the tweak does not seem to be that big).

zherczeg:
You create 100*100=10000 references to strings "b" and "c". The maximum string reference count is 8191 at the moment.

akosthekiss:
Thanks for the explanation. If the loops run to 90 instead of 100 (and the indices are adjusted accordingly), then the output is
```
{"b":"c"}
undefined
done
```
as it should be.

LaszloLango:
@zherczeg what is the performance impact of this patch?

zherczeg:
+0.049% on Sunspider. Actually the new code path is never executed when the SunSpider benchmark runs.

Issue #1774, opened 2017-04-24
target: tizenrt-artik05x: add minimal working port and the guide


glistening:
I would like to introduce a new port `tizenrt-artik05x`.

Tizen RT is a lightweight RTOS-based platform to support low-end IoT devices. The project is under developing in [github tizenrt repo](https://github.com/Samsung/TizenRT). Artik05x is an IoT SoC solution that will be released soon. Please refer to [Artik053](https://github.com/Samsung/TizenRT/blob/master/build/configs/artik053/README.md) for more information.

I will make a PR that contains the minimum porting that can print `Hello world from jerryscript` and a document for building and running.

zherczeg:
+1

akosthekiss:
@glistening #1775 and #1786 have landed some days ago. Do you have plans to submit PRs related to this issue, or has goal been reached and we can close this topic?

glistening:
@akosthekiss I will make another PR for tizenrt. I am thinking of checking and getting recent PRs from NuttX port (e.g., setjmp/longjmp, using jerry-libm instead of system libm, and so on). I will follow your suggestion whether I will make it another issue or consider as a part of minimal port.

akosthekiss:
@glistening it's completely fine to keep this issue open if you still have related plans.

glistening:
Closed since #1822 and #1825 are landed.

PR #1822, opened 2017-05-15
Use external handlers in tizenrt-artik05x target

glistening:
This patch includes followings:
- Use external handlers in jerry-ext.
- Make jerry_main similar to other ports. (including options)
- Merge jerry_port.c into jerry_main.c
  jerry_port is clearly out of jerry-core.

This patch not included following:
- Use jerry-libm, instead of system libm.
  It is already included.
  We choose to use jerry-libm since the result of cube-3d sunspider was inaccurate
  when using NuttX libm.
- Correct setjmp/longjmp implementation.
  I tried to bring the source from NuttX port.
  Unfortunately it does not work in tizenrt-artik05x as it is.
  I will take a closer look the code and make sure it works.


glistening:
Here is my next plan. I will port `setjmp/longjmp` on artik05x (cortex-r4). This architecture seems not support `VSTM` and so on, which were used in setjmp/longjmp on stm32f4 (cortex-m4). After landing this, I would like to close #1774. Then, I will make next step for porting promise on tizenrt-artik05x.

yichoi:
LGTM

robertsipka:
Just a question about the setjmp/longjmp implementation that may be a different from the source of the NuttX port. Is the hard-float mode enabled? The stm32f4 NuttX port's setjmp implementation expects that the hard-float mode is used and stores s16-s31 vfp registers to buffer, pointed with r0 register, and increases the register on size of stored data.

robertsipka:
Sorry, I've just saw the #1823 issue, but the question is still valid, and it is concern both of the vstm and and vldm instructions.

PR #1824, opened 2017-05-15
Initialize an array in test-ext-arg.c

bzsolt:

### DIFF
--- a/tests/unit-ext/test-ext-arg.c
+++ b/tests/unit-ext/test-ext-arg.c
@@ -86,7 +86,7 @@ test_validator1_handler (const jerry_value_t func_obj_val __attribute__((unused)
 
   bool arg1;
   double arg2 = 0.0;
-  char arg3[5];
+  char arg3[5] = { 0 };
   jerry_value_t arg4 = jerry_create_undefined ();
 
   jerryx_arg_t mapping[] =


bzsolt:
In case of some compilers (e.g. gcc-4.9, gcc-5.4) the `TEST_ASSERT (arg3[4] == '\0')` fails at [test-ext-arg.c:118](https://github.com/jerryscript-project/jerryscript/blob/master/tests/unit-ext/test-ext-arg.c#L118) due to uninitialized memory fields.

bzsolt:
I've updated the commit message.

jiangzidong:
@bzsolt I think the root cause is line 118 and line 127
`arg3[4] == '\0'`  should be `arg3[3] == '\0'`. Sorry it's my mistake.

zherczeg:
So we should fix this differently?

jiangzidong:
@zherczeg Yes, I think we can keep the array uninitialized, and just change the code `arg3[4] == '\0'` to `arg3[3] == '\0'`

akosthekiss:
@zherczeg @jiangzidong I think both fixes are needed. Indeed, the asserts checked the wrong index (should check [3] instead of [4]), but leaving the array uninitialized is also not nice. (Actually, exactly because it is uninitialized, there is a (super-low) chance that the garbage in the memory would be exactly 'abc' and the test would pass even if the transform function behaved incorrectly.)

bzsolt:
As I can see, the `jerryx_arg_transform_this_and_args` doesn't terminate the 'abc' with `\0` so this assert ` TEST_ASSERT (arg3[0] == 'a' && arg3[1] == 'b' && arg3[2] == 'c' && arg3[3] == '\0');` still fails.

jiangzidong:
@bzsolt Oh, I just found that. Here comes another problem: should we add `\0` in jerryx_arg_string? I think we should and I will submit a patch, what do you think? @akosthekiss @martijnthe @zherczeg

PR #1826, opened 2017-05-16
CMakeLists maintenance

akosthekiss:
Removed various superfluous declarations, and rewrote some parts to
be nicer or more consistent.

### DIFF
--- a/jerry-core/debugger/jerry-debugger-ws.c
+++ b/jerry-core/debugger/jerry-debugger-ws.c
@@ -13,16 +13,16 @@
  * limitations under the License.
  */
 
+#include "jcontext.h"
+#include "jerry-debugger.h"
+#include "jerryscript-port.h"
+
 #ifdef JERRY_DEBUGGER
 
 #include <arpa/inet.h>
 #include <errno.h>
 #include <fcntl.h>
 #include <unistd.h>
-
-#include "jcontext.h"
-#include "jerry-debugger.h"
-#include "jerryscript-port.h"
 
 /**
  * Debugger socket communication port.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -19,10 +19,25 @@ project (Jerry C ASM)
 set(PLATFORM "${CMAKE_SYSTEM_NAME}")
 string(TOUPPER "${PLATFORM}" PLATFORM)
 
-# Remove rdynamic option
-set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS )
+# Determining compiler
+if(CMAKE_C_COMPILER_ID MATCHES "GNU")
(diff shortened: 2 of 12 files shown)


jerry-core/debugger/jerry-debugger-ws.c
```diff
@@ -13,8 +13,6 @@
  * limitations under the License.
  */
 
-#ifdef JERRY_DEBUGGER
```
zherczeg: What is the reason for this change? #include <arpa/inet.h> will surely fail with jerry_libc, and we need to think about all-in-one builds.

akosthekiss: The reason is "transitive". CMakeLists added all sources of jerry-core to `SOURCE_CORE_FILES` unconditionally but the sources of the debugger. And the reason for that, I guess, was that the sources were completely guarded by `JERRY_DEBUGGER`. However, if the guard was off, the compilation unit was empty, which was a warning/error for C99. I wanted to treat all sources equal: it's "nicer" but also easier to maintain. The common workaround for the empty compilation unit problem is to have at least a typedef somewhere, which is usually ensured by having the includes outside the guard. The jerry-core/ecma subdir uses the same approach at various places, too, e.g., ecma-builtin-promise.c. (We don't add those sources to the list of files conditionally either.)

The strange thing is that Travis CI tests jerry-libc builds and it does _not_ fail... (Although I tend to agree with you that some of the headers here should not be available outside the guard.)

zherczeg: The problam is that not everybody use the cmake build system. Some people just use the files (all-in-one build). E.g. esp8266.

akosthekiss: I see no problem with the non-cmake use case. On the contrary, then those who don't use cmake, don't have to keep in mind which files they may add to the list of sources and which they must not. (Right now, if someone just gathers the list of _all_ files, including the debugging-related ones, and tries to build them without `JERRY_DEBUGGER` defined - that's the non-cmake use case -, then s/he will get an error. With this patch, that error goes away.)

zherczeg: It cannot. You put JERRY_DEBUGGER after #include <arpa/inet.h>, and if that system has no arpa/inet.h (which is not needed if you are not using the debugger), you will get a compile error.

akosthekiss: Fair point. It's strange that no testing caught the issue (or perhaps not that strange, because we have CI for Linux and OSX only). I've moved the platform-dependent includes within the guard.

akosthekiss:
PR rebased and updated according to comments.

PR #1828, opened 2017-05-17
Add pending breakpoints feature to HTML (JavaScript) Debugger client

polaroi8d:


jerry-debugger/jerry-client-ws.html
```diff
@@ -1207,14 +1267,21 @@
   {
     case "b":
     case "break":
-      debuggerObj.setBreakpoint(args[2]);
+      debuggerObj.setBreakpoint(args[2], false);
```
zherczeg: I think "undefined" is same as "false", so we might not even need to pass anything here.

polaroi8d: If we use "false" IMHO its more readable and consistens.

PR #1833, opened 2017-05-18
Lazily create a linked list of context items (alloc()/free() done inside JerryScript)

gabrielschulhof:
This approach has the benefit that it does not require any *a priori*
initialization, and that each context pointer is identified by the way
in which it was created. Additionally, retrieving the context pointer
now requires that the entity responsible for creating/destroying it
(the manager) be given. Since managers are stored in global static
const structures, they should not normally be visible across source
files, and thus there should be no danger that a context item will be
retrieved by the wrong manager and thus cast into the wrong data type.

Since the items are stored in a linked list, their number will be
limited to exactly as many as are needed for a given context, with the
caveat that storing too many on a context will cause slow retrieval.

Thanks @mhdawson for the idea!

This is an alternative that is similar to #1815, but where the header which links context items is not exposed, while keeping the memory structure equally contiguous.


docs/02.API-REFERENCE.md
```diff
@@ -101,6 +101,23 @@ created by API functions has the error flag set.
...
+## jerry_context_data_manager_t
+
+**Summary**
+
+Structure that defines how a user context item will be initialized and deinitialized.
+
+**Prototype**
+
+```c
+typedef struct jerry_context_data_manager
```
LaszloLango: just `typedef struct`

docs/02.API-REFERENCE.md
```diff
@@ -397,37 +334,80 @@ jerry_cleanup (void);
...
 
-## jerry_get_user_context
+## jerry_get_context_data
 
 **Summary**
 
-Retrieve the pointer stored within the current context.
+Retrieve a pointer to the item stored within the current context by the given manager.
+
+**N.B.:** Since internally the pointer to a manager's context data item is linked to the next such pointer in a linked
```
LaszloLango: What is "N.B."? :)

jerry-core/api/jerry.c
```diff
@@ -201,23 +192,39 @@ jerry_cleanup (void)
...
 void *
-jerry_get_user_context (void)
+jerry_get_context_data (const jerry_context_data_manager_t *manager_p)
 {
-  return JERRY_CONTEXT (user_context_p);
-} /* jerry_get_user_context */
+  void *ret = NULL;
+  jerry_context_data_header_t *item_p;
+
+  for (item_p = JERRY_CONTEXT (context_data_p); item_p; item_p = item_p->next_p)
```
LaszloLango: style issue in the condition: `item_p != NULL`

docs/02.API-REFERENCE.md
gabrielschulhof: https://en.wikipedia.org/wiki/Nota_bene

gabrielschulhof: I noticed that the trend is to use `*Note*:` and then indent the rest, so I'll move to that.

gabrielschulhof:
@LaszloLango thanks for reviewing! I've addressed your comments.

jerry-core/jcontext/jcontext.h
```diff
@@ -35,6 +36,18 @@
...
+ * User context item
+ */
+typedef struct jerry_context_data_header
+{
+  struct jerry_context_data_header *next_p; /**< pointer to next context item */
+  const jerry_context_data_manager_t *manager_p; /**< manager responsible for deleting this item */
+} jerry_context_data_header_t;
+
+#define JERRY_CONTEXT_DATA_HEADER_USER_DATA(item_p) \
+  (&(((char *) item_p)[sizeof (jerry_context_data_header_t)]))
```
zherczeg: What about `(uint8_t *) (item_p + 1)` ? Much shorter.

jerry-core/api/jerry.c
```diff
@@ -201,23 +192,39 @@ jerry_cleanup (void)
...
+
+  for (item_p = JERRY_CONTEXT (context_data_p); item_p != NULL; item_p = item_p->next_p)
+  {
+    if (item_p->manager_p == manager_p)
+    {
+      return JERRY_CONTEXT_DATA_HEADER_USER_DATA (item_p);
+    }
+  }
+
+  item_p = jmem_heap_alloc_block (sizeof (jerry_context_data_header_t) + manager_p->bytes_needed);
```
zherczeg: I think this is the critical part here: Is it a good idea to reduce JerryScript space with this? Bigger contexts could be a problem. What do people think about this?

zherczeg: @akosthekiss @janjongboom @LaszloLango @martijnthe

jiangzidong: @zherczeg I am not fully understand. Do you worry that the `bytes_needed` may too large, so it will occupy too much jerry heap?

gabrielschulhof: @zherczeg Aaaah, I see! You wanted to expose the header to let others do the allocation. Still, at the project level the dev can juggle the relative size of the JerryScript heap wrt. whatever other heap they're using. After all, the total amount of memory on any given device is fixed.

Nevertheless, #1815 is still around so we can go with that one.

jerry-core/jcontext/jcontext.h
gabrielschulhof: I'd go with ((jerry_context_data_header_t *) item_p) to make sure that the "+1" has the right semantics.

gabrielschulhof: OTOH it's internal, so the extra cast is probably unnecessary.

jerry-core/api/jerry.c
gabrielschulhof: @jiangzidong I think that's the concern. In #1815 allocation is done outside JerryScript so it can potentially be done on a different heap where there's more room. The tradeoff is that we expose the header or we waste a word with an extra indirection.

jiangzidong: I see, thanks. :)

gabrielschulhof: @zherczeg perhaps wasting that word on the indirection is not so bad, since then the size of `bytes_needed` is always the same and we still keep the benefit of not exposing the header.

gabrielschulhof: Yet another alternative is to re-introduce the *a priori* aspect and initialize each context with a fixed number of extra slots:
```c
void jerry_init (jerry_init_flag_t flags, size_t context_slot_count);
```
Then we could have an array of
```c
struct {
  jerry_context_manager_t *manager_p;
  void *user_data_p
};
```
which we could allocate contiguously with the rest of the context if we stick it to the end of the context, thus saving a further word.

This is kinda limiting though because the value of `context_slot_count` will need to be known ahead of time, and there might be some temptation to make that value large in anticipation of future complex code.

zherczeg: I have no strong preference, but I would like to hear others opinion. If they prefer this one, it is fine by me.

martijnthe: Fine by me to use `jmem_heap_alloc_block` for this.

martijnthe:
@LaszloLango could you re-review? Thanks!

LaszloLango:
@gabrielschulhof could you open an issue for this task? You have two open PRs doing the same. Please summarize them in a GitHub issue. Which one do you prefer and why? BTW, the code looks good in general.

zherczeg:
I think he prefers this one. If you are ok with jerry allocated contexts, it is fine by me as well.

jerry-core/include/jerryscript-core.h
```diff
@@ -195,15 +195,16 @@ typedef jerry_value_t (*jerry_vm_exec_stop_callback_t) (void *user_p);
...
-typedef void *(*jerry_user_context_init_t) (void);
 
 /**
- * Function type for user context deallocation
+ * User context item manager
  */
-typedef void (*jerry_user_context_deinit_t) (void *user_context_p);
+typedef struct
+{
+  void (*init_cb) (void *); /**< callback responsible for initializing a context item */
```
zherczeg: I just realized that in this variant, the context receives the allocated buffer, but without any context. If the data is only cleard, this is enough, but if anything more is needed, this is not enough. If we want to limit the init to clear the data (a valid use case), we can do it and remove the init function.

jerry-core/api/jerry.c
```diff
@@ -167,29 +167,20 @@ jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
...
-
-/**
  * Terminate Jerry engine
  */
 void
 jerry_cleanup (void)
 {
   jerry_assert_api_available ();
 
+  for (jerry_context_data_header_t *this_p = JERRY_CONTEXT (context_data_p), *next_p = NULL; this_p; this_p = next_p)
```
LaszloLango: Please write like this
```c
for (jerry_context_data_header_t *this_p = JERRY_CONTEXT (context_data_p), *next_p = NULL;
     this_p != NULL;
     this_p = next_p)
```

jerry-core/include/jerryscript-core.h
martijnthe: I would like to keep it.

gabrielschulhof: Me too.

gabrielschulhof:
@LaszloLango I prefer this PR because it does not expose the header and it does not require that users perform memory allocation, even though there's the trade-off that it may burden the Jerry heap further.

LaszloLango:
LGTM

jerry-core/include/jerryscript-core.h
zherczeg: I think we should add a user pointer (similar to the other patch) which allows initializing the user data. Without that the init function has very limited capabilities.

gabrielschulhof: @zherczeg if you mean that it should be `void (*init_cb)(void *newly_allocated_p, void *hint_p)` then I don't think that's necessary, because if any user ever needs to use the context for two different reasons, they can just store a default reason in the context data during `init_cb ()`, and change the reason if it's wrong:
```c
/* this calls init_cb () */
my_data_t *my_data_p = jerry_get_context_data (&my_manager);
if (my_data_p->init_hint != MY_INIT_HINT)
{
  reinit_my_data(my_data_p, MY_INIT_HINT);
}
```

Or, they can have two different managers for different kinds of context data. That's what the [unit test](https://github.com/jerryscript-project/jerryscript/pull/1833/files#diff-a91107b501aafed86478d129d0a4d2e3) does.

I tend to think of context data as global static data that cannot be stored in a global static variable because it's different for each Jerry context, and the same compilation unit has to run in two different contexts. As such, I don't believe there should be any more degrees of freedom.

Now, given this analogy, global static variables can be initialized to non-trivial values without any hints, and in our case, the `init_cb` has full access to the JS world of its context, so it can definitely make decisions about how it should initialize itself.

Another implication of adding the hint is that it probably needs to be stored along with the `manager_p` in the header and then passed to `deinit_cb`.

At that point you're probably better off with having two different managers.

jiangzidong: It already has arg `void *` which points to user_data

zherczeg: Ok. However I think in most of the time people will simply zero the buffer, and this could be the default behaviour: pass a zeroed buffer to the init callback. If no init callback is provided (NULL), just zero the buffer.

gabrielschulhof: @zherczeg cool! Can do.

gabrielschulhof:
@LaszloLango @jiangzidong @martijnthe @zherczeg I've made the `init_cb` optional. Please have another look!

docs/02.API-REFERENCE.md
```diff
@@ -101,6 +101,24 @@ created by API functions has the error flag set.
...
+
+Structure that defines how a user context item will be initialized and deinitialized. The default initialization, if
+`init_cb` is NULL, is to zero out the memory.
+
+**Prototype**
+
+```c
+typedef struct
+{
+  void (*init_cb) (void *); /**< callback responsible for initializing a context item, or NULL to zero out the memory */
```
zherczeg: The memory should be always zeroed.

jerry-core/api/jerry.c
```diff
@@ -201,23 +194,43 @@ jerry_cleanup (void)
...
+    }
+  }
+
+  item_p = jmem_heap_alloc_block (sizeof (jerry_context_data_header_t) + manager_p->bytes_needed);
+  item_p->manager_p = manager_p;
+  item_p->next_p = JERRY_CONTEXT (context_data_p);
+  JERRY_CONTEXT (context_data_p) = item_p;
+  ret = JERRY_CONTEXT_DATA_HEADER_USER_DATA (item_p);
+  memset (ret, 0, manager_p->bytes_needed);
+  if (manager_p->init_cb)
```
zherczeg: Please add a newline before memset.

tests/unit-core/test-context-data.c
```diff
@@ -0,0 +1,92 @@
...
+{
+  TEST_INIT ();
+
+  jerry_init (JERRY_INIT_EMPTY);
+
+  TEST_ASSERT (!strcmp (*((const char **) jerry_get_context_data (&manager1)), "item1"));
+  TEST_ASSERT (!strcmp (*((const char **) jerry_get_context_data (&manager2)), "item2"));
+
+  TEST_ASSERT (test_context_data1_new_called);
+  TEST_ASSERT (test_context_data2_new_called);
```
zherczeg: Add these here:

    TEST_ASSERT (!test_context_data1_free_called);
    TEST_ASSERT (!test_context_data2_free_called);

docs/02.API-REFERENCE.md
gabrielschulhof: It actually is always zeroed, but that's not how I documented it. I'll re-word this to reflect that.

tests/unit-core/test-context-data.c
gabrielschulhof: @zherczeg done.

LaszloLango:
still LGTM

martijnthe:
@zherczeg @jiangzidong could you give your final blessing?
I'm dying to start using this 😁 thanks

jiangzidong:
LGTM

martijnthe:
🎉

gabrielschulhof:
Thanks all for reviewing!

PR #1835, opened 2017-05-19
Fix a typo in a variable name

DanielBallaSZTE:
There was a typo in a variable name causing the build to fail when mem-stats was turned on.


polaroi8d:
LGTM (informally)

PR #1841, opened 2017-05-19
Print exception hint in the debugger client when an exception is thrown.

zherczeg:


jerry-core/debugger/jerry-debugger.c
```diff
@@ -715,4 +717,151 @@ jerry_debugger_send_memstats (void)
...
+} /* jerry_debugger_exception_object_to_string */
+
+/**
+ * Send string representation of exception to the client.
+ *
+ * @return true - if the data sent successfully to the debugger client,
+ *         false - otherwise
+ */
+bool
+jerry_debugger_send_exception_string (ecma_value_t exception_value)
```
LaszloLango: ditto

PR #1842, opened 2017-05-22
Update mbedos5 target to ignore jerry-port-defaul

fbmrk:
Update mbedos5 target to use their own jerry-port implementations.


zherczeg:
Could you squash these commits?

PR #1848, opened 2017-05-23
[tizenrt-artik05x] Set log level to JERRY_LOG_LEVEL_DEBUG when using mem-stats or show-opcodes

robertsipka:
It provides detailed output in case of enabled memory statistics and  parser byte-code dumps.
@glistening : Please look at it, thanks!


glistening:
@LaszloLango, @robertsipka I've checked. It works.

yichoi:
LGTM

PR #1849, opened 2017-05-24
Improve instance management in JerryScript.

zherczeg:


zherczeg:
@jiangzidong

jerry-core/jcontext/jcontext.h
```diff
@@ -121,52 +121,65 @@ typedef struct
...
+#endif /* !CONFIG_ECMA_LCACHE_DISABLE */
+
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+
+#ifndef JERRY_GET_CURRENT_INSTANCE
+
+/**
+ * Default function if JERRY_GET_CURRENT_INSTANCE is not defined.
+ */
+#define JERRY_GET_CURRENT_INSTANCE() (jerry_port_get_current_instance ())
```
jiangzidong: Why do we need the JERRY_GET_CURRENT_INSTANCE instead of directly calling jerry_port_get_current_instance?

jerry-core/include/jerryscript-port.h
```diff
@@ -127,12 +127,12 @@ bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p);
...
 /**
- * Get the current instance, which contains the current context, heap and other infomation.
- * Each port should provide its own implementation of this interface.
+ * Get the current instance which contains the current context, heap and other
+ * structures. Each port should provide its own implementation of this interface.
  *
  *Note:
- *    This port function will be called automatically by jerry-core
- *    wnen JERRY_ENABLE_EXTERNAL_CONTEXT is defined. If not, this function will never be called.
+ *    This port function is called by jerry-core wnen JERRY_ENABLE_EXTERNAL_CONTEXT
```
jiangzidong: "when", its my typo.

jerry-core/jcontext/jcontext.h
```diff
@@ -121,52 +121,65 @@ typedef struct
...
+#define JERRY_CONTEXT(field) (JERRY_GET_CURRENT_INSTANCE ()->context.field)
 
 #ifndef JERRY_SYSTEM_ALLOCATOR
 
 static inline jmem_heap_t * __attr_always_inline___
 jerry_context_get_current_heap (void)
 {
-  return (jmem_heap_t *) (jerry_port_get_current_instance ()->heap_p);
+  uint8_t *instance_p = (uint8_t *) JERRY_GET_CURRENT_INSTANCE ();
+  return (jmem_heap_t *) (instance_p + ALIGNED_SIZE_OF_JERRY_INSTANCE);
```
jiangzidong: Can it make sure the jmem_heap_t* address is aligned with JMEM_ALIGNMENT?

I think the heap size is multiple of JMEM_ALIGNMENT, the size of jerry_instance_t is multiple of JMEM_ALIGNMENT, but the address of the heap may not.  Or do I miss something?

jerry-core/jcontext/jcontext.h
zherczeg: Some ports might want to use something else than that function call (e.g. a static variable or a thread local variable).

jerry-core/jcontext/jcontext.h
zherczeg: Right, this is a problem when allocating memory. Bad news.

jerry-core/jcontext/jcontext.h
jiangzidong: In this way, jerry_port_get_current_instance() can return the static variable or thread local variable, right?

zherczeg: You can, but calling a function is expensive if the compiler cannot inline it. I think we should not restrict people here, although the default is using `jerry_port_get_current_instance` of course.

jerry-core/jcontext/jcontext.h
zherczeg: I have been thinking about it. The easiest way to fix it is:

a) passing an alignment requirement to the malloc callback
b) align the returned instance_p and add a function which tells the padding
c) keep the current one, where have to do an dereference when accessing the heap

Probably a is the easiest, since we could simply use posix_memalign. Do you have a better idea?

jiangzidong: a) is not good to me, because 1) pass the burden to user, 2) align-requirement many be changed later, it is better to be transparent to users.

What about the original way, which is similar to c, 
* alloc more (JMEM_ALIGNMENT - 1) bytes to make sure the buffer is big enough that heap can be aligned up from any address.
* set the new start address of heap is 
```
 uint8_t *unaligned_heap_p = ...
 uintptr_t alignment_mask = ~(uintptr_t) (JMEM_ALIGNMENT - 1);
 new_heap_addr = (uint8_t *) (((uintptr_t) unaligned_heap_p + JMEM_ALIGNMENT - 1) & alignment_mask);
```

zherczeg:
Patch updated.

jerry-core/api/jerry.c
```diff
@@ -2260,37 +2260,48 @@ jerry_create_instance (uint32_t heap_size, /**< the size of heap */
...
+
+  uint8_t *byte_p = (uint8_t *) instance_p;
+  byte_p += JERRY_ALIGNUP (sizeof (jerry_instance_t), JMEM_ALIGNMENT);
 
 #ifndef JERRY_SYSTEM_ALLOCATOR
-  uint8_t *unaligned_heap_p = instance_p->lcache_p + lcache_size;
-  uintptr_t alignment_mask = ~(uintptr_t) (JMEM_ALIGNMENT - 1);
-  instance_p->heap_p = (uint8_t *) (((uintptr_t) unaligned_heap_p + JMEM_ALIGNMENT - 1) & alignment_mask);
+  instance_p->heap_p = (jmem_heap_t *) byte_p;
+  instance_p->heap_size = heap_size;
```
jiangzidong: I don't think it solve the alignment problem.

if instance_p is not multiple of 8bytes, `JERRY_ALIGNUP (sizeof (jerry_instance_t), JMEM_ALIGNMENT)` is multiple of 8bytes, then `heap_p = instance_p + JERRY_ALIGNUP (sizeof (jerry_instance_t), JMEM_ALIGNMENT)` is not multiple of 8 bytes

zherczeg: True. Ok I change it.

zherczeg:
Next attempt is uploaded.

jerry-core/include/jerryscript-port.h
LaszloLango: Not fixed yet.

zherczeg:
Typo fixed.

jiangzidong:
still LGTM

LaszloLango:
LGTM

PR #1850, opened 2017-05-24
Python debugger source & display command

DanielBallaSZTE:
Reworked source(src) command to display a given range of the code, defaults to 3.
Added display command which does the same as the source, except it displays the source code everytime a breakpoint is hit.

### DIFF
--- a/jerry-debugger/jerry-client-ws.py
+++ b/jerry-debugger/jerry-client-ws.py
@@ -117,7 +117,7 @@ class JerryFunction(object):
     def __init__(self, is_func, byte_code_cp, source, source_name, line, column, name, lines, offsets):
         self.is_func = is_func
         self.byte_code_cp = byte_code_cp
-        self.source = source
+        self.source = re.split("\r\n|[\r\n]", source)
         self.source_name = source_name
         self.name = name
         self.lines = {}
@@ -312,11 +312,29 @@ class DebuggerPrompt(Cmd):
 
     def do_src(self, args):
         """ Get current source code """
-        if self.disable_args(args):
-            return
+        if args:
+            line_num = src_check_args(args)
+            if line_num >= 0:
+                print_source(self.debugger, line_num)
+            elif line_num == 0:
+                print_source(self.debugger, self.debugger.default_viewrange)
+            else:
+                return
 
-        if self.debugger.last_breakpoint_hit:
-            print(self.debugger.last_breakpoint_hit.function.source)
+    do_source = do_src
+
+    def do_display(self, args):
+        """ Toggle source code display after breakpoints """
+        if args:
+            line_num = src_check_args(args)
+            if line_num >= 0:
+                self.debugger.display = line_num
+            else:
+                return
+
+        else:
+            print("Non-negative integer number expected, 0 turns off this function")
+            return
 
     def do_dump(self, args):
         """ Dump all of the debugger data """
@@ -395,7 +413,7 @@ class DebuggerPrompt(Cmd):
 
     def do_memstats(self, args):
         """ Memory statistics """
-        self.exec_command(args, JERRY_DEBUGGER_MEMSTATS);
+        self.exec_command(args, JERRY_DEBUGGER_MEMSTATS)
         return
 
     do_ms = do_memstats
@@ -448,6 +466,8 @@ class JerryDebugger(object):
         self.active_breakpoint_list = {}
         self.pending_breakpoint_list = []
         self.line_list = Multimap()
+        self.display = 0
+        self.default_viewrange = 3
         self.client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.client_socket.connect((self.host, self.port))
 
@@ -729,6 +749,42 @@ def parse_source(debugger, data):
         logging.debug("No pending breakpoints")
 
 
+def src_check_args(args):
+    try:
+        line_num = int(args)
+        if line_num < 0:
+            print("Error: Non-negative integer number expected")
+            return -1
+
+        return line_num
+    except ValueError as val_errno:
+        print("Error: Non-negative integer number expected: %s" % (val_errno))
+        return -1
+
+
+def print_source(debugger, line_num):
+    last_bp = debugger.last_breakpoint_hit
+    if not last_bp:
+        return
+
+    lines = last_bp.function.source
+    if last_bp.function.source_name:
+        print("Source: %s" % (last_bp.function.source_name))
+
+    if line_num == 0:
+        start = 0
+        end = len(last_bp.function.source) - 1
+    else:
+        start = max(last_bp.line - line_num, 0)
+        end = min(last_bp.line + line_num-1, len(last_bp.function.source)-1)
+
+    for i in range(start, end):
+        if i == last_bp.line - 1:
+            print("%4d > %s" % (i + 1, lines[i]))
+        else:
+            print("%4d   %s" % (i + 1, lines[i]))
+
+
 def release_function(debugger, data):
     byte_code_cp = struct.unpack(debugger.byte_order + debugger.cp_format,
                                  data[3: 3 + debugger.cp_size])[0]
@@ -881,6 +937,8 @@ def main():
                 breakpoint_info += " breakpoint:%d" % (breakpoint[0].active_index)
 
             print("Stopped %s %s" % (breakpoint_info, breakpoint[0]))
+            if debugger.display:
+                print_source(prompt.debugger, debugger.display)
 
             prompt.cmdloop()
             if prompt.quit:
--- a/tests/debugger/do_display.cmd
+++ b/tests/debugger/do_display.cmd
@@ -1,0 +1,13 @@
+b a
+b b
+b c
+b d
+display
+c
+display 2
+c
+display 5435
+c
+display 0
+c
+c
--- a/tests/debugger/do_display.expected
+++ b/tests/debugger/do_display.expected
@@ -1,0 +1,53 @@
+Connecting to: localhost:5001
+Stopped at tests/debugger/do_display.js:20
+(jerry-debugger) b a
+Breakpoint 1 at tests/debugger/do_display.js:15 (in a() at line:15, col:1)
+(jerry-debugger) b b
+Breakpoint 2 at tests/debugger/do_display.js:16 (in b() at line:16, col:1)
+(jerry-debugger) b c
+Breakpoint 3 at tests/debugger/do_display.js:17 (in c() at line:17, col:1)
+(jerry-debugger) b d
+Breakpoint 4 at tests/debugger/do_display.js:18 (in d() at line:18, col:1)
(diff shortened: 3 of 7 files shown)


jerry-debugger/jerry-client-ws.py
```diff
@@ -728,6 +759,35 @@ def parse_source(debugger, data):
     else:
         logging.debug("No pending breakpoints")
 
+def printSource(self, line_num):
+    last_bp = self.debugger.last_breakpoint_hit
+    lines = last_bp.function.source.split('\n',last_bp.function.source.count('\n') - 1)
```
zherczeg: Even if the source is an empty string, the array should have at least one value. if "".split("\n") return with an array which contains an empty string: `[ "" ]` that is ok. If it returns with an empty array `[]` than we should manually add a "" to this array.

And one more question: why do we need a second argument to split?

tests/debugger/do_src.expected
```diff
@@ -9,14 +9,16 @@ Stopped at tests/debugger/do_src.js:20
 (jerry-debugger) s
 Stopped at <unknown>:1
 (jerry-debugger) src
-f = function f() {
+
+1   > f = function f() {
```
zherczeg: Why there is an empty line before this? We also need tests for src with argument and the new display commands.

jerry-debugger/jerry-client-ws.py
```diff
@@ -117,7 +117,7 @@ class JerryFunction(object):
     def __init__(self, is_func, byte_code_cp, source, source_name, line, column, name, lines, offsets):
         self.is_func = is_func
         self.byte_code_cp = byte_code_cp
-        self.source = source
+        self.source = source.split('\n')
```
zherczeg: Is `\n` enough? Perhaps `re.split("\r\n|[\r\n]", source)` would be better.

jerry-debugger/jerry-client-ws.py
```diff
@@ -134,6 +134,9 @@ def __init__(self, is_func, byte_code_cp, source, source_name, line, column, nam
             self.lines[line] = breakpoint
             self.offsets[offset] = breakpoint
 
+        for i in range(len(self.source)):
+            self.source[i] = "%d%4s %s" % (i + 1, " ", self.source[i])
```
zherczeg: We don't need this preparation :) The line index is available at use.

jerry-debugger/jerry-client-ws.py
```diff
@@ -728,6 +759,24 @@ def parse_source(debugger, data):
     else:
         logging.debug("No pending breakpoints")
 
+def printSource(self, line_num):
+    last_bp = self.debugger.last_breakpoint_hit
+    lines = last_bp.function.source
+    if last_bp:
+        print("Source: %s" % (last_bp.function.source_name))
```
zherczeg: please check whether `last_bp.function.source_name` is available. There is no point printing empty lines.

jerry-debugger/jerry-client-ws.py
```diff
@@ -728,6 +759,24 @@ def parse_source(debugger, data):
...
+            end = len(last_bp.function.source) - 1
+        else:
+            start = max(last_bp.line - line_num-1, 0)
+            end = min(last_bp.line + line_num, len(last_bp.function.source)-1)
+
+        for i in range(start, end):
+            if i == last_bp.line - 1:
+                print("%s" % lines[i][:3], ">", lines[i][6:])
+            else:
+                print(lines[i])
```
zherczeg: This is too complex. Just remove the for above, and format the line here.

DanielBallaSZTE:
Added those issued changes.

tests/debugger/do_display.cmd
```diff
@@ -0,0 +1,13 @@
+b a
+b b
+b c
+b d
+display
```
zherczeg: I think display without argument is pointless, I would throw an error.

jerry-debugger/jerry-client-ws.py
```diff
@@ -312,11 +312,38 @@ def do_backtrace(self, args):
...
+
+    do_source = do_src
+
+    def do_display(self, args):
+        """ Toggle source code display after breakpoints """
+        if args:
+            try:
+                line_num = int(args)
+                if line_num < 0:
+                    print("Error: Positive integer number expected")
```
zherczeg: Non-negative integer number expected (0 disables this command).

jerry-debugger/jerry-client-ws.py
```diff
@@ -728,6 +757,27 @@ def parse_source(debugger, data):
     else:
         logging.debug("No pending breakpoints")
 
+def printSource(self, line_num):
+    last_bp = self.debugger.last_breakpoint_hit
+    lines = last_bp.function.source
+    if last_bp:
```
zherczeg: you are using last_bp before this check. Please move `lines = last_bp.function.source` after this check.

tests/debugger/do_display.expected
```diff
@@ -0,0 +1,53 @@
...
+(jerry-debugger) display
+Error: Display should be called with an argument
+(jerry-debugger) c
+Stopped at breakpoint:1 tests/debugger/do_display.js:15 (in a() at line:15, col:1)
+(jerry-debugger) display 2
+(jerry-debugger) c
+Stopped at breakpoint:2 tests/debugger/do_display.js:16 (in b() at line:16, col:1)
+Source: tests/debugger/do_display.js
+15   function a() { print("hi"); }
+16  >function b() { print("welcome"); }
```
zherczeg: I would put a space after >

jerry-debugger/jerry-client-ws.py
zherczeg: Btw we could do an early return here:

    if not last_bp:
      return

This way we don't need to indent the lines below.

jerry-debugger/jerry-client-ws.py
```diff
@@ -728,6 +757,27 @@ def parse_source(debugger, data):
...
+            start = 0
+            end = len(last_bp.function.source) - 1
+        else:
+            start = max(last_bp.line - line_num, 0)
+            end = min(last_bp.line + line_num-1, len(last_bp.function.source)-1)
+
+        for i in range(start, end):
+            spaces = (4-len(str(i+1)))*" "
+            if i == last_bp.line - 1:
+                print("%d%s>%s" % (i + 1, spaces, lines[i]))
```
zherczeg: Space after >

zherczeg: What about:

    if i == last_bp.line - 1:
      string = " > "
    else
      string = "   "
    print("%4d%s%s" % (i + 1, string, lines[i]))

This way you don't need spaces

jerry-debugger/jerry-client-ws.py
```diff
@@ -728,6 +757,27 @@ def parse_source(debugger, data):
     else:
         logging.debug("No pending breakpoints")
 
+def printSource(self, line_num):
```
galpeter: Naming style: not camel case but underscores. Thus the name should be: `print_source`.
Additionally, we should not accept self (as we don't know what is it) but instead the debugger itself as the first argument.

jerry-debugger/jerry-client-ws.py
```diff
@@ -312,11 +312,38 @@ def do_backtrace(self, args):
...
-        if self.disable_args(args):
+        if args:
+            try:
+                line_num = int(args)
+                if line_num < 0:
+                    print("Error: Positive integer number expected")
+                    return
+            except ValueError as val_errno:
+                print("Error: Positive integer number expected, %s" % (val_errno))
+                return
```
galpeter: Maybe we could extract the convert logic to a helper method, so then we would not need to duplicate the code here and in the next method.

jerry-debugger/jerry-client-ws.py
```diff
@@ -729,6 +737,46 @@ def parse_source(debugger, data):
...
+    elif is_src_cmd:
+        return debugger.default_viewrange
+    else:
+        print("Display should be called with an argument")
+        return -1
+
+
+def print_source(debugger, line_num):
+    last_bp = debugger.last_breakpoint_hit
+    if last_bp:
```
zherczeg: One of my comment is missing:

    if not last_bp:
      return

And remove the indentation form the code after.

tests/debugger/do_src.expected
```diff
@@ -8,15 +8,13 @@ Stopped at breakpoint:1 tests/debugger/do_src.js:16 (in f() at line:15, col:1)
 Stopped at tests/debugger/do_src.js:20
 (jerry-debugger) s
 Stopped at <unknown>:1
-(jerry-debugger) src
-f = function f() {
-print('F2') }
+(jerry-debugger) source 0
```
zherczeg: Is `source 0` prints the current line??? Isn't `source 1` should do this?

jerry-debugger/jerry-client-ws.py
```diff
@@ -729,6 +737,48 @@ def parse_source(debugger, data):
...
+        end = len(last_bp.function.source) - 1
+    else:
+        start = max(last_bp.line - line_num, 0)
+        end = min(last_bp.line + line_num-1, len(last_bp.function.source)-1)
+
+    for i in range(start, end):
+        if i == last_bp.line - 1:
+            print("%4d > %s" % (i + 1, lines[i]))
+        else:
+            print("%4d%3s%s" % (i + 1, " ",lines[i]))
```
zherczeg: print("%4d   %s" % (i + 1, lines[i]))

Is good enough :)

jerry-debugger/jerry-client-ws.py
```diff
@@ -729,6 +737,48 @@ def parse_source(debugger, data):
         logging.debug("No pending breakpoints")
 
 
+def src_check_args(debugger, is_src_cmd, args):
```
galpeter: We had a bit of miscommunication, sorry about that. What I meant originally that the try-except block could be extracted to it's own method not the whole thing.

jerry-debugger/jerry-client-ws.py
```diff
@@ -729,6 +737,48 @@ def parse_source(debugger, data):
...
+    if not last_bp:
+        return
+
+    lines = last_bp.function.source
+    if last_bp.function.source_name:
+        print("Source: %s" % (last_bp.function.source_name))
+
+    if line_num == 0:
+        start = 0
+        end = len(last_bp.function.source) - 1
```
galpeter: Do we really need to subtract 1? the range method below will generate numbers in the `[start, end)` range, that is the `end` number will not be reached

DanielBallaSZTE: Yes we do, if it is not added there would be an empty line printed at the end of the source command.

galpeter: I see. ok.

galpeter:
I don't have any other comments, lgtm

PR #1853, opened 2017-05-25
target: mbedos5: Update to mbed OS 5.4.5

janjongboom:
Also ignore the default jerry-port - see also #1847

Tested programs:

* mbed-js-example on K64F
* mbed-js-repl-example on K64F and DISCO_F429NG
* mbed-js-ble-example on F411RE + IDB04A1


janjongboom:
@zherczeg Ah, didn't realize it was moved to jerryx. I'll make a follow up issue.

Issue #1855, opened 2017-05-26
Divide by negative zero does not keep sign.


glistening:
Jerry version:

```
Checked revision: e84109f
Build command: ./tools/build.py --clean
```

OS:

```
Ubuntu 14.04.3 LTS
```

Test case:

```
1 / (0 * (-1))
```

Result:

```
$ node
> 1 / (0 * (-1))
-Infinity

$ jsc
>>> 1 / (0 * (-1))
-Infinity

$ jerry 
jerry> 1 / (0 * (-1))
Infinity
```

glistening:
Close since #1856 landed.

PR #1856, opened 2017-05-26
Handle negative zero in number multiplication

glistening:
Keep sign for zero.
For example, 1 / (0 * (-1)) should be -Infinity, not +Infinity.

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -1772,7 +1772,9 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
             if (-ECMA_INTEGER_MULTIPLY_MAX <= left_integer
                 && left_integer <= ECMA_INTEGER_MULTIPLY_MAX
                 && -ECMA_INTEGER_MULTIPLY_MAX <= right_integer
-                && right_integer <= ECMA_INTEGER_MULTIPLY_MAX)
+                && right_integer <= ECMA_INTEGER_MULTIPLY_MAX
+                && left_value != 0
+                && right_value != 0)
             {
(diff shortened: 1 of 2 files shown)


glistening:
It is related to #1855. When I checked it with jerry-test-suite and jerry-tests, there was no regression. And when I run [benchmarks 10kk, and 1kk](https://github.com/jerryscript-project/jerryscript/blob/master/tests/benchmarks/jerry/), it also shows no regression on x86_64. Could you check the correctness and run sunspider?

zherczeg:
The bug must be elsewhere. -0 is not an integer number in JerryScript. I suspect the division is the problem. Let me check.

jerry-core/vm/vm.c
```diff
@@ -1764,7 +1764,7 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
                                && -(ECMA_INTEGER_MULTIPLY_MAX * ECMA_INTEGER_MULTIPLY_MAX) >= ECMA_INTEGER_NUMBER_MIN,
                                square_of_integer_multiply_max_must_fit_into_integer_value_range);
 
-          if (ecma_are_values_integer_numbers (left_value, right_value))
+          if (left_value != 0 && right_value != 0 && ecma_are_values_integer_numbers (left_value, right_value))
           {
```
zherczeg: Ok, it is a bug with the multiplication but we need to move this check down to the `if` starting with:

    if (-ECMA_INTEGER_MULTIPLY_MAX <= left_integer

Add:

    && left_integer != 0
    && right_integer != 0

glistening:
Ok. I'll move it. Thank you for review.

PR #1858, opened 2017-05-26
New extension: module

gabrielschulhof:
This extension provides two facilities: one for registering modules, and
one for registering module resolvers.

A module is defined as a global static structure containing a pointer
to a string which is the name of the module, and a pointer to a function
which will be called when an instance of the module is needed.

A module resolver is a function that accepts a string holding the name
of the module and returns a `jerry_value_t` if the module was found, or
zero if it was not. The resolver is expected to cache the
`jerry_value_t` it returns for a given string such that, if called with
the same string again, it will return the same `jerry_value_t`.

This extension provides a built-in module resolver which attempts to
load and cache modules that follow the above module definition.

Additionally, this extension provides an API for invoking all module
resolvers in sequence to attempt to resolve the name of a single module.
After one resolver returns a non-zero `jerry_value_t`, the API stops
iterating over the remaining resolvers and returns the `jerry_value_t`.
The API will return a `jerry_value_t` containing an error indicating
that the module was not found if it reaches the end of the list of
resolvers. The error it returns has an extra property `"moduleName"` the
value of which is a string containing the name of the module that the
API was asked to resolve.

### DIFF
--- a/jerry-ext/module/module.c
+++ b/jerry-ext/module/module.c
@@ -1,0 +1,225 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jmem.h"
+#include "jerryscript.h"
+#include "jerryscript-ext/module.h"
+
+static const jerry_char_t *module_name_property_name = (jerry_char_t *) "moduleName";
+static const jerry_char_t *on_resolve_absent = (jerry_char_t *) "Module on_resolve () must not be NULL";
+static const jerry_char_t *module_not_found = (jerry_char_t *) "Module not found";
+
+/**
+ * Create an error related to modules
+ *
+ * Creates an error object of the requested type with the additional property "moduleName" the value of which is a
+ * string containing the name of the module that was requested when the error occurred.
+ *
+ * @return the error
+ */
+static jerry_value_t
+jerryx_module_create_error (jerry_error_t error_type, /**< the type of error to create */
+                            const jerry_char_t *message, /**< the error message */
+                            const jerry_char_t *module_name) /**< the module_name */
+{
+  jerry_value_t ret = jerry_create_error (error_type, message);
+  jerry_value_t property_name = jerry_create_string (module_name_property_name);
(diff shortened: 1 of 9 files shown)


jerry-ext/module/module.c
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  for JERRYX_SECTION_ITERATE (jxm_modules, index, module_p)
+  {
+    if (!strcmp ((char *) module_p->name, (char *) name))
+    {
+      if (module_p->on_resolve)
+      {
+        ret = module_p->on_resolve ();
+        instance_p = (jerryx_module_instance_t *) jmem_heap_alloc_block (sizeof (jerryx_module_instance_t));
```
gabrielschulhof: Should I just use a `jerry_create_object ()` for `instances_pp` and store that as my context data? Then I can set properties on it for each module I wish to cache and I don't need to do any `jmem_heap_alloc_block()`-ing.

gabrielschulhof: ... and the lookup will likely be faster than linear for when there are many modules.

martijnthe:
Thanks @gabrielschulhof !

@jiangzidong @akosthekiss @zherczeg could you help review this?
Thanks a lot.

gabrielschulhof:
@martijnthe @jiangzidong @akosthekiss @zherczeg I'd really like to know what you think about using a JS object instead of a linked list, because I am more and more inclined to change this PR to use a JS object. However, for that to work I need to clear up how it's gonna be with #1859 and #1860, because checking whether a module has already been loaded is done via `jerry_has_property ()`.

jerry-ext/module/module.c
martijnthe: That's fine with me.

I also wonder whether it would be useful if all the resolvers would share the same cache map, instead of each resolver having to keep its own piece of the cache (just move the caching mechanism a level up, into `jerryx_module_resolve`). That way each resolver can just focus on resolving and not have to worry about caching.

gabrielschulhof: OK. I'll modify the PR to work with a JS object after we've resolved #1859.

I was also thinking about moving up the cache but I realized that by doing that I don't even call the resolver at all if I find the instance in the cache. I'm not sure that that's such a good idea. I think the resolver should *always* be called, and *it* should decide whether to return a cached item or not - meaning that *it* has to maintain such a cache.

Debatable, to be sure, and I'll be happy to move up the caching if we agree on that.

martijnthe: > I'm not sure that that's such a good idea. I think the resolver should always be called, and it should decide whether to return a cached item or not - meaning that it has to maintain such a cache.

Why do you think the resolver needs to make the decision to return a cached item?

BTW, a resolver already isn't guaranteed to be called if a resolver that sits before it in the array of resolvers handled the request.

gabrielschulhof: Alright. I guess I don't have any strong reasons. Just a gut feeling that it's better to let the resolvers do their own thing instead of doing too much on their behalf. OTOH ES6 modules *require* caching, so I guess we could/should handle the caching centrally.

Alright. I'll cache higher-up.

gabrielschulhof:
Argh! I accidentally deleted the branch! I'll open a new PR. Sorry everybody!

Issue #1859, opened 2017-05-27
jerry_has_property () returns true even on a newly created object


gabrielschulhof:
The following test fails at `jerry_has_property ()`:
```c
#include "config.h"
#include "jerryscript.h"
#include "test-common.h"

int
main (void)
{
  TEST_INIT ();

  jerry_init (JERRY_INIT_EMPTY);

  jerry_value_t object = jerry_create_object ();
  jerry_value_t prop_name = jerry_create_string_from_utf8 ((jerry_char_t *) "something");

  TEST_ASSERT (!jerry_has_property (object, prop_name));
  TEST_ASSERT (!jerry_has_own_property (object, prop_name));

  jerry_release_value (object);
  jerry_release_value (prop_name);

  jerry_cleanup ();

  return 0;
} /* main */
```

Why would any prototype in `object`'s prototype chain have the property `"something"`?

Either way, I think it'd be good to cover those two APIs with a test.

gabrielschulhof:
OK. NM. The documentation is wrong. `jerry_has_property ()` and `jerry_has_own_property ()` return `jerry_value_t` not `bool`.

LaszloLango:
Wrong button :) sorry

PR #1860, opened 2017-05-27
Clean up property existence check inconsistencies

gabrielschulhof:
`jerry_has_property ()` and `jerry_has_own_property ()` are inconsistent
in that they squash their return value into a `bool` when in fact it is
a `jerry_value_t`, thereby giving the wrong impression that a property
is there when it isn't and encouraging the leaking of `jerry_value_t`s
by disguising them as `bool`s.

Fixes https://github.com/jerryscript-project/jerryscript/issues/1859

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -2767,19 +2767,22 @@ jerry_create_undefined (void);
 
 **Summary**
 
-Checks whether the object or it's prototype objects have the given property.
+Checks whether the object or its prototype objects have the given property.
+
+*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
+is no longer needed.
 
 **Prototype**
 
 ```c
-bool
+jerry_value_t
 jerry_has_property (const jerry_value_t obj_val,
                     const jerry_value_t prop_name_val);
 ```
 
 - `obj_val` - object value
 - `prop_name_val` - property name
-- return value
+- return value - JavaScript boolean value that evaluates to
   - true, if the property exists
   - false, otherwise
 
@@ -2790,8 +2793,10 @@ jerry_has_property (const jerry_value_t obj_val,
   jerry_value_t global_object = jerry_get_global_object ();
   jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) "handler_field");
 
-  bool has_prop = jerry_has_property (global_object, prop_name);
+  jerry_value_t has_prop_js = jerry_has_property (global_object, prop_name);
+  bool has_prop = jerry_get_boolean_value (has_prop_js);
 
+  jerry_release_value (has_prop_js);
   jerry_release_value (prop_name);
   jerry_release_value (global_object);
 }
@@ -2809,17 +2814,20 @@ jerry_has_property (const jerry_value_t obj_val,
 
 Checks whether the object has the given property.
 
+*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
+is no longer needed.
+
 **Prototype**
 
 ```c
-bool
+jerry_value_t
 jerry_has_own_property (const jerry_value_t obj_val,
                         const jerry_value_t prop_name_val);
 ```
 
 - `obj_val` - object value
 - `prop_name_val` - property name
-- return value
+- return value - JavaScript boolean value that evaluates to
   - true, if the property exists
   - false, otherwise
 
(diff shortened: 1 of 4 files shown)


zherczeg:
Looks like an oversight. We should simply remove the `ecma_make_boolean_value` at the end.

gabrielschulhof:
@zherczeg does the spec say that evaluating this may end up calling into JS? If so, it may throw, and this needs to stay a JS boolean - similar to `jerry_get/set_property ()`.

gabrielschulhof:
@zherczeg sounds like [proxy](https://ariya.io/2013/07/es6-and-proxy) might be such an ES6 feature.

zherczeg:
That is the point, it cannot throw.

gabrielschulhof:
@zherczeg I don't understand. If `jerry_has_property ()` ends up invoking JS because the object is a proxy, then that JS may throw, and that error needs to be returned as a `jerry_value_t` with the error flag set.

gabrielschulhof:
@zherczeg consider the following code:
```JS
var x = {}
var y = new Proxy( x, {
	has: function( target, prop ) {
		return prop === "abc";
	}
} );

console.log( "abc" in y );
console.log( "def" in y );

var z = new Proxy( x, {
	has: function( target, prop ) {
		throw new Error( "Peekaboo!" );
	}
} );

console.log( "ghi" in z );
```

Now, if I perform `jerry_has_property (z, anything)`, it will throw - at least that's what it does in Node.js.

gabrielschulhof:
So, the return value of `jerry_has_property ()` and `jerry_has_own_property ()` must be `jerry_value_t` so we can accommodate `Proxy` in the future.

zherczeg:
The problem is we break the api. If is hard to judge what is the right decision here.

gabrielschulhof:
@zherczeg what do you mean "we break the api"? The API is broken right now. `jerry_has_property ()` and `jerry_has_own_property ()` are documented as returning `bool`, and, indeed they are declared as returning `bool`, but in reality they return `jerry_value_t` so now they always return `true` because they return a `jerry_value_t` cast as a bool.

So, both functions work incorrectly. Therefore, `jerry_has_property ()` and `jerry_has_own_property ()` as documented and as they behave now are completely useless, dangerous even, because they always return `true` (useless), and they leak `jerry_value_t`s (dangerous).

Therefore, API commitment or not, any code making calls to these functions is completely broken. If we fix it my way, the user code in question will be no more broken than it is now, and it will continue to compile against new versions because of the implicit inter-castability of `bool` and `jerry_value_t` (which is perhaps the root of the problem) but the documentation will be correct, and the implementation will be more future-proof.

We could mention this breakage at the time of release, pointing people to the updated documentation. Any and all user code making use of `jerry_has_(own_)?property ()` will then have to be updated anyway.

gabrielschulhof:
OTOH, if we fix these APIs to actually return `bool` we'll have to add two new APIs later which return `jerry_value_t`, and we'll have to deprecate these two, which by then will behave incorrectly in some cases because they cannot do anything else but squash thrown errors in order to fit the result into a `bool`.

gabrielschulhof:
On the plus side of actually returning `bool`, the leak will be gone and existing user code will be fixed.

gabrielschulhof:
We could do something like this if we wanted to keep the return value as `bool`:
```c
jerry_value_t jerry_get_last_error ();
```
... and then in `jerry_has_(own_)?property ()`:
```c
/* At the top of the function */
if (JERRY_CONTEXT (last_error))
{
  jerry_release_value (JERRY_CONTEXT (last_error));
  JERRY_CONTEXT (last_error) = 0;
}
...
/* 
 * At the bottom of the function, assuming has_property is now a
 * jerry_value_t, and not a boolean, because of the ES6-compliant
 * implementation.
 */
if (jerry_value_has_error_flag (has_property))
{
  JERRY_CONTEXT (last_error) = jerry_acquire_value (has_property);
  jerry_release_value (has_property);

 /*
  * false? true? I don't know - does having thrown an error mean the
  * property is there or does it mean it is absent? Let's go for
  * "absent".
  */
  return false;
}

ret = jerry_get_boolean_value (has_property);
jerry_release_value (has_property);
return ret;
```
This is actually what ChakraCore [does](https://github.com/Microsoft/ChakraCore/wiki/JsGetAndClearException) in all cases where exceptions might be thrown. I don't like that approach, and, in fact, I like the JerryScript approach of having an error flag much, much better.

zherczeg:
Ok, you convinced me.

PR #1863, opened 2017-05-29
New extension: module

gabrielschulhof:
This extension provides the following facilities:
  - registering modules,
  - registering module resolvers, and
  - an API for retrieving a module instance given its name.

A module is defined as a global static structure containing a pointer
to a string which is the name of the module, and a pointer to a function
which will be called when an instance of the module is needed.

A module resolver is a function that accepts a string holding the name
of the module and returns a `jerry_value_t` if the module was found, or
zero if it was not. If it returns a non-zero value with the error flag
set then the API will pass it through without caching.

This extension provides a built-in module resolver which attempts to
load modules that follow the above module definition.

The API provided by this extension invokes all module resolvers in
sequence to attempt to resolve the name of a single module. After one
resolver returns a non-zero `jerry_value_t`, the API stops iterating
over the remaining resolvers and caches the value if its error flag is
not set. It then returns the `jerry_value_t`. The API will return a
`jerry_value_t` containing an error indicating that the module was not
found if it reaches the end of the list of resolvers. The error it
returns has an extra property `"moduleName"` the value of which is a
string containing the name of the module that the API was asked to
resolve.

### DIFF
--- a/docs/12.EXT-REFERENCE-MODULE.md
+++ b/docs/12.EXT-REFERENCE-MODULE.md
@@ -1,0 +1,186 @@
+#  Module API
+
+This is a JerryScript extension that provides a means of loading modules. Fundamentally, a module is a name (stored as
+a string) that resolves to a `jerry_value_t`. This extension provides the function `jerryx_module_resolve()` which
+accepts the name of the module being requested as well as an array of so-called "resolvers" - functions which satisfy
+the signature `jerryx_module_resolver_t`. The resolvers in the list are called in sequence until one of them returns
+`true` and a `jerry_value_t` in its out parameter. The value is cached if it is not an error, so subsequent requests
+for the same name will not result in additional calls to the resolvers.
+
+The purpose of having resolvers is to be able to account for the fact that different types of modules may be structured
+differently and thus, for each type of module a module resolver must be supplied at the point where an instance of that
+type of module is requested.
+
+Additionally, this extension provides a means of easily defining so-called "native" JerryScript modules which can be
+resolved using the JerryScript native module resolver `jerryx_module_native_resolver()`, which can be passed to
+`jerryx_module_resolve()`. Note, however, that native JerryScript modules are only supported and
+`jerryx_module_native_resolver()` is only compiled in if compiler support for `__attribute__` extensions is present. In
+effect this means that native JerryScript modules are available only when this extension is built with GCC or
+LLVM/clang. In the absence of such support, you may construct alternative module systems and provide your own resolver
+to `jerryx_module_resolve()`.
+
+`jerryscript-ext/module.h` defines the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` only if support for
+native JerryScript modules is available.
+
+## jerryx_module_resolve
+
+**Summary**
+
+Load a copy of a module into the current context or return one that was already loaded if it is found.
+
+Each function in `resolvers_p` will be called in sequence until one returns `true` and fills out its out-parameter with
+the `jerry_value_t` representing the requested module. If the `jerry_value_t` does not have the error flag set it will
+be cached. Thus, on subsequent calls with the same value for `name`, none of the functions in `resolvers_p` will be
+called.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name,
+                       jerryx_module_resolver_t *resolvers_p,
+                       size_t resolver_count);
+```
+
+- `name` - the name of the module to load
+- `resolvers_p` - the list of resolvers to call in sequence
+- `resolver_count` - the number of resolvers in `resolvers_p`
+- return value - `jerry_value_t` representing the module that was loaded, or the error that occurred in the process.
+
+
+## jerryx_module_native_resolver
+
+**Summary**
+
+The resolver for JerryScript modules. A pointer to this function can be passed in the second parameter to
+`jerryx_module_resolve` to search for the module among the JerryScript modules built into the binary. This function is
+available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is defined.
+
+**Prototype**
+
+```c
+bool
+jerryx_module_native_resolver (const jerry_char_t *name,
+                               jerry_value_t *result)
+```
+- `name` - the name of the module to find
+- `result` - out - place where to store the resulting module instance
+- return value - `true` if the module was found and stored in `result`, and `false` otherwise
+
+
+# Module data types
+
+## jerryx_native_module_on_resolve_t
+
+**Summary**
+
+Function pointer type for a function that will create an instance of a native module. This type is only defined if the
+preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is defined.
+
+**Prototype**
+
+```c
+typedef jerry_value_t (*jerryx_native_module_on_resolve_t) (void);
+```
+
+## jerryx_module_resolver_t
+
+**Summary**
+
+Function pointer type for a module resolver
+
+**Prototype**
+
+```c
+typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+```
+
+**Example**
+```c
+bool
+load_and_evaluate_js_file (const jerry_char_t *name, jerry_value_t *result)
+{
+  bool return_value = false;
+  char *js_file_contents = NULL;
+  int file_size = 0;
+  FILE *js_file = fopen (name, "r");
+
+  if (js_file)
+  {
+    /* We have successfully opened the file. Now, we establish its size. */
+    file_size = fseek (js_file, 0, SEEK_END);
+    fseek (js_file, 0, SEEK_SET);
+
+    /* We allocate enough memory to store the contents of the file. */
+    js_file_contents = malloc (file_size);
+    if (js_file_contents)
+    {
+      /* We read the file into memory and call jerry_eval (), assigning the result to the out-parameter. */
+      fread (js_file_contents, file_size, 1, js_file);
+      (*result) = jerry_eval (js_file_contents, file_size, false);
+
+      /* We release the memory holding the contents of the file. */
+      free (js_file_contents);
+      return_value = true;
+    }
+
+    /* We close the file. */
+    fclose (js_file);
+  }
+
+  return return_value;
+}
+```
+
+We can now load JavaScript files:
+```c
+static const jerryx_module_resolver_t resolvers =
+{
+  /* Consult the JerryScript module resolver first, in case the requested module is a compiled-in JerryScript module. */
+  jerryx_module_native_resolver,
+
+  /*
+   * If the requested module is not a JerryScript module, assume it is a JavaScript file on disk and use the above-
+   * defined JavaScript file loader to load it.
+   */
+  load_and_evaluate_js_file
+};
+jerry_value_t js_module = jerryx_module_resolve (requested_module, resolvers, 2);
+```
+
+# Module helper macros
+
+## JERRYX_NATIVE_MODULE
+
+**Summary**
+
+Helper macro to define a JerryScript module. Currently stores the name of the module and its initializer in an
+executable linker section. This macro is available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED`
+is defined.
+
+**Note**: The helper macro must appear at the bottom of a source file, and no semicolon must follow it.
+
+**Prototype**
+```c
+#define JERRYX_NATIVE_MODULE(module_name, on_resolve_cb)
+```
+
+- `module_name` - the name of the module without quotes
+- `on_resolve_cb` - the function of type `jerryx_native_module_on_resolve_t` that will be called when the module needs to be
+loaded.
+
+**Example**
+
+```c
+#include "jerryscript.h"
+#include "jerryscript-ext/module.h"
+
+static jerry_value_t
+my_module_on_resolve (void)
+{
+  return jerry_create_external_function (very_useful_function);
+} /* my_module_on_resolve */
+
+/* Note that there is no semicolon at the end of the next line. This is how it must be. */
+JERRYX_NATIVE_MODULE (my_module, my_module_on_resolve)
+```
--- a/jerry-ext/include/jerryscript-ext/module.h
+++ b/jerry-ext/include/jerryscript-ext/module.h
@@ -1,0 +1,74 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRYX_MODULE_H
+#define JERRYX_MODULE_H
+
+#include "jerryscript.h"
+
+#ifdef __GNUC__
+#define JERRYX_NATIVE_MODULES_SUPPORTED
+#endif /* __GNUC__ */
+
+#ifdef JERRYX_NATIVE_MODULES_SUPPORTED
+#include "jerryscript-ext/section.impl.h"
+
+/**
+ * Declare the signature for the module initialization function.
+ */
+typedef jerry_value_t (*jerryx_native_module_on_resolve_t) (void);
+
+/**
+ * Declare the structure used to define a module. One should only make use of this structure via the
+ * JERRYX_NATIVE_MODULE macro declared below.
+ */
+typedef struct
+{
+  jerry_char_t *name; /**< name of the module */
+  jerryx_native_module_on_resolve_t on_resolve; /**< function that returns a new instance of the module */
+} jerryx_native_module_t;
+
+/**
+ * Declare a helper macro that expands to the declaration of a variable of type jerryx_native_module_t placed into the
+ * specially-named linker section "jerryx_modules" where the JerryScript module resolver
+ * jerryx_module_native_resolver () will look for it.
+ */
+#define JERRYX_NATIVE_MODULE(module_name, on_resolve_cb)                                 \
+  static const jerryx_native_module_t _module JERRYX_SECTION_ATTRIBUTE(jerryx_modules) = \
+  {                                                                                      \
+    .name = ((jerry_char_t *) #module_name),                                             \
+    .on_resolve = (on_resolve_cb)                                                        \
+  };
+
+/**
+ * Declare the JerryScript module resolver so that it may be added to an array of jerryx_module_resolver_t items and
+ * thus passed to jerryx_module_resolve.
+ */
+bool jerryx_module_native_resolver (const jerry_char_t *name, jerry_value_t *result);
+
+#endif /* JERRYX_NATIVE_MODULES_SUPPORTED */
+
+/**
+ * Declare the function pointer type for module resolvers.
+ */
+typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+
+/**
+ * Load a copy of a module into the current context using the provided module resolvers, or return one that was already
+ * loaded if it is found.
+ */
+jerry_value_t jerryx_module_resolve (const jerry_char_t *name, jerryx_module_resolver_t *resolvers, size_t count);
+
+#endif /* !JERRYX_MODULE_H */
--- a/jerry-ext/include/jerryscript-ext/section.impl.h
+++ b/jerry-ext/include/jerryscript-ext/section.impl.h
@@ -1,0 +1,56 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRYX_SECTION_IMPL_H
+#define JERRYX_SECTION_IMPL_H
+
+/**
+ * Define the name of a section
+ *
+ * @param name the name of the section without quotes
+ */
+#ifdef __MACH__
+#define JERRYX_SECTION_NAME(name) "__DATA," #name
+#else /* !__MACH__ */
+#define JERRYX_SECTION_NAME(name) #name
+#endif /* __MACH__ */
+
+/**
+ * Expands to the proper __attribute__(()) qualifier for appending a variable to a section.
+ */
+#define JERRYX_SECTION_ATTRIBUTE(name) \
+  __attribute__ ((used, section (JERRYX_SECTION_NAME (name)), aligned (sizeof (void *))))
+
+/**
+ * Declare references to a section that contains an array of items.
+ *
+ * @param name the name of the section (without quotes)
+ * @param type the type of the elements stored in the array
+ *
+ * This macro declares two extern const variables such that their type is an array of @p type and their names are
+ * constructed by prefixing @p name with "__start_" and "__stop_". They evaluate to the starting and ending address
+ * of the section @p name.
+ */
+#ifdef __MACH__
+#define JERRYX_SECTION_DECLARE(name, type)                                   \
+  extern const type __start_ ## name[] __asm("section$start$__DATA$" #name); \
+  extern const type __stop_ ## name[] __asm("section$end$__DATA$" #name);
+#else /* !__MACH__ */
+#define JERRYX_SECTION_DECLARE(name, type)                    \
+  extern const type __start_ ## name[] __attribute__((weak)); \
+  extern const type __stop_ ## name[] __attribute__((weak));
+#endif /* __MACH__ */
+
+#endif /* !JERRYX_SECTION_IMPL_H */
--- a/jerry-ext/module/module.c
+++ b/jerry-ext/module/module.c
@@ -1,0 +1,217 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string.h>
+#include "jerryscript.h"
+#include "jerryscript-ext/module.h"
+
+static const jerry_char_t *module_name_property_name = (jerry_char_t *) "moduleName";
+static const jerry_char_t *module_not_found = (jerry_char_t *) "Module not found";
+
+/**
+ * Create an error related to modules
+ *
+ * Creates an error object of the requested type with the additional property "moduleName" the value of which is a
+ * string containing the name of the module that was requested when the error occurred.
+ *
+ * @return the error
+ */
+static jerry_value_t
+jerryx_module_create_error (jerry_error_t error_type, /**< the type of error to create */
+                            const jerry_char_t *message, /**< the error message */
+                            const jerry_char_t *module_name) /**< the module name */
+{
+  jerry_value_t ret = jerry_create_error (error_type, message);
+  jerry_value_t property_name = jerry_create_string (module_name_property_name);
+  jerry_value_t property_value = jerry_create_string_from_utf8 (module_name);
+
+  jerry_release_value (jerry_set_property (ret, property_name, property_value));
+  jerry_release_value (property_name);
+  jerry_release_value (property_value);
+  return ret;
+} /* jerryx_module_create_error */
+
+/**
+ * Initialize the module manager extension.
+ */
+static void
+jerryx_module_manager_init (void *user_data_p)
+{
+  *((jerry_value_t *) user_data_p) = jerry_create_object ();
+} /* jerryx_module_manager_init */
+
+/**
+ * Deinitialize the module manager extension.
+ */
+static void
+jerryx_module_manager_deinit (void *user_data_p) /**< context pointer to deinitialize */
+{
+  jerry_release_value (*(jerry_value_t *) user_data_p);
+} /* jerryx_module_manager_deinit */
+
+/**
+ * Declare the context data manager for modules.
+ */
+static const jerry_context_data_manager_t jerryx_module_manager =
+{
+  .init_cb = jerryx_module_manager_init,
+  .deinit_cb = jerryx_module_manager_deinit,
+  .bytes_needed = sizeof (jerry_value_t)
+};
+
+/**
+ * Declare the linker section where module definitions are stored.
+ */
+JERRYX_SECTION_DECLARE (jerryx_modules, jerryx_native_module_t)
+
+/**
+ * Attempt to retrieve a module by name from a cache, and return false if not found.
+ */
+static bool
+jerryx_module_check_cache (jerry_value_t cache, /**< cache from which to attempt to retrieve the module by name */
+                           jerry_value_t module_name, /**< JerryScript string value holding the module name */
+                           jerry_value_t *result) /**< Resulting value */
+{
+  bool ret = false;
+
+  /* Check if the cache has the module. */
+  jerry_value_t js_has_property = jerry_has_property (cache, module_name);
+
+  /* If we succeed in getting an answer, we examine the answer. */
+  if (!jerry_value_has_error_flag (js_has_property))
+  {
+    bool has_property = jerry_get_boolean_value (js_has_property);
+
+    /* If the module is indeed in the cache, we return it. */
+    if (has_property)
+    {
+      (*result) = jerry_get_property (cache, module_name);
+      ret = true;
+    }
+  }
+
+  jerry_release_value (js_has_property);
+
+  return ret;
+} /* jerryx_module_check_cache */
+
+/**
+ * Attempt to cache a loaded module.
+ *
+ * @return the module on success, otherwise the error encountered when attempting to cache. In the latter case, the
+ * @p module is released.
+ */
+static jerry_value_t
+jerryx_module_add_to_cache (jerry_value_t cache, /**< cache to which to add the module */
+                            jerry_value_t module_name, /**< key at which to cache the module */
+                            jerry_value_t module) /**< the module to cache */
+{
+  jerry_value_t ret = jerry_set_property (cache, module_name, module);
+
+  if (jerry_value_has_error_flag (ret))
+  {
+    jerry_release_value (module);
+  }
+  else
+  {
+    jerry_release_value (ret);
+    ret = module;
+  }
+
+  return ret;
+} /* jerryx_module_add_to_cache */
+
+#ifdef JERRYX_NATIVE_MODULES_SUPPORTED
+static const jerry_char_t *on_resolve_absent = (jerry_char_t *) "Module on_resolve () must not be NULL";
+
+/**
+ * Declare and define the default module resolver - one which examines what modules are defined in the above linker
+ * section and loads one that matches the requested name, caching the result for subsequent requests using the context
+ * data mechanism.
+ */
+bool
+jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module */
+                               jerry_value_t *result) /**< [out] where to put the resulting module instance */
+{
+  int index;
+  const jerryx_native_module_t *module_p = NULL;
+
+  /* Look for the module by its name in the list of module definitions. */
+  for (index = 0, module_p = &__start_jerryx_modules[0];
+       &__start_jerryx_modules[index] < __stop_jerryx_modules;
+       index++, module_p = &__start_jerryx_modules[index])
+  {
+    if (module_p->name != NULL && !strcmp ((char *) module_p->name, (char *) name))
+    {
+      /* If we find the module by its name we load it and cache it if it has an on_resolve () and complain otherwise. */
+      (*result) = ((module_p->on_resolve) ? module_p->on_resolve ()
+                                          : jerryx_module_create_error (JERRY_ERROR_TYPE, on_resolve_absent, name));
+      return true;
+    }
+  }
+
+  return false;
+} /* jerryx_module_native_resolver */
+#endif /* JERRYX_NATIVE_MODULES_SUPPORTED */
+
+/**
+ * Resolve a single module using the module resolvers available in the section declared above and load it into the
+ * current context.
+ *
+ * @p name - name of the module to resolve
+ * @p resolvers - list of resolvers to invoke
+ * @p count - number of resolvers in the list
+ *
+ * @return a jerry_value_t containing one of the followings:
+ *   - the result of having loaded the module named @p name, or
+ *   - the result of a previous successful load, or
+ *   - an error indicating that something went wrong during the attempt to load the module.
+ */
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name, /**< name of the module to load */
+                       jerryx_module_resolver_t *resolvers_p, /**< list of resolvers */
+                       size_t resolver_count) /**< number of resolvers in @p resolvers */
+{
+  size_t index;
+  jerry_value_t ret;
+  jerry_value_t instances = *(jerry_value_t *) jerry_get_context_data (&jerryx_module_manager);
+  jerry_value_t module_name = jerry_create_string_from_utf8 (name);
+
+  /* Return the cached instance if present. */
+  if (jerryx_module_check_cache (instances, module_name, &ret))
+  {
+    goto done;
+  }
+
+  /* Try each resolver until one manages to find the module. */
+  for (index = 0; index < resolver_count; index++)
+  {
+    if ((*resolvers_p[index]) (name, &ret))
+    {
+      if (!jerry_value_has_error_flag (ret))
+      {
+        ret = jerryx_module_add_to_cache (instances, module_name, ret);
+      }
+      goto done;
+    }
+  }
+
+  /* If none of the resolvers manage to find the module, complain with "Module not found" */
+  ret = jerryx_module_create_error (JERRY_ERROR_COMMON, module_not_found, name);
+
+done:
+  jerry_release_value (module_name);
+  return ret;
+} /* jerryx_module_resolve */
--- a/tools/update-webpage.sh
+++ b/tools/update-webpage.sh
@@ -33,6 +33,7 @@ CODING_STANDARDS_MD="08.CODING-STANDARDS.md"
 EXT_REFERENCE_ARG_MD="09.EXT-REFERENCE-ARG.md"
 EXT_REFERENCE_HANDLER_MD="10.EXT-REFERENCE-HANDLER.md"
 EXT_REFERENCE_AUTORELEASE_MD="11.EXT-REFERENCE-AUTORELEASE.md"
+EXT_REFERENCE_MODULE_MD="12.EXT-REFERENCE-MODULE.md"
 
 declare -A titles
 
@@ -47,6 +48,7 @@ titles[$CODING_STANDARDS_MD]="Coding Standards"
 titles[$EXT_REFERENCE_ARG_MD]="'Extension API: Argument Validation'"
 titles[$EXT_REFERENCE_HANDLER_MD]="'Extension API: External Function Handlers'"
 titles[$EXT_REFERENCE_AUTORELEASE_MD]="'Extension API: Autorelease Values'"
+titles[$EXT_REFERENCE_MODULE_MD]="'Extension API: Module Support'"
 
 for docfile in $docs_dir/*.md; do
   docfile_base=`basename $docfile`
--- a/tests/unit-ext/CMakeLists.txt
+++ b/tests/unit-ext/CMakeLists.txt
@@ -15,6 +15,8 @@
 cmake_minimum_required (VERSION 2.8.12)
 project (unit-ext C)
 
+set(INCLUDE_UNIT_EXT ${CMAKE_CURRENT_SOURCE_DIR})
+
 # Unit tests main modules
 file(GLOB SOURCE_UNIT_TEST_EXT_MODULES *.c)
 
@@ -32,4 +34,11 @@ foreach(SOURCE_UNIT_TEST_EXT ${SOURCE_UNIT_TEST_EXT_MODULES})
   target_link_libraries(${TARGET_NAME} jerry-ext jerry-core jerry-port-default-minimal)
 
   add_dependencies(unittests-ext ${TARGET_NAME})
+endforeach()
+
+file(GLOB CONTENTS_UNIT_TEST_EXT *)
+foreach(CONTENT_UNIT_TEST_EXT ${CONTENTS_UNIT_TEST_EXT})
+  if(IS_DIRECTORY ${CONTENT_UNIT_TEST_EXT})
+    add_subdirectory(${CONTENT_UNIT_TEST_EXT})
+  endif()
 endforeach()
--- a/tests/unit-ext/module/CMakeLists.txt
+++ b/tests/unit-ext/module/CMakeLists.txt
@@ -13,27 +13,12 @@
 # limitations under the License.
 
 cmake_minimum_required (VERSION 2.8.12)
-set(JERRY_EXT_NAME jerry-ext)
-project (${JERRY_EXT_NAME} C)
+set(JERRYX_MODULE_UNITTEST_NAME unit-test-jerry-module)
+project (${JERRYX_MODULE_UNITTEST_NAME} C)
 
-# Include directories
-set(INCLUDE_EXT "${CMAKE_CURRENT_SOURCE_DIR}/include")
+file(GLOB JERRYX_MODULE_UNIT_TEST_SOURCES *.c)
 
-# Source directories
-file(GLOB SOURCE_EXT_ARG      arg/*.c)
-file(GLOB SOURCE_EXT_MODULE   module/*.c)
-file(GLOB SOURCE_EXT_HANDLER  handler/*.c)
-
-set(SOURCE_EXT
-    ${SOURCE_EXT_ARG}
-    ${SOURCE_EXT_MODULE}
-    ${SOURCE_EXT_HANDLER})
-
-add_library(${JERRY_EXT_NAME} STATIC ${SOURCE_EXT})
-
-target_include_directories(${JERRY_EXT_NAME} PUBLIC ${INCLUDE_EXT})
-
-target_link_libraries(${JERRY_EXT_NAME} jerry-core)
-
-install(TARGETS ${JERRY_EXT_NAME} DESTINATION lib)
-install(DIRECTORY ${INCLUDE_EXT}/ DESTINATION include)
+add_executable(${JERRYX_MODULE_UNITTEST_NAME} ${JERRYX_MODULE_UNIT_TEST_SOURCES})
+set_property(TARGET ${JERRYX_MODULE_UNITTEST_NAME} PROPERTY LINK_FLAGS "${LINKER_FLAGS_COMMON}")
+target_link_libraries(${JERRYX_MODULE_UNITTEST_NAME} jerry-ext jerry-core jerry-port-default-minimal)
+target_include_directories(${JERRYX_MODULE_UNITTEST_NAME} PRIVATE ${INCLUDE_UNIT_EXT})
--- a/tests/unit-ext/module/jerry-module-test.c
+++ b/tests/unit-ext/module/jerry-module-test.c
@@ -1,0 +1,168 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string.h>
+
+#include "jerryscript.h"
+#include "test-common.h"
+#include "jerryscript-ext/module.h"
+
+#ifdef JERRYX_NATIVE_MODULES_SUPPORTED
+
+/* Load a module. */
+const char eval_string1[] = "require ('my_custom_module');";
+
+/* Load a module using a different resolver. */
+const char eval_string2[] = "require ('differently-handled-module');";
+
+/* Load a broken module using the built-in resolver. */
+const char eval_string3[] =
+"(function() {"
+"  var theError;"
+"  try {"
+"    require ('my_broken_module');"
+"  } catch (anError) {"
+"    theError = anError;"
+"  }"
+"  return (((theError.message === 'Module on_resolve () must not be NULL') &&"
+"    (theError.moduleName === 'my_broken_module') &&"
+"    (theError instanceof TypeError)) ? 1 : 0);"
+"}) ();";
+
+/* Load a non-existent module. */
+const char eval_string4[] =
+"(function() {"
+"  var theError;"
+"  try {"
+"    require ('some_missing_module_xyzzy');"
+"  } catch (anError) {"
+"    theError = anError;"
+"  }"
+"  return (((theError.message === 'Module not found') &&"
+"    (theError.moduleName === 'some_missing_module_xyzzy')) ? 1 : 0);"
+"}) ();";
+
+/* Make sure the result of a module load is cached. */
+const char eval_string5[] =
+"(function() {"
+"  var x = require('cache-check');"
+"  var y = require('cache-check');"
+"  return x === y ? 1 : 0;"
+"}) ();";
+
+static bool
+resolve_differently_handled_module (const jerry_char_t *name,
+                                                jerry_value_t *result)
+{
+  if (!strcmp ((char *) name, "differently-handled-module"))
+  {
+    (*result) = jerry_create_number (29);
+    return true;
+  }
+  return false;
+} /* resolve_differently_handled_module */
+
+/*
+ * Define module "cache-check" via its own resolver as an empty object. Since objects are accessible only via references
+ * we can strictly compare the object returned on subsequent attempts at loading "cache-check" with the object returned
+ * on the first attempt and establish that the two are in fact the same object - which in turn shows that caching works.
+ */
+static bool
+cache_check (const jerry_char_t *name,
+             jerry_value_t *result)
+{
+  if (!strcmp ((char *) name, "cache-check"))
+  {
+    (*result) = jerry_create_object ();
+    return true;
+  }
+  return false;
+} /* cache_check */
+
+static jerryx_module_resolver_t resolvers[3] =
+{
+  jerryx_module_native_resolver,
+  resolve_differently_handled_module,
+  cache_check
+};
+
+static jerry_value_t
+handle_require (const jerry_value_t js_function,
+                const jerry_value_t this_val,
+                const jerry_value_t args_p[],
+                const jerry_length_t args_count)
+{
+  (void) js_function;
+  (void) this_val;
+  (void) args_count;
+
+  jerry_value_t return_value = 0;
+  jerry_char_t module_name[256] = "";
+  jerry_size_t bytes_copied = 0;
+
+  TEST_ASSERT (args_count == 1);
+  bytes_copied = jerry_string_to_char_buffer (args_p[0], module_name, 256);
+  if (bytes_copied < 256)
+  {
+    module_name[bytes_copied] = 0;
+    return_value = jerryx_module_resolve (module_name, resolvers, 3);
+  }
+
+  return return_value;
+} /* handle_require */
+
+static void
+assert_number (jerry_value_t js_value, double expected_result)
+{
+  TEST_ASSERT (!jerry_value_has_error_flag (js_value));
+  TEST_ASSERT (jerry_get_number_value (js_value) == expected_result);
+} /* assert_number */
+
+static void
+eval_one (const char *the_string, double expected_result)
+{
+  jerry_value_t js_eval_result = jerry_eval ((const jerry_char_t *) the_string, strlen (the_string), true);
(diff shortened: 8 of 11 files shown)


glistening:
@gabrielschulhof This PR's module syntax looks similar to nodejs's `require`. Do you have reason to use this syntax? (e.g., backward compatibility to existing nodejs style code?) Are you going to implement node.js compliant module system? Why don't we follow ECMA2015 specification module syntax? JerryScript is JavaScript Engine, so at least for me, it seems natural to make jerryscript ECMA specification compliant.

gabrielschulhof:
@glistening I'm not providing any new language structures in this PR. I'm simply providing a way of converting a string to a `jerry_value_t` with module semantics. It is indeed similar to Node.js if that's how you choose to expose it to JS - which is what my unit test does.

I'm certainly in favour of implementing ES6 modules in JerryScript. I can imagine JerryScript exposing an API such as
```c
typedef jerry_value_t (*jerry_module_handler_t) (const jerry_char_t *);
```
and
```c
void jerry_set_module_handler (jerry_module_handler_t handler);
```
or even
```c
void jerry_init_with_module_handler (jerry_init_flag_t flag, jerry_module_handler_t handler);
```
which can be satisfied by `jerryx_module_resolve ()` as provided by this PR.

I am not sufficiently familiar with the JerryScript interpreter to write the parser for the ES6 syntax bits concerning modules, however, I'm fairly certain that we can use this PR to implement ES6 modules if/when the parser is ready and uses an external hook as described above.

martijnthe:
@glistening:

> Why don't we follow ECMA2015 specification module syntax? JerryScript is JavaScript Engine, so at least for me, it seems natural to make jerryscript ECMA specification compliant.

This was discussed briefly in the community call of 2017/04/19.
Here is a snippet from the meeting notes – See https://groups.io/g/jerryscript-dev/message/7

> "... Was discussing with Gabriel, working on building a utility on top of this API, like a module manager, similar to require(). Requires bookkeeping. ES6 spec has abstract module record (for import/export). Modules have state, need to have a handle related to that state. Native bindings also need this. Would it make sense to add something to jerry-core that implements abstract module record? So we can add import/require()/native bindings on top of it."

I remember @zherczeg or someone else from the Szeged folks saying something along the lines of: let's put it into `jerryx` first, we can later move it into `jerry-core`.

My hope is that by putting the basic foundation of module management into JerryScript, that it will become easier to mix and match native binding code from different projects.

ES2015 modules, `import`, etc is a lot more work to build out, so I'm considering this just a start of it.

gabrielschulhof:
@glistening for example, we could write a module resolver for text files (JS source code):
```c
JERRYX_MODULE_RESOLVER(load_source_code)
{
  jerry_value_t ret = 0;
  FILE *module_file = NULL;
  const char absolute_path[4096] = MODULES_ROOT_DIRECTORY;
  strncpy (&absolute_path[sizeof (MODULES_ROOT_DIRECTORY)],
    name, 4096 - sizeof (MODULES_ROOT_DIRECTORY));
  module_file = fopen(absolute_path, "r");
  if (!module_file)
  {
    return 0;
  }
  /* load and ret = jerry_eval () the file */
  fclose (module_file);
  return ret;
}
```
Now, if the text file is an ES6 module, then `ret` will be the result of having loaded an ES6 module, and if the module has imports, then the `load_source_code ()` module resolver will be called recursively, or iteratively from a main loop, or whatever.

glistening:
@gabrielschulhof, @martijnthe  Oh, thank you for the history and explanatory examples.

gabrielschulhof:
@martijnthe we needn't necessarily move the module resolution into jerry-core. We need only that jerry-core expose a hook for informing that some JS code has requested a module.

gabrielschulhof:
Oh, OK, so in my example above we probably can't use `jerry_eval ()` because JS' `eval ()` is [not supposed to be able to handle modules](http://2ality.com/2014/09/es6-modules-final.html#eval-and-modules). So, we need a new Jerry API like `jerry_eval_module (const jerry_char_t *module_definition)` which understands kerywords like `import` and `export` and such.

The bottom line is that I think interpreting new keywords and integrating the concept of modules into the JS engine is something that needs to be done in core, whereas the mechanisms for resolving modules, whether native or text, can be "hooked out" to extension(s).

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,116 @@
...
+
+**Summary**
+
+Load a copy of a module into the current context or return one that was already loaded if it is found.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name);
```
zherczeg: What about `jerry_x_get_module` ?

jerry-ext/include/jerryscript-ext/section.impl.h
```diff
@@ -0,0 +1,84 @@
...
+#ifndef JERRYX_SECTION_IMPL_H
+#define JERRYX_SECTION_IMPL_H
+
+/**
+ * Define the name of a section
+ *
+ * @param name the name of the section without quotes
+ */
+#ifdef __MACH__
+#define JERRYX_SECTION_NAME(name) "__DATA," #name
```
zherczeg: What about those systems where this option is not available?

jerry-ext/include/jerryscript-ext/section.impl.h
```diff
@@ -0,0 +1,84 @@
...
+ * @p item_p as follows:
+ *
+ * @p index will be set to the index of the current item.
+ * @p item_p will point to the current item.
+ *
+ * Usage example:
+ * @code
+ * int index;
+ * widget_t *current_widget_p;
+ * for JERRYX_SECTION_ITERATE (widget_array, index, current_widget_p)
```
zherczeg: For me this is a bit ugly. We should invent a more common macro.

jerry-ext/module/module.c
```diff
@@ -0,0 +1,215 @@
...
+  jerry_value_t property_name = jerry_create_string (module_name_property_name);
+  jerry_value_t property_value = jerry_create_string_from_utf8 (module_name);
+
+  jerry_release_value (jerry_set_property (ret, property_name, property_value));
+  jerry_release_value (property_name);
+  jerry_release_value (property_value);
+  return ret;
+} /* jerryx_module_create_error */
+
+
```
zherczeg: Single newline after functions.

jerry-ext/module/module.c
```diff
@@ -0,0 +1,215 @@
...
+ *
+ * @return the module on success, otherwise the error encountered when attempting to cache. In the latter case, the
+ * @p module is released.
+ */
+static jerry_value_t
+jerryx_module_add_to_cache (jerry_value_t cache, /**< cache to which to add the module */
+                            jerry_value_t module_name, /**< key at which to cache the module */
+                            jerry_value_t module) /**< the module to cache */
+{
+  jerry_value_t ret = jerry_set_property (cache, module_name, module);
```
zherczeg: Do you use a jerry object for caching module names? It is a bit inefficient if many modules are loaded.

jerry-ext/module/module.c
```diff
@@ -0,0 +1,215 @@
...
+
+  /* Return the cached instance if present. */
+  ret = jerryx_module_check_cache (instances, module_name);
+  if (ret != 0)
+  {
+    goto done;
+  }
+
+  /* Try each resolver in the section until one manages to return non-zero. */
+  for JERRYX_SECTION_ITERATE (jxm_resolvers, index, resolver_p)
```
zherczeg: What about dynamic modules (i.e. comes from a JS file).

martijnthe:
@janjongboom could you / other mbedjs people take a look at this and provide feedback? It's the foundation for a native module system. If everyone starts using it, it will hopefully make it easier to re-use native module code in different projects.

jerry-ext/module/module.c
gabrielschulhof: That can be implemented separately: https://github.com/jerryscript-project/jerryscript/pull/1863#issuecomment-304797823

jerry-ext/module/module.c
gabrielschulhof: Using a jerry object avoids having to use `jmem_heap_alloc_block ()`. I can certainly revert to using `jmem_heap_alloc_block ()` but what data structure should I use, especially if we're considering that there may be many modules? I was using a linked list before, but that would probably be even more inefficient.

I'm certainly happy to implement, like a prefix tree or something.

zherczeg: Hm, this is an implementation detail ,so we can change it anytime later. We could start using objects.

docs/11.EXT-REFERENCE-MODULE.md
gabrielschulhof: The convention for extension seems to be `jerryx_<extension>_what_the_api_does ()`. Also, `resolve` is used in the spec to refer to loading modules.

jerry-ext/module/module.c
zherczeg: We still need a plan from API perspective.

jerry-ext/include/jerryscript-ext/section.impl.h
gabrielschulhof: You mean for systems that do not support linker sections?

An alternative for such systems is to use a constructor, for which we now have support in jerry-libc. However, constructors are also stored in a special linker section, so that too might not work either.

An uglier alternative if all else fails is to use code generation and parse the source code to extract `JERRYX_MODULE()` declarations.

Either way, we can define `JERRYX_MODULE()` in various different platform-specific ways.

jerry-ext/module/module.c
gabrielschulhof: That's true, but, aside from support for ES6 modules syntax (like `import` and `export`) I don't think we need anything new from core. So, we just need to agree to throw our support behind a convention for loading modules. We should consider supporting Node.js and AMD.

For example, if we do what Node.js does, then we wrap the contents of a .js file in
```JS
(function (exports, require, module, __filename, __dirname) {
/* contents of JS file */
});
```
And then
```c
jerry_value_t the_function = jerry_eval (wrapped_file_contents);
jerry_call_function (the_function, ...);
```
and retrieve the exports.

AMD uses a different approach. For AMD we need not wrap the original JS in anything, but we need to provide the `define ()` method.
```JS
define( /* optional string with the module name */, [ "dep1", "dep2", "dep3" ],
  function( dep1, dep2, dep3 ) {
    /* Module body goes here. dep1, dep2, and dep3 are loaded asynchronously,
     * and the return value is the module. */
  } );
```

There's an effort to unify all of these into a single universal module definition: https://github.com/umdjs/umd

This is all a bit hairy, to say the least.

jerry-ext/include/jerryscript-ext/section.impl.h
```diff
@@ -0,0 +1,54 @@
...
+#define JERRYX_SECTION_ATTRIBUTE(name) \
+  __attribute__ ((used, section (JERRYX_SECTION_NAME (name)), aligned (sizeof (void *))))
+
+/**
+ * Declare references to a section that contains an array of items.
+ *
+ * @param name the name of the section (without quotes)
+ * @param type the type of the elements stored in the array
+ *
+ * This macro needs to be placed before any occurrence of JERRYX_SECTION_ITERATE, for a given section.
```
jiangzidong: `This macro needs to be placed before any iterations inside a given section`

jerry-ext/module/module.c
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * Initialize the module manager extension.
+ */
+static void
+jerryx_module_manager_init (void *user_data_p)
+{
+  jerry_value_t context = jerry_create_object ();
+  *((jerry_value_t *) user_data_p) = jerry_acquire_value (context);
+  jerry_release_value (context);
+} /* jerryx_module_manager_init */
```
jiangzidong: why not just
```
{
*((jerry_value_t *) user_data_p) = jerry_create_object ();
}
```

jerry-ext/module/module.c
```diff
@@ -0,0 +1,218 @@
...
+ */
+JERRYX_SECTION_DECLARE (jxm_modules, jerryx_module_t)
+
+/**
+ * Attempt to retrieve a module by name from a cache, and return zero if not found.
+ */
+static jerry_value_t
+jerryx_module_check_cache (jerry_value_t cache, /**< cache from which to attempt to retrieve the module by name */
+                          jerry_value_t module_name) /* JerryScript string value holding the module name */
+{
```
jiangzidong: What about using `jerry_get_property (cache, module_name) `, if ret is not undefined, then the module is in tha cache, otherwise not.

jerry-ext/module/module.c
```diff
@@ -0,0 +1,218 @@
...
+  /* Look for the module by its name in the list of module definitions. */
+  for (index = 0, module_p = &__start_jxm_modules[0];
+  &__start_jxm_modules[index] < __stop_jxm_modules;
+  index++, module_p = &__start_jxm_modules[index])
+  {
+    if (!strcmp ((char *) module_p->name, (char *) name))
+    {
+      /* If we find the module by its name we load it and cache it if it has an on_resolve () and complain otherwise. */
+      ret = (module_p->on_resolve) ? module_p->on_resolve () :
+      jerryx_module_create_error (JERRY_ERROR_TYPE, on_resolve_absent, name);
```
jiangzidong: The indention is confusing,

jerry-ext/module/module.c
```diff
@@ -0,0 +1,218 @@
...
+  if (ret != 0)
+  {
+    goto done;
+  }
+
+  /* Try each resolver in the section until one manages to return non-zero. */
+  for (index = 0, resolver_p = &__start_jxm_resolvers[0];
+  &__start_jxm_resolvers[index] < __stop_jxm_resolvers;
+  index++, resolver_p = &__start_jxm_resolvers[index])
+  {
```
jiangzidong: Why do we need multiple resolver-functions? And How can user make sure the order those functions?

I think one resolver function is enough, and user can set its own one if they like.

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+## JERRYX_MODULE_RESOLVER
+
+**Summary**
+
+Define a module resolver. An invocation of this macro is to be followed by the body of a module resolver. The macro
+declares a function of the name given in `cb_name` which accepts a single parameter `const jerry_char_t *name`. It is
+expected that the function returns a non-zero value of type `jerry_value_t` if the resolver is able to associate `name`
+with such a value, or zero if it is not. If the function returns non-zero `jerryx_module_resolve ()` will cache the
```
gabrielschulhof: 0 indicates that that the resolver was unable to resolve the module and that the next resolver should be consulted.

gabrielschulhof: Nevertheless, I can change the signature to return `bool` and the result in an out parameter.

jerry-ext/module/module.c
gabrielschulhof: I wasn't sure actually. I thought `jerry_acquire_value ()` was needed to keep the value around indefinitely.

docs/11.EXT-REFERENCE-MODULE.md
jiangzidong: "return bool or error obj" is better to me

jerry-ext/module/module.c
gabrielschulhof: Actually, a module could intentionally evaluate to `undefined`. The result of loading a module could be anything, including `undefined`. That is, the entire gamut of possible JS values must be available to modules. We cannot pick a possible value and reserve it for ourselves.

jerry-ext/module/module.c
jiangzidong: I think they are equivalent.
object created by `jerry_create_object ` will never be GCed unless you call the `jerry_release_value`.

jerry-ext/module/module.c
gabrielschulhof: Sorry! If I add two spaces before `jerryx_module_create_error` then check-vera.sh complains.

jerry-ext/module/module.c
jiangzidong: Yes, that's why I suggest return error if there is no proper resolver.  Your bool way should also work.

BTW, why do we need multiple resolvers?

jerry-ext/module/module.c
jiangzidong: what about this way
```
a = (b ? c
       : d);
```
which is the example of `08.CODING-STANDARDS.md`

jerry-ext/module/module.c
gabrielschulhof: The idea is that you can have multiple module systems. For example, one resolver takes care of native modules, whereas another takes care of loading text files.

I guess we could chain them if we do something like
```c
typedef jerry_value_t (jerryx_module_resolver_t) (const jerry_char_t *);
jerryx_module_resolver_t
jerryx_module_set_resolver (jerryx_module_resolver_t new_resolver);
```
where `jerryx_module_set_resolver ()` returns the previous resolver - kinda like signal handlers.
The order must not matter.

jerry-ext/module/module.c
gabrielschulhof: For multiple module formats. In fact, my chaining idea from below probably won't work, because we need to assemble the chain at some point.

jerry-ext/module/module.c
jiangzidong: Though you said `order must not matter`, but people may provides resolves whose older matters, and it can work well. But when other rebuild the same project, the order may change and resulting in strange error.

Sorry that I don't totally understand the reason why we need multiple resolver in one project. Could you give me an example?

gabrielschulhof: One example is the built-in loader that uses the list of modules from the section `jxm_modules`. Another example is the one I gave @glistening [above](https://github.com/jerryscript-project/jerryscript/pull/1863#issuecomment-304797823).

As for evaluation order, the idea is that no two resolvers resolve the same module. If somebody adds module "xyzzy.js" to their project, then adds another module "xyzzy.c" which uses `JERRYX_MODULE()` internally, then it is undefined which one they're going to get when they attempt to `jerryx_module_resolve ("xyzzy")`.

jerry-ext/module/module.c
gabrielschulhof: It complains about that too. The best it lets me do is
```
a = (b ? c
: d);
```

jiangzidong: @zherczeg Do you know why?

janjongboom:
@gabrielschulhof @martijnthe Will look at it in the next few days. Traveling right now. But I applaud the idea!

jiangzidong:
@gabrielschulhof Can I say that all modules which use JERRYX_MODULE to register must be handled by native_resolver, right?

gabrielschulhof:
@jiangzidong correct.

jerry-ext/module/module.c
martijnthe: https://github.com/jerryscript-project/jerryscript/pull/1863#issuecomment-304797823 is IMO a very good reason to want to have multiple resolvers.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,55 @@
...
+ */
+jerry_value_t jerryx_module_resolve (const jerry_char_t *name);
+
+/**
+ * Define a module. The section name is cryptic ("jxm") because it has to be less than 16 bytes on Mach-O.
+ */
+#define JERRYX_MODULE(module_name, on_resolve_cb)                        \
+  static jerryx_module_t _module JERRYX_SECTION_ATTRIBUTE(jxm_modules) = \
+  {                                                                      \
+    .name = ((jerry_char_t *)#module_name),                              \
```
LaszloLango: Missing space: `((jerry_char_t *) #module_name)`

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,55 @@
...
+    .on_resolve = (on_resolve_cb)                                        \
+  };
+
+/**
+ * Define a module resolver. The section name is cryptic ("jxm") because it has to be less than 16 bytes on Mach-O.
+ */
+#define JERRYX_MODULE_RESOLVER(cb_name)                                              \
+  static bool cb_name (const jerry_char_t *name, jerry_value_t *result);             \
+  static bool (*__static_global_ ## cb_name) (const jerry_char_t *, jerry_value_t *) \
+  JERRYX_SECTION_ATTRIBUTE(jxm_resolvers) = cb_name;                                 \
```
LaszloLango: ditto: `JERRYX_SECTION_ATTRIBUTE (jxm_resolvers)`

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,117 @@
...
+```c
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name);
+```
+
+- `name` - the name of the module to load.
+- return value - `jerry_value_t` representing the module that was loaded, or the error that occurred in the process.
+
+
+# Modules data types
```
zherczeg: Module data types

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Load a copy of a module into the current context or return one that was already loaded if it is found.
+ */
+jerry_value_t jerryx_module_resolve (const jerry_char_t *name);
+
+/**
+ * Define a module. The section name is cryptic ("jxm") because it has to be less than 16 bytes on Mach-O.
+ */
+#define JERRYX_MODULE(module_name, on_resolve_cb)                        \
+  static jerryx_module_t _module JERRYX_SECTION_ATTRIBUTE(jxm_modules) = \
```
zherczeg: Could this be const data?

jerry-ext/module/module.c
```diff
@@ -0,0 +1,215 @@
...
+ */
+JERRYX_MODULE_RESOLVER (native_resolver)
+{
+  int index;
+  const jerryx_module_t *module_p = NULL;
+
+  /* Look for the module by its name in the list of module definitions. */
+  for (index = 0, module_p = &__start_jxm_modules[0];
+  &__start_jxm_modules[index] < __stop_jxm_modules;
+  index++, module_p = &__start_jxm_modules[index])
```
zherczeg: This code does not follow the style.

jerry-ext/module/module.c
```diff
@@ -0,0 +1,215 @@
...
+    ret = module;
+  }
+
+  return ret;
+} /* jerryx_module_add_to_cache */
+
+/**
+ * Declare and define the default module resolver - one which examines what modules are defined in the above linker
+ * section and loads one that matches the requested name, caching the result for subsequent requests using the context
+ * data mechanism.
```
zherczeg: Does this work if no static modules are defined?

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,55 @@
...
+  };
+
+/**
+ * Define a module resolver. The section name is cryptic ("jxm") because it has to be less than 16 bytes on Mach-O.
+ */
+#define JERRYX_MODULE_RESOLVER(cb_name)                                              \
+  static bool cb_name (const jerry_char_t *name, jerry_value_t *result);             \
+  static bool (*__static_global_ ## cb_name) (const jerry_char_t *, jerry_value_t *) \
+  JERRYX_SECTION_ATTRIBUTE (jxm_resolvers) = cb_name;                                 \
+  static bool cb_name (const jerry_char_t *name, jerry_value_t *result)
```
zherczeg: What is the priority order of the module resolvers? In case if multiple resolvers reacts to the same name.

gabrielschulhof: There is no priority order. The first one that resolves the module wins. If multiple resolvers provide the same module, it is undefined which one you're going to get. It's up to project maintainers to keep their list of modules clean.

I've added a `*Note*:` into the doc.

jerry-ext/module/module.c
gabrielschulhof: Good question. I'll add a test.

jerry-ext/module/module.c
gabrielschulhof: Skimming over the coding standards doc I don't see any section that applies to long for-statement headers. How should I indent this? Keep in mind that `check-vera.sh` does not now complain about indentation.

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: For example, if somebody wants to load module "http", and two differently structured modules resolved by different resolvers both provide "http", then you're gonna get "http" one way or the other. If you're confused as to why the JS API provided by the resulting module is totally different from what you expected, well, check your build scripts and don't load two of the same thing into your project.

gabrielschulhof: Note that in npm this is not a problem, since each package has a unique name, and there are already two different package structures: native addon packages, and plain JS packages.

jerry-ext/module/module.c
gabrielschulhof: [Added](https://github.com/jerryscript-project/jerryscript/pull/1863/files#diff-56a70d8121afefa3ff42d36017116ef3)

jerry-ext/module/module.c
gabrielschulhof: NM. Found it. Let's see if check-vera.sh allows it.

jerry-ext/include/jerryscript-ext/module.h
zherczeg: For a generic module selector, defining the order of resolvers is very important. We cannot expect that names are unique, and we need to let people to define resolve order for symbol clashes. E.g. a module on a file system has a higher priority than a builtin module, or vice versa. We need to figure out something for this.

martijnthe: Hmm, yeah I tend to agree that I'd control over the order in which the resolvers are called.
I'm mostly in interested in having one resolver for native modules and one for source file modules. But I don't want an app's JS code to be able to "override" native module with a source file module. Therefore, the native module resolver should be run before the source file module resolver. For my use case it's fine to decide the priority at build-time, I don't have a need to resolve clashes at runtime or anything like that.

Maybe add a priority argument to `JERRYX_MODULE_RESOLVER`, so one can write something like this?

```
#ifndef MY_RESOLVER_PRIORITY
#  define MY_RESOLVER_PRIORITY (1)
#endif

JERRYX_MODULE_RESOLVER(my_resolver, MY_RESOLVER_PRIORITY)
{
...
}
```

gabrielschulhof: I guess the resolve order is a bit like a search path for modules. For example, first we search the linker section, then we search the file system.

We could give resolvers names, and accept a comma-separated list of names. We could then re-arrange the `"jxm_resolvers"` linker section so that, the next time around, presuming that the requested order will be the same, the resolvers will be in the correct order.

So, like,
```c
jerryx_module_resolve("name", "built-in,js-fs-loader");
```
`NULL` would mean whatever order is there. If there are more resolvers than there are items in the list, and the listed ones are unable to resolve the module, the rest would be consulted in whatever order they're left in the section after the rearrangement completes.

This would make the resolver kinda like a meta-module.

martijnthe: Passing the ordered list of resolvers to use to the `jerryx_module_resolve` would work for my purposes. I'm not a huge fan of passing the resolvers as a comma separated list of strings though..
Maybe it makes sense make the resolver a visible/usable symbol, as opposed to being static, and just it directly in the `jerryx_module_resolve` call?

gabrielschulhof: That could become unwieldy:
```c
static const jerryx_module_resolver_t resolvers = {
#ifdef HAVE_JS_FS_RESOLVER
  jerryx_module_resolver_js_fs,
#endif /* HAVE_JS_FS_RESOLVER */
#fidef HAVE_SOME_OTHER_RESOLVER
  jerryx_module_some_other_resolver,
#endif /* HAVE_SOME_OTHER_RESOLVER */
};
```
and
```c
jerry_value_t = jerryx_module_resolve("something",
                                      resolvers,
                                      sizeof(resolvers)/sizeof(resolvers[0]));
```
Though I highly doubt that the array of resolvers will go too-too large.

gabrielschulhof: TBH I think passing `NULL` as the second parameter would work 99.9% of the time, because any given module is provided by exactly one of the resolvers.

gabrielschulhof: A completely different alternative would be not to expose the resolver at all, and to maintain its capabilities as part of this extension ourselves.

martijnthe: > That could become unwieldy

Doesn't look too bad to me...

Maybe it makes sense to have 2 separate APIs, `jerryx_module_resolve` (simple, no control over ordering) and `jerryx_module_resolve_using_resolvers` (complex, control over ordering) or something.

gabrielschulhof: I'll go with the array and the default `NULL` if it's not that unwieldy. I'll have to change `JERRYX_MODULE_RESOLVER` to accept an existing function rather than declaring a new function.

martijnthe: cool, thanks

zherczeg: I see the reason behind a random order for built-in modules (stored in a linker section for simplicity), but the number of resolvers should be low (1-3 maximum), so I would prefer a static list of them. This list (and its length) should be passed when the resolvers are initialized. The general resolver simply goes through this list.

gabrielschulhof: OK, there's another wrinkle: The results of module resolution are cached. So, if you pass the same `name` to `jerryx_module_resolve ()` at a later time, you're going to get whatever module was loaded the first time around, even if you specify a different order of resolvers, which would result in a different module getting cached if this were the first time around.

bzsolt:
@gabrielschulhof could you please refer to the new docfile in the [tools/update-webpage.sh](https://github.com/jerryscript-project/jerryscript/blob/master/tools/update-webpage.sh)? (this script generates the proper files for the webpage)
Just add the following two lines (or something similar):

```
EXT_REFERENCE_MODULE_MD="11.EXT-REFERENCE-MODULE.md"
...
titles[$EXT_REFERENCE_MODULE_MD]="'Extension API: Module Handling'"
```

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: @zherczeg 
How do you mean "when the resolvers are initialized"? We do not currently have a separate initialization step for the resolvers. With my current (not yet pushed) modification, the list of resolvers is passed to `jerryx_module_resolve ()`, so it becomes
```c
jerry_value_t jerryx_module_resolve (const jerry_char_t *name,
                                     const jerryx_module_resolver_t *resolvers,
                                     size_t count);
```

zherczeg: Sounds good!

zherczeg: I don't see any problem with cached resolvers. If we ever find a module, we need to use it forever (in its current state).

gabrielschulhof:
@bzsolt Done.

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,108 @@
...
+
+Load a copy of a module into the current context or return one that was already loaded if it is found.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name, /**< name of the module to load */
+                       jerryx_module_resolver_t *resolvers_p, /**< list of resolvers */
+					   size_t resolver_count); /**< number of resolvers in @p resolvers */
```
LaszloLango: wrong indentation

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * Define a module. The section name is cryptic ("jxm") because it has to be less than 16 bytes on Mach-O.
+ */
+#define JERRYX_MODULE(module_name, on_resolve_cb)                        \
+  static const jerryx_module_t _module JERRYX_SECTION_ATTRIBUTE(jxm_modules) = \
+  {                                                                      \
+    .name = ((jerry_char_t *) #module_name),                              \
+    .on_resolve = (on_resolve_cb)                                        \
+  };
```
LaszloLango: Please align the backslash characters to the same column

gabrielschulhof:
@zherczeg I switched to a list of resolvers explicitly passed. I got rid of the `"jxm_resolvers"` section, so the list cannot be empty.

jerry-ext/module/module.c
```diff
@@ -0,0 +1,213 @@
...
+ */
+bool jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module */
+                                    jerry_value_t *result) /**< out - where to put the resulting module instance */
```
LaszloLango: `jerry_value_t *result) /**< [out] where to put the resulting module instance */`

jerry-ext/module/module.c
```diff
@@ -0,0 +1,213 @@
...
+  /* Look for the module by its name in the list of module definitions. */
+  for (index = 0, module_p = &__start_jxm_modules[0];
+       &__start_jxm_modules[index] < __stop_jxm_modules;
+       index++, module_p = &__start_jxm_modules[index])
+  {
+    if (!strcmp ((char *) module_p->name, (char *) name))
+    {
+      /* If we find the module by its name we load it and cache it if it has an on_resolve () and complain otherwise. */
+      (*result) = ((module_p->on_resolve) ? module_p->on_resolve ()
+      : jerryx_module_create_error (JERRY_ERROR_TYPE, on_resolve_absent, name));
```
LaszloLango:
```c
(*result) = ((module_p->on_resolve) ? module_p->on_resolve ()
                                    : jerryx_module_create_error (JERRY_ERROR_TYPE, on_resolve_absent, name));
```

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,108 @@
+#  Module API
+
```
jiangzidong: Could you also introduce the concept of "module" and "module reducer" here

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,108 @@
...
+## jerryx_module_resolve
+
+**Summary**
+
+Load a copy of a module into the current context or return one that was already loaded if it is found.
+
```
jiangzidong: We should also tell the user the meaning  of the order in resolvers_list

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,108 @@
...
+The resolver for JerryScript modules. A pointer to this function can be passed in the second parameter to `jerryx_module_resolve` to search for the module among the JerryScript modules built into the binary.
+
+**Prototype**
+
+```c
+bool jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module */
+                                    jerry_value_t *result) /**< out - where to put the resulting module instance */
+```
+- `name` - the name of the module to find
+- `result` - out - place where to store the resulting JS value
```
jiangzidong: `place where to store the resulting module instance`

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,108 @@
...
+typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+```
+
+# Module helper macros
+
+## JERRYX_MODULE
+
+**Summary**
+
+Helper macro to define a module. Declares a variable of type `jerryx_module_t` as global static and places it into the
```
jiangzidong: `Helper macro to define a JerryScript module`

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,108 @@
...
+
+
+# Module data types
+
+## jerryx_module_t
+
+**Summary**
+
+Structure to define a module
+
```
jiangzidong: `module` -> `JerryScript module`

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,108 @@
...
+- `resolvers_p` - the list of resolvers to call
+- `resolver_count` - the number of resolvers in `resolvers_p`
+- return value - `jerry_value_t` representing the module that was loaded, or the error that occurred in the process.
+
+
+## jerryx_module_native_resolver
+
+**Summary**
+
+The resolver for JerryScript modules. A pointer to this function can be passed in the second parameter to `jerryx_module_resolve` to search for the module among the JerryScript modules built into the binary.
```
jiangzidong: `The resolver for JerryScript modules` , we should put it clearer. 
1) what is jerryscript module? I think it is **ONLY** the module which is declared by `JERRYX_MODULE`, right? (We can also put the description in the beginning of the doc)
2) For other types of module, user should provide their own resolver

jerry-ext/module/module.c
```diff
@@ -0,0 +1,213 @@
...
+};
+
+/**
+ * Declare the linker section where module definitions are stored.
+ */
+JERRYX_SECTION_DECLARE (jxm_modules, jerryx_module_t)
+
+/**
+ * Attempt to retrieve a module by name from a cache, and return zero if not found.
+ */
```
jiangzidong: return false

tests/unit-ext/module-empty.c
```diff
@@ -0,0 +1,70 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
```
jiangzidong: what about renaming `module-empty` => "test-ext-module-empty", so that all the unittest case under /unit-ext have similar name pattern

docs/11.EXT-REFERENCE-MODULE.md
gabrielschulhof: @jiangzidong please see below.

docs/11.EXT-REFERENCE-MODULE.md
gabrielschulhof: @jiangzidong please see below.

docs/11.EXT-REFERENCE-MODULE.md
gabrielschulhof: @jiangzidong the introduction at the top of the file now covers both these topics.

tools/update-webpage.sh
```diff
@@ -32,6 +32,7 @@ DEBUGGER_MD="07.DEBUGGER.md"
 CODING_STANDARDS_MD="08.CODING-STANDARDS.md"
 EXT_REFERENCE_ARG_MD="09.EXT-REFERENCE-ARG.md"
 EXT_REFERENCE_HANDLER_MD="10.EXT-REFERENCE-HANDLER.md"
+EXT_REFERENCE_MODULE_MD="11.EXT-REFERENCE-MODULE.mpd"
```
bzsolt: Thank you for wiring the doc, but there is a typo in the extension of filename: `mpd` -> `md`.

tools/update-webpage.sh
```diff
@@ -45,6 +46,7 @@ titles[$DEBUGGER_MD]="Debugger"
 titles[$CODING_STANDARDS_MD]="Coding Standards"
 titles[$EXT_REFERENCE_ARG_MD]="'Extension API: Argument Validation'"
 titles[$EXT_REFERENCE_HANDLER_MD]="'Extension API: External Function Handlers'"
+titles[$EXT_REFERENCE_MODULE_MD]="'Extension API: Module support'"
```
bzsolt: Please start the word "support" with capital letter like the others.

gabrielschulhof:
@jiangzidong Thanks!

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,178 @@
...
+Each function in `resolvers_p` will be called in sequence until one returns `true` and fills out its out-parameter with
+the `jerry_value_t` representing the requested module. If the `jerry_value_t` does not have the error flag set it will
+be cached. Thus, on subsequent calls with the same value for `name`, none of the functions in `resolvers_p` will be
+called.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name, /**< name of the module to load */
```
LaszloLango: Please remove the comments. You described the parameters below.

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,178 @@
...
+## jerryx_module_native_resolver
+
+**Summary**
+
+The resolver for JerryScript modules. A pointer to this function can be passed in the second parameter to `jerryx_module_resolve` to search for the module among the JerryScript modules built into the binary.
+
+**Prototype**
+
+```c
+bool jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module */
```
LaszloLango: ditto

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,178 @@
...
+    /* We allocate enough memory to store the contents of the file. */
+    js_file_contents = malloc (file_size);
+    if (js_file_contents)
+    {
+      /* We read the file into memory and call jerry_eval (), assigning the result to the out-parameter. */
+      fread (js_file_contents, file_size, 1, js_file);
+      (*result) = jerry_eval (js_file_contents, file_size, false);
+
+      /* We release the memory holding the contents of the file. */
+	  free (js_file_contents);
```
zherczeg: Misaligned line.

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,178 @@
...
+    fclose (js_file);
+  }
+
+  return return_value;
+}
+```
+
+We can now load JavaScript files:
+```c
+jerryx_module_resolver_t resolvers =
```
zherczeg: const static

docs/11.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,178 @@
...
+   * defined JavaScript file loader to load it.
+   */
+  load_and_evaluate_js_file
+};
+jerry_value_t js_module = jerryx_module_resolve (requested_module, resolvers, 2);
+```
+
+# Module helper macros
+
+## JERRYX_MODULE
```
zherczeg: These are called native modules above.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,56 @@
...
+    .on_resolve = (on_resolve_cb)                                              \
+  };
+
+/**
+ * Type of a module resolver function
+ */
+typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+
+/**
+ * The implementation of the resolver for modules as defined above
```
zherczeg: Full stop after sentence.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+#define JERRYX_MODULE(module_name, on_resolve_cb)                              \
+  static const jerryx_module_t _module JERRYX_SECTION_ATTRIBUTE(jxm_modules) = \
+  {                                                                            \
+    .name = ((jerry_char_t *) #module_name),                                   \
+    .on_resolve = (on_resolve_cb)                                              \
+  };
+
+/**
+ * Type of a module resolver function
```
zherczeg: Ditto.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,56 @@
...
+ * Structure to define a module
+ */
+typedef struct
+{
+  jerry_char_t *name; /**< name of the module */
+  jerry_value_t (*on_resolve) (void); /**< function that returns a new instance of the module */
+} jerryx_module_t;
+
+/**
+ * Define a module. The section name is cryptic ("jxm") because it has to be less than 16 bytes on Mach-O.
```
zherczeg: What is Mach-O? 3 character is pretty short, we have at least 13 characters more for a bit more descriptive name/

jerry-ext/include/jerryscript-ext/section.impl.h
zherczeg: Ok, so without this macro the whole module system does not work? No workaround at all?

jerry-ext/include/jerryscript-ext/module.h
akosthekiss: Mach-O is Mach object file format

LaszloLango:
Please rebase to the current master

gabrielschulhof:
@LaszloLango I'm working on a way of assembling the array of module definitions by providing a tool that recognizes the pattern `JERRYX_MODULE (...)` and outputting a C file that creates a global variable initialized to the list of modules. I will rebase when I have that in place.

gabrielschulhof:
@LaszloLango this is for the case where `__attribute__((section(...)))` is unavailable - IOW `__GNUC__` is not defined.

jerry-ext/include/jerryscript-ext/section.impl.h
gabrielschulhof: @zherczeg I guess a workaround is possible, but it's hard to test, because I have to build jerry-ext with `-U__GNUC__` (cf. https://github.com/jerryscript-project/jerryscript/issues/1890#issuecomment-311889136)

zherczeg: Hm, come to think of it it would be good overall to test the whole project without GNUC macro. A travis job could be added.

zherczeg: (Btw I am happy with having no native modules if certain compiler features are missing. The loading from file system is still useful.)

gabrielschulhof: @akosthekiss @galpeter @dbatyai @zherczeg @LaszloLango @glistening @martijnthe @jiangzidong @bzsolt @robertsipka @janjongboom So, can we land this without the workaround and add the workaround as part of the solution to https://github.com/jerryscript-project/jerryscript/issues/1890?

gabrielschulhof: @zherczeg I tried precisely that (adding a Travis job) using the command line I mention in my comment, but I'm not familiar enough with the engine to fix those warnings.

gabrielschulhof: @zherczeg Also, `-U__GNUC__` != extensions disabled, because if `__attribute__((something))` appears unconditionally, gcc will still accept that. So, it remains our responsibility to provide for the case where `__GNUC__` is not defined and then to test for that case using travis.

I actually wonder if Ubuntu has a C compiler that does not provide GNU extensions. Then we could do a travis job where we use that as the compiler instead of the default one.

gabrielschulhof:
@zherczeg I have now addressed your review comments. The module system will not work `#ifndef __GNUC__`, although I cannot currently test that because of https://github.com/jerryscript-project/jerryscript/issues/1890#issuecomment-311889136. @LaszloLango I have rebased.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,182 @@
+#  Module API
+
+**Note**: This extension depends on compiler support because it uses `__attribute__` extensions provided only by GCC and
+LLVM/clang. For other compilers, no support has been added yet.
```
zherczeg: Please update this to the current status.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,182 @@
...
+the `jerry_value_t` representing the requested module. If the `jerry_value_t` does not have the error flag set it will
+be cached. Thus, on subsequent calls with the same value for `name`, none of the functions in `resolvers_p` will be
+called.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerryx_module_resolve (const jerry_char_t *name,
+                       *resolvers_p,
```
zherczeg: Type of resolvers?

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,182 @@
...
+Structure to define a JerryScript module
+
+**Prototype**
+
+```c
+typedef struct
+{
+  jerry_char_t *name; /**< name of the module */
+  jerry_value_t (*on_resolve) (void); /**< function that returns a new instance of the module */
+} jerryx_module_t;
```
zherczeg: I think this structure is only used by native modules. Furthermore do we need to make it public? I mean it is needed to be public, but people should never use its members directly. We should emphasize this.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,60 @@
...
+#include "jerryscript-ext/section.impl.h"
+
+/**
+ * Declare the structure used to define a module.
+ */
+typedef struct
+{
+  jerry_char_t *name; /**< name of the module */
+  jerry_value_t (*on_resolve) (void); /**< function that returns a new instance of the module */
+} jerryx_module_t;
```
zherczeg: The native module only part should should be hidden in some way when the support is missing. And than type name should be `jerryx_native_module_t`

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Declare a helper macro that expands to the declaration of a variable of type jerryx_module_t placed into the
+ * specially-named linker section "jerryx_modules" where the JerryScript module resolver
+ * jerryx_module_native_resolver () will look for it.
+ */
+#define JERRYX_MODULE(module_name, on_resolve_cb)                                 \
```
zherczeg: JERRYX_NATIVE_MODULE

docs/12.EXT-REFERENCE-MODULE.md
gabrielschulhof: How do you mean? gcc/clang-only *is* the current status.

docs/12.EXT-REFERENCE-MODULE.md
gabrielschulhof: I guess we need not document this structure, but only the macro. We can then leave a comment in the header emphasizing that the structure not be used directly.

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: I'm not sure we should call it `jerryx_native_module_t` - after all, it's a "JerryScript module" and that's how we document it.

jerry-ext/include/jerryscript-ext/section.impl.h
martijnthe: I'm fine not having the workaround here yet.

docs/12.EXT-REFERENCE-MODULE.md
zherczeg: Please describe what happens when other compilers are used.

docs/12.EXT-REFERENCE-MODULE.md
zherczeg: Ok.

jerry-ext/include/jerryscript-ext/module.h
zherczeg: Is it used for non-native modules?

docs/12.EXT-REFERENCE-MODULE.md
gabrielschulhof: Actually, I just realized that only the JerryScript modules will be affected. If you create modules using some other mechanism and provide a resolver, that will work. So, I'll document *that*.

gabrielschulhof:
It shouldn't be used for non-native modules, because those will have their
own definition mechanism and will have their own resolver too.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,177 @@
+#  Module API
+
+A JerryScript module is defined by calling the macro `JERRYX_MODULE` and supplying it with the name of the module as
+its first parameter and the name of an initialization function as its second parameter.
+
+**Note**: The `JERRYX_MODULE` macro depends on compiler support because it uses `__attribute__` extensions provided
+only by GCC and LLVM/clang to ensure that JerryScript's own module resolver (`jerryx_module_native_resolver`) will
+find a module that was declared with `JERRYX_MODULE`. If such compiler support is absent, currently
+`jerryx_module_native_resolver` will fail to find a module declared with `JERRYX_MODULE`.
```
zherczeg: To tell you the truth I don't really like the compiles but does not work at runtime approach. That is awful to debug. I would prefer a compiler error to raise the attention. What I would prefer: conditional definition of macros / structs / functions which related to native modules, and all of these types should have `native` in their name to make their role clear.

gabrielschulhof: Ha! That's exactly what I had at first :) I just put `#ifdef __GNUC__` at the top of each file involved. OK. I'll add the `native_` and switch back to that approach.

zherczeg: I would alao add a define for this feature when it is enabled, so people can test it at compile time and do a workaround if needed:

    #ifdef __GNUC__
    ...
    #define JERRYX_NATIVE_MODULES_SUPPORTED
    ...
    #endif

gabrielschulhof:
@zherczeg Please have another look. I have defined `JERRYX_NATIVE_MODULES_SUPPORTED` and I have substantially updated the documentation to reflect that native modules are only available when that macro is defined while also emphasizing that the module resolver continues to work even if native modules are unavailable, as long as you provide your own resolver.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,72 @@
...
+
+#ifdef __GNUC__
+#define JERRYX_NATIVE_MODULES_SUPPORTED
+#endif /* __GNUC__ */
+
+#ifdef JERRYX_NATIVE_MODULES_SUPPORTED
+/**
+ * Declare the signature for the module initialization function.
+ */
+typedef jerry_value_t (*jerryx_module_on_resolve_t) (void);
```
zherczeg: Shall we add native prefix to this type?

jerry-ext/include/jerryscript-ext/section.impl.h
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRYX_SECTION_IMPL_H
+#define JERRYX_SECTION_IMPL_H
```
zherczeg: I would add a JERRYX_NATIVE_MODULES_SUPPORTED guard here, or where this file is included.

jerry-ext/include/jerryscript-ext/section.impl.h
```diff
@@ -0,0 +1,56 @@
...
+ * of the section @p name.
+ */
+#ifdef __MACH__
+#define JERRYX_SECTION_DECLARE(name, type)                                   \
+  extern const type __start_ ## name[] __asm("section$start$__DATA$" #name); \
+  extern const type __stop_ ## name[] __asm("section$end$__DATA$" #name);
+#else /* !__MACH__ */
+#define JERRYX_SECTION_DECLARE(name, type)                    \
+  extern const type __start_ ## name[] __attribute__((weak)); \
+  extern const type __stop_ ## name[] __attribute__((weak));
```
zherczeg: This is just a question: why these symbols are weak? Do we expect them to be overwritten by strong symbols?

jerry-ext/module/module.c
```diff
@@ -0,0 +1,216 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <string.h>
+#include "jerryscript.h"
+#include "jerryscript-ext/module.h"
+
+static const jerry_char_t *module_name_property_name = (jerry_char_t *) "moduleName";
+static const jerry_char_t *on_resolve_absent = (jerry_char_t *) "Module on_resolve () must not be NULL";
```
zherczeg: Am I correct that this error text is used when native modules are used? We should not define unused symbols.

jerry-ext/module/module.c
```diff
@@ -0,0 +1,216 @@
...
+  /* Look for the module by its name in the list of module definitions. */
+  for (index = 0, module_p = &__start_jerryx_modules[0];
+       &__start_jerryx_modules[index] < __stop_jerryx_modules;
+       index++, module_p = &__start_jerryx_modules[index])
+  {
+    if (!strcmp ((char *) module_p->name, (char *) name))
+    {
+      /* If we find the module by its name we load it and cache it if it has an on_resolve () and complain otherwise. */
+      (*result) = ((module_p->on_resolve) ? module_p->on_resolve ()
+                                          : jerryx_module_create_error (JERRY_ERROR_TYPE, on_resolve_absent, name));
```
zherczeg: Do we complain about NULL onresolve, but not about NULL names? It is a bit inconsistent to me. Hm, I am fine to ignore modules without names (does strcmp supports NULL arguments?).

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,187 @@
...
+
+```c
+typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+```
+
+**Example**
+```c
+/* An example resolver for JavaScript files */
+bool
+load_and_evaluate_js_file(const jerry_char_t *name, jerry_value_t *result)
```
LaszloLango: missing space before `(`

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,187 @@
...
+
+**Prototype**
+
+```c
+typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+```
+
+**Example**
+```c
+/* An example resolver for JavaScript files */
```
LaszloLango: I think this comment is unnecessary here.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -0,0 +1,187 @@
...
+  {
+    /* We have successfully opened the file. Now, we establish its size. */
+    file_size = fseek (js_file, 0, SEEK_END);
+    fseek (js_file, 0, SEEK_SET);
+
+    /* We allocate enough memory to store the contents of the file. */
+    js_file_contents = malloc (file_size);
+    if (js_file_contents)
+    {
+      /* We read the file into memory and call jerry_eval (), assigning the result to the out-parameter. */
```
LaszloLango: Not sure why, but GitHub marks the end of the line with a red color for me. Is there a trailing whitespace at the end of the line?

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,72 @@
...
+} jerryx_native_module_t;
+
+/**
+ * Declare a helper macro that expands to the declaration of a variable of type jerryx_native_module_t placed into the
+ * specially-named linker section "jerryx_modules" where the JerryScript module resolver
+ * jerryx_module_native_resolver () will look for it.
+ */
+#define JERRYX_NATIVE_MODULE(module_name, on_resolve_cb)                                 \
+  static const jerryx_native_module_t _module JERRYX_SECTION_ATTRIBUTE(jerryx_modules) = \
+  {                                                                               \
```
LaszloLango: Please put the backslash characters to the same column.

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: ✓

jerry-ext/include/jerryscript-ext/section.impl.h
gabrielschulhof: `jerryscript-ext/module.h` refers to this file. I have moved that reference to after the `#ifdef JERRYX_NATIVE_MODULES_SUPPORTED`.

jerry-ext/include/jerryscript-ext/section.impl.h
gabrielschulhof: I believe it's so that if there is no such section, then their value is `NULL` and iterating over the array will short-circuit.

jerry-ext/module/module.c
gabrielschulhof: I have moved this into the `#ifdef JERRYX_NATIVE_MODULES_SUPPORTED` block.

jerry-ext/module/module.c
gabrielschulhof: Good point! I don't think a `NULL` name is OK, because it cannot be requested. I mean, sure, in JS you can write `var nullModule = require(null);` but c'mon, seriously, I can't imagine that being a feature somebody would want.

gabrielschulhof: OTOH, I don't think the resolver should throw an error if it encounters a module with a `NULL` name because that would poison the rest of the linker section, preventing *any* native modules following the defective module from being loaded. So, I will simply add a `NULL`-check before the `strcmp()` because I don't believe it's `NULL`-safe.

docs/12.EXT-REFERENCE-MODULE.md
gabrielschulhof: Nope. I wager it's some kind of parse error.

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: ✓

gabrielschulhof:
@zherczeg @LaszloLango I have addressed your review comments.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,72 @@
...
+
+#include "jerryscript.h"
+
+#ifdef __GNUC__
+#define JERRYX_NATIVE_MODULES_SUPPORTED
+#endif /* __GNUC__ */
+
+#ifdef JERRYX_NATIVE_MODULES_SUPPORTED
+#include "jerryscript-ext/section.impl.h"
+/**
```
zherczeg: Newline before comment.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -0,0 +1,72 @@
...
+  {                                                                                      \
+    .name = ((jerry_char_t *) #module_name),                                             \
+    .on_resolve = (on_resolve_cb)                                                        \
+  };
+
+/**
+ * Declare the JerryScript module resolver so that it may be added to an array of jerryx_module_resolver_t items and
+ * thus passed to jerryx_module_resolve.
+ */
+bool jerryx_module_native_resolver (const jerry_char_t *name, jerry_value_t *result);
```
zherczeg: Newline after function declaration.

jerry-ext/include/jerryscript-ext/section.impl.h
```diff
@@ -0,0 +1,56 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
```
zherczeg: The name section.impl.h does not refer to modules. It is clear that It is an internal header and should not be used, but it is still part of the public API.

jerry-ext/include/jerryscript-ext/section.impl.h
gabrielschulhof: https://github.com/jerryscript-project/jerryscript/pull/1863/files#diff-9ae8ab785adef517e883b35c916b144cR26

jerry-ext/include/jerryscript-ext/section.impl.h
gabrielschulhof: I didn't intend for it to be related to modules. It contains macros that help with linker sections, whatever purpose those linker sections may serve.

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: Added

LaszloLango:
LGTM

PR #1869, opened 2017-05-31
[nuttx-stm32f4] Set log level to JERRY_LOG_LEVEL_DEBUG when using show-opcodes

robertsipka:
It provides detailed output in case of enabled parser byte-code dumps.


yichoi:
LGTM

PR #1872, opened 2017-06-01
Add command line options to exception and display commands

DanielBallaSZTE:
You can now set --display and --exception parameters in the command line.

### DIFF
--- a/jerry-debugger/jerry-client-ws.py
+++ b/jerry-debugger/jerry-client-ws.py
@@ -81,6 +81,10 @@ def arguments_parse():
                         help="disable stop when newline is pressed (default: %(default)s)")
     parser.add_argument("--color", action="store_true", default=False,
                         help="enable color highlighting on source commands (default: %(default)s)")
+    parser.add_argument("--display", action="store", default=None, type=int,
+                        help="set display range")
+    parser.add_argument("--exception", action="store", default=None, type=int, choices=[0, 1],
+                        help="set exception config, usage 1: [Enable] or 0: [Disable]")
 
     args = parser.parse_args()
 
@@ -893,6 +897,12 @@ def main():
     prompt = DebuggerPrompt(debugger)
     prompt.prompt = "(jerry-debugger) "
     prompt.non_interactive = non_interactive
+
+    if args.display:
+        prompt.do_display(args.display)
+
+    if args.exception:
+        prompt.do_exception(args.exception)
 
     while True:
         if not non_interactive and prompt.cont:


jerry-debugger/jerry-client-ws.py
```diff
@@ -81,6 +81,10 @@ def arguments_parse():
                         help="disable stop when newline is pressed (default: %(default)s)")
     parser.add_argument("--color", action="store_true", default=False,
                         help="enable color highlighting on source commands (default: %(default)s)")
+    parser.add_argument("--display", action="store", default="0",
+                        help="set display range (default: %(default)s)")
+    parser.add_argument("--exception", action="store", default="0",
```
zherczeg: Isn't 1 is the default for exception?

jerry-debugger/jerry-client-ws.py
```diff
@@ -81,6 +81,10 @@ def arguments_parse():
                         help="disable stop when newline is pressed (default: %(default)s)")
     parser.add_argument("--color", action="store_true", default=False,
                         help="enable color highlighting on source commands (default: %(default)s)")
+    parser.add_argument("--display", action="store", default="0",
```
zherczeg: I think display should have an empty default.

jerry-debugger/jerry-client-ws.py
```diff
@@ -81,6 +81,10 @@ def arguments_parse():
                         help="disable stop when newline is pressed (default: %(default)s)")
     parser.add_argument("--color", action="store_true", default=False,
                         help="enable color highlighting on source commands (default: %(default)s)")
+    parser.add_argument("--display", action="store", default="",
```
galpeter: The default should be `None`. Also you can specify the type for which the value will be converted automatically (just add the `type=int` as a keyword argument after the `default` entry.

jerry-debugger/jerry-client-ws.py
```diff
@@ -81,6 +81,10 @@ def arguments_parse():
...
+                        help="set display range")
+    parser.add_argument("--exception", action="store", default="",
```
galpeter: As I see the `--exception` can only have two value: `0` and `1`. Could you please update that only there values are allowed (there is a choice action) also the default should be None and we can specify the type to be int. If the choices are present then we would not need for the try-except below and the help can be extended to contain the meaning of the values.

DanielBallaSZTE: Thanks, added it to the code.

jerry-debugger/jerry-client-ws.py
```diff
@@ -81,6 +81,10 @@ def arguments_parse():
                         help="disable stop when newline is pressed (default: %(default)s)")
     parser.add_argument("--color", action="store_true", default=False,
                         help="enable color highlighting on source commands (default: %(default)s)")
+    parser.add_argument("--display", action="store", default=None, type=int,
+                        help="set display range")
+    parser.add_argument("--exception", action="store", default=None, type=int, choices=range(0, 2),
+                        help="set exception config, usage 1: [Enable] or 0: [Disable]:")
```
galpeter: Two minor things:
* No need for range call, just simply: `[0, 1]`
* There is an extra `:` at the end of the help message.

DanielBallaSZTE:
Fixed those issues.

PR #1876, opened 2017-06-02
out_native_pointer_p in jerry_get_object_native_pointer can be NULL

jiangzidong:
In some cases, user just want to know whether the JS object has native pointer, but doesn't care what the pointer is.

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -2100,9 +2100,12 @@ jerry_get_object_native_pointer (const jerry_value_t obj_val, /**< object to get
     return false;
   }
 
-  *out_native_pointer_p = native_pointer_p->data_p;
+  if (out_native_pointer_p != NULL)
+  {
+    *out_native_pointer_p = native_pointer_p->data_p;
+  }
 
-  if (out_native_info_p)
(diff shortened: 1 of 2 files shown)


jerry-core/api/jerry.c
```diff
@@ -2100,7 +2100,10 @@ jerry_get_object_native_pointer (const jerry_value_t obj_val, /**< object to get
     return false;
   }
 
-  *out_native_pointer_p = native_pointer_p->data_p;
+  if (out_native_pointer_p == NULL)
```
galpeter: Should this be `!= NULL` ?

jiangzidong: Sorry for the stupid mistake

jerry-core/api/jerry.c
```diff
@@ -2100,7 +2100,10 @@ jerry_get_object_native_pointer (const jerry_value_t obj_val, /**< object to get
     return false;
   }
 
-  *out_native_pointer_p = native_pointer_p->data_p;
+  if (out_native_pointer_p)
```
zherczeg: != NULL

jiangzidong:
@zherczeg fixed

PR #1877, opened 2017-06-02
Rename files under jerry-core/debugger to avoid build fails

galpeter:
Without the file renaming there are two jerry-debugger.c files.
Thus when the jerry-core archive is created there are two objects
in it with the same name. Generally this does not create any problems.
However if the archive is extracted then the second object file
will overwrite the first one which results in undefined reference
linkage error.

The jerry- prefix was removed from the file names and
fixed the include oreders where it was needed.

### DIFF
--- a/jerry-core/debugger/debugger.c
+++ b/jerry-core/debugger/debugger.c
@@ -14,12 +14,12 @@
  */
 
 #include "byte-code.h"
+#include "debugger.h"
 #include "ecma-builtin-helpers.h"
 #include "ecma-conversion.h"
 #include "ecma-eval.h"
 #include "ecma-objects.h"
 #include "jcontext.h"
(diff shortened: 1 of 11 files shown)


galpeter:
@akosthekiss I've updated the PR

jerry-core/debugger/debugger.c
```diff
@@ -19,7 +19,7 @@
 #include "ecma-eval.h"
 #include "ecma-objects.h"
 #include "jcontext.h"
-#include "jerry-debugger.h"
+#include "debugger.h"
```
akosthekiss: everywhere else, the includes are in alpha order. is this a leftover or intentional?

galpeter: Not, really. But it's fixed now (hopefully).

PR #1878, opened 2017-06-02
Fix gcc 7 fallthrough warnings

galpeter:
When compiling with gcc 7 the current fallthrough comments
are not accepted. This is a bug in gcc 7.

For now disable the fallthrough comment detection.
Disabling the check does not introduce any risk
as previously it was not enabled by default and
vera++ already check these kind of comments.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,6 +219,12 @@ if(ENABLE_STRIP AND NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
   jerry_add_link_flags(-s)
 endif()
 
+# TODO: Remove workaround for gcc 7 bug if the
+# fallthrough comment detection is fixed.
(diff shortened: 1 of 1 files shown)


galpeter:
The problem with this PR is that the vera checker expects the comment at its original place...

zherczeg:
Do we need to update vera as well? Btw why gcc changed its policy? Is it a bug in gcc?

galpeter:
@zherczeg AFAIK they simply introduced a stricter checker for the fallthrough things. And yeas we nee do update vera also. Another idea would be to disable the gcc fallthrough check.

galpeter:
I've been digging around and it seems it's really a gcc bug

galpeter:
@zherczeg, @LaszloLango  I've update the PR to disable the check in gcc for now.

LaszloLango:
Still LGTM

PR #1880, opened 2017-06-06
Update target mbedos5 README.

fbmrk:
You need to use mbed-js-gulp instead of make to generate a main.cpp.

### DIFF
--- a/targets/mbedos5/README.md
+++ b/targets/mbedos5/README.md
@@ -44,52 +44,31 @@ instructions above.
 make is used to automate the process of fetching dependencies, and making sure that
 mbed-cli is called with the correct arguments.
 
+### nodejs
+
+npm is used to install the dependencies in the local node_modules folder.
+
+### gulp
+
+gulp is used to automate tasks, like cloning repositories or generate source files.
+If you create an own project, for more info see [mbed-js-gulp](https://github.com/ARMmbed/mbed-js-gulp).
+
 ### (optional) jshint
 
 jshint is used to statically check your JavaScript code, as part of the build process.
 This ensures that pins you are using in your code are available on your chosen target
 platform.
 
-### Python modules
-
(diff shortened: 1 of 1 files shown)


targets/mbedos5/README.md
```diff
@@ -44,52 +44,32 @@ instructions above.
 make is used to automate the process of fetching dependencies, and making sure that
 mbed-cli is called with the correct arguments.
 
-### (optional) jshint
+### nodejs
 
-jshint is used to statically check your JavaScript code, as part of the build process.
-This ensures that pins you are using in your code are available on your chosen target
-platform.
+a recent version of [node.js](http://nodejs.org/).
```
akosthekiss: - Start sentence with capital letter. (Other sections start with lower case letters because they start with program names that cannot be capitalized.)
- What counts as recent? Is there a known working version or a minimum requirement?

targets/mbedos5/README.md
```diff
@@ -44,52 +44,32 @@ instructions above.
...
-following command invoked from your jerryscript/targets/mbedos5/tools directory:
+[gulp](http://gulpjs.com/).
 
-```bash
-pip install -r requirements.txt
-```
-
-If you have no pip installed but your are on an apt-managed system (e.g.,
-Ubuntu), issue the following command first:
+([mbed-js-gulp](https://github.com/ARMmbed/mbed-js-gulp))
```
akosthekiss: Please add some more text here. Right now, it's just two links in two lines without any explanation. What is gulp, why is there an mbed-js variant of gulp, and why is the latter one in parentheses? (Is that not needed, actually?)

fbmrk:
@akosthekiss updated.

LaszloLango:
@janjongboom @thegecko please check.

thegecko:
LGTM

Issue #1881, opened 2017-06-06
Assertion 'ecma_get_value_type_field (value) == ECMA_TYPE_OBJECT' failed in ecma_get_object_from_value


renatahodovan:
###### Jerry version:

```
Checked revision: 905cd70
Build command: ./tools/build.py --clean --debug --compile-flag=-fsanitize=address --profile=es2015-subset --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --compile-flag=-g --jerry-libc=off --static-link=off --strip=off --system-allocator=on --linker-flag=-fuse-ld=gold  --error-messages=on
```
###### OS:

```
Ubuntu 16.04.2 LTS
```
###### Test case:

```javascript
new Promise(isFinite.toString).catch(Promise.prototype.then)
```

###### Backtrace:

```
ICE: Assertion 'ecma_get_value_type_field (value) == ECMA_TYPE_OBJECT' failed at jerryscript/jerry-core/ecma/base/ecma-helpers-value.c(ecma_get_object_from_value):586.
Error: ERR_FAILED_INTERNAL_ASSERTION
bt
#0  0xf7fd8be9 in __kernel_vsyscall ()
#1  0xf7898ea9 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf789a407 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0x08054cf9 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x0809c789 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x0809c7bc in jerry_assert_fail (assertion=0x815c9a0 "ecma_get_value_type_field (value) == ECMA_TYPE_OBJECT", file=0x815c5c0 "jerryscript/jerry-core/ecma/base/ecma-helpers-value.c", function=0x811d2a0 <__func__.3327.lto_priv.733> "ecma_get_object_from_value", line=586) at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x081108ee in ecma_get_object_from_value (value=72) at jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:586
[...]

PR #1882, opened 2017-06-06
Check if it is object in Promise.prototype.then

jiangzidong:
Fix issue 1881


zherczeg:
Please add #1881 (hashmark) to the commit message.

PR #1883, opened 2017-06-07
add transform functions for integer in jerryx/arg

jiangzidong:

### DIFF
--- a/docs/09.EXT-REFERENCE-ARG.md
+++ b/docs/09.EXT-REFERENCE-ARG.md
@@ -122,6 +122,42 @@ Enum that indicates whether an argument is optional or required.
 - [jerryx_arg_function](#jerryx_arg_function)
 - [jerryx_arg_native_pointer](#jerryx_arg_native_pointer)
 
+## jerryx_arg_round_t
+
+Enum that indicates the rounding policy which will be chosen to transform an integer.
+
+ - JERRYX_ARG_ROUND - use round() method.
+ - JERRYX_ARG_FLOOR - use floor() method.
+ - JERRYX_ARG_CEIL - use ceil() method.
+
+**See also**
+
+- [jerryx_arg_uint8](#jerryx_arg_uint8)
+- [jerryx_arg_uint16](#jerryx_arg_uint16)
+- [jerryx_arg_uint32](#jerryx_arg_uint32)
+- [jerryx_arg_int8](#jerryx_arg_int8)
+- [jerryx_arg_int16](#jerryx_arg_int16)
+- [jerryx_arg_int32](#jerryx_arg_int32)
+
+
+## jerryx_arg_clamp_t
+
+ Indicates the clamping policy which will be chosen to transform an integer.
+ If the policy is NO_CLAMP, and the number is out of range,
+ then the transformer will throw a range error.
+
+ - JERRYX_ARG_CLAMP - clamp the number when it is out of range
+ - JERRYX_ARG_NO_CLAMP - throw a range error
+
+**See also**
+
+- [jerryx_arg_uint8](#jerryx_arg_uint8)
+- [jerryx_arg_uint16](#jerryx_arg_uint16)
+- [jerryx_arg_uint32](#jerryx_arg_uint32)
+- [jerryx_arg_int8](#jerryx_arg_int8)
+- [jerryx_arg_int16](#jerryx_arg_int16)
+- [jerryx_arg_int32](#jerryx_arg_int32)
+
 # Main functions
 
 ## jerryx_arg_transform_this_and_args
@@ -269,6 +305,49 @@ jerryx_arg_transform_object_properties (const jerry_value_t obj_val,
 
 
 # Helpers for commonly used validations
+
+## jerryx_arg_uint8
+
+## jerryx_arg_uint16
+
+## jerryx_arg_uint32
+
+## jerryx_arg_int8
+
+## jerryx_arg_int16
+
+## jerryx_arg_int32
+
+**Summary**
+
+All above jerryx_arg_[u]intX functions are used to create a validation/transformation step
+(`jerryx_arg_t`) that expects to consume one `number` JS argument
+and stores it into a C integer (uint8, int8, uint16, ...)
+
+**Prototype**
+
+Take jerryx_arg_int32 as an example
+
+```c
+static inline jerryx_arg_t
+jerryx_arg_int32 (int32_t *dest,
+                  jerryx_arg_round_t round_flag,
+                  jerryx_arg_clamp_t clamp_flag,
+                  jerryx_arg_coerce_t coerce_flag,
+                  jerryx_arg_optional_t opt_flag);
+```
+
+ - return value - the created `jerryx_arg_t` instance.
+ - `dest` - pointer to the `int32_t` where the result should be stored.
+ - `round_flag` - the rounding policy.
+ - `clamp_flag` - the clamping policy.
+ - `coerce_flag` - whether type coercion is allowed.
+ - `opt_flag` - whether the argument is optional.
+
+**See also**
+
+- [jerryx_arg_transform_this_and_args](#jerryx_arg_transform_this_and_args)
+
 
 ## jerryx_arg_number
 
--- a/jerry-ext/arg/arg-transform-functions.c
+++ b/jerry-ext/arg/arg-transform-functions.c
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+#include <math.h>
+
 #include "jerryscript-ext/arg.h"
 #include "jerryscript.h"
 
@@ -39,14 +41,15 @@ jerryx_arg_transform_optional (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< ava
 } /* jerryx_arg_transform_optional */
 
 /**
- * Tranform a JS argument to a double. Type coercion is not allowed.
+ * The common part in transforming a JS argument to a number (double or certain int) type.
+ * Type coercion is not allowed.
  *
  * @return jerry undefined: the transformer passes,
  *         jerry error: the transformer fails.
  */
-jerry_value_t
-jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
-                                    const jerryx_arg_t *c_arg_p) /**< the native arg */
+static jerry_value_t
+jerryx_arg_transform_number_strict_common (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                                           double *number_p) /**< [out] the number in JS arg */
 {
   jerry_value_t js_arg = jerryx_arg_js_iterator_pop (js_arg_iter_p);
 
@@ -56,21 +59,21 @@ jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p, /**
                                (jerry_char_t *) "It is not a number.");
   }
 
-  double *dest = c_arg_p->dest;
-  *dest = jerry_get_number_value (js_arg);
+  *number_p = jerry_get_number_value (js_arg);
 
   return jerry_create_undefined ();
-} /* jerryx_arg_transform_number_strict */
+} /* jerryx_arg_transform_number_strict_common */
 
 /**
- * Transform a JS argument to a double. Type coercion is allowed.
+ * The common part in transforming a JS argument to a number (double or certain int) type.
+ * Type coercion is allowed.
  *
  * @return jerry undefined: the transformer passes,
  *         jerry error: the transformer fails.
  */
-jerry_value_t
-jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
-                             const jerryx_arg_t *c_arg_p) /**< the native arg */
+static jerry_value_t
+jerryx_arg_transform_number_common (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                                    double *number_p) /**< [out] the number in JS arg */
 {
   jerry_value_t js_arg = jerryx_arg_js_iterator_pop (js_arg_iter_p);
 
@@ -84,13 +87,118 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
                                (jerry_char_t *) "It can not be converted to a number.");
   }
 
-  double *dest = c_arg_p->dest;
-  *dest = jerry_get_number_value (to_number);
+  *number_p = jerry_get_number_value (to_number);
   jerry_release_value (to_number);
 
   return jerry_create_undefined ();
+} /* jerryx_arg_transform_number_common */
+
+/**
+ * Transform a JS argument to a double. Type coercion is not allowed.
+ *
+ * @return jerry undefined: the transformer passes,
+ *         jerry error: the transformer fails.
+ */
+jerry_value_t
+jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                                    const jerryx_arg_t *c_arg_p) /**< the native arg */
+{
+  return jerryx_arg_transform_number_strict_common (js_arg_iter_p, c_arg_p->dest);
+} /* jerryx_arg_transform_number_strict */
+
+/**
+ * Tranform a JS argument to a double. Type coercion is allowed.
+ *
+ * @return jerry undefined: the transformer passes,
+ *         jerry error: the transformer fails.
+ */
+jerry_value_t
+jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                             const jerryx_arg_t *c_arg_p) /**< the native arg */
+{
+  return jerryx_arg_transform_number_common (js_arg_iter_p, c_arg_p->dest);
 } /* jerryx_arg_transform_number */
 
+/**
+ * Helper function to process a double number before converting it
+ * to an integer.
+ *
+ * @return jerry undefined: the transformer passes,
+ *         jerry error: the transformer fails.
+ */
+static jerry_value_t
+jerryx_arg_helper_process_double (double *d, /**< [in, out] the number to be processed */
+                                  double min, /**< the min value for clamping */
+                                  double max, /**< the max value for clamping */
+                                  jerryx_arg_int_option_t option) /**< the converting policies */
+{
+  if (option.clamp == JERRYX_ARG_NO_CLAMP)
+  {
+    if (*d > max || *d < min)
+    {
+      return jerry_create_error (JERRY_ERROR_TYPE,
+                                 (jerry_char_t *) "The number is out of range.");
+    }
+  }
+  else
+  {
+    *d = *d < min ? min : *d;
+    *d = *d > max ? max : *d;
+  }
+
+  if (option.round == JERRYX_ARG_ROUND)
+  {
+    *d = (*d >= 0.0) ? floor (*d + 0.5) : ceil (*d - 0.5);
+  }
+  else if (option.round == JERRYX_ARG_FLOOR)
+  {
+    *d = floor (*d);
+  }
+  else
+  {
+    *d = ceil (*d);
+  }
+
+  return jerry_create_undefined ();
+} /* jerryx_arg_helper_process_double */
+
+/**
+ * Use the macro to define thr transform functions for int type.
+ */
+#define JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE(type, suffix, min, max) \
+  jerry_value_t jerryx_arg_transform_ ## type ## suffix (jerryx_arg_js_iterator_t *js_arg_iter_p, \
+                                                         const jerryx_arg_t *c_arg_p) \
+  { \
+    double tmp; \
+    jerry_value_t rv = jerryx_arg_transform_number ## suffix ## _common (js_arg_iter_p, &tmp); \
+    if (jerry_value_has_error_flag (rv)) \
+    { \
+      return rv; \
+    } \
+    jerry_release_value (rv); \
+    jerryx_arg_int_option_t *options_p = (jerryx_arg_int_option_t *) &c_arg_p->extra_info; \
+    rv = jerryx_arg_helper_process_double (&tmp, min, max, *options_p); \
+    if (jerry_value_has_error_flag (rv)) \
+    { \
+      return rv; \
+    } \
+    *(type ## _t *) c_arg_p->dest = (type ## _t) tmp; \
+    return rv; \
+  }
+
+#define JERRYX_ARG_TRANSFORM_FUNC_FOR_INT(type, min, max) \
+  JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE (type, _strict, min, max) \
+  JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE (type, , min, max)
+
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint8, 0, UINT8_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int8, INT8_MIN, INT8_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint16, 0, UINT16_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int16, INT16_MIN, INT16_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint32, 0, UINT32_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int32, INT32_MIN, INT32_MAX)
+
+#undef JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE
+#undef JERRYX_ARG_TRANSFORM_FUNC_FOR_INT
 /**
  * Tranform a JS argument to a boolean. Type coercion is not allowed.
  *
@@ -294,12 +402,12 @@ jerryx_arg_transform_object_props (jerryx_arg_js_iterator_t *js_arg_iter_p, /**<
  * Define transformer for optional argument.
  */
 #define JERRYX_ARG_TRANSFORM_OPTIONAL(type) \
-jerry_value_t \
-jerryx_arg_transform_ ## type ## _optional (jerryx_arg_js_iterator_t *js_arg_iter_p, \
-                                            const jerryx_arg_t *c_arg_p) \
-{ \
-  return jerryx_arg_transform_optional (js_arg_iter_p, c_arg_p, jerryx_arg_transform_ ## type); \
-}
+  jerry_value_t \
+  jerryx_arg_transform_ ## type ## _optional (jerryx_arg_js_iterator_t *js_arg_iter_p, \
+                                              const jerryx_arg_t *c_arg_p) \
+  { \
+    return jerryx_arg_transform_optional (js_arg_iter_p, c_arg_p, jerryx_arg_transform_ ## type); \
+  }
 
 JERRYX_ARG_TRANSFORM_OPTIONAL (number)
 JERRYX_ARG_TRANSFORM_OPTIONAL (number_strict)
@@ -310,6 +418,21 @@ JERRYX_ARG_TRANSFORM_OPTIONAL (string_strict)
 JERRYX_ARG_TRANSFORM_OPTIONAL (function)
 JERRYX_ARG_TRANSFORM_OPTIONAL (native_pointer)
 JERRYX_ARG_TRANSFORM_OPTIONAL (object_props)
+
+JERRYX_ARG_TRANSFORM_OPTIONAL (uint8)
+JERRYX_ARG_TRANSFORM_OPTIONAL (uint16)
+JERRYX_ARG_TRANSFORM_OPTIONAL (uint32)
+JERRYX_ARG_TRANSFORM_OPTIONAL (int8)
+JERRYX_ARG_TRANSFORM_OPTIONAL (int16)
+JERRYX_ARG_TRANSFORM_OPTIONAL (int32)
+JERRYX_ARG_TRANSFORM_OPTIONAL (int8_strict)
+JERRYX_ARG_TRANSFORM_OPTIONAL (int16_strict)
+JERRYX_ARG_TRANSFORM_OPTIONAL (int32_strict)
+JERRYX_ARG_TRANSFORM_OPTIONAL (uint8_strict)
+JERRYX_ARG_TRANSFORM_OPTIONAL (uint16_strict)
+JERRYX_ARG_TRANSFORM_OPTIONAL (uint32_strict)
+
+#undef JERRYX_ARG_TRANSFORM_OPTIONAL
 
 /**
  * Ignore the JS argument.
--- a/jerry-ext/include/jerryscript-ext/arg.h
+++ b/jerry-ext/include/jerryscript-ext/arg.h
@@ -106,13 +106,52 @@ typedef enum
   JERRYX_ARG_REQUIRED
 } jerryx_arg_optional_t;
 
+/**
+ * Indicates the rounding policy which will be chosen to transform an integer.
+ */
+typedef enum
+{
+  JERRYX_ARG_ROUND, /**< round */
+  JERRYX_ARG_FLOOR, /**< floor */
+  JERRYX_ARG_CEIL /**< ceil */
+} jerryx_arg_round_t;
+
+/**
+ * Indicates the clamping policy which will be chosen to transform an integer.
+ * If the policy is NO_CLAMP, and the number is out of range,
+ * then the transformer will throw a range error.
+ */
+typedef enum
+{
+  JERRYX_ARG_CLAMP,/**< clamp the number when it is out of range */
+  JERRYX_ARG_NO_CLAMP /**< throw a range error */
+} jerryx_arg_clamp_t;
+
 /* Inline functions for initializing jerryx_arg_t */
+
+#define JERRYX_ARG_INTEGER(type) \
+  static inline jerryx_arg_t \
+  jerryx_arg_ ## type (type ## _t *dest, \
+                       jerryx_arg_round_t round_flag, \
+                       jerryx_arg_clamp_t clamp_flag, \
+                       jerryx_arg_coerce_t coerce_flag, \
+                       jerryx_arg_optional_t opt_flag);
+
+JERRYX_ARG_INTEGER (uint8)
+JERRYX_ARG_INTEGER (int8)
+JERRYX_ARG_INTEGER (uint16)
+JERRYX_ARG_INTEGER (int16)
+JERRYX_ARG_INTEGER (uint32)
+JERRYX_ARG_INTEGER (int32)
+
+#undef JERRYX_ARG_INTEGER
+
 static inline jerryx_arg_t
-jerryx_arg_number (double *dest, jerryx_arg_coerce_t conv_flag, jerryx_arg_optional_t opt_flag);
+jerryx_arg_number (double *dest, jerryx_arg_coerce_t coerce_flag, jerryx_arg_optional_t opt_flag);
 static inline jerryx_arg_t
-jerryx_arg_boolean (bool *dest, jerryx_arg_coerce_t conv_flag, jerryx_arg_optional_t opt_flag);
+jerryx_arg_boolean (bool *dest, jerryx_arg_coerce_t coerce_flag, jerryx_arg_optional_t opt_flag);
 static inline jerryx_arg_t
-jerryx_arg_string (char *dest, uint32_t size, jerryx_arg_coerce_t conv_flag, jerryx_arg_optional_t opt_flag);
+jerryx_arg_string (char *dest, uint32_t size, jerryx_arg_coerce_t coerce_flag, jerryx_arg_optional_t opt_flag);
 static inline jerryx_arg_t
 jerryx_arg_function (jerry_value_t *dest, jerryx_arg_optional_t opt_flag);
 static inline jerryx_arg_t
--- a/jerry-ext/include/jerryscript-ext/arg.impl.h
+++ b/jerry-ext/include/jerryscript-ext/arg.impl.h
@@ -17,44 +17,98 @@
 #define JERRYX_ARG_IMPL_H
 
 /* transform functions for each type. */
-jerry_value_t jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                  const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_number_strict_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                           const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_number_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                    const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                           const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_boolean_strict (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                   const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_boolean_strict_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
+
+#define JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL(type) \
+  jerry_value_t jerryx_arg_transform_ ## type (jerryx_arg_js_iterator_t *js_arg_iter_p, \
+                                               const jerryx_arg_t *c_arg_p); \
+  jerry_value_t jerryx_arg_transform_ ## type ## _optional (jerryx_arg_js_iterator_t *js_arg_iter_p, \
                                                             const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_boolean_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                     const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_boolean (jerryx_arg_js_iterator_t *js_arg_iter_p,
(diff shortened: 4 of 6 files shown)


jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +85,91 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+  } \
+  return rv; \
+}
+
+GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint8, 0, 255)
+GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int8, -128, 127)
+GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint16, 0, 65535)
+GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int16, -32767, 32767)
+GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint32, 0, 4294967295)
+GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int32, -2147483648 , 2147483647)
```
martijnthe: Use the macros from stdint.h, `(U)INT..._MAX/MIN` ?

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +85,91 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+#define GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT(type, min, max) \
+jerry_value_t jerryx_arg_transform_ ## type ## _strict (jerryx_arg_js_iterator_t *js_arg_iter_p, \
+                                                        const jerryx_arg_t *c_arg_p) \
+{ \
+  double tmp; \
+  jerry_value_t rv = jerryx_arg_transform_number_strict_common (js_arg_iter_p, &tmp);\
+  if (!jerry_value_has_error_flag (rv)) \
+  { \
+    tmp = jerryx_arg_clamp_double (tmp, min, max); \
+    *(type ## _t *) c_arg_p->dest = (type ## _t) tmp; \
```
martijnthe: Should we round?

jiangzidong: What do you think? I think they both make sense if we tell the rules to user.

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -39,14 +39,15 @@ jerryx_arg_transform_optional (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< ava
...
  *
  * @return jerry undefined: the transformer passes,
  *         jerry error: the transformer fails.
  */
-jerry_value_t
-jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
-                                    const jerryx_arg_t *c_arg_p) /**< the native arg */
+static jerry_value_t
+jerryx_arg_transform_number_strict_common (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                                           double *number_p) /**< OUT: the number in JS arg */
```
LaszloLango: `/**< [out] the number in JS arg */`

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -56,21 +57,21 @@ jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p, /**
...
  *
  * @return jerry undefined: the transformer passes,
  *         jerry error: the transformer fails.
  */
-jerry_value_t
-jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
-                             const jerryx_arg_t *c_arg_p) /**< the native arg */
+static jerry_value_t
+jerryx_arg_transform_number_common (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                                    double *number_p) /**< OUT: the number in JS arg */
```
LaszloLango: /**< [out] the number in JS arg */

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +85,91 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
-  *dest = jerry_get_number_value (to_number);
+  *number_p = jerry_get_number_value (to_number);
   jerry_release_value (to_number);
 
   return jerry_create_undefined ();
+} /* jerryx_arg_transform_number_common */
+
+/**
+ * clamp a double by given min and max value
+ * @return clamped double value
```
LaszloLango: missing newline before `@return`

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +85,91 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
-  double *dest = c_arg_p->dest;
-  *dest = jerry_get_number_value (to_number);
+  *number_p = jerry_get_number_value (to_number);
   jerry_release_value (to_number);
 
   return jerry_create_undefined ();
+} /* jerryx_arg_transform_number_common */
+
+/**
+ * clamp a double by given min and max value
```
LaszloLango: `Clamp a double by given min and max value.`

jerry-ext/arg/arg-transform-functions.c
martijnthe: hmm, yeah depends on the use case..
Maybe add a enum new argument to the `jerryx_arg_(u)int...` helpers, with `JERRYX_ARG_ROUND` / `JERRYX_ARG_FLOOR` (could provide `..._CEIL` later if someone needs it...)?

martijnthe: I guess some in use cases one may also not want the implicit clamping, but instead throw an error if the input value is out of the possible range.

So, to summarize we've got these "dimensions" when converting numbers:
1. target storage type (width & signed-ness)
2. rounding behavior (floor, round or ceil)
3. out-of-range values: clamping vs throwing Range error

... and then the common dimensions that aren't specific to numbers:
4. optionality
5. strictness

I think it's important to be able to cater for 2 & 3.
But adding 2 & 3 as arguments `jerryx_arg_(u)int...` will make the signature quite long.
Maybe it makes sense to pass a "config" struct and choose default values such that passing `{}` as config is a reasonable thing to do?

```

typedef struct {
  jerryx_arg_coerce_t coercion;
  jerryx_arg_optional_t optionality;
  jerryx_arg_rounding_t rounding;
  jerryx_arg_clipping_t clipping;
} jerryx_arg_number_options_t;

//------

uint8_t d = 0;
jerryx_arg_t mapping[] = {
  // {} == "defaults": optional arg + coerce types + round + clip.
  jerryx_arg_uint8(&d, {})
};

//-------
uint8_t d = 0;
jerryx_arg_t mapping[] = {
  // {} == "defaults", all zero
  jerryx_arg_uint8(&d, (jerryx_arg_number_options_t) {
   .clipping = JERRYX_ARG_NO_CLIPPING,
   .coercion = JERRYX_ARG_NO_COERCE
  });
};

```

jerry-ext/arg/arg-transform-functions.c
jiangzidong: Agree, My concern was that the target may not have <stdint.h>, but now I realized that if so, there is also no such `uint8_t` things, and people should write their own stdint header.

jerry-ext/arg/arg-transform-functions.c
jiangzidong: We have 3 kinds of rounding policies and 2 kinds of out-of-range policies. What about define 3*2 = 6 enums. like `JERRYX_ARG_INT_ROUND_ERROR` (means will round and will throw error if out-of-range ), and passing the enum to the c_arg_p->extra_info.

Then in the validator funtion (e.g. `jerryx_arg_transform_uint8_strict`), we can choose different policy depends on the  `c_arg_p->extra_info`. 
In this way, we don't need to provide 24 functions for one int type, just 4 is enough.

martijnthe: I don't like the idea of "flattening" these 2 independent dimensions (rounding and out-of-range policies) onto 1 enum type.
I like your idea of using `extra_info` though!

What if "compress" the struct so that it fits into `extra_info`, by using bitfields?
```
typedef struct {
  jerryx_arg_coerce_t coercion:1;
  jerryx_arg_optional_t optionality:1;
  jerryx_arg_rounding_t rounding:2;
  jerryx_arg_clipping_t clipping:1;
} jerryx_arg_number_options_t;

__Static_assert(sizeof(jerryx_arg_number_options_t) <= sizeof(((jerryx_arg_t *)0)->extra_info), "jerryx_arg_number_options_t must fit into extra_info");
```

Additional benefit: I think this way we can have just *one* internal transform function for all the int types.

jiangzidong: > I think this way we can have just one internal transform function for all the int types.

Do you mean put "target storage type" , "optionality" and "strictness" into the jerryx_arg_number_options_t?
The reason why using multiple transform functions for different type is that some unused functions will not linked into the final exe file, so that the binary size can be saved.

martijnthe: Ok, yeah that makes sense.

Still, from a usage point of view, I think it's nice to have a struct where `{}` results in using reasonable "default" options.

jiangzidong: I cannot just pass {} like ` jerryx_arg_uint8(&d, {})`
In C, the braced list of initializers cannot be empty. 

It should be `jerryx_arg_uint8(&d, (jerryx_arg_int_option_t) {0})`

jiangzidong:
updated @LaszloLango @martijnthe

jerry-ext/arg/arg-transform-functions.c
martijnthe: Sorry, I forgot the cast. `(jerryx_arg_int_option_t) {}` should work too I think.

jiangzidong: No, C doesn't support the empty initializer.  C++ does.

jiangzidong:
CI reports
```
error: type of bit-field ‘round’ is a GCC extension [-Werror=pedantic]
   jerryx_arg_round_t round:2; /**< rounding policy */
   ^
``` 


But in my local machine, it can pass all the tests (gcc version 5.4.0 20160609 (Ubuntu 5.4.0-6ubuntu1~16.04.4))

jerry-ext/include/jerryscript-ext/arg.h
```diff
@@ -90,7 +90,61 @@ typedef enum
...
+
+/**
+ * The structure indicates the options used to transform integer argument.
+ * It will be passed into jerryx_arg_t's  extra_info filed.
+ */
+typedef struct
+{
+  jerryx_arg_round_t round; /**< rounding policy */
+  jerryx_arg_clamp_t clamp; /**< clamping policy */
+} jerryx_arg_int_option_t;
```
martijnthe: Maybe it makes more sense to add the `packed` to the struct type.
AFAIK `packed` isn't applied to "parent" types.

jiangzidong: done

jiangzidong:
@LaszloLango Furthur comments?

docs/09.EXT-REFERENCE-ARG.md
```diff
@@ -98,6 +98,68 @@ Enum that indicates whether an argument is optional or required.
 - [jerryx_arg_function](#jerryx_arg_function)
 - [jerryx_arg_native_pointer](#jerryx_arg_native_pointer)
 
+## jerryx_arg_round_t
+
+Enum that indicates the rounding policy which will be choose to transform a integer.
```
LaszloLango: `an integer`

docs/09.EXT-REFERENCE-ARG.md
```diff
@@ -98,6 +98,68 @@ Enum that indicates whether an argument is optional or required.
...
+- [jerryx_arg_int16](#jerryx_arg_int16)
+- [jerryx_arg_int32](#jerryx_arg_int32)
+
+
+## jerryx_arg_int_option_t
+
+**Summary**
+
+The structure indicates the options used to transform integer argument.
+It will be passed into jerryx_arg_t's  extra_info filed.
```
LaszloLango: `field`

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -39,14 +42,15 @@ jerryx_arg_transform_optional (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< ava
...
  *
  * @return jerry undefined: the transformer passes,
  *         jerry error: the transformer fails.
  */
-jerry_value_t
-jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
-                                    const jerryx_arg_t *c_arg_p) /**< the native arg */
+static jerry_value_t
+jerryx_arg_transform_number_strict_common (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                                           double *number_p) /**< [out]: the number in JS arg */
```
LaszloLango: colon is not needed after [out]

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -56,21 +60,21 @@ jerryx_arg_transform_number_strict (jerryx_arg_js_iterator_t *js_arg_iter_p, /**
...
  *
  * @return jerry undefined: the transformer passes,
  *         jerry error: the transformer fails.
  */
-jerry_value_t
-jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
-                             const jerryx_arg_t *c_arg_p) /**< the native arg */
+static jerry_value_t
+jerryx_arg_transform_number_common (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                                    double *number_p) /**< [out]: the number in JS arg */
```
LaszloLango: ditto

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +88,152 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+jerry_value_t
+jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< available JS args */
+                             const jerryx_arg_t *c_arg_p) /**< the native arg */
+{
+  return jerryx_arg_transform_number_common (js_arg_iter_p,
+                                             c_arg_p->dest);
 } /* jerryx_arg_transform_number */
 
 /**
+ * clamp a double by given min and max value.
```
LaszloLango: `Clamp`

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +88,152 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+                         double max) /**< max value */
+{
+  const double t = d < min ? min : d;
+  return t > max ? max : t;
+} /* jerryx_arg_helper_clamp */
+
+/**
+ * round a double
+ *
+ * @return
```
LaszloLango: Please fix the description

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +88,152 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+ */
+static double
+jerryx_arg_helper_round (double x) /**< the number need to round */
+{
+  return (x >= 0.0) ? floor (x + 0.5) : ceil (x - 0.5);
+} /* jerryx_arg_helper_round */
+
+static jerry_value_t
+jerryx_arg_helper_process_double (double *d,
```
LaszloLango: missing comments

jiangzidong:
@LaszloLango updated

docs/09.EXT-REFERENCE-ARG.md
```diff
@@ -98,6 +98,68 @@ Enum that indicates whether an argument is optional or required.
...
+
+
+## jerryx_arg_clamp_t
+
+Enum that indicates whether it will throw error when the integer's range is
+out of the target type.
+
+ - JERRYX_ARG_CLAMP - it will try to clamp the number into the range.
+ - JERRYX_ARG_NO_CLAMP - it will throw range error.
+**See also**
```
LaszloLango: missing new line

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +88,159 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+
+/**
+ * Helper function to process a double number before converting it
+ * to an integer.
+ *
+ * @return jerry undefined: the transformer passes,
+ *         jerry error: the transformer fails.
+ */
+static jerry_value_t
+jerryx_arg_helper_process_double (double *d, /**< [inout] the number to be processed */
```
LaszloLango: `[in, out]`

jerry-ext/include/jerryscript-ext/arg.h
```diff
@@ -90,7 +90,61 @@ typedef enum
   JERRYX_ARG_REQUIRED
 } jerryx_arg_optional_t;
 
+/**
+ * Indicates the rounding policy which will be choose to transform a integer.
+ */
+typedef enum __attribute__((packed))
```
akosthekiss: The need for an attribute signals that something is non-standard here. Why is this needed? I guess it is to ensure that jerryx_arg_int_option_t fits into extra_info. But that can be solved in other ways. BTW, is it necessary to expose jerryx_arg_int_option_t to the user? Why not have the jerryx_arg_(u)intX functions accept both a jerryx_arg_round_t and a jerryx_arg_clamp_t argument and let those functions encode (pack) the enums into extra_info? E.g., by casting the enum values to integers and shifting-bitmasking them together into a single int value. As a lib user, you could still create mappings easily but without a need for an extra (public) type, e.g.,

```c
jerryx_arg_int8 (&num0, 0, 0, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED)
jerryx_arg_int8 (&num0, JERRYX_ARG_ROUND, 0, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED)
jerryx_arg_int8 (&num0, 0, JERRYX_ARG_NO_CLAMP, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED)
jerryx_arg_int8 (&num0, JERRYX_ARG_ROUND, JERRYX_ARG_NO_CLAMP, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED)
```

jerry-ext/include/jerryscript-ext/arg.impl.h
```diff
@@ -17,42 +17,83 @@
...
-jerry_value_t jerryx_arg_transform_function (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                             const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_function_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                      const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_native_pointer (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                   const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_native_pointer_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                            const jerryx_arg_t *c_arg_p);
+
+#define DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL(type) \
```
jiangzidong: Here I use the prefix DECLARE_ to indicate it is function declaration.And I use GENERATE_ to indicate it is function impl./defination

jerry-ext/include/jerryscript-ext/arg.impl.h
```diff
@@ -17,42 +17,83 @@
...
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (int8)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (uint16)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (int16)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (uint32)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (int32)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (number)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (string)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (boolean)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL (function)
+DECLARE_JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL (native_pointer)
```
jiangzidong: added the `#undef` in the end of the usage. Thanks for reminding.

jerry-ext/include/jerryscript-ext/arg.h
jiangzidong: the discussion with @martijnthe  is in this thread https://github.com/jerryscript-project/jerryscript/pull/1883#discussion_r120578346

The main purpose is that
* we don't want the signature of the jerryx_arg_[u]intX is too long
* jerryx_arg_int_option_t indicates that these are int-specific options from normal number. And people can just pass a empty one to use the default options
* we avoid the extra computation (like  encode (pack) the enums into extra_info) in jerryx_arg_[u]intX since they are inlined.

jiangzidong:
@akosthekiss updated
About the prefix DECLARE_ ,
it is used to indicate the following part is function declaration instead of the definition. And I use GENERATE_ to indicate it is function impl./defination

jerry-ext/include/jerryscript-ext/arg.h
akosthekiss: Having read the discussion, I'm still not convinced about the need for a public option type.
- In the discussion, the option type was planned to contain rounding, clamping, optionality, requiredness - but eventually it just remained with rounding and clamping. Optionality and requiredness are still passed as regular args. So, the functions already take 4 arguments. How much longer is a 5-args signature, when does it count as too long?
- The empty initializer is not C but C++, as also mentioned in the discussion. How better is `{0}` than `0, 0`? (Provided that you'd like to use 0-based defaults instead of writing out the enum values explicitly.)
- I'm not sure that accessing fields of a struct of packed enums is in any ways different from doing the shifting-masking manually (at the machine code level). But if we want to avoid bit ops, we can still have a non-public struct of two uint8_t fields that contain the enum values. Then, at implementation source code level, we only have to write out casting explicitly (to enum and uint8_t), and the compiler will hide the packing.

docs/09.EXT-REFERENCE-ARG.md
```diff
@@ -98,6 +98,69 @@ Enum that indicates whether an argument is optional or required.
...
+- [jerryx_arg_int32](#jerryx_arg_int32)
+
+
+## jerryx_arg_clamp_t
+
+Enum that indicates whether it will throw error when the integer's range is
+out of the target type.
+
+ - JERRYX_ARG_CLAMP - it will try to clamp the number into the range.
+ - JERRYX_ARG_NO_CLAMP - it will throw range error.
```
akosthekiss: (doc) the descriptions are not word by word identical to the comments in the sources. it might be easier to maintain documentation if they would always be kept in sync. (btw, why is "try to clamp"? can clamping not succeed?)

jerry-ext/include/jerryscript-ext/arg.h
```diff
@@ -90,7 +90,53 @@ typedef enum
...
+{
+  JERRYX_ARG_CLAMP,/**< clamp the number when it is out of range */
+  JERRYX_ARG_NO_CLAMP /**< throw a range error */
+} jerryx_arg_clamp_t;
+
+/**
+ * The structure indicates the options used to transform integer argument.
+ * It will be passed into jerryx_arg_t's  extra_info filed.
```
akosthekiss: filed -> field (only if this struct survives)

tests/unit-ext/test-ext-arg.c
```diff
@@ -211,6 +215,113 @@ test_validator2_handler (const jerry_value_t func_obj_val __attribute__((unused)
...
 } /* test_validator2_handler */
 
+/**
+ * args_p[0-2] are uint8, args_p[3-5] are int8, args_p[6-8] are uint32, args_p[9-11] are int32.
+ */
+static jerry_value_t
+test_validator_int1_handler (const jerry_value_t func_obj_val __attribute__((unused)), /**< function object */
+                            const jerry_value_t this_val __attribute__((unused)), /**< this value */
+                            const jerry_value_t args_p[], /**< arguments list */
+                            const jerry_length_t args_cnt) /**< arguments length */
```
akosthekiss: indentation is off by one space

tests/unit-ext/test-ext-arg.c
```diff
@@ -211,6 +215,113 @@ test_validator2_handler (const jerry_value_t func_obj_val __attribute__((unused)
...
+  {
+    jerryx_arg_int8 (&num0, opt_round, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+    jerryx_arg_int8 (&num1, opt_floor, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+    jerryx_arg_int8 (&num2, opt_ceil, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+    jerryx_arg_int8 (&num3, opt_round, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+    jerryx_arg_int8 (&num4, opt_floor, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+    jerryx_arg_int8 (&num5, opt_ceil, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+    jerryx_arg_int8 (&num6, opt_default, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+    jerryx_arg_int8 (&num7, opt_no_clamp, JERRYX_ARG_COERCE, JERRYX_ARG_REQUIRED),
+
```
akosthekiss: why this empty line within the array?

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,160 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+{
+  if (option.clamp == JERRYX_ARG_NO_CLAMP)
+  {
+    if (*d > max || *d < min)
+    {
+      return jerry_create_error (JERRY_ERROR_TYPE,
+                                 (jerry_char_t *) "The number is out of range.");
+    }
+  }
+  else
```
jiangzidong: Does it make sense to perfect the above code after we impl the assert/unreachable/ndebug such things in jerryx?

docs/09.EXT-REFERENCE-ARG.md
jiangzidong: clamping will not fail. I will change that

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,160 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+  }
+
+  return jerry_create_undefined ();
+} /* jerryx_arg_helper_process_double */
+
+/**
+ * Use the macro to define two transform functions for int type.
+ * One is non-strict version and the other is the strict version.
+ */
+#define GENERATE_JERRYX_ARG_TRANSFORM_FUNC_FOR_INT(type, min, max) \
```
jiangzidong: Ok, I will remove those prefix.

jerry-ext/include/jerryscript-ext/arg.impl.h
```diff
@@ -17,41 +17,97 @@
...
+        func = jerryx_arg_transform_ ## type; \
+      } \
+    } \
+    const jerryx_arg_int_option_t int_option = { .round = round_flag, .clamp = clamp_flag }; \
+    const uintptr_t *extra_info_p = (uintptr_t *) &int_option; \
+    return (jerryx_arg_t) \
+    { \
+      .func = func, \
+      .dest = (void *) dest, \
+      .extra_info = *extra_info_p \
```
akosthekiss: couldn't this be simply `.extra_info = *(uintptr_t *) &int_option;` without an extra `extra_info_p` variable?

jerry-ext/arg/arg-transform-functions.c
akosthekiss: OK, leave it for now

jiangzidong:
fixed @akosthekiss

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,120 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+  {
+    *d = ceil (*d);
+  }
+
+  return jerry_create_undefined ();
+} /* jerryx_arg_helper_process_double */
+
+/**
+ * Use the macro to define two transform functions for int type.
+ * One is non-strict version and the other is the strict version.
```
akosthekiss: the comment is outdated

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,120 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+ */
+#define JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE(type, suffix, min, max) \
+  jerry_value_t jerryx_arg_transform_ ## type ## suffix (jerryx_arg_js_iterator_t *js_arg_iter_p, \
+                                                          const jerryx_arg_t *c_arg_p) \
```
akosthekiss: indentation is off by one space

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,120 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+    jerry_release_value (rv); \
+    jerryx_arg_int_option_t *options_p = (jerryx_arg_int_option_t *) &c_arg_p->extra_info; \
+    rv = jerryx_arg_helper_process_double (&tmp, min, max, *options_p); \
+    if (jerry_value_has_error_flag (rv)) \
+    { \
+      return rv; \
+    } \
+    *(type ## _t *) c_arg_p->dest = (type ## _t) tmp; \
+    return rv; \
+  } \
```
akosthekiss: trailing backslash not needed

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,120 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+#define JERRYX_ARG_TRANSFORM_FUNC_FOR_INT(type, min, max) \
+  JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE(type, _strict, min, max) \
+  JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE(type, , min, max)
+
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint8, 0, UINT8_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int8, INT8_MIN, INT8_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint16, 0, UINT16_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int16, INT16_MIN, INT16_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (uint32, 0, UINT32_MAX)
+JERRYX_ARG_TRANSFORM_FUNC_FOR_INT (int32, INT32_MIN , INT32_MAX)
```
akosthekiss: extra spacing (`INT32_MIN , INT32_MAX`)

jerry-ext/include/jerryscript-ext/arg.h
```diff
@@ -90,13 +90,52 @@ typedef enum
...
+
 /* Inline functions for initializing jerryx_arg_t */
+
+#define JERRYX_ARG_INTEGER(type) \
+  static inline jerryx_arg_t \
+  jerryx_arg_ ## type (type ## _t *dest, \
+                      jerryx_arg_round_t round_flag, \
+                      jerryx_arg_clamp_t clamp_flag, \
+                      jerryx_arg_coerce_t coerce_flag, \
+                      jerryx_arg_optional_t opt_flag);
```
akosthekiss: indentation off by one space

jerry-ext/include/jerryscript-ext/arg.impl.h
```diff
@@ -17,41 +17,96 @@
...
-jerry_value_t jerryx_arg_transform_function (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                             const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_function_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                      const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_native_pointer (jerryx_arg_js_iterator_t *js_arg_iter_p,
-                                                   const jerryx_arg_t *c_arg_p);
-jerry_value_t jerryx_arg_transform_native_pointer_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
+
+#define JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL(type) \
+  jerry_value_t jerryx_arg_transform_ ## type  (jerryx_arg_js_iterator_t *js_arg_iter_p, \
```
akosthekiss: extra spacing (`type  (`)

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,120 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+    { \
+      return rv; \
+    } \
+    *(type ## _t *) c_arg_p->dest = (type ## _t) tmp; \
+    return rv; \
+  } \
+
+#define JERRYX_ARG_TRANSFORM_FUNC_FOR_INT(type, min, max) \
+  JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE(type, _strict, min, max) \
+  JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE(type, , min, max)
```
akosthekiss: missing space at `TEMPLATE(` (between end of macro name and opening parenthesis)

jerry-ext/include/jerryscript-ext/arg.impl.h
```diff
@@ -17,41 +17,96 @@
...
+                                                const jerryx_arg_t *c_arg_p); \
+  jerry_value_t jerryx_arg_transform_ ## type ## _optional (jerryx_arg_js_iterator_t *js_arg_iter_p, \
                                                             const jerryx_arg_t *c_arg_p);
+
+#define JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT(type) \
+  JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL(type) \
+  JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL(type ## _strict)
```
akosthekiss: missing space at `OPTIONAL(`

jiangzidong:
fixed and rebased to the latest master @akosthekiss

jerry-ext/arg/arg-transform-functions.c
```diff
@@ -84,14 +87,119 @@ jerryx_arg_transform_number (jerryx_arg_js_iterator_t *js_arg_iter_p, /**< avail
...
+
+  return jerry_create_undefined ();
+} /* jerryx_arg_helper_process_double */
+
+/**
+ * Use the macro to define thr transform functions for int type.
+ */
+#define JERRYX_ARG_TRANSFORM_FUNC_FOR_INT_TEMPLATE(type, suffix, min, max) \
+  jerry_value_t jerryx_arg_transform_ ## type ## suffix (jerryx_arg_js_iterator_t *js_arg_iter_p, \
+                                                        const jerryx_arg_t *c_arg_p) \
```
akosthekiss: indentation still off by one space

jerry-ext/include/jerryscript-ext/arg.impl.h
```diff
@@ -17,45 +17,100 @@
...
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (string)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (boolean)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL (function)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL (native_pointer)
 jerry_value_t jerryx_arg_transform_ignore (jerryx_arg_js_iterator_t *js_arg_iter_p,
                                            const jerryx_arg_t *c_arg_p);
 jerry_value_t jerryx_arg_transform_object_props (jerryx_arg_js_iterator_t *js_arg_iter_p,
                                                  const jerryx_arg_t *c_arg_p);
 jerry_value_t jerryx_arg_transform_object_props_optional (jerryx_arg_js_iterator_t *js_arg_iter_p,
                                                           const jerryx_arg_t *c_arg_p);
```
akosthekiss: why haven't these been refactored to use `JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL`? (that's what I was referring to in my previous review)

tests/unit-ext/test-ext-arg.c
```diff
@@ -346,6 +350,104 @@ test_validator_prop3_handler (const jerry_value_t func_obj_val __attribute__((un
...
+  TEST_ASSERT (num1 == -2);
+  TEST_ASSERT (num2 == -1);
+  TEST_ASSERT (num3 == 2);
+  TEST_ASSERT (num4 == 1);
+  TEST_ASSERT (num5 == 2);
+  TEST_ASSERT (num6 == 127);
+  TEST_ASSERT (num7 == 123);
+
+  jerry_release_value (is_ok);
+  validator_int_count ++;
```
akosthekiss: seems like extra spacing

tests/unit-ext/test-ext-arg.c
```diff
@@ -346,6 +350,104 @@ test_validator_prop3_handler (const jerry_value_t func_obj_val __attribute__((un
...
+  TEST_ASSERT (num5 == -127);
+  TEST_ASSERT (num6 == 0);
+  TEST_ASSERT (num7 == 4294967295);
+  TEST_ASSERT (num8 == 65536);
+  TEST_ASSERT (num9 == -2147483648);
+  TEST_ASSERT (num10 == 2147483647);
+  TEST_ASSERT (num11 == -2147483647);
+
+  jerry_release_value (is_ok);
+  validator_int_count ++;
```
akosthekiss: seems like extra spacing

jerry-ext/include/jerryscript-ext/arg.impl.h
jiangzidong: I mainly focused on the conflict after rebasing to the integer-check patch, and missed this one.. Fixed

jiangzidong:
What about the current update? @akosthekiss

docs/09.EXT-REFERENCE-ARG.md
```diff
@@ -270,6 +306,49 @@ jerryx_arg_transform_object_properties (const jerry_value_t obj_val,
...
+jerryx_arg_int32 (int32_t *dest,
+                  jerryx_arg_round_t round_flag,
+                  jerryx_arg_clamp_t clamp_flag,
+                  jerryx_arg_coerce_t coerce_flag,
+                  jerryx_arg_optional_t opt_flag);
+```
+
+ - return value - the created `jerryx_arg_t` instance.
+ - `dest` - pointer to the `int32_t` where the result should be stored.
+ - `round_flag` - the roudning policy.
```
akosthekiss: roudning -> rounding

jerry-ext/include/jerryscript-ext/arg.impl.h
```diff
@@ -17,44 +17,97 @@
...
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (uint8)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (int8)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (uint16)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (int16)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (uint32)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (int32)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (number)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (string)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (boolean)
+JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT (object_props)
```
akosthekiss: This is not right. `JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL_AND_STRICT` will declare
```
jerryx_arg_transform_object_props
jerryx_arg_transform_object_props_optional
jerryx_arg_transform_object_props_strict
jerryx_arg_transform_object_props_strict_optional
```
but the latter two does not exist. `JERRYX_ARG_TRANSFORM_FUNC_WITH_OPTIONAL` should be used instead.

jiangzidong: My bad. Stupid mistake...

PR #1885, opened 2017-06-08
Add timeout to jerry_debugger_receive

rerobika:
Fixed when debug server uses a CPU core on 100%.

### DIFF
--- a/jerry-core/CMakeLists.txt
+++ b/jerry-core/CMakeLists.txt
@@ -181,6 +181,16 @@ if(FEATURE_DEBUGGER)
     message(FATAL_ERROR "This configuration is not supported. Please build against your system libc to enable the JerryScript debugger.")
   endif()
 
+  # Sleep function availability check
+  INCLUDE (CheckIncludeFiles)
+  CHECK_INCLUDE_FILES (time.h HAVE_TIME_H)
+  CHECK_INCLUDE_FILES (unistd.h HAVE_UNISTD_H)
+  if(HAVE_TIME_H)
+    set(DEFINES_JERRY ${DEFINES_JERRY} HAVE_TIME_H)
+  elseif(HAVE_UNISTD_H)
+    set(DEFINES_JERRY ${DEFINES_JERRY} HAVE_UNISTD_H)
+  endif()
+
   set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_DEBUGGER)
   set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_DEBUGGER_PORT=${FEATURE_DEBUGGER_PORT})
 endif()
--- a/jerry-core/debugger/debugger.c
+++ b/jerry-core/debugger/debugger.c
@@ -13,6 +13,8 @@
  * limitations under the License.
  */
 
+#define _XOPEN_SOURCE 500 /* Required macro for sleep functions */
+
 #include "byte-code.h"
 #include "debugger.h"
 #include "ecma-builtin-helpers.h"
@@ -25,6 +27,12 @@
 
 #ifdef JERRY_DEBUGGER
 
+#ifdef HAVE_TIME_H
+#include <time.h>
+#elif defined (HAVE_UNISTD_H)
+#include <unistd.h>
+#endif /* HAVE_TIME_H */
+
 /**
  * Type cast the debugger send buffer into a specific type.
  */
@@ -36,6 +44,11 @@
  */
 #define JERRY_DEBUGGER_RECEIVE_BUFFER_AS(type, name_p) \
   type *name_p = ((type *) recv_buffer_p)
+
+/**
+ * Sleep time in milliseconds between each jerry_debugger_receive call
+ */
+#define JERRY_DEBUGGER_TIMEOUT 100
 
 /**
  * Free all unreferenced byte code structures which
@@ -200,6 +213,24 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
 
   return success;
 } /* jerry_debugger_send_eval */
+
+/**
+ * Suspend execution for a given time.
+ * Note: If the platform does not have nanosleep or usleep, this function does not sleep at all.
+ */
+static void
+jerry_debugger_sleep (unsigned milliseconds) /**< suspending time */
+{
+#ifdef HAVE_TIME_H
+  nanosleep (&(const struct timespec)
+  {
+    milliseconds / 1000, (milliseconds % 1000) * 1000000L /* Seconds, nanoseconds */
+  }
+  , NULL);
+#elif defined (HAVE_UNISTD_H)
+  usleep ((useconds_t) milliseconds * 1000);
+#endif /* HAVE_TIME_H */
+} /* jerry_debugger_sleep */
 
 /**
  * Check received packet size.
@@ -514,6 +545,7 @@ jerry_debugger_breakpoint_hit (uint8_t message_type) /**< message type */
 
   while (!jerry_debugger_receive ())
   {
+    jerry_debugger_sleep (JERRY_DEBUGGER_TIMEOUT);
   }
 
   JERRY_CONTEXT (debugger_flags) = (uint8_t) (JERRY_CONTEXT (debugger_flags) & ~JERRY_DEBUGGER_BREAKPOINT_MODE);


jerry-core/debugger/debugger.c
```diff
@@ -514,6 +519,10 @@ jerry_debugger_breakpoint_hit (uint8_t message_type) /**< message type */
 
   while (!jerry_debugger_receive ())
   {
+    /* timeout */
+    #ifdef __unix__
+    syscall(SYS_nanosleep, (const struct timespec[]){{0, 100}}, NULL);
```
galpeter: Do we really need to do a syscall here? Can't we just invoke the nanosleep C function?

zherczeg:
Is this a problem? Debugging is not for benchmarking.

rerobika:
I know it's not for benchmarking, but in my opinion the debug server consumes 10 times more resources than it really needs.

jerry-core/debugger/debugger.c
rerobika: I think so, because jerry-libc is turned off so nanosleep is not available and combining jerry-libc with unix-libc would not be the best choice.

zherczeg:
The idea is good but I worry about the compatibility here. What about TizenRT, nuttix and other exotic systems?

jerry-core/debugger/debugger.c
galpeter: But if the jerry libc is turned of then we are using the standard libc which should have the nanosleep function.

jerry-core/debugger/debugger.c
```diff
@@ -514,6 +519,14 @@ jerry_debugger_breakpoint_hit (uint8_t message_type) /**< message type */
 
   while (!jerry_debugger_receive ())
   {
+    /* timeout */
+    #ifdef __unix__
+    syscall (SYS_nanosleep, &(const struct timespec)
+    {
+        0, 100  /*second, millisecond*/
```
akosthekiss: The comment is misleading is many ways. First of all, the `timespec` struct has fields for seconds and nanoseconds. Or, 100 nanoseconds is not 1 millisec but 0.1 microsec.

akosthekiss:
Partly revising myself: as the debugger support cannot even be built with jerry-libc, implementing nanosleep/usleep/sleep in jerry-libc is not needed. The rest of the above still holds ("JerryScript should be kept plain C99", "a standard libc function should be used in jerry-core", "options are nanosleep, usleep, and sleep", ...)

(PS: jerry-core/debugger/debugger-ws.c can act as an example for how to separate includes. Those headers that are available everywhere can go before `#ifdef JERRY_DEBUGGER`, while those that are not available in jerry-libc should go after it. Should that kind of header separation become necessary.)

rerobika:
Thanks for your tips and advice, I'm going to refactor the code as soon as possible. I really appreciate your help @akosthekiss ! :)

rerobika:
@akosthekiss I modified it depending your previous comments. Can you review it?

jerry-core/debugger/debugger.c
```diff
@@ -522,6 +523,22 @@ jerry_debugger_breakpoint_hit (uint8_t message_type) /**< message type */
 } /* jerry_debugger_breakpoint_hit */
 
 /**
+* Suspend execution for received time
+*/
+void jerry_debugger_sleep (int milliseconds) /**< Suspending time */
+{
+  #ifdef HAVE_TIME_H
```
akosthekiss: style: preprocessor directives should start at beginning of line (column 0)

jerry-core/debugger/debugger.c
```diff
@@ -522,6 +523,22 @@ jerry_debugger_breakpoint_hit (uint8_t message_type) /**< message type */
...
+  nanosleep (&(const struct timespec)
+  {
+    milliseconds / 1000, milliseconds % 1000
+  }
+  , NULL);
+  #elif defined (HAVE_UNISTD_H)
+  usleep (milliseconds * 1000);
+  #endif
```
akosthekiss: `#endif` (and `#else`) directives are (almost) always followed by comments (here, it should be `/* HAVE_TIME */`, if I'm not mistaken. please, grep around in jerry-core for `#endif`)

jerry-core/debugger/debugger.c
```diff
@@ -522,6 +523,22 @@ jerry_debugger_breakpoint_hit (uint8_t message_type) /**< message type */
...
 
 /**
+* Suspend execution for received time
+*/
+void jerry_debugger_sleep (int milliseconds) /**< Suspending time */
+{
+  #ifdef HAVE_TIME_H
+  nanosleep (&(const struct timespec)
+  {
+    milliseconds / 1000, milliseconds % 1000
```
akosthekiss: `milliseconds % 1000` is not nanosecs

jerry-core/debugger/debugger.c
```diff
@@ -522,6 +523,22 @@ jerry_debugger_breakpoint_hit (uint8_t message_type) /**< message type */
 } /* jerry_debugger_breakpoint_hit */
 
 /**
+* Suspend execution for received time
+*/
+void jerry_debugger_sleep (int milliseconds) /**< Suspending time */
```
akosthekiss: Moreover, I don't think that this function needs to be exported from this module. It can be kept static, as it's only accessed from this file.

jerry-core/debugger/debugger.h
```diff
@@ -29,6 +39,11 @@
 #define JERRY_DEBUGGER_MESSAGE_FREQUENCY 5
 
 /**
+* Sleep time in milliseconds between each jerry_debugger_receive() call
+*/
+#define JERRY_DEBUGGER_TIMEOUT 100
```
akosthekiss: as this define is used in the source file only, it can be moved there, too (and don't forget the proper indentation of the comment block)

jerry-core/debugger/debugger.h
```diff
@@ -19,6 +19,16 @@
...
 
+#ifdef HAVE_TIME_H
+#define __USE_POSIX199309
+#include <time.h>
+#undef __USE_POSIX199309
+#elif defined (HAVE_UNISTD_H)
+#define __USE_MISC
+#include <unistd.h>
+#undef __USE_MISC
+#endif
```
akosthekiss: several comments:
- all these should go into the source file, not in this header
- the double-underscore defines are very suspicious. usually, you should not touch any of them. please, have a look at https://www.gnu.org/software/libc/manual/html_node/Feature-Test-Macros.html . most probably `_XOPEN_SOURCE` is the macro you are looking for. (also note that "You should define these macros by using ‘#define’ preprocessor directives at the top of your source code files. These directives must come before any #include of a system header file.") (and it does not neet to be undefined IMO)
- don't forget adding comments to preproc directives

rerobika:
I fixed these @akosthekiss!

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +218,23 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
...
 /**
+ * Suspend execution for received time
+ */
+static void
+jerry_debugger_sleep (int milliseconds) /**< Suspending time */
+{
+#ifdef HAVE_TIME_H /* HAVE_TIME */
+  nanosleep (&(const struct timespec)
+  {
+    milliseconds / 1000, milliseconds * 1000000 /* Seconds, nanoseconds */
```
akosthekiss: math is still not good. e.g., in case of 1500 millisecs, this will be 1 sec + 1500 * 10^6 nanosecs (where 500 * 10^6 wuld be correct)

one more note: milliseconds is int, and multiplying it by literal 1000000 will also yield an int, which may overflow on some architectures (although improbable on 32 bit systems). using a long literal (1000000L) can solve the problem for sure (and the second field of the timespec struct is long anyway).

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +218,23 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
...
+  }
+  , NULL);
+#elif defined (HAVE_UNISTD_H) /* HAVE_UNISTD */
```
akosthekiss: no need for the comment here and at the `#if`, only at the end

jerry-core/debugger/debugger.c
```diff
@@ -38,6 +49,11 @@
   type *name_p = ((type *) recv_buffer_p)
 
 /**
+ * Sleep time in milliseconds between each jerry_debugger_receive() call
```
akosthekiss: style: either don't add the parentheses after jerry_debugger_receive or use jerry coding style there as well and add a space between the function name and the parentheses

jerry-core/debugger/debugger.c
```diff
@@ -26,6 +28,15 @@
 #ifdef JERRY_DEBUGGER
 
 /**
+ * HAVE_TIME
+ */
+#ifdef HAVE_TIME_H
+#include <time.h>
+#elif defined (HAVE_UNISTD_H) /* HAVE_UNISTD */
```
akosthekiss: no need for the comment here

jerry-core/CMakeLists.txt
```diff
@@ -250,6 +250,16 @@ if (FEATURE_VM_EXEC_STOP)
...
 
+# Sleep function availability check
+INCLUDE (CheckIncludeFiles)
+CHECK_INCLUDE_FILES (time.h HAVE_TIME_H)
+CHECK_INCLUDE_FILES (unistd.h HAVE_UNISTD_H)
+if(HAVE_TIME_H)
+  set(DEFINES_JERRY ${DEFINES_JERRY} HAVE_TIME_H)
+elseif(HAVE_UNISTD_H)
+  set(DEFINES_JERRY ${DEFINES_JERRY} HAVE_UNISTD_H)
+endif()
```
akosthekiss: all this is debugger specific. please move this to line 184 under the debugger feature check

rerobika:
I've fixed these mistakes.

jerry-core/debugger/debugger.c
```diff
@@ -26,6 +28,15 @@
 #ifdef JERRY_DEBUGGER
 
 /**
+ * HAVE_TIME
+ */
```
akosthekiss: this doc comment is not needed (or not needed in its current form). A `/**` is a doxygen comment for documenting an entity in the code. but we usually don't document the inclusion of headers. (even if we would, we should add descriptive English text in the documentation.)

TL;DR: drop this comment

jerry-core/debugger/debugger.c
```diff
@@ -26,6 +28,15 @@
...
+#ifdef HAVE_TIME_H
+#include <time.h>
+#elif defined (HAVE_UNISTD_H)
+#include <unistd.h>
+#endif /* HAVE_TIME */
```
akosthekiss: this comment is needed, but it should match the predicate of the `#ifdef`, i.e., `#endif /* HAVE_TIME_H */`

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +218,23 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
 } /* jerry_debugger_send_eval */
 
 /**
+ * Suspend execution for received time
```
akosthekiss: here, "received" sounds a bit strange to me. as if we would _receive_ it over network (which we don't). alternatives could be:
- Suspend execution for some time.
- Suspend execution for a given time.
- Suspend execution for milliseconds intervals.
- Platform-independent sleep implementation.

The following note might be worth adding anyway: Note: If the platform does not have nanosleep or usleep, this function does not sleep at all.

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +218,23 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
...
+ */
+static void
+jerry_debugger_sleep (int milliseconds) /**< Suspending time */
```
akosthekiss: usually, we don't capitalize the docstring of parameters (see rest of the code base, or this file).

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +218,23 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
...
+{
+#ifdef HAVE_TIME_H /* HAVE_TIME */
```
akosthekiss: drop this comment

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +218,23 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
...
+  nanosleep (&(const struct timespec)
+  {
+    milliseconds / 1000, (milliseconds % 1000) * 1000000L /* Seconds, nanoseconds */
+  }
+  , NULL);
+#elif defined (HAVE_UNISTD_H) /* HAVE_UNISTD */
```
akosthekiss: drop this comment

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +218,23 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
...
+  usleep ((__useconds_t) milliseconds * 1000);
+#endif /* HAVE_TIME */
```
akosthekiss: match this comment to the `#ifdef` predicate

rerobika: I've fixed these :)

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +215,24 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
...
+jerry_debugger_sleep (int milliseconds) /**< suspending time */
+{
+#ifdef HAVE_TIME_H
+  nanosleep (&(const struct timespec)
+  {
+    milliseconds / 1000, (milliseconds % 1000) * 1000000L /* Seconds, nanoseconds */
+  }
+  , NULL);
+#elif defined (HAVE_UNISTD_H)
+  usleep ((__useconds_t) milliseconds * 1000);
```
akosthekiss: I'm not completely sure about this but it's suspicious: are you sure that `__useconds_t` is the type that should be used here? double underscores usually denote internal things. according to manuals, the declaration of usleep is `int usleep(useconds_t usec);`, i.e., the public type has no underscore prefix. please check

rerobika: You are right! Using __useconds_t is proved to be unnecessary, the public type is appropriate. I'll fix this.

jerry-core/debugger/debugger.c
```diff
@@ -202,6 +215,24 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
 } /* jerry_debugger_send_eval */
 
 /**
+ * Suspend execution for a given time.
+ * Note: If the platform does not have nanosleep or usleep, this function does not sleep at all.
+ */
+static void
+jerry_debugger_sleep (int milliseconds) /**< suspending time */
```
LaszloLango: `unsigned milliseconds`

rerobika: I've fixed it based on your previous comment!

LaszloLango:
LGTM

Issue #1857, opened 2017-05-26
target: mbedos5: Move print to jerryx_handler_print


janjongboom:
Right now we have our own print implementation (#1853). Needs to be replaced by the one from jerry-ext.

akosthekiss:
Closing as #1889 has landed

PR #1889, opened 2017-06-13
target: mbedos5: use jerry-ext

fbmrk:
Update target mbedos5 to use external function handlers: `assert`, `gc` and `print` from jerry-ext.
See #1857 


zherczeg:
@janjongboom please check

akosthekiss:
Beware that the PR changes the behaviour of JS `assert`! Existing implementation throws a JS error, but jerry-ext implementation kills the engine.

janjongboom:
@akosthekiss I don't think people rely on `assert` in that way. Other than that LGTM.

akosthekiss:
@janjongboom thanks for reviewing the PR. I have nothing against the change, just didn't want it to go unnoticed.

PR #1892, opened 2017-06-26
target: mbedos5: unnecessary acquire

fbmrk:
When you cleanup the engine you got an `ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)` error.
There is an unnecessary call of jerry_acquire_value which causes the problem.

Also in the InterruptIn-js.cpp file there is a wrong check of an argument.

### DIFF
--- a/targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/source/InterruptIn-js.cpp
+++ b/targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/source/InterruptIn-js.cpp
@@ -29,7 +29,7 @@ DECLARE_CLASS_FUNCTION(InterruptIn, rise) {
     CHECK_ARGUMENT_COUNT(InterruptIn, rise, (args_count == 1));
 
     // Detach the rise callback when InterruptIn::rise(null) is called
-    if (jerry_value_is_null(args[1])) {
+    if (jerry_value_is_null(args[0])) {
         uintptr_t native_handle;
         jerry_get_object_native_handle(this_obj, &native_handle);
 
@@ -44,6 +44,7 @@ DECLARE_CLASS_FUNCTION(InterruptIn, rise) {
             // Ensure that the EventLoop frees memory used by the callback.
             mbed::js::EventLoop::getInstance().dropCallback(cb_func);
         }
+        jerry_release_value(cb_func);
 
         this_interruptin->rise(0);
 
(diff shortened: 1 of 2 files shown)


akosthekiss:
Please, give a more descriptive commit message:
- the patch clearly does not deal with releasing values only (it seems that there was also a bug in arg checks), and
- the commit message does not tell why the releasing of `f` is necessary (e.g., what issue rose from it not being released).

targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/source/InterruptIn-js.cpp
```diff
@@ -66,6 +66,7 @@ DECLARE_CLASS_FUNCTION(InterruptIn, rise) {
     // Keep track of our callback internally.
     jerry_value_t property_name = jerry_create_string((const jerry_char_t*)"cb_rise");
     jerry_set_property(this_obj, property_name, f);
+    jerry_release_value(f);
```
LaszloLango: this release look unnecessary

fbmrk:
If you do not start the eventloop in mbed, just call the jerry_cleanup function at the end of the application, you got this error message:
`ICE: Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed at ../../jerry-core/ecma/base/ecma-literal-storage.c(ecma_free_string_l.
Error: ERR_FAILED_INTERNAL_ASSERTION`
@akosthekiss I am sorry about the empty message, but I thought it is a simple bug, and my PR is correct and clear solution for that. Probably i was wrong, and there is an other way to fix this error.
I'm working on the solution, and i will update the PR soon.

fbmrk:
@akosthekiss @LaszloLango 
updated

targets/mbedos5/jerryscript-mbed/jerryscript-mbed-event-loop/EventLoop.h
```diff
@@ -52,9 +52,6 @@ class EventLoop {
     Callback<void()> wrapFunction(jerry_value_t f) {
         MBED_ASSERT(jerry_value_is_function(f));
 
-        // not sure if this is necessary?
-        jerry_acquire_value(f);
```
akosthekiss: Should this not go in pair with `jerry_release_value` in `dropCallback`? If we wrote
```js
InterruptIn.raise(f());
InterruptIn.raise(null);
```
then would we not release `f` more than we acquired it?

fbmrk:
@akosthekiss 
If you delete the `jerry_release_value` from `dropCallback`, you get the same error.
My new update possibly make the code more clear, please check.

PR #1894, opened 2017-06-27
target: mbedos5: fix mbedignore

fbmrk:
Ignoring the entire jerry-port/default library causes a link error, because of the need for the defaultx-handler.c source file.


akosthekiss:
Note: editing the PR comment only does not change the commit message.

PR #1895, opened 2017-06-28
Fix a build error

rerobika:
Fixed an error when neither of libraries are found the build fails.

### DIFF
--- a/jerry-core/debugger/debugger.c
+++ b/jerry-core/debugger/debugger.c
@@ -229,6 +229,8 @@ jerry_debugger_sleep (unsigned milliseconds) /**< suspending time */
   , NULL);
 #elif defined (HAVE_UNISTD_H)
   usleep ((useconds_t) milliseconds * 1000);
+#else /* If neither of the libs found */
+  JERRY_UNUSED (milliseconds);
 #endif /* HAVE_TIME_H */
 } /* jerry_debugger_sleep */
 


jerry-core/debugger/debugger.c
```diff
@@ -230,6 +230,7 @@ jerry_debugger_sleep (unsigned milliseconds) /**< suspending time */
 #elif defined (HAVE_UNISTD_H)
   usleep ((useconds_t) milliseconds * 1000);
 #endif /* HAVE_TIME_H */
+  JERRY_UNUSED (milliseconds); /* If neither of the libs found */
```
LaszloLango:
```c
#elif HAVE_UNISTD_H
  usleep ((useconds_t) milliseconds * 1000);
#else /* !HAVE_TIME_H && !HAVE_UNISTD_H */
  JERRY_UNUSED (milliseconds);
#endif /* HAVE_TIME_H */
```

rerobika: I've fixed it based on your previous comment!

jerry-core/debugger/debugger.c
```diff
@@ -227,8 +227,10 @@ jerry_debugger_sleep (unsigned milliseconds) /**< suspending time */
     milliseconds / 1000, (milliseconds % 1000) * 1000000L /* Seconds, nanoseconds */
   }
   , NULL);
-#elif defined (HAVE_UNISTD_H)
+#elif HAVE_UNISTD_H
```
zherczeg: #if defined (HAVE_UNISTD_H) && HAVE_UNISTD_H

akosthekiss: I don't agree either with the proposed change or with the review.
- `#elif HAVE_UNISTD_H` is not correct, as the CMakeLists.txt defines HAVE_UNISTD_H without a value (if it gets defined at all), so this conditional can never be evaluated as 1 (true).
- `#if defined (HAVE_UNISTD_H) && HAVE_UNISTD_H` is unidiomatic and incorrect (because of the same reason, i.e., the `&& HAVE_UNISTD_H` part will always be false.

IMHO, the original form (`#elif defined (HAVE_UNISTD_H)`) is the only correct approach

rerobika:
IMHO finally I've fixed the mistake! :)

LaszloLango:
@zherczeg may we land this fix?

PR #1902, opened 2017-06-30
Add the API examples to the code base

LaszloLango:
After this commit the examples will be checked by 'buidoption-test',
so it helps to catch API changes as soon as possible and update the
documentations.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ if(NOT CMAKE_BUILD_TYPE)
 endif()
 
 # Optional components
+set(EXAMPLES              OFF CACHE BOOL "Build examples?")
 set(JERRY_CMDLINE         ON  CACHE BOOL "Build jerry command line tool?")
 set(JERRY_CMDLINE_MINIMAL OFF CACHE BOOL "Build jerry minimal command line tool?")
 set(JERRY_PORT_DEFAULT    ON  CACHE BOOL "Build default jerry port implementation?")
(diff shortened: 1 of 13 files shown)


akosthekiss:
I'm somewhat afraid of the maintenance burden of keeping the doc and the code in sync. Wouldn't it be better if we could build and run the code "right out of the docs"? Python has the concept of doctests, I was thinking of something similar (although there are various differences between the two languages, so 1-to-1 mapping is certainly not possible). I've drafted something here: https://github.com/akosthekiss/jerryscript/commit/924c8a32594c7e3471b4c722213e71a2ea7d6329

LaszloLango:
IMHO it isn't a high risk. The only thing you need to know is to always update the docs if you update the examples. We can add a note in the files. The biggest gain of this PR is that even if the doc is got out of sync (somehow), the codes under the example folder stays working, because the travis checks it. Furthermore it gives a better experience for newcomers if we give them an easy way to build and run the API examples.

akosthekiss:
But why not do this automatically?

LaszloLango:
@akosthekiss interesting patch. Why didn't make a PR from it before? :) My only concern is the complexity of it. If you think the parsing of [doctest] is stable and don't see any risk there, then I can close this PR. This PR is much simpler.

zherczeg:
Is the other patch makes this patch obsolote?

LaszloLango:
@zherczeg Yes, I think the other patch is better. I'm closing this.

PR #1909, opened 2017-07-04
Add support for doctests

akosthekiss:
Markdown files in the docs/ directory can now be annotated to turn
fenced C code blocks into unit tests. The recognized syntax is:

    [doctest]: # (name="test.c", test="run")

    ```c
    // unit test code
    ```

The commit also fixes the issues revealed during the initial
annotation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -45,6 +45,7 @@ set(JERRY_EXT             ON  CACHE BOOL "Build jerry-ext?")
 set(JERRY_LIBC            ON  CACHE BOOL "Build and use jerry-libc?")
 set(JERRY_LIBM            ON  CACHE BOOL "Build and use jerry-libm?")
 set(UNITTESTS             OFF CACHE BOOL "Build unit tests?")
+set(DOCTESTS              OFF CACHE BOOL "Build doc tests?")
 
 # Optional build settings
 set(ENABLE_ALL_IN_ONE  OFF CACHE BOOL "Enable all-in-one build?")
@@ -53,16 +54,16 @@ set(ENABLE_STATIC_LINK ON  CACHE BOOL "Enable static linking?")
 set(ENABLE_STRIP       ON  CACHE BOOL "Enable stripping all symbols from release binary?")
 
 # Option overrides
-if(JERRY_CMDLINE OR JERRY_CMDLINE_MINIMAL)
+if(JERRY_CMDLINE OR JERRY_CMDLINE_MINIMAL OR UNITTESTS OR DOCTESTS)
   set(JERRY_PORT_DEFAULT ON)
 
-  set(JERRY_PORT_DEFAULT_MESSAGE " (FORCED BY CMDLINE)")
+  set(JERRY_PORT_DEFAULT_MESSAGE " (FORCED BY CMDLINE OR TESTS)")
 endif()
 
-if(JERRY_CMDLINE)
+if(JERRY_CMDLINE OR DOCTESTS)
   set(JERRY_EXT ON)
 
-  set(JERRY_EXT_MESSAGE " (FORCED BY CMDLINE)")
+  set(JERRY_EXT_MESSAGE " (FORCED BY CMDLINE OR TESTS)")
 endif()
 
 if("${PLATFORM}" STREQUAL "DARWIN")
@@ -104,6 +105,7 @@ message(STATUS "JERRY_EXT                 " ${JERRY_EXT} ${JERRY_EXT_MESSAGE})
 message(STATUS "JERRY_LIBC                " ${JERRY_LIBC} ${JERRY_LIBC_MESSAGE})
 message(STATUS "JERRY_LIBM                " ${JERRY_LIBM} ${JERRY_LIBM_MESSAGE})
 message(STATUS "UNITTESTS                 " ${UNITTESTS})
+message(STATUS "DOCTESTS                  " ${DOCTESTS})
 
 # Setup directories
 # Project binary dir
@@ -180,6 +182,8 @@ if (USING_GCC OR USING_CLANG)
 endif()
 if(("${PLATFORM}" STREQUAL "DARWIN"))
   jerry_add_link_flags(-lSystem)
+  set(CMAKE_C_ARCHIVE_CREATE "<CMAKE_AR> Sqc <TARGET> <LINK_FLAGS> <OBJECTS>")
+  set(CMAKE_C_ARCHIVE_FINISH "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
 else()
   jerry_add_link_flags(-Wl,-z,noexecstack)
 endif()
@@ -273,4 +277,9 @@ if(UNITTESTS)
   if(JERRY_EXT)
     add_subdirectory(tests/unit-ext)
   endif()
+endif()
+
+# Doctests
+if(DOCTESTS)
+  add_subdirectory(tests/unit-doc)
 endif()
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -303,7 +303,13 @@ jerry_init (jerry_init_flag_t flags)
 
 **Example**
 
+[doctest]: # ()
+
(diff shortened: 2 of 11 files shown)


akosthekiss:
Motivated by #1902 but with the goal of giving more coverage (not testing api examples only) with less manual work.

tests/unit-doc/CMakeLists.txt
```diff
@@ -0,0 +1,93 @@
...
+    string(TOUPPER ${DOCTEST_ACTION} DOCTEST_ACTION)
+
+    # message(STATUS "doctest found: ${DOCTEST_ACTION} '${DOCTEST_NAME}'")
+    set(DOCTEST_${DOCTEST_ACTION} ${DOCTEST_${DOCTEST_ACTION}} ${DOCTEST_NAME})
+  endif()
+endforeach()
+
+# message(STATUS "doctests to compile: ${DOCTEST_COMPILE}")
+# message(STATUS "doctests to link: ${DOCTEST_LINK}")
+# message(STATUS "doctests to run: ${DOCTEST_RUN}")
```
LaszloLango: Debug comments?

akosthekiss: Yes, leftovers. Will clean them up when/if more review comments arrive.

akosthekiss:
PR updated. Removed the leftover debug messages and added some comments to the new CMakeLists file to document how it's processing the doctests.

CMakeLists.txt
```diff
@@ -53,16 +54,16 @@ set(ENABLE_STATIC_LINK ON  CACHE BOOL "Enable static linking?")
 set(ENABLE_STRIP       ON  CACHE BOOL "Enable stripping all symbols from release binary?")
 
 # Option overrides
-if(JERRY_CMDLINE OR JERRY_CMDLINE_MINIMAL)
+if(JERRY_CMDLINE OR JERRY_CMDLINE_MINIMAL OR UNITTESTS OR DOCTESTS)
```
LaszloLango: Why is the `OR UNITTESTS` condition needed in this patch?

docs/02.API-REFERENCE.md
```diff
@@ -450,9 +467,9 @@ jerry_register_magic_strings  (const jerry_char_ptr_t *ex_str_items_p,
 
   // must be static, because 'jerry_register_magic_strings' does not copy
   static const jerry_length_t magic_string_lengths[] = {
-                                                         (jerry_length_t)strlen (magic_string_items[0]),
-                                                         (jerry_length_t)strlen (magic_string_items[1]),
-                                                         (jerry_length_t)strlen (magic_string_items[2])
+                                                         12,
```
LaszloLango: Why was this change needed?

akosthekiss: `magic_string_lengths` is a `static const` array variable, it cannot be initialized with the return value of a function call, C99 does not allow that. The extracted code did not compile.

CMakeLists.txt
akosthekiss: Both unit-core and unit-ext depends on the jerry-port-default-minimal library. However, somehow this dependency did not appear here until now. (Actually, run-tests.py hid it, because it did not turn off cmdline when building the unit tests, and cmdline forced the building of the default port. So, we have never tried building unittests alone, without the cmdline. If we had, we would have run into a link issue before, too.) So, this is actually an independent addition, but as this line is already touched by this PR, I thought it was OK to fix this little problem here.

LaszloLango: I see. It's OK to fix it in this PR.

docs/02.API-REFERENCE.md
LaszloLango: Thanks for the explanation.

tools/gen-doctest.py
```diff
@@ -0,0 +1,122 @@
+#!/usr/bin/env python
```
LaszloLango: Please try to fix the pylint warnings.

akosthekiss:
@LaszloLango I've updated the PR. The doctest extractor went under a heavy refactoring (now, still in a second commit), which enabled the "fixing" of several pylint warnings. The code got quite some documentation, too. (However, there is one pylint warning remaining: "Too few public methods (1/2) (too-few-public-methods)". I cannot get rid of that, and honestly, I don't think pylint is right here. Some other scripts report the same issue, but instead of "fixing" all that in some artificial ways, I think it would be nice to disable that warning in the future.)

tools/gen-doctest.py
```diff
@@ -0,0 +1,182 @@
...
+        but valid Markdown links/comments that must be added before the fenced code blocks:
+        `[doctest]: # (name="test.c", ...)`. For now, two parameters are valid:
+        `name` determines the filename for the extracted code block (overriding the default
+        auto-numbered naming scheme), and `test` determines the test action to be performed on
+        the extracted code (valid options are "compile", "link", and the default "run").
+        """)
+    parser.add_argument('-d', '--dir', metavar='NAME', default=os.getcwd(),
+                        help='output directory name (default: %(default)s)')
+    parser.add_argument('--dry', action='store_true',
+                        help='don\'t generate files but print file names that would be generated '
```
galpeter: No need to escape the `'` character, just use a `"` around the "outer" string.

tests/unit-doc/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+    set_property(TARGET ${TARGET_NAME} PROPERTY LINK_FLAGS "${LINKER_FLAGS_COMMON}")
+    target_link_libraries(${TARGET_NAME} jerry-ext jerry-core jerry-port-default-minimal)
+  endforeach()
+endif()
+
+# Process "full-fledged" doctests: create an executable from each (named
+# unit-doc-*). Their name prefix (unit-) ensures that the unit test runner
+# script will treat them like any other unit test, i.e., executed them.
+if(NOT ("${DOCTEST_RUN}" STREQUAL ""))
+  foreach(DOCTEST_NAME ${DOCTEST_RUN})
```
galpeter: This block and the previous for-each block looks almost the same. I've only found one difference that is the `TARGET_NAME` will have a different suffix. Would it make sense to have a function which can be used by both for-each block?

tests/unit-doc/CMakeLists.txt
```diff
@@ -0,0 +1,98 @@
...
+    set_property(TARGET ${TARGET_NAME} APPEND_STRING PROPERTY COMPILE_FLAGS "${COMPILE_FLAGS_DOCTEST}")
+    set_property(TARGET ${TARGET_NAME} PROPERTY LINK_FLAGS "${LINKER_FLAGS_COMMON}")
+    target_link_libraries(${TARGET_NAME} jerry-ext jerry-core jerry-port-default-minimal)
+  endforeach()
+endif()
+
+# Process "full-fledged" doctests: create an executable from each (named
+# unit-doc-*). Their name prefix (unit-) ensures that the unit test runner
+# script will treat them like any other unit test, i.e., executed them.
+if(NOT ("${DOCTEST_RUN}" STREQUAL ""))
```
galpeter: Is it really needed to check if the variable is empty? Shouldn't the for-each know what to do with an empty "list"?

tools/gen-doctest.py
akosthekiss: I'm aware of that, but I don't consider good practice to mix the two quotes in a single file. (It's valid, of course. I'm just saying that it's not nice.)

akosthekiss:
PR updated according to the comments of @galpeter

tools/gen-doctest.py
galpeter: Well  we can change all of the other `'` strings to `"` :)

galpeter: but I'm fine with the current one (It's just my personal preference to avoid such escapes as must as possible)

akosthekiss: Yes, it's all about personal preferences. (I made my choice because the repr of strings is delimited by single quotes in python by default, so I try to stick to them whenever possible.) So, if it's not strongly required, I'd keep this single escape here as is.

PR #1910, opened 2017-07-04
Make debugger port runtime configurable

akosthekiss:
Convert debug server port from a compile-time constant to a context
variable. This enables each engine instance (either running in the
same process or in different processes) to listen for connections
at different ports, i.e., multiple engines can be debugged at the
same time on the same machine.


yichoi:
LGTM

PR #1912, opened 2017-07-07
Fix hash calculation for strings created using jerry_create_string_from_utf8

martijnthe:
The hash was being calculated over the bytes *after* the actual string bytes.


martijnthe:
cc @jiangzidong

PR #1915, opened 2017-07-11
Improve pylint runner script

akosthekiss:
- On OSX, `find` is more picky about its arguments than on Linux
  and requires the start directory(-ies) to be explicitly mentioned
  before the expressions (a.k.a. "primaries"). However, until now,
  the `pylint` runner script called `find` without a start dir,
  which caused an error on OSX.

- Until now, `find` looked for python scripts in all subdirectories
  of the project, causing `pylint` to check scripts under `targets/`
  and `tests/test262` (if `test262` was checked out). However, the
  files under `targets/` are less strictly controlled while
  `test262` is completely beyond the control of the project. This
  caused a lot of superfluous lint warnings.

This patch makes the `pylint` runner script call `find` in the
cross-platform way, and also limits the scope of the lint checks to
the `tools/` and `jerry-debugger/` directories.

### DIFF
--- a/tools/check-pylint.sh
+++ b/tools/check-pylint.sh
@@ -14,5 +14,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-find -name "*.py" \
+find ./tools ./jerry-debugger -name "*.py" \
     | xargs pylint --rcfile=tools/pylint/pylintrc


LaszloLango:
I'm not happy with the removal of the `targets` directory from the check.

akosthekiss:
@LaszloLango If I'm not mistaken, only two python files are under `targets`: `./targets/curie_bsp/setup.py` and `./targets/mbedos5/tools/generate_pins.py`. I'm not sure we should/could improve them much, so they will keep on triggering lint warnings.

LaszloLango:
@akosthekiss OK. I double checked and you are right. Only those two python files remained under the targets folder.

If you started to improve the pylint checker then it would be great to tune a bit the config file too. Related to discussion on https://github.com/jerryscript-project/jerryscript/pull/1909#issuecomment-314518688 I suggest to change the required public methods number from 2 to 1. It would be great if we could remove this check from the allowed fail jobs too in this PR.

LaszloLango:
It looks like there are still too many errors to enable the check on Travis. It's OK to do it on a follow up work.

PR #1921, opened 2017-07-14
Fix empty string issue in ext/args

jiangzidong:
Now the js string "" can be converted to C array by jerryx_arg_string

### DIFF
--- a/jerry-ext/arg/arg-transform-functions.c
+++ b/jerry-ext/arg/arg-transform-functions.c
@@ -259,10 +259,10 @@ jerryx_arg_string_common_routine (jerry_value_t js_arg, /**< JS arg */
                                                    target_p,
                                                    target_buf_size);
 
-  if (size == 0 || size == target_buf_size)
+  if ((size == target_buf_size) || (size == 0 && jerry_get_string_length (js_arg) != 0))
   {
     return jerry_create_error (JERRY_ERROR_TYPE,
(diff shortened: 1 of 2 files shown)


jerry-ext/arg/arg-transform-functions.c
```diff
@@ -259,10 +259,17 @@ jerryx_arg_string_common_routine (jerry_value_t js_arg, /**< JS arg */
                                                    target_p,
                                                    target_buf_size);
 
-  if (size == 0 || size == target_buf_size)
+
+  if (size == target_buf_size)
+  {
+    return jerry_create_error (JERRY_ERROR_TYPE,
+                               (jerry_char_t *) "Buffer has no space to store '\0'.");
```
martijnthe: Why the separate error?

jiangzidong: if size == target_buf_size, then there is no space to store `\0`, so we should throw error

martijnthe: Yeah, what I meant is why not the same error message as below? In other words, just fold the conditions together in one `if`:

```
if ((size == target_buf_size) || (size == 0 && jerry_get_string_length (js_arg) != 0)) {
  return jerry_create_error(....);
}
```

jiangzidong: Oh, I know what you mean. will update soon

martijnthe:
lgtm (informally)

PR #1923, opened 2017-07-17
Add API for get the memory stats

jiangzidong:
During our developing runtime based on Jerry, we find that it is needed to get the heap stats inside jerry engine. e.g. to impl `performance.memory` 
https://developer.mozilla.org/en-US/docs/Web/API/Window/performance
https://anonyco.github.io/wiki/apis/timing/properties/memory/
```
console.log(performance.memory)

// Would show, for example
//{
// jsHeapSizeLimit: 767557632,
// totalJSHeapSize: 58054528,
// usedJSHeapSize: 42930044
//}
```
In this PR, I added `jerry_get_memory_stats` in API.

And during test, I found the current `jmem_heap_stats_print` have `divided by 0` issue (block test-heap), and I quickly fixed it (not a decent way)
 
### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -169,6 +169,26 @@ typedef struct
 } jerry_property_descriptor_t;
 ```
 
+## jerry_heap_stats_t
+
+**summary**
+
+Description of JerryScript heap memory stats.
+It is for memory profiling.
+
+**Prototype**
+
+```c
+typedef struct
+{
+  size_t version /**< the version of the stats struct */
+  size_t size; /**< heap total size */
+  size_t allocated_bytes; /**< currently allocated bytes */
+  size_t peak_allocated_bytes; /**< peak allocated bytes */
+  size_t reserved[4]; /**< padding for future extensions */
+} jerry_heap_stats_t;
+```
+
 ## jerry_external_handler_t
 
 **Summary**
@@ -521,6 +541,36 @@ main (void)
 
 - [jerry_init](#jerry_init)
 - [jerry_cleanup](#jerry_cleanup)
+
+
+## jerry_get_memory_stats
+
+**Summary**
+
+Get heap memory stats.
+
+**Prototype**
+
+```c
+bool
+jerry_get_memory_stats (jerry_heap_stats_t *out_stats_p);
+```
+
+- `out_stats_p` - out parameter, that provides the heap statistics.
+- return value
+  - true, if run was successful
+  - false, otherwise. Usually it is because the MEM_STATS feature is not enabled.
+
+**Example**
+
+```c
+jerry_heap_stats_t stats = {0};
+bool get_stats_ret = jerry_get_memory_stats (&stats);
+```
+
+**See also**
+
+- [jerry_init](#jerry_init)
 
 
 ## jerry_gc
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -33,6 +33,7 @@
 #include "ecma-promise-object.h"
 #include "jcontext.h"
 #include "jerryscript.h"
+#include "jmem.h"
 #include "js-parser.h"
 #include "re-compiler.h"
 
@@ -253,6 +254,39 @@ jerry_gc (void)
 
   ecma_gc_run (JMEM_FREE_UNUSED_MEMORY_SEVERITY_LOW);
 } /* jerry_gc */
+
+/**
+ * Get heap memory stats.
+ *
+ * @return true - get the heap stats successful
+ *         false - otherwise. Usually it is because the MEM_STATS feature is not enabled.
+ */
+bool
+jerry_get_memory_stats (jerry_heap_stats_t *out_stats_p) /**< [out] heap memory stats */
+{
+#ifdef JMEM_STATS
+  if (out_stats_p == NULL)
+  {
+    return false;
+  }
+
+  jmem_heap_stats_t jmem_heap_stats = {0};
+  jmem_heap_get_stats (&jmem_heap_stats);
+
+  *out_stats_p = (jerry_heap_stats_t)
+  {
+    .version = 1,
+    .size = jmem_heap_stats.size,
+    .allocated_bytes = jmem_heap_stats.allocated_bytes,
+    .peak_allocated_bytes = jmem_heap_stats.peak_allocated_bytes
+  };
+
+  return true;
+#else
+  JERRY_UNUSED (out_stats_p);
+  return false;
+#endif
+} /* jerry_get_memory_stats */
 
 /**
  * Simple Jerry runner
--- a/jerry-core/include/jerryscript-core.h
+++ b/jerry-core/include/jerryscript-core.h
@@ -159,6 +159,19 @@ typedef struct
 } jerry_property_descriptor_t;
 
 /**
+ * Description of JerryScript heap memory stats.
+ * It is for memory profiling.
+ */
+typedef struct
+{
+  size_t version; /**< the version of the stats struct */
+  size_t size; /**< heap total size */
+  size_t allocated_bytes; /**< currently allocated bytes */
+  size_t peak_allocated_bytes; /**< peak allocated bytes */
+  size_t reserved[4]; /**< padding for future extensions */
+} jerry_heap_stats_t;
+
+/**
  * Type of an external function handler.
  */
 typedef jerry_value_t (*jerry_external_handler_t) (const jerry_value_t function_obj,
@@ -234,6 +247,8 @@ void jerry_register_magic_strings (const jerry_char_ptr_t *ex_str_items_p, uint3
 void jerry_get_memory_limits (size_t *out_data_bss_brk_limit_p, size_t *out_stack_limit_p);
 void jerry_gc (void);
 void *jerry_get_context_data (const jerry_context_data_manager_t *manager_p);
+
+bool jerry_get_memory_stats (jerry_heap_stats_t *out_stats_p);
 
 /**
  * Parser and executor functions.
--- a/jerry-core/jmem/jmem-allocator-internal.h
+++ b/jerry-core/jmem/jmem-allocator-internal.h
@@ -25,7 +25,6 @@
  */
 
 #ifdef JMEM_STATS
-void jmem_heap_get_stats (jmem_heap_stats_t *);
 void jmem_heap_stats_reset_peak (void);
 void jmem_heap_stats_print (void);
 #endif /* JMEM_STATS */
--- a/jerry-core/jmem/jmem.h
+++ b/jerry-core/jmem/jmem.h
@@ -161,6 +161,7 @@ void jmem_stats_free_object_bytes (size_t string_size);
 void jmem_stats_allocate_property_bytes (size_t property_size);
 void jmem_stats_free_property_bytes (size_t property_size);
 
+void jmem_heap_get_stats (jmem_heap_stats_t *);
 #endif /* JMEM_STATS */
 
 jmem_cpointer_t jmem_compress_pointer (const void *pointer_p) __attr_pure___;
--- a/tests/unit-core/test-heap.c
+++ b/tests/unit-core/test-heap.c
@@ -84,7 +84,8 @@ main (void)
   jmem_register_free_unused_memory_callback (test_heap_give_some_memory_back);
 
 #ifdef JMEM_STATS
-  jmem_heap_stats_print ();
+  // TODO: uncomment it after we solve the divide zero issue.
+  // jmem_heap_stats_print ();
 #endif /* JMEM_STATS */
 
   for (uint32_t i = 0; i < test_iters; i++)
--- a/tests/unit-core/test-mem-stats.c
+++ b/tests/unit-core/test-mem-stats.c
@@ -1,0 +1,60 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript.h"
+#include "test-common.h"
+
+#ifdef JMEM_STATS
+
+const char *test_source = (
+                           "var a = 'hello';"
+                           "var b = 'world';"
+                           "var c = a + ' ' + b;"
+                           );
+
+int main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+  jerry_value_t parsed_code_val = jerry_parse ((jerry_char_t *) test_source, strlen (test_source), false);
+  TEST_ASSERT (!jerry_value_has_error_flag (parsed_code_val));
+
+  jerry_value_t res = jerry_run (parsed_code_val);
+  TEST_ASSERT (!jerry_value_has_error_flag (res));
+
+  jerry_heap_stats_t stats = {0};
+  bool get_stats_ret = jerry_get_memory_stats (&stats);
+  TEST_ASSERT (get_stats_ret);
+  TEST_ASSERT (stats.version == 1);
+  TEST_ASSERT (stats.size == 524280);
+
+  TEST_ASSERT (!jerry_get_memory_stats (NULL));
+
+  jerry_release_value (res);
+  jerry_release_value (parsed_code_val);
+
+  jerry_cleanup ();
+
+  return 0;
+} /* main */
+
+#else /* JMEM_STATS */
+
+int
+main (void)
+{
+  return 0;
+} /* main */
+
+#endif /* !JMEM_STATS */
--- a/tests/unit-core/test-poolman.c
+++ b/tests/unit-core/test-poolman.c
@@ -83,7 +83,7 @@ main (void)
   }
 
 #ifdef JMEM_STATS
-  jmem_pools_stats_print ();
+  jmem_heap_stats_print ();
 #endif /* JMEM_STATS */
 
   jmem_finalize ();
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -42,9 +42,9 @@ def get_binary_path(bin_dir_path):
 # Test options for unittests
 JERRY_UNITTESTS_OPTIONS = [
     Options('unittests',
-            ['--unittests', '--jerry-cmdline=off', '--error-messages=on', '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset']),
+            ['--unittests', '--jerry-cmdline=off', '--error-messages=on', '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset', '--mem-stats=on']),
     Options('unittests-debug',
-            ['--unittests', '--jerry-cmdline=off', '--debug', '--error-messages=on', '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset']),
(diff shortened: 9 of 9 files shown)


jiangzidong:
cc @martijnthe

docs/02.API-REFERENCE.md
```diff
@@ -169,6 +169,51 @@ typedef struct
...
+  size_t peak_property_bytes; /**< peak allocated memory for properties */
+
+  size_t skip_count; /**< Number of skip-aheads during insertion of free block */
+  size_t nonskip_count; /**< Number of times we could not skip ahead during
+                         *   free block insertion */
+
+  size_t alloc_count; /**< number of memory allocations */
+  size_t free_count; /**< number of memory frees */
+  size_t alloc_iter_count; /**< Number of iterations required for allocations */
+  size_t free_iter_count; /**< Number of iterations required for inserting free blocks */
```
zherczeg: I would really not make this structure public. It contains implementation details which might change.

LaszloLango: I agree

zherczeg:
We could have a public structure with version number to publish this data. So we could extend the structure later without breaking compatibility.

jiangzidong:
@zherczeg @LaszloLango What about exposing the following struct?
```c
typedef struct
{
  size_t size; /**< heap total size */
  size_t allocated_bytes; /**< currently allocated bytes */
  size_t peak_allocated_bytes; /**< peak allocated bytes */
  uint8_t version;
} jerry_heap_stats_t
```
Because in the current stage, we are mainly interested in those members.

zherczeg:
Something like this. As for version number, it can be part of the struct, or passed directly to the function. I would prefer the latter since it is more robust. Anyway, if it is part of the struct, it must be the first member.

jiangzidong:
Thanks, but I think the function call like `jerry_get_memory_stats(out_stats_p, 1)`, the version number here is a bit weird, because the caller need to know exactly the version of the mem-stats struct.

I will firstly impl the "version inside struct" way.

docs/02.API-REFERENCE.md
```diff
@@ -523,6 +544,40 @@ main (void)
...
+
+- `out_stats_p` - out parameter, that gives the stats of the heap.
+- return value
+  - true, if run was successful
+  - false, otherwise. Usually it is because the MEM_STATS feature is not enabled.
+
+**Example**
+
+```c
+jerry_init (JERRY_INIT_EMPTY | JERRY_INIT_MEM_STATS);
```
martijnthe: I was trying to find how `JERRY_INIT_MEM_STATS` affects the engine but grepping for this define didn't lead to much insight. Can someone point me to how this flag is actually being consumed?

jerry-core/api/jerry.c
```diff
@@ -254,6 +255,34 @@ jerry_gc (void)
...
+ * @return true - get the heap stats successful
+ *         false - otherwise. Usually it is because the MEM_STATS feature is not enabled.
+ */
+bool
+jerry_get_memory_stats (jerry_heap_stats_t *out_stats_p) /**< [out] heap memory stats */
+{
+#ifdef JMEM_STATS
+  jmem_heap_stats_t jmem_heap_stats = {0};
+  jmem_heap_get_stats (&jmem_heap_stats);
+
```
martijnthe: check for `NULL == out_stats_p` and return `false`?

jerry-core/api/jerry.c
```diff
@@ -254,6 +255,34 @@ jerry_gc (void)
...
+  out_stats_p->version = 1;
```
martijnthe: I would make sure to completely initialize the struct.
Personally I like doing this as such:

```
*out_stats_p = (jerry_heap_stats_t) {
  .size-> = ...,
 // ...
 // if more fields are added to the struct in the future, but we forget them here,
 // they will at least be initialized to zero (instead of being garbage).
};
```

fbmrk:
@jiangzidong 
Possibly the `waste_bytes` also could be useful information, for example before a `gc()` call.
What do you think?

jerry-core/include/jerryscript-core.h
```diff
@@ -159,6 +159,18 @@ typedef struct
 } jerry_property_descriptor_t;
 
 /**
+ * Description of JerryScript heap memory stats.
+ * It is for memory profiling.
+ */
+typedef struct
+{
+  size_t version; /**< the version of the stats struct */
```
martijnthe: I'm not really convinced `version` is needed.
IMO, the struct and field types and names are a semantic API contract, but IMO the struct layout isn't an ABI (the exact layout is target- & compiler dependent after all...)
I think it's safe to assume that the users of JerryScript have control over what version of Jerry they use.
If the semantics change (i.e. field is renamed, types are changed, ...) the compiler will tell you.

jiangzidong: @zherczeg @LaszloLango @fbmrk @martijnthe 
What about 
```
typedef struct
{
  size_t size; /**< heap total size */
  size_t allocated_bytes; /**< currently allocated bytes */
  size_t peak_allocated_bytes; /**< peak allocated bytes */
  size_t waste_bytes; /**< bytes waste due to blocks filled partially */
  size_t peak_waste_bytes;
}  jerry_heap_stats_t
```
I think those fields are not much impl-dependent comparing to `alloc_count`, `alloc_iter_count`.
And maybe we can add `byte_code_bytes`, `string_bytes`, and so on.

docs/02.API-REFERENCE.md
jiangzidong: in jmem-allocator.c
```c
void
jmem_finalize (void)
{
  jmem_pools_finalize ();

#ifdef JMEM_STATS
  if (JERRY_CONTEXT (jerry_init_flags) & ECMA_INIT_MEM_STATS)
  {
    jmem_stats_print ();
  }
#endif /* JMEM_STATS */

  jmem_heap_finalize ();
} /* jmem_finalize */
```
I think the only usage of "JERRY_INIT_MEM_STATS" is dumping the mem stats after cleaning up. So you're right, we don't need it here, i will remove it.

jerry-core/include/jerryscript-core.h
zherczeg: The problem is if you dynamically link Jerry, and you compiled your code with an older Jerry, and the system has a newer one, you will experience a crash. Currently most projects statically link Jerry, so this is not a problem, but it might be if some distros will support it.

jiangzidong:
@zherczeg @fbmrk @martijnthe 
updated the patch

jerry-core/include/jerryscript-core.h
martijnthe: > The problem is if you dynamically link Jerry, and you compiled your code with an older Jerry, and the system has a newer one, you will experience a crash.

In the current implementation doesn't know or check the version / how large the caller's `jerry_heap_stats_t` struct is. I think the caller should be passing in the desired/expected version then for this to work?

martijnthe: @zherczeg I don't know how versioning of APIs with dynamic libraries this is usually dealt with, but I wonder whether it makes sense to have a over-arching "get version" API for version matching instead of adding versions for each and every API / ABI.

zherczeg: But you don't know the changes in future versions. At compile time you see the correct version, but when dynamically linked, you get a library with different version.

docs/02.API-REFERENCE.md
```diff
@@ -523,6 +546,36 @@ main (void)
...
+
+- `out_stats_p` - out parameter, that gives the stats of the heap.
+- return value
+  - true, if run was successful
+  - false, otherwise. Usually it is because the MEM_STATS feature is not enabled.
+
+**Example**
+
+```c
+jerry_heap_stats_t stats = {0};
```
zherczeg: I would prefer a textual version number.

docs/02.API-REFERENCE.md
```diff
@@ -169,6 +169,29 @@ typedef struct
...
+typedef struct
+{
+  size_t version /**< the version of the stats struct */
+  size_t size; /**< heap total size */
+  size_t allocated_bytes; /**< currently allocated bytes */
+  size_t peak_allocated_bytes; /**< peak allocated bytes */
+  size_t waste_bytes; /**< bytes waste due to blocks filled partially */
+  size_t peak_waste_bytes; /**< peak wasted bytes */
+} jerry_heap_stats_t;
+
```
LaszloLango: Unnecessary empty line

docs/02.API-REFERENCE.md
jiangzidong: Sorry, a little confused. Do you mean the `version` is a input param and  `jerry_get_mem_stats` should check the `version`? But in this way, user must know the version clearly and explicitly. I doubt if it is a  good way.

jerry-core/api/jerry.c
```diff
@@ -254,6 +255,44 @@ jerry_gc (void)
   ecma_gc_run (JMEM_FREE_UNUSED_MEMORY_SEVERITY_LOW);
 } /* jerry_gc */
 
+#ifdef JMEM_STATS
+void jmem_heap_get_stats (jmem_heap_stats_t *);
```
LaszloLango: Why is this deceleration needed here?

jerry-core/include/jerryscript-core.h
LaszloLango: The heap size is set at compile time. Not sure it is a useful information.

martijnthe: > But you don't know the changes in future versions. At compile time you see the correct version, but when dynamically linked, you get a library with different version.
 
Sure, I understand.

> The heap size is set at compile time. Not sure it is a useful information.

Not if you use `jerry_create_instance`.

jerry-core/api/jerry.c
```diff
@@ -254,6 +255,44 @@ jerry_gc (void)
...
+{
+#ifdef JMEM_STATS
+  if (out_stats_p == NULL)
+  {
+    return false;
+  }
+
+  jmem_heap_stats_t jmem_heap_stats = {0};
+  jmem_heap_get_stats (&jmem_heap_stats);
+
```
martijnthe: Thinking this further:

Say today `sizeof(jerry_heap_stats_t) == 24`.
Say I dynamically link my project against JerryScript of today, so my project does something like:
```
jerry_heap_stats_t stats = {}; // 24 bytes allocated on the stack
jerry_get_memory_stats(&stats); // 24 bytes written
```

Now tomorrow I have a newer JerryScript, which happened to have added some fields to `jerry_heap_stats_t`, so now `sizeof(jerry_heap_stats_t) == 36`.

Now it will be a problem if the project dynamically links to the newer JerryScript:
```
jerry_heap_stats_t stats = {}; // still 24 bytes allocated on the stack
jerry_get_memory_stats(&stats); // 36 bytes written -- WHOOPS, smashing the stack!
```

That's what I meant with "I think the caller should be passing in the desired/expected version then for this to work" in [my earlier comment].

cc @zherczeg 

[my earlier comment]: https://github.com/jerryscript-project/jerryscript/pull/1923#discussion_r128181423

jerry-core/api/jerry.c
jiangzidong: or should I use
```
#define JMEM_ALLOCATOR_INTERNAL
#include "jmem-allocator-internal.h"
```

jerry-core/api/jerry.c
zherczeg: We might talk about the same thing.

    if (jerry_get_memory_stats(VERSION, &stats))

This work regardless of future changes. Each caller gets the data in their expected format. I do think memory statistics is something I expect to change in the future.

jiangzidong: @zherczeg is that means the jerry_get_memory_stats will check the version, if it is not match, then return false ?

zherczeg: Not necessarily. It can support multiple variants since the version determines the request. So the implementation will be a switch-case which fills the appropriate fields, or false if no mem-stats or version is unknown.

jiangzidong: Ok, `if (jerry_get_memory_stats(VERSION, &stats))` the VERSION here should be sth like `JERRY_API_MAJOR_VERSION`. I still feel strange because it is the only API who pass the api version. 
e.g.
`jerry_get_own_property_descriptor` is also getting a block of memory as jerry_property_descriptor_t, but it doesnt pass the api version.

zherczeg: True, but I feel this is something which will likely change in the future. I have an alternative idea: the version can be encoded in the function name: `jerry_get_memory_stats_1` and people can choose the appropriate one. The compiler will only link the appropriate one.

martijnthe: @zherczeg Personally, I was also surprised about your concerns w.r.t. version mismatches when dynamic linking, because – as @jiangzidong also pointed out – the rest of the API is not really taking that concern into account at all. Breaking ABI changes have been made in the past without anyone complaining that their dynamically linked libraries became incompatible.

I also still feel it makes the API pretty ugly and heavy if you have to pass in versions all over the place. I'd rather have a over-arching "JerryScript version" that the caller can use to check for compatibility.

zherczeg: Normally I don't see the reason for versioning, because most functions are simple enough and will likely not change (e.g. I don't think an `is_boolean` will ever change). But I feel memory statistics (as the name suggests) is different. It would be great to have at least a plan for what will happen then.

martijnthe: Another idea:
Add a bit of padding to the struct for future use, for example something like:
```
typedef struct
{
  int version;
  size_t size; /**< heap total size */
  size_t allocated_bytes; /**< currently allocated bytes */
  size_t peak_allocated_bytes; /**< peak allocated bytes */
  uint32_t reserved[4];
}  jerry_heap_stats_t
```

It won't be extensible indefinitely of course, but it's simple and wouldn't require passing in the version in the call. We'd promise not to grow the struct ever. The caller can check the version after the call.

zherczeg: Sounds reasonable. If we ever need a much bigger structure, we can create a new function for it. Please make the type of reserved as `size_t`, not `uint32_t`.

jiangzidong:
Please review again, @zherczeg @LaszloLango @martijnthe

jerry-core/api/jerry.c
LaszloLango: I don't like the reserved fields. But this is only my personal preference. My experience is that they are always remaining intact. This is not an objection. If everyone vote for reserved fields than I can accept that. 

> But I feel memory statistics (as the name suggests) is different. It would be great to have at least a plan for what will happen then.

I think this won't change frequently. @zherczeg do you have any plan with memory statistics or management in the near future?

martijnthe:
lgtm

martijnthe:
@LaszloLango could you please take another look? Thanks!

docs/02.API-REFERENCE.md
```diff
@@ -169,6 +169,29 @@ typedef struct
...
+{
+  size_t version /**< the version of the stats struct */
+  size_t size; /**< heap total size */
+  size_t allocated_bytes; /**< currently allocated bytes */
+  size_t peak_allocated_bytes; /**< peak allocated bytes */
+  size_t waste_bytes; /**< bytes waste due to blocks filled partially */
+  size_t peak_waste_bytes; /**< peak wasted bytes */
+  size_t reserved[4]; /**< the padding for future extension */
+} jerry_heap_stats_t;
+
```
LaszloLango: Unnecessary empty line

docs/02.API-REFERENCE.md
```diff
@@ -169,6 +169,28 @@ typedef struct
...
+```c
+typedef struct
+{
+  size_t version /**< the version of the stats struct */
+  size_t size; /**< heap total size */
+  size_t allocated_bytes; /**< currently allocated bytes */
+  size_t peak_allocated_bytes; /**< peak allocated bytes */
+  size_t waste_bytes; /**< bytes waste due to blocks filled partially */
+  size_t peak_waste_bytes; /**< peak wasted bytes */
```
zherczeg: Do we need wasted data? It seems allocator specific, and we cannot provide for normal allocators.

jiangzidong: It's fbmrk's idea. What do you think? @fbmrk 
Personally, I also agree that the `waste` is much more impl-specific than `allocated` and `total`

fbmrk: It was just a suggestion, i do not insist on it.

jiangzidong: Ok, I will remove it in this patch. If someday we really need it, let's put it into another PR.

jiangzidong: @zherczeg update, please review again @fbmrk @zherczeg

docs/02.API-REFERENCE.md
```diff
@@ -169,6 +169,26 @@ typedef struct
...
+**Prototype**
+
+```c
+typedef struct
+{
+  size_t version /**< the version of the stats struct */
+  size_t size; /**< heap total size */
+  size_t allocated_bytes; /**< currently allocated bytes */
+  size_t peak_allocated_bytes; /**< peak allocated bytes */
+  size_t reserved[4]; /**< the padding for future extension */
```
zherczeg: padding for future extensions

docs/02.API-REFERENCE.md
```diff
@@ -523,6 +543,36 @@ main (void)
...
+Get heap memory stats.
+
+**Prototype**
+
+```c
+bool
+jerry_get_memory_stats (jerry_heap_stats_t *out_stats_p);
+```
+
+- `out_stats_p` - out parameter, that gives the stats of the heap.
```
zherczeg: that provides the heap statistics

jerry-core/api/jerry.c
```diff
@@ -33,9 +33,14 @@
 #include "ecma-promise-object.h"
 #include "jcontext.h"
 #include "jerryscript.h"
+#include "jmem.h"
 #include "js-parser.h"
 #include "re-compiler.h"
 
+#define JMEM_ALLOCATOR_INTERNAL
+#include "jmem-allocator-internal.h"
+#undef JMEM_ALLOCATOR_INTERNAL
```
zherczeg: Why do we need this include? If something is needed, we should move it to jmem.h.

jerry-core/jmem/jmem-heap.c
```diff
@@ -615,12 +615,12 @@ jmem_heap_stats_print (void)
...
-                   heap_stats->alloc_iter_count / heap_stats->alloc_count,
-                   heap_stats->alloc_iter_count % heap_stats->alloc_count * 10000 / heap_stats->alloc_count,
-                   heap_stats->free_iter_count / heap_stats->free_count,
-                   heap_stats->free_iter_count % heap_stats->free_count * 10000 / heap_stats->free_count);
+                   heap_stats->skip_count / (heap_stats->nonskip_count + 1),
+                   heap_stats->skip_count % (heap_stats->nonskip_count + 1) * 10000 / (heap_stats->nonskip_count + 1),
+                   heap_stats->alloc_iter_count / (heap_stats->alloc_count + 1),
+                   heap_stats->alloc_iter_count % (heap_stats->alloc_count + 1) * 10000 / (heap_stats->alloc_count + 1),
+                   heap_stats->free_iter_count / (heap_stats->free_count + 1),
+                   heap_stats->free_iter_count % (heap_stats->free_count + 1) * 10000 / (heap_stats->free_count + 1));
```
zherczeg: I suspect the +1 is needed for avoiding division by zero, but the formula is not correct for non-zero values. I would introduce a helper function for this. Anyway this change seems unrelated, and perhaps we could fix it in another patch.

jiangzidong: yes, to avoid the "divide by 0" exception

zherczeg: What about fixing it in another patch?

jiangzidong: because the unittest `test-poolmain,c` calls `jmem_heap_stats_print`, which will cause error if we don't fix it. (JMEM_STATS is not enabled before, so the issue is not be discovered)

What about comments out the function call, and I will reset the change for function body.

zherczeg: Hm, ok. Lets disable it for now (leave a TODO comment). This issue may affect other things, so it deserves its own patch.

jiangzidong: Ok, updated, please review it again. Thanks

PR #1931, opened 2017-07-19
target: mbedos5: Improve setInterval and setTimeout

fbmrk:
In setInterval and setTimeout there must be a reference increase to the given function.
If you use `jerry_acquire` and you want to cleanup the engine, 
you get an assertion, because this value is released nowhere in the code.
In my opinion it is better to set the function as a property, so it is released automatically.

In launcher.cpp there is wrong error handle. If `returned_value` has error flag, `parsed_code` is not released.

### DIFF
--- a/targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/source/setInterval-js.cpp
+++ b/targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/source/setInterval-js.cpp
@@ -28,11 +28,20 @@ DECLARE_GLOBAL_FUNCTION(setInterval) {
     CHECK_ARGUMENT_TYPE_ALWAYS(global, setInterval, 0, function);
     CHECK_ARGUMENT_TYPE_ALWAYS(global, setInterval, 1, number);
 
-    jerry_acquire_value(args[0]);
     int interval = int(jerry_get_number_value(args[1]));
 
     int id = mbed::js::EventLoop::getInstance().getQueue().call_every(interval, jerry_call_function, args[0], jerry_create_null(), (jerry_value_t*)NULL, 0);
 
+    jerry_value_t result = jerry_set_property_by_index(function_obj_p, id, args[0]);
+
+    if (jerry_value_has_error_flag(result)) {
+        jerry_release_value(result);
+        mbed::js::EventLoop::getInstance().getQueue().cancel(id);
+
+        return jerry_create_error(JERRY_ERROR_TYPE, (const jerry_char_t *) "Failed to run setInterval");
+    }
+
+    jerry_release_value(result);
     return jerry_create_number(id);
 }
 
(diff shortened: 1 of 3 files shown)


fbmrk:
cc @janjongboom

targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/source/setInterval-js.cpp
```diff
@@ -28,11 +28,20 @@ DECLARE_GLOBAL_FUNCTION(setInterval) {
     CHECK_ARGUMENT_TYPE_ALWAYS(global, setInterval, 0, function);
     CHECK_ARGUMENT_TYPE_ALWAYS(global, setInterval, 1, number);
 
-    jerry_acquire_value(args[0]);
     int interval = int(jerry_get_number_value(args[1]));
 
     int id = mbed::js::EventLoop::getInstance().getQueue().call_every(interval, jerry_call_function, args[0], jerry_create_null(), (jerry_value_t*)NULL, 0);
 
+    jerry_value_t result = jerry_set_property_by_index(this_obj, id, args[0]);
```
akosthekiss: But this will pollute the global object, won't it? Basically, it's OK with me to assign `args[0]` to a property _somewhere_ as it will simplify cleanup indeed. However, global object might not be the best choice. But the function object of `setInterval` might work better, perhaps (i.e., `function_obj_p` instead of `this_obj`).

One more question occurred to me: when will these properties be removed? Only at cleanup time? Cannot we drop the reference (delete the property) once the event is canceled (in `clearInterval`)?

targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/source/setTimeout-js.cpp
```diff
@@ -28,11 +28,20 @@ DECLARE_GLOBAL_FUNCTION(setTimeout) {
     CHECK_ARGUMENT_TYPE_ALWAYS(global, setTimeout, 0, function);
     CHECK_ARGUMENT_TYPE_ALWAYS(global, setTimeout, 1, number);
 
-    jerry_acquire_value(args[0]);
     int interval = int(jerry_get_number_value(args[1]));
 
     int id = mbed::js::EventLoop::getInstance().getQueue().call_in(interval, jerry_call_function, args[0], jerry_create_null(), (jerry_value_t*)NULL, 0);
 
+    jerry_value_t result = jerry_set_property_by_index(this_obj, id, args[0]);
```
akosthekiss: Same here, might be better not polluting the global object.

And also the question about dropping the reference: property should be cleared when timeout in explicitly canceled (via `clearTimeout`) _but also_ if/after the event fired (because this is a non-repeating one-shot event, if I'm right). (I'm not sure whether the second scenario can be handled easily, because I don't know whether we get notified about that anywhere and/or whether we still have reference to `function_obj_p` and `id` there.)

fbmrk:
@akosthekiss updated

LaszloLango:
@janjongboom may we land this patch?

LaszloLango:
@janjongboom ping :)

janjongboom:
@LaszloLango LGTM

LaszloLango:
@janjongboom thank you for checking it.

PR #1932, opened 2017-07-19
Add source sending feature to the debugger.

knightburton:
With this feature the debugger webIDE and the python client can able to send a source code to the debugger while that is running in wait mode. This feature can be activated with the --debugger-wait-source switch and the debugger will wait for the source messages. If every message part are received the debugger will continue the exectuion with the initalized options.


jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,63 @@ jerry_debugger_cleanup (void)
   }
 #endif /* JERRY_DEBUGGER */
 } /* jerry_debugger_cleanup */
+
+/**
+ * Sets whether the engine should wait for a source.
+ */
+void
+jerry_debugger_wait_source (void)
```
zherczeg: I feel the name is not the best (it actually runs the code, not just wait for it), but unsure about a better name. Some ideas:

`jerry_debugger_wait_and_run_source_code` or `jerry_debugger_run_client_source_code`?

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,63 @@ jerry_debugger_cleanup (void)
   }
 #endif /* JERRY_DEBUGGER */
 } /* jerry_debugger_cleanup */
+
+/**
+ * Sets whether the engine should wait for a source.
+ */
+void
```
zherczeg: Could be `bool` or `jerry_value_t` to tell about what happened (e.g. the returned value of the jerry_run on success, some error otherwise).

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,63 @@ jerry_debugger_cleanup (void)
...
+    while (true)
+    {
+      if (jerry_debugger_receive (&client_source_data_p))
+      {
+        if (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONNECTED))
+        {
+          break;
+        }
+
+        if (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CLIENT_SOURCE_MODE))
```
zherczeg: Add a comment that the source is arrived.

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,63 @@ jerry_debugger_cleanup (void)
...
+        {
+          jerry_char_t *string_p = (jerry_char_t *) (client_source_data_p + 1);
+          size_t name_size = strlen ((char *) string_p);
```
zherczeg: `(const char *)`

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,63 @@ jerry_debugger_cleanup (void)
...
+
+      jerry_debugger_sleep (JERRY_DEBUGGER_TIMEOUT);
+    }
+
+    JERRY_ASSERT (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CLIENT_SOURCE_MODE)
+                  || !(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONNECTED));
+
+    if (client_source_data_p != NULL)
+    {
+      jmem_heap_free_block (client_source_data_p,
```
zherczeg: Add a comment that the data may partly arrived.

jerry-core/debugger/debugger-ws.c
```diff
@@ -433,16 +433,21 @@ JERRY_STATIC_ASSERT (JERRY_DEBUGGER_MAX_RECEIVE_SIZE < 126,
...
 bool
-jerry_debugger_receive (void)
+jerry_debugger_receive (jerry_debugger_string_data_t **message_data_p) /**< custom message data */
 {
   JERRY_ASSERT (JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONNECTED);
 
+  if ((JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_BREAKPOINT_MODE)
+      || (JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CLIENT_SOURCE_MODE))
+  {
+    JERRY_ASSERT (message_data_p != NULL);
```
zherczeg: Please don't use this construct, write a single assert:

    JERRY_ASSERT (message_data_p != NULL
                  || (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_BREAKPOINT_MODE)
                      && !(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CLIENT_SOURCE_MODE));

jerry-core/debugger/debugger-ws.h
```diff
@@ -64,6 +64,23 @@ typedef struct
...
+} jerry_debugger_receive_string_data_part_t;
+
+/**
+ * String data for eval and client source.
+ */
+typedef struct
+{
+  uint32_t string_size; /**< total size of the client source */
+  uint32_t string_offset; /**< current offset in the client source */
+} jerry_debugger_string_data_t;
```
zherczeg: No type for this message?

jerry-core/debugger/debugger-ws.h
```diff
@@ -85,7 +102,7 @@ bool jerry_debugger_accept_connection (void);
 void jerry_debugger_close_connection (void);
 
 bool jerry_debugger_send (size_t data_size);
-bool jerry_debugger_receive (void);
+bool jerry_debugger_receive (jerry_debugger_string_data_t **message_data_p);
```
zherczeg: Hm, perhaps `jerry_debugger_uint8_data_t` would be better, because in the future we might receive non-string data, although it will still be a byte data. And `jerry_debugger_string_data_t` -> `jerry_debugger_uint8_data_t`

jerry-core/debugger/debugger-ws.c
zherczeg: Another idea: we could create a define:

    #define JERRY_DEBUGGER_RECEIVE_DATA_MODE \
      (JERRY_DEBUGGER_BREAKPOINT_MODE | JERRY_DEBUGGER_CLIENT_SOURCE_MODE)

    JERRY_ASSERT (message_data_p != NULL
                  || (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_RECEIVE_DATA_MODE)

You can use a better symbol name of course.

jerry-core/debugger/debugger.h
```diff
@@ -288,27 +296,23 @@ typedef struct
...
-  uint32_t eval_size; /**< total size of the eval string */
-  uint32_t eval_offset; /**< current offset in the eval string */
-} jerry_debugger_eval_data_t;
+  uint8_t type; /**< type of the message */
+  uint8_t code_size[sizeof (uint32_t)]; /**< total size of the message */
+} jerry_debugger_receive_client_source_first_t;
 
 void jerry_debugger_free_unreferenced_byte_code (void);
 
+void jerry_debugger_sleep (unsigned milliseconds);
```
zherczeg: Since we always pass `JERRY_DEBUGGER_TIMEOUT`, this function could work without argument.

jerry-core/debugger/debugger-ws.h
knightburton: The message type is in the jerry_debugger_string_data_part_t struct. This solution was in the eval data struct, so i did not changed it.

knightburton:
@zherczeg I've updated the PR based on your requests.

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,81 @@ jerry_debugger_cleanup (void)
...
+    JERRY_CONTEXT (debugger_flags) = (uint8_t) (JERRY_CONTEXT (debugger_flags) | JERRY_DEBUGGER_CLIENT_SOURCE_MODE);
+    jerry_debugger_uint8_data_t *client_source_data_p = NULL;
+
+    while (true)
+    {
+      if (jerry_debugger_receive (&client_source_data_p))
+      {
+        if (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONNECTED))
+        {
+          jmem_heap_free_block (client_source_data_p,
```
zherczeg: This is incorrect. `client_source_data_p` can be NULL. What about simply break here and let the `if` handle the free below.

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,81 @@ jerry_debugger_cleanup (void)
...
+                                client_source_data_p->string_size + sizeof (jerry_debugger_uint8_data_t));
+          return false;
+        }
+
+        /* The source arrived. */
+        if (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CLIENT_SOURCE_MODE))
+        {
+          jerry_char_t *string_p = (jerry_char_t *) (client_source_data_p + 1);
```
zherczeg: Add an ASSERT that client_source_data_p is not NULL.

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,81 @@ jerry_debugger_cleanup (void)
...
+
+    if (client_source_data_p != NULL)
+    {
+      /* The data may partly arrived. */
+      jmem_heap_free_block (client_source_data_p,
+                            client_source_data_p->string_size + sizeof (jerry_debugger_uint8_data_t));
+    }
+  }
+
+  return true;
```
zherczeg: I would put the return `false` case here, since that is more frequent.

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,81 @@ jerry_debugger_cleanup (void)
...
+          }
+
+          if (jerry_value_has_error_flag (ret_value))
+          {
+            jmem_heap_free_block (client_source_data_p,
+                                  client_source_data_p->string_size + sizeof (jerry_debugger_uint8_data_t));
+            return false;
+          }
+
+          jerry_release_value (ret_value);
```
zherczeg: It would not be bad to pass this back somehow.

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,81 @@ jerry_debugger_cleanup (void)
...
+                                                                name_size,
+                                                                (string_p + name_size + 1),
+                                                                (client_source_data_p->string_size - name_size - 1),
+                                                                false);
+
+          if (!jerry_value_has_error_flag (ret_value))
+          {
+            jerry_value_t func_val = ret_value;
+            ret_value = jerry_run (func_val);
+            jerry_release_value (func_val);
```
zherczeg: This is a question: I would consider returning with `true` even if `jerry_run` returns with an error, since the `the client source has been processed`. Maybe even if a parse error occures, since the source is successfully received, just it is wrong.

jerry-core/debugger/debugger-ws.c
```diff
@@ -433,16 +440,18 @@ JERRY_STATIC_ASSERT (JERRY_DEBUGGER_MAX_RECEIVE_SIZE < 126,
  *         false - otherwise
  */
 bool
-jerry_debugger_receive (void)
+jerry_debugger_receive (jerry_debugger_uint8_data_t **message_data_p) /**< custom message data */
```
zherczeg: `/**< [out] data received from client */`

jerry-core/debugger/debugger-ws.c
```diff
@@ -433,16 +440,18 @@ JERRY_STATIC_ASSERT (JERRY_DEBUGGER_MAX_RECEIVE_SIZE < 126,
...
 {
   JERRY_ASSERT (JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONNECTED);
 
+  JERRY_ASSERT (message_data_p != NULL
+                || (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_RECEIVE_DATA_MODE)));
```
zherczeg: This is also a question: there is no point passing a `message_data_p != NULL` when we are not in `JERRY_DEBUGGER_RECEIVE_DATA_MODE` mode since we don't process any messages which receives data.

What about:

```
JERRY_ASSERT (message_data_p != NULL ? (JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_RECEIVE_DATA_MODE))
                                     : !(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_RECEIVE_DATA_MODE)));
```

zherczeg: Or you can add two asserts if line is too long.

jerry-core/debugger/debugger-ws.h
```diff
@@ -64,6 +64,23 @@ typedef struct
...
 /**
+ * Incoming message: next message of string data.
+ */
+typedef struct
+{
+  uint8_t type; /**< type of the message */
+} jerry_debugger_receive_uint8_data_part_t;
+
+/**
+ * String data for eval and client source.
```
zherczeg: Byte data for evaluating expressions and receiving client source.

(Note: in case of client source we receive two strings, so String data is not even correct now).

jerry-core/debugger/debugger.c
```diff
@@ -270,53 +263,65 @@ jerry_debugger_process_message (uint8_t *recv_buffer_p, /**< pointer the the rec
 
   if (*expected_message_type_p != 0)
   {
-    JERRY_ASSERT (*expected_message_type_p == JERRY_DEBUGGER_EVAL_PART);
+    JERRY_ASSERT (*expected_message_type_p == JERRY_DEBUGGER_EVAL_PART
+                  || *expected_message_type_p == JERRY_DEBUGGER_CLIENT_SOURCE_PART);
 
-    jerry_debugger_eval_data_t *eval_data_p = (jerry_debugger_eval_data_t *) *message_data_p;
+    jerry_debugger_uint8_data_t *string_data_p = (jerry_debugger_uint8_data_t *) *message_data_p;
```
zherczeg: `uint8_data_p`

jerry-core/debugger/debugger.c
```diff
@@ -504,6 +509,64 @@ jerry_debugger_process_message (uint8_t *recv_buffer_p, /**< pointer the the rec
       return true;
     }
 
+    case JERRY_DEBUGGER_CLIENT_SOURCE:
+    {
+      if (message_size < sizeof (jerry_debugger_receive_client_source_first_t) + 1)
```
zherczeg: `<=` ? No need for +1.

jerry-main/main-unix.c
```diff
@@ -707,6 +719,17 @@ main (int argc,
     }
   }
 
+  if (is_wait_mode)
+  {
+    is_repl_mode = false;
+#ifdef JERRY_DEBUGGER
+    if (!jerry_debugger_wait_and_run_client_source ())
+    {
+      ret_value = jerry_create_error (JERRY_ERROR_COMMON, (jerry_char_t *) "Source parse and run error!");
```
zherczeg: "Connection aborted before source arrived."

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,82 @@ jerry_debugger_cleanup (void)
...
+
+            return JERRY_DEBUGGER_WAIT_AND_RUN_SUCCESS;
+          }
+          else
+          {
+            jmem_heap_free_block (client_source_data_p,
+                                  client_source_data_p->string_size + sizeof (jerry_debugger_uint8_data_t));
+
+            return JERRY_DEBUGGER_WAIT_AND_RUN_FAIL;
```
zherczeg: This is incorrect according to comment: /**< the source has been received */

jerry-core/include/jerryscript-debugger.h
```diff
@@ -28,12 +28,22 @@ extern "C"
  */
 
 /**
+ * Types for the client source wait and run method.
+ */
+typedef enum
+{
+  JERRY_DEBUGGER_WAIT_AND_RUN_FAIL = 0, /**< error occured while receive */
```
zherczeg: /**< source is not received */
We can return early if we are in breakpoint mode, etc., so this would be clearer.

jerry-core/include/jerryscript-debugger.h
```diff
@@ -28,12 +28,22 @@ extern "C"
...
+} jerry_debugger_wait_and_run_type_t;
+
+/**
  * Engine debugger functions.
  */
 bool jerry_debugger_is_connected (void);
 void jerry_debugger_stop (void);
 void jerry_debugger_continue (void);
 void jerry_debugger_stop_at_breakpoint (bool enable_stop_at_breakpoint);
+jerry_debugger_wait_and_run_type_t jerry_debugger_wait_and_run_client_source (jerry_value_t *return_value);
```
zherczeg: Is there a documentation update for this new API function?

jerry-core/include/jerryscript-debugger.h
```diff
@@ -28,12 +28,22 @@ extern "C"
  */
 
 /**
+ * Types for the client source wait and run method.
+ */
+typedef enum
+{
+  JERRY_DEBUGGER_WAIT_AND_RUN_FAIL = 0, /**< error occured while receive */
+  JERRY_DEBUGGER_WAIT_AND_RUN_SUCCESS = 1, /**< the source has been received */
```
zherczeg: I would prefer JERRY_DEBUGGER_WAIT_AND_RUN_RECEIVED to specify that the source is arrived. It does not mean the source execution is successful.

zherczeg: Come to think of it, I would remove the WAIT_AND_RUN since it extends the string too much. What about:

JERRY_DEBUGGER_SOURCE_RECEIVE_FAILED 
JERRY_DEBUGGER_SOURCE_RECEIVED

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,82 @@ jerry_debugger_cleanup (void)
...
+/**
+ * Sets whether the engine should wait and run a source.
+ *
+ * @return FAIL - if the source is not received or parsed
+ *         SUCCESS - if the client source has been received and processed
+ */
+jerry_debugger_wait_and_run_type_t
+jerry_debugger_wait_and_run_client_source (jerry_value_t *return_value) /**< [out] parse and run return value */
+{
+#ifdef JERRY_DEBUGGER
```
zherczeg: Please move `*return_value = jerry_create_undefined ();` before the ifdef so the `return_value` is always filled with something valid.

jerry-core/debugger/debugger-ws.c
```diff
@@ -433,16 +440,18 @@ JERRY_STATIC_ASSERT (JERRY_DEBUGGER_MAX_RECEIVE_SIZE < 126,
...
 
+  JERRY_ASSERT (message_data_p != NULL ? (JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_RECEIVE_DATA_MODE)
+                                       : !(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_RECEIVE_DATA_MODE));
+
   JERRY_CONTEXT (debugger_message_delay) = JERRY_DEBUGGER_MESSAGE_FREQUENCY;
 
   uint8_t *recv_buffer_p = JERRY_CONTEXT (debugger_receive_buffer);
   bool resume_exec = false;
   uint8_t expected_message_type = 0;
-  void *message_data = NULL;
```
zherczeg: I am happy we don't need a workaround for this local variable in the end. This is very good.

jerry-core/include/jerryscript-debugger.h
knightburton: Yes, this is much better.

jerry-core/include/jerryscript-debugger.h
knightburton: Not yet, i'll update the documentation.

docs/07.DEBUGGER.md
```diff
@@ -194,3 +201,33 @@ jerry_debugger_stop_at_breakpoint (bool enable_stop_at_breakpoint)
...
+jerry_debugger_wait_and_run_type_t
+jerry_debugger_wait_and_run_client_source (jerry_value_t *return_value)
+```
+
+**Example**
+
+```c
+  jerry_init (JERRY_INIT_DEBUGGER);
+
+  jerry_value_t wait_and_run_value = jerry_create_undefined ();
```
zherczeg: No value should be passed to the function. In fact it does not even release the value as far as I remember.

jerry-core/api/jerry-debugger.c
```diff
@@ -116,3 +116,81 @@ jerry_debugger_cleanup (void)
...
+/**
+ * Sets whether the engine should wait and run a source.
+ *
+ * @return FAIL - if the source is not received or parsed
+ *         SUCCESS - if the client source has been received and processed
+ */
+jerry_debugger_wait_and_run_type_t
+jerry_debugger_wait_and_run_client_source (jerry_value_t *return_value) /**< [out] parse and run return value */
+{
+  *return_value = jerry_create_undefined ();
```
zherczeg: Add a newline after the statement.

jerry-core/debugger/debugger-ws.h
```diff
@@ -64,6 +64,23 @@ typedef struct
...
+  uint8_t type; /**< type of the message */
+} jerry_debugger_receive_uint8_data_part_t;
+
+/**
+ * Byte data for evaluating expressions and receiving client source.
+ */
+typedef struct
+{
+  uint32_t string_size; /**< total size of the client source */
+  uint32_t string_offset; /**< current offset in the client source */
```
zherczeg: Please chenge them to uint8_...

knightburton:
@zherczeg I've updated the PR. Can you review it again please?

PR #1938, opened 2017-07-26
ecma_delete_property must recreate the hashmap even if one of the properties is still valid

### LINKED ISSUE #1934: build-in array shift/unshift error
run latest `jerry` with following test case, will encounter `Script Error: TypeError`, with `node` no error occur.

jerry is build by 

```bash
python tools/build.py
```
git header

```
d75710c `replace_str_curr_p` must be inspected before dereferencing (#1926)
ce1d555 Fix regression after #1927
```
test case

```js
 var devNum = 38;
 var devQueue = [];
 for (var i = 0; i < devNum; i++) {
     devQueue.push(i);
 }

 for (var j = 0; j < devNum; j++) {
     print(devQueue[0], devQueue.length);
     devQueue.shift();
     devQueue.unshift(1);
     devQueue.shift();
 }
```

rerobika:
Fixes #1934.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers.c
+++ b/jerry-core/ecma/base/ecma-helpers.c
@@ -821,6 +821,11 @@ ecma_delete_property (ecma_object_t *object_p, /**< object */
         if (cur_prop_p->types[1 - i] != ECMA_PROPERTY_TYPE_DELETED)
         {
           /* The other property is still valid. */
+          if (hashmap_status == ECMA_PROPERTY_HASHMAP_DELETE_RECREATE_HASHMAP)
+          {
+            ecma_property_hashmap_free (object_p);
+            ecma_property_hashmap_create (object_p);
+          }
           return;
         }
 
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/base/ecma-property-hashmap.c
```diff
@@ -507,7 +507,10 @@ ecma_property_hashmap_find (ecma_property_hashmap_t *hashmap_p, /**< hashmap */
 
       ecma_property_t *property_p = property_pair_p->header.types + offset;
 
-      JERRY_ASSERT (ECMA_PROPERTY_IS_NAMED_PROPERTY (*property_p));
+      if (!ECMA_PROPERTY_IS_NAMED_PROPERTY (*property_p))
+      {
+        return NULL;
```
zherczeg: Why do we return here? This aborts the loop.

zherczeg:
I have been thinking about this. It is perfectly valid to store the private pointer in the hash map. It should not cause any issues. So the assert may be incorrect. But if this cause an issue at runtime, there is an issue elsewhere.

zherczeg:
After some discussion: the assert is still correct. The problem is how the special value appears there.

LaszloLango:
I agree with @zherczeg. This is not the proper fix for the issue.

rerobika:
I've found the proper fix for this issue!

zherczeg:
Great work!

PR #1937, opened 2017-07-24
Bugfix: should return error immediately in ecma_op_create_typedarray

jiangzidong:
Related issue: #1936

JerryScript-DCO-1.0-Signed-off-by: Zidong Jiang zidong.jiang@intel.com

Issue #1936, opened 2017-07-21
Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed in ecma_free_string_list


renatahodovan:
###### Jerry version:

```
Checked revision: 1526bcc
Build command: ./tools/build.py --clean --debug --profile=es2015-subset --compile-flag=-m32 --system-allocator=on --jerry-libc=off --error-messages=on
```
###### OS:

```
macOS Sierra (10.12.5)
```
###### Test case:

```javascript
Object.defineProperties(constructor.isExtensible, {a: Object.getOwnPropertyDescriptor(Uint8ClampedArray, "length")})
new Int32Array(new ArrayBuffer(), undefined, Date.now())
```

###### Backtrace:

```
ICE: Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed at jerryscript/jerry-core/ecma/base/ecma-literal-storage.c(ecma_free_string_list):43.
Error: ERR_FAILED_INTERNAL_ASSERTION
bt
* thread #1: tid = 0x19fc2d, 0xa1659496 libsystem_kernel.dylib`__pthread_kill + 10, queue = 'com.apple.main-thread', stop reason = signal SIGABRT
  * frame #0: 0xa1659496 libsystem_kernel.dylib`__pthread_kill + 10
    frame #1: 0xa173fc63 libsystem_pthread.dylib`pthread_kill + 90
    frame #2: 0xa15b9a34 libsystem_c.dylib`abort + 141
    frame #3: 0x000e7df7 jerry`jerry_port_fatal(code=ERR_FAILED_INTERNAL_ASSERTION) + 55 at default-fatal.c:71
    frame #4: 0x00019b9b jerry`jerry_fatal(code=ERR_FAILED_INTERNAL_ASSERTION) + 219 at jrt-fatals.c:58
    frame #5: 0x00007b52 jerry`jerry_assert_fail(assertion="ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)", file="jerryscript/jerry-core/ecma/base/ecma-literal-storage.c", function="ecma_free_string_list", line=43) + 114 at jrt-fatals.c:82
    frame #6: 0x0003a155 jerry`ecma_free_string_list(string_list_p=0x00400cc0) + 421 at ecma-literal-storage.c:43
[...]

PR #1939, opened 2017-07-27
Bugfix: the value in test cases should be invariable

jiangzidong:
Fix the bug imported in #1937 

Reason: the #1936 's regression case use the `Date.now()` as a param of a function, and then cause the assert fail. Then I debug on it and fix the original bug. But the value of Data.now() changes with time changes. Do the patch can pass the CI last time, but failed this time.


jiangzidong:
@akosthekiss please review it.

PR #1951, opened 2017-08-02
add Vagrantfile & docs

boneskull:
This PR adds a `Vagrantfile` for building JerryScript on non-Ubuntu-14.04+ host operating systems, as well as instructions.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,25 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+$script = <<SCRIPT
+sudo apt-get update
+sudo apt-get dist-upgrade -y
+sudo apt-get install gcc gcc-arm-none-eabi cmake cppcheck vera++ python -y
+SCRIPT
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "ubuntu/trusty64"
+  config.vm.synced_folder ".", "/home/vagrant/jerryscript"
+  config.vm.provision "shell", inline: $script
+end
--- a/docs/01.GETTING-STARTED.md
+++ b/docs/01.GETTING-STARTED.md
@@ -2,6 +2,8 @@
 
 Currently, only Ubuntu 14.04+ is officially supported as primary development environment.
 
+[Vagrant can also be used](#building-jerryscript-with-vagrant) to build JerryScript.
+
 There are several dependencies, that should be installed manually. The following list is the absolute minimum for building:
 
 - `gcc` or any C99-compliant compiler (native or cross, e.g., arm-none-eabi)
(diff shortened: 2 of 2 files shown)


boneskull:
I may update this to xenial, as I'm having trouble with some targets on trusty.

akosthekiss:
I have to state upfront that I don't know Vagrant. But from the proposed file it seems that it does not need anything special: it performs an apt update and a dist-upgrade and then installs dependencies of jerryscript development (note: not with the "official" install script but maintains the dep list on its own). So it seems to me that it's just a VM wrapper for non-ubuntu or non-apt-based-linux users. Apparently, anyone using this system (Vagrant) should be fairly familiar with the process. So I'm not convinced that it should be incorporated into the project repository.

boneskull:
If your build process requires a certain OS, then having a `Vagrantfile` reduces the friction for those who *aren't* using that OS to begin to contribute to the project.

zherczeg:
Perhaps we need some testing environment to track future breaks.

boneskull:
@zherczeg I haven't tried doing such a thing before, but it seems straightforward enough.  The cost would be that your build time would likely slow significantly if it was run after every commit.  The image ("box") could be cached on Travis-CI, but it still needs to bring up a machine and provision it (including `apt-get update` and `apt-get dist-upgrade`).

Keep in mind that keeping a `Vagrantfile` around is merely a *convenience* for those users on different operating systems.  It is not required to build JerryScript.  Like @akosthekiss wrote, you can bring up your own Ubuntu VM via other means.  To that end, maintaining a test harness for something like this and running it in CI is probably more trouble than it's worth.

boneskull:
If the maintainer(s) of this project are unfamiliar with Vagrant, I'd encourage you to download & install it (w/ prerequisite VirtualBox) to get a feel for it.  Another way to think of it is like a project-specific "virtualenv", where the prerequisites for building the software won't interfere with your host operating system.  

One last remark is that a `Vagrantfile` belongs in the project; it is project-specific, and this is Vagrant's *intent*.  Storing it elsewhere arguably defeats its purpose--the purpose is to provide a project the means to *easily* provision a development environment.  :wink:

zherczeg:
The idea sounds something like docker. Hm, we probably need some kind of maintenace strategy here. Not just for this patch, but in general. Perhaps opening an issue could be better.

Anyway, it is not bad to support different environments. However the project changes a lot in a year, and unmaintained targets can quickly broke. So in some way we should emphasize that some targets are experimental, and may not work. It is still better having something than nothing. Maybe we could move these targets to an experimental/ sub-directory. We should also think about an eviction strategy, e.g. if a target has no update in 1.5-2 years, we can consider it obsolote and remove it. The key thing is: the rules must be clear for everybody. The rules should be part of the documentation.

martijnthe:
Does Vagrant have a public image repo like Docker has? If a fully provisioned builder image can be downloaded, it shouldn't be too slow to run on every commit?

boneskull:
@martijnthe Yes, a box could be pre-provisioned, but that implies maintaining such a thing.  Maintainers would want to either release it regularly with (security) updates, or provision upon `vagrant up` (the current solution).  

Either way, you still have to bring up a machine.  As it stands, this takes 4m15s on my 2.7 GHz MBP.  Once provisioned, the machine takes 1m18s to start up.  I can only speculate how that'd look in CI.

But, what are we doing when we run the build on Travis-CI?  

1.  Bringing up an Ubuntu VM
2.  Installing the necessary prereqs
3.  Executing build and test scripts

This `Vagrantfile` does 1 and 2.  If the provisioning script (i.e. the `.travis.yml`'s `before_install`) was extracted into a shell script, it could be run directly from the `Vagrantfile`.  Then, no JerryScript maintainers need maintain a `Vagrantfile` they don't themselves use.

LaszloLango:
@zherczeg this is not a new target. I don't see any risk here. The patch looks good to me.

Vagrantfile
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+$script = <<SCRIPT
+sudo apt-get update
+sudo apt-get dist-upgrade -y
+sudo apt-get install gcc gcc-arm-none-eabi cmake cppcheck vera++ python -y
```
akosthekiss: This is a duplication of `tools/apt-get-install-deps.sh`. Why not reuse that? As core developers are not using Vagrant, this list will go easily bitrotten.

martijnthe:
@boneskull In the context of another PR (https://github.com/jerryscript-project/jerryscript/pull/1901) I will be adding a Dockerfile. Could this replace the need for the Vagrant file? Windows, macOS and Linux are supported as host OSes by Docker.

The problem with Vagrant is that Travis CI does not support Vagrant, while it does support Docker.

boneskull:
@martijnthe it's a bit of a higher barrier to entry, but likely a better solution if the maintainers are using it

akosthekiss:
@boneskull, as you seem to accept Docker instead of Vagrant, should we/you close this PR?

boneskull:
yes, it should solve the problem of not being able to compile on macOS.

akosthekiss:
Although this PR just got closed, something caught my eye. @boneskull, could you please elaborate on what do you mean by not being able to compile on macOS? JerryScript supports macOS, I'm developing (compiling, running, etc.) it on macOS myself. Travis CI also tests the macOS builds for every commit. If you really have an issue with that platfrom, it may be worth opening a new issue describing the problem along with the SW (+ HW) environment.

boneskull:
@akosthekiss [the getting started guide](https://github.com/jerryscript-project/jerryscript/blob/master/docs/01.GETTING-STARTED.md)

akosthekiss:
@boneskull you have a point. The docs are definitely lacking information. Until they get amended, please have a look at the .travis.yml file to see how macOS is handled (basically, you need homebrew and the tools/brew-install-deps.sh).

PR #1953, opened 2017-08-04
Fix internal jmem_heap_finalize assertion in debugger.

knightburton:
In case of received client source the client_source_data_p was not freed.


jerry-core/api/jerry-debugger.c
```diff
@@ -134,6 +134,7 @@ jerry_debugger_wait_and_run_client_source (jerry_value_t *return_value) /**< [ou
   {
     JERRY_CONTEXT (debugger_flags) = (uint8_t) (JERRY_CONTEXT (debugger_flags) | JERRY_DEBUGGER_CLIENT_SOURCE_MODE);
     jerry_debugger_uint8_data_t *client_source_data_p = NULL;
+    jerry_debugger_wait_and_run_type_t ret_type;
```
zherczeg: Isn't this an undefined variable if the break below is executed? I would initialize it:

    jerry_debugger_wait_and_run_type_t ret_type = JERRY_DEBUGGER_SOURCE_RECEIVE_FAILED;

knightburton: You are right, I have noticed this a few minutes before. Just a moment and I will update the PR.

yichoi:
LGTM

PR #1954, opened 2017-08-04
Reduce the memory consumption of function objects.

zherczeg:
```
function f(a,b)
{
  return this.x + a + b;
}

var a = [];
var o = { x:5 }

  a[i] = f.bind(o)
```

On this test it reduced the memory consumption from 69K to 45K (35% saving).

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -1337,29 +1337,41 @@ ecma_string_raw_chars (const ecma_string_t *string_p, /**< ecma-string */
 } /* ecma_string_raw_chars */
 
 /**
- * Checks whether ecma string is empty or not
+ * Checks whether the string equals to the magic string id.
  *
- * @return true  - if empty
+ * @return true - if the string equals to the magic string id
  *         false - otherwise
  */
-bool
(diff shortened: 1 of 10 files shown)


zherczeg:
New variant of the patch submitted.

New reduction: 29016 (58% reduction)

The original 56 byte consumption is reduced to 16 byte.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1337,29 +1337,41 @@ ecma_string_raw_chars (const ecma_string_t *string_p, /**< ecma-string */
 } /* ecma_string_raw_chars */
 
 /**
+ * Checks whether the string equals to the magic string id.
+ *
+ * @return true - if the string equals to the magic string id
+ *         false - otherwise
+ */
+inline bool __attr_always_inline___
+ecma_string_is_magic_id (const ecma_string_t *string_p, /**< property name */
```
LaszloLango: I think `ecma_compare_ecma_string_to_magic_id` describes better this function. (based on `ecma_compare_ecma_strings`)

LaszloLango:
Nice improvement. The code looks good expect the mentioned function name.

PR #1959, opened 2017-08-11
Fix new ArrayBuffer(length) for various input

glistening:
Currently new ArrayBuffer(length) does not conform to ES2015.

For example, new ArrayBuffer(length) should not throw RangeError
for length = NaN, undefined, negative number, floating point, and so on.

### DIFF
--- a/tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-006.js
+++ b/tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-006.js
@@ -13,15 +13,5 @@
  * limitations under the License.
  */
 
-var name = "";
-
-try
-{
-  var a = new ArrayBuffer(5.5);
-}
-catch (e)
-{
-  name = e.name;
-}
-
-assert(name === "RangeError");
+var a = new ArrayBuffer(5.5);
+assert(a.byteLength === 5);
--- a/tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-012.js
+++ b/tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-012.js
@@ -13,5 +13,14 @@
  * limitations under the License.
  */
 
-var a = new ArrayBuffer("string");
-assert(a.byteLength === 0);
+var name = "";
+try
+{
+  var a = new ArrayBuffer(-1.9);
+}
+catch (e)
+{
+  name = e.name;
+}
+
+assert(name === "RangeError");
--- a/jerry-core/ecma/operations/ecma-arraybuffer-object.c
+++ b/jerry-core/ecma/operations/ecma-arraybuffer-object.c
@@ -75,15 +75,14 @@ ecma_op_create_arraybuffer_object (const ecma_value_t *arguments_list_p, /**< li
 {
   JERRY_ASSERT (arguments_list_len == 0 || arguments_list_p != NULL);
 
-  uint32_t length = 0;
+  ecma_number_t length_num = 0;
 
   if (arguments_list_len > 0)
   {
-    ecma_number_t num;
 
     if (ecma_is_value_number (arguments_list_p[0]))
     {
-      num = ecma_get_number_from_value (arguments_list_p[0]);
+      length_num = ecma_get_number_from_value (arguments_list_p[0]);
     }
     else
     {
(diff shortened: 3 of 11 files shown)


glistening:
@jiangzidong Could you review this patch?

jiangzidong:
@glistening I dont think ES2015 allow the length could be NaN, undefined, negative number, floating point.

http://www.ecma-international.org/ecma-262/6.0/#sec-arraybuffer-length see step 5
```
If SameValueZero(numberLength, byteLength) is false, throw a RangeError exception.
```

If the length is NaN, undefined, negative number, floating point. they will fail the `samevaluezero` check

But interesting part is V8 supports  NaN, undefined, negative number (-1, 0), floating point. I don't know why.

Correct me if I miss sth in the spec.

glistening:
@jiangzidong Please see `ToLength` in step 3. Please see `ToInteger` also, which is called from `ToLength`. It specifies the handling of NaN, negative number, and so on. My tests check like Test262 [testcases](https://github.com/tc39/test262/blob/master/test/built-ins/ArrayBuffer/toindex-length.js), and major JS runtime (node, jsc, ...) passes the tests.

jiangzidong:
@glistening Yes, I know that ToInteger will turn "negative number, NaN and floating number" into non-negative integer. But then it will fail the step5. 
for example, `var a = new ArrayBuffer(3.4);` 
> 2.  Let numberLength be ToNumber(length).

the `numberLength ` is 3.4
> 3. Let byteLength be ToLength(numberLength).

the `byteLength` is 3

> 5. If SameValueZero(numberLength, byteLength) is false, throw a RangeError exception.

`3 !== 3.4`, so it will throw RangeError .

I don't know why test262/v8/spidermonkey don't follow it.

glistening:
@jiangzidong Thank you for correcting me. The first sentence in this PR was wrong.

Test262 and major engines (v8, spidermonkey and JavaScriptCore) seems to follow [ES2017](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-arraybuffer-constructor), not ES2015. ES2017 uses [ToIndex](https://www.ecma-international.org/ecma-262/8.0/index.html#sec-toindex) to describe the expected behaviour of `ArrayBuffer(length)`. You can see it in test26 test case comment also.

> esid: sec-arraybuffer-length
> description: >
>   The `length` parameter is converted to a value numeric index value.
> info: |
>
>   ArrayBuffer( length )
>   1. If NewTarget is undefined, throw a TypeError exception.
>   2. Let byteLength be ? ToIndex(length).
>   3. Return ? AllocateArrayBuffer(NewTarget, byteLength).
>
>   ToIndex( value )
>
>   1. If value is undefined, then
>     a. Let index be 0.
>   2. Else,
>     a. Let integerIndex be ? ToInteger(value).
>     b. If integerIndex < 0, throw a RangeError exception.
>     c. Let index be ! ToLength(integerIndex).
>     d. If SameValueZero(integerIndex, index) is false, throw a RangeError exception.
>   3. Return index.
> 

I will raise this as Issue to get opinion.

jiangzidong:
@glistening Wow, I didn't notice that it's ES2017. Thanks for pointing out.

zherczeg:
(I wouldn't create subsets for such differences, just use the latest standard)

tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-006.js
```diff
@@ -24,4 +24,5 @@ catch (e)
   name = e.name;
 }
 
-assert(name === "RangeError");
+assert(name === "");
```
jiangzidong: if there is no expection, we dont need to check the name. just
```
var a = new ArrayBuffer(5.5);
assert(a.byteLength === 5);
```

tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-012.js
```diff
@@ -0,0 +1,28 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var name = "";
+
+try
+{
+  var a = new ArrayBuffer(-0.9);
```
jiangzidong: can you also give a case where the length is float and below -1?

tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-006.js
glistening: @jiangzidong I've simplified 24.01.02-006.js ~24.01.02-012.js.

tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-012.js
glistening: @jiangzidong Now, `24.01.02-012.js` covers two more cases that the length is float and below -1.

tests/jerry-test-suite/es2015/24/24.01/24.01.02/24.01.02-012.js
```diff
@@ -0,0 +1,23 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var a = new ArrayBuffer(-0.9);
+assert(a.byteLength === 0);
+
+var b = new ArrayBuffer(-1.9);
+assert(b.byteLength === 0);
```
jiangzidong: `-1.9` and `-2.2` should throw error, because `toIndex(length) < 0 `

glistening: Oh, You're right. I will fix right now.

glistening: @jiangzidong I've updated PR.

PR #1966, opened 2017-08-15
Allow "<NUL>" character within string literals in strict mode

bsdelf:
"\0" is a legal character value with a lookahead restriction according to specification:

> The CV of EscapeSequence :: 0 [lookahead ∉ DecimalDigit] is a &lt;NUL&gt; character (Unicode value 0000).

 Therefore, it should not be treated like octal escape sequences in some circumstances.

Test case:

```javascript
'use strict';
var print = print || console.log;
print('\0');
print('\0a');
print('a\0');
print('\0\0\0');
print('foo\0bar');
```

Reference:
- https://www.ecma-international.org/ecma-262/5.1/#sec-7.8.4
- https://www.ecma-international.org/ecma-262/5.1/#sec-B.1.2
- https://github.com/nodejs/node/blob/98ddab411523f83eca9b5311529f8288379e550f/deps/v8/src/parsing/scanner.cc#L1050
- https://github.com/svaarala/duktape/blob/aa401554460db5c006fdb76715dd66467e5c4c3d/src-input/duk_lexer.c#L785

### DIFF
--- a/tests/jerry/strict.js
+++ b/tests/jerry/strict.js
@@ -70,6 +70,52 @@ try
   assert (e instanceof TypeError);
 }
 
+try
+{
+  eval ("'\\" + "101'");
+
+  assert (false);
+} catch (e)
+{
+  assert (e instanceof SyntaxError);
+}
+
+try
+{
+  var str1 = "'\\" + "0'";
+  var str2 = "'\\x" + "00'";
+  eval (str1);
+
+  assert (eval (str1) === eval (str2));
+} catch (e)
+{
+  assert (false);
+}
+
+try
+{
+  var str1 = "'\\" + "0" + "\\" + "0" + "\\" + "0'";
+  var str2 = "'\\x" + "00" + "\\x" + "00" + "\\x" + "00'";
+  eval (str1);
+
+  assert (eval (str1) === eval (str2));
+} catch (e)
+{
+  assert (false);
+}
+
+try
+{
+  var str1 = "'foo\\" + "0" + "bar'";
+  var str2 = "'foo\\x" + "00" + "bar'";
+  eval (str1);
+
+  assert (eval (str1) === eval (str2));
+} catch (e)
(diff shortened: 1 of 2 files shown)


LaszloLango:
Thanks for the fix. Please add some test for it.

bsdelf:
Patch updated, I added some test for both invalid and valid cases.

tests/jerry/strict.js
```diff
@@ -70,6 +70,46 @@ try
...
+} catch (e)
+{
+  assert (false);
+}
+
+try
+{
+  eval ("'\\" + "0" + "\\" + "0" + "\\" + "0'");
+
+  assert (true);
```
LaszloLango: This statement is useless.

tests/jerry/strict.js
```diff
@@ -70,6 +70,46 @@ try
...
+} catch (e)
+{
+  assert (false);
+}
+
+try
+{
+  eval ("'foo\\" + "0" + "bar'");
+
+  assert (true);
```
LaszloLango: This statement is useless.

tests/jerry/strict.js
```diff
@@ -70,6 +70,46 @@ try
...
+} catch (e)
+{
+  assert (e instanceof SyntaxError);
+}
+
+try
+{
+  eval ("'\\" + "0'");
+
+  assert (true);
```
LaszloLango: This statement is useless.

bsdelf: Ah, I see.

LaszloLango:
I'm wondering if we could add asserts to compare the result and not just exception checking.
E.g.: `eval ("'\\" + "0'") === eval ("'\\x" + "00'")` or something similar. What do you think?

bsdelf:
It would be best.

yichoi:
LGTM

Issue #1972, opened 2017-08-18
Type confusion in release builds while setting array length from not a number (003)


corporateshill:
Only an assert stops the function ecma_get_float_from_value from returning pointers from within an ecma value of the wrong type (it's not a float), which will happen in release builds because the assert is compiled out.

It's likely that this has a security impact.

On Ubuntu 16.04.2,

$ python tools/build.py --compile-flag=-m32 --clean --jerry-libc=OFF --system-allocator ON --compile-flag=-ggdb3 --debug --link-lib mcheck

$ cat ArrayLength.js
[].length={valueOf:function(){return Array.prototype.push(1),Object.freeze(Array.prototype);}}

$ MALLOC_CHECK_=1 gdb -q -x mcheck.gdb --args ./build/bin/jerry ArrayLength.js

<snip>

ICE: Assertion 'ecma_get_value_type_field (value) == ECMA_TYPE_FLOAT' failed at /afl/jerryscript/jerry-core/ecma/base/ecma-helpers-value.c(ecma_get_float_from_value):544.
Error: ERR_FAILED_INTERNAL_ASSERTION
[Inferior 1 (process 9992) exited with code 0170]

(gdb) where
  #0  ecma_get_number_from_value (value=0x82d58e7) at /afl/jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:562
  #1  0x080911ad in ecma_op_array_object_set_length (object_p=0x82d43f0, new_value=0x82d44ab, flags=0x0) at /afl/jerryscript/jerry-core/ecma/operations/ecma-array-object.c:170
  #2  0x0808e1fa in ecma_op_object_put (object_p=0x82d43f0, property_name_p=0x82d4520, value=0x82d44ab, is_throw=0x0)
      at /afl/jerryscript/jerry-core/ecma/operations/ecma-objects.c:707
  #3  0x080858c7 in vm_op_set_value.lto_priv.413 (object=0x82d43f3, property=0x82d4522, value=0x82d44ab, is_strict=0x0) at /afl/jerryscript/jerry-core/vm/vm.c:186
  #4  0x08082bcd in vm_loop.lto_priv.409 (frame_ctx_p=0xffffd3f4) at /afl/jerryscript/jerry-core/vm/vm.c:2492
[...]

PR #1985, opened 2017-08-29
Fix ecma_op_array_object_set_length method

rerobika:
So far a freed variable was tested during error flag inspection.
This patch fixes it and #1972 as well.


rerobika:
@dbatyai As it seems, you are right this condition is unnecessary. I'll update this PR.

dbatyai:
As it turns out, it is needed after all, as it can cause side effects. Please keep the original solution.

yichoi:
LGTM

PR #1988, opened 2017-08-30
target: mbedos5: Use math.h from stdlib, not from jerry-libm

janjongboom:
The math.h header file in JerryScript is lacking a number of functions which are required to build for certain mbed OS 5 targets. NUCLEO_F207ZG target fails to build as sqrtf() is not defined in this header file. Exclude this file from the build process, so the math.h from stdlib is included instead.

### DIFF
--- a/targets/mbedos5/template-mbedignore.txt
+++ b/targets/mbedos5/template-mbedignore.txt
@@ -1,6 +1,7 @@
 cmake/*
 docs/*
 jerry-libc/*
+jerry-libm/*
 jerry-main/*
 jerry-port/default/default-date.c
 jerry-port/default/default-fatal.c


targets/mbedos5/template-mbedignore.txt
```diff
@@ -1,6 +1,7 @@
 cmake/*
 docs/*
 jerry-libc/*
+jerry-libm/include/*
 jerry-main/*
```
akosthekiss: Why not exclude the whole jerry-libm directory and let the libm of the toolchain be used? If the header comes from the toolchaim but the sources come from jerry-libm, nasty confusion can happen. (Not tried myself but that's my gut feeling.)

janjongboom: @akosthekiss Good point, thanks. Have updated the PR and verified that it still builds.

akosthekiss: I don't see the new changes in the PR, yet. Problem on my side?

janjongboom:
@akosthekiss Ah, I missed that you didn't +1 it. Sorry.

PR #1992, opened 2017-09-01
Correctly enumerate function property members.

zherczeg:
Functions has several built-in non-enumerable properties, and
they are correctly ignored during enumeration after this patch.

External function prototype is also lazy enumerated.


yichoi:
LGTM

Issue #1995, opened 2017-09-04
Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed in ecma_free_string_list


renatahodovan:
###### Jerry version:

```
Checked revision: 7525006
Build command: ./tools/build.py --clean --debug --profile=es2015-subset --jerry-libc=off
```
###### OS:

```
Ubuntu 16.04.3 LTS
```
###### Test case:

```javascript
Promise.race([""]).$()
```

###### Backtrace:

```
Script Error: ReferenceError
ICE: Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed at jerryscript/jerry-core/ecma/base/ecma-literal-storage.c(ecma_free_string_list):43.
Error: ERR_FAILED_INTERNAL_ASSERTION
bt
#0  0x00000000004853e8 in raise ()
#1  0x00000000004855da in abort ()
#2  0x0000000000400b5d in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-port/default/default-fatal.c:71
#3  0x00000000004298c9 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#4  0x0000000000429919 in jerry_assert_fail (assertion=0x5267e0 "ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)", file=0x526780 "jerryscript/jerry-core/ecma/base/ecma-literal-storage.c", function=0x522d50 <__func__.3969.lto_priv.241> "ecma_free_string_list", line=43) at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#5  0x000000000040ef3b in ecma_free_string_list (string_list_p=0x96c148 <jerry_global_heap+504>) at jerryscript/jerry-core/ecma/base/ecma-literal-storage.c:43
#6  0x000000000040efbc in ecma_finalize_lit_storage () at jerryscript/jerry-core/ecma/base/ecma-literal-storage.c:60
#7  0x000000000046159b in ecma_finalize () at jerryscript/jerry-core/ecma/base/ecma-init-finalize.c:65
#8  0x0000000000478055 in jerry_cleanup () at jerryscript/jerry-core/api/jerry.c:179
[...]

PR #1998, opened 2017-09-05
[unix-main] call jerry_run_all_enqueued_jobs before cleanup

jiangzidong:
Related Issue: #1995

We should also call jerry_run_all_enqueued_jobs even though the script throws exception.

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -817,17 +817,17 @@ main (int argc,
 
     ret_code = JERRY_STANDALONE_EXIT_CODE_FAIL;
   }
-  else
-  {
-    jerry_release_value (ret_value);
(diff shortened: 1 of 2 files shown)


jerry-main/main-unix.c
```diff
@@ -817,19 +817,19 @@ main (int argc,
...
 
-    if (jerry_value_has_error_flag (ret_value))
-    {
-      print_unhandled_exception (ret_value);
-      ret_code = JERRY_STANDALONE_EXIT_CODE_FAIL;
-    }
-  }
   jerry_release_value (ret_value);
 
+  jerry_value_t job_ret = jerry_run_all_enqueued_jobs ();
```
zherczeg: I think we should reuse ret_value.

Issue #1997, opened 2017-09-04
Assertion 'compressed_pointer != JMEM_CP_NULL' in jmem_decompress_pointer


renatahodovan:
###### Jerry version:

```
Checked revision: 7905422
Build command: ./tools/build.py --clean --debug --profile=es2015-subset
```
###### OS:

```
Ubuntu 16.04.3 LTS
```
###### Test case:

```javascript
var a = Promise.resolve();
for (;;)
    Promise.race([a]).then()
$
```

###### Backtrace:

```
ICE: Assertion 'compressed_pointer != JMEM_CP_NULL' failed at jerryscript/jerry-core/jmem/jmem-allocator.c(jmem_decompress_pointer):96.
Error: ERR_FAILED_INTERNAL_ASSERTION
bt
#0  0x000000000047971f in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:59
#1  0x000000000040489e in raise (sig=6) at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:95
#2  0x0000000000404870 in abort () at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:81
#3  0x0000000000406dc4 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x000000000042aa80 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x000000000042aad0 in jerry_assert_fail (assertion=0x480c18 "compressed_pointer != JMEM_CP_NULL", file=0x480b38 "jerryscript/jerry-core/jmem/jmem-allocator.c", function=0x494750 <__func__.3185.lto_priv.290> "jmem_decompress_pointer", line=96) at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x000000000042a8ce in jmem_decompress_pointer (compressed_pointer=0) at jerryscript/jerry-core/jmem/jmem-allocator.c:96
#7  0x00000000004728d0 in ecma_collection_iterator_next (iterator_p=0x7fffffffc890) at jerryscript/jerry-core/ecma/base/ecma-helpers-values-collection.c:332
[...]

PR #2000, opened 2017-09-05
[ecma-collection] Check whether the header's first chunk is NULL

jiangzidong:
Fix Issue: #1997

the reason of the error in #1997 

1, in `ecma_append_to_values_collection`, the collection may be empty at first, and the header->first_chunk is NULL. then it will alloc the chunk, then it will set header's first_chunk. But during the **alloc**,  the alloc may cause GC .
2. GC will traverse the collection item by `ecma_collection_iterator_next` (the Promise needs it), but the original `ecma_collection_iterator_next` will cause error when the header's first chunk is NULL

So we should check whether header's first chunk is NULL in the `ecma_collection_iterator_next`.

The codes in #1997 is 
```
var a = Promise.resolve();
for (;;)
    Promise.race([a]).then()
$
```
Even I fixed the issue, the code will still cause OOM. 
How can I handle it in the test case?

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-values-collection.c
+++ b/jerry-core/ecma/base/ecma-helpers-values-collection.c
@@ -318,7 +318,8 @@ bool
 ecma_collection_iterator_next (ecma_collection_iterator_t *iterator_p) /**< context of iterator */
 {
   if (iterator_p->header_p == NULL
-      || unlikely (iterator_p->header_p->unit_number == 0))
+      || unlikely (iterator_p->header_p->unit_number == 0)
+      || unlikely (iterator_p->header_p->first_chunk_cp == JMEM_CP_NULL))
   {
     return false;
   }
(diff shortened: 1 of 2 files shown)


zherczeg:
Could you add a test case?

LaszloLango:
Probably a finite for loop could work.

jiangzidong:
@zherczeg @LaszloLango Yes, I found that the 200 loop will reproduce the error in the current master branch. I will add the case will 200 times iteration

LaszloLango:
A bit off topic, but congratulations, you have the 2000th PR/issue of the project. :)

jiangzidong:
Wow, wish the project and the community will be more and more prosperous :)

jiangzidong:
@zherczeg @LaszloLango Hi, could you help review the updated patch again? Thanks

yichoi:
LGTM

PR #2002, opened 2017-09-05
Clean up radix conversion in Number toString method

dbatyai:
The radix conversion code path was very messy which made it hard to understand
what was happening inside of it. The code got cleaned up, and a lot of comments
were added that explain what is happening and why.


dbatyai:
The test case is changed because we gained a significant digit, due to how the required digit count is calculated after the patch.

PR #2007, opened 2017-09-06
Bitfields conversions fix for Nuttx build

pmarcinkiew:
A conversion from bool to bitfield causes "may alter its value" warnings
for some versions of GCC used for MCUs. It prevents enabling Nuttx
headers for JerryScript compilation with strict Werror mode.

ECMA_CONVERT_BOOL_TO_FIELD macro was introduced in ecma-helpers. It
provides conversion from boolean to 1/0 constant, which can be assigned
to bitfield without any warnings. It is very ugly solution for not very
smart compiler:
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39170

I enabled Nuttx headers for JerryScript with werrror active, which caused many warning for ECMA structures:
https://travis-ci.org/Samsung/iotjs/jobs/270446192

This commits provides ugly work around for GCC limitations and IoT.js strict warning policy but it will also benefit any project with similar configuration.

JerryScript-DCO-1.0-Signed-off-by: Piotr Marcinkiewicz p.marcinkiew@samsung.com

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers.h
+++ b/jerry-core/ecma/base/ecma-helpers.h
@@ -109,6 +109,13 @@
     JERRY_ASSERT (utf8_ptr != NULL); \
     jmem_heap_free_block ((void *) utf8_ptr, utf8_str_size); \
   }
+/**
+ * This is very ugly macro but there is not better solution,
+ * which will work for bit fields in versions of GCC used in MCU-s
+ * See:
+ * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39170
+ */
+#define ECMA_CONVERT_BOOL_TO_FIELD(x) ((x) ? 1 : 0)
 
 /* ecma-helpers-value.c */
 bool ecma_is_value_direct (ecma_value_t value) __attr_const___;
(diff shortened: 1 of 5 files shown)


jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -109,6 +109,13 @@
     JERRY_ASSERT (utf8_ptr != NULL); \
     jmem_heap_free_block ((void *) utf8_ptr, utf8_str_size); \
   }
+/**
+ * This is very ugly macro but there is not better solution,
+ * which will work for bit fields in versions of GCC used in MCU-s
+ * See:
+ * https://gcc.gnu.org/bugzilla/show_bug.cgi?id=39170
+ */
+#define ECMA_CONVERT_BOOL_TO_FIELD(x) ((x) ? 1 : 0)
```
LaszloLango: Please make this dependent from the GCC version and declare as empty above the problematic version. Do not criticize your own code in the comment, just write that it is a workaround for a certain GCC version.

zherczeg: I think compilers are able to optimize this, so it si not an ugly macro.

yichoi:
I don't think this is relevant with our situation to meet one of our major milestone, the second internal release. If not, please close then focus with our priorities.

robertsipka:
I've ran into this problem in case of IoT.js build on artik053 with enabled jerry-debugger as well. Are you planning to reopen this pull request?

PR #2009, opened 2017-09-08
Add context reset to the debugger.

knightburton:
- The context reset request message can be sent anytime from a client.
- After the message received the engine will call the cleanup and init
  when in the source waiting mode (which means the currently processed file will be executed).
- After the reinitialization is done, the engine will wait
  for a new client connection(rest of the work is the client's responsibility).


jerry-core/debugger/debugger.c
```diff
@@ -586,6 +586,17 @@ jerry_debugger_process_message (uint8_t *recv_buffer_p, /**< pointer the the rec
       return true;
     }
 
+    case JERRY_DEBUGGER_CONTEXT_RESET:
+    {
+      JERRY_DEBUGGER_CHECK_PACKET_SIZE (jerry_debugger_receive_type_t);
+
+      JERRY_CONTEXT (debugger_flags) = (uint8_t) (JERRY_CONTEXT (debugger_flags) | JERRY_DEBUGGER_CONTEXT_RESET_MODE);
```
zherczeg: please make this case similar to the JERRY_DEBUGGER_NO_MORE_SOURCES above. Check mode and remove JERRY_DEBUGGER_CLIENT_SOURCE_MODE flag.

knightburton:
@zherczeg I've updated the PR based on your request, please review it again.

jerry-core/api/jerry-debugger.c
```diff
@@ -135,6 +136,13 @@ jerry_debugger_wait_and_run_client_source (jerry_value_t *return_value) /**< [ou
           break;
         }
 
+        /* Stop executing the current context. */
+        if ((JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONTEXT_RESET_MODE))
+        {
+          ret_type = JERRY_DEBUGGER_CONTEXT_RESET_RECEIVED;
```
zherczeg: I think these flags should be removed (with binary and) after processed. Otherwise they will always trigger the return. Probably true for JERRY_DEBUGGER_CLIENT_NO_SOURCE.

knightburton: I've added the chnages.

PR #2011, opened 2017-09-12
target: mbedos5: correct #1931

fbmrk:
The mentioned patch uses `this_obj` argument to get `setInterval` and
`setTimeout` properties from the global object, but `this_obj` is undefined in these scopes.


LaszloLango:
@janjongboom could you check this PR?

janjongboom:
@akosthekiss Verified on FRDM-K64F. I'm not sure how I missed it too, I tested it on a board I thought, maybe didn't properly switch...

PR #2012, opened 2017-09-15
Add convenience macros for declaring C strings

gabrielschulhof:
JerryScript-DCO-1.0-Signed-off-by: Gabriel Schulhof <gabriel.schulhof@intel.com>

I found I was writing the steps in these macros over and over so I figured these might be common enough to warrant macros.

### DIFF
--- a/jerry-core/include/jerryscript-core.h
+++ b/jerry-core/include/jerryscript-core.h
@@ -300,7 +300,7 @@ bool jerry_get_boolean_value (const jerry_value_t value);
 double jerry_get_number_value (const jerry_value_t value);
 
 /**
- * Functions for string values.
+ * Functions and macros for string values.
  */
 jerry_size_t jerry_get_string_size (const jerry_value_t value);
 jerry_size_t jerry_get_utf8_string_size (const jerry_value_t value);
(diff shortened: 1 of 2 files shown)


jerry-core/include/jerryscript-core.h
```diff
@@ -321,6 +321,16 @@ jerry_size_t jerry_substring_to_utf8_char_buffer (const jerry_value_t value,
...
                                                   jerry_size_t buffer_size);
 
+#define DECLARE_C_STRING_FROM_JERRY_STRING(prefix, value)               \
+  jerry_size_t prefix##_size = jerry_get_string_size ((value));         \
+  jerry_char_t prefix##_string[prefix##_size];                          \
+  jerry_string_to_char_buffer ((value), prefix##_string, prefix##_size)
+
+#define DECLARE_C_STRING_FROM_UTF8_JERRY_STRING(prefix, value)               \
+  jerry_size_t prefix##_size = jerry_get_string_size ((value));              \
+  jerry_char_t prefix##_string[prefix##_size];                               \
```
martijnthe: Hmm, I'm not sure if it's a good idea to have these.
The stack frames will blow up a lot when the strings are large.
Unless the string size is known to be small, I think it's kind of an anti-pattern, tbh.

zherczeg: I have a similar opinion.

gabrielschulhof:
I suppose with large strings you'd want to use the `_substring_` variants repeatedly and achieve a kind of streaming.

I guess we might even want to use the "streaming" approach when looking at file system paths.

Alright. I don't feel strongly about these macros.

PR #2013, opened 2017-09-16
module extension: add support for canonical name resolution

gabrielschulhof:
Before attempting to load a module, each provided resolver must be given an
opportunity to examine the name of the requested module without actually
loading it so as to canonicalize it, in case a module can be referred to by
multiple names.

Then, modules are loaded and cached by their canonical name.

### DIFF
--- a/docs/12.EXT-REFERENCE-MODULE.md
+++ b/docs/12.EXT-REFERENCE-MODULE.md
@@ -2,17 +2,33 @@
 
 This is a JerryScript extension that provides a means of loading modules. Fundamentally, a module is a name (stored as
 a string) that resolves to a `jerry_value_t`. This extension provides the function `jerryx_module_resolve()` which
-accepts the name of the module being requested as well as an array of so-called "resolvers" - functions which satisfy
-the signature `jerryx_module_resolver_t`. The resolvers in the list are called in sequence until one of them returns
-`true` and a `jerry_value_t` in its out parameter. The value is cached if it is not an error, so subsequent requests
-for the same name will not result in additional calls to the resolvers.
+accepts the name of the module being requested as well as an array of so-called "resolvers" - structures containing two
+function pointers: one for a function which computes a canonical name for the requested module or returns a reference
+to the requested name, and one that converts a canonical name to a `jerry_value_t`, thus "resolving" or "loading" the
+requested module.
+
+The resolvers are first called in sequence to each compute the canonical name of the requested module. This is
+accomplished by calling the `get_canonical_name` function pointer they provide. If the function pointer is `NULL`, the
+requested module name is assumed to be what the resolver considers to be its canonical name. `jerryx_module_resolve`
+searches its cache of loaded modules for each canonical name as returned by a `get_canonical_name` function pointer. If
+one of the loaded modules in the cache corresponds to a canonical name, it is returned.
+
+If no cached module is found, `jerryx_module_resolve` calls each resolver's `resolve` function pointer, passing it its
+previously computed interpretation of the requested module's canonical name. If the resolver successfully creates the
+`jerry_value_t` that represents the loaded module, it returns `true` and the `jerry_value_t` in its out parameter.
+
+When `jerryx_module_resolve` receives a value of `true` from a resolver, it stops iterating over the remaining
+resolvers in the sequence and, if the `jerry_value_t` returned from the resolver's `resolve` does not have the error
+flag set, it will add the `jerry_value_t` to its cache under the module's canonical name and return it. Thus, on
+subsequent calls to `jerryx_module_resolve` with a module name whose canonical name is associated with the
+`jerry_value_t`, no `resolve` callback need be called again.
 
 The purpose of having resolvers is to be able to account for the fact that different types of modules may be structured
 differently and thus, for each type of module a module resolver must be supplied at the point where an instance of that
 type of module is requested.
 
 Additionally, this extension provides a means of easily defining so-called "native" JerryScript modules which can be
-resolved using the JerryScript native module resolver `jerryx_module_native_resolver()`, which can be passed to
+resolved using the native JerryScript module resolver `jerryx_module_native_resolver`, which can be passed to
 `jerryx_module_resolve()`. Native modules are registered during application startup and by calling `dlopen()` by means
 of library constructors, support for which can be turned on using the `FEATURE_INIT_FINI` build flag. In the absence of
 such a flag, the module registration and unregistration functions are exposed as global symbols which can be called
@@ -24,10 +40,17 @@ explicitly.
 
 Load a copy of a module into the current context or return one that was already loaded if it is found.
 
-Each function in `resolvers_p` will be called in sequence until one returns `true` and fills out its out-parameter with
-the `jerry_value_t` representing the requested module. If the `jerry_value_t` does not have the error flag set it will
-be cached. Thus, on subsequent calls with the same value for `name`, none of the functions in `resolvers_p` will be
-called.
+For each resolver passed in via `resolvers_p`, its `get_canonical_name` function pointer gets called in order to
+establish the resolver's interpretation of what the canonical name for the module should be. If `get_canonical_name` is
+`NULL`, it is assumed that the requested module's name as passed in is its canonical name.
+
+Then, for each resolver passed in via `resolvers_p`, its `resolve` function pointer gets called with its interpretation
+of what the module's canonical name should be, as computed in the previous step.
+
+If the resolver's `resolve` function pointer returns `true`, the `jerry_value_t` returned in its out-parameter will be
+returned by `jerryx_module_resolve` as the result of the request. If no error flag is set on the `jerry_value_t` it
+will be cached under its canonical name so as to avoid loading the same module twice in the event of a subsequent call
+to `jerryx_module_resolve` with a module name whose canonical name matches an already loaded module.
 
 **Prototype**
 
@@ -48,56 +71,79 @@ jerryx_module_resolve (const jerry_char_t *name,
 
 **Summary**
 
-The resolver for native JerryScript modules. A pointer to this function can be passed in the second parameter to
-`jerryx_module_resolve` to search for the module among the native JerryScript modules loaded so far.
+The resolver for native JerryScript modules. A pointer to this structure can be passed in the second parameter to
+`jerryx_module_resolve` to search for the module among the native JerryScript modules built into the binary. This
+function is available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is defined.
 
 **Prototype**
 
 ```c
-bool
-jerryx_module_native_resolver (const jerry_char_t *name,
-                               jerry_value_t *result)
+extern jerry_module_resolver_t jerryx_native_module_resolver;
 ```
-- `name` - the name of the module to find
-- `result` - out - place where to store the resulting module instance
-- return value - `true` if the module was found and stored in `result`, and `false` otherwise
-
 
 # Module data types
 
-## jerryx_native_module_on_resolve_t
+## jerryx_module_get_canonical_name_t
 
 **Summary**
 
-Function pointer type for a function that will create an instance of a native module.
+The function pointer type for converting a module's requested name to its canonical name.
 
 **Prototype**
 
 ```c
-typedef jerry_value_t (*jerryx_native_module_on_resolve_t) (void);
+typedef jerry_value_t (*jerryx_module_get_canonical_name_t) (const jerry_value_t name);
+```
+
+## jerryx_module_resolve_t
+
+**Summary**
+
+Function pointer type for module resolution.
+
+**Prototype**
+
+```c
+typedef bool (*jerryx_module_resolve_t) (const jerry_value_t canonical_name,
+                                        jerry_value_t *result);
 ```
 
 ## jerryx_module_resolver_t
 
 **Summary**
 
-Function pointer type for a module resolver
+Structure defining a module resolver.
 
 **Prototype**
 
 ```c
-typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+typedef struct
+{
+  jerryx_module_get_canonical_name_t get_canonical_name_p;
+  jerryx_module_resolve_t resolve_p;
+} jerryx_module_resolver_t;
 ```
+
+- `get_canonical_name_p` - function pointer to be called when the canonical name corresponding to the requested name
+of a module must be established.
+- `resolve_p` - function pointer to be called when a module with the given canonical name needs to be converted to the
+`jerry_value_t` that will become the loaded module.
 
 **Example**
 ```c
-bool
-load_and_evaluate_js_file (const jerry_char_t *name, jerry_value_t *result)
+static bool
+load_and_evaluate_js_file (const jerry_value_t name, jerry_value_t *result)
 {
   bool return_value = false;
   char *js_file_contents = NULL;
   int file_size = 0;
-  FILE *js_file = fopen (name, "r");
+
+  jerry_size_t name_size = jerry_get_utf8_string_size (name);
+  jerry_char_t name_string[name_size + 1];
+  jerry_string_to_utf8_char_buffer (name, name_string, name_size);
+  name_string[name_size] = 0;
+
+  FILE *js_file = fopen (name_string, "r");
 
   if (js_file)
   {
@@ -124,23 +170,44 @@ load_and_evaluate_js_file (const jerry_char_t *name, jerry_value_t *result)
 
   return return_value;
 }
+
+static jerry_value_t
+canonicalize_file_path (const jerry_value_t name)
+{
+  jerry_value_t absolute_path;
+
+  /**
+   * Since a file on the file system can be referred to by multiple relative paths, but only by one absolute path, the
+   * absolute path becomes the canonical name for the module. Thus, to establish this canonical name, we must search
+   * name for "./" and "../", follow symlinks, etc., then create absolute_path via jerry_create_string () and return
+   * it, because it is the canonical name for this module. Thus, we avoid loading the same JavaScript file twice.
+   */
+
+  return absolute_path;
+}
+
+static jerryx_module_resolver_t js_file_loader
+{
+  canonicalize_file_path,
+  load_and_evaluate_js_file
+};
 ```
 
 We can now load JavaScript files:
 ```c
-static const jerryx_module_resolver_t resolvers[] =
+static const jerryx_module_resolver_t *resolvers[] =
 {
   /*
-   * Consult the JerryScript native module resolver first, in case the requested module is a native JerryScript
+   * Consult the resolver for native JerryScript modules first, in case the requested module is a native JerryScript
    * module.
    */
-  jerryx_module_native_resolver,
+  &jerryx_module_native_resolver,
 
   /*
    * If the requested module is not a native JerryScript module, assume it is a JavaScript file on disk and use the
    * above-defined JavaScript file loader to load it.
    */
-  load_and_evaluate_js_file
+  &js_file_loader
 };
 jerry_value_t js_module = jerryx_module_resolve (requested_module, resolvers, 2);
 ```
--- a/jerry-ext/include/jerryscript-ext/module.h
+++ b/jerry-ext/include/jerryscript-ext/module.h
@@ -30,7 +30,7 @@ typedef jerry_value_t (*jerryx_native_module_on_resolve_t) (void);
 typedef struct jerryx_native_module_t
 {
   const jerry_char_t *name_p; /**< name of the module */
-  const jerryx_native_module_on_resolve_t on_resolve; /**< function that returns a new instance of the module */
+  const jerryx_native_module_on_resolve_t on_resolve_p; /**< function that returns a new instance of the module */
   struct jerryx_native_module_t *next_p; /**< pointer to next module in the list */
 } jerryx_native_module_t;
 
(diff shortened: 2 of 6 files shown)


docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -87,23 +104,39 @@ typedef jerry_value_t (*jerryx_native_module_on_resolve_t) (void);
...
 **Prototype**
 
 ```c
-typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+typedef struct
+{
+  jerry_value_t (*get_canonical_name) (const jerry_value_t name);
+  bool (*resolve) (const jerry_value_t canonical_name,
+                   jerry_value_t *result);
+};
```
LaszloLango: `} jerryx_module_resolver_t;`

gabrielschulhof:
@LaszloLango fixed.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -130,20 +163,36 @@ load_and_evaluate_js_file (const jerry_char_t *name, jerry_value_t *result)
 
   return return_value;
 }
+
+static jerry_value_t
+canonicalize_file_path (const jerry_value_t name)
+{
+  jerry_value_t resulting_path;
+
+  /* Search name for "./" and "../", follow symlinks, etc., then create resulting_path via jerry_create_string () */
```
zherczeg: Full stop after sentence.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -130,20 +163,36 @@ load_and_evaluate_js_file (const jerry_char_t *name, jerry_value_t *result)
...
+};
 ```
 
 We can now load JavaScript files:
 ```c
 static const jerryx_module_resolver_t resolvers[] =
 {
   /* Consult the JerryScript module resolver first, in case the requested module is a compiled-in JerryScript module. */
-  jerryx_module_native_resolver,
+  &jerryx_module_native_resolver,
```
zherczeg: Why do you need & ?

jerry-ext/module/module.c
```diff
@@ -137,33 +131,52 @@ jerryx_module_add_to_cache (jerry_value_t cache, /**< cache to which to add the
...
+jerryx_resolve_native_module (const jerry_value_t name, /**< name of the module */
                                jerry_value_t *result) /**< [out] where to put the resulting module instance */
 {
   int index;
   const jerryx_native_module_t *module_p = NULL;
 
+  bool return_value = false;
+
+  jerry_size_t c_module_name_size = jerry_get_utf8_string_size (name);
+  jerry_char_t c_module_name[c_module_name_size];
```
zherczeg: While this allocation is simple in source code level, I always has doubts about its efficiency. You need to move a lot of data on the stack.

jerry-ext/module/module.c
```diff
@@ -180,29 +193,60 @@ jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module
...
+                       const jerry_size_t count) /**< length of list */
 {
   size_t index;
-  jerry_value_t ret;
-  jerry_value_t instances = *(jerry_value_t *) jerry_get_context_data (&jerryx_module_manager);
-  jerry_value_t module_name = jerry_create_string_from_utf8 (name);
+  jerry_value_t ret = 0;
+  const jerryx_module_resolver_t *resolver_p;
+  jerry_value_t instances;
+  jerry_value_t canonical_names_static[5];
```
zherczeg: Lets use a define for this.

jerry-ext/module/module.c
```diff
@@ -180,29 +193,60 @@ jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module
...
+    {
+      goto bail;
+    }
+  }
+
+  instances = *(jerry_value_t *) jerry_get_context_data (&jerryx_module_manager);
+
+  for (index = 0; index < count; index++)
+  {
+    canonical_names[index] = 0;
```
zherczeg: 0? Not undefined?

jerry-ext/module/module.c
```diff
@@ -180,29 +193,60 @@ jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module
...
 
   /* Return the cached instance if present. */
-  if (jerryx_module_check_cache (instances, module_name, &ret))
+  for (index = 0; index < count; index++)
   {
-    goto done;
+    resolver_p = resolvers_p[index];
+    canonical_names[index] =
+    resolver_p->get_canonical_name == NULL ? jerry_acquire_value (name) :
+    resolver_p->get_canonical_name (name);
```
zherczeg: Interesting alignment.

docs/12.EXT-REFERENCE-MODULE.md
gabrielschulhof: Because `jerryx_module_native_resolver` is now a structure stored in a global variable, and this is an array of structure pointers, so I will update the declaration of `resolvers` a few lines up.

jerry-ext/module/module.c
gabrielschulhof: So, should I use `jmem_heap_alloc_block_null_on_error()` instead?

jerry-ext/module/module.c
gabrielschulhof: What do you mean "use a define"?

jerry-ext/module/module.c
gabrielschulhof: If I move the else under `jerry_acquire_value ()` I get this from check-vera.sh:
```
./jerry-ext/module/module.c:235: error: Indentation: 45 -> 4.
```
So, that's why it's like this.

jerry-ext/module/module.c
zherczeg: I am just asking what is its cost. We could introduce convenience stuff for them (I saw you already created some macros for it).

jerry-ext/module/module.c
zherczeg: `#define CANONICAL_NAMES_LOCAL_SIZE 4` or something. I think 4 is enough (probably 2 is the most frequent in practice).

jerry-ext/module/module.c
zherczeg: Use round brackets.

jerry-ext/module/module.c
```diff
@@ -137,33 +131,52 @@ jerryx_module_add_to_cache (jerry_value_t cache, /**< cache to which to add the
...
 
-  return false;
-} /* jerryx_module_native_resolver */
+  return return_value;
+} /* jerryx_resolve_native_module */
+
+jerryx_module_resolver_t jerryx_module_native_resolver =
+{
+  NULL,
+  jerryx_resolve_native_module
```
martijnthe:
```
.get_canonical_name = NULL,
.resolve = jerryx_resolve_native_module,
```

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -22,6 +22,24 @@
...
 #endif /* __GNUC__ */
 
+/**
+ * Declare the structure for module resolvers.
+ */
+typedef struct
+{
+  jerry_value_t (*get_canonical_name) (const jerry_value_t name); /**< module name to canonicalize */
+  bool (*resolve) (const jerry_value_t canonical_name, /**< requested module's canonical name */
+                   jerry_value_t *result); /**< resulting module */
```
martijnthe: What's your rationale for passing around the name strings as `jerry_value_t`s?
I think most module resolvers will convert it right back to a C string.
Is it to avoid having to deal with memory management of the C strings?

gabrielschulhof: Yeah, and I was actually thinking that, for path resolution, you might not even convert the whole value to a C string, but you would rather use fixed size chunks to gradually retrieve the string and maybe assemble the canonical name in a similarly gradual fashion. Then, unless you had a directory with a really, really, really long name, you wouldn't really be allocating all that much.

gabrielschulhof: Call it streamed path resolution.

gabrielschulhof: Another very important reason: the module cache is stored as a JavaScript object, so I have to convert a C string back to a `jerry_value_t` in order to check if that JavaScript object has a module at the module name's key.

martijnthe: Ah I see, yeah that makes sense.

gabrielschulhof:
@martijnthe @LaszloLango @zherczeg I have updated the patch to reflect the changes to the moduling system.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -92,20 +92,27 @@ void jerryx_native_module_register (jerryx_native_module_t *module_p);
...
  */
-typedef bool (*jerryx_module_resolver_t) (const jerry_char_t *name, jerry_value_t *result);
+extern jerryx_module_resolver_t jerryx_module_native_resolver;
 
 /**
  * Load a copy of a module into the current context using the provided module resolvers, or return one that was already
  * loaded if it is found.
  */
-jerry_value_t jerryx_module_resolve (const jerry_char_t *name, const jerryx_module_resolver_t *resolvers, size_t count);
+jerry_value_t jerryx_module_resolve (const jerry_value_t name, /**< module's name */
```
zherczeg: Ditto

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -92,20 +92,27 @@ void jerryx_native_module_register (jerryx_native_module_t *module_p);
...
- * Declare the JerryScript module resolver so that it may be added to an array of jerryx_module_resolver_t items and
- * thus passed to jerryx_module_resolve.
+ * Declare the structure for module resolvers.
  */
-bool jerryx_module_native_resolver (const jerry_char_t *name, jerry_value_t *result);
+typedef struct
+{
+  jerry_value_t (*get_canonical_name) (const jerry_value_t name);  /**< module name to canonicalize */
+  bool (*resolve) (const jerry_value_t canonical_name,  /**< requested module's canonical name */
+                   jerry_value_t *result); /**< resulting module */
```
zherczeg: We do not use comments in header files.

jerry-ext/module/module.c
```diff
@@ -167,26 +167,38 @@ static const jerry_char_t *on_resolve_absent = (jerry_char_t *) "Module on_resol
...
-bool
-jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module */
-                               jerry_value_t *result) /**< [out] where to put the resulting module instance */
+static bool
+jerryx_resolve_native_module (const jerry_value_t canonical_name, /**< canonical name of the module */
+                              jerry_value_t *result) /**< [out] where to put the resulting module instance */
 {
   const jerryx_native_module_t *module_p = NULL;
 
+  jerry_size_t name_size = jerry_get_utf8_string_size (canonical_name);
```
zherczeg: Shall we call tostring on the name? Or check if it is a string.

jerry-ext/module/module.c
```diff
@@ -202,29 +214,51 @@ jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module
  *   - an error indicating that something went wrong during the attempt to load the module.
  */
 jerry_value_t
-jerryx_module_resolve (const jerry_char_t *name, /**< name of the module to load */
-                       const jerryx_module_resolver_t *resolvers_p, /**< list of resolvers */
+jerryx_module_resolve (const jerry_value_t name, /**< name of the module to load */
+                       const jerryx_module_resolver_t **resolvers_p, /**< list of resolvers */
```
zherczeg: `**` ? A pointer to a list of pointers which points to the resolvers? Why not an array of resolvers?

jerry-ext/module/module.c
```diff
@@ -202,29 +214,51 @@ jerryx_module_native_resolver (const jerry_char_t *name, /**< name of the module
...
+  /**
+   * Establish the canonical name for the requested module. Each resolver presents its own canonical name. If one of
+   * the canonical names matches a cached module, it is returned as the result.
+   */
+  for (index = 0; index < resolver_count; index++)
+  {
+    resolver_p = resolvers_p[index];
+    canonical_names[index] = ((resolver_p && resolver_p->get_canonical_name != NULL) ?
+      resolver_p->get_canonical_name (name):
+      jerry_acquire_value (name));
```
zherczeg: This is not how we align ?:

tests/unit-ext/module/jerry-module-test.c
```diff
@@ -61,41 +61,61 @@ const char eval_string5[] =
 "}) ();";
 
 static bool
-resolve_differently_handled_module (const jerry_char_t *name,
+resolve_differently_handled_module (const jerry_value_t name,
                                                 jerry_value_t *result)
```
zherczeg: Comment is missing, spaces are wrong.

jerry-ext/module/module.c
gabrielschulhof: We shouldn't need to. This function should not be called directly, even though a pointer to it is available. It should only be called by `jerryx_module_resolve ()` and it should only ever receive a `jerry_value_t` stored in the array of canonical module names created by `jerryx_module_resolve ()`. In fact, this function is not even documented. Only the structure `jerryx_module_native_resolver` is documented, and only the availability of it as a global data symbol is documented, not its contents.

jerry-ext/module/module.c
gabrielschulhof: Because the resolvers are stored in-contiguously, and we do not want to make copies of them, right?

jerry-ext/module/module.c
gabrielschulhof: It's pretty tough not breaking 120 columns with this line. I'm still trying to figure out how to align it.

gabrielschulhof: I renamed `resolver_p` to `res_p` then it fits.

gabrielschulhof:
@zherczeg I have addressed your review comments.

jerry-ext/module/module.c
zherczeg: Are they used multiple times? Or why they are scattered?

I think one application use one resolver list, but this is not much of a code size increase, so lleave it be.

jerry-ext/module/module.c
zherczeg: I would move the condition into a bool value rather than shoretning the name.

jerry-ext/module/module.c
gabrielschulhof: Resolvers may be declared global in separate .c files (jerryx_module_native_resolver` certainly is), and they are collected into an array only once - the place that calls `jerryx_module_resolve ()`.

gabrielschulhof:
@zherczeg instead of a boolean I went straight for storing the function pointers in variables.

gabrielschulhof:
... aaand I forgot to key the cache on the canonical name. Fixed now.

gabrielschulhof:
@zherczeg could you or somebody with Travis access please restart the lint job? It seems to have been killed because it had stalled.

gabrielschulhof:
Thanks!

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -29,9 +29,9 @@ typedef jerry_value_t (*jerryx_native_module_on_resolve_t) (void);
  */
 typedef struct jerryx_native_module_t
 {
-  const jerry_char_t *name_p; /**< name of the module */
-  const jerryx_native_module_on_resolve_t on_resolve; /**< function that returns a new instance of the module */
-  struct jerryx_native_module_t *next_p; /**< pointer to next module in the list */
+  const jerry_char_t *name_p;
+  const jerryx_native_module_on_resolve_t on_resolve;
+  struct jerryx_native_module_t *next_p;
```
LaszloLango: Why did you remove the comments from here?

jerry-ext/module/module.c
```diff
@@ -14,11 +14,13 @@
  */
 
 #include <string.h>
+#include "jmem.h"
```
LaszloLango: I think we should not use internal headers in `jerry-ext`. I don't see what are you using from it. Please remove it if possible.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -92,20 +92,27 @@ void jerryx_native_module_register (jerryx_native_module_t *module_p);
...
 
 /**
- * Declare the JerryScript module resolver so that it may be added to an array of jerryx_module_resolver_t items and
- * thus passed to jerryx_module_resolve.
+ * Declare the structure for module resolvers.
  */
-bool jerryx_module_native_resolver (const jerry_char_t *name, jerry_value_t *result);
+typedef struct
+{
+  jerry_value_t (*get_canonical_name) (const jerry_value_t name);
```
LaszloLango: Missing comments

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: @zherczeg [said](https://github.com/jerryscript-project/jerryscript/pull/2013#discussion_r140710783) we didn't add such `/**<` comments to the header files. Or did I misinterpret?

jerry-ext/include/jerryscript-ext/module.h
gabrielschulhof: I need to know for sure. Do we, or do we not add `/**<` comments after structure members and function arguments? How do we document function pointer structure members, since the formal parameters of the function signature also needs to be documented?

jerry-ext/include/jerryscript-ext/module.h
LaszloLango: Yes, I think there was a misunderstanding here. We do not add comments to declarations, but only to definitions. The type definitions are usually in the header files, so we must put the comments there. Please don't remove these comments.

zherczeg: Sorry, I mean only for function arguments, not for types.

gabrielschulhof:
@LaszloLango fixed up the comments and added some `_p`s.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -92,20 +92,41 @@ void jerryx_native_module_register (jerryx_native_module_t *module_p);
 void jerryx_native_module_unregister (jerryx_native_module_t *module_p);
 
 /**
- * Declare the JerryScript module resolver so that it may be added to an array of jerryx_module_resolver_t items and
- * thus passed to jerryx_module_resolve.
+ * Declare the function pointer type for canonical name resolution.
+ */
+typedef jerry_value_t (*jerryx_module_get_canonical_name_t) (const jerry_value_t name); /**< The name for which to
+                                                                                         * compute the canonical
+                                                                                         * name */
```
zherczeg: Align the start of the comment lines to the same column.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -92,20 +92,41 @@ void jerryx_native_module_register (jerryx_native_module_t *module_p);
...
+
+/**
+ * Declare the function pointer type for module resolution.
  */
-bool jerryx_module_native_resolver (const jerry_char_t *name, jerry_value_t *result);
+typedef bool (*jerryx_module_resolve_t) (const jerry_value_t canonical_name, /**< The module's canonical name */
+                                         jerry_value_t *result); /**< The resulting module, if the function returns
+                                                                  * true */
```
zherczeg: Ditto. And similar to below.

tests/unit-ext/test-ext-module-canonical.c
```diff
@@ -0,0 +1,93 @@
...
+get_canonical_name (const jerry_value_t name)
+{
+  jerry_size_t name_size = jerry_get_string_size (name);
+  jerry_char_t name_string[name_size + 1];
+  jerry_string_to_char_buffer (name, name_string, name_size);
+  name_string[name_size] = 0;
+
+  return (!strcmp ((char *) name_string, "batman") ? jerry_acquire_value (name):
+         (!strcmp ((char *) name_string, "bruce-wayne") ? jerry_create_string ((jerry_char_t *) "batman"):
+         jerry_create_undefined ()));
```
zherczeg: Fix the style of these lines.

tests/unit-ext/test-ext-module-canonical.c
```diff
@@ -0,0 +1,93 @@
...
+{
+  (void) argc;
+  (void) argv;
+
+  const jerryx_module_resolver_t *resolver = &canonical_test;
+
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t batman = jerry_create_string ((jerry_char_t *) "batman");
+  jerry_value_t bruce_wayne = jerry_create_string ((jerry_char_t *) "bruce-wayne");
```
zherczeg: I hope these are not copyrighted.

gabrielschulhof: Good point. I'll use [Liberator](http://pdsh.wikia.com/wiki/Liberator), which is public domain.

zherczeg: They are still, however, "use at your own risk" characters. - this doesn't sound entirely public

gabrielschulhof: *sigh* Alice and Bob it is.

gabrielschulhof:
@zherczeg @LaszloLango please have another look. I updated the test to use `#define`-ed strings to store the alias and the actual name at the top, and I replaced the ternary composition with `if`/`else if`/`else`. I also changed the resolver initializer to use the `{ .member_name = value }` notation.

LaszloLango:
still LGTM

zherczeg:
Still ok.

Issue #1952, opened 2017-08-03
Remove stringification of JERRYX_NATIVE_MODULE module_name argument?


martijnthe:
See https://github.com/jerryscript-project/jerryscript/blob/master/jerry-ext/include/jerryscript-ext/module.h#L51

`module_name` is stringified, which will cause the argument to not be evaluated by the preprocessor but used verbatim.

After having used the macro in practice, I found that it's not desirable to stringify it, because:

- it prevents one from creating macros that call the `JERRYX_NATIVE_MODULE` macro and pass in a variable module name

- it makes it impossible to single-source the module name, i.e. I cannot write something like:
```
#define MY_MODULE_NAME my_module
#define MY_MODULE_NAME_STR STRINGIFY(MY_MODULE_NAME)

void
foo (void)
{
  const jerry_value_t m = call_require(MY_MODULE_NAME_STR);
  // ...
}

// module will actually be named "MY_MODULE_NAME"...
JERRYX_NATIVE_MODULE(MY_MODULE_NAME, ...)   
```

Therefore I propose to remove the stringification (`#`) from the macro.
I realize it's a breaking change, but the API is pretty new and the migration path is super simple.

Thoughts @gabrielschulhof ?

cc @jiangzidong

jiangzidong:
@martijnthe I agree with you. And I think we can "break" the API.

gabrielschulhof:
@martijnthe @jiangzidong I'm working on this to help out Tim Harvey. I'll have something in a bit.

zherczeg:
Ok for me.

gabrielschulhof:
@martijnthe I found a [solution](https://gcc.gnu.org/onlinedocs/cpp/Stringizing.html#Stringizing) which will allow both the use of `#` in the macro implementation and passing in a macro that evaluates to a non-quoted string.

Basically, if we do
```c
#define JERRYX_NATIVE_MODULE(module_name, on_resolve_cb) \
  JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb)

#define JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb) \
/* actual implementation of macro */
```
Then we can use it like so:
```c
#define MY_MODULE_NAME my_module_name

static jerry_value_t Init (void) {return jerry_create_number (42);}

JERRYX_NATIVE_MODULE(MY_MODULE_NAME, Init)
```

The key is to have a macro call a macro.

[...]

PR #2014, opened 2017-09-18
module: Expose more of the native module implementation

gabrielschulhof:
Only the declarations related to the linker section need be behind the
`JERRYX_NATIVE_MODULES_SUPPORTED` flag. When the flag is absent or `0`,
it is now possible to pass an array of `jerryx_native_module_t`
structures to a new macro `JERRYX_NATIVE_MODULES_INIT()` which will
initialize the global variables `__start_jerryx_modules` and
`__stop_jerryx_modules`, just as the linker would if it were a gcc
linker.

Fixes https://github.com/jerryscript-project/jerryscript/issues/1952

### DIFF
--- a/docs/12.EXT-REFERENCE-MODULE.md
+++ b/docs/12.EXT-REFERENCE-MODULE.md
@@ -13,14 +13,15 @@ type of module is requested.
 
 Additionally, this extension provides a means of easily defining so-called "native" JerryScript modules which can be
 resolved using the JerryScript native module resolver `jerryx_module_native_resolver()`, which can be passed to
-`jerryx_module_resolve()`. Note, however, that native JerryScript modules are only supported and
-`jerryx_module_native_resolver()` is only compiled in if compiler support for `__attribute__` extensions is present. In
-effect this means that native JerryScript modules are available only when this extension is built with GCC or
-LLVM/clang. In the absence of such support, you may construct alternative module systems and provide your own resolver
-to `jerryx_module_resolve()`.
+`jerryx_module_resolve()`. Note, however, that automatic loading of native JerryScript modules is only supported if
+compiler support for `__attribute__` extensions is present. In effect this means that automatic loading for native
+JerryScript modules is available only when this extension is built with GCC or LLVM/clang. In the absence of such
+support, you may manually initialize the list of native modules recognized by the native resolver by passing an array
+of `jerryx_native_module_t` structures to the macro `JERRYX_NATIVE_MODULES_INIT()` preferably during execution of
+`main()`. Note that it is not necessary to call `jerry_init()` before calling the macro.
 
-`jerryscript-ext/module.h` defines the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` only if support for
-native JerryScript modules is available.
+`jerryscript-ext/module.h` defines the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` and sets it to `1` only
+if compiler support for automatic loading of native JerryScript modules is available.
 
 ## jerryx_module_resolve
 
@@ -53,8 +54,12 @@ jerryx_module_resolve (const jerry_char_t *name,
 **Summary**
 
 The resolver for JerryScript modules. A pointer to this function can be passed in the second parameter to
-`jerryx_module_resolve` to search for the module among the JerryScript modules built into the binary. This function is
-available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is defined.
+`jerryx_module_resolve()` to search for the module among the JerryScript modules built into the binary.
+
+**Note**: If compiler support for automatic loading of native modules is not available, indicated by the fact that the
+preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is either absent or set to `0`, you must define an array of
+`jerryx_native_module_t` structures and pass it to the macro `JERRYX_NATIVE_MODULES_INIT()` before calling
+`jerryx_module_resolve()` with `jerryx_module_native_resolver()`.
 
 **Prototype**
 
@@ -74,8 +79,7 @@ jerryx_module_native_resolver (const jerry_char_t *name,
 
 **Summary**
 
-Function pointer type for a function that will create an instance of a native module. This type is only defined if the
-preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is defined.
+Function pointer type for a function that will create an instance of a native module.
 
 **Prototype**
 
@@ -150,13 +154,31 @@ jerry_value_t js_module = jerryx_module_resolve (requested_module, resolvers, 2)
 
 # Module helper macros
 
+## JERRYX_NATIVE_MODULE_INIT
+
+**Summary**
+
+Expands to the initializer of a variable of type `jerryx_native_module_t`.
+
+**Prototype**
+
+```c
+#define JERRYX_NATIVE_MODULE_INIT(module_name, on_resolve_cb)
+```
+
+- `module_name` - literal string giving the module's name
+- `on_resolve_cb` - the function of type `jerryx_native_module_on_resolve_t` that will be called when the module needs
+to be loaded.
+
 ## JERRYX_NATIVE_MODULE
 
 **Summary**
 
 Helper macro to define a JerryScript module. Currently stores the name of the module and its initializer in an
-executable linker section. This macro is available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED`
-is defined.
+executable linker section.
+
+**Note**: This macro is available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED`
+is defined and set to 1.
 
 **Note**: The helper macro must appear at the bottom of a source file, and no semicolon must follow it.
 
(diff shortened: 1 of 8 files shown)


jiangzidong:
@gabrielschulhof in the emx patch #1901 , we met this issue, and did a similar (not same) way to solve it. 
Emx cannot simulate the linker vars like `__start_jerryx_modules` and `__stop_jerryx_modules`, but it can simulate the gcc constructor attribute. So for those "platforms" who dont support link __start/strop_xxx, we use the constructor attribute to register the native modules from different source files. In this way, we don't need to register them explicitly in centralization. 

You can see the https://github.com/jerryscript-project/jerryscript/blob/569a0d95fe1811bea68a9cef94e38c8adaf16f8f/jerry-ext/module/module.c#L74 and https://github.com/jerryscript-project/jerryscript/blob/569a0d95fe1811bea68a9cef94e38c8adaf16f8f/jerry-ext/include/jerryscript-ext/module.h#L48
Any review comments are appreciated!

And for those platforms who dont even support the constructor attr, seems we had to register module explicitly in centralization like your patch.

gabrielschulhof:
I don't really want to add the second fallback of using a constructor, because that requires features from both the compiler and the C library.

zherczeg:
I am ok with the constructor system as long as it does not increase the complexity too much. These are exclusive mechanisms. Anyway the constructor support can be added later.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -13,14 +13,15 @@ type of module is requested.
...
+compiler support for `__attribute__` extensions is present. In effect this means that automatic loading for native
+JerryScript modules is available only when this extension is built with GCC or LLVM/clang. In the absence of such
+support, you may manually initialize the list of native modules recognized by the native resolver by passing an array
+of `jerryx_native_module_t` structures to the macro `JERRYX_NATIVE_MODULES_INIT()` preferably during execution of
+`main()`. Note that it is not necessary to call `jerry_init()` before calling the macro.
 
-`jerryscript-ext/module.h` defines the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` only if support for
-native JerryScript modules is available.
+`jerryscript-ext/module.h` defines the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` and sets it to `1` only
+if compiler support for automatic loading of native JerryScript modules is available.
```
zherczeg: We may explicitly define JERRYX_NATIVE_MODULES_SUPPORTED as non-zero when available since the zero check determines that it is not available.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -53,8 +54,12 @@ jerryx_module_resolve (const jerry_char_t *name,
 **Summary**
 
 The resolver for JerryScript modules. A pointer to this function can be passed in the second parameter to
-`jerryx_module_resolve` to search for the module among the JerryScript modules built into the binary. This function is
-available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is defined.
+`jerryx_module_resolve()` to search for the module among the JerryScript modules built into the binary.
+
+**Note**: If compiler support for automatic loading of native modules is not available, indicated by the fact that the
+preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED` is either absent or set to `0`, you must define an array of
```
zherczeg: Here the zero check explicitly mentioned.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -150,13 +154,31 @@ jerry_value_t js_module = jerryx_module_resolve (requested_module, resolvers, 2)
...
+## JERRYX_NATIVE_MODULE_INIT
+
+**Summary**
+
+Expands to the initializer of a variable of type `jerryx_native_module_t`.
+
+**Prototype**
+
+```c
+#define JERRYX_NATIVE_MODULE_INIT(module_name, on_resolve_cb)
```
zherczeg: I don't really like that we will have a JERRYX_NATIVE_MODULE_INIT and JERRYX_NATIVE_MODULE macros, which both inits something in different scenarios. The name of both should be more explicit about what they actually do.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -41,25 +40,56 @@ typedef struct
...
+ */
+#define JERRYX_NATIVE_MODULE (module_name, on_resolve_cb)
+
+/**
+ * Declare a macro that will allow users to initialize the global variables from an array of jerryx_native_module_t
+ * structures.
+ */
+#define JERRYX_NATIVE_MODULES_INIT(modules)                                      \
+  __start_jerryx_modules = (modules);                                            \
+  __stop_jerryx_modules = &(modules)[sizeof ((modules)) / sizeof ((modules)[0])]
```
zherczeg: I would prefer to have an explicit size. sizeof(modules) might be unavailable if the init is done by a function.

gabrielschulhof:
I can use `__GNUC__` to detect whether linker sections are supported or not, but how do I detect that constructors are supported, but linker sections are not?

docs/12.EXT-REFERENCE-MODULE.md
gabrielschulhof: I actually want it to be conditionally defined in the header file, because I want the build system to be able to override it so, for example, I can define it and set it to zero so I can tests the implementation without linker sections even on gcc, where it would normally be defined and set to 1.

docs/12.EXT-REFERENCE-MODULE.md
gabrielschulhof: I guess IRL it would not be defined. Only for the purposes of testing will it ever be defined but zero. So, should I leave "absent" and remove "set to &#96;0&#96;"?

zherczeg:
Probably the same was as linker sections: based on people feedback we can add defines for those environments where it is supported.

docs/12.EXT-REFERENCE-MODULE.md
zherczeg: Probably yes. Although I like `#if defined M && M` In JerryScript we just test the existence of macros. Hence their value (1, 0, whatever) is irrelevant. Perhaps just remove all explicit number mentionings.

gabrielschulhof: OK, you're right. Maybe we should go one step further and not publish such feature macros at all. The only thing we need to document is that "if you know your compiler doesn't support linker section notation, (and if you know your compiler/C library doesn't support constructors), then call `JERRYX_NATIVE_MODULES_INIT()`" (still to be properly named).

gabrielschulhof:
That's just it though. I know about `__GNUC__`, but I don't yet know about `__HAS_INIT_FINI__` or whatever its name. I think we should add support for constructors later and separately from this PR, because it's too both expose the native resolver and add support for a third (after section and explicit) loading mechanism.

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -150,13 +154,31 @@ jerry_value_t js_module = jerryx_module_resolve (requested_module, resolvers, 2)
...
+to be loaded.
+
 ## JERRYX_NATIVE_MODULE
 
 **Summary**
 
 Helper macro to define a JerryScript module. Currently stores the name of the module and its initializer in an
-executable linker section. This macro is available only if the preprocessor directive `JERRYX_NATIVE_MODULES_SUPPORTED`
-is defined.
+executable linker section.
```
glistening: @gabrielschulhof I have a question because we may use jerryx module on linux based platform. If we run jerryscript on platform that support dynamically library loading (e.g., `dlopen`), `JERRYX_NATIVE_MODULE` macro may not store any information in an executable linker section, and it will do nothing. Do you assume the jerryx module will work on statically linked system only? Could you please correct me if I don't understand correctly?

gabrielschulhof: @glistening if you're using jerryscript in an environment that supports `dlopen()` then we should definitely add support for constructor-based loading.

docs/12.EXT-REFERENCE-MODULE.md
zherczeg: Not necessary. If somebody creates a generic library, he may want to support multiple module representations.

docs/12.EXT-REFERENCE-MODULE.md
zherczeg: I feel this is getting out of hand. Perhaps we should move a step back and re-think what we want to do. Here is my thoughts:

a) The module resolver system with the canonical name support is good. It definitely covers the typical use case.
b) The native system should focus only the static modules. No dynamic loading.
c) We should provide a system for dynamic modules. Question: why do we need constructors? When the module is loaded, a startup routine should run. That startup can be defined by name (e.g.: module name underscore start(), or simply start() since dlsym has a .so argument).

gabrielschulhof: @zherczeg using constructors unifies the process of dynamic loading with that of static loading, and, like linker sections, makes it possible to avoid having to maintain an array with all possible modules.

I'm starting to think that perhaps we should support only constructors, and we should drop support for linker sections altogether, because if the C library or the compiler does not support constructors/destructors, then those functions can be called manually, since at that point the use of a global array is unavoidable.

gabrielschulhof:
I think we're gonna go with #2018 on this one.

PR #2015, opened 2017-09-18
Parse functions directly

zherczeg:
This patch adds direct function source code parsing, which is useful to avoid source code duplications. The patch also improves the Function constructor.


jerry-core/parser/js/js-parser.c
```diff
@@ -1966,15 +1966,120 @@ parser_free_literals (parser_list_t *literal_pool_p) /**< literals */
 } /* parser_free_literals */
 
 /**
+ * Parse function arguments
+ */
+static void
+parser_parse_function_arguments (parser_context_t *context_p,
+                                 lexer_token_type_t end_type) /**< context */
```
LaszloLango: The comments of the arguments should be fixed.

zherczeg:
Patch updated, thank you for the review.

PR #2016, opened 2017-09-18
Fixes uint16_t overflow in ecma_new_ecma_string_from_utf8_converted_to_cesu8 ()

rerobika:
This unreported test issue revealed an assertion in jmem_heap_finalize ().
During the conversion the lot of additional information what a cesu8 represented string needs caused overflow while setting the new ecma_string_t variable's length. This patch fixes this issue and the mentioned test case is available here:
https://gist.github.com/rerobika/3bd590fdcf664a3fcfcc98f11b14c74e

### DIFF
--- a/tests/unit-core/test-strings.c
+++ b/tests/unit-core/test-strings.c
@@ -26,7 +26,7 @@
 #define test_subiters (128)
 
 /* Max bytes in string. */
-#define max_bytes_in_string (16 * 1024)
+#define max_bytes_in_string (65 * 1024)
 #define max_code_units_in_string (max_bytes_in_string)
 
 typedef enum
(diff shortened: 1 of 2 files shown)


zherczeg:
Good patch, but perhaps a test case could be added

rerobika:
@zherczeg I've updated this PR and added a test case for it by increasing the `max_byte_in_string` macro  which can trigger the same effect as the mentioned `.js` file.

tests/unit-core/test-strings.c
```diff
@@ -26,7 +26,7 @@
 #define test_subiters (128)
 
 /* Max bytes in string. */
-#define max_bytes_in_string (16 * 1024)
+#define max_bytes_in_string (65 * 1024)
```
zherczeg: This minor change make the bug appear? wow

rerobika: Exactly.

Issue #1990, opened 2017-08-31
Assertion 'lit_is_valid_cesu8_string (string_p, string_size)' failed in ecma_new_ecma_string_from_utf8


renatahodovan:
###### Jerry version:

```
Checked revision: 0ef2418
Build command: ./tools/build.py --clean --debug
```
###### OS:

```
Ubuntu 16.04.3 LTS
```
###### Test case:

```javascript
unescape("%f׊nd")
```
[Download test case](https://github.com/jerryscript-project/jerryscript/files/1266743/test.txt)

###### Backtrace:

```
ICE: Assertion 'lit_is_valid_cesu8_string (string_p, string_size)' failed at jerryscript/jerry-core/ecma/base/ecma-helpers-string.c(ecma_new_ecma_string_from_utf8):154.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
(gdb) bt
#0  0x00000000005e4d28 in raise ()
#1  0x00000000005e4eda in abort ()
#2  0x000000000057ec65 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-port/default/default-fatal.c:71
#3  jerry_fatal (code=code@entry=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#4  0x000000000057f350 in jerry_assert_fail (assertion=<optimized out>, file=<optimized out>, function=<optimized out>, line=<optimized out>)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#5  0x000000000040eac9 in ecma_new_ecma_string_from_utf8 (string_p=0x9c6458 <jerry_global_heap+296> "%f\327ýd", string_size=5)
    at jerryscript/jerry-core/ecma/base/ecma-helpers-string.c:154
#6  0x00000000004b65c6 in ecma_builtin_global_object_unescape.isra.12 (arg=<optimized out>)
    at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-global.c:1336
#7  ecma_builtin_global_dispatch_routine (builtin_routine_id=<optimized out>, this_arg_value=<optimized out>, arguments_list=<optimized out>, 
    arguments_number=<optimized out>) at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:215
[...]

PR #2019, opened 2017-09-20
Fixes pattern parsing in ecma_builtin_global_object_unescape ()

rerobika:
Issue #1990 revealed an unhandled corner case while parsing pattern. This patch fixes it and also adds a test case.


yichoi:
LGTM

PR #2022, opened 2017-09-22
Implement arrow function parsing.

zherczeg:
Note: the special this behaviour of arrow functions is not implemented.


zherczeg:
(Increases binary size by 300 bytes if it is enabled, and no performance change on sunspider)

PR #2025, opened 2017-09-26
Template literal

zherczeg:


zherczeg:
When this feature is enabled from es6, the code size increase by 32 byte. No other effect (except the usual noise).

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -590,6 +584,112 @@ parser_check_arrow_function (parser_context_t *context_p) /**< context */
...
 
+#ifndef CONFIG_DISABLE_ES2015_TEMPLATE_STRINGS
+
+/**
+ * Parse template literal.
+ */
+static void
+parser_parse_template_literal (parser_context_t *context_p) /**< context */
+{
+  bool empty_head = true;
```
LaszloLango: A bit nitpicky, but `is_empty_head` would be better.

PR #2028, opened 2017-09-27
Fix pending breakpoints in python debugger client

DanielBallaSZTE:
There was an issue where pending breakpoints were incorrectly tried to be assigned to invalid files, since they missed the source name check. It's also worth mentioning that pending breakpoints were not "moved" correctly to active status, just got copied there instead, making them appear in the pending list still.
This patch fixes both of these issues.

### DIFF
--- a/jerry-debugger/jerry-client-ws.py
+++ b/jerry-debugger/jerry-client-ws.py
@@ -226,8 +226,17 @@ class DebuggerPrompt(Cmd):
         """ Insert breakpoints on the given lines or functions """
         if args == "":
             print("Error: Breakpoint index expected")
+        elif ':' in args:
+            try:
+                args_second = int(args.split(':', 1)[1])
+                if args_second < 0:
+                    print("Error: Positive breakpoint index expected")
+                else:
+                    set_breakpoint(self.debugger, args, False)
+            except ValueError as val_errno:
+                print("Error: Positive breakpoint index expected: %s" % val_errno)
         else:
-            set_breakpoint(self.debugger, args)
+            set_breakpoint(self.debugger, args, False)
 
     do_b = do_break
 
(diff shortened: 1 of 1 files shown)


zherczeg:
```
+TypeError: argument of type 'NoneType' is not iterable
```

DanielBallaSZTE:
@zherczeg I've fixed the issue, plus added an additional check if the user would over-index the pending breakpoint.

jerry-debugger/jerry-client-ws.py
```diff
@@ -826,14 +826,25 @@ def parse_source(debugger, data):
     # Try to set the pending breakpoints
     if len(debugger.pending_breakpoint_list) != 0:
         logging.debug("Pending breakpoints list: %s", debugger.pending_breakpoint_list)
+        indexes = []
+        bp_list = debugger.pending_breakpoint_list
+        source_lines = func_desc["source"].count('\n')
```
zherczeg: The source should be split into lines already, shouldn't it?

DanielBallaSZTE:
@zherczeg I've fixed the mentioned issue, furthermore added a check, if the user tries to give invalid input for breakpoint creation.

jerry-debugger/jerry-client-ws.py
```diff
@@ -826,14 +835,30 @@ def parse_source(debugger, data):
...
+        for breakpoint in bp_list:
+            for src in debugger.function_list.values():
+                if src.source_name == bp_list[breakpoint].source_name:
+                    source_lines = len(src.source)
+                else:
+                    source_lines = 0
+
+            if bp_list[breakpoint].line:
+                if bp_list[breakpoint].line <= source_lines:
+                    indexes.append(breakpoint)
```
zherczeg: Just a question: can't we delete the breakpoint here?

jerry-debugger/jerry-client-ws.py
```diff
@@ -826,14 +835,30 @@ def parse_source(debugger, data):
     # Try to set the pending breakpoints
     if len(debugger.pending_breakpoint_list) != 0:
         logging.debug("Pending breakpoints list: %s", debugger.pending_breakpoint_list)
+        indexes = []
+        bp_list = debugger.pending_breakpoint_list
+
+        for breakpoint in bp_list:
+            for src in debugger.function_list.values():
+                if src.source_name == bp_list[breakpoint].source_name:
```
zherczeg: Does function breakpoints has source names? Is it an empty string?

DanielBallaSZTE: Function breakpoints don't have source names.

DanielBallaSZTE:
@zherczeg I've updated the PR, made the set_breakpoint function have a return value, that way we can decide whether the function breakpoint was set in a file or not, therefore we can just simply delete the pending breakpoint if the pending breakpoint was added.
I would still keep the checks in the source:line breakpoint style, since it needs less time checking if it should be added to the correct file than trying to set every pending breakpoint in every file. Sadly since the function breakpoints have no names associated to them, we can not add a check to them.

PR #2032, opened 2017-09-29
Free the GC mark bit to increase the number of object types.

zherczeg:


zherczeg:
The effect is noise: -0.808% RSS, +0.071% perf, and no binary size change.

zherczeg:
The key effect is adding 8 more possible object types.

PR #2035, opened 2017-10-03
Fix possible uninitialized value

gabrielschulhof:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-function.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-function.c
@@ -159,7 +159,7 @@ ecma_builtin_function_dispatch_construct (const ecma_value_t *arguments_list_p, 
   ECMA_STRING_TO_UTF8_STRING (arguments_str_p, arguments_buffer_p, arguments_buffer_size);
   ECMA_STRING_TO_UTF8_STRING (function_body_str_p, function_body_buffer_p, function_body_buffer_size);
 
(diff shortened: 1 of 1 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-function.c
```diff
@@ -159,7 +159,7 @@ ecma_builtin_function_dispatch_construct (const ecma_value_t *arguments_list_p,
   ECMA_STRING_TO_UTF8_STRING (arguments_str_p, arguments_buffer_p, arguments_buffer_size);
   ECMA_STRING_TO_UTF8_STRING (function_body_str_p, function_body_buffer_p, function_body_buffer_size);
 
-  ecma_compiled_code_t *bytecode_data_p;
+  ecma_compiled_code_t *bytecode_data_p = NULL;
```
zherczeg: This is an output argument, how can it be uninitialized?

LaszloLango: I agree, this initialization is unnecessary. It looks to me that `parser_parse_script` always sets the output argument.

gabrielschulhof: A certain compiler [complains](https://travis-ci.org/gabrielschulhof/zephyr.js/jobs/282261720#L937) under certain circumstances and then stops complaining if I explicitly initialize it.

PR #2036, opened 2017-10-04
Replace JERRY_JS_PARSER feature guard with JERRY_DISABLE_JS_PARSER

akosthekiss:
All feature guards of jerry-core are deciding about the inclusion
or exclusion of a feature based on the state (defined or undefined)
of a macro -- except for the JS parser guard, which requires
`JERRY_JS_PARSER` to be defined with a value of either 0 or 1. This
has some issues:
- The engine cannot be built with a "clean" compiler invocation,
  i.e., without any `-D` command line macro definitions. This is
  painful for targets that must use a different build system from
  the project's own python/cmake-based one.
- Some build systems in targets and even some code in jerry-code
  are already confused about the different semantics of
  `JERRY_JS_PARSER`, and simply define it without a value and make
  decisions based on the macro being simply defined or not.

This patch renames the guard to `JERRY_DISABLE_JS_PARSER` and makes
use of it in jerry-core based on its state, not based on its value.
As obvious from the guard name, the default for the JS parser is
that it is included in the build.

The patch also touches those targets in the repository that
explicitly defined the original macro (correctly or incorrectly).


yichoi:
@akosthekiss please rebase it to the master

akosthekiss:
Rebased, conflict resolved.

akosthekiss:
I had to touch a completely independent part of the code base (cppcheck) to ensure that the PR passes. The change is in a separate commit, I'm happy to break it out into a new PR if that's the conclusion.

zherczeg:
I don't think this is a good idea. The point of the original approach was that those who don't use cmake must be aware of this macro.

akosthekiss:
There is no other feature in the project that mandates the declaration of a macro on the compiler command line. What necessitates the JS parser to be different? My arguments against that:
- It is easy to forget that this is a feature that uses a completely different semantics from the rest. So easy, that it is already being forgotten and misused. Some targets declare it with 0/1, but some only declare it without a value. (Note: I have no idea how those targets work then.) Core code is also already misusing it at some places.
- Those who use the cmake build system are allowed to be unaware of this macro. The cmake build system treats the JS parser just like any other feature: it has a sensible default for it, and if the `FEATURE_JS_PARSER` cmake macro is undefined, it uses that default (which is `ON`). So, those who use cmake can rely on defaults for all features, but those who (have to) use a different build system can rely on defaults for all but one feature. What's the point in that?
- It is a very nice feature (warning: concept overload) to be able to compile the project in a clean way, without any command line options (or with as few as possible). The next example is going to be a bit demagogue, but: it should be quite the norm that the project just builds with `cc -o jerry.o jerry-all-in-one.c` (yes, there may be need for include path specification, or C standard version declaration, etc. -- the example is oversimplifying intentionally).

zherczeg:
I am not sure I follow, since these are the arguments why this particular feature must be different from other macros.

zherczeg:
(I think by FEATURE_JS_PARSER you refer to JERRY_JS_PARSER)

akosthekiss:
@zherczeg Now, I'm not sure I follow. But let me try and go into details. Basically, there are 3 ways to build the project (from command line):
- The python way: `tools/build.py` orchestrates the whole build process by invoking cmake to create makefiles and invoking make to do the all the compilation and linking. The python script has several command line arguments to control whether various features of the project are enabled or disabled (by passing appropriate macros the cmake). And all these feature arguments have defaults. (Relevant: `--js-parser ON|OFF` with ON as default.)
- The cmake way: if someone doesn't want or cannot use the python build script but still can use cmake. From the cmakelists, cmake creates the makefiles. The cmakelists declare various variables that control features of the project by adding/not adding preprocessor macros to the compiler command line. These variables can come from the command line of cmake via `-D...` constructs but also have defaults declared in cmakelists. (Relevant: `-DFEATURE_JS_PARSER=ON|OFF` with ON as default.) (Note: the python script actually sets these cmake variables based on its command line options.)
- The "no/other build system" or direct way: sometimes you are forced to simply pass the sources of jerry into another build system. Then the only way to control the features is to define preprocessor macros manually. For most of the features, this is simply either defining or not defining a macro. Not defining a macro means requesting the opposite semantics that is indicated by the macro name (i.e., if the macro name contains ENABLE, not defining the macro means disabling the feature; and the other way around with disable). (BUT! Relevant: `-DJERRY_JS_PARSER=0|1`)

So, both the python and cmake ways allow ALL features to be controlled in a consistent way with sensible defaults. In the direct way, all features can be controlled by defined/undefined macros named ENABLE/DISABLE *but* the JS parser feature. The difference in the naming and in the semantics and in the lack of a default is quite an inconsistency, which is targeted by this PR.

zherczeg:
The inconsistency is intentional. A while ago we tried both enable/disable macros for the parser, and there were groups who complained. I don't remember which side you took (both had valid arguments), but I remember that the final decision was that everybody needs to be aware that parser can be disabled, and you explicitly need to specify you want the parser or not. Honestly, I don't want to restart the debate.

akosthekiss:
HONESTLY, I DO WANT to restart the debate.

akosthekiss:
And I'm willing to reconsider my past arguments, if I had any and if I took any sides. The requirement that "everybody needs to be aware that parser can be disabled" simply is not true and does not hold because in 2 out of 3 ways of building the project you can be as unaware as you like. I don't think that you ever explicitly write `tools/build.py --js-parser=on` -- simply because you don't have to, because you have a default in place -- and you only ever write `--js-parser=off` if you want to opt out.

zherczeg:
Since the current one is the result of a conclusion, I would like to hear others opinion first.

akosthekiss:
Please, link that conclusion.

zherczeg:
https://github.com/jerryscript-project/jerryscript/pull/1639
the patch does not reveal anything, so probably it was spread in irc/issues etc

akosthekiss:
It's really hard to take that PR as a conclusion. It's a patch with no description other than its title, and there is only a single review by you stating that people would not expect the parser to be optional. This is self-reference.

(BTW, people would not expect standard EcmaScript features to be optional in an ES-compliant engine, still we have them optional. And none of those features have mandatory macros.)

zherczeg:
I haven't noticed that the parser was disabled by default before that patch. Enabling by default is better. I am still neutral to this change.

zherczeg:
@janjongboom @jiangzidong @martijnthe since this patch affects your build systems, please let us know your opinion. 

tl;dr; The patch propose a default enabled parser rather than setting the parser explicitly.

akosthekiss:
Note: there is still an open question whether to break out the topmost, logically independent commit into a separate PR?

zherczeg:
You don't need to. How much is the growth of the log due to removing --quiet?

akosthekiss:
4076 lines (see https://travis-ci.org/jerryscript-project/jerryscript/jobs/2841525149)

martijnthe:
> @janjongboom @jiangzidong @martijnthe since this patch affects your build systems, please let us know your opinion.
> tl;dr; The patch propose a default enabled parser rather than setting the parser explicitly.

I'm not too worried about making changes to our build if it improves the quality.

That said, I personally much more prefer to use the *value* of a define as the condition (`#if`) instead of the existence of a define (`#ifdef`).

I came to prefer this the hard way... Sometimes you make a typo in the place where you write the `#define`. If you use `#if` the preprocessor will yell at you and you find out about your typo. While if you use `#ifdef`, your typo will go unnoticed...

So long story short, if changes are being made to these guards, I would prefer to change them to use `#if` everywhere instead of `#ifdef`.

akosthekiss:
Does that mean that we should rewrite the whole code base where guards are used? (The goal of this PR was increasing consistency across build methods by adding defaults where they were missed (JS parser) and using the same style of guards. The change of all `#if[n]def`s to `#if`s achieves the same but with a significantly higher code churn.)

janjongboom:
@zherczeg I'm fine with this too, but I agree with Martijn on the `#ifdef` vs `#if`.

zherczeg:
Memories are coming back. The reason why the parser define is so special is that this is the only feature which you can disable, and still not part of the minimal profile. So if somebody new come, and wants to create a super minimal jerry, it is enough to take the sources and the minimal profile, and create the smallest possible jerryscript, isn't it? But that is actually not true. This form raises the visibility of this feature which is good for those who has no time to read the documentation.

akosthekiss:
> good for those who has no time to read the documentation

Sorry but this is pure nonsense. This project is a JavaScript engine for highly constrained embedded devices. There is no use case scenario, according to my best understanding, where Average Joe grabs and builds the vanilla project on its own. It is an engine which needs an embedding environment to be useful for anything. Average Joe, read the fine manual.

(BTW, does that last sentence mean that cmakers have more time for reading, by definition?)

martijnthe:
> Does that mean that we should rewrite the whole code base where guards are used? (The goal of this PR was increasing consistency across build methods by adding defaults where they were missed (JS parser) and using the same style of guards. The change of all #if[n]defs to #ifs achieves the same but with a significantly higher code churn.)

I guess this PR would already cause churn in the configuration area, so if your goal is to make it consistent, I'd go for using `#if` everywhere. Configuration usually happens in one place in the project, so shouldn't be too bad to update a client project I think?

> why the parser define is so special

Yeah... I don't know, I don't think it's a very strong story TBH. There are easier and better ways to make new comers at ease... And I also kind of agree with @akosthekiss that if you're a C developer, you're probably used to having to create a `config.h` of sorts.

akosthekiss:
> I guess this PR would already cause churn in the configuration area

Just did a quick estimate on the churn of the `#if` approach and compared it to the here-proposed solution. Rewriting all guards to `#if`s would touch some 1700 lines (give or take, but I guess that it's a lower bound) in jerry-core only, while bringing the JS parser feature in line with the existing `#if[n]def` approach was done in ~50 lines in this PR (+47 −53, to be exact), but that already includes the fixup of all in-tree targets, and related and unrelated tool script updates, too. That's a ~35x multiplier (minimum).

And one more thing occurred to me. Right now, we have some 33(?) features/builtins/etc in jerry-core that are guarded by macros. I think (hope) that even if we changed to an `#if`-based approach, we wouldn't force users (==those who compile the library) to pass all these 30+ macros to the compiler on the command line (CFLAGS, whatnot), only those, which would deviate from their defaults. Then, we would need to define those defaults somewhere (in some config.h), as their name would not encode the semantics anymore (e.g., instead of having a macro like `JERRY_DISABLE_FEATURE_X` describing an enabled-by-default feature, we would have a macro `JERRY_FEATURE_X` and also needed to have `#ifndef JERRY_FEATURE_X` / `#define JERRY_FEATURE_X 1` / `#endif` in the config file to describe what the default is).

Now, the problem is that this config header (to be included by all files in jerry-core) should be an internal header of jerry-core. (IMO, at least. I've never seen a built library distributed together with the config header that was used for the build.) But right now, there are some "externals" who use some ENABLE/DISABLE macros in their code: all of jerry-ext, jerry-port, and jerry-main have guarded code blocks depending on jerry-core configuration macros. To me, that's clearly a layering violation and some bad smell in design or in build system - but it also has a practical effect: either all these components would have to include jerry-core's config.h or they would need to duplicate the default definition logic for the used macros. None of these two approaches sounds good to me (I've been fighting against internal header accesses from outside for some time, while default duplication and syncing is quite error prone).

OK, that was my two cents. Thanks for bearing with me.

martijnthe:
Thanks for running the numbers @akosthekiss. Does sound like a lot of churn. I'm on the fence. If consistency is the goal, I'd just stick to `#ifdef` everywhere for now.

> either all these components would have to include jerry-core's config.h or they would need to duplicate the default definition logic for the used macros

FWIW, one other option: I've seen other projects use `-include` (see https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html#Preprocessor-Options) to include the default config.h.

akosthekiss:
@martijnthe re: `-include`: Ofc you can prepend the config.h to every source you compile. But in this case, you'd have to prepend the config.h of jerry-core to the sources of jerry-ext (or jerry-main, or jerry-port). In my eyes, this is just a tad better than including the config.h directly. (Having said that, I know that adding macro definitions to the command line is also far from perfect.)

@zherczeg Can we conclude?

zherczeg:
What do you mean?

akosthekiss:
@zherczeg we've had "I'm fine with this too" and "I'd just stick to #ifdef everywhere for now" (and I'm taking the liberty to skip the "but" parts from the quotes). Is this enough feedback from target maintainers or is anything else still needed to conclude whether to land or reject this PR?

zherczeg:
I am still uncertain. I hoped for stronger opinions.

There are two types of features in jerry:
a) disabled by default
b) listed in minimal.profile

The correct solution is (and this was before) is make the parser disabled by default (this is from source code perspective, cmake and other build systems can use a different default). But this cause(d) confusion when jerry is inlined by a project. Making the parser enabled by default keeps its special behaviour without signaling it.

I kind of worry that we start the cycle again and rework this check in every month.

What is your opinion?

akosthekiss:
My opinions
- "cmake and other build systems can use a different default": I strongly disagree. All build techniques should behave alike, as much as possible. And defaults definitely fall into this category.
- "The correct solution is (and this was before) is make the parser disabled by default": I disagree again. That makes the default build unusable in the general case. (Being able to run snapshots only is/should not be the general case for a JS engine, even if it is targeted for constrained devices.) None of the other features hinder functionality. So it's completely OK to have some features default to enabled while others default to disabled.
- "start the cycle again and rework this check in every month": we haven't even completed a full circle yet.
  - First, we had `#ifdef`-based parser feature with an inconsistently named macro and disabled by default at source level but enabled by default in cmake builds. (With no other feature having different defaults for different build techniques.)
  - Then we had `#if`-based parser feature (the technique not used for any other feature) with no default at source level and still enabled by default in cmake.
  - Now I propose to have an `#ifndef`-based parser feature with consistent naming and defaults across all features and build techniques.
- "Making the parser enabled by default keeps its special behaviour without signaling it.": I don't know what should be signalled.
- Final opinion: we should allow a default build, that does not require the declaration of any feature macros for the compiler. (This, we don't have ATM.)

PR #2042, opened 2017-10-11
Rework error to use a global slot to store the error value.

zherczeg:
This change frees up the error bit in ecma_value_t, which allows to define 4 more value types (e.g. symbols). To keep API compatibility we introduce a box for values with error flag.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-value.c
+++ b/jerry-core/ecma/base/ecma-helpers-value.c
@@ -28,15 +28,10 @@
  * @{
  */
 
-/**
- * Masking the type and flags
- */
-#define ECMA_VALUE_FULL_MASK (ECMA_VALUE_TYPE_MASK | ECMA_VALUE_ERROR_FLAG)
-
 JERRY_STATIC_ASSERT (ECMA_TYPE___MAX <= ECMA_VALUE_TYPE_MASK,
                      ecma_types_must_be_less_than_mask);
 
-JERRY_STATIC_ASSERT ((ECMA_VALUE_FULL_MASK + 1) == (1 << ECMA_VALUE_SHIFT),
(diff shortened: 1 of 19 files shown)


yichoi:
LGTM

akosthekiss:
It's post mortem, but: What's the result of this change on code size, memory consumption, and running time? Moreover, what's the maintenance cost of this change on lib users? (I see several changes in jerry-main which most probably have to be adapted to all targets/users, which does not happen in this patch, so I think all targets just went broken?)

zherczeg:
Yes, these are the interesting questions. Let me reply here:
binary size: 2.5K decrease (140->137.5K) so it somehow affected the GCC internal optimizations a lot. There is noise in memory (every allocation change affects gc run) and runtime (probably the large binary change), but nothing unusal.

There is one change in jerry-main, which was a bug in there, but it wasn't revealed before: we remove the error flag from a value in a function, but forgot to notify the caller about it, so it frees the object with the error flag, which causes a crash now. The error flag removal intentionally expects a pointer, and not a value, and has the freedom to do anything with the object, and after the flag removal the new value must be used not the old one. So this is a bug.

This patch does not change the API at all, although it would be good to do some redesign before 2.0. Snapshot API is messy for example. Parser API shows aging as well.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -589,34 +586,30 @@ ecma_get_object_from_value (ecma_value_t value) /**< ecma value */
 } /* ecma_get_object_from_value */
 
 /**
- * Invert a boolean value
+ * Get pointer to error reference from ecma value
```
martijnthe: @zherczeg: the docstring and function name suggests that it's getting the error from a value, but isn't it the other way around?

PR #2043, opened 2017-10-13
Introduce a way to directly save snapshot functions and load them back

galpeter:
Added two new api functions:
* jerry_parse_and_save_function_snapshot
* jerry_load_function_snapshot_at

The jerry_parse_and_save_function_snapshot function allows
creating snapshots from snapshot arguments and body source.

The jerry_load_function_snapshot_at function enables loading
back functions from a given snapshot as a JS function object.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -4170,6 +4170,76 @@ main (void)
 - [jerry_exec_snapshot](#jerry_exec_snapshot)
 
 
+## jerry_parse_and_save_function_snapshot
+
+**Summary**
+
+Generate function snapshot from the specified source code
+with the given arguments.
+
+The function arguments and function body are
+passed as separated arguments.
+
+**Prototype**
+
+```c
+size_t
+jerry_parse_and_save_function_snapshot (const jerry_char_t *source_p,
+                                        size_t source_size,
+                                        const jerry_char_t *args_p,
+                                        size_t args_size,
+                                        bool is_strict,
+                                        uint32_t *buffer_p,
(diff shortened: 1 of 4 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -4318,6 +4388,95 @@ main (void)
 - [jerry_parse_and_save_snapshot](#jerry_parse_and_save_snapshot)
 
 
+## jerry_load_function_snapshot_at
+
+**Summary**
+
+Load the selected snapshot function from the specified buffer as a function object.
```
zherczeg: I would mention that the lexical env is always the global lexical env.

jerry-core/api/jerry-snapshot.c
```diff
@@ -494,24 +495,26 @@ snapshot_load_compiled_code (const uint8_t *base_addr_p, /**< base address of th
 
 #endif /* JERRY_ENABLE_SNAPSHOT_EXEC */
 
+#ifdef JERRY_ENABLE_SNAPSHOT_SAVE
```
LaszloLango: I think we should do the same here as in other API functions. Move the guard into the function's body and use `JERRY_UNUSED`, if snapshot save is disabled. We usually prefer no-op rather than a compile time liner error.

LaszloLango: Oh, my bad. It is only a static function.

docs/02.API-REFERENCE.md
galpeter: @zherczeg I've updated the docs.

PR #2045, opened 2017-10-18
Fix API docs

galpeter:
The #2043 introduced a bit of error in the API
docs as there was missing backticks.


yichoi:
LGTM

PR #2044, opened 2017-10-17
target: mbedos5: Added DigitalIn & Serial js bindings

fluentart:
This is now working:

```
var a = new DigitalIn(BUTTON1); 
print(a.read());
```

Also added the other missing class constructors to lib_drivers.h as commented out lines. Lets get them all done!

JerryScript-DCO-1.0-Signed-off-by: Rouan van der Ende rouan.vanderende@iotnxt.com

PR #2050, opened 2017-10-21
target: mbedos5: Add support for DigitalIn

janjongboom:
Follow up from #2044 

Verified on FRDM-K64F.

Thanks @fluentart

### DIFF
--- a/targets/mbedos5/README.md
+++ b/targets/mbedos5/README.md
@@ -11,6 +11,7 @@ so far:
 - [Nordic Semiconductor NRF52 Development Kit](https://developer.mbed.org/platforms/Nordic-nRF52-DK/)
 - [NXP Freedom K64F](https://developer.mbed.org/platforms/FRDM-K64F/)
 - [STM NUCLEO F401RE](https://developer.mbed.org/platforms/ST-Nucleo-F401RE/)
+- [STM NUCLEO F411RE](https://developer.mbed.org/platforms/ST-Nucleo-F411RE/)
 - [Silicon Labs EFM32 Giant Gecko](https://developer.mbed.org/platforms/EFM32-Giant-Gecko/)
 
 ## Features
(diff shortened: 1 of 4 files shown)


akosthekiss:
@janjongboom As is, the PR cannot be merged because of "administrative" reasons. Right now, the commit identifies itself as "Rouan van der Ende authored, Jan Jongboom committed". And even Travis catches this and reports "Signed-off-by message is incorrect. The following line should be at the end of the 3c3cfe6 commit's message: 'JerryScript-DCO-1.0-Signed-off-by: Rouan van der Ende rouan@8bo.org'.". (And obviously, it would be strange if you signed it as Rouan.) So, right now, the PR is in violation of the DCO (https://github.com/jerryscript-project/jerryscript/blob/master/DCO.md). You might `--reset-author` if both you and @fluentart are OK with that (and then the commit would align with DCO items b or c).

fluentart:
Same as other PR, I used the wrong email address (work one, instead of my github email). Sorry for the mixup.

glistening:
@janjongboom The recent PRs (#2050, #2051) in mbedos port are about external APIs. They seems beyond the scope of jerryscript. I am not sure jerryscript repository is the right location to put those code. It may be ok if it is an example. But if you have a plan to provide several APIs, what about pushing them in another repository?

yichoi:
Closing. I don't think this is common APIs to be included JerryScript main repository which to be balanced with APIs from other targets.

janjongboom:
@glistening @yichoi @akosthekiss I've been thinking about this... I'm not sure where they belong. It's all APIs that are part of the *core* Mbed OS, just like InterruptIn, DigitalOut, etc.

But actually if I look at it from a functional perspective, I'd not have the mbedos5 target in core jerry at all, but rather as a library you include from an Mbed OS project. Maybe I should refactor that out.

PR #2051, opened 2017-10-21
target: mbedos5: Add serial support

janjongboom:
Follow up from #2044 

Please do not land yet, am verifying it this week.

### DIFF
--- a/targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/Serial-js.h
+++ b/targets/mbedos5/jerryscript-mbed/jerryscript-mbed-drivers/Serial-js.h
@@ -1,0 +1,22 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


akosthekiss:
@janjongboom As is, the PR cannot be merged because of "administrative" reasons. Right now, the commit identifies itself as "Rouan van der Ende authored, Jan Jongboom committed". And even Travis catches this and reports "Signed-off-by message is incorrect. The following line should be at the end of the 456848a commit's message: 'JerryScript-DCO-1.0-Signed-off-by: Rouan van der Ende rouan@8bo.org'.". (And obviously, it would be strange if you signed it as Rouan.) So, right now, the PR is in violation of the DCO (https://github.com/jerryscript-project/jerryscript/blob/master/DCO.md). You might `--reset-author` if both you and @fluentart are OK with that (and then the commit would align with DCO items b or c).

fluentart:
Yeah sorry my bad, I missed the part where you had to signoff with the same email address used on the github account.

yichoi:
Closing. I don't think this is common APIs to be included JerryScript main repository which to be balanced with APIs from other targets.

Issue #2053, opened 2017-10-24
Math.random() generates same results


tosyu:
Because `srand()` is not used on initialization. Without it, the `Math.random()` function generates the same results every time (not every call, but every time the environment runs).

This works on other JS engines like V8 (nodejs). IMHO we should look to achieve same results to be consistent with JS platform

akosthekiss:
@tosyu I'm not sure about the reference to other JS engines, because initialising the libc random generator should be the responsibility of the application, not the library (the JS engine). That being said, however, it's true that `srand` is not called anywhere, not even in jerry-main. Will you be submitting a patch to fix that?

zherczeg:
do we have srand in jerry libc?

akosthekiss:
Note: interestingly, tests/unit-core/test-common.h calls `srand`.

@zherczeg yes, we do.

tosyu:
@akosthekiss I did not find anything in the documentation about initializing random number generator with `srand()`, so for an application developer this would require to make assumptions about `Math.random()` implementation, which in my opinion is a "no-no". Either it should be fixed in the code or it should be written in the documentation that `srand()` needs to be called

akosthekiss:
@tosyu I see no way for fixing this issue in code (as code of the engine) because library should not (re-)initialise the RNG (might have already been done by the application, and so the lib could cause conflicts). It can and should be fixed in the code of the jerry-main application(s) though. I also agree that it could be documented that `Math.random` is based on libc's `rand`, which would imply the need for initialisation by `srand` (and that could also be written explicitly). (No idea though where to put that note, as we have no dedicated documentation for the JS objects/functions as we assume that everything is documented in the standard. Perhaps the documentation of `jerry_init` in the API docs could have some similar notes. I'm also open to suggestions.)

tosyu:
@akosthekiss IMHO `docs/03.API-EXAMPLE.md` would be sufficient to put the information about `srand()`; as an example

akosthekiss:
@tosyu That could work, too. Are you up for submitting a PR for this or shall we put that on our todo list?

tosyu:
@akosthekiss please put it on the todo list for now

PR #2056, opened 2017-10-25
Add info to documentation about random numbers

DanielBallaSZTE:
Issue #2053 has highlighted the fact that random numbers are always generated with the same seed.
An example of generating different random numbers, other than the original seed, has been added to the documentation.

### DIFF
--- a/docs/03.API-EXAMPLE.md
+++ b/docs/03.API-EXAMPLE.md
@@ -551,6 +551,50 @@ Value of x is 12
 Value of x is 17
 ```
 
+## Step 8. Changing the seed of pseudorandom generated numbers
+
+If you want to change the seed of `Math.random()` generated numbers, you have to initialize the seed value with `srand`.
+A recommended method is using `jerry_port_get_current_time()` or something based on a constantly changing value, therefore every run produces truly random numbers.
+
+[doctest]: # ()
+
+```c
+#include <string.h>
+#include <stdlib.h>
+#include "jerryscript.h"
+#include "jerryscript-port.h"
+#include "jerryscript-ext/handler.h"
+
+int
+main (void)
+{
+  /* Initialize srand value */
+  srand ((unsigned) jerry_port_get_current_time ());
+
+  /* Generate a random number, and print it */
+  const jerry_char_t script[] = "var a = Math.random (); print(a)";
+  size_t script_size = strlen ((const char *) script);
+
+  /* Initialize the engine */
+  jerry_init (JERRY_INIT_EMPTY);
+
+  /* Register the print function */
+  jerryx_handler_register_global ((const jerry_char_t *) "print",
+                                  jerryx_handler_print);
+
+  /* Evaluate the script */
+  jerry_value_t eval_ret = jerry_eval (script, script_size, false);
+
+  /* Free the JavaScript value returned by eval */
+  jerry_release_value (eval_ret);
+
+  /* Cleanup the engine */
+  jerry_cleanup ();
+
+  return 0;
+}
+```
+
 ## Further steps
 
 For further API description, please visit [API Reference page](https://jerryscript-project.github.io/jerryscript/api-reference/) on [JerryScript home page](https://jerryscript-project.github.io/jerryscript/).
--- a/jerry-main/main-unix-minimal.c
+++ b/jerry-main/main-unix-minimal.c
@@ -71,6 +71,7 @@ int
 main (int argc,
       char **argv)
 {
+  srand ((unsigned) jerry_port_get_current_time ());
   if (argc <= 1 || (argc == 2 && (!strcmp ("-h", argv[1]) || !strcmp ("--help", argv[1]))))
   {
     print_help (argv[0]);
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -424,6 +424,7 @@ int
 main (int argc,
       char **argv)
 {
+  srand ((unsigned) jerry_port_get_current_time ());
   const char *file_names[argc];
   int files_counter = 0;
 
--- a/targets/curie_bsp/jerry_app/quark/main.c
+++ b/targets/curie_bsp/jerry_app/quark/main.c
@@ -136,6 +136,7 @@ void eval_jerry_script (int argc, char *argv[], struct tcmd_handler_ctx *ctx)
 
 void jerry_start ()
 {
+  srand ((unsigned) jerry_port_get_current_time ());
   jerry_init (JERRY_INIT_EMPTY);
   jerry_value_t global_obj_val = jerry_get_global_object ();
   jerry_value_t print_func_name_val = jerry_create_string ((jerry_char_t *) "print");
--- a/targets/mbed/source/main.cpp
+++ b/targets/mbed/source/main.cpp
@@ -24,6 +24,7 @@ static Serial pc (USBTX, USBRX); //tx, rx
 
 static int jerry_task_init (void)
 {
+  srand ((unsigned) jerry_port_get_current_time ());
   int retcode;
 
   DECLARE_JS_CODES;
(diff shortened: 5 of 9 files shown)


docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,47 @@ Value of x is 12
...
+main (void)
+{
+  const jerry_char_t script[] = "var a = Math.random(); print(a);";
+  size_t script_size = strlen ((const char *) script);
+
+  bool ret_value = jerry_run_simple (script, script_size, JERRY_INIT_EMPTY);
+
+  return (ret_value ? 0 : 1);
+}
+```
```
akosthekiss: I'm not convinced that a negative example is needed. A positive example where Math.random is used and srand is properly called should be enough with proper explanation.

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,47 @@ Value of x is 12
...
+
+Running it would always output the same randomly generated numbers, as the seed is always the same.
+However, adding srand would use a different seed, therefore different random numbers are generated.
+
+```c
+#include <string.h>
+#include "jerryscript.h"
```
akosthekiss: Let's try prevent example code from bitrot, add `[doctest]: # ()` before the code block. This will guide the "doctest" testing phase to extract the example code and compile-link-run it (and will signal error if the code is not OK).

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,47 @@ Value of x is 12
...
+
+int
+main (void)
+{
+  const jerry_char_t script[] = "var a = Math.random(); print(a);";
+  size_t script_size = strlen ((const char *) script);
+  int my_random_seed = 123456;
```
akosthekiss: I'm a bit worried, this example will not print different numbers on different runs. In that sense it's just like the "negative example", only starting from a different (but also fixed) seed. I wonder whether the `double jerry_port_get_current_time (void);` Date Port API function could be used to initialize the seed.

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,47 @@ Value of x is 12
...
+
+  srand ((unsigned) my_random_seed);
+  bool ret_value = jerry_run_simple (script, script_size, JERRY_INIT_EMPTY);
+
+  return (ret_value ? 0 : 1);
```
akosthekiss: (style) those parentheses are not really needed

docs/03.API-EXAMPLE.md
DanielBallaSZTE: Yeah, we could use time. What I meant to point out was the fact that there is no srand call in Jerry, therefore you have to manually take care about generating truly random numbers.

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,32 @@ Value of x is 12
 Value of x is 17
 ```
 
+## Changing the seed of pseudorandom generated numbers
+
+If you want to change the seed of Math.random() generated numbers, you have to initialize srand with a seed value.
```
akosthekiss: Better wording perhaps: "you have to initialize the seed value with srand."

Also: code (or function references) are typeset differently from the rest of the text: use backticks around Math.random() and srand.

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,32 @@ Value of x is 12
...
+A recommended method is using `jerry_port_get_current_time(void)` or something based on a constantly changing value, therefore every run produces truly random numbers.
```
akosthekiss: I don't think that (void) is needed.

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,32 @@ Value of x is 12
...
+#include "jerryscript.h"
+#include "jerryscript-port.h"
+
+int
+main (void)
+{
+  const jerry_char_t script[] = "var a = Math.random ();";
+  size_t script_size = strlen ((const char *) script);
+
+  srand ((unsigned) (jerry_port_get_current_time ()));
```
akosthekiss: `(unsigned) jerry_port_get_current_time ()` should be enough (no need for the extra parentheses)

also in all the "real" `main` functions

targets/nuttx-stm32f4/jerry_main.c
```diff
@@ -346,6 +346,8 @@ int main (int argc, FAR char *argv[])
 int jerry_main (int argc, char *argv[])
 #endif
 {
+  srand ((unsigned) (jerry_port_get_current_time ()));
```
akosthekiss: nuttx's `jerry_port_get_current_time` implementation is a simple `return 0;`, so calling `srand` does not help a lot here

targets/particle/source/main.cpp
```diff
@@ -70,6 +70,7 @@ js_delay (const jerry_value_t func_value, /**< function object */
 static void
 init_jerry ()
 {
+  srand ((unsigned) (jerry_port_get_current_time ()));
```
akosthekiss: I cannot find which port implementation the particle target is using. Does it have a sensible `jerry_port_get_current_time` implementation? (There is a chance that the target went bitrotten. If that's the case, we should not try touch it up now with `srand` but let a target maintainer bring it up to speed first -- or add `srand` at the same time.)

targets/tizenrt-artik053/apps/jerryscript/jerry_main.c
```diff
@@ -320,6 +320,7 @@ static jerry_log_level_t jerry_log_level = JERRY_LOG_LEVEL_ERROR;
 static int
 jerry_cmd_main (int argc, char *argv[])
 {
+  srand ((unsigned) (jerry_port_get_current_time ()));
```
akosthekiss: the implementation is dummy for this target, too. calling srand does not help here either

DanielBallaSZTE:
@akosthekiss Thanks for the review, I've updated the pull request.

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,47 @@ Value of x is 12
 Value of x is 17
 ```
 
+## Changing the seed of pseudorandom generated numbers
```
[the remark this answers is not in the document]
akosthekiss: I'm still not completely content with the example. I'd like to ask two more things: Please add a "Step 8." prefix to the section title, as this doc is a step-by-step guide (as mentioned before).

docs/03.API-EXAMPLE.md
```diff
@@ -551,6 +551,32 @@ Value of x is 12
...
+```c
+#include <string.h>
+#include <stdlib.h>
+#include "jerryscript.h"
+#include "jerryscript-port.h"
+
+int
+main (void)
+{
+  const jerry_char_t script[] = "var a = Math.random ();";
```
akosthekiss: The other thing that bothers me is that the example code is a bit "alone", i.e., it does not build on the info introduced by the previous steps. At least a print should be added to the example script so that when someone really copies this code out into a source file and links it to jerry, they get some output (a value changing randomly on each run) on display. It's not as simple as a `jerry_run_simple` ofc, but not that complex either. Something along the lines:

```c
  /* Initialize engine */

  /* Register 'print' function from the extensions */
  jerryx_handler_register_global ((const jerry_char_t *) "print",
                                  jerryx_handler_print);
  /* Evaluate script */
  jerry_value_t eval_ret = jerry_eval (script, script_size, false);

  /* Free JavaScript value, returned by eval */
  jerry_release_value (eval_ret);

  /* Cleanup engine */
  jerry_cleanup ();

  return 0;
```

(+ `print` in the script and the `srand` call ofc (preferably with a comment highlighting what's happening), and it may also need the inclusion of `"jerryscript-ext/handler.h"`)

DanielBallaSZTE:
@akosthekiss Thank you for your review, I've updated the commit message and the pull request as well, according to it.

PR #2057, opened 2017-10-25
Move snapshot generation function to the snapshot tool

galpeter:
Now there is a snapshot tool which can merge snapshots
but was unable to generate snapshots by itself.

This change moves the snapshot generation utilities
to the snapshot tool.


jerry-main/main-unix-snapshot.c
```diff
@@ -73,6 +93,237 @@ read_file (uint8_t *input_pos_p, /**< next position in the input buffer */
...
+  if (snapshot_file_p == NULL)
+  {
+    jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Error: Unable to write snapshot file: '%s'\n", output_file_name_p);
+    return JERRY_STANDALONE_EXIT_CODE_FAIL;
+  }
+
+  fwrite (output_buffer, sizeof (uint8_t), snapshot_size, snapshot_file_p);
+  fclose (snapshot_file_p);
+
+  printf ("Created snapshot file: '%s' (%d bytes)\n", output_file_name_p, snapshot_size);
```
akosthekiss: cppcheck signals signed/unsigned int mismatch here ("warning(invalidPrintfArgType_sint): %d in format string (no. 2) requires a signed integer given in the argument list.")

jerry-main/main-unix-snapshot.c
```diff
@@ -73,6 +93,237 @@ read_file (uint8_t *input_pos_p, /**< next position in the input buffer */
...
+    if (literal_file_p == NULL)
+    {
+      jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Error: Unable to write literal file: '%s'\n", save_literals_file_name_p);
+      return JERRY_STANDALONE_EXIT_CODE_FAIL;
+    }
+
+    fwrite (output_buffer, sizeof (uint8_t), literal_buffer_size, literal_file_p);
+    fclose (literal_file_p);
+
+    printf ("Created literal file: '%s' (%d bytes)\n", save_literals_file_name_p, literal_buffer_size);
```
akosthekiss: ... and here

zherczeg:
I like this change

galpeter:
@akosthekiss I've updated the PR.

jerry-main/main-unix-snapshot.c
```diff
@@ -73,6 +93,237 @@ read_file (uint8_t *input_pos_p, /**< next position in the input buffer */
...
+               .help = "print generated opcodes"),
+  CLI_OPT_DEF (.id = OPT_GENERATE_CONTEXT, .opt = "c", .longopt = "context",
+               .help = "specify the execution context of the snapshot: "
+                       "global or eval (default: global).", .meta = "MODE"),
+  CLI_OPT_DEF (.id = OPT_GENERATE_LITERAL_LIST, .longopt = "save-literals-list-format",
+               .meta = "FILE",
+               .help = "export literals found in parsed JS input (in list format)"),
+  CLI_OPT_DEF (.id = OPT_GENERATE_LITERAL_C, .longopt = "save-literals-c-format",
+               .meta = "FILE",
+               .help = "export literals found in parsed JS input (in C source format)"),
```
akosthekiss: the original `--save-snapshot-for-{global,eval}` options have already been rewritten to a new `--context={global,eval}` format. wouldn't it make sense to have something similar for `--save-literals-{c,list}-format`, too? e.g., as simple as `--save-literals={c,list}`.

tools/run-tests.py
```diff
@@ -237,16 +237,22 @@ def run_jerry_debugger_tests(options):
 
     return ret_build | ret_test
 
+def update_test_args_with_snapshot_tool(jerry_path, test_args):
+    if "--snapshot-tool" in test_args:
+        snapshot_idx = test_args.index("--snapshot-tool")
+        test_args[snapshot_idx] = "--snapshot-tool=%s-snapshot" % (jerry_path)
```
akosthekiss: is this rewrite really needed? patching up the cli options this way is a bit creepy, and the same effect could be achieved by simply adding this "deduce snapshot tool path from engine path" logic to the run-test-suite.sh script. (explanation following there...)

tools/runners/run-test-suite.sh
```diff
@@ -44,13 +44,21 @@ then
...
+if [[ "$1" =~ ^--snapshot-tool=.* ]]
 then
     TEST_FILES="$TEST_FILES.snapshot"
     TEST_FAILED="$TEST_FAILED.snapshot"
     TEST_PASSED="$TEST_PASSED.snapshot"
     IS_SNAPSHOT=true;
-    shift
+
+    SNAPSHOT_TOOL=${1#--snapshot-tool=}
+    SNAPSHOT_TOOL=(${SNAPSHOT_TOOL//,/ })
```
akosthekiss: ...following from above: you'd just have `SNAPSHOT_TOOL=$ENGINE-snapshot` here. (agreed, it looses a bit of flexibility by not allowing for snapshot tools placed anywhere in the file system for those who run this script manually. but who does get restricted by that, actually? run-test.py does force to have jerry and jerry-snapshot side-by-side anyway)

jerry-main/main-unix-snapshot.c
galpeter: the logic is different, for --save-literals* you need to specify the output file as argument. This way you can generate the literal file and also the snapshot.

jerry-main/main-unix-snapshot.c
```diff
@@ -73,6 +93,237 @@ read_file (uint8_t *input_pos_p, /**< next position in the input buffer */
...
+  (void) argc;
+
+  bool is_save_literals_mode_in_c_format = false;
+  bool is_snapshot_mode_for_global = true;
+  jerry_init_flag_t flags = JERRY_INIT_EMPTY;
+
+  uint32_t number_of_files = 0;
+  uint8_t *source_p = input_buffer;
+  size_t source_length = 0;
+  const char *output_file_name_p = "js.snapshot";
```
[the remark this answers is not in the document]
galpeter: IMHO having different default output names makes the intent more clear.
But we can change it to a common one. Currently merge uses "merged.snapshot",  so what do you think would be the best common name?

tools/runners/run-test-suite.sh
galpeter: Certainly we can simply assume the path, but I wanted to give the test site runner more freedom if it is executed manually. 

But I'll update this part.

jerry-main/main-unix-snapshot.c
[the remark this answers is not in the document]
akosthekiss: "js.snapshot" is fine by me for both. it'd be a quite simple and elegant default.

jerry-main/main-unix-snapshot.c
akosthekiss: oh that's right

tools/runners/run-test-suite.sh
```diff
@@ -50,7 +50,14 @@ then
...
     IS_SNAPSHOT=true;
-    shift
+
+    SNAPSHOT_TOOL=${ENGINE}-snapshot
+    if [ ! -x $SNAPSHOT_TOOL ]
+    then
+        echo "$0: $SNAPSHOT_TOOL: not an executable"
+        exit 1
+    fi
+    shift;
```
akosthekiss: no need for the semicolon (shouldn't be used elsewhere either)

jerry-main/main-unix-snapshot.c
```diff
@@ -74,6 +121,227 @@ read_file (uint8_t *input_pos_p, /**< next position in the input buffer */
...
+
+  fwrite (output_buffer, sizeof (uint8_t), snapshot_size, snapshot_file_p);
+  fclose (snapshot_file_p);
+
+  printf ("Created snapshot file: '%s' (%lu bytes)\n", output_file_name_p, (unsigned long) snapshot_size);
+
+  if (save_literals_file_name_p != NULL)
+  {
+    const size_t literal_buffer_size = jerry_parse_and_save_literals ((jerry_char_t *) source_p,
+                                                                       source_length,
```
akosthekiss: seems that the arg indentation is off by one space

tools/runners/run-test-suite.sh
akosthekiss: semicolon still here but removed from after IS_SNAPSHOT=true?

akosthekiss: ping on this minor change request

PR #2068, opened 2017-10-31
Fix jerry-libc's srand function

DanielBallaSZTE:
The previous implementation of srand was wrong in jerry-libc.
It set all 4 values to the seed, not modifying anything in them, causing random values to be repeated pretty often.
This approach fixes the mentioned issue.

### DIFF
--- a/jerry-libc/jerry-libc.c
+++ b/jerry-libc/jerry-libc.c
@@ -276,10 +276,14 @@ rand (void)
 void
 srand (unsigned int seed) /**< new seed */
 {
-  libc_random_gen_state[0] =
-  libc_random_gen_state[1] =
-  libc_random_gen_state[2] =
-  libc_random_gen_state[3] = seed;
+  libc_random_gen_state[0] = (uint32_t) ((seed * 14316555781)
+                                         + (seed * 1183186591)
+                                         + (seed * 622729787)
+                                         + (seed * 338294347));
+
+  libc_random_gen_state[1] = 842502087;
+  libc_random_gen_state[2] = 3579807591;
+  libc_random_gen_state[3] = 273326509;
 } /* srand */
 
 /**


DanielBallaSZTE:
I'm unsure how to add a test case for this issue though, so any suggestions are appreciated.

jerry-libc/jerry-libc.c
```diff
@@ -27,7 +27,11 @@
 /**
  * State of pseudo-random number generator
  */
-static uint32_t libc_random_gen_state[4] = { 1455997910, 1999515274, 1234451287, 1949149569 };
+static uint32_t initial_states[3] = { 842502087, 3579807591, 273326509};
+static uint32_t _x;
+static uint32_t _y;
+static uint32_t _z;
+static uint32_t _w;
```
akosthekiss: The patch identifies correctly that `srand` should not set all state elements to the same value. However, that does not mean that the whole RNG, together with its state representation and `rand` implementation should be rewritten (there is no problem with any of those).

Moreover, this change here does not only rewrite the state representation but introduces a new issue by leaving `_x`, `_y`, `_z`, and `_w` uninitialized.

Please, revert this block to the original code version (with `libc_random_gen_state`).

jerry-libc/jerry-libc.c
```diff
@@ -257,17 +261,12 @@ strlen (const char *s)
...
 
-  libc_random_gen_state[0] = libc_random_gen_state[1];
-  libc_random_gen_state[1] = libc_random_gen_state[2];
-  libc_random_gen_state[2] = libc_random_gen_state[3];
-
-  libc_random_gen_state[3] ^= libc_random_gen_state[3] >> 19;
-  libc_random_gen_state[3] ^= intermediate;
-
-  return (int) (libc_random_gen_state[3] % (RAND_MAX + 1));
+  return (int) (0x7FFFFFFF & (_w = (_w ^ (_w >> 19)) ^ (t ^ (t >> 8))));
```
akosthekiss: As explained above, this refactoring is also unnecessary, please revert. (Here is also a newly introduced issue: this formula does not take `RAND_MAX` into consideration, which is a libc standard-mandated macro and `rand` must work with that. So, if we changed our `stdlib.h` that would not affect this code. So, please, revert.)

jerry-libc/jerry-libc.c
```diff
@@ -276,10 +275,14 @@ rand (void)
...
-  libc_random_gen_state[2] =
-  libc_random_gen_state[3] = seed;
+  _x = (unsigned) ((seed * 14316555781)
+       + (seed * 1183186591)
+       + (seed * 622729787)
+       + (seed * 338294347));
+
+  _y = initial_states[0];
+  _z = initial_states[1];
+  _w = initial_states[2];
```
akosthekiss: Here, this approach is not bad. It only needs to be adapted to the original state representation. Something like:

```c
  libc_random_gen_state[0] = (unsigned) ((seed * 14316555781)
                                         + (seed * 1183186591)
                                         + (seed * 622729787)
                                         + (seed * 338294347));

  libc_random_gen_state[1] = 842502087;
  libc_random_gen_state[2] = 3579807591;
  libc_random_gen_state[3] = 273326509;
```

DanielBallaSZTE:
@akosthekiss Hmm, you actually are right, it's a bit of an overkill to change the whole rand implementation. Thank you for your review.

jerry-libc/jerry-libc.c
```diff
@@ -27,7 +27,7 @@
 /**
  * State of pseudo-random number generator
  */
-static uint32_t libc_random_gen_state[4] = { 1455997910, 1999515274, 1234451287, 1949149569 };
+ static uint32_t libc_random_gen_state[4] = { 1455997910, 1999515274, 1234451287, 1949149569 };
```
akosthekiss: This leading space seems to be a leftover / addition by mistake. Probably Travis does not like it either.

PR #2078, opened 2017-11-09
Add fast path to ecma_op_object_get with magic string.

zherczeg:
This patch causes a similar effect than my other patch a while ago: binary 139880 -> 137532 and -13% RSS. This is the effect of code / data layout change in memory. No other major effect.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -666,7 +666,12 @@ ecma_op_object_get_own_data_prop (ecma_object_t *object_p, /**< the object */
 } /* ecma_op_object_get_own_data_prop */
 
 /**
- * [[Get]] ecma object's operation
+ * [[Get]] operation of ecma object
+ *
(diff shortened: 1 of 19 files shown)


LaszloLango:
@zherczeg please check the Travis failures.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -703,6 +703,47 @@ ecma_op_object_get (ecma_object_t *object_p, /**< the object */
   return ecma_make_simple_value (ECMA_SIMPLE_VALUE_UNDEFINED);
 } /* ecma_op_object_get */
 
+/**
+ * [[Get]] ecma object's operation using magic string name
+ *
+ * See also:
+ *          ECMA-262 v5, 8.6.2; ECMA-262 v5, Table 8
+ *
+ * @return ecma value
```
LaszloLango: Please add more details here. For example it when it will return undefined and when an object value.

PR #2079, opened 2017-11-09
[jerryx-module]Fix a bug in module name comparison

jiangzidong:
In the original code, if the native module name is "abcd", and the new module name is "ab", then the new module will be treated as native module.


jiangzidong:
@gabrielschulhof @martijnthe

martijnthe:
LGTM (informally)

I'm guessing @gabrielschulhof used `strncmp` (instead of `strcmp`) for safety?

PR #2084, opened 2017-11-10
Add grouping macro guards for disabling engine features

akosthekiss:
Until now, the engine's feature set was configurable either via
`CONFIG_*` macro guards defined individually on compiler command
line or via profiles (which are text files listing macro guards,
picked up by the cmake build system and turning them into compiler
command line options). And the features under profile control are
all enabled by default (i.e., all macros are `CONFIG_DISABLE_*`).

This causes a maintenance issue when new features are added to the
engine, because the disabling macros have to be added to all
profiles that don't include the new features. This can even cause
"compatibility break" for applications that embed JerryScript but
don't use the cmake or the python build system, because then
profiles are unavailable and all feature disabling guards have to
be explicitly passed to the compiler. (I.e., if such an application
wants to use the ES5.1 feature set, it must define all the ES2015
disable macros; if the engine is developed further and a new ES2015
feature gets implemented, then the new feature will sneak into the
application's binary unless its own build system is changed to add
the new feature guard.) Even the in-repo example Curie BSP target
seems to have suffered from this maintenance problem.

This patch introduces two new grouping macro guards that enable the
disabling of all ES5.1 builtins and all ES2015 features. As the
grouping logic is in config.h, the maintenance of non-cmake-based
build systems becomes easier (and there is no change for the python
and cmake-based build systems).


martijnthe:
Good idea!

DanielBallaSZTE:
Great patch, LGTM informally.

jerry-core/config.h
```diff
@@ -17,6 +17,33 @@
 #define CONFIG_H
 
 /**
+ * Group of builtin-related features that can be disabled together.
+ */
+#ifdef CONFIG_DISABLE_BUILTINS
+# define CONFIG_DISABLE_ANNEXB_BUILTIN
```
LaszloLango: wrong style: no space needed after `#`

akosthekiss: The rest of this header uses this type of indentation. I did not want to mix styles within a single file so went for the space.

LaszloLango: Legacy... You can update those line too if you want.

dbatyai: Just to stir things up a bit, why do we hate indentation in preproc conditions in the first place? Personally I would always prefer a well indented block over a giant blob of unindented mess anytime.

akosthekiss: We have 7 header + 3 source files that carry this legacy. Can I suggest to have this file in a uniform style for now and have a follow up style fix PR that does nothing else but gets rid of all the legacy hashmark-spaces everywhere?

LaszloLango: OK, let's land it this way and discuss the spaces later.

PR #2086, opened 2017-11-10
WIP: Illustrate `return this` reference error

gabrielschulhof:

### DIFF
--- a/tests/unit-core/test-return-this.c
+++ b/tests/unit-core/test-return-this.c
@@ -1,0 +1,64 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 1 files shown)


gabrielschulhof:
@brianjjones the test in this example seeks to properly capture the situation whereby returning `this` seems to retain a reference to the object containing the function. The test doesn't fail because of memory that should have been released but wasn't, nor does it fail because of a refcount that is too high. Perhaps this is because I didn't properly cover the usage from JS of the values created natively.

Note that when [returning `this`](https://github.com/jerryscript-project/jerryscript/pull/2086/files#diff-87118d72e4920de564a10d4e0ebe27c4R34) you have to acquire the value, otherwise the test fails with a reference count underflow.

Could you please compare this test with the ZJS [use case](https://github.com/01org/zephyr.js/blob/master/src/zjs_event.c#L115-L134) and let me know if there are any differences in the way the JS values are being used from JS, or in the way they are acquired/released on the native side?

brianjjones:
Thanks for the explanation, certainly has helped tracking down the issue in our code.

PR #2088, opened 2017-11-10
Try to determine compiler version if not set by CMake

### LINKED ISSUE #2085: riot-stm32f4 fails to build
Hi,

I want to build jerryscript for RIOT. If I try to build the source via
```
make -f ./targets/riot-stm32f4/Makefile.riot
```
I get the following errors
```c
[  0%] Building C object jerry-core/CMakeFiles/jerry-core.dir/ecma/operations/ecma-regexp-object.c.obj
/home/frey/Desktop/Projects/jerryscript/jerry-core/ecma/operations/ecma-regexp-object.c: In function 're_match_regexp':
/home/frey/Desktop/Projects/jerryscript/jerry-core/ecma/operations/ecma-regexp-object.c:747:7: error: this statement may fall through [-Werror=implicit-fallthrough=]
       {
       ^
/home/frey/Desktop/Projects/jerryscript/jerry-core/ecma/operations/ecma-regexp-object.c:801:7: note: here
       case RE_OP_CAPTURE_GROUP_START:
       ^~~~
/home/frey/Desktop/Projects/jerryscript/jerry-core/ecma/operations/ecma-regexp-object.c:933:14: error: this statement may fall through [-Werror=implicit-fallthrough=]
         bc_p = old_bc_p;
         ~~~~~^~~~~~~~~~
/home/frey/Desktop/Projects/jerryscript/jerry-core/ecma/operations/ecma-regexp-object.c:938:7: note: here
       case RE_OP_CAPTURE_GREEDY_GROUP_END:
       ^~~~
cc1: some warnings being treated as errors
```
Compiling jerryscript via
```
python tools/build.py
```
works fine. Am I missing something?

Thanks in advance!

dbatyai:
Fixes #2085

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,11 @@ if(CMAKE_C_COMPILER_ID MATCHES "TI")
   set(USING_TI 1)
 endif()
 
+# Try to determine compiler version if not automatically set by CMake
+if(USING_GCC AND NOT CMAKE_C_COMPILER_VERSION)
+  EXECUTE_PROCESS(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CMAKE_C_COMPILER_VERSION)
+endif()
+
 # Determining build type
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "MinSizeRel")


CMakeLists.txt
```diff
@@ -129,6 +129,11 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${PROJECT_BINARY_DIR}/lib/")
 # Remove rdynamic option
 set(CMAKE_SHARED_LIBRARY_LINK_C_FLAGS )
 
+# Try to determine compiler version if not automatically set by CMake
+if(NOT CMAKE_C_COMPILER_VERSION)
+  EXECUTE_PROCESS(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CMAKE_C_COMPILER_VERSION)
```
akosthekiss: Two comments:
- Compiler-related scripts are around line 22-33.
- How general is this process execution, how generally is `-dumpversion` supported by compilers?

dbatyai: I'll move it up. Regarding the support, after checking, TI doesn't seem to support it, and although Clang accepts the flag it may not output the correct version. I can restrict it to `USING_GCC` if that is okay with you.

akosthekiss: Sounds OK. But we'll first need confirmation from @mfrey whether it solves the problem. 

(Note: I'm puzzled. How come that cmake cannot identify the gcc version? Gcc support has been there in cmake for ages. If we can query the version via `-dumpversion`, why cannot cmake? Is this some known cmake vs gcc version incompatibility?)

dbatyai: I'm guessing the issue is because of `cmake/toolchain_external.cmake:20`.
It says [here] (https://cmake.org/cmake/help/v3.0/module/CMakeForceCompiler.html) that `It also bypasses the check for working compiler and basic compiler information tests`.

dbatyai: (Now that we're at it, I'm not quite sure why we're using the external toolchain for the riot-stm32f4 target when we have an stm32f4 toolchain, nevertheless it wouldn't make a difference here, as we're using CMAKE_FORCE_C_COMPILER there as well.)

CMakeLists.txt
```diff
@@ -32,6 +32,11 @@ if(CMAKE_C_COMPILER_ID MATCHES "TI")
   set(USING_TI 1)
 endif()
 
+# Try to determine compiler version if not automatically set by CMake
+if(USING_GCC AND NOT CMAKE_C_COMPILER_VERSION)
+  EXECUTE_PROCESS(COMMAND ${CMAKE_C_COMPILER} -dumpversion OUTPUT_VARIABLE CMAKE_C_COMPILER_VERSION)
```
akosthekiss: (the previous thread became hidden, so let's continue here)

I have serious doubts that this is a proper way to go. Actually, I have serious doubts about the whole 'external' toolchain file and the `CMAKE_FORCE_C_COMPILER`. If one can pass `-DCMAKE_TOOLCHAIN_FILE=cmake/toolchain_external.cmake -DEXTERNAL_CMAKE_SYSTEM_PROCESSOR=armv7l -DEXTERNAL_CMAKE_C_COMPILER=arm-none-eabi-gcc -DEXTERNAL_CMAKE_C_COMPILER_ID=GNU` to cmake, why cannot they simply pass `-DCMAKE_SYSTEM_NAME=MCU -DCMAKE_SYSTEM_PROCESSOR=armv7l -DCMAKE_C_COMPILER=arm-none-eabi-gcc`? If I'm not mistaken, it is not mandatory to specify a toolchain file. And `CMAKE_C_COMPILER` can also be set without 'forcing' it (see https://cmake.org/Wiki/CMake_FAQ#How_do_I_use_a_different_compiler.3F). And then, I guess, `CMAKE_C_COMPILER_VERSION` would be correctly detected without any workarounds.

akosthekiss: I was thinking of something like this: https://github.com/akosthekiss/jerryscript/commit/1534146857d5f1b7b835d270a7ed6ed9c278002e . (This is speculative though as I cannot validate these changes.)

dbatyai: I agree, it looks better. I'm going to close this then.

PR #2089, opened 2017-11-10
Remove the 'external' toolchain file

akosthekiss:
The 'external' toolchain file does nothing but transitively sets
some cmake system variables from values received on the command
line, and forcibly sets the C compiler. However, the same cmake
system variables can be directly set via the command line, together
with the C compiler, and specifying a toolchain is not a must.
Thus, this patch drops the superfluous 'external' toolchain file
and updates cmake-based targets to invoke cmake in a simpler form.

Related changes in this commit:
- While updating the cmake invocations, all the command line
  arguments have been reviewed and simplified (removed those, which
  did not change the defaults).
- Removed unnecessary forced C compiler settings from some
  toolchain files.


akosthekiss:
Related to #2085, follow-up to #2088. The PR is speculative, needs confirmation from ESP8266, Mbed, Particle, RIOT, TizenRT, and Zephyr maintainers + from the original issue's reporter.

akosthekiss:
@dbatyai Good point. Updated.

dbatyai:
I tested the PR, the compiler version is correctly identified, but now the problem is that the arm-none-eabi toolchain fails the working compiler test, due to an undefined reference to `_exit`. In this case the [docs](https://cmake.org/Wiki/CMake_Cross_Compiling#The_toolchain_file) suggest to use the force compiler option. It seems we've come full circle.

akosthekiss:
Doesn't sound good. But let me ask one more experiment: what if we added `-DCMAKE_C_COMPILER_WORKS=TRUE` after `-DCMAKE_C_COMPILER=...`? (Or `set(CMAKE_C_COMPILER_WORKS TRUE)` in the toolchain files, depending on the target.) Would that still run compiler identification (to properly set version) but disable the checks that fail? (Because, if I see it right, these are two separate steps (the logs have separate lines for: "The C compiler identification is ..." / "Check for working C compiler: ...").

sources:
- https://stackoverflow.com/questions/10599038/can-i-skip-cmake-compiler-tests-or-avoid-error-unrecognized-option-rdynamic/30642130#30642130
- https://github.com/Kitware/CMake/blob/master/Modules/CMakeTestCCompiler.cmake

dbatyai:
I checked, it seems setting `CMAKE_C_COMPILER_WORKS` solves the issue, the compiler is still identified correctly.

akosthekiss:
Thanks! I'll update the PR soon.

akosthekiss:
PR updated (sets `CMAKE_C_COMPILER_WORKS` for all non-OS-based targets).

rerobika:
I've tested this PR in ESP8266-12E and it works properly. Nice job!

akosthekiss:
Rebased onto master, which was a good thing to do as the new target build jobs have revealed that a `CMAKE_C_COMPILER_WORKS` define was missing for Zephyr. Commit fixed, PR updated.

PR #2091, opened 2017-11-13
Refactor CLI argument handling and test runner logic of run-tests.py

akosthekiss:
On argument handling:
- Merged all signoff check options into one, option value chooses
  between strict/tolerant check variants.
- Added sensible metavars to options with arguments (consistent
  with metavars of build.py).
- Reordered options to match the order the checks are executed in.
- Added `--all` alias to `--precommit` check option.
- Beautified code (always placed `help` on separate line, removed
  unnecessary arguments of `add_argument` that just repeated their
  defaults).

On test runner logic:
There was too many code duplication on how checks were executed, it
got refactored. No change in semantics.

The only change was in Travis CI-related invocations,
`--check-signed-off-travis` had to be rewritten to
`--check-signed-off=travis`.

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -17,6 +17,7 @@
 from __future__ import print_function
 
 import argparse
+import collections
 import os
 import subprocess
 import sys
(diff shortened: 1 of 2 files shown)


tools/run-tests.py
```diff
@@ -331,56 +340,34 @@ def run_buildoption_test(options):
...
-        ret = run_test262_test_suite(options)
-
-    if not ret and (options.all or options.unittests):
-        ret = run_unittests(options)
-
-    if not ret and (options.all or options.buildoption_test):
-        ret = run_buildoption_test(options)
-
-    sys.exit(ret)
+class Check(object):
```
galpeter: For such cases there are namedtuples: https://docs.python.org/2/library/collections.html#collections.namedtuple

tools/run-tests.py
```diff
@@ -130,33 +130,42 @@ def get_binary_path(bin_dir_path):
...
-                        help='Run signed-off check in tolerant mode')
-    parser.add_argument('--check-signed-off-travis', action='store_true', default=False,
-                        help='Run signed-off check in tolerant mode if on Travis CI and not checking a pull request')
-    parser.add_argument('--check-cppcheck', action='store_true', default=False, help='Run cppcheck')
-    parser.add_argument('--check-doxygen', action='store_true', default=False, help='Run doxygen')
-    parser.add_argument('--check-pylint', action='store_true', default=False, help='Run pylint')
-    parser.add_argument('--check-vera', action='store_true', default=False, help='Run vera check')
-    parser.add_argument('--check-license', action='store_true', default=False, help='Run license check')
-    parser.add_argument('--check-magic-strings', action='store_true', default=False,
+    parser.add_argument('--check-signed-off', metavar='TYPE', choices=['strict', 'tolerant', 'travis'], nargs='?', const='strict',
```
galpeter: Pylint checker reports too long line (over 120 chars). In fact there are other places not related to this PR, but we can fix that in a later PR.

tools/run-tests.py
```diff
@@ -331,56 +340,34 @@ def run_buildoption_test(options):
...
-    sys.exit(ret)
+class Check(object):
+    def __init__(self, enabled, runner, *args):
+        self.enabled = enabled
+        self.runner = runner
+        self.args = args
 
+def main(options):
+    checks = [
+        Check(options.check_signed_off, run_check, [settings.SIGNED_OFF_SCRIPT] + {'tolerant': ['--tolerant'], 'travis': ['--travis']}.get(options.check_signed_off, [])),
```
galpeter: Pylint checker reports too long line (over 120 chars)

tools/run-tests.py
akosthekiss: BTW, namedtuples don't allow `*args`, do they? (making classes more powerful/future-proof)

akosthekiss:
Updated (line length limit taken into account, but class not rewritten to namedtuple)

tools/run-tests.py
galpeter: It would be great to use namedtuple here (and also at the Options, again another PR imho). The problem currently is that the pylint returns a warning: `Too few public methods (0/2) (too-few-public-methods)` which of course we could change.

What if we do this?:
```
_Check = namedtuple("Check", ["enabled", "runner", "args"])
def Check(enabled, runner, *args):
    return _Check(enabled, runner, args)
```
With this we have support for the `*args` also.

akosthekiss: Honestly, the only thing the pylint warning signals to me is that we don't have pylint configured properly. I think pylint's own documentation is quite open about this:

> What Pylint says is not to be taken as gospel and Pylint isn’t smarter than you are: it may warn you about things that you have conscientiously done. Pylint tries hard to report as few false positives as possible for errors, but it may be too verbose with warnings.

In another project, I've been experimenting with reasonable settings for pylint and this was that worked for us: https://github.com/renatahodovan/grammarinator/blob/master/.pylintrc (It's quite smaller and also more maintainable than what we have now in JerryScript.)

On `Check` and `_Check`: for the current purposes, the suggested approach is functionally equivalent to the PR's approach but less pythonic and/or less OO: values returned by `Check` are not instances of `Check`.

akosthekiss: I've pushed an experimental commit on top of the original showing how a (IMO) better maintained pylintrc would work. Please have a look.

LaszloLango:
IMHO pylint config change should go to a separate PR.

akosthekiss:
@galpeter OK, I've followed the review and I've rewritten `Check` to be a `namedtuple`.  (Note be taken, I still don't like those constructs. :) )

@LaszloLango Ofc, the pylintrc commit was just an experiment.

Issue #2094, opened 2017-11-14
dynamic-stack-buffer-overflow in jerry_value_is_syntax_error


renatahodovan:
###### Jerry version:

```
Checked revision: 6b1ed6e0
Build command: CC=clang ./tools/build.py --clean --debug --compile-flag=-fsanitize=address --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --jerry-libc=off --static-link=off --lto=off --error-messages=on
```
###### OS:

```
Ubuntu 17.10
macOS High Sierra (10.13.1)
```
###### Test case:

It seems that jerry produces the same failure with any kind of syntax errors when compiling it with clang. E.g.:

```javascript
function a() {
```

###### Backtrace:

```
=================================================================
==1467==ERROR: AddressSanitizer: dynamic-stack-buffer-overflow on address 0xff934fcb at pc 0x080abfde bp 0xff934f88 sp 0xff934b60
READ of size 12 at 0xff934fcb thread T0
    #0 0x80abfdd in __interceptor_strcmp.part.32 crtstuff.c:?
    #1 0x816f8e3 in jerry_value_is_syntax_error jerryscript/jerry-main/main-unix.c:122
    #2 0x816e8e9 in print_unhandled_exception jerryscript/jerry-main/main-unix.c:154 (discriminator 1)
    #3 0x816d99f in main jerryscript/jerry-main/main-unix.c:790
    #4 0xf7be3985 in __libc_start_main ??:?
    #5 0x806e177 in _start ??:?

Address 0xff934fcb is located in stack of thread T0
SUMMARY: AddressSanitizer: dynamic-stack-buffer-overflow crtstuff.c:? in __interceptor_strcmp.part.32
Shadow bytes around the buggy address:
  0x3ff269a0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ff269b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ff269c0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ff269d0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ff269e0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x3ff269f0: 00 00 00 00 ca ca ca ca 00[03]cb cb cb cb cb cb
[...]

PR #2095, opened 2017-11-14
Fixed dynamic-stack-buffer-overflow in jerry_value_is_syntax_error

rerobika:
Fixes issue #2094, which introduced an error caused by jerry_value_is_syntax_error.
The problem was that the function used strcmp instead of strncmp while checking if err_str_buf contains syntax error and it caused buffer overflow.

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -109,6 +109,14 @@ jerry_value_is_syntax_error (jerry_value_t error_value) /**< error value */
   }
 
   jerry_size_t err_str_size = jerry_get_string_size (error_name);
+  const char syntax_error_str[] = "SyntaxError";
+
+  if (err_str_size != strlen (syntax_error_str) - 1)
+  {
+    jerry_release_value (error_name);
+    return false;
+  }
+
   jerry_char_t err_str_buf[err_str_size];
 
   jerry_size_t sz = jerry_string_to_char_buffer (error_name, err_str_buf, err_str_size);
@@ -119,7 +127,7 @@ jerry_value_is_syntax_error (jerry_value_t error_value) /**< error value */
     return false;
   }
 
-  if (!strcmp ((char *) err_str_buf, "SyntaxError"))
+  if (!strncmp ((char *) err_str_buf, syntax_error_str, sizeof (syntax_error_str) - 1))
   {
     return true;
   }
--- a/tests/jerry/fail/regression-test-issue-2094.js
+++ b/tests/jerry/fail/regression-test-issue-2094.js
@@ -1,0 +1,15 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function a() {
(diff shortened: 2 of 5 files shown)


rerobika:
Nuttx and Tizenrt  targets also contains these potential error possibilities. If this solutions seems approved for main-unix.c I'll update the targets's as well.

jerry-main/main-unix.c
```diff
@@ -119,7 +119,8 @@ jerry_value_is_syntax_error (jerry_value_t error_value) /**< error value */
     return false;
   }
 
-  if (!strcmp ((char *) err_str_buf, "SyntaxError"))
+  char syntax_error_str[] = "SyntaxError";
```
LaszloLango: `const char syntax_error_str[] = "SyntaxError";`

rerobika:
Targets are updated as well!

LaszloLango:
still lgtm

jerry-main/main-unix.c
```diff
@@ -119,7 +119,8 @@ jerry_value_is_syntax_error (jerry_value_t error_value) /**< error value */
     return false;
   }
 
-  if (!strcmp ((char *) err_str_buf, "SyntaxError"))
+  const char syntax_error_str[] = "SyntaxError";
+  if (!strncmp ((char *) err_str_buf, syntax_error_str, sizeof (syntax_error_str) - 1))
```
akosthekiss: This approach does not solve the issue completely. If the error object's name property is a prefix of "SyntaxError" then `strncmp` will still read beyond the end of `err_str_buf`. The following snippet triggers a dynamic-stack-buffer-overflow error even with this patch applied:

```js
var e = new SyntaxError('dummy');
e.name = 'Syntax';
throw e;
```

Properly, it should also be checked whether `err_str_size == sizeof (syntax_error_str) - 1`. If not, we can do an early return. (Actually, this can happen quite early, even before allocating `err_str_buf` in line 112.)

akosthekiss: Note: please, add the new snippet to the test suite, too.

rerobika: Good point!

jerry-main/main-unix.c
```diff
@@ -109,6 +109,14 @@ jerry_value_is_syntax_error (jerry_value_t error_value) /**< error value */
   }
 
   jerry_size_t err_str_size = jerry_get_string_size (error_name);
+  const char syntax_error_str[] = "SyntaxError";
+
+  if (err_str_size != strlen (syntax_error_str) - 1)
+  {
+    jerry_release_value (error_name);
+    return true;
```
akosthekiss: `return true;` ? This function is called `jerry_value_is_syntax_error`. This code does the opposite of what's expected.

tests/jerry/fail/regression-test-issue-2094.js
```diff
@@ -0,0 +1,29 @@
...
+
+name = "";
+try
+{
+  var b = eval (a);
+} catch (e) {
+  name = e.name;
+}
+
+assert (name == "SyntaxError")
```
akosthekiss: This new code path actually does not check what it should. If written as above, the test would pass even with the nonpatched jerry. (The stack buffer overflow error is reported against code in unix-main.c if there is an _unhandled_ SyntaxError exception in the JS. If the exception is caught, then jerry-core will deal with it and the bogus code jerry-main will not be executed.)

So, we'll need one test file with the originally reported content (`function a() {`), and another with the code snippet shown in the conversation of this PR. (Both as `jerry/fail/regression-test-issue-XXXX.js`. For the first snippet, XXXX=2094, for the second it can be either 2095 - because it was first reported here -, or 2094bis - because its a variant of the issue reported in issue 2094.)

jerry-main/main-unix.c
rerobika: Ohh, that's awkward. I'll fix it.

tests/jerry/fail/regression-test-issue-2094.js
rerobika: I see your point and I can agree with it. I'll update it.

LaszloLango:
still LGTM after the recent updates.

Issue #2080, opened 2017-11-09
Changes from #2042 break semantic API compatibility


martijnthe:
Before PR #2042, `jerry_value_clear_error_flag` and `jerry_value_set_error_flag` would take a `jerry_value_t *` and set a flag onto it. After each of these calls return, the `jerry_value_t` that the pointer points to would still be a reference to the same JS thing as before.

After PR #2042, after the `jerry_value_set_error_flag` returns, the `jerry_value_t` is updated to reference a *different* value (the error "box"). Same with `jerry_value_clear_error_flag`, it expects to take an `jerry_value_t *` pointing to such an error "box" and after the call returns, the `jerry_value_t` will contain the unboxed value.

Even though this change is source and binary compatible, semantically it's not compatible (and thus breaking the API, IMHO).

One of the more serious implications is that the client code needs to be aware of the "swapping" of the `jerry_value_t`:

Contrived example:
```
jerry_value_t sth = ...;

sth = jerry_acquire_value(sth);  // this worked fine before #2042
jerry_value_set_error_flag(&sth);

// Whoops! After PR #2042 the boxed value is not acquired...
return sth;
```

Unless we really need to free up the error flag bit in the `jerry_value_t` *now*, I'd revert #2042, I'm afraid it will break people's projects in unexpected ways. At the very least, the docs should be updated – this change is more than a transparent implementation detail.

If the freed up bit is actually really needed now, perhaps create a new API with a name that better matches what happens under the hood, mark the old one deprecated. The old one can call the new implementation, but at least this way there's a mechanism to get people's attention after pulling from upstream.

cc @zherczeg @akosthekiss

martijnthe:
cc @jiangzidong

jiangzidong:
I can provide another example which we met in our real project.

the code is like
```c
{
  jerry_value_t rv = get_a_jerry_value_from_some_api(...);

  // in the js_check_and_log_if_error, we check if rv is error, 
  // if so, call jerry_value_clear_error_flag, and print the error log 
  if (js_check_and_log_if_error(rv)){
    // error handle
  } else {
     // normal handle
  }

  jerry_release_value(rv);
}
```

[...]

PR #2096, opened 2017-11-14
Add api: jerry_get_value_without_error_flag

jiangzidong:
It is used to get the reference value from a error value

Related issue #2080 

If the input value is an error, then the api will return the acquired its reference value. If the input value is not an error, the api will return the acquired input value. So the caller need to release the return value.

It is a necessary supplement for jerry_value_clear_error_flag, because the jerry_value_clear_error_flag will change the input value directly and release the error-box automatically, which maybe not what developers expect.

@zherczeg @martijnthe 

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -1486,6 +1486,49 @@ jerry_value_set_error_flag (jerry_value_t *value_p);
 - [jerry_value_t](#jerry_value_t)
 
 
+## jerry_get_value_without_error_flag
+
+**Summary**
+
+If the input value is an error value, then return a new reference to its referenced value.
+Otherwise, return a new reference to the value itself.
+
+*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) 
+when it is no longer needed.
+
+**Prototype**
+
+```c
+jerry_value_t 
+jerry_get_value_without_error_flag (jerry_value_t value)
+```
+
+- `value` - api value
+
+**Example**
+
+```c
+{
+  jerry_value_t value;
+  ... // create or acquire value
+  
+  jerry_value_set_error_flag (&value);
+  
+  jerry_value_t real_value = jerry_get_value_without_error_flag (value);
+  ... // process the real_value. Different from `jerry_value_clear_error_flag`, 
+      // the error `value` will not be automatically released after calling 
+      // `jerry_get_value_without_error_flag`.
+
+  jerry_release_value (value);
+  jerry_release_value (real_value);
+}
+```
+
+**See also**
+
+- [jerry_acquire_value](#jerry_acquire_value)
+- [jerry_value_clear_error_flag](#jerry_value_clear_error_flag)
+
 # Getter functions of 'jerry_value_t'
 
 Get raw data from API values.
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -850,6 +850,20 @@ jerry_value_set_error_flag (jerry_value_t *value_p)
 } /* jerry_value_set_error_flag */
 
 /**
+ * If the input value is an error value, then return a new reference to its referenced value.
+ * Otherwise, return a new reference to the value itself.
+ *
+ * Note:
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return the real value of the jerry_value
+ */
+jerry_value_t jerry_get_value_without_error_flag (jerry_value_t value) /**< api value */
+{
+  return jerry_acquire_value (jerry_get_arg_value (value));
+} /* jerry_get_value_without_error_flag */
+
+/**
  * Get boolean from the specified value.
  *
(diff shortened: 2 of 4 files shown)


jiangzidong:
I will add the doc if the reviewers are OK with the api's signature and semantic

martijnthe:
@akosthekiss @zherczeg what do you think?

jerry-core/include/jerryscript-core.h
```diff
@@ -349,6 +349,11 @@ jerry_value_t jerry_acquire_value (jerry_value_t value);
 void jerry_release_value (jerry_value_t value);
 
 /**
+ * Get the 'real' value of a jerry_value
+ */
+jerry_value_t jerry_get_value (jerry_value_t value);
```
akosthekiss: Should be declared with the other error-related functions (around lines 295-297)?

jerry-core/api/jerry.c
```diff
@@ -1037,6 +1037,20 @@ jerry_release_value (jerry_value_t value) /**< API value */
...
 /**
+ * If the input value is a error value, then return its reference value.
+ * Otherwise, return itself.
+ *
+ * Note:
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return the real value of the jerry_value
+ */
+jerry_value_t jerry_get_value (jerry_value_t value)
```
akosthekiss: The name of the newly proposed function does not really reveal what it's doing. IMO. Would `jerry_get_value_without_error_flag` be too long to be useful? Or just `jerry_value_without_error_flag`? (Would align better with existing error-related API functions.)

jiangzidong: I prefer the `jerry_get_value_without_error_flag`, which is clear because of the `get`

docs/02.API-REFERENCE.md
```diff
@@ -1486,6 +1486,48 @@ jerry_value_set_error_flag (jerry_value_t *value_p);
...
 
+## jerry_get_value_without_error_flag
+
+**Summary**
+
+If the input value is a error value, then return its reference value.
+Otherwise, return itself.
+
+Note that the returned value must be freed with jerry_release_value, 
+when it is no longer needed.
```
akosthekiss: Please use the "standard" note:

```
*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
is no longer needed.
```
(style and reference)

martijnthe:
@zherczeg any further comments?

docs/02.API-REFERENCE.md
```diff
@@ -1486,6 +1486,48 @@ jerry_value_set_error_flag (jerry_value_t *value_p);
 - [jerry_value_t](#jerry_value_t)
 
 
+## jerry_get_value_without_error_flag
+
+**Summary**
+
+If the input value is a error value, then return its reference value.
+Otherwise, return itself.
```
zherczeg: Otherwise return a new reference to itself.

jiangzidong: done

jiangzidong:
Updated 
@zherczeg

martijnthe:
Can this be merged? @zherczeg @akosthekiss 
Thanks!

zherczeg:
Yes, from my side.

docs/02.API-REFERENCE.md
```diff
@@ -1486,6 +1486,48 @@ jerry_value_set_error_flag (jerry_value_t *value_p);
 - [jerry_value_t](#jerry_value_t)
 
 
+## jerry_get_value_without_error_flag
+
+**Summary**
+
+If the input value is a error value, then return its reference value.
+Otherwise return a new reference to itself.
```
zherczeg: If the input value is an error value, then return a new reference to its referenced value.
Otherwise, return a new reference to the value itself.

also update the comment for the function.

docs/02.API-REFERENCE.md
```diff
@@ -1486,6 +1486,48 @@ jerry_value_set_error_flag (jerry_value_t *value_p);
...
+      // `jerry_get_value_without_error_flag`.
+
+  jerry_release_value (value);
+  jerry_release_value (real_value);
+}
+```
+
+**See also**
+
+- [jerry_value_clear_error_flag](#jerry_value_clear_error_flag)
```
zherczeg: Add acquire_value as reference. Actually this function is closer to acquire_value than clear_error_flag.

jiangzidong:
@zherczeg Sorry that I missed your last comments. Updated.

martijnthe:
@zherczeg could you take a look again? Thank you!

LaszloLango:
I'm neutral with this change, but is this still needed after #2131?

jiangzidong:
Yes, I think it is still needed. Because in our case, we just want to get the error value, we dont want to change the ref_count of the error-box.

jiangzidong:
I will merge it since @zherczeg  already gives LGTM.

PR #2101, opened 2017-11-15
Add 'jerry_job_queue_is_empty' function to the API.

LaszloLango:
This API function could be useful on systems which use eventloop and
real asyncron calls, like IoT.js.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -870,49 +870,6 @@ jerry_eval (const jerry_char_t *source_p,
 - [jerry_create_external_function](#jerry_create_external_function)
 - [jerry_external_handler_t](#jerry_external_handler_t)
 
-## jerry_run_all_enqueued_jobs
-
-**Summary**
-
-Run enqueued Promise jobs until the first thrown error or until all get executed.
-
-**Prototype**
-
-```c
-jerry_value_t
-jerry_run_all_enqueued_jobs (void)
-```
-
-- return value - result of last executed job, may be error value.
-
-**Example**
-
-[doctest]: # ()
-
-```c
-#include <string.h>
-#include "jerryscript.h"
-
-int
-main (void)
-{
-  jerry_init (JERRY_INIT_EMPTY);
-
-  const jerry_char_t script[] = "new Promise(function(f,r) { f('Hello, World!'); }).then(function(x) { print(x); });";
-  size_t script_size = strlen ((const char *) script);
-
-  jerry_value_t parsed_code = jerry_parse (script, script_size, false);
-  jerry_value_t script_value = jerry_run (parsed_code);
-  jerry_value_t job_value = jerry_run_all_enqueued_jobs ();
-
-  jerry_release_value (job_value);
-  jerry_release_value (script_value);
-  jerry_release_value (parsed_code);
-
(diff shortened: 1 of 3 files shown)


jerry-core/api/jerry.c
```diff
@@ -2492,6 +2472,42 @@ jerry_resolve_or_reject_promise (jerry_value_t promise, /**< the promise value *
 } /* jerry_resolve_or_reject_promise */
 
 /**
+ * Check whether is job queue empty or not.
```
DanielBallaSZTE: typo: `Check whether job queue is empty` instead of `Check whether is.. `

jerry-core/api/jerry.c
```diff
@@ -2492,6 +2472,42 @@ jerry_resolve_or_reject_promise (jerry_value_t promise, /**< the promise value *
...
+ */
+inline bool __attr_always_inline___
+jerry_job_queue_is_empty (void)
+{
+#ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
+  return JERRY_CONTEXT (job_queue_head_p) == NULL;
+#else /* CONFIG_DISABLE_ES2015_PROMISE_BUILTIN */
+  return true;
+#endif /* !CONFIG_DISABLE_ES2015_PROMISE_BUILTIN */
+}
```
DanielBallaSZTE: Missing function comment from the end.

docs/02.API-REFERENCE.md
```diff
@@ -2235,6 +2192,78 @@ jerry_resolve_or_reject_promise (jerry_value_t promise,
...
+## jerry_job_queue_is_empty
+
+**Summary**
+
+Checks whether the job queue is empty.
+
+**Prototype**
+
+```c
+bool jerry_job_queue_is_empty (void);
```
DanielBallaSZTE: Function type and name should be in seperate lines.

jiangzidong:
@LaszloLango Could you give a quick example that in which cases we need call this API?

LaszloLango:
@jiangzidong https://github.com/Samsung/iotjs/issues/1292

jiangzidong:
So it is more like a debug tool, right?

LaszloLango:
No, it is not a debug tool.

zherczeg:
Why it is not enough to simply run the jobqueue? I suspect it is nop if the queue is empty.

jiangzidong:
sorry I got confused. 

the logic about the jobqueue and eventqueue should be
```
while(if there is pending event like timer, fs, network, ...)
{
   wait the event triggered
   call the corresponding event handler (usually will call js callback, but its ok to call a pure native handler)
   call jerry_run_all_enqueued_jobs, which will make sure all the enqueued job are executed in sequence
}
```

So I don't know when to use the `jerry_job_queue_is_empty`.

Back to the iotjs issue, the workaround is counting the pendingPromiseObj, note that the pendingPromiseObj will not be added into the jobqueue, only the resolved promise will be added into the jobqueue. For example. if the async readfile takes 5s to finish reading, then during the 5s, we have the pendingPromiseObj, but the jobqueue is still empty

LaszloLango:
@jiangzidong no worries. I found a better fix this. I don't need this new API anymore :) Sorry for the inconvenience.

PR #2102, opened 2017-11-16
Add Travis CI jobs for build testing several targets

akosthekiss:
Hitherto, code under the `targets` directory was not tested and so
its maintenance was sometimes speculative. This commit adds build
testing for several targets to prevent them from bit rotting.
Targets covered by this commit are: ESP8266, Mbed, Mbed OS 5,
NuttX, RIOT, Tizen RT, and Zephyr.

Some issues were revealed and fixed:
- ESP8266: added missing include for `uint32_t` typedef.
- Tizen RT: replaced missing `str_to_uint` with `strtol`.

### DIFF
--- a/targets/esp8266/Makefile.travis
+++ b/targets/esp8266/Makefile.travis
@@ -1,0 +1,51 @@
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+# Default target for running the build test outside the Travis CI environment.
+all:
+	$(MAKE) install
+	$(MAKE) script
+
+
+## Targets for installing build dependencies of the ESP8266 JerryScript target.
+
+# Install tools via apt.
+install-apt-get-deps:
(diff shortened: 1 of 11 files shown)


targets/esp8266/Makefile.travis
```diff
@@ -0,0 +1,51 @@
...
+	wget https://github.com/esp8266/esp8266-wiki/raw/master/libs/libhal.a -O ../ESP8266_SDK/lib/libhal.a
+
+# Perform all the necessary (JerryScript-independent) installation steps.
+install: install-apt-get-deps install-xtensa-kx106-gcc install-espressif-sdk
+
+
+## Targets for building ESP8266 with JerryScript.
+
+# Build the firmware (ESP8266 with JerryScript).
+script:
```
LaszloLango: I'd add `install` target as a dependency of this target to prevent the parallel "build" of `script` and `install`

LaszloLango: If you agree in my opinion, then please do it in the other new makefiles too.

akosthekiss: That never happens (I mean the parallel execution of script and install). If `all` is called as the default target of make, then it explicitly lists the order of execution by invoking sub-makes in a specific order (install, script). While `.travis.yml` calls make twice, first with `install` target in the `install` stage and then with `script` target in script stage. (The make targets intentionally reflect the names of the Travis CI job stage names to highlight which target is called in which stage.) If `install` was a dependency of `script`, `install` would run again in script stage.

See:
```yml
    - env: JOBNAME="ESP8266 Build Test"
      cache: ccache
      install: make -f ./targets/esp8266/Makefile.travis install
      script: make -f ./targets/esp8266/Makefile.travis script
```

LaszloLango: I see now, why you did it in this way. It is good, leave as is.

LaszloLango:
>This PR does NOT want to deal with the above issues but they are left for target maintainers for consideration.

@akosthekiss Please open an issue for them.

PR #2113, opened 2017-11-17
Fix api inconsistency for promises

tkeri:
The chage makes the jerry_value_is_promise api method
available in all cases, just like other promise api methods

JerryScript-DCO-1.0-Signed-off-by: Tamas Keri tkeri@inf.u-szeged.hu

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -715,8 +715,6 @@ jerry_value_is_object (const jerry_value_t value) /**< api value */
   return ecma_is_value_object (jerry_get_arg_value (value));
 } /* jerry_value_is_object */
 
-#ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
-
 /**
  * Check if the specified value is promise.
  *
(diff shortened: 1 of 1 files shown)


tkeri:
This incosystency causes build error when the '--unittest' flag is enable,
because the 'jerry_value_is_promise' is undefine.

jiangzidong:
LGTM (informally).

akosthekiss:
(but please fix the vera++-reported style issue)

tkeri:
Done

jerry-core/api/jerry.c
```diff
@@ -727,14 +725,16 @@ bool
...
   jerry_assert_api_available ();
-
+#ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
   jerry_value_t promise = jerry_get_arg_value (value);
   return (ecma_is_value_object (promise)
           && ecma_is_promise (ecma_get_object_from_value (promise)));
+#else /* !CONFIG_DISABLE_ES2015_PROMISE_BUILTIN */
+  JERRY_UNUSED (value);
+  return false;
+#endif /* CONFIG_DISABLE_ES2015_PROMISE_BUILTIN */
```
akosthekiss: Just spotted that the preproc else/endif comments have been swapped. I think the endif should have the !CONFIG_DISABLE comment. Could you please fix this too before landing? Sorry for the late hit.

Issue #2111, opened 2017-11-16
 Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed in jmem_heap_finalize


renatahodovan:
###### Jerry version:

```
Checked revision: 7692aa9d
Build command: tools/build.py --clean --debug --profile=es2015-subset
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
var value = {};
var poisonedThen = Object.defineProperty({}, 'then', {
  get: function() {
    throw value;
  }
});
var promise = new Promise(function(resolve) {
  resolve(poisonedThen);
});

```

###### Backtrace:

```
ICE: Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed at jerryscript/jerry-core/jmem/jmem-heap.c(jmem_heap_finalize):181.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x00000000004809d6 in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:59
59	  SYSCALL_2
(gdb) bt
#0  0x00000000004809d6 in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:59
#1  0x000000000040ca61 in raise (sig=6) at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:98
#2  0x000000000040ca33 in abort () at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:81
#3  0x0000000000405d8c in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x000000000044371c in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x000000000044376e in jerry_assert_fail (assertion=0x487600 "JERRY_CONTEXT (jmem_heap_allocated_size) == 0", 
    file=0x487588 "jerryscript/jerry-core/jmem/jmem-heap.c", 
    function=0x494560 <__func__.3199.lto_priv.392> "jmem_heap_finalize", line=181) at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x0000000000442d33 in jmem_heap_finalize () at jerryscript/jerry-core/jmem/jmem-heap.c:181
[...]

PR #2115, opened 2017-11-20
Fix error-free issue in promise

jiangzidong:
Fix issue: #2107

JerryScript-DCO-1.0-Signed-off-by: Zidong Jiang zidong.jiang@intel.com

PR #2116, opened 2017-11-20
Fix bug in promise resolve handler

jiangzidong:
Should get the error_value if `then` prop is an error.

Fix issue: #2111


yichoi:
LGTM

jiangzidong:
@LaszloLango @yichoi Could you help review this one? It is very similar to #2115 (same root cause)

Issue #2105, opened 2017-11-16
Assertion 'ecma_is_value_empty (completion_value)' failed in vm_var_decl


renatahodovan:
###### Jerry version:

```
Checked revision: 7692aa9d
Build command: ./tools/build.py --clean --debug
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript

Object.freeze (this)
{
eval('{ function boo() {} }')
}
```

###### Backtrace:

```
ICE: Assertion 'ecma_is_value_empty (completion_value)' failed at jerryscript/jerry-core/vm/opcodes.c(vm_var_decl):58.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x0000000000475246 in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:59
59	  SYSCALL_2
(gdb) bt
#0  0x0000000000475246 in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:59
#1  0x000000000040ddfe in raise (sig=6) at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:98
#2  0x000000000040ddd0 in abort () at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:81
#3  0x0000000000407129 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x000000000044dc48 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x000000000044dc9a in jerry_assert_fail (assertion=0x475538 "ecma_is_value_empty (completion_value)", 
    file=0x475508 "jerryscript/jerry-core/vm/opcodes.c", function=0x48afc8 <__func__.3502.lto_priv.46> "vm_var_decl", line=58)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x0000000000401bc0 in vm_var_decl (frame_ctx_p=0x7fffffffd740, var_name_str_p=0x695a28 <jerry_global_heap+2024>)
    at jerryscript/jerry-core/vm/opcodes.c:58
#7  0x0000000000433630 in vm_init_loop (frame_ctx_p=0x7fffffffd740) at jerryscript/jerry-core/vm/vm.c:675
[...]

PR #2121, opened 2017-11-23
Fix undefined behaviour of global object freezing

DanielBallaSZTE:
Freezing the global object and later trying to add properties to it caused an assertion error (issue #2105).
This patch fixes the issue.


yichoi:
LGTM

Issue #2110, opened 2017-11-16
Assertion '((jmem_cpointer_t) uint_ptr) == uint_ptr' failed in jmem_decompress_pointer


renatahodovan:
###### Jerry version:

```
Checked revision: 7692aa9d
Build command: tools/build.py --clean --debug --profile=es2015-subset
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
var arrowFn = () => {};
arrowFn.hasOwnProperty('caller');
```

###### Backtrace:

```
ICE: Assertion '((jmem_cpointer_t) uint_ptr) == uint_ptr' failed at jerryscript/jerry-core/jmem/jmem-allocator.c(jmem_decompress_pointer):100.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0x00000000004809d6 in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:59
59	  SYSCALL_2
(gdb) bt
#0  0x00000000004809d6 in syscall_2 () at jerryscript/jerry-libc/target/posix/jerry-asm.S:59
#1  0x000000000040ca61 in raise (sig=6) at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:98
#2  0x000000000040ca33 in abort () at jerryscript/jerry-libc/target/posix/jerry-libc-target.c:81
#3  0x0000000000405d8c in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x000000000044371c in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x000000000044376e in jerry_assert_fail (assertion=0x4874a8 "((jmem_cpointer_t) uint_ptr) == uint_ptr", 
    file=0x487400 "jerryscript/jerry-core/jmem/jmem-allocator.c", 
    function=0x4944b0 <__func__.3201.lto_priv.432> "jmem_decompress_pointer", line=100)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x0000000000442b4e in jmem_decompress_pointer (compressed_pointer=3670022)
    at jerryscript/jerry-core/jmem/jmem-allocator.c:100
[...]

PR #2122, opened 2017-11-23
Add object type checking to instantiation

DanielBallaSZTE:
It wasn't properly checked if the given object was an arrow function, therefore it always got wrongly casted into `ecma_extended_object_t`.
Fixes issue #2110
*plus fixing a typo in the comments

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -730,7 +730,7 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
 } /* ecma_op_function_construct */
 
 /**
- * Lazy instantation of non-builtin ecma function object's properties
+ * Lazy instantiation of non-builtin ecma function object's properties
  *
  * Warning:
  *         Only non-configurable properties could be instantiated lazily in this function,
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -783,11 +783,23 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
...
+#ifndef CONFIG_DISABLE_ES2015_ARROW_FUNCTION
+    if (ecma_get_object_type (object_p) == ECMA_OBJECT_TYPE_ARROW_FUNCTION)
+    {
+      ecma_arrow_function_t *arrow_func_p = (ecma_arrow_function_t *) object_p;
+      bytecode_data_p = ECMA_GET_NON_NULL_POINTER (const ecma_compiled_code_t,
+                                                   arrow_func_p->bytecode_cp);
+    }
+    else
+    {
+#endif /* CONFIG_DISABLE_ES2015_ARROW_FUNCTION */
```
zherczeg: I don't really like when `{}` blocks are enclosed in ifs. It would be good to avoid it somehow. Even the original code could be duplicated here.

DanielBallaSZTE: Quite honestly I don't know which one of those two is the lesser evil, I'd stick with the current solution, as it contains less code duplication.

LaszloLango: IMHO, it is better in the current form.

PR #2123, opened 2017-11-24
Remove trivial ecma_number arithmetic functions

rerobika:
The affected function calls have been replaced with the appropriate arithmetic operands.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-number.c
+++ b/jerry-core/ecma/base/ecma-helpers-number.c
@@ -520,7 +520,7 @@ ecma_number_get_prev (ecma_number_t num) /**< ecma-number */
 
   if (ecma_number_is_negative (num))
   {
-    return ecma_number_negate (ecma_number_get_next (num));
+    return -ecma_number_get_next (num);
   }
 
   uint32_t biased_exp = ecma_number_get_biased_exponent_field (num);
@@ -555,7 +555,7 @@ ecma_number_get_next (ecma_number_t num) /**< ecma-number */
 
   if (ecma_number_is_negative (num))
   {
-    return ecma_number_negate (ecma_number_get_prev (num));
+    return -ecma_number_get_prev (num);
   }
 
   uint32_t biased_exp = ecma_number_get_biased_exponent_field (num);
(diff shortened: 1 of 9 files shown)


LaszloLango:
I don't see any improvement here. This is just a noise. I think these inlining are done by compiler automatically, because the binary size did not increase.

rerobika:
@LaszloLango I've updated the result with the run time standard deviation and highlighted the tests and results where the gain is magnitude greater then the std deviation. That's why I think it's more then noise.

rerobika:
Also I've compared the two binaries (with patch and without patch) and there is difference between them. It's means inlining these functions has real effect.

jerry-core/ecma/base/ecma-helpers-number.c
```diff
@@ -687,7 +687,7 @@ ecma_number_calc_remainder (ecma_number_t left_num, /**< left operand */
  *
  * @return number - result of addition.
  */
-ecma_number_t
+inline ecma_number_t __attr_always_inline___
```
zherczeg: Please just remove these functions.

rerobika:
@zherczeg Your suggestion have been applied!

jerry-core/vm/vm.c
```diff
@@ -1696,8 +1696,8 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
           if (ecma_is_value_float_number (left_value)
               && ecma_is_value_number (right_value))
           {
-            ecma_number_t new_value = ecma_number_add (ecma_get_float_from_value (left_value),
-                                                       ecma_get_number_from_value (right_value));
+            ecma_number_t new_value = ecma_get_float_from_value (left_value) +
+            ecma_get_number_from_value (right_value);
```
zherczeg: This is not nice alignment. What is the difference between float value and number value?

rerobika: Number value not specifies that the stored value is float or integer. Unfortunately it is not a nice alignment, but this is what vera++ requires.

zherczeg: I think you can use `()` brackets to help aligning

rerobika: It worked! Thanks for your advice!

PR #2126, opened 2017-11-28
Fix conversion warnings

robertsipka:
NuttX and artik053 build - compiling with strict Werror=conversion - fail when jerry-debugger option is enabled.
This patch based on #2007, because most of them are fixed earlier within that PR, but it was closed before the land.

Credit: Piotr Marcinkiewicz p.marcinkiew@samsung.com

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -1084,7 +1084,6 @@ jerry_create_boolean (bool value) /**< bool value from which a jerry_value_t wil
 {
   jerry_assert_api_available ();
 
-  value = jerry_get_arg_value (value);
   return jerry_return (ecma_make_boolean_value (value));
 } /* jerry_create_boolean */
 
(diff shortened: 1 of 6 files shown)


jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -110,6 +110,11 @@
     jmem_heap_free_block ((void *) utf8_ptr, utf8_str_size); \
   }
 
+/**
+ * Convert boolean to bitfield value.
+ */
+#define ECMA_CONVERT_BOOL_TO_FIELD(x) ((x) ? 1 : 0)
```
glistening: @robertsipka What about `ECMA_BOOL_TO_BITFIELD`? 
i) At least, for me, I can't understand what does it mean by `FIELD`.
ii) There is `ECMA_STRING_TO_UTF8_STRING` in same file. We can remove `CONVERT`.

robertsipka:
Good point, I've updated it accordingly, thanks.

jerry-core/api/jerry.c
```diff
@@ -1879,14 +1879,16 @@ jerry_define_own_property (const jerry_value_t obj_val, /**< object value */
 
   ecma_property_descriptor_t prop_desc = ecma_make_empty_property_descriptor ();
 
-  prop_desc.is_enumerable_defined = prop_desc_p->is_enumerable_defined;
-  prop_desc.is_enumerable = prop_desc_p->is_enumerable_defined ? prop_desc_p->is_enumerable : false;
+  prop_desc.is_enumerable_defined = ECMA_BOOL_TO_BITFIELD (prop_desc_p->is_enumerable_defined);
+  prop_desc.is_enumerable = ECMA_BOOL_TO_BITFIELD (prop_desc_p->is_enumerable_defined ?
+                                                        prop_desc_p->is_enumerable : false);
```
zherczeg: Wrong style.
```
prop_desc.is_enumerable = ECMA_BOOL_TO_BITFIELD (prop_desc_p->is_enumerable_defined ? prop_desc_p->is_enumerable
                                                                                    : false);
```

robertsipka: @zherczeg : Thanks, I've updated it.

jerry-core/api/jerry.c
```diff
@@ -1070,7 +1070,7 @@ jerry_create_boolean (bool value) /**< bool value from which a jerry_value_t wil
 {
   jerry_assert_api_available ();
 
-  value = jerry_get_arg_value (value);
+  value = jerry_get_arg_value (value) ? true : false;
```
LaszloLango: I think the correct fix is to remove only the original line.
`jerry_get_arg_value (jerry_value_t value)`, but it is called with a bool and not a `jerry_value_t`.

zherczeg: Good catch, I think this is incorrect. Probably should be removed.

robertsipka:
@LaszloLango , @zherczeg  : Thanks for the review, I've updated the PR.

zherczeg:
LGTM

PR #2131, opened 2017-11-30
Fix clearing of error reference

LaszloLango:
'ecma_clear_error_reference' must increase the reference of the returned
ecma value referenced by the error if there are more than one reference.


LaszloLango:
#2080 and #2096 might be related.

rerobika:
Just an observation. There is a typo 'ecama' in the  commit message.

martijnthe:
👍 lgtm

zherczeg:
good catch

PR #2136, opened 2017-12-04
Move _XOPEN_SOURCE macro to a more generic place

galpeter:
Building with all-in-one and enabled debugger caused
errors as the sleep functions were not defined.

By moving the _XOPEN_SOURCE macro to a generic place
we make sure that the required sleep functions are
present in every case.


LaszloLango:
@akosthekiss please check.

PR #2137, opened 2017-12-04
Fix the mbed Travis build

galpeter:
On Travis the Python (by default) is a bit old
and causes problems when the yotta is being used.

Switching to Python 2.7.13 solves half of the problem.

The other problem is the pyOpenSSL 17.5 and Linux
library incompatibilities. Thus we force the pyOpenSSL
to a pre-17.5 version.


yichoi:
LGTM

PR #2139, opened 2017-12-05
Force python2 when running test262

galpeter:
The test262 python script is not python 3 compatible
so we force it to be executed with python2.

Additionally print out the reason if the execution
failed.


yichoi:
LGTM

PR #2141, opened 2017-12-07
Optimize string concatenation.

zherczeg:
This patch adds two new string concatenation functions:
ecma_append_chars_to_string and ecma_append_magic_string_to_string

The former appends a magic string the and latter appends a cesu8 byte
array to a string. These two and the original ecma_concat_ecma_strings
free (dereference) their first argument which simplifies string handling
in most cases.


zherczeg:
Binary size: 137172 -> 136428. The rest is noise.

jerry-core/vm/opcodes-ecma-arithmetics.c
```diff
@@ -132,18 +132,29 @@ opfunc_addition (ecma_value_t left_value, /**< left value */
...
+    if (ECMA_IS_VALUE_ERROR (str_right_value))
+    {
+      ecma_deref_ecma_string (string1_p);
+      return str_right_value;
+    }
+
+    ecma_string_t *string2_p = ecma_get_string_from_value (str_right_value);
 
-    ret_value = ecma_make_string_value (concat_str_p);
+    string1_p = ecma_concat_ecma_strings (string1_p, string2_p);
```
glistening: @zherczeg I think `ecma_deref_ecma_string (string1_p)` is needed before assigning concat'd strings into `string1_p`.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-json.c
```diff
@@ -158,58 +147,43 @@ ecma_builtin_helper_json_create_separated_properties (ecma_collection_header_t *
  *         Returned value must be freed with ecma_free_value.
  */
 ecma_value_t
-ecma_builtin_helper_json_create_formatted_json (ecma_string_t *left_bracket_p, /**< left bracket*/
-                                                ecma_string_t *right_bracket_p, /**< right bracket*/
+ecma_builtin_helper_json_create_formatted_json (ecma_char_t left_bracket, /**< left bracket character */
```
LaszloLango: Why don't we use `lit_utf8_byte_t` for `left_bracket` and `right_bracket`? It looks like you are casting it anyway.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -1131,11 +1104,13 @@ ecma_builtin_string_prototype_object_replace_loop (ecma_builtin_replace_search_c
       {
         /* No more matches */
         ecma_string_t *appended_string_p;
+
+        ecma_ref_ecma_string (result_string_p);
```
LaszloLango: I don't see why is this ref needed here. Can you explain it?

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-json.c
zherczeg: The idea was that we don't reduce anything since byte/short/int consumes 1 machine register, and I added an assert. But if you prefer I can change it.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
zherczeg: There is an `ecma_deref_ecma_string (result_string_p);` below. Actually this part of the code should be rewritten as a whole, but I felt it could be done in a follow-up patch.

LaszloLango: Yes, it should be done in a follow-up patch. Thanks for the explanation.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-json.c
LaszloLango: IMHO the code will be a bit more readable that way. Please change it.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1337,8 +1400,8 @@ inline bool __attr_always_inline___
 ecma_compare_ecma_string_to_magic_id (const ecma_string_t *string_p, /**< property name */
                                       lit_magic_string_id_t id) /**< magic string id */
 {
-  return (ECMA_STRING_GET_CONTAINER (string_p) == ECMA_STRING_CONTAINER_MAGIC_STRING
-          && string_p->u.magic_string_id == id);
+  return (string_p->u.magic_string_id == id
```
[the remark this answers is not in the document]
zherczeg: I changed it to check the lower-chance part first (and it does not require `&` operation).

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -690,16 +645,124 @@ ecma_concat_ecma_strings (ecma_string_t *string1_p, /**< first ecma-string */
...
+    return string1_p;
+  }
+
+  const lit_utf8_byte_t *cesu8_string2_p;
+  lit_utf8_size_t cesu8_string2_size;
+  lit_utf8_size_t cesu8_string2_length;
+
+  lit_utf8_byte_t uint32_to_string_buffer[ECMA_MAX_CHARS_IN_STRINGIFIED_UINT32];
+
+  switch (ECMA_STRING_GET_CONTAINER (string2_p))
```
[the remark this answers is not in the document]
zherczeg: I don't think so, that is the hotspot part. There are key differences (e.g. setting `string1_rehash_needed`) and passing around a lot of pointers is costly.

PR #2125, opened 2017-11-28
Improve multiplication in vm

rerobika:
This patch improves the ecma_integers_values multiplication by checking if the multiplier or multiplicand is power of 2.
If it is it uses right shift instead of multiplication.

JerryScript-DCO-1.0-Signed-off-by: Robert Fancsik frobert@inf.u-szeged.hu

Benchmark | RSS (bytes) | Perf (sec) |
----: | ----: | ----: | 
3d-cube.js | 53248 -> 53248 : 0.000% | 0.824 -> 0.824 : +0.069% | 
3d-raytrace.js | 114102 -> 106496 : +6.666% | 1.063 -> 1.058 : +0.458% | 
access-binary-trees.js | 49152 -> 49152 : 0.000% | 0.565 -> 0.564 : +0.042% | 
access-fannkuch.js | 16384 -> 16384 : 0.000% | 2.134 -> 2.129 : +0.234% | 
access-nbody.js | 24576 -> 24576 : 0.000% | 1.131 -> 1.134 : -0.271% | 
bitops-3bit-bits-in-byte.js | 12288 -> 12288 : 0.000% | 0.592 -> 0.583 : +1.465% | 
bitops-bits-in-byte.js | 12288 -> 12288 : 0.000% | 0.789 -> 0.784 : +0.621% | 
bitops-bitwise-and.js | 12288 -> 12288 : 0.000% | 0.975 -> 0.971 : +0.370% | 
bitops-nsieve-bits.js | 110592 -> 110592 : 0.000% | 1.323 -> 1.319 : +0.339% | 
controlflow-recursive.js | 81920 -> 81920 : 0.000% | 0.387 -> 0.385 : +0.563% | 
crypto-aes.js | 61440 -> 61440 : 0.000% | 0.938 -> 0.938 : -0.034% | 
crypto-md5.js | 126976 -> 126976 : 0.000% | 0.644 -> 0.639 : +0.840% | 
crypto-sha1.js | 81920 -> 81920 : 0.000% | 0.643 -> 0.635 : +1.190% | 
date-format-tofte.js | 36864 -> 36864 : 0.000% | 0.726 -> 0.728 : -0.217% | 
date-format-xparb.js | 36864 -> 36864 : 0.000% | 0.410 -> 0.411 : -0.322% | 
math-cordic.js | 16384 -> 16384 : 0.000% | 1.255 -> 1.243 : +0.965% | 
[...]

PR #2142, opened 2017-12-07
Improve multiplication in vm

rerobika:
This patch improves the ecma_integers_values multiplication by checking if the multiplier or multiplicand is power of 2.
If it is it uses right shift instead of multiplication.


LaszloLango:
Note: this is a re-uploaded patch after a master revert. Related PR: #2125

yichoi:
sorry for my mistake. LGTM

PR #2144, opened 2017-12-08
Bugfix: add missing stdlib.h include for rand() usage in ecma-builtin-math.c

martijnthe:
`rand ()` is being used in `ecma-builtin-math.c`, but the `stdlib.h` header is not being included.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-math.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-math.c
@@ -14,6 +14,7 @@
  */
 
 #include <math.h>
+#include <stdlib.h>
 
 #include "ecma-alloc.h"
 #include "ecma-builtins.h"


jerry-core/ecma/builtin-objects/ecma-builtin-math.c
```diff
@@ -14,6 +14,7 @@
  */
 
 #include <math.h>
+#include <stdlib.h>
```
rerobika: If I'm right, `stdlib.h` comes from [this]( https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/jrt/jrt-libc-includes.h#L23.)  header, which will be included later.

martijnthe: Ah, you're right, thanks!
The problem is in our `stdlib.h` header, it's missing `rand`.

Issue #2143, opened 2017-12-08
Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed in ecma_free_string_list


renatahodovan:
###### Jerry version:

```
Checked revision: 90f2473f
Build command: tools/build.py --clean --debug --profile=es2015-subset
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
var a = new Uint8Array(2);
a[0] = { valueOf : function() { throw "intoint" ; } };

```

###### Backtrace:

```
ICE: Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed at jerryscript/jerry-core/ecma/base/ecma-literal-storage.c(ecma_free_string_list):43.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0xf7fd5db9 in __kernel_vsyscall ()
(gdb) bt
#0  0xf7fd5db9 in __kernel_vsyscall ()
#1  0xf7ca47e2 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf7ca5f51 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0x0835ed6e in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x0822aa4b in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x0822aac1 in jerry_assert_fail (assertion=0x839e880 <.str.2> "ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)", 
    file=0x839e7a0 <.str.1> "jerryscript/jerry-core/ecma/base/ecma-literal-storage.c", 
    function=0x839e8e0 <__func__.ecma_free_string_list> "ecma_free_string_list", line=43)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x081d36ba in ecma_free_string_list (string_list_p=0xf6300710)
    at jerryscript/jerry-core/ecma/base/ecma-literal-storage.c:43
#7  0x081d3432 in ecma_finalize_lit_storage () at jerryscript/jerry-core/ecma/base/ecma-literal-storage.c:60
[...]

PR #2147, opened 2017-12-12
[Bugfix] free error in typearray set value routine

jiangzidong:
Issue: #2143

### DIFF
--- a/jerry-core/ecma/operations/ecma-typedarray-object.c
+++ b/jerry-core/ecma/operations/ecma-typedarray-object.c
@@ -26,6 +26,7 @@
 #include "ecma-gc.h"
 #include "ecma-globals.h"
 #include "ecma-helpers.h"
+#include "jcontext.h"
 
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -897,6 +898,11 @@ ecma_op_typedarray_set_index_prop (ecma_object_t *obj_p, /**< a TypedArray objec
     return true;
   }
 
+  if (ECMA_IS_VALUE_ERROR (error))
```
LaszloLango: What about merging the two if statements like this?

```c
if (ECMA_IS_VALUE_ERROR (error))
{
  ecma_free_value (JERRY_CONTEXT (error_value));
  return false;
}

JERRY_ASSERT (ecma_is_value_empty (error));
return true;
```

jiangzidong: done

galpeter:
lgtm (the failed check is the mbed build which was moved to allowed failure group)

PR #2153, opened 2017-12-20
Rework ecma collection.

zherczeg:
Greatly simplify the iterator part and make it compatible with 32 bit cpointers.

It seems the patch has little effect on SunSpider. Rather it seems it pushes the binary in the runtime side (+0.3% overall), but increase the binary size by 200 bytes.


jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1023,26 +1024,22 @@ typedef double ecma_number_t;
...
-  /** Number of elements in the collection */
-  ecma_length_t unit_number;
-
-  /** Compressed pointer to first chunk with collection's data */
-  jmem_cpointer_t first_chunk_cp;
-
-  /** Compressed pointer to last chunk with collection's data */
-  jmem_cpointer_t last_chunk_cp;
+  jmem_cpointer_t first_chunk_cp; /**< compressed pointer to first chunk with collection's data */
+  jmem_cpointer_t last_chunk_cp; /** Compressed pointer to last chunk with collection's data */
```
dbatyai: Missing < in comment.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1023,26 +1024,22 @@ typedef double ecma_number_t;
...
+  ecma_length_t item_count; /**< number of items in the collection */
 } ecma_collection_header_t;
 
+/**
+ * String is not a valid array index.
```
dbatyai: Could you please explain this comment? Maybe I'm missing something, but it seems out of context.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1023,26 +1024,22 @@ typedef double ecma_number_t;
...
  * Description of non-first chunk in a collection's chain of chunks
  */
 typedef struct
 {
-  /** Characters */
-  lit_utf8_byte_t data[ sizeof (uint64_t) - sizeof (jmem_cpointer_t) ];
-
-  /** Compressed pointer to next chunk */
-  jmem_cpointer_t next_chunk_cp;
+  ecma_value_t items[ECMA_COLLECTION_CHUNK_ITEMS + 1];
```
dbatyai: Could you please add a comment here that clarifies why the +1 value is added and what purpose it serves.

zherczeg:
I made a micro benchmark of this patch (since SunSpider does not use for-in):

```
function test() {
  var a = [];
     a[i] = i;

  var sum = 0;
  for (j = 0; j < 1000; j++) {
      for (i in a)
          sum += i;
  }
}

test();
```

On x86-32, the old code runs in 1.5 sec, the new is 0.7 sec.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -74,6 +74,7 @@ typedef enum
   ECMA_TYPE_STRING = 2, /**< pointer to description of a string */
   ECMA_TYPE_OBJECT = 3, /**< pointer to description of an object */
   ECMA_TYPE_ERROR = 7, /**< pointer to description of an error reference */
+  ECMA_TYPE_COLLECTION_CHUNK = ECMA_TYPE_ERROR, /**< pointer to description of a collection chunk */
   ECMA_TYPE___MAX = ECMA_TYPE_ERROR /** highest value for ecma types */
```
LaszloLango: ECMA_TYPE___MAX and ECMA_TYPE_COLLECTION_CHUNK have the same value. Is this intentional? IMHO it is confusing.

zherczeg: Yes. The error is used only by the external API, while collection is used only internally. Currently I don't want to waste a type on collection.

Issue #2103, opened 2017-11-16
Outdated or incorrect build instructions for several targets


akosthekiss:
During the scripting of target build test automation for Travis CI (#2102), several issues have been identified:
- [ ] ESP8266:
  - The apt dependencies list was overly extensive (at least compared to what Travis CI had already installed on the worker image), sometimes even erroneous (e.g., didn't have to install any `libc6-dev*` packages and actually couldn't install `libtool-bin` because such a package was not available).
  - The creation of the `/opt/Espressif` directory is described together with the `dependencies` step although it belongs to the crosstool-NG building step. Moreover, it should be noted that neither the building of ct-ng nor the execution of ct-ng to build a cross toolchain requires elevated privileges (it can be done somewhere in the user's home directory, not under `/opt` only).
  - Building crosstool-NG didn't work for me exactly as described, `make` detected recursion and bailed out.
  - When building the firmware with JerryScript, why must `BIN_PATH` be specified? Couldn't it have at least a default to be the same as `BUILD_DIR`?
- [ ] Mbed OS 5:
  - The target needs heavy maintenance. As is, building an example app or a library from sources present under `targets/mbedos5` with the Makefile is not possible. `make all` errors because there is no `main` in the code base, `make library` errors because launcher sources require `jerry-targetjs.h` that are explicitly not generated for libraries. Then the readme suggests to do a gulp-based build, but that build system uses a pinned JerryScript version (https://github.com/ARMmbed/mbed-js-gulp/blob/master/main.js, `const JERRYSCRIPT_REVISION = '5cdb98c75ec231806677a8af0b16d508b1033eb2';`) not master. So that seems never building with latest jerry. The target should always be able to test latest developments.
- [ ] NuttX:
  - The readme doesn't mention which version of NuttX was last seen working. (7.22 worked for me, see makefile.)
  - The readme doesn't describe how configuration can be automated (without menuconfig that requires user interaction). Moreover, the readme suggests fetching the kconfig sources from a "3rd party" while the nuttx/tools repo contains its own kconfig sources.
- [ ] Particle: I couldn't get it to build at all.
- [ ] RIOT:
  - The readme doesn't pin the last known working version of RIOT. (2017.10 worked for me.)
- [ ] Tizen RT:
  - The readme doesn't pin the last known working version of Tizen RT. (1.1_Public_Release worked for me but only if I adapted the romfs patch.) 
[...]

PR #2154, opened 2017-12-20
Modify the build method of NuttX target.

rtakacs:
Currently, NuttX compiles the whole JerryScript (with the target app) by the targets/nuttx-stm32f4/Makefile. I think, this is not the best because there are many compile/feature-options (in build.py) that are not supported neither in Makefile nor in KConfig.

This development helps to build JerryScript separately from NuttX. In this case it's simple enough just build the target app (targets/nuttx-stm32f4/jerry-main.c) by NuttX and use the JerryScript's static libraries.

The compile/feature-options maintenance could be better because it doesn't need to duplicate them in the Makefile and in the KConfig files.

### DIFF
--- a/targets/nuttx-stm32f4/README.md
+++ b/targets/nuttx-stm32f4/README.md
@@ -5,20 +5,21 @@ This folder contains files to run JerryScript on
 
 ### How to build
 
-#### 1. Setting up the build environment for STM32F4-Discovery board
+#### 1. Setup the build environment for STM32F4-Discovery board
 
-Clone JerryScript and NuttX into jerry-nuttx directory
+Clone the necessary projects into a `jerry-nuttx` directory. The last tested working version of NuttX is 7.22.
 
-```
-mkdir jerry-nuttx
-cd jerry-nuttx
+```sh
+# Create a base folder for all the projects.
+mkdir jerry-nuttx && cd jerry-nuttx
+
 git clone https://github.com/jerryscript-project/jerryscript.git
-git clone https://bitbucket.org/nuttx/nuttx.git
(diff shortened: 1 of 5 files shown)


robertsipka:
I think it is a good approach to use CMake to build jerryscript. Could you please check #2103, and also update the NuttX related parts?

rtakacs:
@robertsipka Thanks for your comment. I've updated the README.md based on #2103 NuttX part.

targets/nuttx-stm32f4/README.md
```diff
@@ -30,66 +31,87 @@ jerry-nuttx
...
+##### 4.2 Enable JerryScript using kconfig-frontend
+
+`kconfig-frontend` could be useful if there are another options that should be enabled or disabled in NuttX.
 
 ```
-# assume you are in jerry-nuttx folder
 sudo apt-get install gperf flex bison libncurses-dev
+
+# Assuming you are in jerry-nuttx folder.
 git clone https://github.com/jameswalmsley/kconfig-frontends.git
```
robertsipka: As akosthekiss mentioned in #2103, the nuttx/tools repo contains its own kconfig sources. I think that cloning the `https://bitbucket.org/nuttx/tools.git` would be better.

rtakacs: @robertsipka Sorry, I miss that :) I will modify the README.md file.

zherczeg:
I think this can land after a few minor changes.

rtakacs:
@zherczeg @robertsipka Okay, I've updated the README.md with all the missing information that are mentioned in #2103.

robertsipka:
LGTM (informally)

PR #2157, opened 2018-01-02
Fix random number generation on ESP8266

rerobika:
This patch uses the onboard RTC for generating random seed. It also improves the print handler to support float values.
The rom segment is slightly increased to fit to the latest master.

### DIFF
--- a/targets/esp8266/user/jerry_extapi.c
+++ b/targets/esp8266/user/jerry_extapi.c
@@ -73,9 +73,20 @@ DELCARE_HANDLER(print) {
         printf("%s ", buffer);
         free (buffer);
       }
-      else
+      else if (jerry_value_is_number (args_p[cc]))
       {
-        printf ("(%d) ", args_p[cc]);
+        double number = jerry_get_number_value (args_p[cc]);
+        if ((int) number == number)
+        {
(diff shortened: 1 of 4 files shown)


targets/esp8266/user/jerry_extapi.c
```diff
@@ -73,9 +73,20 @@ DELCARE_HANDLER(print) {
...
         free (buffer);
       }
-      else
+      else if (jerry_value_is_number (args_p[cc]))
       {
-        printf ("(%d) ", args_p[cc]);
+        double number = jerry_get_number_value (args_p[cc]);
+        if ((int) number == number)
+        {
+          printf ("%d", number);
```
robertsipka: Does not this cause a warning here? Is it prints a correct number at all? Not just rubbish?

rerobika: I tested it, and it caused an issue. Thanks for your suggestion, I'm going to fix it.

rerobika: Fixed, now it works as intended.

robertsipka:
LGTM (informally)

yichoi:
LGTM

Issue #1832, opened 2017-05-18
C API for ArrayBuffer/TypedArray


martijnthe:
As per Gilles'/@barbibulle 's request on the mailing list:

>> Do you have any expectations of the interface? Which abilities are urgent in your application? Those feedbacks will help us design the API.
>
> what I'd expect from the public interface is to be able to:
> 1/ get to the underlying memory buffer of an ArrayBuffer and TypedArray, using something  equivalent to what's in ecma-arraybuffer-object.h and:
>  * ecma_arraybuffer_get_buffer (ecma_object_t *obj_p);
>  * ecma_arraybuffer_get_length (ecma_object_t *obj_p);
>  * ecma_typedarray_get_buffer (ecma_object_t *typedarray_p);
>  * ecma_typedarray_get_length (ecma_object_t *typedarray_p);
>  * ecma_typedarray_get_element_size_shift (ecma_object_t *typedarray_p);
>
> 2/ Ability to query whether a jerry_value_t is an instance of ArrayBuffer or is a TypedArray. Something equivalent to:
>  * ecma_is_arraybuffer (ecma_value_t val);
>  * ecma_is_typedarray (ecma_value_t target);
>
> 3/ Maybe also a convenience function for creating an ArrayBuffer with a single call (probably not necessary for TypedArrays). Something equivalent to:
>  * ecma_arraybuffer_new_object (ecma_length_t lengh);
>
> 4/ It would be nice to also have DataView implemented, since Uint32Array is always with the > platform's endianness, it makes it hard to parse external data structures that are big endian when running on little-endian platforms.
>
> When interfacing with native APIs, including filesystems, network interfaces, etc, ArrayBuffers are very convenient, and the ability to easily exchange buffers with JS code is extremely useful.

jiangzidong:
Rough plan for C API:

## ArrayBuffer

Though JS spec did provide buffer-access ability for ArrayBuffer, here, in native API, we provide them.

* `bool jerry_value_is_arraybuffer(jerry_value_t);`
* `jerry_value_t jerry_create_arraybuffer(jerry_length_t);`
* `jerry_length_t jerry_arraybuffer_set_buffer(uint8_t *buf_p, jerry_length_t buf_size); ` It will **copy** the content in buf_p into the buffer of arraybuffer. return the number of actually copied value. 
[...]

PR #2161, opened 2018-01-05
Introduce the Array Buffer C API

galpeter:
Add C API to work with Array Buffers.
The following methods are added:
- jerry_value_is_arraybuffer
- jerry_create_arraybuffer
- jerry_arraybuffer_write
- jerry_arraybuffer_read
- jerry_get_arraybuffer_byte_length

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -993,6 +993,44 @@ jerry_value_is_array (const jerry_value_t value)
 
 - [jerry_release_value](#jerry_release_value)
 
+## jerry_value_is_arraybuffer
+
+**Summary**
+
+Returns whether the given `jerry_value_t` is an ArrayBuffer object.
+
+**Prototype**
+
+```c
+bool
+jerry_value_is_arraybuffer (const jerry_value_t value)
+```
+
+- `value` - api value
+- return value
+  - true, if the given `jerry_value_t` is an ArrayBuffer object.
+  - false, otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t value;
+  ... // create or acquire value
+
+  if (jerry_value_is_arraybuffer (value))
+  {
+    ...
+  }
+
+  jerry_release_value (value);
+}
+```
+
+**See also**
+
+- [jerry_create_arraybuffer](#jerry_create_arraybuffer)
+
 
 ## jerry_value_is_boolean
 
@@ -2389,6 +2427,42 @@ jerry_create_array (uint32_t size);
 
 - [jerry_set_property_by_index](#jerry_set_property_by_index)
 - [jerry_get_property_by_index](#jerry_get_property_by_index)
+
+
+## jerry_create_arraybuffer
+
+**Summary**
+
+Create a jerry_value_t representing an ArrayBuffer object.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_arraybuffer (jerry_length_t size);
+```
+
+ - `size` - size of the ArrayBuffer to create **in bytes**
+ - return value - the new ArrayBuffer as a `jerry_value_t`
+
+**Example**
+
+```c
+{
+  jerry_value_t buffer_value = jerry_create_arraybuffer (15);
+
+  ... // use the ArrayBuffer
+
+  jerry_release_value (buffer_value);
+}
+```
+
+**See also**
+
+- [jerry_arraybuffer_read](#jerry_arraybuffer_read)
+- [jerry_arraybuffer_write](#jerry_arraybuffer_write)
+- [jerry_value_is_arraybuffer](#jerry_value_is_arraybuffer)
+- [jerry_release_value](#jerry_release_value)
 
 
 ## jerry_create_boolean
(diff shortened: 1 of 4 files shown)


galpeter:
As per #1832 I've created an API proposal for the Array Buffers.

Some notes:
* The `jerry_arraybuffer_map` call is a bit dangerous but useful in case we want to reduce memory usage with array buffers. I'm not sure we want this.
* I was wondering on a way to have the array buffer use an external (non-jerry heap memory) for which we could have the map operation. I'll try to dig around this a bit.
* The names of the functions was changed a bit, but maybe there are better names.

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,266 @@ main (void)
...
+**Summary**
+
+The function allows acces to the contents of the Array Buffer directly.
+
+WARNING! This operation is for expert use only! The programmer must
+ensure that the returned memory area is used correctly. That is
+there is no out of bounds reads or writes.
+
+After using the pointer the [jerry_arraybuffer_unmap][#jerry_arraybuffer_unmap)
+function must be called.
```
robertsipka: Wrong bracket

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,266 @@ main (void)
...
+- [jerry_arraybuffer_write](#jerry_arraybuffer_write)
+- [jerry_arraybuffer_unmap](#jerry_arraybuffer_unmap)
+
+
+## jerry_arraybuffer_unmap
+
+**Summary**
+
+Release the Array Buffer after a [jerry_arraybuffer_map][#jerry_arraybuffer_map)
+function call.
```
robertsipka: ditto

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,266 @@ main (void)
...
+- return value
+  - pointer to the Array Buffer's data area.
+  - 0 if the `value` is not an Array Buffer object.
+
+**Example**
+
+```c
+{
+  jerry_value_t buffer = jerry_create_arraybuffer (22);
+  uint8_t * const data = jerry_arraybuffer_map (buffer);
```
robertsipka: Put a space only before the asterisk character

jerry-core/api/jerry.c
```diff
@@ -2624,6 +2625,214 @@ jerry_set_vm_exec_stop_callback (jerry_vm_exec_stop_callback_t stop_cb, /**< per
...
+  if (!ecma_is_arraybuffer (buffer))
+  {
+    return NULL;
+  }
+
+  jerry_acquire_value (value);
+
+  ecma_object_t *buffer_p = ecma_get_object_from_value (buffer);
+  lit_utf8_byte_t *mem_buffer_p = ecma_arraybuffer_get_buffer (buffer_p);
+  return (uint8_t * const) mem_buffer_p;
```
robertsipka: ditto

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,266 @@ main (void)
...
+
+**Example**
+
+```c
+{
+  jerry_value_t array_buffer;
+  // acquire the Array Buffer from somewhere
+
+  uint8_t * const data = jerry_arraybuffer_map (array_buffer);
+
```
robertsipka: ditto

jerry-core/include/jerryscript-core.h
```diff
@@ -441,6 +441,24 @@ jerry_instance_t *jerry_create_instance (uint32_t heap_size, jerry_instance_allo
...
+jerry_length_t jerry_arraybuffer_write (const jerry_value_t value,
+                                        jerry_length_t offset,
+                                        const uint8_t *buf_p,
+                                        jerry_length_t buf_size);
+jerry_length_t jerry_arraybuffer_read (const jerry_value_t value,
+                                       jerry_length_t offset,
+                                       uint8_t *buf_p,
+                                       jerry_length_t buf_size);
+jerry_length_t jerry_arraybuffer_get_length (const jerry_value_t value);
+uint8_t * jerry_arraybuffer_map (const jerry_value_t value);
```
robertsipka: ditto

tests/unit-core/test-arraybuffer.c
```diff
@@ -0,0 +1,284 @@
...
+      "array.buffer");
+    jerry_value_t buffer = jerry_eval ((jerry_char_t *) eval_arraybuffer_src_p,
+                                        strlen (eval_arraybuffer_src_p),
+                                        true);
+
+    TEST_ASSERT (!jerry_value_has_error_flag (buffer));
+    TEST_ASSERT (jerry_value_is_arraybuffer (buffer));
+    TEST_ASSERT (jerry_arraybuffer_get_length (buffer) == 20);
+
+    uint8_t * const data = jerry_arraybuffer_map (buffer);
```
robertsipka: ditto

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,266 @@ main (void)
...
+  // fill the data with values
+  for (int i = 0; i < 20; i++)
+  {
+    data[i] = i * 2;
+  }
+
+  jerry_value_t buffer;
+  // ... create the Array Buffer or acuiqre it from somewhere.
+
```
robertsipka: *acquire

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,266 @@ main (void)
...
+- [jerry_create_arraybuffer](#jerry_create_arraybuffer)
+- [jerry_arraybuffer_write](#jerry_arraybuffer_write)
+- [jerry_arraybuffer_get_length](#jerry_arraybuffer_get_length)
+
+
+## jerry_arraybuffer_map
+
+**Summary**
+
+The function allows acces to the contents of the Array Buffer directly.
```
robertsipka: access

galpeter:
I've updated the PR with the requested changes. Also I've removed the `_map` and `_unmap` functions as they are high-risk functions and I'll will create another PR which introduces ArrayBuffers with external (non Jerry heap) memory usage.

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,170 @@ main (void)
...
+
+**Summary**
+
+Get the length of the Array Buffer in bytes.
+
+**Prototype**
+
+```c
+jerry_length_t
+jerry_arraybuffer_get_length (const jerry_value_t value);
```
zherczeg: There is a `jerry_get_array_length` function already. I am thinking whether this could be combined with it. Maybe just introducing a jerry_get_length() function which works with any special object (list in the comment what types it supports). But that could increase the binary size. I am unsure.

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,170 @@ main (void)
...
+
+**See also**
+- [jerry_create_arraybuffer](#jerry_create_arraybuffer)
+
+
+## jerry_arraybuffer_read
+
+**Summary**
+
+Read the contents of the Array Buffer into a buffer.
```
zherczeg: Copy the portion of the Array Buffer into a user provided buffer.

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,170 @@ main (void)
...
+- [jerry_create_arraybuffer](#jerry_create_arraybuffer)
+- [jerry_arraybuffer_write](#jerry_arraybuffer_write)
+- [jerry_arraybuffer_get_length](#jerry_arraybuffer_get_length)
+
+
+## jerry_arraybuffer_write
+
+**Summary**
+
+Write the contents of a buffer into the Array Buffer.
```
zherczeg: copy again

docs/02.API-REFERENCE.md
```diff
@@ -4679,3 +4753,170 @@ main (void)
...
+calculated in the following way: `min(array buffer length - offset, buf_size)`.
+
+**Prototype**
+
+```c
+jerry_length_t
+jerry_arraybuffer_write (const jerry_value_t value,
+                        jerry_length_t offset,
+                        const uint8_t *buf_p,
+                        jerry_length_t buf_size);
```
zherczeg: wrong alignment

jerry-core/api/jerry.c
```diff
@@ -2624,6 +2625,163 @@ jerry_set_vm_exec_stop_callback (jerry_vm_exec_stop_callback_t stop_cb, /**< per
...
+  jerry_value_t buffer = jerry_get_arg_value (value);
+
+  if (!ecma_is_arraybuffer (buffer))
+  {
+    return 0;
+  }
+
+  ecma_object_t *buffer_p = ecma_get_object_from_value (buffer);
+  jerry_length_t length = ecma_arraybuffer_get_length (buffer_p);
+  jerry_length_t copy_count = JERRY_MIN (length - offset, buf_size);
```
zherczeg: What happens it offset > length? In some range, it will be a big number, and the min mitigates the issue, but if offset is 0xffffffff that could be exploited.

I would add an early return if offset >= length.

jerry-core/api/jerry.c
```diff
@@ -2624,6 +2625,163 @@ jerry_set_vm_exec_stop_callback (jerry_vm_exec_stop_callback_t stop_cb, /**< per
...
```
zherczeg: offset issue again

docs/02.API-REFERENCE.md
galpeter: The `jerry_arraybuffer_get_length` returns the `byteLength` property (which is specified ad construction time) and the arraybuffer does not have a `length` property. I'm not really sure we should remove this function, but we could change the name to have `byte_length` suffix?

galpeter: Maybe a name like: `jerry_get_arraybuffer_byte_length` ? so we'll have a common naming scheme for "getters".

zherczeg: Hm, actually sounds good.

galpeter:
@zherczeg I've updated the PR with the requested changes, and also added two new api tests.

docs/02.API-REFERENCE.md
```diff
@@ -2391,6 +2429,42 @@ jerry_create_array (uint32_t size);
...
+Create a jerry_value_t representing an ArrayBuffer object.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_arraybuffer (jerry_length_t size);
+```
+
+ - `size` - size of the ArrayBuffer to create
```
zherczeg: byte size. I think we need to emphasize this.

tests/unit-core/test-arraybuffer.c
```diff
@@ -0,0 +1,269 @@
...
+{
+  JERRY_UNUSED (func_obj_val);
+  JERRY_UNUSED (this_val);
+
+  if (jerry_value_is_boolean (args_p[0])
+      && jerry_get_boolean_value (args_p[0]))
+  {
+    return jerry_create_boolean (true);
+  }
+  else
```
zherczeg: No need else after return.

galpeter:
@zherczeg  updated PR based on requests.

martijnthe:
Hey @galpeter, thanks for adding this. Looks great.
The one thing I'm missing is a non-copy read API (basically getting a pointer to the internal buffer).
The need for this was raised by multiple people in the discussion here:
https://github.com/jerryscript-project/jerryscript/issues/1832
Fine to add it later, I just wanted to bring it up again and make sure there's nothing in the way to do so later.

zherczeg:
Please check #2162 . The idea is that we allow arraybuffer to be created for external byte arrays. This allows communication between jerryscript instances, direct access for screen memory, etc. However there is no plan for accessing data of internally created arraybuffers.

martijnthe:
@zherczeg got it, makes sense, thanks!

galpeter:
@martijnthe initially I had that non-copy api which returned the internal buffer pointer, but that was to risky so I've changed the logic to external buffers which can be seen in #2162.

PR #2128, opened 2017-11-29
Update build guide for TizenRT 1.1

glistening:
Previous guide is out of date.
Updated guide works with TizenRT 1.1.

JerryScript-DCO-1.0-Signed-off-by: Sanggyu Lee sg5.lee@samsung.com

PR #2164, opened 2018-01-09
Modify the build method of TizenRT target

rtakacs:
Introduced a cmake/toolchain_mcu_artik053.cmake file that defines all the target specific compiler options.

Modified the Makefile.tizenrt to do not copy the created static libraries to the TizenRT folder. Instead, the application builder Makefile (tizenrt-artik053/apps/jerryscript/Makefile) copies the
required static libraries to TizenRT.

This modification helps to build JerryScript with the Python based build-script. In this case the user can use more build options.

```sh
tools/build.py \
    --clean \
    --lto=OFF \
    --jerry-cmdline=OFF \
    --all-in-one=OFF \
    --jerry-libc=OFF \
    --mem-heap=70 \
    --profile=es2015-subset \
    --toolchain=${PWD}/cmake/toolchain_mcu_artik053.cmake \
    --compile-flag="--sysroot=${PWD}/../TizenRT/os"
```

### DIFF
--- a/cmake/toolchain_mcu_artik053.cmake
+++ b/cmake/toolchain_mcu_artik053.cmake
@@ -12,27 +12,11 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-JERRYHEAP	?= 16
-BUILD_DIR	?= build
+set(CMAKE_SYSTEM_NAME TizenRT)
+set(CMAKE_SYSTEM_PROCESSOR armv7l)
+set(CMAKE_SYSTEM_VERSION ARTIK053)
 
-EXT_CFLAGS += -I. -isystem ../TizenRT/os/include
+set(FLAGS_COMMON_ARCH -mcpu=cortex-r4 -mfpu=vfpv3 -fno-builtin -fno-strict-aliasing -fomit-frame-pointer -fno-strength-reduce -Wall -Werror -Wshadow -Wno-error=conversion)
 
-.PHONY: libjerry clean
-
-all: libjerry
-
-libjerry:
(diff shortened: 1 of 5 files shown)


rtakacs:
Note: the patch also removes unused/unnecessary codes.

robertsipka:
Nice patch! Could you please also update the README with the followings:

1. Extend the description of the JerryScript build method with the Python based build-script.
2. akosthekiss mentioned a few request in #2103:
- The README suggest to checkout the `1.1_Public_Release` branch of the TizenRT. Please write there that it is the last known version (and the romfs patch is needed).
- The PPA-based GCC6 version works also, the GCC4.9 is a lower bound. Please update it accordingly.

LGTM (informally) after these minor fixes.

glistening:
@robertsipka I already updated for second item in #2128.

robertsipka:
@glistening : Indeed, I missed that, thanks.

rtakacs:
@robertsipka I've updated the README.md according to your requirements.

rtakacs:
@hs0225 Could you try out this PR on ARTIK053?

glistening:
@rtakacs I would like to specify the requirement of "trying out this PR by @hs0225"? Because perhaps you already build and test jerryscript with this PR on ARTIK053. Do you want a cross check?

rtakacs:
@glistening Yes, I've already tested this PR on ARTIK053 (just build and flash, and run jerry with no test). It worked for me, so maybe it's enough just to check the modifications.
(Travis is also green, so probably everything is OK :))

glistening:
(informally) LGTM

yichoi:
LGTM

PR #2166, opened 2018-01-16
Add version fields for debugger configuration

galpeter:
By adding version information to the debugger protocol
it is possible to report if the debugger client and server
have different expectations on debugger workings (opcodes, types, etc.).

### DIFF
--- a/jerry-core/debugger/debugger.h
+++ b/jerry-core/debugger/debugger.h
@@ -24,6 +24,11 @@
 /* JerryScript debugger protocol is a simplified version of RFC-6455 (WebSockets). */
 
 /**
+ * JerryScript debugger protocol version.
+ */
+#define JERRY_DEBUGGER_VERSION (1)
+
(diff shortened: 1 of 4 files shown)


jerry-core/debugger/debugger.h
```diff
@@ -23,6 +23,11 @@
 
 /* JerryScript debugger protocol is a simplified version of RFC-6455 (WebSockets). */
 
+/**
+ * JerryScript debugger protocol version.
+ */
+#define JERRY_DEBUGGER_VERSION (4)
```
zherczeg: Let's start with version 0.

galpeter: I would start with 1 if that's not a problem for you (zero is not really a version number imho).

robertsipka:
LGTM (informally)

yichoi:
LGTM

Issue #2134, opened 2017-12-04
Parsing Problem for Regex containing {} brackets


billyauhk:
I was trying to put a code compiled with Emscripten and run them on both node.js and iot.js, however iot.js fails to parse a particular regular expression. After tracing the issue I think JerryScript is not-compliant with other ECMA Script implementations.

The line given by Emscripten is (matching a JavaScript function):
`var sourceRegex = /^function\s*[a-zA-Z$_0-9]*\s*\(([^)]*)\)\s*{\s*([^*]*?)[\s;]*(?:return\s*(.*?)[;\s]*)?}$/;`
The minified code which produces a parsing error in jerryscript is:
`var sourceRegex = /{}/;`

On other JavaScript engines (like v8, node.js and the script console on my Firefox), the above minified code passed the parsing, while in JerryScript (and thus iot.js), it results in a parsing error.

Section 15.10 of ECMA262 seems to allow the above regular expression as the "{" (or "}" character) can be interpreted as an Atom instead of a start of QuantifierPrefix:
https://www.ecma-international.org/ecma-262/5.1/#sec-15.10

galpeter:
Hi!
I've checked the standard and for me it seems that the Atom should not accept the "{" character (as the PatternCharacter can be a SourceCharacter but not any of `^ $ \ . * + ? ( ) [ ] { } |`) thus the engine is ECMA compliant. Still that is true that other engines allow this extra element.

zherczeg:
ES6 maybe?

billyauhk:
Thanks @galpeter for a deeper look into the standard. I now agree that the current implementation is ECMA compliant (and other engines implement this as sort of extension).

(Re @zherczeg ) Seems ES6 also excluded "{" as an Atom (Atom -> PatternCharacter -> SourceCharacter but not SyntaxCharacter), and SyntaxCharacter is simply the same list as in ES5.1.

The standard tests for ECMA262 also use the escaped versions:
https://github.com/tc39/test262/blob/master/harness/nativeFunctionMatcher.js

However, I would like to know if such an extension could be provided upon giving JerryScript a setting (either in build-time or as a runtime argument) that provides the compatibility with other engines? This may up to active developers as JerryScript should be kept small to fit in the embedded systems.

Maybe the developers could:
1) Close the issue; or
2) Re-label this as a feature request.

galpeter:
[...]

PR #2169, opened 2018-01-22
Change '{' regex parsing to work similarly to other engines (#2134)

szledan:
Changed regex parsing to be able to handle opening braces
as other engines do, and also added a compile flag which
can disable this behaviour.

JerryScript-DCO-1.0-Signed-off-by: Szilard Ledan szledan@inf.u-szeged.hu


jerry-core/parser/regexp/re-parser.c
```diff
@@ -881,9 +881,35 @@ re_parse_next_token (re_parser_ctx_t *parser_ctx_p, /**< RegExp parser context *
     case LIT_CHAR_QUESTION:
     case LIT_CHAR_ASTERISK:
     case LIT_CHAR_PLUS:
+    {
+      return ecma_raise_syntax_error (ECMA_ERR_MSG ("Invalid RegExp token."));
+    }
     case LIT_CHAR_LEFT_BRACE:
     {
+#ifdef ENABLE_REGEXP_STRICT_MODE
```
galpeter: What if we switch around the logic and do something like this?
```c
    case LIT_CHAR_PLUS:
    case LIT_CHAR_LEFT_BRACE:
    {
#ifndef ENABLE_REGEXP_STRICT_MODE
      const lit_utf8_byte_t *input_curr_p = parser_ctx_p->input_curr_p;

      lit_utf8_decr (&parser_ctx_p->input_curr_p);
      if (!ecma_is_value_empty (re_parse_iterator (parser_ctx_p, out_token_p)))
      {
          parser_ctx_p->input_curr_p = input_curr_p;

          out_token_p->type = RE_TOK_CHAR;
          out_token_p->value = ch;
          ret_value = re_parse_iterator (parser_ctx_p, out_token_p);

          if (!ecma_is_value_empty (ret_value))
          {
            parser_ctx_p->input_curr_p = input_curr_p;
            ret_value = ECMA_VALUE_EMPTY;
          }
          break;
      }
#else /* ENABLE_REGEXP_STRICT_MODE */
      return ecma_raise_syntax_error (ECMA_ERR_MSG ("Invalid RegExp token."));
#endif /* !ENABLE_REGEXP_STRICT_MODE */
```
so we can remove the 884-886 line additions

szledan: I think we can't remove these lines, because this patch handles only the `{` braces issue and doesn't catch other iterator characters.

galpeter: ahhh yeah! I was to eager to simplify.

jerry-core/CMakeLists.txt
```diff
@@ -17,23 +17,24 @@ set(JERRY_CORE_NAME jerry-core)
...
+set(FEATURE_CPOINTER_32_BIT    OFF     CACHE BOOL   "Enable 32 bit compressed pointers?")
+set(FEATURE_DEBUGGER           OFF     CACHE BOOL   "Enable JerryScript debugger?")
+set(FEATURE_ERROR_MESSAGES     OFF     CACHE BOOL   "Enable error messages?")
+set(FEATURE_EXTERNAL_CONTEXT   OFF     CACHE BOOL   "Enable external context?")
+set(FEATURE_JS_PARSER          ON      CACHE BOOL   "Enable js-parser?")
+set(FEATURE_MEM_STATS          OFF     CACHE BOOL   "Enable memory statistics?")
+set(FEATURE_MEM_STRESS_TEST    OFF     CACHE BOOL   "Enable mem-stress test?")
+set(FEATURE_PARSER_DUMP        OFF     CACHE BOOL   "Enable parser byte-code dumps?")
+set(FEATURE_PROFILE            "es5.1" CACHE STRING "Use default or other profile?")
+set(FEATURE_REGEXP_STRICT_MODE OFF     CACHE BOOL   "Enable regexp strict mode?")
```
zherczeg: Do we change the current behaviour here? Personally I would prefer FEATURE_REGEXP_COMPATIBILITY_MODE, and keep it OFF by default.

szledan: Yes, now the compatibility mode is the default, so the regexp engine accepts this `/{}/` regex pattern while the strict mode does not.
Shall I set the strict mode as default?
(Personally I would prefer the extended behaviour as default, like other engines.)

zherczeg: I would like to hear others opinion (kind of voting). I usually prefer to keep the current behaviour but I am not that strict.

LaszloLango: I don't have strong preferences. I think the average user won't notice the compatibility features only if they dig into the build system. So enable compatibility mode by default sounds reasonable for me.

robertsipka: I agree with @szledan and @LaszloLango, the compatibility features should be enabled by default, but I would change the name also. So my vote is to set the `FEATURE_REGEXP_COMPATIBILITY_MODE` to `ON` instead of setting the `FEATURE_REGEXP_STRICT_MODE` to `OFF`.

robertsipka: Sorry, I rethought it, the `FEATURE_REGEXP_STRICT_MODE` with `OFF` is fine. The other version may cause problems/different behaviors if somebody want to compile without CMake. LGTM (informally)

PR #2171, opened 2018-01-23
Print the return value of the engine in case of TizenRT.

rtakacs:
Since TizenRT doesn't have any commands to get the return value of a process,
added a debug print line that shows the return value.

Similar solution was introduced in IoT.js ([iotjs_main_tizenrt.c#L167](https://github.com/Samsung/iotjs/blob/master/src/platform/tizenrt/iotjs_main_tizenrt.c#L167))

### DIFF
--- a/targets/tizenrt-artik053/apps/jerryscript/jerry_main.c
+++ b/targets/tizenrt-artik053/apps/jerryscript/jerry_main.c
@@ -483,6 +483,21 @@ jerry_cmd_main (int argc, char *argv[])
 } /* jerry_cmd_main */
 
 /**
+ * Run JerryScript and print its return value.
+ */
+static int
+jerry(int argc, char *argv[])
+{
+  int ret_code = jerry_cmd_main(argc, argv);
+
+#ifdef CONFIG_DEBUG_VERBOSE
+  jerry_port_log(JERRY_LOG_LEVEL_DEBUG, "JerryScript result: %d\n", ret_code);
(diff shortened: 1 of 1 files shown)


robertsipka:
LGTM (informally)

targets/tizenrt-artik053/apps/jerryscript/jerry_main.c
```diff
@@ -482,6 +482,21 @@ jerry_cmd_main (int argc, char *argv[])
...
+/**
+ * Run JerryScript and print its return value.
+ */
+static int
+jerry(int argc, char *argv[])
+{
+  int ret_code = jerry_cmd_main(argc, argv);
+
+#ifdef CONFIG_DEBUG_VERBOSE
+  printf("JerryScript Result: %d\n", ret_code);
```
zherczeg: lowercase "result"

glistening: @rtakacs What about [jerry_port_log](https://github.com/jerryscript-project/jerryscript/blob/master/targets/tizenrt-artik053/apps/jerryscript/jerry_main.c#L497), insted of `printf`?

rtakacs: @glistening Good idea, thanks.
@zherczeg Fixed.

yichoi:
LGTM

Issue #2173, opened 2018-01-24
Is there a jerry_string_to_char_buffer() that will fill a buffer, even if jerry string is too big?


mofosyne:
Specifically this check `(ecma_string_get_size (str_p) > buffer_size)` is exiting as expected if the jerry value is larger than the buffer.

However it would be nice if there was another function that would be more tolerant of too big jerry string, by simply dropping the extra characters.

I'm not too sure on the best option of dealing with this. Should this be a seperate function like `jerry_string_to_char_buffer_unsafe()`?

Note: If you are approaching this, then also check if ecma_string_copy_to_cesu8_buffer() actually checks buffer_size(), as memcpy uses the jerry string size rather than declared buffer size (or the smallest of either). 

https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/api/jerry.c#L1489

mofosyne:
The other alternative is a way to change the length of the jerry string temporarily (without copying the value), before placing it into jerry_string_to_char_buffer(). However I'm not aware of the best approach to do this.

mofosyne:
Hmmm... I solved my issue by using jerry_substring_to_char_buffer() using this:

```
  // Copy to buffer
  message_size = jerry_substring_to_char_buffer (message,                               /**< input string value */
                                                0,                                      /**< position of the first character */
                                                jerry_get_string_size(message),      /**< position of the last character */
                                                (unsigned char *)logger->temp_buffer,   /**< [out] output characters buffer */
                                                (sizeof(logger->temp_buffer) - 1)       /**< size of output buffer */
                                                );

```

We can close this ticket. But you may want to consider if a seperate function should be included to encourage usage if anyone needs a permissive version of jerry_string_to_char_buffer()

e.g.

```
jerry_size_t
jerry_string_to_char_buffer_permissive (const jerry_value_t value, /**< input string value */
                                jerry_char_t *buffer_p, /**< [out] output characters buffer */
                                jerry_size_t buffer_size) /**< size of output buffer */
{
    return jerry_substring_to_char_buffer (value, 0, jerry_get_string_size(message), buffer_p, buffer_size);
}
```

LaszloLango:
@mofosyne as you wrote `jerry_substring_to_char_buffer` should be used in your use case.

mofosyne:
[...]

PR #2174, opened 2018-01-24
Update documentation for jerry string to buffer function with suggestions to use substring

mofosyne:
In some use cases, you want to reliably copy jerry strings to buffer, even if it doesn't fit target buffer, but is acceptable to lose some bytes. In those cases, the documentation will now suggest using the substring function as an alternative instead.

https://github.com/jerryscript-project/jerryscript/issues/2173

JerryScript-DCO-1.0-Signed-off-by: Brian Khuu mofosyne@gmail.com

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -1531,13 +1531,13 @@ jerry_value_set_error_flag (jerry_value_t *value_p);
 If the input value is an error value, then return a new reference to its referenced value.
 Otherwise, return a new reference to the value itself.
 
-*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) 
+*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value)
 when it is no longer needed.
 
 **Prototype**
 
 ```c
-jerry_value_t 
+jerry_value_t
 jerry_get_value_without_error_flag (jerry_value_t value)
 ```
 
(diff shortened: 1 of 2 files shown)


yichoi:
@mofosyne write proper titles and descriptions

mofosyne:
Will this work?

LaszloLango:
@mofosyne please fix the sign off in the commit message:

> Signed-off-by message is incorrect. The following line should be at the end of the 79929df commit's message: 'JerryScript-DCO-1.0-Signed-off-by: Brian Khuu mofosyne@gmail.com'.

LaszloLango:
@mofosyne you updated the title and description on GitHub only. Please update the commit message accordingly and (force) update your branch.

mofosyne:
Hmmmm... ah so now the commit build check is fixed then.

docs/02.API-REFERENCE.md
```diff
@@ -1824,6 +1824,11 @@ enough for the whole string.
 *Note*: Does not put '\0' to the end of string, the return value identifies
 the number of valid bytes in the output buffer.
 
+*Note*: If jerry value is larger than target buffer, then it will fail to copy.
```
zherczeg: I would say: `If the size of the string in jerry value is larger than the size of the target buffer, the copy will fail.`

docs/02.API-REFERENCE.md
```diff
@@ -1824,6 +1824,11 @@ enough for the whole string.
...
+Use [jerry_substring_to_char_buffer](#jerry_substring_to_char_buffer) instead if
```
zherczeg: `To copy a substring the [jerry_substring_to_char_buffer](#jerry_substring_to_char_buffer) API function is recommended instead.`

zherczeg: Please also fix the other comments.

mofosyne:
adjusted

zherczeg:
One more thing, please squash the commits.

mofosyne:
Oh my bad. I though `git commit --amend` should have squashed it.

mofosyne:
Should be ready now

PR #2176, opened 2018-01-25
Add support for aborts.

zherczeg:
Aborts are similar to exceptions except they are not caught by catch
and finally blocks. Callbacks should honor aborts as well and return
them without processing them. Aborts are never thrown by JavaScript
code.

In the future certain events such as out-of-memory condition may
also throw aborts.

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -82,10 +82,10 @@ static const char * const wrong_args_msg_p = "wrong type of argument";
  * Assert that it is correct to call API in current state.
  *
  * Note:
- *         By convention, there can be some states when API could not be invoked.
+ *         By convention, there are some states when API could not be invoked.
  *
- *         While, API can be invoked jerry_api_available flag is set,
- *         and while it is incorrect to invoke API - it is not set.
+ *         The API can be and only be invoked when the ECMA_STATUS_API_AVAILABLE
+ *         flag is set.
  *
  *         This procedure checks whether the API is available, and terminates
  *         the engine if it is unavailable. Otherwise it is a no-op.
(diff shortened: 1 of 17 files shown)


tests/unit-core/test-exceptions.c
```diff
@@ -0,0 +1,175 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
```
martijnthe: +1 to splitting the huge test-api.c into smaller test files.

I'd go even further and split this into test-abort.c and test-exec-stop.c.
At a glance they look like independent tests, no?

martijnthe:
@zherczeg could you elaborate a bit on when one would use this new API vs the existing `jerry_port_fatal`?

tests/unit-core/test-exceptions.c
zherczeg: For me that is a bit too much. Others?

zherczeg:
On embedded systems. Currently a jerry_abort or exit leaves too much resources open. Especially this is a problem for normal exits. Also the debugger should be able to "reset" a device by throwing such error.

martijnthe:
Thanks for clarifying.

jerry-core/api/jerry.c
```diff
@@ -846,10 +867,24 @@ jerry_value_set_error_flag (jerry_value_t *value_p)
...
 
+/**
+ * Set both the abort and error flags if the value is not an error reference.
+ */
+void
+jerry_value_set_abort_flag (jerry_value_t *value_p)
+{
+  jerry_assert_api_available ();
+
+  if (!ecma_is_value_error_reference (*value_p))
```
LaszloLango: What happens if a reference error. Shouldn't we set the abort flag in such cases as well?

zherczeg: It is possible, but the description says it ignores error references. Shall we change it?

LaszloLango: I would change it.

zherczeg:
I did the requested changes.

PR #2177, opened 2018-01-25
Introduce C API to query the type of an Error object

galpeter:
New api function:
* jerry_get_error_type

Additionally update a few places where this new function can be used.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -23,6 +23,9 @@ Possible types of an error:
  - JERRY_ERROR_TYPE - type error
  - JERRY_ERROR_URI - URI error
 
+There is also a special value `JERRY_ERROR_NONE` which is not an error type
+this value can only be returned by the [jerry_get_error_type](#jerry_get_error_type).
+
 ## jerry_feature_t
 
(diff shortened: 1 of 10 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -2555,6 +2601,11 @@ jerry_create_boolean (bool value);
 
 Create new JavaScript error object.
 
+Important! The `error_type` argument *must not be*
+`JERRY_ERROR_NONE` as it does not make sense.
```
zherczeg: "as it does not make sense." - delete this part

jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -30,10 +30,36 @@
  * \addtogroup exceptions Exceptions
  * @{
  */
+typedef struct
+{
+  ecma_standard_error_t error_type;
+  ecma_builtin_id_t error_prototype_id;
+} ecma_error_mapping_t;
+
+ecma_error_mapping_t ecma_error_mappings[] =
```
zherczeg: Make this const.

galpeter:
@zherczeg Fixed & rebased.

PR #2187, opened 2018-01-31
Make source scrolling command in python debugger

DanielBallaSZTE:
Source is now scrollable after writing `scroll`.
Keys are `q` to quit, `w` to scroll up `s` to scroll down.

### DIFF
--- a/jerry-debugger/jerry-client-ws.py
+++ b/jerry-debugger/jerry-client-ws.py
@@ -24,6 +24,7 @@ import select
 import socket
 import struct
 import sys
+import math
 
 # Messages sent by the server to client.
 JERRY_DEBUGGER_CONFIGURATION = 1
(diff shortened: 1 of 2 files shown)


jerry-debugger/jerry-client-ws.py
```diff
@@ -889,6 +915,17 @@ def print_source(debugger, line_num):
...
+            if start + offset < 0:
+                debugger.src_offset += debugger.src_offset_diff
+                offset += debugger.src_offset_diff
+            elif end + offset > len(last_bp.function.source):
+                debugger.src_offset -= debugger.src_offset_diff
+                offset -= debugger.src_offset_diff
+
+            start = max(start + offset, 0)
+            end = min(end + offset, len(last_bp.function.source))
+
```
zherczeg: I think one newline is enough here.

DanielBallaSZTE: Yes, you are right, thanks :)

PR #2191, opened 2018-02-01
Check if there are any characters left after an escape sequence

DanielBallaSZTE:
Fixes #2180, #2192 

JerryScript-DCO-1.0-Signed-off-by: Robert Fancsik frobert@inf.u-szeged.hu
JerryScript-DCO-1.0-Signed-off-by: Daniel Balla dballa@inf.u-szeged.hu

Issue #2192, opened 2018-02-01
heap-buffer-overflow in lit_read_code_unit_from_hex


renatahodovan:
###### Jerry version:

```
Checked revision: 918eb22a
Build command: ./tools/build.py --clean --debug --compile-flag=-fsanitize=address --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --compile-flag=-g --jerry-libc=off --static-link=off --strip=off --system-allocator=on --linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
JSON.parse('"' + '\\u');
```

The test case is quite similar to #2180 but the failure happens in a different module.
The place and the type of the failure is the same as in #2140 but the test cases and the traces are completely different.

###### Backtrace:

```
=================================================================
==15239==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xf5f0065b at pc 0x0822d726 bp 0xff814e78 sp 0xff814e6c
READ of size 1 at 0xf5f0065b thread T0
    #0 0x822d725 in lit_read_code_unit_from_hex jerryscript/jerry-core/lit/lit-char-helpers.c:443:9
    #1 0x82b7ae9 in ecma_builtin_json_parse_string jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:181:17
    #2 0x82b5cf3 in ecma_builtin_json_parse_next_token jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:435:9
    #3 0x82b477d in ecma_builtin_json_parse_value jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:550:3
    #4 0x82b1ac7 in ecma_builtin_json_parse jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:822:31
    #5 0x82b12a8 in ecma_builtin_json_dispatch_routine jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.inc.h:26:1
    #6 0x81ecc14 in ecma_builtin_dispatch_routine jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.inc.h:135:1
    #7 0x81ebb55 in ecma_builtin_dispatch_call jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.c:844:17
[...]

Issue #2180, opened 2018-01-30
heap-buffer-overflow in ecma_builtin_json_parse_string


renatahodovan:
###### Jerry version:

```
Checked revision: 918eb22a
Build command: ./tools/build.py --clean --debug --compile-flag=-fsanitize=address --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --compile-flag=-g --jerry-libc=off --static-link=off --strip=off --system-allocator=on --linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```
###### OS:

```
Ubuntu 17.10
```
###### Test case:

```javascript
JSON.parse('"' + '\\');
```

###### Backtrace:

```
=================================================================
==20840==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xf5d0065a at pc 0x5665f4ca bp 0xffa22e68 sp 0xffa22e58
READ of size 1 at 0xf5d0065a thread T0
    #0 0x5665f4c9 in ecma_builtin_json_parse_string jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:164
    #1 0x566607e9 in ecma_builtin_json_parse_next_token jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:435
    #2 0x56660d80 in ecma_builtin_json_parse_value jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:550
    #3 0x56661acf in ecma_builtin_json_parse jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.c:822
    #4 0x5665ef92 in ecma_builtin_json_dispatch_routine jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-json.inc.h:26
    #5 0x566c6de5 in ecma_builtin_dispatch_routine jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.inc.h:135
    #6 0x566c74d1 in ecma_builtin_dispatch_call jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.c:844
    #7 0x566d06a9 in ecma_op_function_call jerryscript/jerry-core/ecma/operations/ecma-function-object.c:342
    #8 0x5667697e in opfunc_call jerryscript/jerry-core/vm/vm.c:425
    #9 0x56683328 in vm_execute jerryscript/jerry-core/vm/vm.c:2871
[...]

PR #2194, opened 2018-02-02
Avoid using an unitialized variable

robertsipka:
@DanielBallaSZTE : This bug is appeared after your PR (#2191). Please check that this patch covers your original intent.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-json.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-json.c
@@ -183,7 +183,7 @@ ecma_builtin_json_parse_string (ecma_json_token_t *token_p) /**< token argument 
         case LIT_CHAR_LOWERCASE_U:
         {
           ecma_char_t code_unit;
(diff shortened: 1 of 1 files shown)


zherczeg:
I think the issue is not "avoiding an uninitialized variable", but the code was simply wrong. So please change the description. Also pleas add test case since it seems this is not tested.

robertsipka:
I think both. The code was wrong, but also allowed that if the first part of the condition is false then the uninitialized value of the `code_unit` is used. Besides this I agree with you, I will rename it, and add test cases also.

zherczeg:
Please add a test case. I think an eval should do it.

robertsipka:
@zherczeg : Sorry for the late response. I've looked over it and actually I do not know what to check that we have not checked yet. The `jerry` returns with SyntaxError in case of invalid unicode escape sequence, which is an expected behaviour. In #2191, a test case has already added which covers the #2180 and #2192 issues, but it was a wrong and a partial fix, the problem is still exist in some cases without this patch, e.g. the following input : `JSON.parse('"' + '\\u23');` will cause the same issue in the current master. Should I add another test with the input above?

robertsipka:
#2201 also fixes this bug, so I am closing this PR.

PR #2195, opened 2018-02-02
Add C API to query the type of a JS value

galpeter:
New API function:
 * jerry_value_get_type

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -767,6 +767,60 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
 } /* jerry_value_is_undefined */
 
 /**
+ * Perform the base type of the JavaScript value.
+ *
+ * @return jerry_type_t value
+ */
+jerry_type_t
+jerry_value_get_type (const jerry_value_t value) /**< input value to check */
+{
+  jerry_assert_api_available ();
+
+  jerry_value_t argument = jerry_get_arg_value (value);
+  lit_magic_string_id_t lit_id = ecma_get_typeof_lit_id (argument);
+
+  JERRY_ASSERT (lit_id != LIT_MAGIC_STRING__EMPTY);
+
+  switch (lit_id)
+  {
+    case LIT_MAGIC_STRING_UNDEFINED:
+    {
+      return JERRY_TYPE_UNDEFINED;
+    }
+    case LIT_MAGIC_STRING_BOOLEAN:
+    {
+      return JERRY_TYPE_BOOLEAN;
+    }
+    case LIT_MAGIC_STRING_NUMBER:
+    {
+      return JERRY_TYPE_NUMBER;
+    }
+    case LIT_MAGIC_STRING_STRING:
+    {
+      return JERRY_TYPE_STRING;
+    }
+    case LIT_MAGIC_STRING_FUNCTION:
+    {
+      return JERRY_TYPE_FUNCTION;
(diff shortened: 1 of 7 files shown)


jerry-core/api/jerry.c
```diff
@@ -764,6 +764,35 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
...
+  jerry_assert_api_available ();
+
+  jerry_type_t return_type;
+  lit_magic_string_id_t lit_id = ecma_get_typeof_lit_id (jerry_get_arg_value (value));
+
+  JERRY_ASSERT (lit_id != LIT_MAGIC_STRING__EMPTY);
+
+  switch (lit_id)
+  {
+    case LIT_MAGIC_STRING_UNDEFINED: return_type = JERRY_TYPE_UNDEFINED;  break;
```
LaszloLango: What about JERRY_TYPE_NULL?

galpeter: As the ecma 262 5.1 reference describes the NULL in JS as object. However, I'm in favour of having a JERRY_TYPE_NULL, but then the typeof name could be a bit misleading.

LaszloLango: I see. Leave as it is.

galpeter:
I'll change this to have a bit different name and meaning than the 'typeof' operation to allow the null type to be returned. Stay tuned.

galpeter:
@LaszloLango I've updated the PR

jerry-core/api/jerry.c
```diff
@@ -766,6 +766,53 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
...
+
+  jerry_type_t return_type;
+  jerry_value_t argument = jerry_get_arg_value (value);
+  lit_magic_string_id_t lit_id = ecma_get_typeof_lit_id (argument);
+
+  JERRY_ASSERT (lit_id != LIT_MAGIC_STRING__EMPTY);
+
+  switch (lit_id)
+  {
+    case LIT_MAGIC_STRING_UNDEFINED:
```
zherczeg: You have to put `{}` around cases, as the style checker also complains about it.

jerry-core/api/jerry.c
```diff
@@ -766,6 +766,63 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
...
+    {
+      return_type = JERRY_TYPE_UNDEFINED;
+      break;
```
zherczeg: please use return

galpeter:
@zherczeg I've updated the PR, please check.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -187,6 +187,7 @@ void ecma_value_assign_number (ecma_value_t *value_p, ecma_number_t ecma_number)
 void ecma_free_value (ecma_value_t value);
 void ecma_fast_free_value (ecma_value_t value);
 void ecma_free_value_if_not_object (ecma_value_t value);
+lit_magic_string_id_t ecma_get_typeof_lit_id (ecma_value_t value);
```
zherczeg: One more minor thing: since this is not typeof, lets rename this function to `ecma_get_value_type_lit_id`.

galpeter: This function implements the typeof function as this was extracted from the op_typeof call. I'm using this in both places.

Issue #2200, opened 2018-02-10
Heap buffer overflow while parsing malformed JSON hex escape sequence w/o closing quote


martijnthe:
Jerry version:

Checked revision: `e8608707b6d9486022b4d72b280303923945fdeb`
Build command: `./tools/build.py --clean --debug --compile-flag=-fsanitize=address --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --compile-flag=-g --jerry-libc=off --static-link=off --strip=off --system-allocator=on --error-messages=on --profile=es2015-subset`

OS:

macOS 10.12.6 (16G1212)

Test case:

```
JSON.parse('"\\ubad');
```

Backtrace:

```
==8770==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x0240155e at pc 0x001850dc bp 0xbffb4418 sp 0xbffb4414
READ of size 1 at 0x0240155e thread T0
    #0 0x1850db in lit_read_code_unit_from_hex lit-char-helpers.c:443
    #1 0x1e5412 in ecma_builtin_json_parse_string ecma-builtin-json.c:195
    #2 0x1e3821 in ecma_builtin_json_parse_next_token ecma-builtin-json.c:450
    #3 0x1e1fcb in ecma_builtin_json_parse_value ecma-builtin-json.c:565
    #4 0x10cdce in ecma_builtin_json_parse ecma-builtin-json.c:837
    #5 0x10c5cb in ecma_builtin_json_dispatch_routine ecma-builtin-json.inc.h:26
    #6 0x13d396 in ecma_builtin_dispatch_routine ecma-builtins.inc.h:135
    #7 0x13caff in ecma_builtin_dispatch_call ecma-builtins.c:844
    #8 0x7e3af in ecma_op_function_call ecma-function-object.c:342
    #9 0x22181f in opfunc_call vm.c:425
    #10 0x1dbd77 in vm_execute vm.c:2871
    #11 0x1572c4 in vm_run vm.c:2951
    #12 0x7e8df in ecma_op_function_call ecma-function-object.c:405
    #13 0x22181f in opfunc_call vm.c:425
    #14 0x1dbd77 in vm_execute vm.c:2871
    #15 0x1572c4 in vm_run vm.c:2951
    #16 0x55f82 in vm_run_global vm.c:232
    #17 0x53dc1 in jerry_run jerry.c:562
    #18 0x4d58f in main main-unix.c:611
[...]

PR #2201, opened 2018-02-10
Fix buffer overrun while parsing malformed JSON hex escape sequence

martijnthe:
Fixes https://github.com/jerryscript-project/jerryscript/issues/2200

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-json.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-json.c
@@ -40,6 +40,11 @@
 #define BUILTIN_UNDERSCORED_ID json
 #include "ecma-builtin-internal-routines-template.inc.h"
 
+/**
+ * The number of expected hexidecimal characters in a hex escape sequence
+ */
+#define ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH (4)
+
 /** \addtogroup ecma ECMA
  * @{
  *
@@ -182,13 +187,18 @@ ecma_builtin_json_parse_string (ecma_json_token_t *token_p) /**< token argument 
         }
         case LIT_CHAR_LOWERCASE_U:
         {
-          ecma_char_t code_unit;
-          if ((end_p - current_p >= 2) && !(lit_read_code_unit_from_hex (current_p + 1, 4, &code_unit)))
+          if ((end_p - current_p <= ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH))
           {
             return;
           }
 
-          current_p += 5;
+          ecma_char_t code_unit;
+          if (!(lit_read_code_unit_from_hex (current_p + 1, ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH, &code_unit)))
+          {
+            return;
+          }
+
+          current_p += ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH + 1;
 
           lit_utf8_byte_t char_buffer[LIT_UTF8_MAX_BYTES_IN_CODE_UNIT];
           buffer_size += lit_code_unit_to_utf8 (code_unit, char_buffer);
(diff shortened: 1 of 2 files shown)


martijnthe:
cc @DanielBallaSZTE @zherczeg @renatahodovan

martijnthe:
@akosthekiss Any idea what's wrong with the sign-off signature?
What CI reports is exactly what's at the end of the commit message...

zherczeg:
What do you think about #2194 ?

martijnthe:
@zherczeg ah, I had not seen there was already a PR for it.

I think #2194 is still not quite correct. I think the comparison should include the count for 1 char for the `u` and 4 chars for the hex value. It's also missing a test case as you pointed out.

In my PR I also replaced the various uses of magic number `4` with a `#define`, which is a bit better IMO.

martijnthe:
> @akosthekiss Any idea what's wrong with the sign-off signature?
What CI reports is exactly what's at the end of the commit message...
Should there be a newline at the end perhaps?
Or is it choking on the é?

OK, the problem is the `é` was represented differently in the commit message text vs author field...

robertsipka:
@martijnthe : Yes you are right, I've missed to increase with 1. I also strongly support the removal of magic constants. But the mistake that was made earlier to fix this kind of issue is still there. I've suggest you to modify the condition as in #2194 instead of adding a new one before.

jerry-core/ecma/builtin-objects/ecma-builtin-json.c
```diff
@@ -182,13 +187,17 @@ ecma_builtin_json_parse_string (ecma_json_token_t *token_p) /**< token argument
         }
         case LIT_CHAR_LOWERCASE_U:
         {
+          if ((end_p - current_p < ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH + 1)) {
+            return;
+          }
+
           ecma_char_t code_unit;
           if ((end_p - current_p >= 2) && !(lit_read_code_unit_from_hex (current_p + 1, 4, &code_unit)))
```
robertsipka: You should use ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH here too.

martijnthe: Oh whoops, I intended to remove that condition there. It would be redundant otherwise.

martijnthe:
@robertsipka OK, fixed it.

Should I also create a `#define` to replace the magic `1` (the length of the `u`)?
What do you think?

jerry-core/ecma/builtin-objects/ecma-builtin-json.c
```diff
@@ -182,13 +187,17 @@ ecma_builtin_json_parse_string (ecma_json_token_t *token_p) /**< token argument
         }
         case LIT_CHAR_LOWERCASE_U:
         {
+          if ((end_p - current_p < ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH + 1)) {
+            return;
+          }
+
           ecma_char_t code_unit;
-          if ((end_p - current_p >= 2) && !(lit_read_code_unit_from_hex (current_p + 1, 4, &code_unit)))
+          if (!(lit_read_code_unit_from_hex (current_p + 1, 4, &code_unit)))
```
robertsipka: You've forgot to replace 4 with ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH here.

robertsipka: I do not think it is necessary. LGTM (informally) after this minor fix.

jerry-core/ecma/builtin-objects/ecma-builtin-json.c
```diff
@@ -182,13 +187,17 @@ ecma_builtin_json_parse_string (ecma_json_token_t *token_p) /**< token argument
         }
         case LIT_CHAR_LOWERCASE_U:
         {
+          if ((end_p - current_p < ECMA_JSON_HEX_ESCAPE_SEQUENCE_LENGTH + 1)) {
+            return;
```
zherczeg: what about <= ? Then the +1 is unnecessary.

martijnthe: Done

jerry-core/ecma/builtin-objects/ecma-builtin-json.c
martijnthe: Done

martijnthe:
OK, I also satisfied the linter now and the CI build is ✅

PR #2205, opened 2018-02-14
[WIP] @jerryscript/debugger TypeScript library + node.js based CLI

martijnthe:
WIP -- just putting it out here so we can leverage the Travis CI, raise visibility and potentially draw in help from others.


martijnthe:
@grgustaf I've put up the skeleton project here. I've put up a brief README.md to hopefully help you get going. Let's chat soon.

jerry-debugger/src/lib/cdt-proxy.ts
```diff
@@ -0,0 +1,65 @@
...
+
+export interface ChromeDevToolsProxyServerOptions extends WebSocket.ServerOptions {
+}
+
+export class ChromeDevToolsProxyServer {
+  constructor(options: ChromeDevToolsProxyServerOptions) {
+    const wss = new WebSocketServer(options);
+    const rpcServer = new rpc.Server(wss);
+    const api: Crdp.CrdpServer = rpcServer.api();
+
```
martijnthe: @grgustaf this is currently not working at all.

I found the `noice-json-rpc` and `chrome-remote-debug-protocol` projects (both by the same author) while searching for existing libraries we could re-use. It looks very useful but I didn't get the example code to work out of the box just copy & pasting it. I didn't spend much time on it so I probably missed something silly.

jerry-debugger/src/lib/jrs-protocol-constants.ts
```diff
@@ -0,0 +1,70 @@
...
+    JERRY_DEBUGGER_NO_MORE_SOURCES: 10,
+    JERRY_DEBUGGER_CONTEXT_RESET: 11,
+    JERRY_DEBUGGER_CONTINUE: 12,
+    JERRY_DEBUGGER_STEP: 13,
+    JERRY_DEBUGGER_NEXT: 14,
+    JERRY_DEBUGGER_GET_BACKTRACE: 15,
+    JERRY_DEBUGGER_EVAL: 16,
+    JERRY_DEBUGGER_EVAL_PART: 17,
+  },
+};
```
martijnthe: @knightburton I lifted this from iotjscode. Think that should be fine, license-wise? Or should this bear the Samsung license banner..? IANAL...

knightburton: @martijnthe I'm not good at license related questions either, maybe @akosthekiss? IMHO, this should bear the license.

jerry-debugger/yarn.lock
```diff
@@ -0,0 +1,3002 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
```
zherczeg: What is the purpose of this file? Do we download and run scripts from the internet inside corporate/university firewall?

martijnthe: It's yarn's equivalent of `package-lock.json`:
https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

https://docs.npmjs.com/files/package-lock.json

zherczeg: Looks dangerous from security perspective.

martijnthe: Same thing here:
https://github.com/Samsung/iotjscode/blob/master/package-lock.json

martijnthe: JFYI, if you don't want to download stuff from the interwebs, it's possible to create an ["offline mirror"] as well and check that in.

It's also possible to roll up the sources and its the dependencies into a single blob of JS (executable by Node.js) by using something like rollup or webpack and commit that build output if you want to avoid having to pull in the dependencies from the npm/yarn package registries on the web.

["offline mirror"]: https://yarnpkg.com/blog/2016/11/24/offline-mirror/

zherczeg: I am worried about the review time. I need to check each source one-by-one.

martijnthe: > I need to check each source one-by-one.

You mean the sources of all the dependencies? Yeah, that will take a while :P

zherczeg: Alternative is using sandboxing like twitter:
http://www.nodesummit.com/prior-video/node-summit-2017-twitter-lite-powered-by-node-james-bellenger/

But commiting a full virtual machine seems overkill.

zherczeg:
It would not be bad if the resulting library would work with a stock IoT.js. I could imagine this similar to the debugger server, a platform independent part which builds the messages and a platform dependent which transmits them. The platform independent part could be pure JS.

jerry-debugger/yarn.lock
martijnthe: Yeah, I agree it's overkill to commit that here (great presentation, btw, thanks for sharing).

I think "trust" is generic and applies to any executable code that you did not write yourself.
People can run it in a VM if they have trust concerns.

I like to limit the number of dependencies we drag in, but at the same time, I'd like to use existing tools (i.e. TypeScript compiler, unit test framework, websocket library, etc.) and not write everything myself.

martijnthe:
@grgustaf I've merged your work. Great work so far!
I also added a fix for the issue in the test.

I also updated the description of this PR with a list of TODOs.

jerry-debugger/src/lib/jrs-protocol-constants.ts
HBehrens: I wonder how one will detect if/when these run out of sync with the counter-part in `jerry-core/debugger/debugger.h`. IMO, one should be generated from the other. This might as well answer the question about the license, then.

martijnthe: cc @grgustaf

martijnthe:
@grgustaf jfyi, I've added a commit here to trigger linting, building and testing the TypeScript code in CI -- at the moment it's failing with -- I'm guessing you've fixed that already in your PR:
```
$ jest
 PASS  src/lib/__tests__/cdt-proxy-http.test.ts (5.895s)
  ● Console
    console.log src/lib/cdt-proxy-http.ts:64
      Warning: unhandled URL: http://127.0.0.1:9229/json/foo
 FAIL  src/cli/__tests__/index.test.ts
  ● Test suite failed to run
    src/cli/cli.ts (65,3): Cannot invoke an expression whose type lacks a call signature. Type '(<TResult1 = void, TResult2 = never>(onfulfilled?: ((value: void) => TResult1 | PromiseLike<TResu...' has no compatible call signatures.
    src/cli/cli.ts (72,13): Parameter 'err' implicitly has an 'any' type.      
```

martijnthe:
@zherczeg @knightburton @akosthekiss: this is starting to become quite a massive PR: +5000 lines and we're not yet at feature parity with the existing JS/python debugger clients. So I'm starting to wonder what the (review) strategy is going to be going forward.

My preference would be to review what we have got and get it merged (even though it's not feature complete) and do smaller follow up PRs for the remaining tasks.

FYI, there's quite a bit of unit test coverage (although not yet enough). I've set up a linter and also added a Travis job to run the tests, linter and build it.

grgustaf:
@martijnthe I think you can click off the next three checklist items.

martijnthe:
@grgustaf sweeet

martijnthe:
Demo / screencast of the WIP JrS <=> Chrome DevTools debugging proxy using it with stock IoT.js:
https://youtu.be/Xe9r9maWmdI

Can someone on your team respond to my last message?
(cc @zherczeg @knightburton @yichoi @akosthekiss @DanielBallaSZTE @polaroi8d )

> @zherczeg @knightburton @akosthekiss: this is starting to become quite a massive PR: ~2000 lines (not counting the generated yarn.lock file) and we're not yet at feature parity with the existing JS/python debugger clients. So I'm starting to wonder what the (review) strategy is going to be going forward.
>
> My preference would be to review what we have got and get it merged (even though it's not feature complete) and do smaller follow up PRs for the remaining tasks.
>
> FYI, there's quite a bit of unit test coverage (although not yet enough). I've set up a linter and also added a Travis job to run the tests, linter and build it.

knightburton:
@martijnthe First of all, nice demo! I aggre with you in that this should be merged and do smaller PRs in the future, however I don't know what the others think about this. Also, I'll review the current progress (I am not a reviewer in the jerry but I can give informals).

jerry-debugger/src/lib/cdt-proxy.ts
```diff
@@ -0,0 +1,219 @@
...
+    // FIXME: probably not quite right, can include ../.. etc.
+    this.jsfile = options.jsfile || 'untitled.js';
+
+    server.listen(this.port);
+
+    const wss = new WebSocketServer({ server });
+    const rpcServer = new rpc.Server(wss);
+    this.api = rpcServer.api();
+
+    wss.on('connection', function connection(ws, req) {
```
knightburton: I think this callback could be an arrow function.

jerry-debugger/src/lib/cdt-proxy.ts
```diff
@@ -0,0 +1,219 @@
...
+  unused() {
+    // FIXME: pretend to use these to get around lint error for now
+    this.skipAllPauses, this.pauseOnExceptions, this.asyncCallStackDepth;
+  }
+
+  scriptParsed(script: JerryMessageScriptParsed) {
+    let name = script.name;
+    if (!name) {
+      // FIXME: make file / module name available to use here
+      name = 'untitled' + this.lastRuntimeScript++;
```
knightburton: Until fix, I would use template literals instead of concatenation.

jerry-debugger/src/lib/jrs-protocol-constants.ts
```diff
@@ -0,0 +1,73 @@
...
+export const JERRY_DEBUGGER_OUTPUT_RESULT = 25;
+export const JERRY_DEBUGGER_OUTPUT_RESULT_END = 26;
+
+export const JERRY_DEBUGGER_EVAL_OK = 1;
+export const JERRY_DEBUGGER_EVAL_ERROR = 2;
+
+export const JERRY_DEBUGGER_OUTPUT_OK = 1;
+export const JERRY_DEBUGGER_OUTPUT_WARNING = 2;
+export const JERRY_DEBUGGER_OUTPUT_ERROR = 3;
+
```
knightburton: The output type list is outdated, there are two new type in the jerry, output debug and trace.

jerry-debugger/src/lib/protocol-handler.ts
```diff
@@ -0,0 +1,535 @@
...
+    this.byteConfig.littleEndian = Boolean(data[3]);
+    this.version = data[4];
+
+    if (this.byteConfig.cpointerSize !== 2 && this.byteConfig.cpointerSize !== 4) {
+      this.abort('compressed pointer must be 2 or 4 bytes long');
+    }
+
+    if (this.version !== SP.JERRY_DEBUGGER_VERSION) {
+      this.abort('incorrect target debugger version detected: ' + this.version
+                 + ' expected: ' + SP.JERRY_DEBUGGER_VERSION);
```
knightburton: This could be done with template literals.

zherczeg:
What about moving this code into jerry-debugger/jerry-client-ts subdirectory?

jerry-debugger/src/lib/protocol-handler.ts
```diff
@@ -0,0 +1,535 @@
...
+        this.abort('the first message must be configuration');
+        return;
+      }
+    }
+
+    const handler = this.functionMap[message[0]];
+    if (handler) {
+      handler.call(this, message);
+    } else {
+      this.abort('unhandled protocol message type: ' + message[0]);
```
knightburton: I think this could be a template literal.

jerry-debugger/src/lib/protocol-handler.ts
```diff
@@ -0,0 +1,535 @@
...
+    return this.activeBreakpoints[breakpointId];
+  }
+
+  findBreakpoint(scriptId: number, line: number, column: number = 0) {
+    if (scriptId <= 0 || scriptId >= this.lineLists.length) {
+      throw new Error('invalid script id');
+    }
+    const lineList = this.lineLists[scriptId];
+    if (!lineList[line]) {
+      throw new Error('no breakpoint found for line: ' + line);
```
knightburton: This could be a template literal, too.

martijnthe:
> What about moving this code into jerry-debugger/jerry-client-ts subdirectory?

Yeah, that sounds good to me.

martijnthe:
@zherczeg this is done now:

> What about moving this code into jerry-debugger/jerry-client-ts subdirectory?

@grgustaf has wrapped up the eval work so I think now is a good moment to call it a PR, do final reviews and get it merged.

@knightburton want to give it a final review pass? Your previous comments have been addressed by @grgustaf.

Thanks all!

zherczeg:
It seems non-related commits are also part of this patch. Please squash it and rebase it to master.

grgustaf:
Note: Before rebase was commit ad58be4c727fb931724efc6f2790c307797fccc6

knightburton:
LGTM (informal)
@martijnthe I am fine with the current state!

grgustaf:
Just fixed up Signed-off-by to pass Travis with the last tweak.

zherczeg:
Correct me if I am wrong, but this code downloads and runs 3rd party software on the user system. We need to (legally) protect us (devs + companies involved) from any harm done by these 3rd party software tools. I am not a lawyer, but here is my idea: it would be good to display a text before the first use, which tells that we are not responsible, and the user must accept it (y/N, so no is the default). But this might be not enough, so it would be good to hear others as well. @daeyeon, what do you think?

The other option is go as the rest of the jerry: if these tools provided by standard packages, we could make it a requirement, and the user is responsible to provide them. Just like the compiler or python: if these tools do harmful things, it is not our responsibility.

martijnthe:
> this code downloads and runs 3rd party software on the user system.

Running `npm install` or `yarn install` downloads (3rd party) library dependencies from the registry of the [npm] package manager. These dependencies are listed in the [package.json] file in this PR.
This is exactly the same mechanism that you are also using in IoT.js Code. In the "How to start" steps you also instruct users to run `npm install` and then `npm start` etc. which will download and run 3rd party library code that was downloaded from npm.

How do you deal with the concerns you are raising over there?
I did not see any warnings whatsoever when using the IoT.js Code project.

[npm]: https://www.npmjs.com/
[package.json]: https://github.com/jerryscript-project/jerryscript/pull/2205/files#diff-15186e89a04a4f1547debb0df66c4b48

zherczeg:
iotjscode is a website from the user perspective. No need to install anything to use. And the browser is the line of defense there. We share the code behind the website, but that is not official part of JerryScript or IoT.js. Only the python and html debuggers are official, although I want to remove the html one if this code lands.

Following this logic an option could be to make this a highly recommended but independent project, and we can add references to it in the documentation. I do think this is a very useful project! But from stakeholder perspective it introduces new risks which could backfire (on me in particular), even if it is unintended.

Unless somebody else with management right says that this is ok. Hence I added @daeyeon . Lets wait his opinion.

yichoi:
I would like to make JerryScript light as much as possible therefore this jerryscript repo should maintain minimum core logics only. I respect this PR since I understand it is great work which elaborate the efficiency for developers. My suggestion is to maintain this features to separate repo under jerryscript-project, since I can make new repository for this purpose.

martijnthe:
@yichoi thanks for your response.

Creating a separate repo under the `jerryscript-project` organization makes sense to me.

It also raises a couple questions:

1. How to detect breakages caused by changes in the protocol/interface? A change in the C debugger code (in this repo) can cause the clients (in the new repo) to stop working. If the client(s) are in another repo it's harder to make it part of the CI test set up. Any suggestions?

2. What about the other debugger clients (`jerry-client-ws.py` and `jerry-client-ws.html`)? Should they be moved as well? We have the intention to re-use the same test cases for the python `jerry-client-ws.py` to ensure the same behavior. If the code is split, this becomes harder to do.

3. Can we (Intel contributors myself and @grgustaf for now) get (push) access to this new repo?

4. When splitting code off, it becomes more important to design the protocols and interfaces such that forwards and backwards compatibility can be maintained. So far this has not really been the case with the JerryScript debugger protocol.

5. How to name the new repo :)

zherczeg:
1) Do you have a good suggestion for this? How fast the other repos can follow a protocol change?

2) I would like to delete `jerry-client-ws.html`. The `jerry-client-ws.py` is small and we should provide something for our users, so I would keep it. Good for testing as well.

4) This is important. In the early phases of the implementation we just explored the requirements, so it was kind of a learning phase. Large changes of the protocol was expected. Now based on these experiences we should start working on defining a real jerryscript debugger protocol interface.

Some questions:
1) Are we happy with the current flow of data or we should change it?
2) Keep 1 byte message type, or introduce 2 byte types, or 1 byte type / 1 byte subtype systems.
3) We should define holes for message groups (e.g. breakpoint mode, etc. ).
4) Perhaps user messages should be introduced in some way (at least a hole reserved for them).

martijnthe:
### Repo / Logistics

>> 1. How to detect breakages caused by changes in the protocol/interface? A change in the C debugger code (in this repo) can cause the clients (in the new repo) to stop working. If the client(s) are in another repo it's harder to make it part of the CI test set up. Any suggestions?
> 1. Do you have a good suggestion for this?

I'm thinking now something along these lines: what if the `jerry-client-ts` repo has a file with the git hash of the latest `jerryscript` that the library is known to work with. The CI scripts for the `jerry-client-ts` repo can then use that to clone `jerryscript`, build and use the `jerry` binary from it, re-use the tests, etc.

> How fast the other repos can follow a protocol change?

I suspect the `jerry-client-ts` repo will be "playing catch-up" all the time. How fast it can follow depends on how important the support is/becomes and how many people can contribute. As long as there's sane versioning + proper back/forwards compatibility arrangements in the protocol, being slightly out of sync shouldn't be the end of the world.

>> 2. What about the other debugger clients (jerry-client-ws.py and jerry-client-ws.html)? Should they be moved as well? We have the intention to re-use the same test cases for the python jerry-client-ws.py to ensure the same behavior. If the code is split, this becomes harder to do.
> 2. I would like to delete jerry-client-ws.html. The jerry-client-ws.py is small and we should provide something for our users, so I would keep it. Good for testing as well.

OK. That means we'll need to download (git clone) the python debugger tests from the main `jerryscript` project as part of the CI process of the new `jerry-client-ts` repo if we want to run those tests against the TypeScript version of the client. No big deal, I think, as long as the `jerryscript` git hash is "pinned" in some file in the `jerry-client-ts` repo. Could perhaps use git submodules? maybe too heavy if it's only required for running tests? OTOH, I guess during development you will want to run the tests locally too, so perhaps including `jerryscript` as git submodule inside `jerry-client-ts` isn't that bad.

### Protocol (re)design

>   1. Are we happy with the current flow of data or we should change it?

@grgustaf what are your thoughts so far, having worked intimately with the protocol?

One thing that comes to mind is that perhaps the "chunking" of messages is more a "transport concern" and not so much an debugger-application level concern. But changing that would mean quite a large overhaul and I'm not sure whether that's worth doing.

>   2. Keep 1 byte message type, or introduce 2 byte types, or 1 byte type / 1 byte subtype systems.
>   3. We should define holes for message groups (e.g. breakpoint mode, etc. ).
>   4. Perhaps user messages should be introduced in some way (at least a hole reserved for them).

I think these are good ideas.
I would vote for 1 byte type + 1 byte subtype (instead of creating holes in a flat space).
Reserving one type value for user messages makes senses to me.

Some thoughts re. protocol versioning:
I'd like to see either a) semver or b) a simple base version integer (equality checked) + capabilities bitset. Both approaches are useful for maintaining back/forwards compatibility. In my personal experience, I've enjoyed using capability bitsets. Both sides can indicate what each side has support for and then the AND'ed bitset is the set of capabilities that is actually used.

Another (large topic) to talk about is future features for the protocol (inspecting the scope chain, heap profiling, performance profiling, etc. etc...). Perhaps one step at a time ;)

grgustaf:
Sorry, I wasn't expecting a flood of activity here today and didn't look. :) I'll have to think about this more.

One rather minor thing off-hand is I was recently thinking how there 6 or 7 debugger messages that can be multipart and are implemented w/ e.g. SOURCE_CODE_NAME and SOURCE_CODE_NAME_END types. It would be nice if the fragmentation were handled with a single mechanism like an "end" flag that works across all packet types and could be handled once for all. 

There's also an implicit order that the packets will be sent in which I don't think is really documented but is effectively part of the protocol. I had to make assumptions about those packets always coming in the basically the same order (trying not be more strict than necessary). This might be fine but could just benefit by being "written in stone". So for example, a Configuration message always comes first. When a new source is parsed, Source Code and Source Code Name messages come. IIRC when the final Source Code packet comes, the Source Code Name is expected to already be complete. Then Breakpoint Line/Offset Lists come and may be a little out of sync, but by the time the Byte Code CP message comes then equal numbers of breakpoint lines and offsets must have been reported. This is maybe the kind of information that a UML Sequence Diagram captures.

I agree as the protocol version has quickly gone from 1 to 2 to 3, I was thinking something needs to be made more stable. It seemed unfortunate that some message numbers were changed in version 2 because it meant incompatibility with version 1. I guess this what you mean by leaving holes for expansion.

Interacting now with Chrome Dev Tools gives us a long list of features that would be useful to add to the JrS debugger, to fully exploit the potential. So we have a lot of proposed changes coming and a version mechanism allowing more finesse would be useful. Perhaps a set of required base features that are called "version 1" but capability flags for what's beyond that. Then someday an expanded set of base features could be called "version 2" (I'm thinking of Android API levels). I guess obviously you want to keep the JrS device a minimal as possible so if anything's going to be flexible about version it should be the client app. So we could potentially support a range of versions w/ one client.

As I look back over both of your comments, I may have been entirely repeating you in my own words... ;)

martijnthe:
@yichoi can you create the new repo?

Name suggestions:
- `jerryscript-debugger-ts` (my favorite -- long but descriptive. `ts` is often used to indicate "TypeScript" related things)
- `jerryscript-debug-ts`
- `jerry-client-ts` (current folder name)
- ...

Who will get push access? Can me and @grgustaf get push access as well?

I think we can start putting the code from this PR in the new repo + set up CI, etc and in parallel work on protocol design improvements.

martijnthe:
@yichoi thank you!

Do you also have control over the Travis account? If so, could you enable Travis for the new repo? I will work on setting up a .travis.yml coming week.

martijnthe:
@yichoi could you also add the new repo to the "JerryScript" team, so that the existing maintainers also have access to the new repo?

cc @LaszloLango

grgustaf:
Resubmitted with slight tweaks to README.md and package.json as a pull request to the new jerryscript-debugger-ts project as jerryscript-project/jerryscript-debugger-ts#4.

martijnthe:
> Do you also have control over the Travis account? If so, could you enable Travis for the new repo? I will work on setting up a .travis.yml coming week.

@akosthekiss already took care of this.

grgustaf:
This code has been merged into the new repo via jerryscript-project/jerryscript-debugger-ts#4, so this can be closed.

PR #2207, opened 2018-02-14
Multiple nexts with one command

DanielBallaSZTE:
Make a next command more gdb like.
If an argument is given `next 10`, it does 10 nexts.

### DIFF
--- a/jerry-debugger/jerry-client-ws.py
+++ b/jerry-debugger/jerry-client-ws.py
@@ -25,6 +25,7 @@ import socket
 import struct
 import sys
 import math
+import time
 
 # Expected debugger protocol version.
 JERRY_DEBUGGER_VERSION = 1
@@ -265,7 +266,22 @@ class DebuggerPrompt(Cmd):
 
     def do_next(self, args):
         """ Next breakpoint in the same context """
-        self._exec_command(args, JERRY_DEBUGGER_NEXT)
+        if not args:
+            args = 0
+        else:
+            try:
+                args = int(args)
+                if args <= 0:
+                    raise ValueError(args)
+            except ValueError as val_errno:
+                print("Error: expected a positive integer: %s" % val_errno)
+                return
+
+            args = min(args, len(self.debugger.last_breakpoint_hit.function.lines) -
+                       self.debugger.last_breakpoint_hit.function.line) - 1
+            self.debugger.repeats_remain = args
+
+        self._exec_command("", JERRY_DEBUGGER_NEXT)
         self.cont = True
 
     do_n = do_next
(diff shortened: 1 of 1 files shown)


jerry-debugger/jerry-client-ws.py
```diff
@@ -273,7 +274,19 @@ def do_step(self, args):
 
     def do_next(self, args):
         """ Next breakpoint in the same context """
-        self.exec_command(args, JERRY_DEBUGGER_NEXT)
+        if args == "":
+            args = 0
```
robertsipka: The preferred way to check that a string is empty (if you know that your variable is a string) is this: `if not args:`

If your variable could also be some other type then you should use this form `if args == "":`, but as I saw in the following code, it always waiting for string and then you cast that into an integer.

jerry-debugger/jerry-client-ws.py
```diff
@@ -273,7 +274,19 @@ def do_step(self, args):
...
+        else:
+            try:
+                args = int(args)
+                args = min(args, len(self.debugger.last_breakpoint_hit.function.lines) -
+                           self.debugger.last_breakpoint_hit.function.line) - 1
+                self.debugger.nexts_remain = args
+            except ValueError as val_errno:
```
robertsipka: I do not think it's necessary to put it all in a try block, because you only use cast in the first command.

jerry-debugger/jerry-client-ws.py
```diff
@@ -273,7 +274,19 @@ def do_step(self, args):
...
     def do_next(self, args):
         """ Next breakpoint in the same context """
-        self.exec_command(args, JERRY_DEBUGGER_NEXT)
+        if args == "":
+            args = 0
+        else:
+            try:
+                args = int(args)
+                args = min(args, len(self.debugger.last_breakpoint_hit.function.lines) -
+                           self.debugger.last_breakpoint_hit.function.line) - 1
```
robertsipka: This condition can not be negative here?

jerry-debugger/jerry-client-ws.py
```diff
@@ -273,7 +274,19 @@ def do_step(self, args):
...
+                self.debugger.nexts_remain = args
+            except ValueError as val_errno:
+                print("Error: expected a positive integer: %s" % val_errno)
```
robertsipka: What about those case when the args value is a zero or negative number?

DanielBallaSZTE:
@robertsipka I've updated the PR

jerry-debugger/jerry-client-ws.py
```diff
@@ -561,6 +577,7 @@ def __init__(self, address):
         self.src_offset_diff = 0
         self.client_socket = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
         self.client_socket.connect((self.host, self.port))
+        self.nexts_remain = 0
```
zherczeg: I would call this as repeat, so we could extend this to step etc. later.

PR #2212, opened 2018-02-19
Exclude hashmaps from GC marks.

### LINKED ISSUE #2146: misuse of type in ecma_gc_mark_property function which can lead to memory corruption
in ecma_gc_mark function

```
      JERRY_ASSERT (prop_iter_p->types[0] == ECMA_PROPERTY_TYPE_HASHMAP
                    || ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));

      ecma_gc_mark_property ((ecma_property_pair_t *) prop_iter_p, 0);
      ecma_gc_mark_property ((ecma_property_pair_t *) prop_iter_p, 1);
```
it call ecma_gc_mark_property function.

```
static void
ecma_gc_mark_property (ecma_property_pair_t *property_pair_p, /**< property pair */
                       uint32_t index) /**< property index */
...
```
but type of argument can be hashmap. this make no problem in most situation because header.types[0] will be always 4( ECMA_PROPERTY_TYPE_HASHMAP ) and types[1] will be 0 and handled by function in this case . but types[1] of hashmap is shiftcounter so it can be bigger number in very specific case (  in below code, when max_property_count is bigger than LIT_STRING_HASH_LIMIT   ... )

```
//jerry-core/ecma/base/ecma-property-hashmap.c
void
ecma_property_hashmap_create (ecma_object_t *object_p) /**< object */
{
......
......
  if (max_property_count <= LIT_STRING_HASH_LIMIT)
  {
    hashmap_p->header.types[1] = 0;
  }
  else
  {
    while (max_property_count > LIT_STRING_HASH_LIMIT)
    {
      shift_counter++;
      max_property_count >>= 1;
    }
  }

  hashmap_p->header.types[1] = shift_counter;
.....
```

if shift_counter is bigger than 0 then type misuse bug code can be executed

```
static void
ecma_gc_mark_property (ecma_property_pair_t *property_pair_p, /**< property pair */
                       uint32_t index) /**< property index */
{
...
      ecma_value_t value = property_pair_p->values[index].value;

```

zherczeg:
Fixes #2146.

The issue is valid, but the test case does not work anymore. I cannot add a test case myself.

### DIFF
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -381,10 +381,15 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
   {
     ecma_property_header_t *prop_iter_p = ecma_get_property_list (object_p);
 
+    if (prop_iter_p != NULL && prop_iter_p->types[0] == ECMA_PROPERTY_TYPE_HASHMAP)
+    {
+      prop_iter_p = ECMA_GET_POINTER (ecma_property_header_t,
+                                      prop_iter_p->next_property_cp);
+    }
+
     while (prop_iter_p != NULL)
     {
-      JERRY_ASSERT (prop_iter_p->types[0] == ECMA_PROPERTY_TYPE_HASHMAP
-                    || ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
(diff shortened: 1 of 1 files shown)


yichoi:
LGTM

PR #2217, opened 2018-02-21
Add tests for the es5.1 profile to tools/run-tests.py

pmarkee:

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -44,7 +44,19 @@ JERRY_UNITTESTS_OPTIONS = [
              '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset']),
     Options('doctests-debug',
             ['--doctests', '--jerry-cmdline=off', '--debug', '--error-messages=on',
-             '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset'])
+             '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset']),
+    Options('unittests-es5.1',
+            ['--unittests', '--jerry-cmdline=off', '--error-messages=on', '--snapshot-save=on',
+             '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es5.1', '--mem-stats=on']),
+    Options('unittests-es5.1-debug',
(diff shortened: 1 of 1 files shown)


tools/run-tests.py
```diff
@@ -261,6 +279,9 @@ def run_jerry_tests(options):
         if '--profile=es2015-subset' not in job.build_args:
             skip_list.append(r"es2015\/")
 
+        if '--profile=es5.1' not in job.build_args:
```
galpeter: There is no skip list for es5.1, so we don't need this part.

LaszloLango:
@pmarkee please rebase your PR.

pmarkee:
@LaszloLango done

galpeter:
@LaszloLango @pmarkee I've created the #2228 PR to try and resolve the too big travis output problem.

galpeter:
@pmarkee could you please rebase the PR? The test runner quiet mode has landed which should fix the Travis output problem(s).

tools/run-tests.py
```diff
@@ -44,7 +44,23 @@ def get_binary_path(bin_dir_path):
              '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset']),
     Options('doctests-debug',
             ['--doctests', '--jerry-cmdline=off', '--debug', '--error-messages=on',
-             '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset'])
+             '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset']),
+    Options('unittests',
+            ['--unittests', '--jerry-cmdline=off', '--error-messages=on', '--snapshot-save=on', '--snapshot-exec=on', '--vm-exec-stop=on', '--profile=es2015-subset', '--mem-stats=on']),
```
galpeter: It seems that these lines are too long (some of them are above 120 characters). We should extract the common options to a different variable and just concatenate the rest where needed.

pmarkee: Thanks for the review! Maybe we should do this to other parts of the code as well, it would make the lines much shorter and easier to read. For now I think we should just break the lines that are too long, and apply your suggestion in a separate PR afterwards.

galpeter: okay, that works for me

yichoi:
LGTM

PR #2220, opened 2018-02-23
Check _XOPEN_SOURCE macro before defining

galpeter:
In jrt.h the _XOPEN_SOURCE macro is defined.
However if the compiler already specifies this as
a default define, it can lead to compiler warnings
or errors.

By adding a macro check the warnings/errors can be
avoided.

### DIFF
--- a/jerry-core/jrt/jrt.h
+++ b/jerry-core/jrt/jrt.h
@@ -16,8 +16,11 @@
 #ifndef JRT_H
 #define JRT_H
 
+#if !defined (_XOPEN_SOURCE) || _XOPEN_SOURCE < 500
+#undef _XOPEN_SOURCE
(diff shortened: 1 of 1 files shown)


jerry-core/jrt/jrt.h
```diff
@@ -16,8 +16,11 @@
 #ifndef JRT_H
 #define JRT_H
 
+#if !defined(_XOPEN_SOURCE) || _XOPEN_SOURCE < 500
```
LaszloLango: > ./jerry-core/jrt/jrt.h:19: error: there should be exactly one space before left parentheses

galpeter:
@LaszloLango I've update the PR.

yichoi:
LGTM

PR #2221, opened 2018-02-26
Improve magic string handling.

zherczeg:
Remove unnecessary ref / deref calls when magic strings are used.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -220,7 +220,7 @@ ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 stri
 
     if (magic_string_ex_id < lit_get_magic_string_ex_count ())
     {
-      return ecma_get_magic_string_ex (magic_string_ex_id);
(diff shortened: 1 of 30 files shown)


zherczeg:
600 byte code size (135916 - > 135316), and +0.343% perf (mostly noise).

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -494,15 +513,18 @@ ecma_new_ecma_string_from_magic_string_ex_id (lit_magic_string_ex_id_t id) /**<
...
+ * Returns the constant assigned to the length magic string.
+ *
+ * Note:
+ *   Calling ecma_deref_ecma_string on the returned pointer is optional.
  *
  * @return pointer to ecma-string descriptor
  */
 ecma_string_t *
-ecma_new_ecma_length_string (void)
+ecma_get_length_string (void)
```
galpeter: Would it make sense to get rid of this function? IMHO there is no real benefit to have this in this form. We could simply call the get_magic_string directly where needed.

zherczeg: I will do it in another PR if that is good.

zherczeg: #2231

galpeter:
LGTM

PR #2222, opened 2018-02-26
Bugfix: check context manager deinit callback for NULL before calling it

martijnthe:
The deinit_cb should be allowed to be NULL.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -122,8 +122,9 @@ typedef uint32_t jerry_value_t;
 **Summary**
 
 Structure that defines how a context data item will be initialized and deinitialized. JerryScript zeroes out the memory
-for the item by default, and if the `init_cb` field is not NULL, it will be called with the pointer to the memory  as
-an additional custom initializer.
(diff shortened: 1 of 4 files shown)


martijnthe:
cc @gabrielschulhof

zherczeg:
While the code is correct technically, is it a vaild use case that you don't want to deallocate your allocated memory?

martijnthe:
> While the code is correct technically, is it a vaild use case that you don't want to deallocate your allocated memory?

Guess it depends what you mean with "your allocated memory".
The memory that got allocated by virtue of setting the `bytes_needed` field of the `jerry_context_data_header_t` is already being deallocated automatically in the lines right after the deinit callback.

I think not having to run any user deinit code (and only automatically deallocating the memory) is a pretty common use case?

zherczeg:
You are right. I haven't noticed that.

gabrielschulhof:
@martijnthe you should also update the [documentation](https://github.com/jerryscript-project/jerryscript/blob/master/docs/02.API-REFERENCE.md#jerry_context_data_manager_t)

LaszloLango:
@martijnthe travis CI should be green after a rebase.

martijnthe:
OK, updated the documentation + doc strings as well.

PR #2228, opened 2018-02-28
Add quiet mode for test runner(s)

galpeter:
In one of the previous PR we have encountered a problem,
where the Travis cuts off the test execution
as the generated log file is too big.

By adding a quiet mode for the test runners, we will
only report the failing tests.


galpeter:
This problem was encountered in #2217

yichoi:
LGTM

PR #2232, opened 2018-03-01
Improve property search.

zherczeg:
Optimize some functions for the two types (direct/non-direct) of strings.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -1708,8 +1708,13 @@ ecma_string_compare_to_property_name (ecma_property_t property, /**< property na
     return ecma_property_to_string (property, prop_name_cp) == string_p;
   }
 
+  if (ECMA_IS_DIRECT_STRING (string_p))
+  {
+    return false;
+  }
+
   ecma_string_t *prop_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t, prop_name_cp);
-  return ecma_compare_ecma_strings (prop_name_p, string_p);
+  return ecma_compare_ecma_non_direct_strings (prop_name_p, string_p);
 } /* ecma_string_compare_to_property_name */
 
 /**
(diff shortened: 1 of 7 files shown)


zherczeg:
+0.87% perf imporvement, no other changes

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1799,6 +1804,45 @@ ecma_compare_ecma_strings (const ecma_string_t *string1_p, /* ecma-string */
...
 
+/**
+ * Compare two non-direct ecma-strings
+ *
+ * @return true - if strings are equal;
+ *         false - otherwise
+ */
+inline bool __attr_always_inline___
+ecma_compare_ecma_non_direct_strings (const ecma_string_t *string1_p, /* ecma-string */
+                                      const ecma_string_t *string2_p) /* ecma-string */
```
galpeter: Style for the comment should be: `/**< .... */`

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -36,48 +36,33 @@
...
 JERRY_STATIC_ASSERT (ECMA_OBJECT_FLAG_EXTENSIBLE == (ECMA_OBJECT_FLAG_BUILT_IN_OR_LEXICAL_ENV << 1),
                      ecma_extensible_flag_must_follow_the_built_in_flag);
 
-/**
- * The ecma object ref one must follow the extensible flag.
- */
 JERRY_STATIC_ASSERT (ECMA_OBJECT_REF_ONE == (ECMA_OBJECT_FLAG_EXTENSIBLE << 1),
                      ecma_object_ref_one_must_follow_the_extensible_flag);
 
-/**
```
galpeter: Do we really want to remove these comments?

jerry-core/ecma/base/ecma-property-hashmap.c
```diff
@@ -492,6 +485,62 @@ ecma_property_hashmap_find (ecma_property_hashmap_t *hashmap_p, /**< hashmap */
...
+                                                                           pair_list_p[entry_index]);
+
+        ecma_property_t *property_p = property_pair_p->header.types + offset;
+
+        JERRY_ASSERT (ECMA_PROPERTY_IS_NAMED_PROPERTY (*property_p));
+
+        if (property_pair_p->names_cp[offset] == property_name_cp
+            && ECMA_PROPERTY_GET_NAME_TYPE (*property_p) == prop_name_type)
+        {
+#ifndef JERRY_NDEBUG
```
galpeter: Why do we need the `#ifndef` here? (and at the other places)

jerry-core/ecma/base/ecma-helpers.c
zherczeg: No other file has them, since the static assert has a description inside. It is basically just a duplication.

jerry-core/ecma/base/ecma-property-hashmap.c
zherczeg: Because asserts implemented in a interesting way and references its variables without evaluating them in release mode.

PR #2233, opened 2018-03-02
Support large string constants in 32 bit cpointer mode.

zherczeg:
After this patch, all sunspider tests run when cpointer 32 is enabled.


jerry-core/parser/js/js-parser-limits.h
```diff
@@ -32,10 +32,18 @@
...
+#else /* !JERRY_CPOINTER_32_BIT */
+#define PARSER_MAXIMUM_STRING_LIMIT 65535
+#endif /* JERRY_CPOINTER_32_BIT */
+
 /* Maximum string length.
- * Limit: 65535. */
+ * Limit: PARSER_MAXIMUM_STRING_LIMIT. */
 #ifndef PARSER_MAXIMUM_STRING_LENGTH
-#define PARSER_MAXIMUM_STRING_LENGTH 65535
+#define PARSER_MAXIMUM_STRING_LENGTH PARSER_MAXIMUM_STRING_LIMIT
```
galpeter: If it is the same as the other define, do we need it at all?

zherczeg: Yes, because the maximum is the default configuration, although we allow lower values.

galpeter: I see

PR #2235, opened 2018-03-05
Eliminate early memory deallocations in case of NuttX and TizenRT

rtakacs:
The error value is released in the print_unhandled_exception function, however
that value is used and released later.


robertsipka:
LGTM (informally)

rtakacs:
Note: the expected fail tests cause arm_dataabort on artik053 without this patch.

PR #2239, opened 2018-03-11
Support static snapshots.

zherczeg:
Unlike normal snapshots, no part of a static snapshot is loaded into the RAM when executed from ROM. Static snapshots rely heavily on external magic strings.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -4736,6 +4736,77 @@ main (void)
 - [jerry_init](#jerry_init)
 - [jerry_cleanup](#jerry_cleanup)
 - [jerry_exec_snapshot](#jerry_exec_snapshot)
+- [jerry_parse_and_save_static_snapshot](#jerry_parse_and_save_static_snapshot)
+
+
+## jerry_parse_and_save_static_snapshot
+
+**Summary**
+
+Generate static snapshot from the specified source code.
+
+Unlike normal snaphots static snaphots are fully executed from ROM. Not
+even their header is loaded into the RAM. However they can only depend
+on magic strings and 28 bit integer numbers. Regular expression literals
+are not supported as well.
+
+**Prototype**
+
+```c
+size_t
+jerry_parse_and_save_static_snapshot (const jerry_char_t *source_p,
+                                      size_t source_size,
+                                      bool is_for_global,
+                                      bool is_strict,
+                                      uint32_t *buffer_p,
+                                      size_t buffer_size);
+```
+
+- `source_p` - script source, it must be a valid utf8 string.
+- `source_size` - script source size, in bytes.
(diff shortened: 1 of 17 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -4736,6 +4736,77 @@ main (void)
...
+                                      uint32_t *buffer_p,
+                                      size_t buffer_size);
+```
+
+- `source_p` - script source, it must be a valid utf8 string.
+- `source_size` - script source size, in bytes.
+- `is_for_global` - snapshot would be executed as global (true) or eval (false).
+- `is_strict` - strict mode
+- `buffer_p` - buffer to save snapshot to.
+- `buffer_size` - the buffer's size.
```
galpeter: What does the size mean here? Byte size? or element size?

docs/02.API-REFERENCE.md
```diff
@@ -4736,6 +4736,77 @@ main (void)
...
+
+## jerry_parse_and_save_static_snapshot
+
+**Summary**
+
+Generate static snapshot from the specified source code.
+
+Unlike normal snaphots static snaphots are fully executed from ROM. Not
+even their header is loaded into the RAM. However they can only depend
+on magic strings and 28 bit integer numbers. Regular expression literals
```
galpeter: I think we should emphasise that the magic strings during the snapshot generation must be the same as the magic strings during execution.

jerry-core/ecma/operations/ecma-function-object.h
```diff
@@ -43,6 +43,17 @@ ecma_op_create_arrow_function_object (ecma_object_t *scope_p, const ecma_compile
...
+const ecma_compiled_code_t *
+ecma_op_function_get_compiled_code (ecma_extended_object_t *function_p);
+
+#ifndef CONFIG_DISABLE_ES2015_ARROW_FUNCTION
+const ecma_compiled_code_t *
+ecma_op_arrow_function_get_compiled_code (ecma_arrow_function_t *arrow_function_p);
+#endif /* !CONFIG_DISABLE_ES2015_ARROW_FUNCTION */
+
+ecma_value_t
+ecma_op_function_has_instance (ecma_object_t *func_obj_p, ecma_value_t value);
```
galpeter: Why do we need to move this function prototype?

docs/02.API-REFERENCE.md
zherczeg: Byte size just like the other cases (this is mostly a copy-paste). And the plan is rewriting the saving API anyway.

jerry-core/ecma/operations/ecma-function-object.h
zherczeg: Just to keep the .c side order. Not a critical change though.

yichoi:
LGTM

PR #2245, opened 2018-03-19
Move the sleep function to jerry-port

imiklos:
Now the jerry-debugger uses the jerry-port's sleep, therefore if there are systems that don't support
usleep or nanosleep they can now define their own function.

### DIFF
--- a/jerry-core/include/jerryscript-port.h
+++ b/jerry-core/include/jerryscript-port.h
@@ -139,6 +139,13 @@ double jerry_port_get_current_time (void);
 struct jerry_instance_t *jerry_port_get_current_instance (void);
 
 /**
+ * Makes the process sleep for a given time.
+ */
+#ifdef JERRY_DEBUGGER
+void jerry_port_sleep (uint32_t sleep_time);
+#endif /* JERRY_DEBUGGER */
+
+/**
  * @}
  */
 
--- a/docs/05.PORT-API.md
+++ b/docs/05.PORT-API.md
@@ -112,6 +112,15 @@ Allow user to provide external buffer for jerry instance (which includes an isol
 struct jerry_instance_t *jerry_port_get_current_instance (void);
 ```
 
+## Sleep
+
+```c
+/**
+ * Makes the process sleep for a given time.
+ */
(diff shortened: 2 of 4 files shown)


jerry-port/default/default-date.c
```diff
@@ -74,3 +83,21 @@ double jerry_port_get_current_time (void)
   return 0.0;
 #endif /* __GNUC__ */
 } /* jerry_port_get_current_time */
+
+#ifdef JERRY_DEBUGGER
+#include <jrt.h>
+
```
robertsipka: You've added this include twice. Remove from here.

akosthekiss: why is jrt.h required in default-date.c at all? that seems like a layering violation to me. if that's only for the sake of uint32_t, there are standard libc headers that could be included.

jerry-core/include/jerryscript-port.h
```diff
@@ -141,6 +141,9 @@ struct jerry_instance_t *jerry_port_get_current_instance (void);
 /**
  * @}
  */
+#ifdef JERRY_DEBUGGER
+void jerry_port_sleep (uint32_t sleep_time);
+#endif /* JERRY_DEBUGGER */
```
akosthekiss: this is outside the comment group. should be within.

imiklos:
@robertsipka @akosthekiss Thank you for the reviews. I updated the PR.

imiklos:
@LaszloLango Thank you for the review. Moved the port function to a new file and updated the `docs/05.PORT-API.md`.

yichoi:
LGTM

jerry-core/include/jerryscript-port.h
```diff
@@ -138,6 +138,13 @@ double jerry_port_get_current_time (void);
  */
 struct jerry_instance_t *jerry_port_get_current_instance (void);
 
+/**
+ * Makes the process sleep for a given time.
```
zherczeg: What is the unit of time?

imiklos: The unit of time is millisecond.

Issue #2229, opened 2018-03-01
Add debugger support with other platforms


jimmy-huang:
The current debugger only works on Linux's BSD socket, I am trying to get the debugger to work on Zephyr since it recently added BSD socket support.  However, it has slightly different headers for using the socket library and also it currently only supports a subset of the API.  It would be nice if we can decouple all those socket APIs like accept(), send(), recv() and sleep() into jerry_port APIs as they can be platform-dependent.   Are there any plans to add debugger support for other platforms?

Here's the working here, basically I replaced the debugger-ws.c with my own debugger-ws-zephyr.c when building for Zephyr:
https://github.com/jimmy-huang/jerryscript/tree/js-debugger

zherczeg:
We also have been thinking about it. TizenRT also needs platform specific changes. Our problem is maintenance: there will be a lot of code duplication.

Anyway this is something we plan to do later in this year.

martijnthe:
If a lot of code literally has to be duplicated, I think we need to look at factoring out the common parts and coming up with a better interface for providing the "transport" for the debugger communication. I guess that's also what @jimmy-huang is hinting at.

When reworking this, I think we should make it generic enough to also support entirely different types of communications. I'm thinking about USB, UART, BLE, etc.

Where sensible, I think it would also be a good idea to create JerryScript gratified "transport profiles" for each transport where applicable. For example, in the case of USB, it could mean using the same name/class/subclass/endpoints. That way, the debugger client tooling would only have to have one USB protocol implementation. Note that for the TCP/IP network based transport this "profile" is already there just by virtue of it being designed directly on top of the websocket protocol.

Thoughts?

zherczeg:
Current requirement is min 64 byte packet with reliable transport, so BLE is probably excluded.

I am neutral about profiles.

martijnthe:
> Current requirement is min 64 byte packet with reliable transport, so BLE is probably excluded.

FWIW, plain GATT (layer immediately below the "application" protocol) allows for MTUs up to 510 bytes GATT payload but not all platforms/stacks support it up to that size. AFAIK, iOS goes up to 155 bytes. On top of that, the default is only 20 or so and an MTU exchange needs to happen first, which some BLE device stacks don't support (and even some poor Android devices also don't)...

Anyhow, BLE is probably not the first candidate just because it's more involved than the other ones (USB, UART, ...)

jimmy-huang:
[...]

PR #2247, opened 2018-03-20
Debugger Transport APIs and Zephyr port

jimmy-huang:
This is patch to tackle issue #2229, basically we want the ability to have debugger support different platforms and different transports.

The current debugger implementation only supports BSD sockets, and it's not portable to different OSes that have different platform APIs.  Besides, other platforms may want to use another transport layer like USB or Bluetooth instead of TCP sockets.  The goal is to create a generic transport interface that different platforms can be used, and provide the actual implementation for running on their devices.  This patch provides a struct that defines the function signature that each transport can be used to register their own implementation of the socket APIs using the jerry_port_xxx interface and a Zephyr port is provided as a sample.

This is a WIP patch and a reference implementation, but I would like use this as a starting point for discussion on how to improve the APIs from here, would like to have feedbacks from @zherczeg and @martijnthe and JerryScript team after reviewing them.


martijnthe:
@jimmy-huang some builds are failing (click "Details" to learn more) ^^^

jerry-core/include/jerryscript-port.h
```diff
@@ -138,6 +139,50 @@ double jerry_port_get_current_time (void);
...
+  jerry_debugger_conn_errors_t (*send) (jerry_debugger_conn_t *connection_p, const void *data_p, size_t data_len, ssize_t *bytes_sent);
+  jerry_debugger_conn_errors_t (*receive) (jerry_debugger_conn_t *connection_p, void *data_p, size_t max_len, ssize_t *bytes_received);
+  jerry_debugger_conn_errors_t (*close) (jerry_debugger_conn_t *connection_p);
+};
+
+/**
+ * The default transport layer that uses TCP socket
+ * other transports can be added like usb, bluetooth
+ */
+jerry_debugger_transport_t *jerry_debugger_init_socket_transport(uint16_t tcp_port);
```
martijnthe: Should this be in `jerryscript-port.h`?
It's specific to the socket transport implementation, no?

jerry-port/default/default-socket.c
```diff
@@ -0,0 +1,246 @@
...
+};
+
+/**
+ * Create and return the socket transport on the provided port for the debugger
+ *
+ * @return the transport created
+ */
+jerry_debugger_transport_t*
+jerry_debugger_init_socket_transport(uint16_t tcp_port) /**< server port number */
+{
```
martijnthe: Perhaps add an assert to make sure it's only called once (since the `socket_transport` state is global/singleton).

targets/zephyr/src/jerry-port.c
```diff
@@ -16,9 +16,13 @@
...
 
 #include <zephyr.h>
+#ifdef JERRY_DEBUGGER
+#include <sys/fcntl.h>
+#include <net/socket.h>
+#endif
 
 #include "jerryscript-port.h"
-
+#include "jrt.h"
```
martijnthe: What is used from this header?
It's a private/internal header.

martijnthe:
@zherczeg could you take a look at the design?

jerry-port/default/default-socket.c
martijnthe: Hmm I see, ok, I guess it's ok.

There's nothing precluding one from creating a dynamically created transport (and having a transport-specific cleanup function to free/destroy the transport), so I'm happy.

targets/zephyr/src/jerry-port.c
martijnthe: Sounds reasonable.

martijnthe: > JERRY_UNUSED

Hmm I guess it's not part of the "API"

Maybe just use `(void) variable_that_is_unused`?

jerry-core/api/jerry-debugger.c
```diff
@@ -93,13 +93,18 @@ jerry_debugger_stop_at_breakpoint (bool enable_stop_at_breakpoint) /**< enable/d
  * Debugger server initialization. Must be called after jerry_init.
  */
 void
-jerry_debugger_init (uint16_t port) /**< server port number */
+jerry_debugger_init (jerry_debugger_transport_t *transport_p) /**< transport */
```
zherczeg: Is this port specific definition part of the public API?

jerry-core/include/jerryscript-debugger.h
```diff
@@ -17,6 +17,7 @@
 #define JERRYSCRIPT_DEBUGGER_H
 
 #include <stdbool.h>
+#include "jerryscript-port.h"
```
zherczeg: This is definitely not good here.

jerry-core/debugger/debugger-ws.c
```diff
@@ -65,24 +60,20 @@ typedef struct
  * Close the socket connection to the client.
  */
 static void
-jerry_debugger_close_connection_tcp (bool log_error) /**< log error */
+jerry_debugger_close_connection_transport (void)
```
zherczeg: Is there a plan to log connection error in the future?

jerry-core/debugger/debugger-ws.c
```diff
@@ -291,14 +293,14 @@ jerry_process_handshake (int client_socket, /**< client socket */
 
   text_p = "HTTP/1.1 101 Switching Protocols\r\nUpgrade: websocket\r\nConnection: Upgrade\r\nSec-WebSocket-Accept: ";
 
-  if (!jerry_debugger_send_tcp ((const uint8_t *) text_p, strlen (text_p))
-      || !jerry_debugger_send_tcp (request_buffer_p + sha1_length + 1, 27))
+  if (!jerry_debugger_send_transport ((const uint8_t *) text_p, strlen (text_p))
+      || !jerry_debugger_send_transport (request_buffer_p + sha1_length + 1, 27))
```
zherczeg: This is quite websocket specific. As far as I know websockets is available only through normal sockets.

jerry-core/include/jerryscript-port.h
```diff
@@ -138,6 +138,54 @@ double jerry_port_get_current_time (void);
...
+/*
+ * Debugger Port API
+ */
+
+/**
+ * A forward declaration of various debugger structures.
+ */
+typedef struct jerry_debugger_transport_t jerry_debugger_transport_t;
+typedef struct jerry_debugger_config_t jerry_debugger_config_t;
+typedef struct jerry_debugger_conn_t jerry_debugger_conn_t;
```
zherczeg: Why do we need to forward declare these functions?

jerry-core/jcontext/jcontext.h
```diff
@@ -116,9 +116,9 @@ typedef struct
   jmem_cpointer_t debugger_byte_code_free_tail; /**< tail of byte code free linked list */
   uint32_t debugger_flags; /**< debugger flags */
   uint16_t debugger_receive_buffer_offset; /**< receive buffer offset */
-  uint16_t debugger_port; /**< debugger socket communication port */
   uint8_t debugger_message_delay; /**< call receive message when reaches zero */
-  int debugger_connection; /**< holds the file descriptor of the socket communication */
+  jerry_debugger_transport_t *debugger_transport_p; /**< holds the pointer to the debugger transport */
+  jerry_debugger_conn_t *debugger_connection_p; /**< holds the pointer to the debugger connection */
```
zherczeg: Do we need to separate these two?

jerry-core/api/jerry-debugger.c
jimmy-huang: Do you mean the jerry_debugger_transport_t type?  Yes, this is intended and part of the jerry_port_ api where each platform have to provide their own definition, hence why it was forward declared, so it can build.

jerry-core/debugger/debugger-ws.c
jimmy-huang: yeah, we can just change to all the handlers Apis to take the same log_error parameter

jerry-core/debugger/debugger-ws.c
jimmy-huang: i could leave it with just  _tcp if it's ok, @martijnthe ?

jerry-core/include/jerryscript-port.h
jimmy-huang: these needs to be forward declared because it's up to the jerry_port layers to provide the definitions for these, it will run into compile errors if debugger is not enabled.

jerry-core/jcontext/jcontext.h
jimmy-huang: we can probably keep just the transport obj, and have the connection be part of the transport, @martijnthe ?

jerry-core/api/jerry-debugger.c
zherczeg: Unfortunately "it can build" is not a strong enough reason for something to be part of the public API. 

While I am ok with having a structure for communication, the public API part needs to be independent of everything: no jerry port dependency, no compiler define dependency, etc.

I think the first question is: do we want multiple transport layer support, or the recv/send/close could be provided by jerry port.

jerry-core/include/jerryscript-debugger.h
jimmy-huang: We can change the init() to take a void*, thus avoid pulling in jerryscript-port.h

jerry-core/api/jerry-debugger.c
jimmy-huang: @zherczeg When I originally started the the port, it was a simpler patch to just port the socket APIs, but I think @martijnthe really like to support multiple transports, so we kinds modified it to take a struct of handlers.  I mean we can submit a simpler patch to ports only socket calls and then a separate patch later to rework the APIs to support multiple transports, I think it's up to you and the JerryScript maintainers to make this kind of decision.  Honestly, my goal is to at least get the websockets working on Zephyr as part of this effort.  If we move the the struct definitions and APIs to from jerryscript-port to jerryscript-debugger.h, would that work for you?

jerry-core/jcontext/jcontext.h
martijnthe: Yeah, I think so.

I don't see a use case for being able to distinguish the transport and the connection.

I guess in a multi-Jerry-context one could have multiple connections, sharing the same transport implementation. But one could work around that by just creating separate `jerry_debugger_transport_t`s instead I guess.

jerry-core/include/jerryscript-port.h
martijnthe: The struct definitions itself is up to the implementation of the transport. Initially, @jimmy-huang had used `void *`, but I suggested to use forward declarations. That way the compiler will still help with ensuring the correct types are used.

jerry-core/debugger/debugger-ws.c
martijnthe: > As far as I know websockets is available only through normal sockets.

At this point I'm regarding the websocket framing and handshaking to be part of the JrS debugger protocol as you yourself suggested a year ago: https://github.com/jerryscript-project/jerryscript/pull/1557#issuecomment-277245868

jerry-core/debugger/debugger-ws.c
martijnthe: It looks like only this function has such an option.

jerry-core/include/jerryscript-debugger.h
martijnthe: Or repeat the forward declaration?
Or move the forward declaration into an API header?
@zherczeg What's the best do you think?

martijnthe: Or repeat the forward declaration?
Or move the forward declaration into an API header and include that?
@zolWhat's the best do you think?

jerry-core/api/jerry-debugger.c
martijnthe: > Is this port specific definition part of the public API?

The *definition* should __NOT__ be part of the public API, only the forward declaration should be. That way there is no dependency. (Until one actually want to use the debugger feature, in which case one has to pick an implementation which would contain the definition of the struct as well.)

> I think the first question is: do we want multiple transport layer support

Yes, at least in our project we want to mix and match between 3 transports at run-time.

IMHO, a stronger reason for wanting a more dynamic transport interface as opposed to extending the "port surface" with static functions for the transport, is that static functions would make combining, composing and re-using transport implementations harder.

For example in our case, we want USB, BLE and a "pseudo transport" that just gathers the debug info (i.e. for printing human consumable exception traces). Imagine that someone already wrote each of these transports separately. Each transport would define the same set of functions/symbols. If I want to create a new transport that combines/chains/... them, I'd have to rename each function to avoid conflicts. Even if a simple "switch" between 2 transports would involve renaming each function.

jerry-core/debugger/debugger-ws.c
zherczeg: It was a long time ago when I worked on this. But it seems a reasonable idea. Probably the handshake could be an optional feature.

zherczeg:
```
JERRY_DEBUGGER_INIT_SEND_MESSAGE
JERRY_DEBUGGER_SET_SEND_MESSAGE_SIZE_FROM_TYPE
JERRY_DEBUGGER_SET_SEND_MESSAGE_SIZE

jerry_debugger_accept_connection
jerry_debugger_close_connection
jerry_debugger_send
jerry_debugger_receive

jerry_debugger_send_header_t
```
It seems these types/functions are used by the debugger. If we want to stick to the websocket package format, perhaps the first step would be to move certain of these types into the main debugger, reducing  the content of `debugger-ws.h` as much as possible. If not, the transport system must provide a way to provide these declarations.

zherczeg:
By the way the send / receive already working in different ways, since the send aware of the low-level websocket header format while it is stripped from the receive part. Perhaps the `JERRY_DEBUGGER_INIT_SEND_MESSAGE` could be removed first, and moved internally to `jerry_debugger_send` in `debugger-ws.c`. One less macro to worry about.

martijnthe:
Re. keeping websocket packet framing:

I'm OK with either, keeping it or moving it out. I do think that it's a tad bit wasteful / annoying if the transport itself also (inherently) takes care of packet framing (I'm thinking about BLE for example where every byte saved in the air is important).
It was just your past comment that made me take the initial position of "OK the websocket framing is part of the JrS debug protocol".

If we decide to move it out, then I agree, there's quite a bit of rework needed to separate the WS related stuff out and into the transport side of the fence.

If we decide to keep WS and regard it as part of the JrS debug protocol, I'm not sure why it's needed to refactor the `debugger-ws.c`/`debugger.c` split in order to progress with the transport layer abstraction. IMHO, cleaning up those files would be better off as a separate PR.

@jimmy-huang thoughts?

@zherczeg So it sounds like a decision needs to be made about "websocket framing is part of JrS debug protocol": yes or no. Any other people that need to chime in to help make that decision?

jimmy-huang:
@martijnthe Yeah, personally, i rather keep the WS at least for now, because initially I think most of the platforms will just re-use the same WS protocol, and like you said, it's a lot of work for each port to add these WS.   I think the WS and both the TCP socket can be consider as a transport, so for now if we just move the TCP logic into a tcp transport, then later we can then move the WS into a separate transport as well.  But I am okay with either as well.

zherczeg:
I would first remove as much macros / types from debugger-ws.h as possible. Probably in separate PR-s first. JERRY_DEBUGGER_INIT_SEND_MESSAGE could go away easily (send should write this byte). The JERRY_DEBUGGER_SET_SEND_MESSAGE_SIZE_FROM_TYPE and 
JERRY_DEBUGGER_SET_SEND_MESSAGE_SIZE macros probably could go away by adding a size parameter to send. Basically the transport protocol should specify how much (maximum) bytes should be reserved for the header in the send buffer. This should be a constant of course (if less bytes are needed for the actual message, just internally adjust the start before sending the packet).

jimmy-huang:
@zherczeg where does jerry_debugger_send_header_t go?  Should it be moved to debugger.h? or it's a internal thing that's in debugger-ws.c?  I am guessing the former?

jimmy-huang:
@zherczeg I submitted another patch #2262 to remove some of these macros.  Is the goal here to make the debugger-ws.c as the default transport and remove all the macros and structurs in debuggr-ws.h.  What about these structs? 

jerry_debugger_send_header_t
jerry_debugger_receive_uint8_data_part_t
jerry_debugger_uint8_data_t

martijnthe:
> Basically the transport protocol should specify how much (maximum) bytes should be reserved for the header in the send buffer. This should be a constant of course (if less bytes are needed for the actual message, just internally adjust the start before sending the packet). The size passed to send excludes this amount of course.

@jimmy-huang @zherczeg Alternative idea: instead of letting the transport specify the max header size, how about if the debugger instead would ask the transport for a pointer to a send buffer? The transport itself would take care of reserving space for a header. In case sending happens async, it could also save a copy to another buffer + lower the peak memory usage a little (perhaps not relevant though). Downside is that getting a buffer would become a function call instead of just getting a pointer + offset from the context, which in the simplest of projects would be at a fixed memory location.

I'm not feeling strongly about either, but wanted to share the idea.

jimmy-huang:
@zherczeg I've started rebase my patch on top of your patch #2266, I've took out the Zephyr port for now to simplify the patch.   Is there a need to move out the rest of the stuff in debugger-ws.h? for example, the jerry_debugger_uint8_data_t structs and jerry_debugger_compute_sha1() could move to debugger.h?

JERRY_DEBUGGER_MAX_BUFFER_SIZE could also be defined somewhere else? or only set it if not defined in the transport layer.

targets/tizenrt-artik053/apps/jerryscript/jerry_main.c
```diff
@@ -508,6 +508,17 @@ jerryx_port_handler_print_char (char c) /**< the character to print */
...
 
+/**
+ * Dummy function to return the transport.
+ *
+ * @return NULL
+ */
+jerry_debugger_transport_t *
+jerry_port_init_socket_transport (uint16_t tcp_port) /**< server port number */
+{
+  return NULL;
```
martijnthe: How was the tizen debugger connection working before?

martijnthe:
@zherczeg @jimmy-huang  thoughts?

> the jerry_debugger_uint8_data_t structs and jerry_debugger_compute_sha1() could move to debugger.h?

`jerry_debugger_compute_sha1()` is websocket specific? I think it should stay in the ws files.

> I guess there are a little more refactoring needed for the transport layer as it's calling internal APIs like JERRY_CONTEXT, so those would needed be moved out of there as well.

Yeah, at a glance, it looks like it's used to get 1) information that the transport itself already has/knows (i.e. max sizes, connection pointer, ...) 2) the send/receive buffers.

Re. send/receive buffers: I wonder if it makes more sense to let the transport "own" the buffers and let JrS *ask* the transport for a buffer when it needs one. That way the data can be copied directly into the buffer that ends up being used by the communications stack, instead of writing it into the JCONTEXT buffer and later on copying it to the buffer of the comms stack (most comms stacks that I've seen, bring their own buffer management APIs).

`JERRY_DEBUGGER_MAX_BUFFER_SIZE`: I think that should also be part of the transport interface (`jerry_debugger_transport_t`).

targets/tizenrt-artik053/apps/jerryscript/jerry_main.c
jimmy-huang: @martijnthe I actually don't think the tizen and nuttx sample apps built with debugger enabled.  I checked the Makefiles and grepped their directory and don't see it compiles with the debugger flags.

jimmy-huang:
@martijnthe I've moved the generic debugger logic into the debugger.c so the transport layer doesn't have to depend on any of the internal context.  I've try to remove as many dependencies from the transport except for as the jerry_debugger_compute_sha1(), which I think it's okay to leave it as is, and other platforms can build with it if needed since it is not really platform-specific.  In ZJS, I have another patch that basically ports the websocket transport layer to Zephyr as you can see here https://github.com/intel/zephyr.js/pull/1865

zherczeg:
> I guess there are a little more refactoring needed for the transport layer as it's calling internal APIs like JERRY_CONTEXT, so those would needed be moved out of there as well.

This is the next step (patch). We need to invent something for that.

jimmy-huang:
@zherczeg I am passing the buffer to the transport layer, the debugger_receive () is actually a little tricky as it has to jerry_debugger_process_message () while keeping track of the offset of the buffer, so I had to pass that to the transport layer and have it updated as well,  see @f66f0b48859f8dcb8ca1adf78129d39b91895b87

zherczeg:
What about using calls to the debugger to set the global variables? This way the transport layer can set variables without access to the global variables. It would be good to do in a separate patch.

jerry-core/include/jerryscript-port.h
```diff
@@ -138,6 +138,39 @@ double jerry_port_get_current_time (void);
...
+                   size_t *data_size,
+                   uint32_t *buffer_offset);
+};
+
+/**
+ * The default transport layer that uses TCP socket
+ * other transports can be added like usb, bluetooth
+ */
+struct jerry_debugger_transport_t *jerry_port_init_socket_transport (uint16_t tcp_port);
+
```
martijnthe: I'm not 100% sure whether this should be in `jerryscript-port.h` because it's specific to an implementation of a transport (of a TCP socket transport in this case).

This is comparable/similar to the various `jerry_port_default_...` functions in https://github.com/jerryscript-project/jerryscript/blob/master/jerry-port/default/include/jerryscript-port-default.h

jerry-port/default/default-debugger-transport.c
```diff
@@ -13,15 +13,15 @@
  * limitations under the License.
  */
 
-#include "debugger.h"
-#include "jcontext.h"
 #include "jerryscript-port.h"
+#include "jmem.h"
```
martijnthe: Do you still need `jmem.h` now?

jerry-port/default/default-debugger-transport.c
```diff
@@ -13,15 +13,15 @@
...
 
 #ifdef JERRY_DEBUGGER
 
 #include <arpa/inet.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <string.h>
```
martijnthe: After this PR, I would split this file up such that you have one platform-agnostic (shared) file and one with the (Linux) platform specific stuff separated out into a different file. That way you can re-use the shared file as-is instead of having to copy-paste so much code in a project that uses TCP/websockets but doesn't have a Linux-y API (such as Zephyr).

martijnthe:
Passing the receive buffer to the transport to write the received data into is seems quite natural and logical to me actually. Updating the offset seems awkward to me, I left a comment on a suggestion for a nicer solution.

@zherczeg I think this is getting close to being done. Can you take another serious look?

jerry-port/default/default-debugger-transport.c
jimmy-huang: this is because for the handshake part, it is allocating some buffer using JMEM_DEFINE_LOCAL_ARRAY, on other platforms like Zephyr, I just call into malloc, but for the Jerryscript default transport, it might make sense to get the memory from the JrS heap since malloc may not be present.  @zherczeg?

jimmy-huang:
@martijnthe yeah, when i tried to refactor it, it was a little awkward, since I wasn't really sure that loop to constantly adjust the offset to the receive buffer was part of the debugger protocol or part of the websocket protocol, and I trying to preserve the originally behavior without breaking the core logic.  I am pretty sure I can enhance it a bit, maybe if @zherczeg can take a look at it and see what the receive API should look like, then I can adapt it to use that API?

jerry-port/default/default-debugger-transport.c
jimmy-huang: Yeah, that makes sense

jerry-core/include/jerryscript-port.h
jimmy-huang: @martijnthe thanks for catching that, i updated patch

martijnthe:
> have multiple transports at the same time? Like what @martijnthe described?

I think only one can be "actively" used at the same time. What we would like for our use case is be able to switch between transports (i.e. USB vs BLE) w/o recompiling anything. In other words, at runtime decide what transport to pass into `jerry_debugger_init`.

zherczeg:
@jimmy-huang I would not use .something fields. Simply create some functions such as `jerry_debugger_set_transmit_sizes` with 3 arguments. Such as @martijnthe said, we should not be able to switch transport after it is set for a jerry instance. I would like to some suggestions how could we remove jcontext.h dependency from debugger-ws.c and do it in a separate patch.

martijnthe:
@zherczeg 

> I would like to some suggestions how could we remove jcontext.h dependency from debugger-ws.c and do it in a separate patch.

I'm not sure if I understand. This PR removes the jcontext.h dependency from debugger-ws.c already? Would you like to see this PR split in 2 separate ones? If so, what should be left out?

Please clarify. Thank you.

jimmy-huang:
@zherczeg @martijnthe Updated with jerry_debugger_set_transmit_sizes function

jimmy-huang:
@zherczeg Can take take a another look at the patch?

zherczeg:
The problem is not with the patch. The problem lies in nicely restructuring the debugger interface. I spent some hours to decouple the ws part from jcontext, but I was not satisfied with the results. It worked, but the code was ugly. Unfortunately this must be the first step. So this is the challenge here. If you have time, you can try it.

jimmy-huang:
@zherczeg Ok, that's essentially what I am doing with one of the commits, so I separate it out as a separate patch, maybe this is what you are looking?

https://github.com/jimmy-huang/jerryscript/tree/refactor-ws

zherczeg:
Not exactly. I would like to be able to chain this at the top of other interfaces, e.g. tcp. Maybe move it to some jerry extension or port. I am not sure.

jerry-core/debugger/debugger-ws.h
```diff
@@ -38,18 +38,12 @@ typedef struct
 /**
  * Byte data for evaluating expressions and receiving client source.
  */
-typedef struct
+typedef struct jerry_debugger_uint8_data_t
```
akosthekiss: This change seems to be unjustified. Current code base adds names to those structs only that are recursive. Here, it's not the case.

jerry-core/include/jerryscript-port.h
```diff
@@ -150,6 +150,30 @@ double jerry_port_get_current_time (void);
  */
 struct jerry_instance_t *jerry_port_get_current_instance (void);
 
+/*
+ * Debugger Port API
+ */
+
+/* forward declaration */
+struct jerry_debugger_uint8_data_t;
```
akosthekiss: This forward declaration also seems to be unnecessary, especially in this public header. The type is only used in sources internal to jerry-core. If I see it right.

jimmy-huang:
@zherczeg @akosthekiss Thanks for the feedback, been out for a while and just got back from vacation.  We actually have a work-around building debugger on top of Zephyr's native posix network stack, since they are almost identical.  So in our project, we have this custom header that get injected when debugger is turned on.  I can probably submit a patch to the zephyr target as well, is this something you guys want? Since at this point, we don't need to support other transports besides tcp.  This will not have any changes on the API too.

 https://github.com/intel/zephyr.js/blob/master/src/jerry-port/arpa/inet.h

zherczeg:
Unfortunately I have no time to work on this at the moment. The whole debugger interface needs a major redesign to satisfy all requirements. I would recommend to use your workaround until we can do this.

zherczeg:
https://github.com/zherczeg/jerryscript/tree/debugger_transport

Something to start with. It is far from compiling though. This will take a long time to finish.

zherczeg:
Please check #2421

zherczeg:
Hi @jimmy-huang, after #2426 lands you can freely add your low-level transmission protocols to JerryScript. Please check the patch if you have time.

jimmy-huang:
@zherczeg Yeah, thanks for the heads up, I am actually ramping up on another project atm, but I'll try to look at this later when I have time, if you think i should close this PR first, I am okay with it and I'll re-submit a pr later.

jimmy-huang:
closing

PR #2248, opened 2018-03-21
Remove ineffective codes

haesik:
Results of assert statements are always true,
I think we don't need these assert statements anymore

JerryScript-DCO-1.0-Signed-off-by: Haesik Jun haesik.jun@samsung.com


LaszloLango:
@haesik could you explain why do you think those asserts are unnecessary?

haesik:
~LIT_UTF8_3_BYTE_CODE_POINT_MAX is 0, so the chr don't affect the result.
The assert statement is always true whatever the value of chr is..
If the assert statement has a different meaning, please let me know.

yichoi:
LGTM

PR #2249, opened 2018-03-21
Add tests for TypedArrays by using indexing API.

imiklos:

### DIFF
--- a/tests/unit-core/test-typedarray.c
+++ b/tests/unit-core/test-typedarray.c
@@ -303,6 +303,121 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
   }
 } /* test_typedarray_complex_creation */
 
+/**
+ * Test get/set/delete property by index.
+ */
+static void test_property_by_index (test_entry_t test_entries[])
+{
+  int test_int_numbers[5] = {-5, -70, 13, 0, 56};
+  double test_double_numbers[5] = {-83.153, -35.15, 0, 13.1, 89.8975};
+  uint8_t test_uint_numbers[5] = {83, 15, 36, 0, 43};
+
+  for (uint32_t i = 0; test_entries[i].constructor_name != NULL; i++)
+  {
+    jerry_value_t test_number;
+    uint32_t test_numbers_length = sizeof (test_int_numbers) / sizeof (int);
+    jerry_value_t typedarray = jerry_create_typedarray (test_entries[i].typedarray_type, test_numbers_length);
+    jerry_typedarray_type_t type = jerry_get_typedarray_type (typedarray);
+
+    jerry_value_t set_result;
+    jerry_value_t get_result;
+
+    switch (type)
+    {
+      case JERRY_TYPEDARRAY_INT8:
+      case JERRY_TYPEDARRAY_INT16:
+      case JERRY_TYPEDARRAY_INT32:
+      {
+        for (uint8_t j = 0; j < test_numbers_length; j++)
+        {
+          test_number = jerry_create_number (test_int_numbers[j]);
+          TEST_ASSERT (!jerry_delete_property_by_index (typedarray, j));
+          set_result = jerry_set_property_by_index (typedarray, j, test_number);
+          get_result = jerry_get_property_by_index (typedarray, j);
+
+          TEST_ASSERT (jerry_value_is_boolean (set_result));
+          TEST_ASSERT (jerry_get_boolean_value (set_result));
+          TEST_ASSERT (!jerry_delete_property_by_index (typedarray, j));
+          TEST_ASSERT (jerry_get_number_value (get_result) == test_int_numbers[j]);
+
+          jerry_release_value (test_number);
+          jerry_release_value (set_result);
+          jerry_release_value (get_result);
+        }
+        break;
+      }
+      case JERRY_TYPEDARRAY_FLOAT32:
+      case JERRY_TYPEDARRAY_FLOAT64:
+      {
+        for (uint8_t j = 0; j < test_numbers_length; j++)
+        {
+          test_number = jerry_create_number (test_double_numbers[j]);
+          TEST_ASSERT (!jerry_delete_property_by_index (typedarray, j));
+          set_result = jerry_set_property_by_index (typedarray, j, test_number);
+          get_result = jerry_get_property_by_index (typedarray, j);
+
+          TEST_ASSERT (jerry_value_is_boolean (set_result));
+          TEST_ASSERT (jerry_get_boolean_value (set_result));
+          TEST_ASSERT (!jerry_delete_property_by_index (typedarray, j));
+
+          double epsilon = pow (10, -5);
+          double get_abs = fabs (jerry_get_number_value (get_result) - test_double_numbers[j]);
+          TEST_ASSERT (get_abs < epsilon);
+
+          jerry_release_value (test_number);
+          jerry_release_value (set_result);
+          jerry_release_value (get_result);
+
+          /* Testing positive and negative infinity */
+          for (uint8_t k = 0; k < 2; k++)
+          {
+            jerry_value_t inf = jerry_create_number_infinity (k);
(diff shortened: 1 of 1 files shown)


tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,44 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+
+  jerry_value_t ja = jerry_create_number (3);
+  jerry_value_t jb = jerry_create_number (6);
+  jerry_value_t jc = jerry_create_number (50);
+  jerry_value_t global_obj_val = jerry_get_global_object ();
+
+  for (uint32_t i = 0; test_entries[i].constructor_name != NULL; i++)
+  {
+    jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) test_entries[i].constructor_name);
+    jerry_value_t prop_value = jerry_get_property (global_obj_val, prop_name);
```
galpeter: There is no need to access the typedarray via the global object, we have a typedarray constructor function: http://jerryscript.net/api-reference/#jerry_create_typedarray.

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,44 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+    jerry_value_t length_arg = jerry_create_number (test_entries[i].element_count);
+
+    jerry_value_t typedarray = jerry_construct_object (prop_value, &length_arg, 1);
+
+    jerry_release_value (prop_name);
+    jerry_release_value (prop_value);
+    jerry_release_value (length_arg);
+
+    jerry_set_property_by_index (typedarray, 0, ja);
```
galpeter: The `..set_property_...` call returns a value, which we also need to check.

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,44 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+    jerry_set_property_by_index (typedarray, 1, jb);
+    jerry_set_property_by_index (typedarray, 2, jc);
+
+    TEST_ASSERT (jerry_get_number_value (ja) == jerry_get_number_value (jerry_get_property_by_index (typedarray, 0)));
```
galpeter: As the `jerry_get_property_by_index` returns a `jerry_value_t` we need to release that (see: http://jerryscript.net/api-reference/#jerry_get_property_by_index)

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,44 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+    TEST_ASSERT (jerry_get_number_value (jc) == jerry_get_number_value (jerry_get_property_by_index (typedarray, 2)));
+    TEST_ASSERT (!jerry_delete_property_by_index (typedarray, 1));
```
galpeter: We should also test a case where we are removing an element on an unset index (for example: the 5th index was not set, but we still try to delete it)

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,44 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+    jerry_value_t length_arg = jerry_create_number (test_entries[i].element_count);
+
+    jerry_value_t typedarray = jerry_construct_object (prop_value, &length_arg, 1);
+
+    jerry_release_value (prop_name);
+    jerry_release_value (prop_value);
+    jerry_release_value (length_arg);
+
+    jerry_set_property_by_index (typedarray, 0, ja);
+    jerry_set_property_by_index (typedarray, 1, jb);
```
galpeter: We also need to test the case when the index is out of bounds, which should return what?

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,44 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
 
+/**
+ * Test get/set/delete property by index.
+ */
+static void test_property_by_index (test_entry_t test_entries[])
+{
+
+  jerry_value_t ja = jerry_create_number (3);
+  jerry_value_t jb = jerry_create_number (6);
+  jerry_value_t jc = jerry_create_number (50);
```
galpeter: We should improve this to test more values not just simple integers. We need negative numbers, zero, inf/-inf, double values. Having multiple values we can create an array of test values (similar to the test_entries array) thus we would not need multiple variables.

imiklos:
@galpeter Thank you for the reviews. I updated the PR.

imiklos:
@galpeter Thanks for the reviews. I updated the PR.

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,102 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+    jerry_typedarray_type_t type = jerry_get_typedarray_type (typedarray);
+
+    switch (type)
+    {
+      case JERRY_TYPEDARRAY_INT8:
+      case JERRY_TYPEDARRAY_INT16:
+      case JERRY_TYPEDARRAY_INT32:
+      for (uint32_t j = 0; j < test_numbers_length; j+=2)
+      {
+        test_numbers[j] = jerry_create_number (0 - jerry_get_number_value (test_numbers[j]));
```
zherczeg: What happens with the original number? I know you don't need to free them, but still you should according the documentation.

Btw, you could have a simple const number list, not an ecma_value_t list. This way you can avoid the release problem.

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,102 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+      TEST_ASSERT (!jerry_delete_property_by_index (typedarray, j));
+      jerry_value_t set_result = jerry_set_property_by_index (typedarray, j, test_numbers[j]);
+      jerry_value_t get_result = jerry_get_property_by_index (typedarray, j);
+
+      TEST_ASSERT (jerry_value_is_boolean (set_result));
+      TEST_ASSERT (jerry_get_boolean_value (set_result));
+      TEST_ASSERT (!jerry_delete_property_by_index (typedarray, j));
+      if (type == JERRY_TYPEDARRAY_FLOAT32 || type == JERRY_TYPEDARRAY_FLOAT64)
+      {
+        double epsilon = pow (10,-7);
```
zherczeg: Missing space before -7.

imiklos:
@galpeter @zherczeg I updated the PR.

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,107 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+      {
+        test_numbers[j] = jerry_create_number (test_double_numbers[j]);
+      }
+      break;
+      default:
+      for (uint8_t j = 0; j < test_numbers_length; j++)
+      {
+        test_numbers[j] = jerry_create_number (test_uint_numbers[j]);
+      }
+      break;
```
zherczeg: Question: can we do this in one step? Create a single jerry value and set/get it?

imiklos:
@zherczeg I updated the PR, made changes to set and get values in one step.

tests/unit-core/test-typedarray.c
```diff
@@ -303,6 +303,115 @@ test_typedarray_complex_creation (test_entry_t test_entries[], /**< test cases *
...
+
+    jerry_value_t set_result;
+    jerry_value_t get_result;
+
+    switch (type)
+    {
+      case JERRY_TYPEDARRAY_INT8:
+      case JERRY_TYPEDARRAY_INT16:
+      case JERRY_TYPEDARRAY_INT32:
+      for (uint8_t j = 0; j < test_numbers_length; j++)
```
zherczeg: I think the style excepts two spaces indentation after `case`

imiklos:
@zherczeg I updated the PR, fixed the indentation after case.

PR #2259, opened 2018-03-28
Rework snapshot generation API.

zherczeg:


docs/02.API-REFERENCE.md
```diff
@@ -54,6 +54,48 @@ Possible compile time enabled feature types:
...
+static snapshots. The literals of these snapshots are limited to magic
+strings, and 28 bit signed integers, so their constant pools do not need
+to be loaded into the memory. Hence these snapshots can be executed from
+ROM.
+
+***Important note:*** the magic strings set by
+[jerry_register_magic_strings](#jerry_register_magic_strings) must be the
+same when the snapshot is generated and executed. Furthermore the
+`copy_bytecode` flag of [jerry_exec_snapshot_at](#jerry_exec_snapshot_at)
+must be set to `false`.
```
martijnthe: I have not yet looked at the changes to implementation, but:
Is there any detection/protection "up front" to prevent a snapshot from executing when the magic strings in the run time don't match? I'm thinking of something like a field in the snapshot header with a hash of all the magic strings, or something along those lines.

zherczeg:
Question: shall we remove JERRY_SNAPSHOT_SAVE_EVAL ?

Reason: it does not work.

Consider the folowing function:
`function f(a) { eval("something"); }`
The engine is aware that the `eval` might access `a`, so it creates a lexical context with a binding called `a`.

Consider its native variant:
`function f(a) { native_eval("something"); }`
Here the engine does not aware of a presence of an `eval`, and places `a` into a register and forgets its name. Accessing `a` from `native_eval` will result an undefined reference.

I am also not aware of a use case where it is useful.

docs/02.API-REFERENCE.md
martijnthe: Yeah, that makes sense.
So what will happen now if there is a mismatch? Will it fail gracefully? (at least not crash?)

zherczeg: May crash if there is a magic string with higher index the current maximum index (magic_string[idx] cause buffer overflow). Other than that, the identifiers / strings will be wrong.

martijnthe: Maybe it makes sense to add the type of snapshot (static vs non-static vs ...) to the header of the snapshot? That way one can at least check the header and not call `jerry_exec_snapshot` in case it's a static snapshot and it's uncertain whether the snapshot was generated with the same set of magic strings.

martijnthe:
Re. `native_eval` scope example: doesn't that apply to any use of `jerry_eval` in general?

docs/02.API-REFERENCE.md
zherczeg: Good point. We could add a getter function which returns the generator flags (static/strict). Probably in a follow-up patch though (you can do it if you have time :) ).

zherczeg:
@martijnthe yes, good point. I checked the code and realized that `is_direct` is set to `false` in both cases, which means they are evaluated globally now. Therefore there is no executing difference between calling jerry_run or jerry_eval. Furthermore the snapshot eval flag has no effect, and its description is wrong.

martijnthe:
Ah, OK, it's already an indirect eval.
I'm fine with removing `JERRY_SNAPSHOT_SAVE_EVAL`.

docs/02.API-REFERENCE.md
martijnthe: Later is fine.

I don't see the snapshot header version being bumped in this PR.
I would do that at the very least?

zherczeg: There is no reason to do it (this is strictly an API change) unless people agree to remove the eval thing.

Anyway I noticed one difference: in strict mode, global eval does not create new variables in the global scope unless you use the `this.var` form (creates a new local context below the global context). This could be a reason to keep the eval. 

What shall we do?

zherczeg: > Didn't snapshots before this change always include string literals even if they were magic
 > strings?

Static snapshort support was landed in another change: #2239 

Would be too much for this patch.

martijnthe: 👍 ignore me!

zherczeg:
A context is always inconvenient to use :( The sanpshot saving is used on desktop anyway.

martijnthe:
> The sanpshot saving is used on desktop anyway.

JFYI, in our case it's not (and I suspect in Fitbit's case it's also used on-device).

zherczeg:
Does the device has 64K RAM?

martijnthe:
I can't give any specific numbers, but in our case there's enough to do it on-device for our use-case, yes.

zherczeg:
Well if you have several mbytes of ram 9 arguments should not be a problem.

martijnthe:
No, it's not a problem. I'm fine with it.
Just wanted to inform you about that it's not only used on a desktop.

PR #2261, opened 2018-03-29
Fix accessing the contents of a direct string

galpeter:
In the `ecma_string_get_chars` method
the contents of a direct string is accessed incorrectly.
It tries to extract the magic string id from the
string pointer but the direct string does not need
this step.


galpeter:
I've did an update: now there is a test for it.

PR #2266, opened 2018-04-03
Simplify debugger-ws.h

zherczeg:
Remove several macros and types from it. This change simplify the required debugger interface.


martijnthe:
cc @jimmy-huang

jerry-core/debugger/debugger-ws.c
```diff
@@ -321,6 +337,25 @@ jerry_debugger_accept_connection (void)
   struct sockaddr_in addr;
   socklen_t sin_size = sizeof (struct sockaddr_in);
 
+  uint8_t *payload_p = JERRY_CONTEXT (debugger_send_buffer) + JERRY_DEBUGGER_WEBSOCKET_HEADER_SIZE;
+  JERRY_CONTEXT (debugger_send_buffer_payload_p) = payload_p;
+
+  uint8_t max_send_size = (JERRY_DEBUGGER_MAX_BUFFER_SIZE - JERRY_DEBUGGER_WEBSOCKET_HEADER_SIZE);
+  if (max_send_size > 125)
+  {
+    max_send_size = 125;
```
martijnthe: Add `#define` for these magics + a comment why it is 125?

jerry-core/debugger/debugger-ws.c
```diff
@@ -453,6 +488,7 @@ bool
 jerry_debugger_receive (jerry_debugger_uint8_data_t **message_data_p) /**< [out] data received from client */
 {
   JERRY_ASSERT (JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONNECTED);
+  JERRY_ASSERT (JERRY_CONTEXT (debugger_max_receive_size) <= 125);
```
martijnthe: Same magic again.

jerry-core/debugger/debugger.h
```diff
@@ -341,9 +340,8 @@ typedef struct
  */
 typedef struct
 {
-  jerry_debugger_send_header_t header; /**< message header */
   uint8_t type; /**< type of the message */
-  jerry_debugger_frame_t frames[JERRY_DEBUGGER_SEND_MAX (jerry_debugger_frame_t)]; /**< frames */
+  jerry_debugger_frame_t frames[1]; /**< frames */
```
martijnthe: Why `[1]` and not `[]` (also in the other structs)?
Aren't these variable sized arrays?

jerry-core/debugger/debugger.c
```diff
@@ -796,37 +800,42 @@ jerry_debugger_send_string (uint8_t message_type, /**< message type */
...
-    string_length -= max_fragment_len;
-    string_p += max_fragment_len;
-  }
-
-  if (sub_type != JERRY_DEBUGGER_NO_SUBTYPE)
-  {
-    string_length += 1;
+    string_length -= max_byte_count;
+    string_p += max_byte_count;
   }
```
martijnthe: Ah nice catch

jerry-core/debugger/debugger.h
martijnthe: Isn't that part of C99 (the README claims that the project uses this standard)?

https://en.wikipedia.org/wiki/Flexible_array_member

> In previous standards of the C language, it was common to declare a zero-sized array member instead of a flexible array member. The GCC compiler explicitly accepts zero-sized arrays for such purposes.

Maybe use that instead?

zherczeg: Which one? Coverity has no problem with it:
https://travis-ci.org/jerryscript-project/jerryscript/jobs/361615057

zherczeg: I just realized that the definition is wrong (non-packed form), and it only works because it is never used. Perhaps we could replace the second member with a comment.

martijnthe: Yeah, that makes sense to me.

zherczeg: I finally chose `[]`. I hope it is really part of C99.

zherczeg:
Exactly my plan.

PR #2267, opened 2018-04-03
Fix bug in stringToCesu8 conversion for 0x7ff

grgustaf:


grgustaf:
Travis seemed to just fail downloading the Zephyr SDK. See #2268 for a suggestion on how to avoid that.

martijnthe:
> Could you add a test case for it?

@zherczeg this bug is in the .html debugger. Are there existing test cases for it somewhere?

zherczeg:
Hm, I haven't noticed that it is html code :)

No test case for that at the moment (and probably will be deleted in the future).

PR #2269, opened 2018-04-04
Add finalize_cb to jerry_context_data_manager_t

martijnthe:
This patch adds a new `finalize_cb` callback to `jerry_context_data_manager_t`.
The callback is run as the very last thing in `jerry_cleanup`, after the VM has been torn down entirely.
There was already the `deinit_cb`, which is run while the VM is still in the process of being torn down.
The reason the `deinit_cb` is not always sufficient is that there may still be objects alive (because they still being referenced) that have native pointers associated with the context manager that is being `deinit`'ed.
As a result, the `free_cb`s for those objects can get called *after* the associated context manager's `deinit_cb` is run. This makes cleanup of manager state that is depended on by the live objects impossible to do in the `deinit_cb`. That type of cleanup can only be done when all values have been torn down completely.

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -210,24 +210,35 @@ jerry_cleanup (void)
   }
 #endif /* JERRY_DEBUGGER */
 
-  for (jerry_context_data_header_t *this_p = JERRY_CONTEXT (context_data_p), *next_p = NULL;
+  for (jerry_context_data_header_t *this_p = JERRY_CONTEXT (context_data_p);
        this_p != NULL;
-       this_p = next_p)
+       this_p = this_p->next_p)
   {
-    next_p = this_p->next_p;
     if (this_p->manager_p->deinit_cb)
     {
       this_p->manager_p->deinit_cb (JERRY_CONTEXT_DATA_HEADER_USER_DATA (this_p));
     }
-    jmem_heap_free_block (this_p, sizeof (jerry_context_data_header_t) + this_p->manager_p->bytes_needed);
   }
 
 #ifndef CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
   ecma_free_all_enqueued_jobs ();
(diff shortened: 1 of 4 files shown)


martijnthe:
@gabrielschulhof thoughts / comments?

@zherczeg I will add tests + docs once the general idea, naming, etc. have been agreed upon.

jerry-core/api/jerry.c
```diff
@@ -210,21 +210,31 @@ jerry_cleanup (void)
...
     if (this_p->manager_p->deinit_cb)
     {
       this_p->manager_p->deinit_cb (JERRY_CONTEXT_DATA_HEADER_USER_DATA (this_p));
     }
-    jmem_heap_free_block (this_p, sizeof (jerry_context_data_header_t) + this_p->manager_p->bytes_needed);
   }
 
   ecma_finalize ();
   jmem_finalize ();
   jerry_make_api_unavailable ();
```
martijnthe: Oh shoot I need to move `jmem_finalize ()` to the end, because `jmem_heap_free_block` is used below.
Is that a problem?

zherczeg: Don't think so.

zherczeg:
The name is ok for me.

martijnthe:
> The name is ok for me.

OK, thanks.

I extended the unit tests + updated the docs.

zherczeg:
Patch is good. But I have a generic question: since the functionality is changed for `deinit_cb`, does this affect existing jerry extensions?

martijnthe:
> But I have a generic question: since the functionality is changed for deinit_cb, does this affect existing jerry extensions?

@zherczeg Which change are you referring to? That the deallocation of the buffer happens a little bit later? I don't think it affects existing jerry extensions.

martijnthe:
OK, I rebased it.

zherczeg:
Actually it happens earlier if I understand your patch. But ok, if they are not affected.

martijnthe:
Yeah, no they are not affected.

PR #2271, opened 2018-04-05
Use jerry_is_feature_enabled function instead of macros where possible

tdusnoki:


tests/unit-core/test-mem-stats.c
```diff
@@ -26,36 +25,33 @@ const char *test_source = (
...
+    jerry_release_value (parsed_code_val);
+
+    jerry_cleanup ();
+
+    return 0;
+  }
+  else
+  {
+    return 0;
+  }
```
robertsipka: The else is not necessary, call return 0 only. You can also outsource it from the if statement.

tests/unit-core/test-mem-stats.c
```diff
@@ -16,7 +16,6 @@
 #include "jerryscript.h"
 #include "test-common.h"
 
-#ifdef JMEM_STATS
 
 const char *test_source = (
```
robertsipka: I am not sure about this to keep it here, If it is only used within this file. Now it works because it is not static, but I think it should be declare as "static" here (I suppose it is not used elsewhere). Then the GCC will warn you, if it is not used (JERRY_FEATURE_MEM_STATS is not enabled). I suggest you to move this variable inside the function, right before it is used.

jerry-main/main-unix.c
```diff
@@ -237,23 +236,27 @@ wait_for_source_callback (const jerry_char_t *resource_name_p, /**< resource nam
...
 {
-  jerry_value_t ret_val = jerry_parse_named_resource (resource_name_p,
-                                                      resource_name_size,
-                                                      source_p,
-                                                      source_size,
-                                                      false);
+  jerry_value_t ret_val = jerry_create_undefined ();
 
-  if (!jerry_value_has_error_flag (ret_val))
+  if (jerry_is_feature_enabled (JERRY_FEATURE_DEBUGGER))
```
galpeter: There is no need to check this here, because this method will be only called when the debugger is enabled and it is registered via the `jerry_debugger_wait_for_client_source` call.

tests/unit-core/test-mem-stats.c
```diff
@@ -16,46 +16,39 @@
 #include "jerryscript.h"
 #include "test-common.h"
 
-#ifdef JMEM_STATS
 
-const char *test_source = (
+int main (void)
+{
+  if (jerry_is_feature_enabled (JERRY_FEATURE_MEM_STATS))
```
galpeter: IMHO: It would be better to reverse the logic: check if the feature is not enabled and if so return with 0 (early exit). That way you do not need to indent the whole code in the main.

PR #2273, opened 2018-04-09
Support abort in the debugger.

zherczeg:
Aborts are not caught by catch/finally blocks, so it is possible to stop a script using the debugger.


zherczeg:
@DanielBallaSZTE

martijnthe:
@grgustaf fyi protocol changes incoming

PR #2275, opened 2018-04-10
Reduce code duplication between Array.forEach, some, and every

mmatyas:
Their implementation only differs in the stop condition and the
final return value.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -1891,19 +1891,33 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
 } /* ecma_builtin_array_prototype_object_last_index_of */
 
 /**
- * The Array.prototype object's 'every' routine
- *
- * See also:
- *          ECMA-262 v5, 15.4.4.16
+ * Type of array routine.
+ */
+typedef enum
+{
+  ARRAY_ROUTINE_EVERY, /**< Array.every: ECMA-262 v5, 15.4.4.16 */
+  ARRAY_ROUTINE_SOME, /**< Array.some: ECMA-262 v5, 15.4.4.17 */
+  ARRAY_ROUTINE_FOREACH, /**< Array.forEach: ECMA-262 v5, 15.4.4.18 */
+  ARRAY_ROUTINE__COUNT /**< count of the modes */
+} array_routine_mode;
+
+/**
+ * Applies the provided function to each element of the array as long as
+ * the return value stays empty. The common function for 'every', 'some'
+ * and 'forEach' of the Array prototype.
  *
  * @return ecma value
  *         Returned value must be freed with ecma_free_value.
  */
 static ecma_value_t
-ecma_builtin_array_prototype_object_every (ecma_value_t this_arg, /**< this argument */
-                                           ecma_value_t arg1, /**< callbackfn */
-                                           ecma_value_t arg2) /**< thisArg */
+ecma_builtin_array_apply (ecma_value_t this_arg, /**< this argument */
+                          ecma_value_t arg1, /**< callbackfn */
+                          ecma_value_t arg2, /**< thisArg */
+                          array_routine_mode mode) /**< array routine mode */
+
 {
+  JERRY_ASSERT (mode < ARRAY_ROUTINE__COUNT);
+
   ecma_value_t ret_value = ECMA_VALUE_EMPTY;
 
   /* 1. */
@@ -1958,9 +1972,13 @@ ecma_builtin_array_prototype_object_every (ecma_value_t this_arg, /**< this argu
         ECMA_TRY_CATCH (call_value, ecma_op_function_call (func_object_p, arg2, call_args, 3), ret_value);
 
         /* 7.c.iii */
-        if (!ecma_op_to_boolean (call_value))
+        if (mode == ARRAY_ROUTINE_EVERY && !ecma_op_to_boolean (call_value))
         {
           ret_value = ECMA_VALUE_FALSE;
+        }
+        else if (mode == ARRAY_ROUTINE_SOME && ecma_op_to_boolean (call_value))
+        {
+          ret_value = ECMA_VALUE_TRUE;
         }
 
         ECMA_FINALIZE (call_value);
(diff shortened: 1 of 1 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1890,19 +1890,49 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
 /**
- * The Array.prototype object's 'every' routine
- *
- * See also:
- *          ECMA-262 v5, 15.4.4.16
- *
- * @return ecma value
- *         Returned value must be freed with ecma_free_value.
+ * If the argument is convertible to boolean true, returns ECMA_VALUE_TRUE,
+ * otherwise returns ECMA_VALUE_EMPTY.
```
galpeter: To specify the return value we usually use the `@returns` element in the docs. (please see a method above)

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1890,19 +1890,49 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
  */
 static ecma_value_t
-ecma_builtin_array_prototype_object_every (ecma_value_t this_arg, /**< this argument */
-                                           ecma_value_t arg1, /**< callbackfn */
-                                           ecma_value_t arg2) /**< thisArg */
+ecma_builtin_convert_true_maybe (ecma_value_t var)
```
galpeter: No comment for the input argument.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1890,19 +1890,49 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
+  }
+
+  return ECMA_VALUE_EMPTY;
+} /* ecma_builtin_convert_true_maybe */
+
+/**
+ * If the argument is convertible to boolean false, returns ECMA_VALUE_FALSE,
+ * otherwise returns ECMA_VALUE_EMPTY.
+ */
+static ecma_value_t
```
galpeter: The comments are the same as above.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1890,19 +1890,49 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
+  }
+
+  return ECMA_VALUE_EMPTY;
+} /* ecma_builtin_convert_false_maybe */
+
+/**
+ * Applies the provided function to each element of the array as long as
+ * the return value stays undefined. If the return value stays undefined
+ * after processing the whole array, a default value can be returned.
+ */
```
galpeter: Just like above, the `@return` should be used.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1890,19 +1890,49 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
+static ecma_value_t
+ecma_builtin_array_apply (ecma_value_t this_arg, /**< this argument */
```
galpeter: In this function the comments like: `/* 1. */` are referring to which part of the standard? We should mention that in the comments of the function (just like in the every/some/etc.) case.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1890,19 +1890,49 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
+                          ecma_value_t arg1, /**< callbackfn */
+                          ecma_value_t arg2, /**< thisArg */
+                          ecma_value_t (*check_stop_cb)(ecma_value_t), /**< retval update fn */
+                          ecma_value_t default_retval /**< default if empty */ )
```
galpeter: IMHO, I would avoid using the callback method to do the work. What we could do is to only pass an enum value (something like: ARRAY_EVERY, _SOME, _FOR_EACH) and based on this value we can calculate the stop and default return values.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1890,19 +1890,49 @@ ecma_builtin_array_prototype_object_last_index_of (ecma_value_t this_arg, /**< t
...
+  {
+    return ECMA_VALUE_FALSE;
+  }
+
+  return ECMA_VALUE_EMPTY;
+} /* ecma_builtin_convert_false_maybe */
+
+/**
+ * Applies the provided function to each element of the array as long as
+ * the return value stays undefined. If the return value stays undefined
```
galpeter: In the comment the `undefined` should be `empty`, as each callback returns `ECMA_VALUE_EMPTY`. The empty value is different from the undefined.

mmatyas:
Thanks, I've updated the patch to use a routine mode enum.

mmatyas:
Rebased to master to fix the Travis build.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1958,9 +1972,19 @@ ecma_builtin_array_prototype_object_every (ecma_value_t this_arg, /**< this argu
...
+          {
+            ret_value = ECMA_VALUE_FALSE;
+          }
+        }
+        else if (mode == ARRAY_ROUTINE_SOME)
+        {
+          if (ecma_op_to_boolean (call_value))
+          {
+            ret_value = ECMA_VALUE_TRUE;
+          }
```
rerobika: This `if` structure can be merged like this: 
```c
if (mode == ARRAY_ROUTINE_EVERY || mode == ARRAY_ROUTINE_SOME)
{
    ret_value = ecma_op_to_boolean (call_value) ? ECMA_VALUE_TRUE : ECMA_VALUE_FALSE;
}
```
Therefore `ecma_op_to_boolean (call_value)` is performed only once in both cases.

mmatyas: Ah there's a small issue actually, there are cases when the `ret_value` shouldn't change. The `if` structure can still be reduced to
```c
        if (mode == ARRAY_ROUTINE_EVERY && !ecma_op_to_boolean (call_value))
        {
            ret_value = ECMA_VALUE_FALSE;
        }
        else if (mode == ARRAY_ROUTINE_SOME && ecma_op_to_boolean (call_value))
        {
            ret_value = ECMA_VALUE_TRUE;
        }

PR #2278, opened 2018-04-12
Improve stopping at errors.

zherczeg:


LaszloLango:
Please a bit more detail. What is exactly improved? It is not clear for me.

zherczeg:
Currently if an exception is thrown, it stops with "exception throw detected" after each function return during exception handling. After this patch it will stop only if a new exception is caught.

yichoi:
LGTM

PR #2280, opened 2018-04-12
Reduce code duplication between Array.reduce and reduceRight

mmatyas:
Their code differs only in handling the array index.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -2404,171 +2404,20 @@ ecma_builtin_array_prototype_object_filter (ecma_value_t this_arg, /**< this arg
 } /* ecma_builtin_array_prototype_object_filter */
 
 /**
- * The Array.prototype object's 'reduce' routine
+ * Reduces the Array starting from left or right
  *
  * See also:
- *          ECMA-262 v5, 15.4.4.21
+ *         Array.prototype.reduce
+ *         Array.prototype.reduceRight
  *
  * @return ecma value
  *         Returned value must be freed with ecma_free_value.
  */
 static ecma_value_t
-ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this argument */
-                                            const ecma_value_t args[], /**< arguments list */
-                                            ecma_length_t args_number) /**< number of arguments */
-{
-  ecma_value_t ret_value = ECMA_VALUE_EMPTY;
-  ecma_value_t callbackfn = (args_number > 0) ? args[0] : ECMA_VALUE_UNDEFINED;
-  ecma_value_t initial_value = (args_number > 1) ? args[1] : ECMA_VALUE_UNDEFINED;
-
-  /* 1. */
-  ECMA_TRY_CATCH (obj_this,
-                  ecma_op_to_object (this_arg),
-                  ret_value);
-
-  ecma_object_t *obj_p = ecma_get_object_from_value (obj_this);
-
-  /* 2. */
-  ECMA_TRY_CATCH (len_value,
-                  ecma_op_object_get_by_magic_id (obj_p, LIT_MAGIC_STRING_LENGTH),
-                  ret_value);
-
-  ECMA_OP_TO_NUMBER_TRY_CATCH (len_number, len_value, ret_value);
-
-  /* 3. */
-  uint32_t len = ecma_number_to_uint32 (len_number);
-
-  /* 4. */
-  if (!ecma_op_is_callable (callbackfn))
-  {
-    ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("Callback function is not callable."));
-  }
-  else
-  {
-    ecma_object_t *func_object_p;
-
-    JERRY_ASSERT (ecma_is_value_object (callbackfn));
-    func_object_p = ecma_get_object_from_value (callbackfn);
-    ecma_value_t accumulator = ECMA_VALUE_UNDEFINED;
-
-    /* 5. */
-    if (len_number == ECMA_NUMBER_ZERO && ecma_is_value_undefined (initial_value))
-    {
-      ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("Initial value cannot be undefined."));
-    }
-    else
-    {
-      /* 6. */
-      uint32_t index = 0;
-
-      /* 7.a */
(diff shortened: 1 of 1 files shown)


mmatyas:
This reduces the binary size by about 2000 bytes on my machine (libjerry-core.a: 3301510 ->3299494).

mmatyas:
Rebased to master to fix the Travis build.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2472,7 +2475,8 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
         bool k_present = false;
 
         /* 8.b */
-        while (!k_present && index < len && ecma_is_value_empty (ret_value))
+        bool has_items_left = start_from_left ? (index < len) : (index != UINT32_MAX);
```
zherczeg: has_items_left ? I also don't like the `UINT32_MAX` comparison, sometimes the compilers decides that certain overflowed values cannot happen.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2493,7 +2497,8 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
           ECMA_FINALIZE (current_value);
 
           /* 8.b.iv */
-          index++;
+          index = start_from_left ? (index + 1) : (index - 1);
```
zherczeg: You could have a direction or something similar.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
zherczeg: Would it be possible to use an indexing from right? So go from 0 - len, and substract the current value from the len? I would not use int64 because it is not a native type on 32 bit systems.

mmatyas: Actually the whole `has_items_left` could be replaced with a counter -- I'll push an update.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2458,8 +2459,11 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
     }
     else
     {
+      ecma_value_t accumulator = ECMA_VALUE_UNDEFINED;
+
       /* 6. */
-      uint32_t index = 0;
+      uint32_t index = start_from_left ? 0 : (len - 1);
```
zherczeg: Ok my idea looks like the following. This variable is always initialized with 0.
`uint32_t index = 0;`

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2472,7 +2476,7 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
         bool k_present = false;
 
         /* 8.b */
-        while (!k_present && index < len && ecma_is_value_empty (ret_value))
+        while (!k_present && items_visited < len && ecma_is_value_empty (ret_value))
```
zherczeg: This is always `index < len`, just like in the original.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2493,7 +2497,8 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
           ECMA_FINALIZE (current_value);
 
           /* 8.b.iv */
-          index++;
```
zherczeg: This is `index++`.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2472,7 +2476,7 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
         bool k_present = false;
 
         /* 8.b */
-        while (!k_present && index < len && ecma_is_value_empty (ret_value))
+        while (!k_present && items_visited < len && ecma_is_value_empty (ret_value))
         {
           /* 8.b.i */
           ecma_string_t *index_str_p = ecma_new_ecma_string_from_uint32 (index);
```
zherczeg: And here comes the catch: depending on `start_from_left` we read `index` or `len - 1 - index`. We can create a local variable for len - 1.

zherczeg: This way no need int64 magic, no need to handle negative numbers.

mmatyas:
@zherczeg updated, what do you think?

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2475,7 +2479,9 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
         while (!k_present && index < len && ecma_is_value_empty (ret_value))
         {
           /* 8.b.i */
-          ecma_string_t *index_str_p = ecma_new_ecma_string_from_uint32 (index);
+          ecma_string_t *index_str_p = ecma_new_ecma_string_from_uint32 (start_from_left
```
zherczeg: This is not the correct style.
```
(a ? b
   : c)
```

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2507,18 +2513,20 @@ ecma_builtin_array_prototype_object_reduce (ecma_value_t this_arg, /**< this arg
       /* 9. */
       ecma_value_t current_index;
 
-      for (; index < len && ecma_is_value_empty (ret_value); index++)
+      while (index < len && ecma_is_value_empty (ret_value))
```
zherczeg: Is it a personal preference to remove the for or something else?

mmatyas: It was changed due to the `index` update in the previous versions. With that part reverted, the `for` might be a better choice indeed.

yichoi:
LGTM

PR #2282, opened 2018-04-16
Rework jerry_parse function.

zherczeg:
Remove jerry_parse_named_resource, merge its arguments to jerry_parse and change is_strict argument to an option list for possible future extensions.

### DIFF
--- a/jerry-core/include/jerryscript-core.h
+++ b/jerry-core/include/jerryscript-core.h
@@ -99,6 +99,15 @@ typedef enum
 } jerry_feature_t;
 
 /**
+ * Option flags for jerry_parse and jerry_parse_function functions.
+ */
+typedef enum
+{
+  JERRY_PARSE_NO_OPTS = 0, /**< no options passed */
+  JERRY_PARSE_STRICT_MODE = (1 << 0), /**< enable strict mode */
+} jerry_parse_opts_t;
+
+/**
  * Character type of JerryScript.
  */
 typedef uint8_t jerry_char_t;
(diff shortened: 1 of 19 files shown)


jerry-core/include/jerryscript-core.h
```diff
@@ -98,6 +98,14 @@ typedef enum
   JERRY_FEATURE__COUNT /**< number of features. NOTE: must be at the end of the list */
 } jerry_feature_t;
 
+/**
+ * Option flags for jerry_parse and jerry_parse_function functions.
+ */
+typedef enum
+{
+  JERRY_PARSE_STRICT_MODE = (1 << 0), /**< enable strict mode */
+} jerry_parse_opts_t;
```
LaszloLango: what about this?
```c
typedef enum
{
  JERRY_PARSE_EMPTY = 0, /**< no flags */
  JERRY_PARSE_STRICT_MODE = (1 << 0), /**< enable strict mode */
} jerry_parse_flag_t;
```
So it will be similar to `jerry_init_flag_t`.

zherczeg: I want to phase out the `flag` term from the project. The option bits are much better description.

zherczeg: (You can also see that in the snapshot gen/exec flags already landed).

LaszloLango: And what about adding JERRY_PARSE_EMPTY or JERRY_PARSE_NO_OPTIONS? We should avoid using of literals in call sites of `jerry_parse`.

zherczeg: Good idea. NO_OPTS added. I can rename it to NO_OPTIONS if you want.

PR #2292, opened 2018-04-19
Rename the jerry_value_clear_error_flag function

imiklos:
Rename the function to represent it's real functionality.


zherczeg:
I have been thinking about fully removing this function. It has always been a source of confusion.

PR #2303, opened 2018-05-02
Add checking return value

haesik:
ecma_collection_iterator_init and ecma_collection_iterator_next can return null pointer
so, return value should be checked in for loop

JerryScript-DCO-1.0-Signed-off-by: Haesik Jun haesik.jun@samsung.com


zherczeg:
Could you add a test case for it? I am curious what triggered this issue.

haesik:
@zherczeg There is no test case, It was triggered by a static analysis tool of Samsung.
Could you let me know if there is no possibility to return a NULL pointer?

haesik:
Ok, thank you for the explanation..

PR #2306, opened 2018-05-02
Add Windows support

### LINKED ISSUE #2211: Windows port
Hello JerryScript,

I have a little question. :)

I'm comparing embeddable javascript engines. I need a js engine which is available on linux, macos, android, ios and windows.

I cannot see windows support now. Is it in progress or is it planned?

lygstate:
Fixes #2211

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,10 @@ if(CMAKE_C_COMPILER_ID MATCHES "TI")
   set(USING_TI 1)
 endif()
 
+if(CMAKE_C_COMPILER_ID MATCHES "MSVC")
+  set(USING_MSVC 1)
+endif()
+
 # Determining build type
 if(NOT CMAKE_BUILD_TYPE)
   set(CMAKE_BUILD_TYPE "MinSizeRel")
@@ -94,8 +98,21 @@ if(USING_TI)
   set(ENABLE_STRIP_MESSAGE       " (FORCED BY COMPILER)")
 endif()
 
+if(USING_MSVC)
+  set(JERRY_LIBC         OFF)
+  set(JERRY_LIBM         OFF)
+  set(ENABLE_ALL_IN_ONE  ON)
+  set(ENABLE_STRIP       OFF)
+
+  set(JERRY_LIBC_MESSAGE         " (FORCED BY COMPILER)")
+  set(JERRY_LIBM_MESSAGE         " (FORCED BY COMPILER)")
+  set(ENABLE_ALL_IN_ONE_MESSAGE  " (FORCED BY COMPILER)")
+  set(ENABLE_STRIP_MESSAGE       " (FORCED BY COMPILER)")
+endif()
+
 # Status messages
 message(STATUS "CMAKE_BUILD_TYPE            " ${CMAKE_BUILD_TYPE})
+message(STATUS "CMAKE_C_COMPILER_ID         " ${CMAKE_C_COMPILER_ID})
 message(STATUS "CMAKE_SYSTEM_NAME           " ${CMAKE_SYSTEM_NAME})
 message(STATUS "CMAKE_SYSTEM_PROCESSOR      " ${CMAKE_SYSTEM_PROCESSOR})
 message(STATUS "ENABLE_ALL_IN_ONE           " ${ENABLE_ALL_IN_ONE} ${ENABLE_ALL_IN_ONE_MESSAGE})
@@ -190,7 +207,7 @@ if(("${PLATFORM}" STREQUAL "DARWIN"))
   jerry_add_link_flags(-lSystem)
   set(CMAKE_C_ARCHIVE_CREATE "<CMAKE_AR> Sqc <TARGET> <LINK_FLAGS> <OBJECTS>")
   set(CMAKE_C_ARCHIVE_FINISH "<CMAKE_RANLIB> -no_warning_for_no_symbols -c <TARGET>")
-else()
+elseif(NOT USING_MSVC)
   jerry_add_link_flags(-Wl,-z,noexecstack)
 endif()
 
@@ -199,9 +216,16 @@ if(JERRY_LIBC)
   jerry_add_link_flags(-nostdlib)
 endif()
 
+if(USING_MSVC)
+  jerry_add_link_flags(/OPT:NOREF)
+  # Disable MSVC warning 4996 globally because it stops us from using
+  # standard C functions.
+  jerry_add_compile_flags(/wd4996)
+endif()
+
 # Turn off stack protector
 if (USING_GCC OR USING_CLANG)
-jerry_add_compile_flags(-fno-stack-protector)
+  jerry_add_compile_flags(-fno-stack-protector)
 endif()
 
 if (USING_GCC OR USING_CLANG)
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,30 @@
+version: 1.0.{build}
+pull_requests:
+  do_not_increment_build_number: true
+branches:
+  except:
+  - coverity_scan
+  - gh_pages
+skip_tags: true
+image:
+  - Visual Studio 2017
+configuration:
+  - Debug
+  - Release
+platform:
+  - x64
+  - Win32
+init:
+  - cmd: |
+      cmake -version
+before_build:
+  - cmd: |
+      mkdir build
+      cd build
+      if "%PLATFORM%"=="Win32" cmake -G"Visual Studio 15 2017" ..
+      if "%PLATFORM%"=="x64" cmake -G"Visual Studio 15 2017 Win64" ..
+
+build:
+  project: build\Jerry.sln
+  parallel: true
+  verbosity: minimal
--- a/jerry-core/CMakeLists.txt
+++ b/jerry-core/CMakeLists.txt
@@ -146,9 +146,7 @@ endif()
 set(INCLUDE_THIRD_PARTY_VALGRIND "${CMAKE_SOURCE_DIR}/third-party/valgrind")
 
 # build mode specific compile/link flags
-if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
-  set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_NDEBUG)
-endif()
+set(DEFINES_JERRY ${DEFINES_JERRY} $<$<NOT:$<CONFIG:Debug>>:JERRY_NDEBUG>)
 
 # Jerry heap-section
 if(DEFINED JERRY_HEAP_SECTION_ATTR)
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
@@ -84,7 +84,7 @@ ecma_builtin_helper_object_to_string (const ecma_value_t this_arg) /**< this arg
      'Null' or one of possible object's classes.
      The string with null character is maximum 27 characters long. */
   const lit_utf8_size_t buffer_size = 27;
-  lit_utf8_byte_t str_buffer[buffer_size];
+  JERRY_VLA (lit_utf8_byte_t, str_buffer, buffer_size);
 
   lit_utf8_byte_t *buffer_ptr = str_buffer;
 
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -1362,9 +1362,10 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
   const bool obj_is_builtin = ecma_get_object_is_builtin (obj_p);
 
   const size_t bitmap_row_size = sizeof (uint32_t) * JERRY_BITSINBYTE;
-  uint32_t names_hashes_bitmap[ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size];
+  const size_t names_hashes_bitmap_size = ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size;
+  JERRY_VLA (uint32_t, names_hashes_bitmap, names_hashes_bitmap_size);
 
-  memset (names_hashes_bitmap, 0, sizeof (names_hashes_bitmap));
+  memset (names_hashes_bitmap, 0, names_hashes_bitmap_size * sizeof (names_hashes_bitmap[0]));
 
   for (ecma_object_t *prototype_chain_iter_p = obj_p;
        prototype_chain_iter_p != NULL;
@@ -1458,8 +1459,9 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
 
     ecma_value_t *ecma_value_p = ecma_collection_iterator_init (prop_names_p);
 
-    uint32_t own_names_hashes_bitmap[ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size];
-    memset (own_names_hashes_bitmap, 0, sizeof (own_names_hashes_bitmap));
+    const size_t own_names_hashes_bitmap_size = ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size;
+    JERRY_VLA (uint32_t, own_names_hashes_bitmap, own_names_hashes_bitmap_size);
+    memset (own_names_hashes_bitmap, 0, own_names_hashes_bitmap_size * sizeof (own_names_hashes_bitmap[0]));
 
     while (ecma_value_p != NULL)
     {
--- a/jerry-core/include/jerryscript-compiler.h
+++ b/jerry-core/include/jerryscript-compiler.h
@@ -47,6 +47,24 @@ extern "C"
 
 #endif /* __GNUC__ */
 
+#ifdef _MSC_VER
+
+/*
+ * Compiler-specific macros relevant for Microsoft Visual C/C++ Compiler.
+ */
+#define JERRY_ATTR_DEPRECATED __declspec(deprecated)
+#define JERRY_ATTR_NOINLINE __declspec(noinline)
+#define JERRY_ATTR_NORETURN __declspec(noreturn)
+
+/*
+ * Microsoft Visual C/C++ Compiler doesn't support for VLA, using _alloca
+ * instead.
+ */
+void * __cdecl _alloca (size_t _Size);
+#define JERRY_VLA(type, name, size) type *name = (type *) (_alloca (sizeof (type) * size))
+
+#endif /* _MSC_VER */
+
 /*
  * Default empty definitions for all compiler-specific macros. Define any of
  * these in a guarded block above (e.g., as for GCC) to fine tune compilation
@@ -109,6 +127,10 @@ extern "C"
 #ifndef JERRY_UNLIKELY
 #define JERRY_UNLIKELY(x) (x)
 #endif /* !JERRY_UNLIKELY */
+
+#ifndef JERRY_VLA
+#define JERRY_VLA(type, name, size) type name[size]
+#endif /* !JERRY_VLA */
 
 /**
  * @}
--- a/jerry-ext/arg/arg.c
+++ b/jerry-ext/arg/arg.c
@@ -115,7 +115,7 @@ jerryx_arg_transform_object_properties (const jerry_value_t obj_val,/**< the JS 
     return jerry_create_error (JERRY_ERROR_TYPE, (jerry_char_t *) "Not an object.");
   }
 
-  jerry_value_t prop[name_cnt];
+  JERRY_VLA (jerry_value_t, prop, name_cnt);
 
   for (jerry_length_t i = 0; i < name_cnt; i++, name_p++)
   {
@@ -160,7 +160,7 @@ jerryx_arg_transform_array (const jerry_value_t array_val, /**< points to the JS
     return jerry_create_error (JERRY_ERROR_TYPE, (jerry_char_t *) "Not an array.");
   }
 
-  jerry_value_t arr[c_arg_cnt];
+  JERRY_VLA (jerry_value_t, arr, c_arg_cnt);
 
   for (jerry_length_t i = 0; i < c_arg_cnt; i++)
   {
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -417,13 +417,13 @@ main (int argc,
       char **argv)
 {
   srand ((unsigned) jerry_port_get_current_time ());
-  const char *file_names[argc];
+  JERRY_VLA (const char *, file_names, argc);
   int files_counter = 0;
 
   jerry_init_flag_t flags = JERRY_INIT_EMPTY;
 
-  const char *exec_snapshot_file_names[argc];
-  uint32_t exec_snapshot_file_indices[argc];
+  JERRY_VLA (const char *, exec_snapshot_file_names, argc);
+  JERRY_VLA (uint32_t, exec_snapshot_file_indices, argc);
   int exec_snapshots_count = 0;
 
   bool is_parse_only = false;
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,4 @@ docs/doxygen
 # Tests
 tests/test262/
 tests/unit-doc/*.c
+.vs
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
@@ -639,7 +639,7 @@ ecma_date_to_string_format (ecma_number_t datetime_number, /**< datetime */
   };
 
   const uint32_t date_buffer_length = 34;
-  lit_utf8_byte_t date_buffer[date_buffer_length];
+  JERRY_VLA (lit_utf8_byte_t, date_buffer, date_buffer_length);
 
   lit_utf8_byte_t *dest_p = date_buffer;
 
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -3045,7 +3045,7 @@ vm_run (const ecma_compiled_code_t *bytecode_header_p, /**< byte-code data heade
   frame_ctx.call_operation = VM_NO_EXEC_OP;
 
   /* Use JERRY_MAX() to avoid array declaration with size 0. */
-  ecma_value_t stack[JERRY_MAX (call_stack_size, 1)];
+  JERRY_VLA (ecma_value_t, stack, JERRY_MAX (call_stack_size, 1));
   frame_ctx.registers_p = stack;
 
   return vm_execute (&frame_ctx, arg_list_p, arg_list_len);
--- a/jerry-ext/module/module.c
+++ b/jerry-ext/module/module.c
@@ -176,7 +176,7 @@ jerryx_resolve_native_module (const jerry_value_t canonical_name, /**< canonical
   const jerryx_native_module_t *module_p = NULL;
 
   jerry_size_t name_size = jerry_get_utf8_string_size (canonical_name);
-  jerry_char_t name_string[name_size];
+  JERRY_VLA (jerry_char_t, name_string, name_size);
   jerry_string_to_utf8_char_buffer (canonical_name, name_string, name_size);
 
   /* Look for the module by its name in the list of module definitions. */
@@ -214,7 +214,7 @@ jerryx_module_resolve_local (const jerry_value_t name, /**< name of the module t
   size_t index;
   size_t canonical_names_used = 0;
   jerry_value_t instances;
-  jerry_value_t canonical_names[resolver_count];
+  JERRY_VLA (jerry_value_t, canonical_names, resolver_count);
   jerry_value_t (*get_canonical_name_p) (const jerry_value_t name);
   bool (*resolve_p) (const jerry_value_t canonical_name,
                      jerry_value_t *result);


jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -84,7 +84,7 @@ ecma_builtin_helper_object_to_string (const ecma_value_t this_arg) /**< this arg
      'Null' or one of possible object's classes.
      The string with null character is maximum 27 characters long. */
   const lit_utf8_size_t buffer_size = 27;
-  lit_utf8_byte_t str_buffer[buffer_size];
+  jerry_alloca(str_buffer, lit_utf8_size_t, buffer_size);
```
zherczeg: These (this and below) are fixed size arrays, they should be supported by the compiler.

jerry-main/main-unix.c
```diff
@@ -283,7 +283,7 @@ wait_for_source_callback (const jerry_char_t *resource_name_p, /**< resource nam
                           size_t resource_name_size, /**< size of resource name */
                           const jerry_char_t *source_p, /**< source code */
                           size_t source_size, /**< source code size */
-                          void *user_p __attribute__((unused))) /**< user pointer */
+                          void *user_p __attr_unused___) /**< user pointer */
```
zherczeg: this file should not rely anything internal

zherczeg: Sounds like a bad design.

akosthekiss: I'm not sure about any of the comments here. How about writing full sentences, ideas, justifications, explanations?

@lygstate: `__attr_unused___` does _not_ come from jerryscript-port.h. That public header writes out all attributes in full, e.g., `__attribute__((noreturn))`. Moreover, it does not make use of the `unused` attribute, neither in its full form nor in the shorthand notation. What's even more strange is that no core part of JerryScript uses the `__attr_unused___` macro, only an internal header of jerry-libc does, which can not possibly leak out to main-unix.c.

@zherczeg: There is no such bad design in jerryscript-port.h. Which can be verified by taking just a bit closer look at the header.

lygstate: I added it into jerryscript-port.h, cause Visual Studio doesnn't suport for __attribute__((noreturn)) or something alike, cause they are porting code, so I moved into jerryscript-port.h @akosthekiss .
Any better idea I appreciate.

lygstate:
I am still confused why https://travis-ci.org/jerryscript-project/jerryscript/jobs/374265701 still failed?
Did I miss something?

CMakeSettings.json
```diff
@@ -0,0 +1,4 @@
+{
+  "configurations": [
+  ]
+}
```
akosthekiss: Is this file mandatory? I see that this is related to MSVC's cmake support, but some blog about that topic mentions "If your CMake project requires additional settings...". And right now, this file seems quite empty.

BTW, a newline is missing from the end of the last line.

jerry-core/CMakeLists.txt
```diff
@@ -145,10 +145,7 @@ endif()
 # Valgrind
 set(INCLUDE_THIRD_PARTY_VALGRIND "${CMAKE_SOURCE_DIR}/third-party/valgrind")
 
-# build mode specific compile/link flags
-if(NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
-  set(DEFINES_JERRY ${DEFINES_JERRY} JERRY_NDEBUG)
-endif()
+set (DEFINES_JERRY "$<$<NOT:$<CONFIG:Debug>>:JERRY_NDEBUG>")
```
akosthekiss: Why is this change needed? The two expressions seem to be (roughly) equivalent and the change seems to be unrelated to Windows support. I'd not support the rewrite if not justified. (Moreover it's actually not equivalent, `set(DEFINES_JERRY ${DEFINES_JERRY} "$<$<NOT:$<CONFIG:Debug>>:JERRY_NDEBUG>")` would be.)

jerry-libc/CMakeLists.txt
```diff
@@ -39,7 +39,10 @@ add_library(${JERRY_LIBC_NAME} STATIC ${SOURCE_LIBC} ${TARGET_SPECIFIC_LIBC_SOUR
 target_compile_definitions(${JERRY_LIBC_NAME} PRIVATE ${DEFINES_LIBC})
 target_include_directories(${JERRY_LIBC_NAME} PRIVATE ${INCLUDE_LIBC})
 target_include_directories(${JERRY_LIBC_NAME} SYSTEM PUBLIC "${CMAKE_SOURCE_DIR}/jerry-libc/include")
-target_link_libraries(${JERRY_LIBC_NAME} -lgcc)
+
+if(NOT USING_MSVC)
+  target_link_libraries(${JERRY_LIBC_NAME} -lgcc)
+endif()
```
akosthekiss: If jerry-libc gets disabled for MSVC by default, this change will become unnecessary.

jerry-libc/include/stdlib.h
```diff
@@ -28,9 +28,15 @@ extern "C"
...
+#define NO_RETURN_ATTRIBUTE __declspec(noreturn)
+#else
+#define NO_RETURN_ATTRIBUTE __attribute__ ((noreturn))
+#endif
 
-void __attribute__ ((noreturn)) exit (int);
-void __attribute__ ((noreturn)) abort (void);
+
+void NO_RETURN_ATTRIBUTE exit (int);
+void NO_RETURN_ATTRIBUTE abort (void);
```
akosthekiss: If jerry-libc gets disabled for MSVC by default, this change will become unnecessary.

jerry-libc/jerry-libc-defs.h
```diff
@@ -23,11 +23,19 @@
...
+#define __attr_noreturn___ __declspec(noreturn)
+#define __attr_noinline___ __declspec(noinline)
+#define __attr_weak___
+#else
 #define __attr_unused___   __attribute__((unused))
 #define __attr_used___     __attribute__((used))
 #define __attr_noreturn___ __attribute__((noreturn))
 #define __attr_noinline___ __attribute__((noinline))
 #define __attr_weak___     __attribute__((weak))
+#endif
```
akosthekiss: If jerry-libc gets disabled for MSVC by default, this change will become unnecessary.

jerry-libm/CMakeLists.txt
```diff
@@ -18,11 +18,17 @@ project (${JERRY_LIBM_NAME} C)
...
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} /wd4146")
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} /wd4018")
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} /wd4293")
+else()
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-error=sign-compare")
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-error=sign-conversion")
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-sign-conversion")
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-sign-compare")
+  set(COMPILE_FLAGS_LIBM "${COMPILE_FLAGS_LIBM} -Wno-strict-aliasing")
+endif()
```
akosthekiss: If jerry-libm gets disabled for MSVC by default, this change will become unnecessary.

jerry-libm/include/math.h
```diff
@@ -22,9 +22,12 @@ extern "C"
...
 /* General Constants. */
-#define INFINITY    (1.0/0.0)
-#define NAN         (0.0/0.0)
-#define HUGE_VAL    INFINITY
+#define _HUGE_ENUF  1e+300
+#define INFINITY   ((float)(_HUGE_ENUF * _HUGE_ENUF))
+#define HUGE_VAL   ((double)INFINITY)
+#define HUGE_VALF  ((float)INFINITY)
+#define HUGE_VALL  ((long double)INFINITY)
+#define NAN        ((float)(INFINITY * 0.0F))
```
akosthekiss: If jerry-libm gets disabled for MSVC by default, this change will become unnecessary.

jerry-libm/log.c
```diff
@@ -105,11 +107,11 @@ log (double x)
   {
     if (((hx & 0x7fffffff) | lx) == 0) /* log(+-0) = -inf */
     {
-      return -two54 / zero;
+      return -INFINITY;
     }
     if (hx < 0) /* log(-#) = NaN */
     {
-      return (x - x) / zero;
+      return NAN;
```
akosthekiss: If jerry-libm gets disabled for MSVC by default, this change will become unnecessary.

jerry-core/jrt/jrt.h
```diff
@@ -109,6 +125,9 @@ void __noreturn jerry_unreachable (const char *file, const char *function, const
     jerry_unreachable (__FILE__, __func__, __LINE__); \
   } while (0)
 #else /* JERRY_NDEBUG */
+#if defined(_MSC_VER)
+#define JERRY_ASSERT(x) do {} while(0)
```
akosthekiss: Why is this variant necessary? The non-MSVC variant seems to be identical to this.

jerry-core/jrt/jrt.h
```diff
@@ -31,6 +31,21 @@
 /*
  * Attributes
  */
+#ifdef _MSC_VER
+#define __noreturn __declspec(noreturn)
+#define __attr_noinline___ __declspec(noinline)
+#define __attr_return_value_should_be_checked___
+#define __attr_hot___
+#ifndef __attr_always_inline___
```
akosthekiss: Is it possible that some variant of MSVC (pre)defines `__attr_always_inline___`, `__attr_const___`, or `__attr_pure___ `? If not, the `#ifndef`s are unnecessary.

jerry-core/jcontext/jcontext.c
```diff
@@ -38,7 +38,13 @@ jerry_context_t jerry_global_context;
 /**
  * Global heap.
  */
-jmem_heap_t jerry_global_heap __attribute__ ((aligned (JMEM_ALIGNMENT))) JERRY_GLOBAL_HEAP_SECTION;
+#ifdef _MSC_VER
+#define _memory_alignment
+#else
+#define _memory_alignment __attribute__((aligned(JMEM_ALIGNMENT)))
+#endif
```
akosthekiss: This should definitely be moved to and handled by jrt.h. And renamed to match existing naming scheme. Something like:

```c
#define __attr_aligned___(ALIGNMENT) __attribute__((aligned(ALIGNMENT)))
```

jerry-main/main-unix.c
zherczeg: I worried that the macro was exposed, but it is added in this patch, which should not. I am happy!

lygstate: @zherczeg Help me have a look at the CI, why it's still failed.

jerry-core/include/jerryscript-port.h
```diff
@@ -20,6 +20,32 @@
...
+#endif
+#define __format_type
+#else
+#define __attr_unused___ __attribute__((unused))
+#define __attr_used___ __attribute__((used))
+#ifndef __noreturn
+#define __noreturn __attribute__((noreturn))
+#endif
+#define __format_type __attribute__((format(printf, 2, 3)))
+#endif
```
akosthekiss: I see what's happening here and I also see why. I think it's not right this way, but my problem is that I cannot come up with a good alternative suggestion right now. The issue with this change is that jerryscript-port.h is there to define an API for port implementors to implement and for jerry-core to rely on. It's not to define an API that extensions, jerry-core users, etc. can use. At least that's how it used to be, and it was cleaner that way IMO.

But first things first: some definitions above can certainly be removed if jerry-libc and -libm are disabled for MSVC, let's see what remains.

PS: `jerry_alloca` should be all caps, provided that it remains.

PPS: `__format_type` should be something like `__attr_format___(...) __attribute__((format(__VA_ARGS__)))`. Only `__noreturn` deviates from the `__attr_xxx___` pattern and we should not introduce mor deviations IMO.

PPPS: Now I see why you wrote before that `__attr_unused___` comes from jerryscript-port.h. But I still think it shouldn't and that there should be a better/nicer way.

zherczeg:
It says:
```
jerry-core/vm/vm.c:3048: style(obsoleteFunctionsalloca): Obsolete function 'alloca' called. In C99 and later it is recommended to use a variable length array instead.
```
Please do as the message says.

jerry-core/include/jerryscript-port.h
```diff
@@ -20,6 +20,32 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#if defined(_MSC_VER)
```
akosthekiss: Somewhere it's `#if defined(_MSC_VER)`, somewhere else it's `#ifdef _MSC_VER`. The two are equivalent but it's best to keep a consistent style.

jerry-core/include/jerryscript-port.h
lygstate: 1. jerry_alloca -> JERRY_ALLOCA ?
2. Define __attribute__ to empty macro? is this an acceptable idea?
PPPS: Looking for better suggestion, I am still not familier with jerry coding stardand.

jerry-core/include/jerryscript-port.h
```diff
@@ -20,6 +20,32 @@
...
+#include <malloc.h>
+#endif
+
+#if defined(_MSC_VER)
+#define jerry_alloca(name, type, size) type *name = (type*)(alloca(sizeof(type)*size))
```
akosthekiss: As noticed already, cppcheck is picky about `alloca`. And by default cppcheck checks all preprocessor configurations, so even the MSVC-specific ones. It can either be tweaked not to check those configurations (perhaps by passing `-U_MSC_VER` to cppcheck, see tools/check-cppcheck.sh) or by adding `obsoleteFunctionsalloca` to the list of suppressed errors (see tools/cppcheck/suppressions-list).

CMakeSettings.json
lygstate: OK, Indeed, this file comes from other people, I may remove it:)

jerry-core/include/jerryscript-port.h
akosthekiss: 1. yes
2. That might work. It's quite a trick. It could even be done from cmake files, adding a `__attribute__(x)=` to the compile definitions (or to `DEFINES_JERRY` perhaps).

jerry-core/jrt/jrt.h
akosthekiss: I'm not sure I understand. There is no function in use here. `JERRY_UNUSED (x)` is just a macro defined some lines above and gets expanded by the preprocessor to `((void) (x))`. What does your linker report?

jerry-core/include/jerryscript-port.h
lygstate: @akosthekiss if the jerry api does not use __attribute__ then, in CMake should be fine.

jerry-core/jrt/jrt.h
lygstate: do you see, ((void) (x));
If the x is a function call, the x got to be expanded and this function would be called.

akosthekiss: Actually, it would never be called because it is in an `if (false)` block. Until now, all compilers were able to optimize that away before it reached the linker. Do you have optimizations enabled?

If that does not help, It might still be easier/simpler to define `JERRY_UNUSED` to nothing in the case of MSVC.

akosthekiss: The more I think about it the less correct the MSVC compiler/linker behaviour seems. Even if the compiler is not optimizing the `if (false)` away, what can be there in `x` that's not linkable? We do test the debug configuration with assertions enabled and everything is fine both on Linux and on macOS. Which means that `x` will be evaluated, and if it contains function calls, they will be linked and called. This suggests me that we are not using anithing that's not there. So, the question still stands: what does your linker report / complain about?

jerry-core/include/jerryscript-port.h
akosthekiss: BTW, I'd vote for the simpler form (`#ifdef _MSC_VER`), which is much more widely used in the existing code base than the longer form, which is mostly used when or-ing/and-ing of preprocessor predicates is necessary.

lygstate: I prefer defined because if we want to add expression then we modify less code

jerry-libm/log.c
lygstate: Yeap , I wanna libm could be enable in msvc, then we can create very tiny jerryscript engine (without msvc runtime depends, that's a dependence hell, if you have experience with it)

jerry-core/include/jerryscript-port.h
akosthekiss: But a consistent style is easier for the maintainers to maintain. (And, well, if we modify the predicate, we modify the same line(s), no matter which style we use, don't we?)

jerry-libc/CMakeLists.txt
lygstate: this is necessary， msvc have no gcc

jerry-core/include/jerryscript-port.h
lygstate: OK， I hope reviews be more specific next time

jerry-core/jrt/jrt.h
lygstate: I am consider that if defines to empty， then its not a concrete statement along with a comma； anyway， empty is also acceptable

jerry-libc/CMakeLists.txt
akosthekiss: Trust me, it isn't. Jerry-libc won't work for Windows. So better disable it in the root cmakelists.txt completely for MSVC builds, and then this cmakelists.txt will not even be considered during the build.

jerry-core/jrt/jrt.h
akosthekiss: Please, _do_ copy here what the linker reports as error.

CMakeSettings.json
lygstate: Removed

jerry-main/main-unix.c
lygstate: Fixed.

CMakeLists.txt
```diff
@@ -191,12 +205,22 @@ if(("${PLATFORM}" STREQUAL "DARWIN"))
...
 endif()
 
 # Turn off linking to compiler's default libc, in case jerry-libc is used
-if(JERRY_LIBC)
-  jerry_add_link_flags(-nostdlib)
+if(NOT USING_MSVC)
+  if(JERRY_LIBC)
+    jerry_add_link_flags(-nostdlib)
+  endif()
+endif()
```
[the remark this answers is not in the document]
lygstate: Fixed. LIBC & LIBM used under Win32.

jerry-libm/log.c
lygstate: Fixed. LIBC & LIBM used under Win32.

lygstate:
About the JERRY_ASSERT, The orignal patch from PKarics karicska@gmail.com have linkage error.
And after I revised the CMake script, it doesn't appear anymore, sorry for the distrub @akosthekiss

jerry-core/jrt/jrt.h
lygstate: Question, why the orignal version have #ifdef ?
That's just a copy of the orignal version.
```
#define __noreturn __attribute__((noreturn))
#define __attr_noinline___ __attribute__((noinline))
#define __attr_return_value_should_be_checked___ __attribute__((warn_unused_result))
#define __attr_hot___ __attribute__((hot))
#ifndef __attr_always_inline___
#define __attr_always_inline___ __attribute__((always_inline))
#endif /* !__attr_always_inline___ */
#ifndef __attr_const___
#define __attr_const___ __attribute__((const))
#endif /* !__attr_const___ */
#ifndef __attr_pure___
#define __attr_pure___ __attribute__((pure))
#endif /* !__attr_pure___ */
```

jerry-core/jcontext/jcontext.c
lygstate: moved

jerry-core/include/jerryscript-port.h
lygstate: Fixesd

jerry-core/jrt/jrt.h
akosthekiss: Good question. This code dates back to the very beginning of the project (https://github.com/jerryscript-project/jerryscript/commit/afe242e7a3d96ac1353fecc1bc43eea5649bb5c3). For some 3.5 years now, nobody has dared to touch this. Most probably, the original `#ifndefs` can be removed as well. May be a candidate for a follow-up refactoring patch.

jerry-core/CMakeLists.txt
akosthekiss: I've found the same SO post last night. CMake's multi-configuration generators are tricky to set config at build time, it's quite unexpected. I'm still not completely getting it how the evaluation of generator expressions can be postponed until build time, but let's accept that this is the suggested approach.

This change still deserves some comments:
- style: no space between set and (
- style: I'm not sure the quotes are necessary around the generator expression
- logic: as mentioned above, the rewrite is not functionally equivalent; an equivalent rewrite is `set(DEFINES_JERRY ${DEFINES_JERRY} $<$<NOT:$<CONFIG:Debug>>:JERRY_NDEBUG>)`

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1362,7 +1362,7 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
   const bool obj_is_builtin = ecma_get_object_is_builtin (obj_p);
 
   const size_t bitmap_row_size = sizeof (uint32_t) * JERRY_BITSINBYTE;
-  uint32_t names_hashes_bitmap[ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size];
+  JERRY_ALLOCA(names_hashes_bitmap, uint32_t, ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size);
 
   memset (names_hashes_bitmap, 0, sizeof (names_hashes_bitmap));
```
akosthekiss: Cppcheck complains about this memset and rightfully so. Previously `names_hashes_bitmap` was guaranteed to be a VLA, for which sizeof gave meaningful result. But for MSVC, it becomes a pointer, so this memset will only clear the first some bytes of the alloca'd array, as sizeof will give the size of the pointer, not the size of the allocated array. `ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size * sizeof (uint32_t)` will always work, even if its much uglier than before.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1458,7 +1458,7 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
 
     ecma_value_t *ecma_value_p = ecma_collection_iterator_init (prop_names_p);
 
-    uint32_t own_names_hashes_bitmap[ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size];
+    JERRY_ALLOCA(own_names_hashes_bitmap, uint32_t, ECMA_OBJECT_HASH_BITMAP_SIZE / bitmap_row_size);
     memset (own_names_hashes_bitmap, 0, sizeof (own_names_hashes_bitmap));
```
akosthekiss: Same issue as above. To be solved similarly.

lygstate:
OK at least it pasaed the CI, I would opt out libc libm in this pull request

CMakeLists.txt
```diff
@@ -216,7 +236,11 @@ elseif(USING_CLANG)
 endif()
 
 if(JERRY_LIBC)
-  jerry_add_compile_flags(-Werror)
+  if(USING_MSVC)
+    jerry_add_compile_flags(/Wall) # TODO: We may turn warning to error by /WX
+  else()
+    jerry_add_compile_flags(-Werror)
+  endif()
```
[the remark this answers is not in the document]
lygstate: they are inside libc libm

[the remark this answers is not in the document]
lygstate: they are inside libc libm

CMakeLists.txt
```diff
@@ -94,10 +98,22 @@ if(USING_TI)
...
 
+if(USING_MSVC)
+  set(JERRY_LIBC         OFF)
+  set(JERRY_LIBM         OFF)
+  set(ENABLE_STRIP       OFF)
+  set(ENABLE_ALL_IN_ONE  ON)
+
+  set(JERRY_LIBC_MESSAGE         " (FORCED BY COMPILER)")
+  set(JERRY_LIBM_MESSAGE         " (FORCED BY COMPILER)")
+  set(ENABLE_ALL_IN_ONE_MESSAGE  " (FORCED BY COMPILER)")
```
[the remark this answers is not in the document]
akosthekiss: I don't understand what you mean by that. You set four configuration variables (above the empty line; those that control libc, libc, final binary stripping, and all-in-one compilation) but you only "document" three of them by setting the corresponding `_MESSAGE` variables (below the empty line). All four should have a `_MESSAGE`, i.e., you should set `ENABLE_STRIP_MESSAGE`, too.

lygstate:
@akosthekiss At least getting it to be compiled are necessary, but jerryscript need open appveyor first.
I have experience with appveyor.

lygstate:
@akosthekiss Merge this first then add functional of appveyor.

lygstate:
@akosthekiss Appveyor also added, I am not sure it's valid.

appveyor.yml
```diff
@@ -0,0 +1,29 @@
...
+  - x64
+  - Win32
+init:
+  - cmd: |
+      cmake -version
+before_build:
+  - cmd: |
+      cd /d %APPVEYOR_BUILD_FOLDER%
+      if "%PLATFORM%"=="Win32" mkdir build & cd build & cmake ..
+      if "%PLATFORM%"=="x64" mkdir build & cd build & cmake ..
```
akosthekiss: There is no difference between the handling of the two platforms. And I also think that we are in the right directory when the command is executed. So I think that
```yaml
before_build:
  - cmd: |
      mkdir build
      cd build
      cmake ..
```
would be enough (and even easier to read perhaps).

appveyor.yml
```diff
@@ -0,0 +1,29 @@
...
+
+build_script:
+  - cmd: |
+      cd /d %APPVEYOR_BUILD_FOLDER%
+      "%ProgramFiles(x86)%\MSBuild\14.0\Bin\amd64\MSBuild.exe" /verbosity:minimal build/Jerry.sln /p:Configuration="%CONFIGURATION%" /p:Platform="%PLATFORM%"
```
akosthekiss: The path to msbuild seems to be too hardcoded. AFAIK and as far as I can see from AppVeyor examples, `msbuild ...` should simply work.

But perhaps an even simple solution exists?
```yaml
build:
  project: build\Jerry.sln
  parallel: true
  verbosity: minimal
```

appveyor.yml
```diff
@@ -0,0 +1,29 @@
...
+  - x64
+  - Win32
+init:
+  - cmd: |
+      cmake -version
+before_build:
+  - cmd: |
+      cd /d %APPVEYOR_BUILD_FOLDER%
+      if "%PLATFORM%"=="Win32" mkdir build & cd build & cmake -G"Visual Studio 15 2017" ..
+      if "%PLATFORM%"=="x64" mkdir build & cd build & cmake -G"Visual Studio 15 2017 Win64" ..
```
akosthekiss: Even if there is some conditional part on the platform part, there is a bit too much repetition here. As pointed out earlier.
```yaml
before_build:
  - cmd: |
      cd /d %APPVEYOR_BUILD_FOLDER%
      mkdir build
      cd build
      if "%PLATFORM%"=="Win32" cmake -G"Visual Studio 15 2017" ..
      if "%PLATFORM%"=="x64" cmake -G"Visual Studio 15 2017 Win64" ..
```
This does not repeat what does not need repeating.

BTW, why is the `cd` command necessary? Have you tried without it? Other cmake-based projects I've seen don't use this construct and work just fine.

appveyor.yml
```diff
@@ -0,0 +1,29 @@
...
+
+build_script:
+  - cmd: |
+      cd /d %APPVEYOR_BUILD_FOLDER%
+      MSBuild /verbosity:minimal build/Jerry.sln /p:Configuration="%CONFIGURATION%" /p:Platform="%PLATFORM%"
```
akosthekiss: Two questions here:
- Is the `cd` necessary? See above comment.
- Is `cmd` necessary at all or is just `project`, `verbosity`, and `parallel` enough? See previous comment with example.

lygstate: @akosthekiss no enough, I need to speicify the Platform.

appveyor.yml
lygstate: @akosthekiss  I am not sure you way it correct,  it's better to have appeyvor ci to verify it.

lygstate: using cmd is better if we want to add more functional in the future.

appveyor.yml
akosthekiss: I don't think that's correct. Just tried to simplify the configuration: https://github.com/akosthekiss/jerryscript/commit/9b4cbc9ba04a3b26ad57a1d80e959ab7a2f257d0 . It works well, as expected. The AppVeyor CI passes both configuration and platform to msbuild when using the `build` approach: https://ci.appveyor.com/project/akosthekiss/jerryscript/build/1.0.7 . You are welcome to use the patch.

appveyor.yml
```diff
@@ -0,0 +1,30 @@
...
+  - cmd: |
+      mkdir build
+      cd build
+      if "%PLATFORM%"=="Win32" cmake -G"Visual Studio 15 2017" ..
+      if "%PLATFORM%"=="x64" cmake -G"Visual Studio 15 2017 Win64" ..
+
+build:
+  project: build\Jerry.sln
+  parallel: true
+  verbosity: minimal
```
akosthekiss: Missing new line at end of file

jerry-core/include/jerryscript-port.h
```diff
@@ -62,7 +77,7 @@ typedef enum
  *
  * Example: a libc-based port may implement this with exit() or abort(), or both.
  */
-void jerry_port_fatal (jerry_fatal_code_t code) __attribute__((noreturn));
+__noreturn void jerry_port_fatal (jerry_fatal_code_t code);
```
akosthekiss: Is there a specific reason to move the `__noreturn` attribute to the beginning of the declaration from the end?

jerry-core/include/jerryscript-port.h
```diff
@@ -20,6 +20,21 @@
...
+#define JERRY_ALLOCA(name, type, size) type name[size]
+#endif
+
+#ifdef _MSC_VER
+#define __noreturn __declspec(noreturn)
+#define __attr_format___(...)
+#else
+#define __noreturn __attribute__ ((noreturn))
+#define __attr_format___(...) __attribute__ ((format (__VA_ARGS__)))
+#endif
```
akosthekiss: FYI: I've proposed a solution to handle the pollution of -port.h in #2313 . Should that be accepted, then these compiler-specific definitions could move to the new header. (Because as it was discussed, they simply don't look right here. Especially `JERRY_ALLOCA`, as it kind of "abuses" jerryscript-port.h to make sure that the macro is visible everywhere.)

jerry-core/include/jerryscript-port.h
akosthekiss: What syntax is accepted? Is `__declspec() type funcname()` the only correct way? In jrt.h, several functions are declared as `type __noreturn funcname()` suggesting that MSVC also accepts `type __declspec() funcname()`. If that's the case, I suggest to place attributes consistently here, too. (And perhaps even some lines below, for `jerry_port_log`.)

akosthekiss:
@zherczeg @LaszloLango I think that this PR starts getting a good shape (pending changes should #2313 get accepted). As noted in that other PR, naming things does matter, so I raise a similar question here, too. Are we happy with the suggested new `JERRY_ALLOCA` name (wherever it may be defined)? Should it be
- `JERRY_ALLOCA`, as proposed here, or
- `JERRY_VLA`, as most compilers do support VLAs in fact, MSVC being the only outlier currently that needs `alloca`?

Should we use
- lower case (although I suggested to have it upper case, but jerryscript-comp.h has everything else all lower case),
- some double underscore prefix instead of the `JERRY_` prefix?

IMO we should try to ensure that we conclude on a logical and consistently applied naming scheme.

lygstate:
Rebased by the contribution from @akosthekiss, an good patch about jerryscript-compiler.h
Now I only need add JERRY_ALLOCA for fixing the compiler warnings.

jerry-core/include/jerryscript-compiler.h
```diff
@@ -47,6 +47,11 @@ extern "C"
 
 #endif /* __GNUC__ */
 
+#ifdef _MSC_VER
+void * __cdecl _alloca (size_t _Size);
+#define JERRY_ALLOCA(name, type, size) type *name = (type *)(_alloca (sizeof(type)*size))
```
akosthekiss: Note 1: Cppcheck apparently doesn't check the contents of the macro but please follow JerryScript coding guidelines there, too: `type *name = (type *) (_alloca (sizeof (type) * size))` (If I'm not mistaken. Reference: https://github.com/jerryscript-project/jerryscript/blob/master/docs/08.CODING-STANDARDS.md )

Note 2: In a previous version of this PR, weren't there definitions for some attributes, too? According to the docs, MSVC has declspecs like `align`, `deprecated`, `noinline`, `noreturn`. Can they be applied here?

Note 3: In the other PR, it's been discussed whether ALLOCA or VLA is the name to use. I'm not sure whether we've reached conclusion but let's keep in mind that the nema still might change. (@zherczeg, opinion?)

Note 4: Current param order of the macro is `(name, type, size)`. In both implementations (MSVC, GCC) the order of syntactic elements is `type ... name ... size` so it would be more natural for the macro arguments to follow this order, too.

jerry-ext/arg/arg.c
```diff
@@ -16,7 +16,7 @@
 #include "arg-internal.h"
 #include "jerryscript-ext/arg.h"
 #include "jerryscript.h"
-
+#include "jerryscript-port.h"
```
akosthekiss: Is this still necessary? `jerryscript.h` includes `jerryscript-compiler.h`, so the macros are visible. Moreover, `jerryscript-port.h` should only be included when port functions are used (which is not the case here).

Same comment goes to module.c, too.

lygstate: include jerryscript-compiler.h instead? OK, give a try.

akosthekiss: No, there should be no need for any new include: jerryscript-compiler.h should already be included transitively.

jerry-core/include/jerryscript-compiler.h
akosthekiss: For the record: I also vote for VLA

akosthekiss:
@lygstate Could you please elaborate on what "there is a lot of changes about JERRY_ATTR_ALWAYS_INLINE" means?

CMakeLists.txt
```diff
@@ -199,9 +216,16 @@ if(JERRY_LIBC)
   jerry_add_link_flags(-nostdlib)
 endif()
 
+if(USING_MSVC)
+  jerry_add_link_flags(/OPT:NOREF)
+  # Disable msvc  warning 4996 warning globally because it's stop us to use standard
+  # c function, refer to https://msdn.microsoft.com/en-us/library/ttcz0bys.aspx
```
akosthekiss: I'd suggest some grammar fixes even though I'm not a native speaker myself:

Disable MSVC warning 4996 globally because it stops us from using standard C functions

BTW, I've read the linked doc and found:
- "C4996 occurs when the compiler encounters a function or variable that is marked as deprecated. ... You can turn this warning off with the warning pragma or the /wd4996 command line option.": If that's the case, then adding the deprecated declspec in -compiler.h will not be really useful.
- "Some POSIX functions in the CRT have been renamed to conform with C99 and C++03 rules": I've always thought that we are C99 conforming. What are the problematic functions?

lygstate: That's the nasty thing of MSVC, such as `fopen` would raise warning.
it's would recommend you to use fopen_s.
And there is a lot more, such as strcpy_s and so on.
basically it's about the standard  ISO/IEC TR24731
http://www.open-std.org/jtc1/sc22/wg14/www/docs/n1225.pdf

lygstate:
@akosthekiss  MSVC have __forceinline
https://msdn.microsoft.com/en-us/library/z8y1yy88.aspx,
and when the type is potiner. __forceinline can only be
placed at the very beginning.

lygstate:
@akosthekiss ABOUT always_inline refer to change
https://github.com/lygstate/jerryscript/commit/06b45c8476f761b305d19460bc50d81b1260327c

akosthekiss:
@lygstate I've taken a quick look at the linked commit. I'm not sure that it will be possible to get accepted as is. It removes `inline` specifiers, which changes the code for non-MSVC builds. Moving around attribute markups is one thing (although it would have been nice to get it right in one go, for all attributes, when -compiler.h got introduced), but removing inline specifiers is another -- and that cannot happen.

akosthekiss:
That may be true, but that's a MSVCism that must not affect the project as a whole.

lygstate:
OK then I would give up that apply

lygstate:
@zherczeg any feedback? @akosthekiss  or just merge as is?

akosthekiss:
@lygstate I'm still voting for a name change: JERRY_ALLOCA to JERRY_VLA .

lygstate:
OK then vla

jerry-core/include/jerryscript-compiler.h
```diff
@@ -110,6 +120,10 @@ extern "C"
 #define JERRY_UNLIKELY(x) (x)
 #endif /* !JERRY_UNLIKELY */
 
+#ifndef JERRY_VLA
+#define JERRY_VLA(type, name, size) type name[size]
+#endif /* JERRY_VLA */
```
akosthekiss: style: comment needs a ! prefix, i.e. `/* !JERRY_VLA */`

lygstate: good point

lygstate: done and also fixes GCC

jerry-core/include/jerryscript-compiler.h
```diff
@@ -45,7 +45,17 @@ extern "C"
...
+
+#ifdef _MSC_VER
+void * __cdecl _alloca (size_t _Size);
+
+#define JERRY_ATTR_DEPRECATED __declspec(deprecated)
+#define JERRY_ATTR_NOINLINE __declspec(noinline)
+#define JERRY_ATTR_NORETURN __declspec(noreturn)
+
+#define JERRY_VLA(type, name, size) type *name = (type *) (_alloca (sizeof (type) * size))
+#endif /* !_MSC_VER */
```
akosthekiss: This update has brought in unintended changes. The comments after `#endif`s "repeat" the condition. Here, and above at GCC's conditional block, the condition is whether `__GNUC__` or `_MSC_VER` is defined, so that's what should be repeated at the corresponding endifs. Below, however, the condition is whether JERRY_VLA is _not_ defined, so the endif's comment is `!JERRY_VLA`.

yichoi:
Before landing this, please think about CI for windows. otherwise this could be one time working version.

yichoi:
Ok. any other suggestions? instead of appveyor

lygstate:
no other alternatives

akosthekiss:
@yichoi Are you aware of some shortcomings of AppVeyor? It's worked well for me in other projects, and we have already tried the current patch in personal AppVeyor accounts and worked well. We can set up a free project account and then we can have that running continuously.

yichoi:
No. I was wondering the availability of other alternatives, in order to avoid usual decision. Thanks for your efforts @lygstate

PR #2312, opened 2018-05-06
Fix documentation of port API

akosthekiss:
The patch fixes both the markdown documentation and the doc comment
in the port header.

The patch also removes an internal configuration macro guard
exposed by the public port header.

### DIFF
--- a/docs/05.PORT-API.md
+++ b/docs/05.PORT-API.md
@@ -11,7 +11,8 @@ It is questionable whether a library should be able to terminate an application.
  *
  * @param code gives the cause of the error.
  *
- * Note: jerry expects the function not to return.
+ * Note:
+ *      Jerry expects the function not to return.
  *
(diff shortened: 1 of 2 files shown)


docs/05.PORT-API.md
```diff
@@ -83,13 +84,24 @@ typedef struct
  *
  * @return true  - if success
  *         false - otherwise
+ *
+ * Note:
+ *      This port function is called by jerry-core when
+ *      CONFIG_DISABLE_DATE_BUILTIN is _not_ defined. Otherwise this function is
+ *      not used.
```
zherczeg: Note should be before return (similar cases below).

akosthekiss: If that's the case then the doc of `jerry_port_fatal` will have to be fixed as it has `@param` before the note. Should we take that approach?

yichoi:
LGTM

PR #2305, opened 2018-05-02
Splitting the debugger and console part of the python debugger client.

TamasZakor:
JerryScript-DCO-1.0-Signed-off-by: Tamas Zakor ztamas@inf.u-szeged.hu

PR #2313, opened 2018-05-07
Add jerryscript-compiler.h public header to cover compiler incompatibilities

akosthekiss:
In general, public headers should not have compiler-specific
constructs but both the core and the port headers have attributes,
which are non-standard. It's better to factor out such constructs
to a common place (a new header) and hide them behind macros, which
can then be defined on a per-compiler basis.

This patch moves the existing definitions of function attributes and
likely/unlikely builtins to the new header. At the same time, it
unifies the names of these attribute defines. Moreover, it touches
on jerry-main and removes the uses of `__attribute__((unused))`
entirely and replaces them with the elsewhere used `(void) ...`
pattern.

### DIFF
--- a/jerry-core/include/jerryscript-compiler.h
+++ b/jerry-core/include/jerryscript-compiler.h
@@ -1,0 +1,120 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef JERRYSCRIPT_COMPILER_H
+#define JERRYSCRIPT_COMPILER_H
+
+#ifdef __cplusplus
+extern "C"
+{
+#endif /* __cplusplus */
+
+/** \addtogroup jerry-compiler Jerry compiler compatibility components
+ * @{
+ */
+
+#ifdef __GNUC__
+
+/*
+ * Compiler-specific macros relevant for GCC.
+ */
+#define JERRY_ATTR_ALIGNED(ALIGNMENT) __attribute__((aligned(ALIGNMENT)))
+#define JERRY_ATTR_ALWAYS_INLINE __attribute__((always_inline))
+#define JERRY_ATTR_CONST __attribute__((const))
+#define JERRY_ATTR_DEPRECATED __attribute__((deprecated))
+#define JERRY_ATTR_FORMAT(...) __attribute__((format(__VA_ARGS__)))
+#define JERRY_ATTR_HOT __attribute__((hot))
+#define JERRY_ATTR_NOINLINE __attribute__((noinline))
+#define JERRY_ATTR_NORETURN __attribute__((noreturn))
+#define JERRY_ATTR_PURE __attribute__((pure))
+#define JERRY_ATTR_SECTION(SECTION) __attribute__((section(SECTION)))
+#define JERRY_ATTR_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
+
+#define JERRY_LIKELY(x) __builtin_expect(!!(x), 1)
+#define JERRY_UNLIKELY(x) __builtin_expect(!!(x), 0)
+
+#endif /* __GNUC__ */
+
+/*
+ * Default empty definitions for all compiler-specific macros. Define any of
+ * these in a guarded block above (e.g., as for GCC) to fine tune compilation
+ * for your own compiler. */
+
+/*
+ * Attributes
+ */
+#ifndef JERRY_ATTR_ALIGNED
+#define JERRY_ATTR_ALIGNED(ALIGNMENT)
+#endif /* !JERRY_ATTR_ALIGNED */
+
+#ifndef JERRY_ATTR_ALWAYS_INLINE
+#define JERRY_ATTR_ALWAYS_INLINE
+#endif /* !JERRY_ATTR_ALWAYS_INLINE */
+
+#ifndef JERRY_ATTR_CONST
+#define JERRY_ATTR_CONST
+#endif /* !JERRY_ATTR_CONST */
+
+#ifndef JERRY_ATTR_DEPRECATED
+#define JERRY_ATTR_DEPRECATED
+#endif /* !JERRY_ATTR_DEPRECATED */
+
+#ifndef JERRY_ATTR_FORMAT
+#define JERRY_ATTR_FORMAT(...)
+#endif /* !JERRY_ATTR_FORMAT */
+
+#ifndef JERRY_ATTR_HOT
+#define JERRY_ATTR_HOT
+#endif /* !JERRY_ATTR_HOT */
+
+#ifndef JERRY_ATTR_NOINLINE
+#define JERRY_ATTR_NOINLINE
+#endif /* !JERRY_ATTR_NOINLINE */
+
+#ifndef JERRY_ATTR_NORETURN
+#define JERRY_ATTR_NORETURN
+#endif /* !JERRY_ATTR_NORETURN */
+
+#ifndef JERRY_ATTR_PURE
+#define JERRY_ATTR_PURE
+#endif /* !JERRY_ATTR_PURE */
+
+#ifndef JERRY_ATTR_SECTION
+#define JERRY_ATTR_SECTION(SECTION)
+#endif /* !JERRY_ATTR_SECTION */
+
+#ifndef JERRY_ATTR_WARN_UNUSED_RESULT
+#define JERRY_ATTR_WARN_UNUSED_RESULT
+#endif /* !JERRY_ATTR_WARN_UNUSED_RESULT */
+
+/*
+ * Condition likeliness, unlikeliness
+ */
+#ifndef JERRY_LIKELY
+#define JERRY_LIKELY(x) (x)
+#endif /* !JERRY_LIKELY */
+
+#ifndef JERRY_UNLIKELY
+#define JERRY_UNLIKELY(x) (x)
+#endif /* !JERRY_UNLIKELY */
+
+/**
+ * @}
+ */
+
+#ifdef __cplusplus
+}
+#endif /* __cplusplus */
+#endif /* !JERRYSCRIPT_COMPILER_H */
--- a/jerry-core/include/jerryscript-port.h
+++ b/jerry-core/include/jerryscript-port.h
@@ -20,6 +20,8 @@
 #include <stdint.h>
 #include <stdio.h>
 
+#include "jerryscript-compiler.h"
+
 #ifdef __cplusplus
 extern "C"
 {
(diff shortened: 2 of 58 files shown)


akosthekiss:
This PR is motivated by #2305, to find a way for not polluting the port header
with compiler-specific `#ifdef`s and not-really-port-API-related macros (like
`JERRY_ALLOCA` and `__attr_format___`).

Open questions:
- The new header does not use any prefixes for the symbols it defines. This was
  no real issue until these symbols were internal to jerry-core but should they
  not follow the `jerry_` pattern now? (However, that would cause a lot of
  non-functional change to the code base.)
- What exactly should this new header contain? Is it OK to cover compiler
  incompatibilities only (attributes, builtins), or should it also expose some
  "useful" but non-compiler-specific defines (e.g., `JERRY_UNUSED`)? In that
  case, is the name of this header OK, or should it rather be jerryscript-rt.h
  or jerryscript-util.h?
- There are additional compiler-specific parts in jerry-core but not in the
  public headers. Should they also be factored out to a common place somehow?
  Should there be perhaps a public header with compiler-specific parts (the one
  proposed in this PR) but also an internal one? Perhaps the public header
  should not even contain everything what's proposed here?

N.B.: Two of the above questions deal with naming things "only". But it is
important to get to a good conclusion now since the names/IDs will stick and be
with us for a while. They will become part of the public API and will be hard to
change. So, this PR needs _comments_!

jerry-core/include/jerryscript-comp.h
```diff
@@ -0,0 +1,56 @@
...
+#define __attr_noreturn___ __attribute__((noreturn))
+#define __attr_pure___ __attribute__((pure))
+#define __attr_section___(SECTION) __attribute__((section(SECTION)))
+#define __attr_warn_unused_result___ __attribute__((warn_unused_result))
+
+/*
+ * Condition likeliness, unlikeliness
+ */
+#define likely(x) __builtin_expect(!!(x), 1)
+#define unlikely(x) __builtin_expect(!!(x), 0)
```
zherczeg: The original code has many ifdefs. Why do we remove them? __builtin_expect is very much gcc sepcific.

akosthekiss: Actually, not that many (or not that many useful). The attribute-related ifndefs are legacy to be removed, I'm quite sure. I'v tried to track back their origin last week and it turned out that they are leftovers from some 3.5+ years. They don't seem to be any useful but nobody dared to touch them, like ever. Now I dare...

As for the `__builtin_expect` issue. It is partially right that I have been sloppy, but it was also partially intentional. The builtins are compiler dependent indeed, but our current macro guards are imprecise / overly strict. E.g., clang also has that builtin, but the existing `__GNUC__` guard will exclude clang, if I'm not mistaken.

So, for now at least, I deliberately dropped the macro guards as gcc and clang are the officially maintained and tested compilers (Travis CI is green!), accepting that we may see some follow up patches from hopefully-not-too-upset users of different compilers. But this time, we will hopefully have more precise guards and only where they are needed and all in one place. At least, that was my motivation here.

LaszloLango:
Thanks for this patch. Here is my opinion
1) I think the prefix is need. `jerry_attr_` or `__jerry_attr_` are both good for me.
2) I do not have strong preferences, but `jerryscript-util.h` with commonly used defines could be great for JerryScript users.
3) Yes, they should and the new header should contain all of the compiler specific parts.

akosthekiss:
@LaszloLango thanks for the feedback.
- How about `_jerry_attr_xxx_` and `_jerry_[un]likely_` then? Note that with the double leading underscores we have been in violation of the C standard right from the beginning (which reserves those identifiers for the standard).
- -util.h is fine by me, too.
- If there is going to be an internal compiler-specific header as well, then we have to decide what goes there and what remains here in the public one.

akosthekiss:
Rebased and updated. Changed the name of the new header to jerryscript-util.h and the placement of attributes in -port.h. Still looking for feedback on the naming of the now published macros (and/or anything else).

akosthekiss:
Anyone interested, please speak up. I'll wait now a bit for feedbacks so that the header doesn't have to be renamed too many times.

akosthekiss:
@zherczeg `-defines.h` can also be an option, but it must be `jerryscript-defines.h` (not `jerry-`, which is deprecated). And let me add `-defs.h` to the list of options (it might be just me but `-defines.h` is just a bit long for me -- but that's highly subjective).

martijnthe:
Some thoughts:

- prefix: I'd be in favor of having a prefix to avoid clashes with other code bases / libraries.
- naming / what should it contain: in my experience, the more generic a file name is, the higher the likelihood of it becoming a messy catch-all over time. `...-defines.h` or `...-util.h` is too generic, IMHO. I'd suggest having multiple, smaller files with very clear names, i.e. `jerryscript-attributes.h`, `jerryscript-builtins.h`, etc. I can also imagine a `jerryscript-compiler.h` file with both macros for attributes as well as builtins.

zherczeg:
`jerryscript-compiler.h` sounds good to me

lygstate:
+1 for `jerryscript-compiler.h`, it's really was handling compiler incompatible.

akosthekiss:
It seems we have come (almost) full circle as the original header name was `jerryscript-comp.h`, where comp stood for comp{iler,atibility,onent}. `-compiler.h` it will be then.

lygstate:
jerry_attr_*
& jerry_likely
& jerry_vla

zherczeg:
I would prefer JERRY_COMPILER_XXX. AFAIK underscore prefix is reserved for C internals by the standard. Furthermore since this is a public header for internal use (there is not even plan for documenting it), I prefer longer names to reduce the chance of symbol clashes.

akosthekiss:
@zherczeg Actually, single underscores are only reserved by the standard at the file scope (so, not everywhere, e.g., you may have underscored local variables). Too bad that our macros will indeed be defined at the file scope. On the other hand, that did not prevent us from using double underscore prefixes before (`__attr_`), which is an even more serious violation of the standard (`__` is undonditionally reserved by the standard). So, I could still accept single underscores, but I'm not pushing for it if it's only me.

Regarding publicness, I'm not sure that it will be that internal. I think jerry-ext, jerry-port, or jerry-main may want to use these. And then all users may want to use it. So, we may even want to document it.

For now, I'm voting for the shorter and lowercase versions for the sake of usability. As the `jerry_` prefix namespace is in our hands, symbol clashes are not an issue.

jerry-core/include/jerryscript-compiler.h
```diff
@@ -0,0 +1,56 @@
...
+#define JERRY_ATTR_NORETURN __attribute__((noreturn))
+#define JERRY_ATTR_PURE __attribute__((pure))
+#define JERRY_ATTR_SECTION(SECTION) __attribute__((section(SECTION)))
+#define JERRY_ATTR_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
+
+/*
+ * Condition likeliness, unlikeliness
+ */
+#define JERRY_LIKELY(x) __builtin_expect(!!(x), 1)
+#define JERRY_UNLIKELY(x) __builtin_expect(!!(x), 0)
```
zherczeg: I like these names. But my question still stands: what happens on non-GCC systems? (Btw clang supports GNU macro)

lygstate: MSVC doesn't support, define it to empty.

jerry-core/include/jerryscript-port.h
```diff
@@ -96,7 +98,7 @@ typedef enum
  * Example: a libc-based port may implement this with vfprintf(stderr) or
  * vfprintf(logfile), or both, depending on log level.
  */
-void jerry_port_log (jerry_log_level_t level, const char *format, ...) __attribute__ ((format (printf, 2, 3)));
+void JERRY_ATTR_FORMAT (printf, 2, 3) jerry_port_log (jerry_log_level_t level, const char *format, ...);
```
lygstate: jerry_port_log new line?

jerry-core/include/jerryscript-compiler.h
akosthekiss: Which will (have to) happen in #2306

jerry-core/include/jerryscript-port.h
akosthekiss: Please explain, I'm not sure I know what you mean.

jerry-core/include/jerryscript-compiler.h
akosthekiss: @zherczeg My previous answer is hopefully still an answer:
> So, for now at least, I deliberately dropped the macro guards as gcc and clang are the officially maintained and tested compilers (Travis CI is green!), accepting that we may see some follow-up patches from hopefully-not-too-upset users of different compilers. But this time, we will hopefully have more precise guards and only where they are needed and all in one place.

So, I'm happy to take the blame for some fallout and review the upcoming PRs that add support for non-gcc and non-clang compatible compilers.

jerry-core/include/jerryscript-port.h
lygstate:
```
JERRY_ATTR_FORMAT (printf, 2, 3) 
```
Are to long in a single line with jerry_port_log (jerry_log_level_t level, const char *format, ...);

Are the following style acceptble?
```c
void JERRY_ATTR_FORMAT (printf, 2, 3)
jerry_port_log (jerry_log_level_t level, const char *format, ...);

```

akosthekiss: That 's a possible way to write it but our line length limits are 120 chars. In the headers, we usually keep one declaration in one line if the limit allows, so I see no point in changing the style here (especially because it was written in a single line before the change).

jerry-core/include/jerryscript-compiler.h
akosthekiss: Well, OK, let's think ahead a bit. Should the macros in `-compiler.h` be opt-in or opt-out?

Opt-in:
```c
/* Compiler-specific macros relevant for GCC. */
#ifdef __GNUC__
#define JERRY_ATTR_ALIGNED(ALIGNMENT) __attribute__((aligned(ALIGNMENT)))
// snip...
#define JERRY_LIKELY(x) __builtin_expect(!!(x), 1)
// snip...
#endif

/* Default empty definitions for all compiler-specific macros. Define any of
 * these in a guarded block above (e.g., as for GCC) to fine tune compilation
 * for your own compiler. */
#ifndef JERRY_ATTR_ALIGNED
#define JERRY_ATTR_ALIGNED(ALIGNMENT)
#endif

#ifndef JERRY_LIKELY
#define JERRY_LIKELY(x)
#endif
```
This is safe as all macros default to the empty string, so all compilers are supported by default. But the default will be sub-optimal then.

Opt-out:
```c
/* Default definitions for compiler-specific macros, whatever is supported by
 * GCC. If your compiler doesn't support any of these or uses different
 * notation, redefine them in a guarded block below. */
#define JERRY_ATTR_ALIGNED(ALIGNMENT) __attribute__((aligned(ALIGNMENT)))
// snip...
#define JERRY_LIKELY(x) __builtin_expect(!!(x), 1)
// snip...

/* Compiler-specific macros for MSVC. Define almost everything as empty because
 * of the differences in the __attribute__ and __declspec syntax. */
#ifdef _MSC_VER
#undef JERRY_ATTR_ALIGNED
#define JERRY_ATTR_ALIGNED(ALIGNMENT)

#undef JERRY_LIKELY
#define JERRY_LIKELY(x)
#endif
```
This is optimal for the officially supported/tested compilers (GCC and Clang). But it needs more maintenance if anything changes (e.g., all compiler blocks have to be adapted if a new attribute gets used, otherwise we risk breaking them).

Or will the reality become much more complex than opt-on or opt-out?

lygstate: I like the Opt-in one:)

akosthekiss:
I've rebased the PR and even added defaults for non-GCC systems.

lygstate:
+1

yichoi:
LGTM

PR #2315, opened 2018-05-07
Use logging macros in jerry-core/debugger

akosthekiss:
...instead of directly calling the logging port API function.


yichoi:
LGTM

PR #2316, opened 2018-05-08
Simplify code paths within PARSER_DUMP_BYTE_CODE

akosthekiss:
Also fix an uncovered issue in the reporting of switching to strict
mode.

### DIFF
--- a/jerry-core/parser/js/js-parser-util.c
+++ b/jerry-core/parser/js/js-parser-util.c
@@ -86,17 +86,18 @@ void
 parser_flush_cbc (parser_context_t *context_p) /**< context */
 {
   uint8_t flags;
+  uint16_t last_opcode = context_p->last_cbc_opcode;
 
-  if (context_p->last_cbc_opcode == PARSER_CBC_UNAVAILABLE)
+  if (last_opcode == PARSER_CBC_UNAVAILABLE)
   {
     return;
   }
 
   context_p->status_flags |= PARSER_NO_END_LABEL;
 
-  if (PARSER_IS_BASIC_OPCODE (context_p->last_cbc_opcode))
+  if (PARSER_IS_BASIC_OPCODE (last_opcode))
   {
-    cbc_opcode_t opcode = (cbc_opcode_t) context_p->last_cbc_opcode;
+    cbc_opcode_t opcode = (cbc_opcode_t) last_opcode;
 
     JERRY_ASSERT (opcode < CBC_END);
     flags = cbc_flags[opcode];
@@ -106,7 +107,7 @@ parser_flush_cbc (parser_context_t *context_p) /**< context */
   }
   else
   {
-    cbc_ext_opcode_t opcode = (cbc_ext_opcode_t) PARSER_GET_EXT_OPCODE (context_p->last_cbc_opcode);
+    cbc_ext_opcode_t opcode = (cbc_ext_opcode_t) PARSER_GET_EXT_OPCODE (last_opcode);
 
     JERRY_ASSERT (opcode < CBC_EXT_END);
     flags = cbc_ext_flags[opcode];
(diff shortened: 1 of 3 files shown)


jerry-core/parser/js/js-parser-util.c
```diff
@@ -572,7 +542,9 @@ parser_emit_cbc_backward_branch (parser_context_t *context_p, /**< context */
 #ifdef PARSER_DUMP_BYTE_CODE
   if (context_p->is_show_opcodes)
   {
-    JERRY_DEBUG_MSG ("  [%3d] %s\n", (int) context_p->stack_depth, name);
+    JERRY_DEBUG_MSG ("  [%3d] %s\n",
+                     (int) context_p->stack_depth,
+                     PARSER_IS_BASIC_OPCODE (opcode) ? cbc_names[opcode] : cbc_ext_names[opcode]);
```
zherczeg: Are you sure `PARSER_IS_BASIC_OPCODE (opcode)` is correct?
`opcode = (uint16_t) PARSER_GET_EXT_OPCODE (opcode);` above modifies the opcode

akosthekiss: Hmm, I've overlooked this. I'll take the approach of `parser_emit_cbc_forward_branch` then.

akosthekiss:
PR rebased and updated according to the review.

jerry-core/parser/js/js-parser-util.c
```diff
@@ -524,9 +505,7 @@ parser_emit_cbc_backward_branch (parser_context_t *context_p, /**< context */
                                  uint32_t offset) /**< destination offset */
 {
   uint8_t flags;
-#ifdef PARSER_DUMP_BYTE_CODE
-  const char *name;
-#endif /* PARSER_DUMP_BYTE_CODE */
+  uint32_t extra_byte_code_increase;
```
zherczeg: I don't really like this isnce it affects the non-dump code path. Why can't we leave the original here?

zherczeg: Or create an early boolean which tells its a basic or extended opcode. That is still somewhat less code.

akosthekiss: My motivation for this whole PR was to reduce the scattering of `PARSER_DUMP_BYTE_CODE`-guarded blocks, and group all dump-related code into bigger chunks. It turned out that most of the "scattering" was caused by variables that were introduced for the sake of `PARSER_DUMP_BYTE_CODE` only. At most places, it was possible to get rid of these variables, but here, my original approach obviously overlooked a re-assignment to `opcode`, so it had to be changed. As there already exists a similar function, `parser_emit_cbc_forward_branch `, which starts almost identically to this function, `parser_emit_cbc_backward_branch`, I thought it would be beneficial to use the same approach here as well. Please, see here: https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/parser/js/js-parser-util.c#L419 .

akosthekiss: re. "Or create an early boolean which tells its a basic or extended opcode.": wouldn't that be even _more_ code? This integer, which contains 0 or 1, can be directly added to another integer (to `byte_code_size`). But if this would be a bool, then we would either need an `if`, or a `+ (early_boolean ? 1 : 0)`. That seems to be more complex that `+ extra_byte_code_increase`. (Or did I get your suggestion wrong?)

zherczeg: No, since this bool is only available when dump byte code is enabled.

akosthekiss: That doesn't really help the goals of the PR then. To replace one guarded variable with another. What do you think about the other function I mentioned `parser_emit_cbc_forward_branch`. Is that a bad example?

zherczeg: Yes, since the approach for forward and backward branches are different. As for forward branches their target is unknown, but it is known for backward branches. This is why we have two separate functions for them.

My problem with your approach is removing code optimizations for a feature which is very rarely enabled. If less ifdefs would be the aim of Jerry, the whole code base would look different.

akosthekiss: Well, only for the sake of argument, here are the parts of the code that affect the performance (let's assume that we throw away all asserts, all dumps, and don't consider the unchanged parts). Current `parser_emit_cbc_backward_branch`:
```c
context_p->byte_code_size++;
context_p->byte_code_size += 2;
```

Proposed `parser_emit_cbc_backward_branch`:
```c
uint32_t extra_byte_code_increase;
extra_byte_code_increase = 1;
context_p->byte_code_size += 2 + extra_byte_code_increase;
```

I'm not sure which one is more optimized: two increments through indirection, or a local variable, a local addition, and one increment through indirection. Whichever wins, both seem like very tiny micro-optimizations to me. (And although I know the aims of JerryScript, maintainability is/should be also on the priority list, even if not even near the top.)

But I can just drop this chunk from the commit.

akosthekiss:
PR updated.

jerry-core/parser/js/js-parser-util.c
```diff
@@ -167,18 +167,11 @@ parser_flush_cbc (parser_context_t *context_p) /**< context */
...
-    {
-      name_p = cbc_ext_names[PARSER_GET_EXT_OPCODE (context_p->last_cbc_opcode)];
-    }
-
-    JERRY_DEBUG_MSG ("  [%3d] %s", (int) context_p->stack_depth, name_p);
+    JERRY_DEBUG_MSG ("  [%3d] %s",
+                     (int) context_p->stack_depth,
+                     PARSER_IS_BASIC_OPCODE (context_p->last_cbc_opcode)
+                     ? cbc_names[context_p->last_cbc_opcode]
+                     : cbc_ext_names[PARSER_GET_EXT_OPCODE (context_p->last_cbc_opcode)]);
```
zherczeg: I realized this coding style is new for me. It is usually:
```
(a ? b
   : c)
```
Is it frequently used? In only found something like this in `re-compiler.c` but that looks even more strange.

akosthekiss:
Code rewritten to follow the suggested style for `?:`. PR rebased and updated.

yichoi:
LGTM

PR #2320, opened 2018-05-09
Fix some incorrect documentation

mmatyas:


jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -544,7 +544,7 @@ ecma_builtin_helper_string_prototype_object_index_of (ecma_value_t this_arg, /**
  *         - The ecma_builtin_helper_string_prototype_object_index_of helper routine.
  *         - The ecma_builtin_string_prototype_object_replace_match helper routine.
  *
- * @return uint32_t - the normalized value of the index
+ * @return bool - whether there is a match for the search string
```
akosthekiss: Would it be possible to add an `[out]` marker to the docstring of the `ret_index_p` param? As that's where this function is "returning" the found index now, instead of the return value. (Note: `/**< [out] ... */` is the proper doxygen syntax.)

PR #2323, opened 2018-05-10
Change the SDK on target ESP8266

rerobika:
The reasons for the change are the following:
  * The communication interfaces (UART, SPI) API documentation is defective.
  * The usage of the provided API functions are rather complicated. For example a simple method like receiving a character from UART buffer requires  an interrupt handler implementation.
  * The SDK uses FreeRTOS 7.5.2 which is outdated.

To overcome these difficulties the new [SDK](https://github.com/SuperHouse/esp-open-rtos) should be used for the ESP8266 target.

The new SDK solves the mentioned issues and provides many of benefits like:
  * Already existing modules like FatFs, SDIO, MbedTLS, SPIFFS can be used while making a new application.
  * It contains a `new-libc` which expands the toolchain's libc provided functionalities.
  * The builtin LwIP module provides POSIX-like interface thereby the jerry-debugger is now available for usage.
  * The communication interfaces' documentation and the provided API is easy to use and embed to an apllication.
  * Active developer community.

Besides the SDK change this patch:
  * Improves the JavaScript source processing.
  * Adds the possibility to redirect `stdout`.
  * Allows the usage of jerry-debugger.
  * Separates and refactors the already existing external handlers to a modular folder structure.
  * Extends the usage of `JERRY_CONST_DATA` macro.
  * Attempts to follow the JerryScript coding style as where as possible


rerobika:
The ESP8266 jobs fails on Travis, I'm going to fix it ASAP.

rerobika:
Thanks for the review. According to your comments I can agree that this PR should be splitted, especially for the consts handling. The SDK change and and the modularization quite depends on each other so I would keep them together.

So, I will close this PR and separate it based on your thoughts.

akosthekiss:
@rerobika Well, you could leave (could have left) this PR open to track the development of the SDK change and only open a new PR for the jerry-core const changes. (Reopening the PR is still possible.)

PR #2327, opened 2018-05-14
Remove 'useless' variable from function.bind

galpeter:
The `args_length` variable in the `ecma_builtin_function_prototype_object_bind`
method is not needed, the `arguments_number` variable can be
directly used.


yichoi:
LGTM

Issue #2213, opened 2018-02-20
JerryScript 2.0


zherczeg:
It would be good to release JerryScript 2.0 this year. No specific deadline so far.

What we should do before that is reworking the API. Some ideas:

- delete obsolote functions
- rework error flag to be an error value (error can only be a result of something, and cannot be used as input except unboxing its value).

martijnthe:
> delete obsolote functions

Which ones are those?

Quick search for "deprecated" through the code base revealed:
```
JERRY_DEPRECATED_API
bool jerry_get_object_native_handle (const jerry_value_t obj_val, uintptr_t *out_handle_p);
JERRY_DEPRECATED_API
void jerry_set_object_native_handle (const jerry_value_t obj_val, uintptr_t handle_p,
                                     jerry_object_free_callback_t freecb_p);
```

```
#pragma message ("using jerry-api.h directly is deprecated, please use jerryscript.h")
```

```
#pragma message ("using jerry-port.h directly is deprecated, please use jerryscript-port.h")
```

```
  JERRY_INIT_MEM_STATS_SEPARATE  = (1u << 3), /**< deprecated, an unused placeholder now */
  JERRY_INIT_DEBUGGER            = (1u << 4), /**< deprecated, an unused placeholder now */
```

Anything else (other than the "old"/current error flag APIs)?

zherczeg:
Mostly these. Maybe the parse variant where the source is not provided, but that is not necessary. We should emphasize that pass source info if available, and pass NULL if not.

It will be a much smaller scope API rework that JerryScript 1.0.

galpeter:
I've got a few suggestions for extra API functions/improvements:
* JSON C API - parse/stringify
* Regex construction from API / maybe even exec/test
* Allow jerry_is_feature_enabled to query builtin/es2015-subset objects
* Add DataView for es2015-subset

please note I'm not sure how widely they would be used or needed.

galpeter:
(IMHO) If we want to remove functions we should mark them as obsolete as a first step. This way we can (hopefully) notify the users that there will be a bigger change. Related to this I think we should consider writing a migration guide or at least a change list between v1 and v2.

zherczeg:
Snapshot generation API  rework #2259

Lets just collect the API rework patches here, and later we can write a changelog using these references.

zherczeg:
Snapshot executing API rework #2270

zherczeg:
Jerry parse rework: #2282

[...]

PR #2329, opened 2018-05-14
Remove legacy jerry_get_memory_limits API function and unused configuration macros

akosthekiss:
Also remove the legacy `jerry_get_memory_limits` API function as it
does not report either useful or valid information.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -665,47 +665,6 @@ main (void)
 - [jerry_parse_and_save_literals](#jerry_parse_and_save_literals)
 
 
-## jerry_get_memory_limits
-
-**Summary**
-
-Gets configured memory limits of JerryScript.
-
-**Prototype**
-
-```c
-void
-jerry_get_memory_limits (size_t *out_data_bss_brk_limit_p,
-                         size_t *out_stack_limit_p);
-```
-
-- `out_data_bss_brk_limit_p` - out parameter, that gives the maximum size of data + bss + brk sections.
-- `out_stack_limit_p` - out parameter, that gives the maximum size of the stack.
-
-**Example**
-
-[doctest]: # ()
-
-```c
-#include "jerryscript.h"
-
-int
-main (void)
-{
-  jerry_init (JERRY_INIT_EMPTY);
-
-  size_t stack_limit;
-  size_t data_bss_brk_limit;
-  jerry_get_memory_limits (&stack_limit, &data_bss_brk_limit);
-}
-```
-
-**See also**
-
-- [jerry_init](#jerry_init)
-- [jerry_cleanup](#jerry_cleanup)
-
-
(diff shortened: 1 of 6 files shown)


akosthekiss:
**NOTE: THIS IS AN API-BREAKING CHANGE!** (At least, theoretically.)

In practice, I cannot think of any user of JerryScript who would have used the `jerry_get_memory_limits` function. It has been introduced in Feb 11, 2015 (35fa39ceb8fc79b6ce33c07d753f3ff2c8d3d39a) but hasn't been of any use since Feb 13, 2015 (43ea53b1d7aad510bc547fc30c9912daddb6828b), when its counterpart `jrt_set_mem_limits` was macrod out to never get enabled ever again.

akosthekiss:
Speaking of config macros and breaking changes: is the `CONFIG_ECMA_GLOBAL_ENVIRONMENT_DECLARATIVE` setting used / useful? The only place where this is used at is the following. Which doesn't seem to be too standard. It's also been introduced several years ago (Sep 2014 (fbcd393962decd30fb2eb51f92e63dfa92031046)) but I'm not aware of any (current?) use cases.

```c
void
ecma_init_global_lex_env (void)
{
#ifdef CONFIG_ECMA_GLOBAL_ENVIRONMENT_DECLARATIVE
  JERRY_CONTEXT (ecma_global_lex_env_p) = ecma_create_decl_lex_env (NULL);
#else /* !CONFIG_ECMA_GLOBAL_ENVIRONMENT_DECLARATIVE */
  ecma_object_t *glob_obj_p = ecma_builtin_get (ECMA_BUILTIN_ID_GLOBAL);

  JERRY_CONTEXT (ecma_global_lex_env_p) = ecma_create_object_lex_env (NULL, glob_obj_p, false);

  ecma_deref_object (glob_obj_p);
#endif /* CONFIG_ECMA_GLOBAL_ENVIRONMENT_DECLARATIVE */
} /* ecma_init_global_lex_env */
```

I'm happy to include the removal of the macro and the associated code paths in this patch but feedback would be appreciated on this first.

zherczeg:
#2213

zherczeg:
I would emphasize the removal of `jerry_get_memory_limits` rather than macros in the title and commit log. Other than that is ok for me.

akosthekiss:
@zherczeg And what about the question raised in the comment: what's the deal with CONFIG_ECMA_GLOBAL_ENVIRONMENT_DECLARATIVE ?

zherczeg:
That macro makes an ecmascript program behave like a function without the argument and `return` support. It is non-standard.

Currently you can create functions directly, and I would recommend that if you want to use such a feature.

akosthekiss:
I've updated the PR according to the feedbacks:
- PR title and commit message changed
- removed CONFIG_ECMA_GLOBAL_ENVIRONMENT_DECLARATIVE from the code base
- rebased onto latest master

Issue #2251, opened 2018-03-24
float vs double


DJLinux:
(Sorry about my bad english)
Hello I build JerryScript with CodeBlocks for Windows and Linux and Raspberry PI 32/ 64-bit 
With support for double "-d" and float "-f" so the resulting libs are:
JerryScript-32-f.dll, JerryScript-32-d.dll, JerryScript-64-f.dll, JerryScript-64-d.dll
libJerryScript-32-f.so, libJerryScript-32-d.so, libJerryScript-64-f.so, libJerryScript-64-d.so

Of course for the FLOAT32 build I disabled the JS date functionality and it works all so far.

But I wonder me that jerry_create_value and jerry_get_value are hard coded for "double" ?
I mean if I build JerryScript  for my retro computer I have only float my retro compiler 
don't have the data type double  at all.

So why does JerryScript  use double if 
(CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32)
are defined ?

Thank you DJ

zherczeg:
The public API cannot rely on compiler defines. In most environments, when you include a library header, you don't provide the list of compiler defines.

The solution I can suggest is redefine double with float in your computer environment. If you don't have the type, it should not cause any problems.

DJLinux:
#if (CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64)

x = TO_ECMA_NUMBER_T (fabs (x));
x = TO_ECMA_NUMBER_T (acos (x));
x = TO_ECMA_NUMBER_T (asin (x));
x = TO_ECMA_NUMBER_T (atan (x));
x = TO_ECMA_NUMBER_T (ceil (x));
x = TO_ECMA_NUMBER_T (cos (x));
x = TO_ECMA_NUMBER_T (exp (x));
x = TO_ECMA_NUMBER_T (floor (x));
x = TO_ECMA_NUMBER_T (log (x));
...
#else
x = TO_ECMA_NUMBER_T (absf (x));
x = TO_ECMA_NUMBER_T (acosf (x));
x = TO_ECMA_NUMBER_T (asinf (x));
x = TO_ECMA_NUMBER_T (atanf (x));
x = TO_ECMA_NUMBER_T (ceilf (x));
x = TO_ECMA_NUMBER_T (cosf (x));
x = TO_ECMA_NUMBER_T (expf (x));
[...]

PR #2330, opened 2018-05-15
Use ecma_number_t instead of double where possible

akosthekiss:
Some code paths explicitly used double instead of ecma_number_t
even though the values assigned from or compared to were of type
ecma_value_t.

Related to #2251


yichoi:
LGTM

PR #2332, opened 2018-05-15
Remove jerry-libc

akosthekiss:
Rationale:
- There is no port under targets/ that would use it. All of them
  turn it off when building.
- That's no surprise, as jerry-libc supports no barebone MCUs but
  posix targets with syscalls only. Actually, that's Linux only,
  because macOS builds have turned off the use of jerry-libc a
  while ago.
- And there is no point in maintaining a highly restricted set of
  libc functions: as soon as someone wants to use JerryScript in a
  scenario that needs more functions than jerry-main, they have to
  choose a different libc (most problably the compiler's default
  one).

I think that we should not keep supporting an otherwise unused
library for the purposes of jerry-main on arm/x86/x64-linux  only.


lygstate:
agreed， we could use external tiny libc instead

zherczeg:
How does this affect measurements?

akosthekiss:
None of the measurements at https://jerryscript-project.github.io/jerryscript-test-results/?view=home use jerry-libc either.

zherczeg:
For me, it is ok to remove jerry libc.

yichoi:
I respect your opinion however I do not make sure what would happen in the near future. Please let me have more time to think about it.

zherczeg:
Perhaps we could close this bug then, and reopen if a positive decision is made.

daeyeon:
Assuming taking jerry in very restrictive environments where libc isn't used, with jerry-libc, jerry can be given as a minimal-sized static binary. IMHO, it needs to consider whether jerry still have the policy to support such an environment.

akosthekiss:
I think JerryScript, the engine, still has the goal to support restricted environments. Actually, that's its main goal IMO.

But that cannot really be achieved with jerry-libc as it hasn't been supporting restricted environments for quite a long time: it only supports x86/x64/arm-linux, none of which counts as a highly restricted environment. (Actually, if I'm not mistaken, in such platforms where dynamic linking is available, linking to a shared libc is more efficient, as that shared libc is already loaded into the memory anyway, one doesn't have to link the whole thing statically to each program and load it over and over again.)

Moreover, the function set of jerry-libc is very limited. If JerryScript is linked to an application that's only a little but more complex than jerry (the example test runner app in jerry-main), then jerry-libc must be replaced even on the supported platforms.

Finally, on the restricted environments (NuttX, TizenRT, MbedOS, FreeRTOS) with no desktop-like OS, where usually a single static firmware image is built and flashed indeed, well, there jerry-libc cannot be used as it simply doesn't support them. In those environments, even nowadays, all use cases use the libc that comes with the compiler.

Bottom line is: it's not jerry-libc that enables building the jerryscript engine into a small static binary. It's only a maintenance burden.

zherczeg:
> Bottom line is: it's not jerry-libc that enables building the jerryscript engine into a small static binary. It's only a maintenance burden.

This is unfortunately incorrect. You cannot make a small static binary without jerry libc. You probably thought about dynamic binary.

akosthekiss:
@zherczeg No, I meant what I wrote. It's not jerry-libc that will make your statically linked binary small (except for one single artificial case, perhaps). A _final binary_ wherein JerryScript the engine is used consists of at least 3 components: the libc, the engine, and the application. And the application is not under the control of this project, meaning it may potentially use a lot more libc functions than the engine or than the _example_ jerry-main application, in which case jerry-libc is useless. So, jerry-libc may enable building a small linux-based static jerry-main but that's all (that's the artificial use case). Roughly anywhere else the libc is out of our hands too, so whether the static binary is small or not does not depend on jerry-libc.

daeyeon:
There is another choice instead of replacing jerry-lib with any libc. It's also possible to add only functions the application relies on. (That could be stupid on most platforms though.) Imagine there is an application having its own specific libc-sets just like jerry engine.

As mentioned above, whether jerry needs to consider such an unusual or not-modern case is a question mark. If not necessary, it seems right to reduce that burden.

akosthekiss:
@daeyeon The scenario you've mentioned (to add functions specific to the application) results in an application-specific libc, which means it should rather be developed together with the application, not with one of its library components (like jerryscript).

And, at least as far as I know, it is not managable or is very fragile to have one set of libc functions for jerry-core to build against and use a different libc for the application. According to my best understanding, all components of an application should be built against the same libc headers/objects.

So, I'd still vote for the "reduce that burder" option.

akosthekiss:
PR has been rebased to latest master, conflicts have been resolved.

akosthekiss:
Rebased again to resolve new conflicts. Thanks for the discussions to all parties involved. Are we getting closer to a decision about the PR?

akosthekiss:
Rebased to prevent bitrotting.

akosthekiss:
Rebased to prevent bitrotting.

Has anyone found a real use case for jerry-libc yet?...

akosthekiss:
Regular pinging, conflict-resolving, and bitrot-preventing rebase.

akosthekiss:
👍 🙇 ✌️

PR #2333, opened 2018-05-16
Move (almost all) Travis CI jobs to container-based infrastructure

akosthekiss:
Container-based VMs promise significantly faster boot times.

More or less related changes:
- Added a JOBNAME to all jobs.
- Added `install-noapt` target to several Makefile.travis files to
  avoid `sudo apt-get install ...` steps. Those installations are
  now handled by the apt addon of Travis. The `install` targets are
  kept anyway to keep the makefiles self-contained.
- Removed a legacy workaround from the Coverity Scan job as it
  isn't necessary anymore to fiddle with the cerificates of
  scan.coverity.com.
- Fixed the Mbed and the Zephyr jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,18 +1,12 @@
 language: c
 
-# Default test platform: Ubuntu Trusty with sudo support.
+# Default environment: Container-based (sudo-less) Ubuntu Trusty 14.04.
 os: linux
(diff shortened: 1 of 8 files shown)


yichoi:
LGTM

PR #2335, opened 2018-05-16
Revert the Mbed Travis CI job to sudo-based VM to avoid sporadic OOMs

akosthekiss:


akosthekiss:
Gcc on CBI seems to nondeterministically die during the Mbed job, most probably because of OOM conditions (CBI has less memory than sudo-enabled VMs). To avoid the need of repeated manual restarts (which always solves the problem eventually, but it's cumbersome), let's switch back to sudo-enabled VM for Mbed. Sorry for the noise.

PR #2337, opened 2018-05-17
Fix parameter constness differences in function declarations and definitions

akosthekiss:


yichoi:
LGTM

PR #2339, opened 2018-05-18
Simplify run-time feature checks in unit tests

akosthekiss:

### DIFF
--- a/tests/unit-core/test-arraybuffer.c
+++ b/tests/unit-core/test-arraybuffer.c
@@ -69,25 +69,6 @@ assert_handler (const jerry_value_t func_obj_val, /**< function object */
 } /* assert_handler */
 
 /**
- * Checks whether global object has arraybuffer.
- */
-static bool
-arraybuffer_is_available (void)
-{
-  jerry_value_t global_obj_val = jerry_get_global_object ();
-  jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) "ArrayBuffer");
-
-  jerry_value_t prop_value = jerry_has_property (global_obj_val, prop_name);
-  bool has_prop = jerry_get_boolean_value (prop_value);
-
-  jerry_release_value (global_obj_val);
-  jerry_release_value (prop_name);
-  jerry_release_value (prop_value);
-
-  return has_prop;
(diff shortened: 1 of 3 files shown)


LaszloLango:
IMHO the unit tests should use the related API function instead: http://jerryscript.net/api-reference/#jerry_is_feature_enabled

akosthekiss:
During some development, I've run into this:
```
/home/travis/build/akosthekiss/jerryscript/tests/unit-core/test-typedarray.c: In function ‘main’:
/home/travis/build/akosthekiss/jerryscript/tests/unit-core/test-typedarray.c:475:7: error: ‘byte_length’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
       TEST_ASSERT (byte_length == element_count);
       ^
/home/travis/build/akosthekiss/jerryscript/tests/unit-core/test-typedarray.c:472:22: note: ‘byte_length’ was declared here
       jerry_length_t byte_length;
                      ^
lto-wrapper: /usr/bin/gcc returned 1 exit status
/usr/bin/ld: lto-wrapper failed
collect2: error: ld returned 1 exit status
```

It's strange that this hasn't appeared before, but the compiler is actually right. If jerry-core is built with the es5.1 profile, the body of some API functions is macrod out to be dummy. But some of them have out parameters, which are not filled in in this case. And LTO correctly recognizes this when jerry-core is linked to the unit tests. (Note: there are runtime feature checks in the unit tests, but the compiler cannot infer info from such checks.)

In the above example, the following code causes the problem:
```c
jerry_value_t buffer = jerry_get_typedarray_buffer (array, &offset, &byte_length);
TEST_ASSERT (byte_length == element_count);
```

@LaszloLango I'm not sure whether calling `jerry_is_feature_enabled` would help the LTO problem as that's also a runtime check.

akosthekiss:
BTW have y'all seen the appveyor checking out on the PR? :) ( @lygstate @yichoi )

akosthekiss:
It has turned out that the simpler `jerry_is_feature_enabled` approach worked as well. Thanks for the tip @LaszloLango . PR updated

lygstate:
see it need a badge on the home Page

yichoi:
LGTM

PR #2342, opened 2018-05-18
Cleanup code around JERRY_UNREACHABLEs

akosthekiss:
`JERRY_UNREACHABLE`s often signal code structure that could be
improved: they can usually either be rewritten to `JERRY_ASSERT`s
or eliminated by restructuring loops, `if`s or `#if`s. Roughly,
the only valid occurences are in default cases of `switch`es.

Moreover, it is dangerous to write meaningful code after
`JERRY_UNREACHABLE` because it pretends as if there was a way to
recover from an impossible situation.

This patch rewrites/eliminates `JERRY_UNREACHABLE`s where possible
and removes misleading code from after them.

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -100,11 +100,7 @@ static const char * const wrong_args_msg_p = "wrong type of argument";
 static inline void JERRY_ATTR_ALWAYS_INLINE
 jerry_assert_api_available (void)
 {
-  if (JERRY_UNLIKELY (!(JERRY_CONTEXT (status_flags) & ECMA_STATUS_API_AVAILABLE)))
-  {
-    /* Terminates the execution. */
-    JERRY_UNREACHABLE ();
-  }
+  JERRY_ASSERT (JERRY_CONTEXT (status_flags) & ECMA_STATUS_API_AVAILABLE);
 } /* jerry_assert_api_available */
 
 /**
@@ -178,11 +174,8 @@ jerry_throw (jerry_value_t value) /**< return value */
 void
 jerry_init (jerry_init_flag_t flags) /**< combination of Jerry flags */
 {
-  if (JERRY_UNLIKELY (JERRY_CONTEXT (status_flags) & ECMA_STATUS_API_AVAILABLE))
-  {
-    /* This function cannot be called twice unless jerry_cleanup is called. */
-    JERRY_UNREACHABLE ();
-  }
+  /* This function cannot be called twice unless jerry_cleanup is called. */
+  JERRY_ASSERT (!(JERRY_CONTEXT (status_flags) & ECMA_STATUS_API_AVAILABLE));
 
   /* Zero out all members. */
   memset (&JERRY_CONTEXT (JERRY_CONTEXT_FIRST_MEMBER), 0, sizeof (jerry_context_t));
@@ -834,21 +827,16 @@ jerry_value_get_type (const jerry_value_t value) /**< input value to check */
     {
       return JERRY_TYPE_FUNCTION;
     }
-    case LIT_MAGIC_STRING_OBJECT:
+    default:
     {
+      JERRY_ASSERT (lit_id == LIT_MAGIC_STRING_OBJECT);
+
       /* Based on the ECMA 262 5.1 standard the 'null' value is an object.
        * Thus we'll do an extra check for 'null' here.
        */
       return ecma_is_value_null (argument) ? JERRY_TYPE_NULL : JERRY_TYPE_OBJECT;
     }
-    default:
-    {
-      JERRY_UNREACHABLE ();
-      break;
-    }
   }
-
-  return JERRY_TYPE_NONE;
 } /* jerry_value_get_type */
 
 /**
--- a/jerry-core/ecma/base/ecma-helpers-value.c
+++ b/jerry-core/ecma/base/ecma-helpers-value.c
@@ -714,11 +714,6 @@ ecma_copy_value (ecma_value_t value)  /**< value description */
 {
   switch (ecma_get_value_type_field (value))
   {
-    case ECMA_TYPE_DIRECT:
-    case ECMA_TYPE_DIRECT_STRING:
-    {
-      return value;
-    }
     case ECMA_TYPE_FLOAT:
     {
       ecma_number_t *num_p = (ecma_number_t *) ecma_get_pointer_from_ecma_value (value);
@@ -737,8 +732,10 @@ ecma_copy_value (ecma_value_t value)  /**< value description */
     }
     default:
     {
-      JERRY_UNREACHABLE ();
-      return ECMA_VALUE_UNDEFINED;
+      JERRY_ASSERT (ecma_get_value_type_field (value) == ECMA_TYPE_DIRECT
+                    || ecma_get_value_type_field (value) == ECMA_TYPE_DIRECT_STRING);
+
+      return value;
     }
   }
 } /* ecma_copy_value */
@@ -905,13 +902,6 @@ ecma_free_value (ecma_value_t value) /**< value description */
 {
   switch (ecma_get_value_type_field (value))
   {
-    case ECMA_TYPE_DIRECT:
-    case ECMA_TYPE_DIRECT_STRING:
-    {
-      /* no memory is allocated */
-      break;
-    }
-
     case ECMA_TYPE_FLOAT:
     {
       ecma_number_t *number_p = (ecma_number_t *) ecma_get_pointer_from_ecma_value (value);
@@ -934,7 +924,10 @@ ecma_free_value (ecma_value_t value) /**< value description */
 
     default:
     {
-      JERRY_UNREACHABLE ();
+      JERRY_ASSERT (ecma_get_value_type_field (value) == ECMA_TYPE_DIRECT
+                    || ecma_get_value_type_field (value) == ECMA_TYPE_DIRECT_STRING);
+
+      /* no memory is allocated */
       break;
     }
   }
--- a/jerry-core/ecma/base/ecma-helpers.c
+++ b/jerry-core/ecma/base/ecma-helpers.c
@@ -784,8 +784,9 @@ ecma_free_property (ecma_object_t *object_p, /**< object the property belongs to
       ecma_free_value_if_not_object (ECMA_PROPERTY_VALUE_PTR (property_p)->value);
       break;
     }
-    case ECMA_PROPERTY_TYPE_NAMEDACCESSOR:
+    default:
     {
+      JERRY_ASSERT (ECMA_PROPERTY_GET_TYPE (*property_p) == ECMA_PROPERTY_TYPE_NAMEDACCESSOR);
 #ifdef JERRY_CPOINTER_32_BIT
       ecma_getter_setter_pointers_t *getter_setter_pair_p;
       getter_setter_pair_p = ECMA_GET_POINTER (ecma_getter_setter_pointers_t,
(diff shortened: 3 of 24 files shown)


akosthekiss:
depends on #2339

akosthekiss:
With #2339 merged, the PR now passes on Travis CI. Rebased.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -738,7 +738,7 @@ ecma_copy_value (ecma_value_t value)  /**< value description */
     default:
     {
       JERRY_UNREACHABLE ();
-      return ECMA_VALUE_UNDEFINED;
+      break;
```
zherczeg: This leads to undefined result.

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -797,7 +797,7 @@ ecma_free_property (ecma_object_t *object_p, /**< object the property belongs to
     default:
     {
       JERRY_UNREACHABLE ();
-      return;
+      break;
```
zherczeg: Why do you change this?

jerry-core/ecma/base/ecma-property-hashmap.c
```diff
@@ -539,8 +539,6 @@ ecma_property_hashmap_find (ecma_property_hashmap_t *hashmap_p, /**< hashmap */
       JERRY_ASSERT (entry_index != start_entry_index);
 #endif /* !JERRY_NDEBUG */
     }
-
-    JERRY_UNREACHABLE ();
```
zherczeg: Why can this be removed?

jerry-core/ecma/operations/ecma-objects-general.c
```diff
@@ -179,8 +179,6 @@ ecma_op_general_object_delete (ecma_object_t *obj_p, /**< the object */
     /* 5. */
     return ECMA_VALUE_FALSE;
   }
-
-  JERRY_UNREACHABLE ();
```
zherczeg: Is it ok to remove this?

jerry-core/vm/vm.c
```diff
@@ -2681,7 +2675,7 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
         default:
         {
           JERRY_UNREACHABLE ();
-          continue;
+          break;
```
zherczeg: Why do you change this?

jerry-core/ecma/base/ecma-helpers-value.c
akosthekiss: No, it doesn't cause undefined behaviour here either:
- https://github.com/jerryscript-project/jerryscript/blob/9eb9fbf5f12fba7511481d0cabb25a2020d7b710/jerry-core/vm/vm.c#L358
- https://github.com/jerryscript-project/jerryscript/blob/9eb9fbf5f12fba7511481d0cabb25a2020d7b710/jerry-core/ecma/builtin-objects/ecma-builtin-internal-routines-template.inc.h#L211

As I wrote even in the commit message: we should not try to return any fallback value after a failed assert. Assertion error means something disastrous, some design flaw in the program. It's not something we should even try to recover from. (If it's fixable then it's not an assertion.)

jerry-core/ecma/base/ecma-helpers.c
akosthekiss: Same as above. Don't try to do anything meaningful after a failed assert. Style guide seems to expect a `break` at the end of every switch case, even in `default` (at least that's how most of the code base looks like), so be it. But after the unreachable point, nothing should happen.

jerry-core/ecma/operations/ecma-objects-general.c
akosthekiss: Yes, above this line is an if/else if/else block with all branches returning. If we'd ever get here that would not be a program design flaw but a compiler error.

jerry-core/vm/vm.c
akosthekiss: Same as above.

jerry-core/ecma/base/ecma-property-hashmap.c
akosthekiss: Because this is a while(true) loop with `return`s the only way out. There is no unreachable assert after the next while(true) loop either.

jerry-core/api/jerry.c
```diff
@@ -100,11 +100,7 @@ static const char * const wrong_args_msg_p = "wrong type of argument";
 static inline void JERRY_ATTR_ALWAYS_INLINE
 jerry_assert_api_available (void)
 {
-  if (JERRY_UNLIKELY (!(JERRY_CONTEXT (status_flags) & ECMA_STATUS_API_AVAILABLE)))
-  {
-    /* Terminates the execution. */
-    JERRY_UNREACHABLE ();
-  }
+  JERRY_ASSERT (JERRY_CONTEXT (status_flags) & ECMA_STATUS_API_AVAILABLE);
```
LaszloLango: I think this is not equal to the old code. `JERRY_ASSERT` has no effect in release builds, but `JERRY_UNREACHABLE` has (with GNUC).

akosthekiss: Which is quite the point of the change. Why should there be an effect in release builds and in GCC only?

jerry-core/api/jerry.c
```diff
@@ -847,8 +840,6 @@ jerry_value_get_type (const jerry_value_t value) /**< input value to check */
       break;
     }
   }
-
-  return JERRY_TYPE_NONE;
```
zherczeg: It is strange the checkers does not complain (for the the UNREACHABLE is just a function call). We should move this up to the default case.

jerry-core/ecma/base/ecma-helpers-value.c
zherczeg: But still looks ugly. I would recommend merge it with one other option (e.g `case ECMA_TYPE_OBJECT` above), and check with an assertion that the type is really ECMA_TYPE_OBJECT. This technique has already used several places in the code base.

jerry-core/ecma/base/ecma-helpers.c
zherczeg: Merge it please just like above.

jerry-core/ecma/builtin-objects/ecma-builtin-internal-routines-template.inc.h
```diff
@@ -209,6 +209,7 @@ DISPATCH_ROUTINE_ROUTINE_NAME (uint16_t builtin_routine_id, /**< built-in wide r
     default:
     {
       JERRY_UNREACHABLE ();
+      break;
```
zherczeg: If we don't want to do anything meaningful, why do we add a break here?

jerry-core/ecma/operations/ecma-objects-general.c
zherczeg: I would remove the last `else` statement then, and simply do the return.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1809,10 +1809,10 @@ ecma_object_get_class_name (ecma_object_t *obj_p) /**< object */
         default:
         {
           JERRY_UNREACHABLE ();
+          break;
```
zherczeg: Merge this with "ECMA_PSEUDO_ARRAY_ARGUMENTS"

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1144,14 +1142,11 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
       }
       default:
       {
-        JERRY_TRACE_MSG ("UNKNOWN opcode (%u)!\n", (unsigned int) op);
-        return ecma_raise_common_error (ECMA_ERR_MSG ("Unknown RegExp opcode."));
+        JERRY_UNREACHABLE ();
+        break;
```
zherczeg: This could also be merged.

jerry-core/vm/vm.c
```diff
@@ -354,10 +328,30 @@ vm_construct_literal_object (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
...
 #endif /* !CONFIG_DISABLE_REGEXP_BUILTIN */
+
+  JERRY_ASSERT (bytecode_p->status_flags & CBC_CODE_FLAGS_FUNCTION);
+
+  ecma_object_t *func_obj_p;
+
+#ifndef CONFIG_DISABLE_ES2015_ARROW_FUNCTION
+  if (!(bytecode_p->status_flags & CBC_CODE_FLAGS_ARROW_FUNCTION))
+  {
+#endif /* !CONFIG_DISABLE_ES2015_ARROW_FUNCTION */
```
zherczeg: Can't re reorganize this with an `unlikely` to use the ifdef only once?

akosthekiss: An `unlikely` is a runtime condition even if the branch predictor gets hints from the developer. An `#if[n]def` guarding out that condition surely wont execute a code that's not even there.

akosthekiss:
@zherczeg I've followed most of your suggestions, i.e., unreachable default cases got merged into previous "real" cases with appropriate asserts and now there is no code anymore after any of the remaining JERRY_UNREACHABLEs.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-helpers.c
```diff
@@ -90,8 +90,7 @@ ecma_typedarray_helper_get_shift_size (uint8_t builtin_id) /**< the builtin id o
 #endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64 */
     default:
     {
-      JERRY_ASSERT (false);
-      return ECMA_BUILTIN_ID__COUNT;
+      JERRY_UNREACHABLE ();
```
LaszloLango: This might cause warnings with certain compilers, if the `JERRY_UNREACHABLE` defined as empty. What about the windows port? I think it should complain about the missing return statements.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1142,16 +1142,8 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
         }
         return ECMA_VALUE_FALSE; /* fail */
       }
-      default:
-      {
-        JERRY_TRACE_MSG ("UNKNOWN opcode (%u)!\n", (unsigned int) op);
-        return ecma_raise_common_error (ECMA_ERR_MSG ("Unknown RegExp opcode."));
```
LaszloLango: I think this isn't a good decision to change this runtime error to an assert.

akosthekiss: Please explain. This is the virtual machine of the regex engine executing the opcodes that can only come from "us", i.e., there is no legitimate way for a user to pass in arbitrary byte sequence. If we generate regex byte code propery and interpret it properly, too, then we should not possibly find an unknown opcode here ever. We are not preparing the "main" VM for unknown opcode either. If anything like that happens then we have a design flaw. If I'm not misunderstanding something.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-helpers.c
LaszloLango: Thanks for the explanation. It is good for me if all of the supported compilers deal with this.

jerry-core/ecma/operations/ecma-regexp-object.c
LaszloLango: Fine, good to me.

akosthekiss:
Thanks for the reviews. Let me squash the commits before landing the PR.

akosthekiss:
PR squashed and rebased

PR #2346, opened 2018-05-19
Don't let doctests pollute the source tree

akosthekiss:
The patch changes the cmakelists of the doctests to extract the
test sources from the markdown docs into the binary tree.


akosthekiss:
It's just a recent flakyness of the unit tests job. I've restarted it and will look into the underlying flakyness problem.

yichoi:
LGTM

PR #2348, opened 2018-05-20
Fixes Visual C/C++ warnings.

lygstate:


jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -573,7 +573,7 @@ re_match_regexp (re_matcher_ctx_t *re_ctx_p, /**< RegExp matcher context */
           {
             JERRY_ASSERT (ecma_is_value_boolean (match_value));
             /* restore saved */
-            memcpy (re_ctx_p->saved_p, saved_bck_p, size);
+            memcpy ((void *) (re_ctx_p->saved_p), saved_bck_p, size);
```
akosthekiss: I'm a bit worried that this patch would only hide things. `re_ctx_p` is a pointer to `struct re_matcher_ctx_t`, and `saved_p` is `const lit_utf8_byte_t **saved_p`, i.e., a (non-const) pointer to a (non-const) pointer to a `const lit_utf8_byte_t`. If anything, the first parameter should be cast to `(lit_utf8_byte_t **) re_ctx_p->saved_p` but actually I'm not even sure why MSVC warns about this: the pointers in the array should not be treated as constant (if I'm not mistaken).

jerry-core/jmem/jmem.h
```diff
@@ -199,7 +199,7 @@ void jmem_run_free_unused_memory_callbacks (jmem_free_unused_memory_severity_t s
   { \
     JERRY_ASSERT (var_name ## ___size != 0); \
     \
-    jmem_heap_free_block (var_name, var_name ## ___size); \
+    jmem_heap_free_block ((void *) (var_name), var_name ## ___size); \
```
akosthekiss: This is something that certainly should _not_ be changed. Same comments hold as for the change above. If anything is to be changed then it's line 1261 of ecma-regexp-object.c, where `saved_p` is declared as a local array of `const lit_utf8_byte_t *` elements. (MSVC reports the warning for line 1428, but 1261 is the root cause.) Again, I'm not sure why MSVC considers this a problem.

akosthekiss:
I've been thinking a bit more about this. I don't claim to be the definitive reference for the C language but these two last const-related warnings seem suspicious to me. If someone would confirm that MSVC is treating `const**` in a non-standard way then I'd suggest not to change the C code base but to add the suppression of the C4090 warnings to the build system.

zherczeg:
What about reporting this issue to microsoft?

zherczeg:
This bug is closed because of inactivity. Please reopen it if you fix the issues above.

PR #2349, opened 2018-05-21
Consolidate compiler-specific parts of CMakeLists

akosthekiss:
The cmakelists contained various compiler-specific parts scattered
across the file. This patch moves them to as few conditional blocks
as possible.

Additional changes:
- `-Werror` does not depend on whether jerry-libc is enabled.
- Some stylistic fixes.


yichoi:
LGTM

PR #2351, opened 2018-05-22
Buildable as shared libs

wille-io:
All libraries were built as STATIC regardless of ENABLE_STATIC_LINK turned OFF.


wille-io:
Also LTO does not happen on shared libs. cmake option should be set OFF with a warning maybe?

LaszloLango:
The current PR commits contains a lot of unrelated files, like cmake cache files, outputs, compiled binaries, etc. Please remove them.

wille-io:
Sorry. Visual Studio Code messed up.

jerry-core/CMakeLists.txt
```diff
@@ -277,7 +277,7 @@ endif()
 math(EXPR MEM_HEAP_AREA_SIZE "${MEM_HEAP_SIZE_KB} * 1024")
 set(DEFINES_JERRY ${DEFINES_JERRY} CONFIG_MEM_HEAP_AREA_SIZE=${MEM_HEAP_AREA_SIZE})
 
-add_library(${JERRY_CORE_NAME} STATIC ${SOURCE_CORE_FILES})
+add_library(${JERRY_CORE_NAME} ${SOURCE_CORE_FILES})
```
zherczeg: Is this make this always dynamic?

akosthekiss:
The macOS part was tricky (at least for me, for the first time): the linker on macOS does not work exactly as on Linux, it doesn't really like undefined symbols in shared libs (and jerry-core uses symbols from jerry-port-default, which is another library and not directly linked to jerry-core). That's the root of the problem in a nutshell, more about it can be read at https://stackoverflow.com/questions/3695234/what-is-the-deal-with-undefined-symbols-in-a-shared-library-or-dylib .

I've a commit on top of the head of this PR at https://github.com/akosthekiss/jerryscript/commit/df9c0783492462dc494ec8bf55518e34e62d013b or https://github.com/akosthekiss/jerryscript/tree/pr2351 . It fixes the macOS issue, makes additional changes to the build system, and reverts changes that should not really happen. Please, use it as a review and/or just cherry-pick it on top of the PR commits.

PS: Don't forget to squash all the commits of the PR at the end. You can keep my signoff line in the commit message above yours.

akosthekiss:
@wille-io Please _do_ squash all these commits into a single one and rebase it on top of latest master. (The project doesn't use non-linear merges.)

akosthekiss:
@wille-io is this happening? On irc, I've sent you some steps to be followed, copied here:

I'd suggest doing the squashing and the rebasing in two steps
1) first squash:
    1) `git rebase -i 9eb9fbf5f12fba7511481d0cabb25a2020d7b710`
    2) do whatever rebase is asking of you: keep (pick) the topmost commit and mark all the rest to be squashed into that, then edit the commit message and make it read as one meaningful text. keep the signoff(s) at the end
2) then rebase:
    1) `git rebase origin/master`
3) finally force push your branch to github to update it:
    1) `git push -f`
    2) note: it is generally better to work on a branch than on master. it's easier to keep different topics of work separated
    3) if you get into any trouble while rebasing and/or squashing, you can always do a `git rebase --abort` and get back to where you started

PR #2353, opened 2018-05-22
WIP: Add automatic code size measurements to PRs

akosthekiss:
The patch makes use of the commenting capabilities and the recent
Ubuntu support of AppVeyor.

[skip ci]

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,25 +6,78 @@ branches:
   - coverity_scan
   - gh_pages
 skip_tags: true
+
+# Build matrix setup: various Visual Studio jobs and one Ubuntu job.
 image:
   - Visual Studio 2017
+  - Ubuntu
 configuration:
-  - Debug
+  # - Debug
   - Release
 platform:
   - x64
-  - Win32
-init:
-  - cmd: |
-      cmake -version
-before_build:
-  - cmd: |
-      mkdir build
-      cd build
-      if "%PLATFORM%"=="Win32" cmake -G"Visual Studio 15 2017" ..
-      if "%PLATFORM%"=="x64" cmake -G"Visual Studio 15 2017 Win64" ..
+  # - Win32
 
-build:
-  project: build\Jerry.sln
-  parallel: true
-  verbosity: minimal
+matrix:
+  exclude:
+    - image: Ubuntu
+      configuration: Debug
+    - image: Ubuntu
+      platform: Win32
+
+for:
+  # Steps of a Visual Studio job.
+  - matrix:
+      except:
+        - image: Ubuntu
+    init:
+      - cmake -version
+    before_build:
+      - if "%PLATFORM%"=="Win32" cmake -G"Visual Studio 15 2017" -Bbuild -H.
+      - if "%PLATFORM%"=="x64" cmake -G"Visual Studio 15 2017 Win64" -Bbuild -H.
+    build:
+      project: build\Jerry.sln
+      parallel: true
+      verbosity: minimal
+
+  # Steps for the Ubuntu job.
+  # Running this job is on an opt-in basis. The commit message has to contain
+  # the "[run measurements]" string somewhere to make it run.
+  - matrix:
+      only:
+        - image: Ubuntu
+    init:
+      - |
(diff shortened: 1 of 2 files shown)


rerobika:
First of all it's a great patch, seems really useful.

Secondly, it would be nice to see the binary size differences directly without any calculation.

My suggestion is:

| jerry variant	| master @ 9eb9fbf (text+data) | 88bfb1f @ 88bfb1f (text+data)	|
| ------ | ------ | ---- |
| arm-linux-minimal	| 76835+0 |77000+12  (increase 165 B + increase 12 B)|
| arm-linux-es5.1 |	138055+16 |	138000+0 (decrease 55 B + decrease 16 B)|
| arm-linux-es2015-subset | 150848+16 | 150848+16 |

LaszloLango:
Why AppVeyor? Any practical reason? I think we can do the same with Travis CI.

zherczeg:
This tool seems spamming comments. Could we make it to just display the last measurement somewhere? E.g. in the green box at the bottom with the "squash and merge" button?

akosthekiss:
@LaszloLango AFAIK this cannot be solved with Travis CI. Here is my understanding of the topic: you can comment on a GitHub PR via REST API but only if you have access credentials. You can create access tokens on your GitHub profile, encrypt it for Travis CI, and add the encrypted token as an environment variable either via the yml file or via the web interface of Travis CI -- BUT: pull requests don't get the encrypted values. Which is quite reasonable as access tokens are like passwords and a malicious pull request could just print the access token to the logs if it had access to them.

So, branch builds get access to secure tokens but then there is no PR to comment on, while PR builds get no access to secure tokens that would allow commenting. (I'm not even listing the opening of a sec hole as a third option.) And this is the situation both for Travis CI and Circle CI. This would also be true for AppVeyor had it no notification possibility that enables commenting on pull requests. So, all the above is handled internally by the service that has already been authenticated and we only have to compose the comment text. If I read all the docs properly, AppVeyor is the only CI service that has this option as of today.

@zherczeg Unfortunately, adding things to the "checks" part of the page is even harder if not impossible. It requires authentication (same issues as above) and may not even allow as much or as formatted text as the comments.

zherczeg:
Whatever is the last variant, it looked ugly in an automatic mail. Anyway it would be good if no mail would be sent when the bot adds its comment, and remove its old comments. Will be very hard to go through on the comments of a long patch.

akosthekiss:
@zherczeg Please specify what looked ugly in the mail, and in which email client. And whether GitHub sends email about the comment of the bot or not is not under the control of the bot itself. It's controlled by the per user settings in GH. But I don't think that it's possible to disable mail notifications about a single user's comments.

akosthekiss:
After some tweaking and twisting, I could make the commenting of the AVbot conditional. If the commit message of the topmost commit contains `[run measurements]` then the size measurements will get executed and the bot will post a comment with the results but if that string is missing then the measurement CI job will not even run and the bot will not comment.

akosthekiss:
@LaszloLango Ah, I though it would cause the least disruption in the comment stream if the bot's messages are small but I can simply remove the sup tags from around the table if normal font size is requested.

LaszloLango:
I vote for the normal font size.

akosthekiss:
So, the concept has been proven to work: AV can be used to provide automated feedback on the size effect of PRs. But just because something can be done doesn't necessarily mean that it must/should be done. Is anyone in favor of this PR, or against it? Any further needs or ideas for further development?

zherczeg:
Personally I really don't like the comment based approach. I would not like the CI if it would do things this way.

Suggestion: the logs are saved by the bots, so we can check them if we want to know the change. Lets just add a new bot which task is producing these results. No other output.

akosthekiss:
@zherczeg Just double checking: do you mean that the CI shouldn't comment even if explicitly requested? (With the current approach, the CI is not commenting anything by default. It's only commenting if the `[run measurements]` text is in the commit message.)

LaszloLango:
My concern about this approach is that we will forget to remove the comment (`[run measurements]`) before merge and it will spam the commits on the master. Printing the measurements only into the the CI logs is enough for me too.

akosthekiss:
It will do _what_? The GitHubPullRequest notification provider will not really have anywhere to comment in case of a branch build.

Anyway, I'm kinda feeling the pushback. Closing.

PR #2355, opened 2018-05-23
Fix issues revealed by cppcheck 1.83

akosthekiss:
A more recent cppcheck has revealed some extra issues not detected
by the old one running on CI.

### DIFF
--- a/jerry-main/cli.c
+++ b/jerry-main/cli.c
@@ -362,7 +362,7 @@ cli_print_help (const char *help) /**< the help message to print */
     cli_print_prefix (help, length);
 
     help += length;
-    while (*help == ' ' && *help != 0)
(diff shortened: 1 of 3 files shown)


jerry-main/cli.c
```diff
@@ -362,7 +362,7 @@ cli_print_help (const char *help) /**< the help message to print */
     cli_print_prefix (help, length);
 
     help += length;
-    while (*help == ' ' && *help != 0)
+    while (*help == ' ')
```
zherczeg: If `*help == ' '` then it cannot be 0, since space is 32, not 0.

akosthekiss: Yes, that's what was spotted by cppcheck and is fixed now.

PR #2362, opened 2018-05-28
Maintenance and merging of Valgrind and Valgrind-Freya code paths

akosthekiss:
None of the code paths have been tested for long and especially the
Freya code paths have been long abandoned. This patch:
- Merges Valgrind-Freya into Valgrind code path (there should be no
  need to choose between them, Valgrind should work just fine).
- Removes leftover code (`VALGRIND_FREYA_CHECK_MEMPOOL_REQUEST` and
  `valgrind_freya_mempool_request`).
- Adds `JMEM_` prefix to Valgrind-related macros (to correctly
  leave the `VALGRIND_` prefix to Valgrind).
- Moves the definition of the Valgrind-related macros to a common
  header to avoid duplication.

Note: Adding a CI job to perform Valgrind Memchecks is left for
follow-up as it turns out to be excessively slow (>50 mins for a
`--jerry-tests --jerry-test-suite` run, and even a simple
`--jerry-tests` may get terminated prematurely because of timeout
issues).


yichoi:
LGTM

PR #2363, opened 2018-05-29
Reverse the generation of line info.

zherczeg:


tests/unit-core/test-backtrace.c
```diff
@@ -198,5 +198,46 @@ main (void)
 
   jerry_cleanup ();
 
+  jerry_init (JERRY_INIT_EMPTY);
```
galpeter: Would it be possible to move this new test to it's own function? That way it is clear that it is a single test case.

yichoi:
LGTM

PR #2370, opened 2018-06-01
Add better support for proper installation after build

akosthekiss:
- Add `--install` option to `tools/build.py`.
- Make use of `--install` in `tools/run-tests.py` by testing the
  installed the executables instead of those in the build tree.

Related changes:
- Collect unit test binaries in the `tests` subdir of the build
  tree instead of `bin`.
- The `ls`-based collection of the unit test binaries had some
  shortcomings hitherto unrevealed (it didn't filter for files so
  it could potentially "collect" dictionaries, too), which has now
  been replaced with a more stable `find`-based solution.


tests/unit-ext/module/CMakeLists.txt
```diff
@@ -24,5 +24,6 @@ endif()
 
 add_executable(${JERRYX_MODULE_UNITTEST_NAME} ${JERRYX_MODULE_UNIT_TEST_SOURCES})
 set_property(TARGET ${JERRYX_MODULE_UNITTEST_NAME} PROPERTY LINK_FLAGS "${LINKER_FLAGS_COMMON}")
+set_property(TARGET ${TARGET_NAME} PROPERTY RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/tests")
```
robertsipka: There is no `TARGET_NAME` defined here.

PR #2371, opened 2018-06-01
Use binary mode in snapshot tool for fopen

galpeter:
In the snapshot tool the files were opened
in text mode. However the snapshot files
are binary files thus it is advised to use the
binary mode when opening the files.

Specifying the binary mode is a must on Windows
platform otherwise the read/write operations
are inserting extra '\r' characters.

Also update jerry-libc to accept the 'b' modifier.

### DIFF
--- a/jerry-main/main-unix-snapshot.c
+++ b/jerry-main/main-unix-snapshot.c
@@ -92,7 +92,7 @@ static size_t
 read_file (uint8_t *input_pos_p, /**< next position in the input buffer */
            const char *file_name) /**< file name */
 {
-  FILE *file = fopen (file_name, "r");
+  FILE *file = fopen (file_name, "rb");
 
   if (file == NULL)
   {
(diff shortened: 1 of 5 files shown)


jerry-main/main-unix-snapshot.c
```diff
@@ -92,7 +92,7 @@ static size_t
 read_file (uint8_t *input_pos_p, /**< next position in the input buffer */
            const char *file_name) /**< file name */
 {
-  FILE *file = fopen (file_name, "r");
+  FILE *file = fopen (file_name, "rb");
```
akosthekiss: The same comment goes here as for main-unix.c and main-unix-test.c: the caller of `read_file` should decide whether its a text or binary mode open.

jerry-main/main-unix-snapshot.c
```diff
@@ -357,7 +357,7 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
       return JERRY_STANDALONE_EXIT_CODE_FAIL;
     }
 
-    FILE *literal_file_p = fopen (save_literals_file_name_p, "w");
+    FILE *literal_file_p = fopen (save_literals_file_name_p, "wb");
```
akosthekiss: Isn't this a text file?

galpeter: Yes, but I want to preserve the line endings so they are the same on Windows and Linux.

jerry-main/main-unix-snapshot.c
galpeter: The reason for always reading a binary file is to avoid any line ending conversion on Windows, even when we are reading text files.

galpeter:
Created a PR to improve the license checker: https://github.com/jerryscript-project/jerryscript/pull/2391

galpeter:
@akosthekiss I've updated the PR, the CRLF problems are solved.

PR #2373, opened 2018-06-02
Rework command line handling of build tool

akosthekiss:
In `tools/build.py`:
- For the sake of readability, group CLI arguments as general build
  options, options to control the building of components, and
  component-specific options.
- To prevent duplications, remove the defaults from those CLI
  arguments that correspond to CMake options and have defaults in
  any of the CMakeLists. Should any of the defaults change, they
  will have to be changed at a single place only. (Those options
  that are not set on the command line of `tools/build.py` are not
  passed as options to `cmake` either.)
- Convert `--unittests` and `--doctests` to ON/OFF options like the
  rest of the component switches.
- Touch on some of the help messages of the CLI arguments.

Other changes:
- The change in `--unittests` and `--doctests` is a slightly CLI-
  breaking change of `tools/build.py`. Thus, follow up on this in
  `tools/run-tests.py`.
- Move `ENABLE_ALL_IN_ONE` into `jerry-core` as it is not a general
  option but specific to that component.
- Remove the forcing of `ENABLE_ALL_IN_ONE` for some compilers/
  platforms as it is still an option, not a hard requirement.


akosthekiss:
Note (to self/maintainers): this will conflict with #2370. Whichever PR gets landed first, the other will have to be rebased and conflicts resolved.

yichoi:
@akosthekiss rebase please

tools/pylint/pylintrc
```diff
@@ -319,7 +319,7 @@ max-returns=6
 max-branches=15
 
 # Maximum number of statements in function / method body
-max-statements=50
+max-statements=75
```
LaszloLango: :)

PR #2377, opened 2018-06-02
Direct alloc/dealloc implementations for ecma_number_t

akosthekiss:
As a result of earlier developments, there remained no other
instatiations of the `ALLOC`, `DEALLOC`, and `DECLARE_ROUTINES_FOR`
templates. So, it's simpler to write the alloc/dealloc routines
directly for `ecma_number_t` as well.


akosthekiss:
Note: The proposed change gives equivalent code to the currently existing implementation. I've tried adding the `inline` and `JERRY_ATTR_ALWAYS_INLINE` modifiers to the functions the same way they are used at the other allocs/deallocs, however, that resulted in no speed improvement but increased the binary size 133440 -> 133752.

PR #2378, opened 2018-06-03
targets: zephyr: Separate CMake definitions from options.

pfalcon:
Add a space between definition and options substitution groups, otherwise
they make expand to something like -DFOO-DBAR.


pfalcon:
Without this, building the latest master fails with:

~~~
<command-line>:0:19: error: ISO C99 requires whitespace after the macro name [-Werror]
~~~

pfalcon:
Thanks for quick turnaround!

PR #2381, opened 2018-06-05
Improve SonarQube analysis to cover more code path.

LaszloLango:

### DIFF
--- a/tools/check-sonarqube.sh
+++ b/tools/check-sonarqube.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if [[ "${TRAVIS_REPO_SLUG}" == "jerryscript-project/jerryscript"
+  && ${TRAVIS_BRANCH} == "master"
+  && ${TRAVIS_EVENT_TYPE} == "push" ]]
+then
+  git fetch --unshallow
+  build-wrapper-linux-x86-64 --out-dir bw-output \
+    ./tools/build.py --error-messages=on \
+                     --jerry-cmdline-snapshot=on \
+                     --jerry-debugger=on \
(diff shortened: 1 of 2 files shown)


tools/check-sonarqube.sh
```diff
@@ -0,0 +1,49 @@
...
+                 --jerry-cmdline-snapshot=on \
+                 --jerry-debugger=on \
+                 --jerry-libc=off \
+                 --line-info=on \
+                 --mem-stats=on \
+                 --profile=es2015-subset \
+                 --snapshot-save=on \
+                 --snapshot-exec=on \
+                 --valgrind=on \
+                 --vm-exec-stop=on
```
akosthekiss: style: lines seem to be two spaces off (if the style in the then branch is to be followed)

tools/check-sonarqube.sh
```diff
@@ -0,0 +1,49 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+if [[ "${TRAVIS_REPO_SLUG}" == "jerryscript-project/jerryscript"
+  && ${TRAVIS_BRANCH} == "master"
+  && ${TRAVIS_EVENT_TYPE} == "push" ]]
+then
+  git fetch --unshallow
```
akosthekiss: Is this really necessary? I've found only one mention of `git fetch --unshallow` in SonarQube docs and that was in the description of the [Branch plugin](https://docs.sonarqube.org/display/PLUG/Branch+Plugin). Are we using that plugin or do we need this for some other reason?

LaszloLango: Yes it is. It was not included in the first patch, but I had to add it in a followup.

LaszloLango: See: #2325
There was warning in the analysis log.
>WARN: Shallow clone detected, no blame information will be provided. You can convert to non-shallow with 'git fetch --unshallow'.

Full log: https://travis-ci.org/jerryscript-project/jerryscript/jobs/377608402

LaszloLango:
@akosthekiss thanks for the review. I've updated the PR.

Note: I also modified to use the short hash for version number.

akosthekiss:
@LaszloLango Why did you truncate the hash? Although 7 hex digits are still highly unlikely to collide but a full git hash is even more unlikely to ever yield a collision.

LaszloLango:
@akosthekiss I prefer the short hash, because I think it looks nicer on the sonarcloud.io page. It is only my personal preference, so I can change it back if I am the only one who prefer it. :)

akosthekiss:
I vote for the use of full git hashes for the sake of unambiguous identification.

LaszloLango:
@akosthekiss looks like others are not interested in this conversation, so I changed back to full git hashes. :)

PR #2385, opened 2018-06-06
Add operational mode for jerry_gc API call.

zherczeg:


docs/02.API-REFERENCE.md
```diff
@@ -705,9 +717,11 @@ Performs garbage collection.
 
 ```c
 void
-jerry_gc (void);
+jerry_gc (jerry_gc_mode_t mode);
 ```
 
+- `mode` - operational mode, see [jerry_gc_mode_t](#jerry_gc_mode_t)
+
 **Example**
```
LaszloLango: Please update the example as well.

docs/02.API-REFERENCE.md
```diff
@@ -68,6 +68,18 @@ Option bits for [jerry_parse](#jerry_parse) and
  - JERRY_PARSE_NO_OPTS - no options passed
  - JERRY_PARSE_STRICT_MODE - enable strict mode
 
+## jerry_gc_mode_t
+
+Set garbage collection operational mode
```
akosthekiss: I'm no native English speaker but isn't this "operation mode" or "operating mode"?

docs/02.API-REFERENCE.md
```diff
@@ -705,13 +717,31 @@ Performs garbage collection.
...
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t object_value = jerry_create_object ();
+  jerry_release_value (object_value);
+
+  jerry_gc (JERRY_GC_SEVERITY_LOW);
+
+  jerry_cleanup ();
+}
+
```
akosthekiss: Extra empty line.

jerry-ext/handler/handler-gc.c
```diff
@@ -31,6 +31,13 @@ jerryx_handler_gc (const jerry_value_t func_obj_val, /**< function object */
   (void) args_p; /* unused */
   (void) args_cnt; /* unused */
 
-  jerry_gc ();
+  jerry_gc_mode_t mode = JERRY_GC_SEVERITY_LOW;
+
+  if (args_cnt > 0 && jerry_value_to_boolean (args_p[0]))
```
akosthekiss: Because of this change, neither `args_cnt` nor `args_p` are unused anymore. (See above.)

jerry-ext/handler/handler-gc.c
```diff
@@ -31,6 +31,13 @@ jerryx_handler_gc (const jerry_value_t func_obj_val, /**< function object */
...
+  {
+    mode = JERRY_GC_SEVERITY_HIGH;
+  }
```
akosthekiss: This all could go into a `?:` expression to avoid making the initialization of `mode` potentially a dead expression.

```c
  jerry_gc_mode_t mode = args_cnt > 0 && jerry_value_to_boolean (args_p[0]) ? JERRY_GC_SEVERITY_HIGH
                                                                            : JERRY_GC_SEVERITY_LOW;
```

akosthekiss: BTW, the doc string of the function should be updated describing how it deals with its arguments. Something like:

The routine invokes the GC with high severity if it has at least one argument and the first argument is equivalent to boolean true; it calls the GC with low severity otherwise.

docs/02.API-REFERENCE.md
zherczeg: According to google, `operation mode` is used 262M sites, while `operational mode` 117M sites. Both are used by high profile (e.g. microsoft or google technical docs) sites. I think both are ok, so I don't want to change unless you insist on it.

zherczeg: I checked Cambridge dictionary:

operation - noun
operational - adjective
operating - present participle

I would prefer `operational` here.

zherczeg: Google: The present participle is a participle that ends in ing. It can be used with the auxilliary verb 'to be' to form the continuous tense. It always takes the 'ing' form of the verb, even irregular verbs have an '...ing' form, in fact virtually all English words that end with 'ing' are present participles.

I think `operating mode` is also correct, if we emphasize how garbage collection is performed. The `operational mode` simply emphasize the behavior.

zherczeg:
Patch updated.

docs/02.API-REFERENCE.md
zherczeg: Done. It runs with doctest as well now.

docs/02.API-REFERENCE.md
zherczeg: E.g:
operating system - a system which is running (emphasizing continuity)
operational system - a system which is capable of running (not relevant if it is doing it now)
operation system - double noun, usually a phrase or not used at all. I think the latter is true here.

dbatyai: As you've said, saying something is 'operational' means it is capable of running. Saying that the mode is operational means that the thing that is running will be able to run in that mode, but also implies that a 'not operational' mode exists, which 'won't be able to run'. This is not the case, as the mode here means a specific a way of operation, and not whether the GC can actually run, or is running.

I'm not sure what behavior you'd like to emphasize by using `operational mode`, yet it feels instinctively wrong to me, and I'd say the suggestions made by @akosthekiss would be the the correct way to go.

zherczeg: I didn't exactly say that. Tricky way of turning my words :)

Adjectives extends the meaning of words, while nouns do not. Think about the difference of "wooden house" and "tree house". The former uses an adjective to extend a noun, while the latter is simply a phrase. 

I found "operational mode" in a well known English dictionary:
https://en.oxforddictionaries.com/definition/standby
An operational mode of an electrical appliance in which the power is switched on but the appliance is not actually functioning.

I could not find the same thing with "operation mode".

zherczeg:
Any more comments for this patch?

docs/02.API-REFERENCE.md
akosthekiss: I've consulted an English teacher and operation was preferred over operational. But still no native speaker chiming in on either side. As this is no API function name but a doc text only, I wont be pushing further.

PR #2391, opened 2018-06-08
Improve license checker line ending validation

galpeter:
The license checker previously assumed that the
lines of the license will always end with \n
characters. However when checking a file
it could happen that other line endings are
returned (should only happen for test files) thus
the checker can incorrectly report invalid license
as the line endings are incorrect.

Additional note #1: in Python when reading a file
in text mode it can happen that the line endings are
converted to the host system's line ending.
However on Travis the conversion did not happen when
using the open built-in method. By switching to the
io.open call the conversion is enforced and
all line endings are converted to '\n' regardless of
the host system's line ending.

Additional note #2: it is possible that there
are input test files which are not utf-8 conformant
(eg.: to test the parser). These files can't be read
as utf-8 strings and an exception would occur.
By ignoring these errors the tool can check
the file's license. In the license text there is no
invalid utf-8 character so the check will work
correctly.

### DIFF
--- a/tools/check-license.py
+++ b/tools/check-license.py
@@ -16,6 +16,7 @@
 
 from __future__ import print_function
 
+import io
 import os
 import re
 import sys
@@ -78,7 +79,7 @@ def main():
             for fname in files:
                 if any(fname.endswith(ext) for ext in EXTENSIONS):
                     fpath = os.path.join(root, fname)
-                    with open(fpath) as curr_file:
+                    with io.open(fpath, 'r', errors='ignore') as curr_file:
                         if not LICENSE.search(curr_file.read()):
                             print('%s: incorrect license' % fpath)
                             is_ok = False


tools/check-license.py
```diff
@@ -78,7 +79,7 @@ def main():
             for fname in files:
                 if any(fname.endswith(ext) for ext in EXTENSIONS):
                     fpath = os.path.join(root, fname)
-                    with open(fpath) as curr_file:
+                    with io.open(fpath, 'r', errors='ignore', newline='') as curr_file:
```
akosthekiss: Why not simply use universal newlines (`newline=None`, or just don't mention `newline` at all)? Then python will see all newlines as `\n` and the `LICENSE` regex can be left unchanged.

(BTW, wouldn't that be simply equivalent to `open(fpath, 'rU')`, without need for `io`?)

galpeter: The problem was that Travis (seemingly) did not use the universal newlines that's why there was an license error reported in #2371 , on my system the conversion is performed and there is no error reported. For the `rU` I got a deprecation warning with Python3.

akosthekiss: Could you please double check? Here are my experiments (on top of this PR's commit):
- https://github.com/akosthekiss/jerryscript/commit/180980eb433a2c3d677f38a9accf565dfaaf24c6
- https://travis-ci.org/akosthekiss/jerryscript/builds/390173690

galpeter:
@akosthekiss did a triple-check and it worked as you suggested (removed the newline parameter). Guess that at a previous check there was some incorrect modifications on my side. Thanks!

I've updated the PR.

PR #2392, opened 2018-06-08
 Rework 128 bit arithmetics of ecma_utf8_string_to_number

akosthekiss:
* Rewritten 128-bit integer "type" to use two 64-bit ints (they are enough), and made it a proper struct instead of an array.
* Rewritten all single-bit shift loops to multi-bit shifts with the help of CLZ, and used `__builtin_clzll` where available.
* Simplified (and documented) 128-bit DIV10 operation.
* Renamed 128-bit integer handling macros to use simpler names. (And removed unused macros that were laying around.)

This is the result of a longer experiment and rewrites part of `ecma_utf8_string_to_number` quite heavily. Thus, for now, the commits are not squashed, allowing reviewers to take a closer look at each rewriting step separately. Should the PR be accepted, all this will be squashed, of course.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-conversion.c
+++ b/jerry-core/ecma/base/ecma-helpers-conversion.c
@@ -26,308 +26,220 @@
  *
  * \addtogroup ecmahelpers Helpers for operations with ECMA data types
  * @{
- *
+ */
+
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT64
+
+/**
  * \addtogroup ecmahelpersbigintegers Helpers for operations intermediate 128-bit integers
  * @{
  */
 
 /**
- * Check that parts of 128-bit integer are 32-bit.
+ * 128-bit integer type
  */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_CHECK_PARTS_ARE_32BIT(name) \
-{ \
-  JERRY_ASSERT (name[0] <= UINT32_MAX); \
-  JERRY_ASSERT (name[1] <= UINT32_MAX); \
-  JERRY_ASSERT (name[2] <= UINT32_MAX); \
-  JERRY_ASSERT (name[3] <= UINT32_MAX); \
-}
+typedef struct
+{
+  uint64_t lo; /**< low 64 bits */
+  uint64_t hi; /**< high 64 bits */
+} ecma_uint128_t;
 
 /**
- * Declare 128-bit integer.
+ * Round high part of 128-bit integer to uint64_t
  */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER(name) uint64_t name[4] = { 0, 0, 0, 0 }
-
-/**
- * Declare 128-bit in-out argument integer.
- */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_ARG(name) uint64_t name[4]
-
-/**
- * Initialize 128-bit integer with given 32-bit parts
- */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_INIT(name, high, mid_high, mid_low, low) \
-{ \
-  name[3] = high; \
-  name[2] = mid_high; \
-  name[1] = mid_low; \
-  name[0] = low; \
- \
-  ECMA_NUMBER_CONVERSION_128BIT_INTEGER_CHECK_PARTS_ARE_32BIT (name); \
-}
-
-/**
- * Copy specified 128-bit integer
- */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_COPY(name_copy_to, name_copy_from) \
-{ \
-  ECMA_NUMBER_CONVERSION_128BIT_INTEGER_CHECK_PARTS_ARE_32BIT (name_copy_to); \
-  ECMA_NUMBER_CONVERSION_128BIT_INTEGER_CHECK_PARTS_ARE_32BIT (name_copy_from); \
-  \
-  name_copy_to[0] = name_copy_from[0]; \
-  name_copy_to[1] = name_copy_from[1]; \
-  name_copy_to[2] = name_copy_from[2]; \
-  name_copy_to[3] = name_copy_from[3]; \
-}
-
-/**
- * Copy high and middle parts of 128-bit integer to specified uint64_t variable
- */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_ROUND_HIGH_AND_MIDDLE_TO_UINT64(name, uint64_var) \
-{ \
-  ECMA_NUMBER_CONVERSION_128BIT_INTEGER_CHECK_PARTS_ARE_32BIT (name); \
-  uint64_var = ((name[3] << 32u) | (name[2])) + (((name[1] >> 31u) != 0 ? 1 : 0)); \
-}
-
-/**
- * Copy middle and low parts of 128-bit integer to specified uint64_t variable
- */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_ROUND_MIDDLE_AND_LOW_TO_UINT64(name, uint64_var) \
-{ \
-  ECMA_NUMBER_CONVERSION_128BIT_INTEGER_CHECK_PARTS_ARE_32BIT (name); \
-  uint64_var = (name[1] << 32u) | (name[0]); \
-}
-
-/**
- * Check if specified 128-bit integers are equal
- */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_ARE_EQUAL(name1, name2) \
-  ((name1)[0] == (name2[0]) \
-   && (name1)[1] == (name2[1]) \
-   && (name1)[2] == (name2[2]) \
-   && (name1)[3] == (name2[3]))
-
-/**
- * Check if bits [lowest_bit, 128) are zero
- */
-#define ECMA_NUMBER_CONVERSION_128BIT_INTEGER_IS_HIGH_BIT_MASK_ZERO(name, lowest_bit) \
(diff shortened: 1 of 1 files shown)


akosthekiss:
Here are the results:

Benchmark | Heap (bytes) | Stack (bytes) | RSS (bytes) | Perf (sec) |
----: | ----: | ----: | ----: | ----: | 
3d-cube.js | 16376 -> 16376 : 0.000% | 15456 -> 15456 : 0.000% | 53248 -> 53248 : 0.000% | 0.806 -> 0.808 : -0.225% | 
3d-raytrace.js | 121600 -> 121600 : 0.000% | 2784 -> 2784 : 0.000% | 106496 -> 106496 : 0.000% | 1.031 -> 1.049 : -1.688% | 
access-binary-trees.js | 32752 -> 32752 : 0.000% | 3224 -> 3224 : 0.000% | 49152 -> 49152 : 0.000% | 0.571 -> 0.568 : +0.498% | 
access-fannkuch.js | 4968 -> 4968 : 0.000% | 1504 -> 1340 : +10.904% | 16384 -> 20480 : -25.000% | 2.092 -> 2.089 : +0.113% | 
access-nbody.js | 6248 -> 6248 : 0.000% | 1480 -> 1408 : +4.865% | 24576 -> 24576 : 0.000% | 1.079 -> 1.098 : -1.778% | 
bitops-3bit-bits-in-byte.js | 704 -> 704 : 0.000% | 1448 -> 1256 : +13.260% | 12288 -> 12288 : 0.000% | 0.512 -> 0.514 : -0.263% | 
bitops-bits-in-byte.js | 656 -> 656 : 0.000% | 1448 -> 1256 : +13.260% | 12288 -> 12288 : 0.000% | 0.666 -> 0.669 : -0.505% | 
bitops-bitwise-and.js | 496 -> 496 : 0.000% | 1224 -> 1032 : +15.686% | 12288 -> 12288 : 0.000% | 0.922 -> 0.925 : -0.285% | 
bitops-nsieve-bits.js | 98304 -> 98304 : 0.000% | 1448 -> 1256 : +13.260% | 110592 -> 110592 : 0.000% | 1.127 -> 1.130 : -0.239% | 
controlflow-recursive.js | 888 -> 888 : 0.000% | 67752 -> 67752 : 0.000% | 77824 -> 81920 : -5.263% | 0.371 -> 0.372 : -0.039% | 
crypto-aes.js | 33056 -> 33056 : 0.000% | 2148 -> 2148 : 0.000% | 61440 -> 65536 : -6.667% | 0.881 -> 0.875 : +0.622% | 
crypto-md5.js | 106488 -> 106488 : 0.000% | 1892 -> 1892 : 0.000% | 126976 -> 131072 : -3.226% | 0.597 -> 0.605 : -1.244% | 
crypto-sha1.js | 65528 -> 65528 : 0.000% | 1568 -> 1376 : +12.245% | 81920 -> 81920 : 0.000% | 0.589 -> 0.596 : -1.121% | 
date-format-tofte.js | 16376 -> 16376 : 0.000% | 1928 -> 1736 : +9.959% | 36864 -> 36864 : 0.000% | 0.702 -> 0.707 : -0.763% | 
date-format-xparb.js | 16384 -> 16384 : 0.000% | 2392 -> 2200 : +8.027% | 36864 -> 40960 : -11.111% | 0.407 -> 0.404 : +0.773% | 
math-cordic.js | 1776 -> 1776 : 0.000% | 1448 -> 1324 : +8.564% | 12288 -> 16384 : -33.333% | 1.165 -> 1.181 : -1.332% | 
math-partial-sums.js | 1232 -> 1232 : 0.000% | 1408 -> 1216 : +13.636% | 12288 -> 16384 : -33.333% | 0.734 -> 0.737 : -0.428% | 
math-spectral-norm.js | 6088 -> 6088 : 0.000% | 1572 -> 1572 : 0.000% | 20480 -> 20480 : 0.000% | 0.547 -> 0.553 : -1.027% | 
string-base64.js | 90792 -> 90792 : 0.000% | 1480 -> 1288 : +12.973% | 139264 -> 139264 : 0.000% | 1.533 -> 1.531 : +0.103% | 
string-fasta.js | 16376 -> 16376 : 0.000% | 1432 -> 1240 : +13.408% | 28672 -> 28672 : 0.000% | 1.341 -> 1.340 : +0.075% | 
Geometric mean: | 0.000% | +7.693% | -5.398% | -0.435% | 

Binary sizes (bytes)
be59d0a224:133440 (this is the latest master)
a3f2817d53:132640 (this is top of this PR)

Binary size and stack peak improvements are as expected. I cannot explain RSS deterioration though, especially as heap peak number haven't changed.

zherczeg:
I would expect a performance improve from this patch, rather than a binary size. Working with 64 bit values on 32 bit systems (our current target) sounds taxing though.

akosthekiss:
But the posted results are from 32 bit ARM, and clearly there is no effective change in performance.

Notes:
- Change of the binary size was expected as code contained dead parts suboptimally designed (like loops that never executed actually).
- Change of stack size was expected as the used data structure was unnecesarrily large (twice the size needed), which resulted variables not fitting in registers and spilled onto stack.
- I'm not sure I understand the "taxing 64 bit" comment. The existing code uses four 64 bit integers to store a 128 bit value only to use the upper 32 bits for addition carry. The PR uses only two 64 bit values. I cannot see how this can be more taxing. Moreover, clearly the builtin support for 64 bit arithmetics will be better on any platform than one crafted manually from 32 bit values, even on 32 bit systems, e.g., it will be able to use addition with carry, which is not available to the abstraction of C.

zherczeg:
32 bit systems have no native 64 bit types, so the compiler needs to do some emulation. The rest sounds unrelated to 32-64 bit types, so they can be fixed without changing types.

akosthekiss:
@zherczeg Just to be clear on the point: you suggest to keep using four 64 bit values to store a single 128 bit integer instead of the here-proposed two 64 bit values?

zherczeg:
four 32 bit values = 128 bit.

akosthekiss:
@zherczeg 1) I'd really appreciate if discussion happened in full sentences between parties. Opinions underpinned with justifications. Showing some respect. Etc.

2) Could you please elaborate (e.g., by giving an example) on how the cascaded addition of four 32 bit integers could be performed efficiently? In C. Because my position is that 
```c
uint64_t alo, ahi, blo, bhi;
alo += blo;
ahi += bhi;
if (alo < blo) ahi++;
```
is always better than
``` c
uint32_t a0,a1,a2,a3,b0,b1,b2,b3;
a0+=b0;
a1+=b1;
if (a0 < b0) a1++;
a2+=b2;
if (a1 < b1) a2++;
a3+=b3;
if (a2 < b2) b3++;
```
because even on 32 bit systems the "emulation" of the 64 bit arithmetics built in the compiler more clever than the handwritten C code. Because it will translate to add-32bit and add-32bit-with-carry (exact opcode differs accross architectures but these two insns are almost always present). Whether the cascaded uint32 additions will be optimized out is highly questionable. Thus, I'd go even further if possible and used a 128bit integer C type, were it available, even if emulated on any platform. Unfortunately, it is not available in the standard only in extensions.

zherczeg:
I am not sure your second example is correct:
```
// let: a0 = 0xffffffff
// let: b0 = 1
a0+=b0; // a0 = 0
// let: a1 = 0xffffffff
// let: b1 = 0
a1+=b1; // a1 unchanged: 0xffffffff
if (a0 < b0) a1++; // a1 = 0

// After this point a carry flag is lost because a1 >= b1
```

Another question what compilers do with this:
```
if (a0 < b0) a1++;
```
Since both numbers are non-negative, this cannot happen. I remember 'int32 * 10 < 0' considered impossible by some GCC compiler because the int32 was known to be non-negative.

Another question: is the standard says we need to use 128 bit numbers or it is JerryScript specific?

(I am sorry I want to understand what and why is happening here, it looks like this is an interesting theoretical problem, and by just talking about it, we might find a much better approach)

akosthekiss:
Indeed, the second snippet is buggy. And that's not even the only error there, I just realize. I was typing it from a phone, if that's an excuse. But it shows the problem quite well in my opinion. With no concept of carry, multi-part addition in C is much of a trouble / unnecessary boilerplate.

As for the question about the result of an "overflowing" expression, I think the C standard has the answer: [C99 6.2.5 Types, paragraph 9](http://www.open-std.org/jtc1/sc22/WG14/www/docs/n1256.pdf)

> A computation involving unsigned operands can never overflow, because a result that cannot be represented by the resulting unsigned integer type is reduced modulo the number that is one greater than the largest value that can be represented by the resulting type

That is, it is very much possible, `a0 + b0` can wrap around nicely and `a0 + b0` can be less than `b0`, which signals the carry.

As for the JS standard on the conversion, I haven't looked it up first, I have to admit. My primary goal was to improve the code while keeping the functionality unchanged. But the comment before `ecma_utf8_string_to_number` refers to [ES5.1 9.3.1](https://www.ecma-international.org/ecma-262/5.1/#sec-9.3.1), which I have looked up now and it expects 20 decimal digits of precision while `double` is only precise to ~15 decimal digits. This may explain why the implementation is not using the 53 bit mantissa of `double` but a longer one. Also the comment of the float32 code path ("Less precise conversion") suggests that the trivial algorithm that works with `ecma_number_t` may not be precise enough (i.e., it may be OK for float32 where we wave our hands when it comes to precision anyway but not OK for the float64 case, which has to be standard-conforming).

zherczeg:
I tried the (a+b)<b check with O3, and it seems it was not optimized out to `1`. Ok let it be this way.

Would it be a good idea to use 64+32 bit, since 96 bit seems enough for 20 digits.

akosthekiss:
There was an intentional step-up from 96 bits to 128 bits 4 years ago (https://github.com/jerryscript-project/jerryscript/commit/f0ab69b01a1806a51abe9478033c95922560ab03). I'm not sure we may want to regress.

akosthekiss:
Thanks for the reviews. Let me squash the commits.

PR #2395, opened 2018-06-11
Improve instanceof operator.

zherczeg:


LaszloLango:
What is improved in this patch? binary size, memory usage, performance? Could you add a bit more description?

zherczeg:
A recursion is eliminated. It effect should be negligible, but it still an improvement.

akosthekiss:
The PR is OK with me in general. But while improving this function, could you also touch on its doc comment? I tried to get some context from around the patch and found that this function is documented as if it were the implementation of `[[Call]]`.

zherczeg:
The doc comment is still correct since this is still the [[call]] operation, and the sources are still the same. The implementation of [[call]] is not mentioned there.

akosthekiss:
I'm not sure about this. The touched function is named `ecma_op_function_has_instance` and is only called from `ecma_op_object_has_instance`, which has the comment "[[HasInstance]] ecma object's operation". How is this related to `[[call]]`?

zherczeg:
Ah, I haven't noticed the comment issue. Ok I will change it.

PR #2397, opened 2018-06-14
Add literal property to the byte code list.

zherczeg:
Initalizing a property with a constant is frequent in object initializers.


zherczeg:
This is a new variant of an existing CBC_SET_PROPERTY opcode and uses the same VM_OC_SET_PROPERTY vm opcode.

PR #2398, opened 2018-06-14
Fix literal free when show opcodes is enabled.

zherczeg:
A literal might be not freed if it is stored in a register, and PARSER_DUMP_BYTE_CODE is enabled, but opcodes are not shown.

It is easy to trigger, but there is no test for this:
`function f(){var \u0042 = true;};`


akosthekiss:
@zherczeg If there is a known problematic test case why not add it to the test suite? (Even if it hardly triggers an issue with the current test execution settings, it may be valuable to save it with a comment attached about the details.)

PR #2400, opened 2018-06-18
Fix parse_print_final_cbc dump when arguments is used

zherczeg:
Currently the byte code start is incorrectly set when a non-strict arguments object is present, and a random memory area is dumped as byte code.

Example:
```
function abc(a,b) {
  var c = 6;
  return arguments[0] + b + c;
}
```

The begin is wrong and several CBC_EXT_NOP opcodes are printed at the end.

### DIFF
--- a/jerry-core/parser/js/js-parser.c
+++ b/jerry-core/parser/js/js-parser.c
@@ -1228,10 +1228,6 @@ parse_print_final_cbc (ecma_compiled_code_t *compiled_code_p, /**< compiled code
   }
 
   byte_code_start_p += (unsigned int) (literal_end - register_end) * sizeof (ecma_value_t);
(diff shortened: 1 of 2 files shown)


LaszloLango:
It would be great if we could check the opcode dump for each commit on the CI

akosthekiss:
Just like with the previous PR, if there is a known misbehaving test case, it would be nice to add it to the regression test suite.

zherczeg:
Unfortunately in the current environment it is not even remotely possible to test this.

akosthekiss:
My only suggestion is to have the test case saved for potential future use (annotated with a descriptive comment). Should we ever get to a state where we can test such issues, the test case wont be lost. And even if this doesn't happen any time soon, the test case should not fail on existing execution tests.

zherczeg:
Test case added. Unfortunately we probably forget about this test.

PR #2401, opened 2018-06-18
Add restart command to the debugger

knightburton:
With this feature the use can restart the actual debug session
(similar to the multiple source context reset) within a client.

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -412,6 +412,25 @@ instance_alloc (size_t size,
 
 #endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
 
+/**
+ * Inits the engine and the debugger
+ */
+static void
+init_engine (jerry_init_flag_t flags, /**< initialized flags for the engine */
+             bool debug_server, /**< enable the debugger init or not */
+             uint16_t debug_port) /**< the debugger port */
+{
+  jerry_init (flags);
+  if (debug_server)
+  {
+    jerry_debugger_init (debug_port);
+  }
+
+  register_js_function ("assert", jerryx_handler_assert);
+  register_js_function ("gc", jerryx_handler_gc);
+  register_js_function ("print", jerryx_handler_print);
+} /* init_engine */
+
 int
 main (int argc,
       char **argv)
@@ -591,15 +610,7 @@ main (int argc,
 
 #endif /* JERRY_ENABLE_EXTERNAL_CONTEXT */
 
-  jerry_init (flags);
-  if (start_debug_server)
-  {
-    jerry_debugger_init (debug_port);
-  }
-
-  register_js_function ("assert", jerryx_handler_assert);
-  register_js_function ("gc", jerryx_handler_gc);
-  register_js_function ("print", jerryx_handler_print);
+  init_engine (flags, start_debug_server, debug_port);
 
   jerry_value_t ret_value = jerry_create_undefined ();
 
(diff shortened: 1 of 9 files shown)


jerry-main/main-unix.c
```diff
@@ -629,99 +629,148 @@ main (int argc,
...
+    }
 
-        jerry_init (flags);
-        jerry_debugger_init (debug_port);
+    bool restart = false;
 
-        register_js_function ("assert", jerryx_handler_assert);
-        register_js_function ("gc", jerryx_handler_gc);
-        register_js_function ("print", jerryx_handler_print);
+    if (jerry_is_feature_enabled (JERRY_FEATURE_DEBUGGER))
```
zherczeg:
```
if (jerry_is_feature_enabled (JERRY_FEATURE_DEBUGGER))
    && jerry_value_is_abort (ret_value))
```

jerry-main/main-unix.c
```diff
@@ -629,99 +629,148 @@ main (int argc,
...
+        jerry_release_value (abort_value);
       }
     }
 
+    if (!restart)
+    {
+      break;
+    }
+
+    jerry_cleanup ();
```
zherczeg: What happens with ret_value?

jerry-main/main-unix.c
```diff
@@ -629,99 +629,148 @@ main (int argc,
...
+
+    jerry_init (flags);
+    jerry_debugger_init (debug_port);
+
+    register_js_function ("assert", jerryx_handler_assert);
+    register_js_function ("gc", jerryx_handler_gc);
+    register_js_function ("print", jerryx_handler_print);
```
zherczeg: This looks like a code duplication (init engine). Would be good to move it into a static function.

jerry-main/main-unix.c
knightburton: The `ret_value` is released in case of "r353t" match, otherwise that value is needed in the global error checking, around line `840`.

knightburton:
@zherczeg I've updated the PR.
If the `--debugger-wait-source` is enabled and the `r353t` abort is received (or any other abort) the engine closes the connection with `No more client source` message. Is this good for us or not?

zherczeg:
Closes the connection and restart jerry and wait for sources again? Sounds ok.

knightburton:
@zherczeg I've updated the PR, now the restart works properly in wait mode, too.

jerry-main/main-unix.c
```diff
@@ -629,99 +640,138 @@ main (int argc,
...
+    if (jerry_is_feature_enabled (JERRY_FEATURE_DEBUGGER) && jerry_value_is_abort (ret_value))
+    {
+      jerry_value_t abort_value = jerry_get_value_from_error (ret_value, false);
+      if (jerry_value_is_string (abort_value))
+      {
+        jerry_char_t str_buf[6];
+        jerry_value_t str_val = jerry_value_to_string (abort_value);
+        jerry_size_t str_size = jerry_get_string_size (str_val);
+        jerry_size_t str_end = jerry_string_to_char_buffer (str_val, str_buf, str_size);
+        assert (str_end == str_size);
```
zherczeg: No assert please!
Abort can have anything. Just check if str size is 5, and compare it.

jerry-main/main-unix.c
```diff
@@ -629,99 +640,138 @@ main (int argc,
...
+        str_buf[str_end] = 0;
+
+        if (strcmp ("r353t", (const char *) (str_buf)) == 0)
```
zherczeg: I would prefer memcmp here. It would make `str_buf[str_end] = 0;` unnecessary.

knightburton:
@zherczeg done, removed the assert and used memcmp instead of strcmp.

jerry-main/main-unix.c
```diff
@@ -629,99 +640,136 @@ main (int argc,
...
-        register_js_function ("print", jerryx_handler_print);
+    if (jerry_is_feature_enabled (JERRY_FEATURE_DEBUGGER) && jerry_value_is_abort (ret_value))
+    {
+      jerry_value_t abort_value = jerry_get_value_from_error (ret_value, false);
+      if (jerry_value_is_string (abort_value))
+      {
+        jerry_char_t str_buf[6];
+        jerry_value_t str_val = jerry_value_to_string (abort_value);
+        jerry_size_t str_size = jerry_get_string_size (str_val);
+        jerry_size_t str_end = jerry_string_to_char_buffer (str_val, str_buf, str_size);
```
zherczeg: Better but there is still not 100%

+        jerry_size_t str_size = jerry_get_string_size (str_val);
+        jerry_size_t str_end = jerry_string_to_char_buffer (str_val, str_buf, str_size);

Please check the str_size before jerry_string_to_char_buffer because this might be a memory overwrite.

I also think that `jerry_size_t str_end` is unnecessary.

knightburton: You are right.

PR #2403, opened 2018-06-19
Fix jerry_create_error_from_value

imiklos:
Fixed the release issue, added some test cases for the function

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -954,20 +954,19 @@ jerry_create_error_from_value (jerry_value_t value, /**< api value */
      * binary size rather than performance. */
     if (!jerry_value_is_abort (value))
     {
-      return (release ? value : jerry_acquire_value (value));
+      return release ? value : jerry_acquire_value (value);
     }
 
     value = jerry_get_value_from_error (value, release);
     release = true;
   }
 
-  jerry_value_t ret_val = ecma_create_error_reference (value, true);
-  if (release)
+  if (!release)
   {
-    jerry_release_value (value);
+    value = ecma_copy_value (value);
   }
 
-  return ret_val;
+  return ecma_create_error_reference (value, true);
 } /* jerry_create_error_from_value */
 
 /**
--- a/tests/unit-core/test-api-set-and-clear-error-flag.c
+++ b/tests/unit-core/test-api-set-and-clear-error-flag.c
@@ -16,6 +16,16 @@
 #include "jerryscript.h"
 #include "test-common.h"
 
+static void
+compare_str (jerry_value_t value, const char *str_p, size_t str_len)
+{
+  jerry_size_t size = jerry_get_string_size (value);
+  JERRY_ASSERT (str_len == size);
+  JERRY_VLA (jerry_char_t, str_buff, size);
+  jerry_string_to_utf8_char_buffer (value, str_buff, size);
+  JERRY_ASSERT (!memcmp (str_p, str_buff, str_len));
+} /* compare_str */
+
 int
 main (void)
 {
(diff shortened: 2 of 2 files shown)


tests/unit-core/test-api-set-and-clear-error-flag.c
```diff
@@ -31,6 +31,51 @@ main (void)
...
+  jerry_char_t str_buff[str_size];
+  jerry_string_to_utf8_char_buffer (str, str_buff, str_size);
+  str_buff[str_size] = '\0';
+  jerry_value_t error = jerry_create_error_from_value (str, true);
+  str = jerry_get_value_from_error (error, true);
+  str_size = jerry_get_string_size (str);
+  jerry_char_t str_buff2[str_size];
+  jerry_string_to_utf8_char_buffer (str, str_buff2, str_size);
+  str_buff2[str_size] = '\0';
+  JERRY_ASSERT (!strcmp ((char *) str_buff, (char *) str_buff2));
```
zherczeg: I would directly compare to "Pterodactylus", no need to get it from the str value.

jerry-core/api/jerry.c
```diff
@@ -954,20 +954,19 @@ jerry_create_error_from_value (jerry_value_t value, /**< api value */
      * binary size rather than performance. */
     if (!jerry_value_is_abort (value))
     {
-      return (release ? value : jerry_acquire_value (value));
+      return release ? value : jerry_acquire_value (value);
     }
 
     value = jerry_get_value_from_error (value, release);
     release = true;
```
zherczeg: Just a question: is there an existing test for this case?

tests/unit-core/test-api-set-and-clear-error-flag.c
```diff
@@ -31,6 +31,51 @@ main (void)
...
+  jerry_value_t num = jerry_create_number (3.1415926);
+  double dNum = jerry_get_number_value (num);
+  jerry_value_t num2 = jerry_create_error_from_value (num, true);
+  num2 = jerry_get_value_from_error (num2, true);
+  double dNum2 = jerry_get_number_value (num2);
+  JERRY_ASSERT (dNum == dNum2);
+  jerry_release_value (num2);
+
+  num = jerry_create_number (3.1415926);
+  dNum = jerry_get_number_value (num);
```
zherczeg: Perhaps you can simply use a local variable here:

```
double test_num = 3.1415926
jerry_value_t num = jerry_create_number (test_num);
jerry_value_t num2 = jerry_create_error_from_value (num, false);
JERRY_ASSERT (jerry_value_is_error (num2));
jerry_release_value (num);
num2 = jerry_get_value_from_error (num2, true);
JERRY_ASSERT (jerry_get_number_value (num2) == test_num);
jerry_release_value (num2);
```

The other test cases should be updated as well.

zherczeg: This could be an interesting test case as well:

```
double test_num = 3.1415926
jerry_value_t num = jerry_create_number (test_num);
jerry_value_t num2 = jerry_create_error_from_value (num, false);
JERRY_ASSERT (jerry_value_is_error (num2));
jerry_release_value (num);
jerry_value_t num3 = jerry_create_error_from_value (num2, false);
JERRY_ASSERT (jerry_value_is_error (num3));
jerry_release_value (num2);
num2 = jerry_get_value_from_error (num3, true);
JERRY_ASSERT (jerry_get_number_value (num3) == test_num);
jerry_release_value (num3);
```

With both `jerry_create_error_from_value (num2, false);` and `jerry_create_error_from_value (num2, true);`

jerry-core/api/jerry.c
imiklos: No, there isn't. I will add it, too.

imiklos:
@zherczeg I updated the PR with the changes you requested.

tests/unit-core/test-api-set-and-clear-error-flag.c
```diff
@@ -31,6 +31,82 @@ main (void)
...
+
+  jerry_value_t str = jerry_create_string ((jerry_char_t *) "Pterodactylus");
+  jerry_value_t error = jerry_create_error_from_value (str, true);
+  str = jerry_get_value_from_error (error, true);
+  jerry_size_t str_size = jerry_get_string_size (str);
+  jerry_char_t str_buff[str_size];
+  jerry_string_to_utf8_char_buffer (str, str_buff, str_size);
+  str_buff[str_size] = '\0';
+  JERRY_ASSERT (!strcmp ("Pterodactylus", (char *) str_buff));
+  jerry_release_value (str);
```
zherczeg: This part from `jerry_size_t str_size = jerry_get_string_size (str);` repeated many times. Perhaps we could introduce a helper function for it. Something like a `void compare_str_to(ecma_value_t value, const char *str_p, size_t str_len)`

tests/unit-core/test-api-set-and-clear-error-flag.c
```diff
@@ -31,6 +31,82 @@ main (void)
...
   jerry_release_value (err_val);
+  jerry_release_value (obj_val);
+
+  jerry_value_t str = jerry_create_string ((jerry_char_t *) "Pterodactylus");
+  jerry_value_t error = jerry_create_error_from_value (str, true);
+  str = jerry_get_value_from_error (error, true);
+  jerry_size_t str_size = jerry_get_string_size (str);
+  jerry_char_t str_buff[str_size];
+  jerry_string_to_utf8_char_buffer (str, str_buff, str_size);
+  str_buff[str_size] = '\0';
```
zherczeg: `str_buff[str_size] = '\0';` this is a memory overwrite since `jerry_char_t str_buff[str_size];` is defined this way. I would recommend to remove `str_buff[str_size] = '\0';` and use `memcmp()` instead of `strcmp()` below, since `memcmp()` has a size argument. Furthermore check the string size with a separate assert right after `jerry_get_string_size`.

You can also add a:
```
const char *pterodactylus_p = "Pterodactylus";
const size_t pterodactylus_size = strlen(pterodactylus_p);
```
at the beginning and use these values for assertion checks.

imiklos:
@zherczeg I updated the PR

tests/unit-core/test-api-set-and-clear-error-flag.c
```diff
@@ -16,6 +16,16 @@
 #include "jerryscript.h"
 #include "test-common.h"
 
+static void
+compare_str (jerry_value_t value, const char *str_p, size_t str_len)
+{
+  jerry_size_t size = jerry_get_string_size (value);
+  JERRY_ASSERT (str_len == size);
+  jerry_char_t str_buff[size];
```
zherczeg: I just remember that there is a JERRY_VLA macro for defining such arrays to allow windows compatibility.

imiklos: Sure, I changed it to ```JERRY_VLA``` and updated the PR.

PR #2405, opened 2018-06-19
Support for %TypedArray%.prototype.set(typedArray [, offset]).

AnthonyCalandra:
This patch allows developers to set a typedarray given a source typedarray. This patch
attempts to follow section 22.2.3.22.2 in the ECMAScript spec as closely as possible.

JerryScript-DCO-1.0-Signed-off-by: AnthonyCalandra anthony@anthony-calandra.com

### DIFF
--- a/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
+++ b/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
@@ -648,29 +648,145 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
 } /* ecma_builtin_typedarray_prototype_reverse */
 
 /**
+ * The %TypedArray%.prototype object's 'set' routine for a typedArray source
+ *
+ * See also:
+ *          ES2015, 22.2.3.22, 22.2.3.22.2
+ *
+ * @return ecma value of undefined if success, error otherwise.
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_op_typedarray_set_with_typedarray (ecma_value_t this_arg, /**< this argument */
+                                        ecma_value_t arr_val, /**< typedarray object */
+                                        ecma_value_t offset_val) /**< offset value */
+{
+  /* 6.~ 8. targetOffset */
+  ecma_number_t target_offset_num;
+  if (!ecma_is_value_empty (ecma_get_number (offset_val, &target_offset_num)))
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid offset"));
+  }
+
+  if (ecma_number_is_nan (target_offset_num))
+  {
+    target_offset_num = 0;
+  }
+
+  if (target_offset_num <= -1.0 || target_offset_num >= (ecma_number_t) UINT32_MAX + 0.5)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid offset"));
+  }
+
+  ecma_object_t *target_typedarray_p = ecma_get_object_from_value (this_arg);
+  ecma_object_t *src_typedarray_p = ecma_get_object_from_value (arr_val);
+
+  /* 9. targetBuffer */
+  ecma_object_t *target_arraybuffer_p = ecma_typedarray_get_arraybuffer (target_typedarray_p);
+  lit_utf8_byte_t *target_buffer_p = ecma_typedarray_get_buffer (target_typedarray_p);
+
+  /* 11. targetLength */
+  ecma_length_t target_length = ecma_typedarray_get_length (target_typedarray_p);
+
+  /* 12. srcBuffer */
+  ecma_object_t *src_arraybuffer_p = ecma_typedarray_get_arraybuffer (src_typedarray_p);
+  lit_utf8_byte_t *src_buffer_p = ecma_typedarray_get_buffer (src_typedarray_p);
+
+  /* 15. targetType */
+  lit_magic_string_id_t target_class_id = ecma_object_get_class_name (target_typedarray_p);
+
+  /* 16. targetElementSize */
+  uint8_t target_shift = ecma_typedarray_get_element_size_shift (target_typedarray_p);
+  uint8_t target_element_size = (uint8_t) (1 << target_shift);
+
+  /* 17. targetByteOffset */
+  ecma_length_t target_byte_offset = ecma_typedarray_get_offset (target_typedarray_p);
+
+  /* 19. srcType */
+  lit_magic_string_id_t src_class_id = ecma_object_get_class_name (src_typedarray_p);
+
+  /* 20. srcElementSize */
+  uint8_t src_shift = ecma_typedarray_get_element_size_shift (src_typedarray_p);
+  uint8_t src_element_size = (uint8_t) (1 << src_shift);
+
+  /* 21. srcLength */
+  ecma_length_t src_length = ecma_typedarray_get_length (src_typedarray_p);
+  uint32_t src_length_uint32 = ecma_number_to_uint32 (src_length);
+
+  if ((ecma_number_t) src_length_uint32 != src_length)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid source length"));
+  }
+
+  /* 22. srcByteOffset */
+  ecma_length_t src_byte_offset = ecma_typedarray_get_offset (src_typedarray_p);
+
+  /* 23. */
+  uint32_t target_offset_uint32 = ecma_number_to_uint32 (target_offset_num);
+
+  if ((int64_t) src_length_uint32 + target_offset_uint32 > target_length)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid range of index"));
+  }
+
+  /* 24.d, 25. srcByteIndex */
+  ecma_length_t src_byte_index = 0;
+
+  if (src_arraybuffer_p != target_arraybuffer_p)
+  {
+    src_byte_index = src_byte_offset;
+  }
+
+  /* 26. targetByteIndex */
+  uint32_t target_byte_index = target_offset_uint32 * target_element_size + target_byte_offset;
+
+  /* 27. limit */
+  uint32_t limit = target_byte_index + target_element_size * src_length_uint32;
+
+  if (src_class_id == target_class_id)
+  {
+    memmove (target_buffer_p + target_byte_index, src_buffer_p + src_byte_index,
+             target_element_size * src_length_uint32);
+  }
+  else
+  {
+    while (target_byte_index < limit)
+    {
+      ecma_number_t elem_num = ecma_get_typedarray_element (src_buffer_p + src_byte_index, src_class_id);
+      ecma_set_typedarray_element (target_buffer_p + target_byte_index, elem_num, target_class_id);
+      src_byte_index += src_element_size;
+      target_byte_index += target_element_size;
+    }
+  }
+
+  return ECMA_VALUE_UNDEFINED;
+} /* ecma_op_typedarray_set_with_typedarray */
+
+/**
  * The %TypedArray%.prototype object's 'set' routine
  *
  * See also:
  *          ES2015, 22.2.3.22, 22.2.3.22.1
  *
- * @return ecma value of undefined.
+ * @return ecma value of undefined if success, error otherwise.
+ *         Returned value must be freed with ecma_free_value.
  */
 ecma_value_t
 ecma_builtin_typedarray_prototype_set (ecma_value_t this_arg, /**< this argument */
                                        ecma_value_t arr_val, /**< array object */
                                        ecma_value_t offset_val) /**< offset value */
 {
-  /* 1. */
-  if (ecma_is_typedarray (arr_val))
-  {
-    /* 22.2.3.22.2 %TypedArray%.prototype(typedArray [, offset ]) is not supported */
-    return ecma_raise_type_error (ECMA_ERR_MSG ("TypedArray.set(typedArray [,offset]) is not supported."));
-  }
-
   /* 2.~ 4. */
   if (!ecma_is_typedarray (this_arg))
   {
     return ecma_raise_type_error (ECMA_ERR_MSG ("Argument 'this' is not a TypedArray."));
+  }
+
+  /* 1. */
+  if (ecma_is_typedarray (arr_val))
+  {
+    /* 22.2.3.22.2 */
+    return ecma_op_typedarray_set_with_typedarray (this_arg, arr_val, offset_val);
   }
 
   /* 6.~ 8. targetOffset */
--- a/tests/jerry/es2015/typedArray-set-with-typedArray.js
+++ b/tests/jerry/es2015/typedArray-set-with-typedArray.js
@@ -1,0 +1,94 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var a = new Int32Array([1, 2, 3, 4, 5]);
+var b = new Int32Array(5);
+
+try {
+  a.set(b, 123456);
+  assert(1 === 0); // Should not get here.
+} catch (e) {
+  assert(e instanceof RangeError);
+}
+
+b.set(a);
+assert(b.join() === '1,2,3,4,5');
+try {
+  b.set(a, 1);
+  assert(1 === 0); // Should not get here.
+} catch (e) {
+  assert(e instanceof RangeError);
+}
+
+b.set(new Int32Array([99, 98]), 2);
+assert(b.join() === '1,2,99,98,5');
+
+b.set(new Int32Array([99, 98, 97]), 2);
+assert(b.join() === '1,2,99,98,97');
+
+try {
+  b.set(new Int32Array([99, 98, 97, 96]), 2);
+  assert(1 === 0); // Should not get here.
+} catch (e) {
+  assert(e instanceof RangeError);
+}
+
+try {
+  b.set([101, 102, 103, 104], 4);
+  assert(1 === 0); // Should not get here.
+} catch (e) {
+  assert(e instanceof RangeError);
+}
+
+//  ab = [ 0, 1, 2, 3, 4, 5, 6, 7 ]
+//  a1 = [ ^, ^, ^, ^, ^, ^, ^, ^ ]
+//  a2 =             [ ^, ^, ^, ^ ]
+var ab = new ArrayBuffer(8);
+var a1 = new Uint8Array(ab);
+for (var i = 0; i < a1.length; i += 1) {
+  a1.set([i], i);
+}
+
+var a2 = new Uint8Array(ab, 4);
+a1.set(a2, 2);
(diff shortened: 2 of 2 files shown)


jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,135 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid source length"));
+  }
+
+  /* 22. srcByteOffset */
+  ecma_length_t src_byte_offset = ecma_typedarray_get_offset (src_typedarray_p);
+
+  /* 23. */
+  uint32_t target_offset_uint32 = ecma_number_to_uint32 (target_offset_num);
+  if ((int64_t) src_length_uint32 + target_offset_uint32 > target_length)
```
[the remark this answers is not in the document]
AnthonyCalandra: I want to make sure that the `offset` is within range of the target typedarray. For example, if the target has length 3 and the offset is 1000, it should throw a `RangeError`. I cast it to `int64_t` to account for overflow.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,135 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  {
+    ret_val = ecma_raise_range_error (ECMA_ERR_MSG ("Invalid range of index"));
+  }
+
+  /* 24.d, 25. srcByteIndex */
+  ecma_length_t src_byte_index;
+  ecma_object_t *src_arraybuffer_clone_p = NULL;
+  bool src_target_same_value = ecma_op_same_value (ecma_make_object_value (src_arraybuffer_p),
+                                                   ecma_make_object_value (target_arraybuffer_p));
+  if (src_target_same_value)
```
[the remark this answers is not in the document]
AnthonyCalandra: That's true, I could probably just copy the buffer and operate on that instead of cloning an entire typedarray. I'll give it a try.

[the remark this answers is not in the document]
AnthonyCalandra: I've tried to do this but I'm running into an assertion error. The code I have looks like:
```c
size_t target_clone_heap_size = (sizeof (lit_utf8_byte_t)) * target_element_size * target_length;
lit_utf8_byte_t *target_clone_buffer_p = (lit_utf8_byte_t *) jmem_heap_alloc_block (target_clone_heap_size);
memmove (target_clone_buffer_p, ecma_arraybuffer_get_buffer (target_arraybuffer_p), target_length);
src_buffer_p = target_clone_buffer_p + src_byte_offset;
// ...
```
And the assertion error is:
```c
ICE: Assertion '(uintptr_t) data_space_p % JMEM_ALIGNMENT == 0' failed at /Users/acalandra/jerryscript/jerry-core/jmem/jmem-heap.c(jmem_heap_alloc_block_internal):335.
```
Any idea how I can resolve this?

----
I would almost prefer the clone approach because it's closer to the spec and the code definitely looks cleaner.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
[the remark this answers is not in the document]
zherczeg: Hm, I checked that partial copy is not allowed. That is strange. Anyway this is an ok check then.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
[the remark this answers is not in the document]
AnthonyCalandra: I agree that this seems like a nop. When I read the spec it added the cloning step because of side-effects. I've been trying to think of a case where this would be an issue but I haven't come up with an example.

jerry-core/ecma/operations/ecma-arraybuffer-object.c
```diff
@@ -61,6 +61,34 @@ ecma_arraybuffer_new_object (ecma_length_t length) /**< length of the arraybuffe
...
+  ext_object_p->u.class_prop.extra_info = ECMA_ARRAYBUFFER_INTERNAL_MEMORY;
+  ext_object_p->u.class_prop.class_id = LIT_MAGIC_STRING_ARRAY_BUFFER_UL;
+  ext_object_p->u.class_prop.u.length = src_arraybuffer_length;
+
+  lit_utf8_byte_t *buf = (lit_utf8_byte_t *) (ext_object_p + 1);
+  memcpy (buf, src_arraybuffer_buffer_p, src_arraybuffer_length);
+
+  return object_p;
+} /* ecma_arraybuffer_clone_object */
+
```
AnthonyCalandra: Waiting to see if cloning an ArrayBuffer is necessary -- will remove if not.

AnthonyCalandra:
I've added the `memmove` optimization and added a test to account for overlapping setting of elements.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,124 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  ECMA_OP_TO_NUMBER_TRY_CATCH (target_offset_num, offset_val, ret_val);
+
+  if (ecma_number_is_nan (target_offset_num))
+  {
+    target_offset_num = 0;
+  }
+
+  if (target_offset_num <= -1.0 || target_offset_num >= (ecma_number_t) UINT32_MAX + 0.5)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid offset"));
```
zherczeg: Normally you cannot return from an ECMA_OP_TO_NUMBER_TRY_CATCH (or any similar macros). But here it would be better just not use this macro, so you could return anytime.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,124 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+
+  /* 22. srcByteOffset */
+  ecma_length_t src_byte_offset = ecma_typedarray_get_offset (src_typedarray_p);
+
+  /* 23. */
+  uint32_t target_offset_uint32 = ecma_number_to_uint32 (target_offset_num);
+
+  if ((int64_t) src_length_uint32 + target_offset_uint32 > target_length)
+  {
+    ret_val = ecma_raise_range_error (ECMA_ERR_MSG ("Invalid range of index"));
```
zherczeg: You could return here as well.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,124 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+    src_byte_index = src_byte_offset;
+  }
+
+  /* 26. targetByteIndex */
+  uint32_t target_byte_index = target_offset_uint32 * target_element_size + target_byte_offset;
+
+  /* 27. limit */
+  uint32_t limit = target_byte_index + target_element_size * src_length_uint32;
+
+  if (src_class_id == target_class_id && target_byte_index < limit && ecma_is_value_empty (ret_val))
```
zherczeg: No need |ecma_is_value_empty (ret_val).
What is the purpose of target_byte_index < limit ? It is used in the loop.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,124 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  {
+    memmove (target_buffer_p + target_byte_index, src_buffer_p + src_byte_index,
+             target_element_size * src_length_uint32);
+  }
+  else
+  {
+    while (target_byte_index < limit && ecma_is_value_empty (ret_val))
```
zherczeg: Add an assert that src and target buffers cannot be the same.

zherczeg:
Ok, I get now where the buffer cloning comes from. They wanted to simplify the pseudo code in the standard so they only need forward direction data copy.  This is how you write a standard. But not how you write code. Fortunately for in-buffer copy the type of src and target must match obviously, and memmove handles the forward/backward copy. This is its key difference from memcpy, where the result is unspecified for in-buffer copy.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: You're right, checking `target_byte_index < limit` is not needed here.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: Why is this required? The assertion fails if I want to do the following:
```js
var ab = new ArrayBuffer(8);
var a1 = new Uint8Array(ab);
var a2 = new Uint32Array(ab, 4);
a1.set(a2, 2);
```
`a1` and `a2` have different class names/IDs so the while loop gets executed even though they have the same buffer. Seems reasonable to me.

AnthonyCalandra:
Addressed all comments and I added a unit test for the case of forward/backward copying when the two buffers have different class IDs (thus using the while-loop algorithm).

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,120 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+ * @return ecma value of undefined.
+ */
+static ecma_value_t
+ecma_op_typedarray_set_with_typedarray (ecma_value_t this_arg, /**< this argument */
+                                        ecma_value_t arr_val, /**< typedarray object */
+                                        ecma_value_t offset_val) /**< offset value */
+{
+  /* 6.~ 8. targetOffset */
+  ecma_number_t target_offset_num;
+  if (!ecma_is_value_empty (ecma_get_number (offset_val, &target_offset_num)))
```
zherczeg: This is not correct. Put the result into a temporary value, and if it is an error, return with it. Do not create another error. Otherwise do nothing, so no need to free ECMA_VALUE_xxx values.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,120 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid offset"));
+  }
+
+  if (ecma_number_is_nan (target_offset_num))
+  {
+    target_offset_num = 0;
+  }
+
+  if (target_offset_num <= -1.0 || target_offset_num >= (ecma_number_t) UINT32_MAX + 0.5)
```
zherczeg: Rounding question: you check <= -1 and MAX+0.5. For consistency, it should be:
```
if (target_offset_num <= -0.5 || target_offset_num >= (ecma_number_t) UINT32_MAX + 0.5)
```
or
```
if (target_offset_num <= -1.0 || target_offset_num >= (ecma_number_t) UINT32_MAX + 1.0)
```
It depends on the rounding rules specified by the standard. I think C rounding is truncating by default.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,120 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  /* 19. srcType */
+  lit_magic_string_id_t src_class_id = ecma_object_get_class_name (src_typedarray_p);
+
+  /* 20. srcElementSize */
+  uint8_t src_shift = ecma_typedarray_get_element_size_shift (src_typedarray_p);
+  uint8_t src_element_size = (uint8_t) (1 << src_shift);
+
+  /* 21. srcLength */
+  ecma_length_t src_length = ecma_typedarray_get_length (src_typedarray_p);
+  uint32_t src_length_uint32 = ecma_number_to_uint32 (src_length);
```
zherczeg: I think ecma_length_t is uint32_t. No need a conversion here. Definitely not a to double and back to int, which is rather costly.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,120 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+
+  if ((ecma_number_t) src_length_uint32 != src_length)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid source length"));
```
zherczeg: This is impossible.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,120 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  }
+
+  /* 22. srcByteOffset */
+  ecma_length_t src_byte_offset = ecma_typedarray_get_offset (src_typedarray_p);
```
zherczeg: This is used later, please move it before use.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,120 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+
+  if ((int64_t) src_length_uint32 + target_offset_uint32 > target_length)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid range of index"));
+  }
+
+  /* 24.d, 25. srcByteIndex */
+  ecma_length_t src_byte_index = 0;
+
+  if (src_arraybuffer_p != target_arraybuffer_p)
```
zherczeg: Is src_byte_index is always 0 if the two array buffers are the same? I don't think this is true because of the subarray mechanics. This is probably not covered by a test, so please test it.

AnthonyCalandra: This is true according to the spec: https://www.ecma-international.org/ecma-262/6.0/#sec-%typedarray%.prototype.set-typedarray-offset

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: This is the same rounding check I do as https://github.com/jerryscript-project/jerryscript/pull/2405/files#diff-f81c66825ff663e22d61a8e1fb0fa8a9R797, so I just copied it over. I don't know much about rounding rules so I left it alone.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: OK, so what happened was when we decided to go with the `memmove` approach and ditch the cloning, the spec sets `srcByteIndex` to zero because it clones the buffer starting at `srcByteOffset`. Since we don't do the cloning, we always need to start at `srcByteOffset` bytes in the buffer, and the tests didn't cover that edge-case.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: I tried with latest clang and gcc on godbolt to try out the rounding rules and I think `if (target_offset_num <= -1.0 || target_offset_num >= (ecma_number_t) UINT32_MAX + 1.0)` is what we want. Not sure why the original author used 0.5 for the latter condition though.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,89 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+{
+  ecma_object_t *target_typedarray_p = ecma_get_object_from_value (this_arg);
+  ecma_object_t *src_typedarray_p = ecma_get_object_from_value (arr_val);
+
+  /* 9. targetBuffer */
+  ecma_object_t *target_arraybuffer_p = ecma_typedarray_get_arraybuffer (target_typedarray_p);
+  lit_utf8_byte_t *target_buffer_p = ecma_arraybuffer_get_buffer (target_arraybuffer_p);
+
+  /* 11. targetLength */
+  ecma_length_t target_length = ecma_typedarray_get_length (target_typedarray_p);
```
zherczeg: Move target_length after target_length definition.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,89 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
...
+  if ((int64_t) src_length + target_offset_uint32 > target_length)
+  {
+    return ecma_raise_range_error (ECMA_ERR_MSG ("Invalid range of index"));
+  }
+
+  /* 22. srcByteOffset */
+  ecma_length_t src_byte_offset = ecma_typedarray_get_offset (src_typedarray_p);
+
+  /* 24.d, 25. srcByteIndex */
+  ecma_length_t src_byte_index = src_byte_offset;
```
zherczeg: I think src_byte_offset is unnecessary.

AnthonyCalandra: Do you mean `src_byte_index`? I was on the fence about that one because it gets mutated in the while loop but I guess it doesn't matter at this point.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: After which definition?

AnthonyCalandra:
Addressed all remaining comments.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -647,6 +647,86 @@ ecma_builtin_typedarray_prototype_reverse (ecma_value_t this_arg) /**< this argu
   return ecma_copy_value (this_arg);
 } /* ecma_builtin_typedarray_prototype_reverse */
 
+/**
+ * The %TypedArray%.prototype object's 'set' routine for a typedArray source
+ *
+ * See also:
+ *          ES2015, 22.2.3.22, 22.2.3.22.2
+ *
+ * @return ecma value of undefined.
```
LaszloLango: This is inaccurate. Please update.
```
 * @return ecma value of undefined if success, error otherwise.
```

AnthonyCalandra: Done.

PR #2407, opened 2018-06-20
Add start to backtrace for debugger

Yuyupo:
With this you can add a start where you want to start listing the backtraces, not just maximum depth.

### DIFF
--- a/jerry-core/debugger/debugger.h
+++ b/jerry-core/debugger/debugger.h
@@ -376,6 +376,7 @@ typedef struct
 typedef struct
 {
   uint8_t type; /**< type of the message */
+  uint8_t min_depth[sizeof (uint32_t)]; /**< minimum depth*/
   uint8_t max_depth[sizeof (uint32_t)]; /**< maximum depth (0 - unlimited) */
 } jerry_debugger_receive_get_backtrace_t;
 
--- a/jerry-debugger/jerry-client-ws.py
+++ b/jerry-debugger/jerry-client-ws.py
@@ -200,13 +200,14 @@ class JerryFunction(object):
 
 
 class DebuggerPrompt(Cmd):
-
+    # pylint: disable=too-many-instance-attributes,too-many-arguments
     def __init__(self, debugger):
         Cmd.__init__(self)
         self.debugger = debugger
         self.stop = False
         self.quit = False
         self.cont = True
+        self.min_depth = 0
         self.non_interactive = False
         self.client_sources = []
 
@@ -348,22 +349,36 @@ class DebuggerPrompt(Cmd):
     def do_backtrace(self, args):
         """ Get backtrace data from debugger """
         max_depth = 0
+        self.min_depth = 0
 
         if args:
+            args = args.split(" ")
             try:
-                max_depth = int(args)
-                if max_depth <= 0:
-                    print("Error: Positive integer number expected")
-                    return
+                if len(args) == 2:
+                    self.min_depth = int(args[0])
+                    max_depth = int(args[1])
+                    if max_depth <= 0 or self.min_depth < 0:
+                        print("Error: Positive integer number expected")
+                        return
+                    if self.min_depth > max_depth:
+                        print("Error: Start depth needs to be lower than or equal to max depth")
+                        return
+                else:
+                    max_depth = int(args[0])
+                    if max_depth <= 0:
+                        print("Error: Positive integer number expected")
+                        return
+
             except ValueError as val_errno:
                 print("Error: Positive integer number expected, %s" % (val_errno))
                 return
 
-        message = struct.pack(self.debugger.byte_order + "BBIB" + self.debugger.idx_format,
+        message = struct.pack(self.debugger.byte_order + "BBIB" + self.debugger.idx_format + self.debugger.idx_format,
                               WEBSOCKET_BINARY_FRAME | WEBSOCKET_FIN_BIT,
-                              WEBSOCKET_FIN_BIT + 1 + 4,
+                              WEBSOCKET_FIN_BIT + 1 + 4 + 4,
                               0,
                               JERRY_DEBUGGER_GET_BACKTRACE,
+                              self.min_depth,
                               max_depth)
         self.debugger.send_message(message)
         self.stop = True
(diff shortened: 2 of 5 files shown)


jerry-core/debugger/debugger.h
```diff
@@ -376,6 +376,7 @@ typedef struct
 typedef struct
 {
   uint8_t type; /**< type of the message */
+  uint8_t start; /**< starting point (0 - max_depth)*/
```
zherczeg: This should be:
`uint8_t min_depth[sizeof (uint32_t)]; /**< minimum depth */`

Similar to:
`uint8_t max_depth[sizeof (uint32_t)]; /**< maximum depth (0 - unlimited) */`

Yuyupo:
@zherczeg Changed what was requested

jerry-debugger/jerry-client-ws.py
```diff
@@ -348,22 +348,30 @@ def do_delete(self, args):
...
         max_depth = 0
+        min_depth = 0
 
         if args:
+            args = args.split(" ")
             try:
-                max_depth = int(args)
-                if max_depth <= 0:
+                min_depth = int(args[0])
+                max_depth = int(args[1])
```
LaszloLango: What happens if there is` args` length is 1 and there is no second element?
Both `bt max_value` and `bt min_value max_value` commands should work.

LaszloLango: I see now, this is intentional and the `bt` command has two mandatory arguments. IMHO this should be changed. One argument should mean a `max_value` and two arguments should mean `min_value max_value` pair.

zherczeg: I don't mind having two mandatory arguments.

zherczeg: Sorry I misunderstood you. Yes, in the python debugger, only one argument must be mandatory. In the message, both values must be sent.

tests/debugger/do_backtrace.expected
```diff
@@ -14,10 +14,15 @@ out: function test
...
 Stopped at tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
+(jerry-debugger) bt 1 2
+Frame 0: tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
+Frame 1: tests/debugger/do_backtrace.js:33 (in test() at line:30, col:1)
 (jerry-debugger) bt
 Frame 0: tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
 Frame 1: tests/debugger/do_backtrace.js:33 (in test() at line:30, col:1)
 Frame 2: tests/debugger/do_backtrace.js:40
+(jerry-debugger) bt 1 1
+Frame 0: tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
```
zherczeg: When you specify a min, then maximum (max-min) items could be sent. Here (1-1)==0, so the backtrace should be empty (empty backtrace is valid, you need to test it!!!)

So "bt 1 2" above should only send 1 item. Please also test a large number, e.g. "bt 1000 1001".

tests/debugger/do_backtrace.expected
```diff
@@ -14,10 +14,15 @@ out: function test
 Stopped at tests/debugger/do_backtrace.js:33 (in test() at line:30, col:1)
 (jerry-debugger) s
 Stopped at tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
+(jerry-debugger) bt 1 2
+Frame 0: tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
```
LaszloLango: Is this intentional? I thought this should be `Frame 1: tests/debugger/do_backtrace.js:33 (in test() at line:30, col:1)`

tests/debugger/do_backtrace.cmd
```diff
@@ -4,9 +4,14 @@ next
...
+bt 1 2
 bt
+bt 2
 n
 n
 s
 backtrace
+bt 4 4
+bt 600 919
+bt 3 500
```
LaszloLango: What happens if `min` greater than the `max`? Can we test it?

Yuyupo:
@LaszloLango Fixed what was requested, and added the test case that was missing.

tests/debugger/do_backtrace.expected
```diff
@@ -14,10 +14,15 @@ out: function test
 Stopped at tests/debugger/do_backtrace.js:33 (in test() at line:30, col:1)
 (jerry-debugger) s
 Stopped at tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
+(jerry-debugger) bt 1 2
+Frame 0: tests/debugger/do_backtrace.js:33 (in test() at line:30, col:1)
```
LaszloLango: `Frame 1`

tests/debugger/do_backtrace.expected
```diff
@@ -14,10 +14,15 @@ out: function test
 Stopped at tests/debugger/do_backtrace.js:33 (in test() at line:30, col:1)
 (jerry-debugger) s
 Stopped at tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
+(jerry-debugger) bt 1 2
+Frame 1: tests/debugger/do_backtrace.js:23 (in foo() at line:21, col:1)
```
LaszloLango: the frame index is still wrong

jerry-core/debugger/debugger.h
```diff
@@ -376,6 +376,7 @@ typedef struct
 typedef struct
 {
   uint8_t type; /**< type of the message */
+  uint8_t min_depth[sizeof (uint32_t)]; /**< minimum depth*/
```
LaszloLango: Why don't we use `uint16_t` for both `min_depth` and `max_depth`? @zherczeg, what do you think?

zherczeg: Currently only uint8, cpointer, and uint32 types are used by the protocol. Adding new types would require adding new encoders/decoders for clients. This is certainly possible, but for this it does not seem worth the effort.

LaszloLango: I see. Thanks for the explanation.

jerry-debugger/jerry-client-ws.py
```diff
@@ -348,22 +349,36 @@ def do_delete(self, args):
...
-                    print("Error: Positive integer number expected")
-                    return
+                if len(args) == 2:
+                    self.min_depth = int(args[0])
+                    max_depth = int(args[1])
+                    if max_depth <= 0 or self.min_depth < 0:
+                        print("Error: Positive integer number expected")
+                        return
+                    if self.min_depth > max_depth:
+                        print("Error: Start needs to be lower than depth")
```
zherczeg: `Error: start depth needs to be less equal than max depth`

Yuyupo:
@zherczeg Fixed that minor issue

PR #2408, opened 2018-06-20
Remove ECMA_LEXICAL_ENVIRONMENT_OBJECT_BOUND.

zherczeg:
Only the global object bound to the root node of the lexical environment tree does not have a provideThis flag, and ecma_op_implicit_this_value() falls back to the global object. Hence we get the same effect regardless of provideThis.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -613,9 +613,8 @@ typedef enum
   /* Types between 0 - 12 are ecma_object_type_t which can have a built-in flag. */
 
   ECMA_LEXICAL_ENVIRONMENT_DECLARATIVE = 13, /**< declarative lexical environment */
-  ECMA_LEXICAL_ENVIRONMENT_OBJECT_BOUND = 14, /**< object-bound lexical environment */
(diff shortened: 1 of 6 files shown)


jerry-core/ecma/base/ecma-globals.h
```diff
@@ -613,8 +613,7 @@ typedef enum
   /* Types between 0 - 12 are ecma_object_type_t which can have a built-in flag. */
 
   ECMA_LEXICAL_ENVIRONMENT_DECLARATIVE = 13, /**< declarative lexical environment */
-  ECMA_LEXICAL_ENVIRONMENT_OBJECT_BOUND = 14, /**< object-bound lexical environment */
-  ECMA_LEXICAL_ENVIRONMENT_THIS_OBJECT_BOUND = 15, /**< object-bound lexical environment
+  ECMA_LEXICAL_ENVIRONMENT_THIS_OBJECT_BOUND = 14, /**< object-bound lexical environment
                                                    *   with provideThis flag */
```
akosthekiss: second line of comment seems to be off by one space. (legacy issue but could be fixed now)

zherczeg:
Done. Thanks for the review.

PR #2409, opened 2018-06-21
Fix escape sequence parsing in lexer_compare_identifier_to_current.

zherczeg:
Furthermore do not allow escape sequences in object initializer
get/set functions.


rerobika:
LGTM (informally)

robertsipka:
LGTM

PR #2414, opened 2018-06-29
Rework function call.

zherczeg:
Furthermore add a construct flag, which disallows calling certain functions without new. Constructing bound arrow functions correctly throws error now.

It seems the change has close to no effect:
Binary: 50 byte reduction, perf: -0.004%, mem: 0.000%

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -403,6 +403,47 @@ ecma_op_function_has_instance (ecma_object_t *func_obj_p, /**< Function object *
 } /* ecma_op_function_has_instance */
 
 /**
+ * Sets the construct flag in the arguments list pointer.
+ *
+ * @return arguments list pointer with the construct flag
+ */
+static inline const ecma_value_t * JERRY_ATTR_ALWAYS_INLINE
+ecma_op_function_set_construct_flag (const ecma_value_t *arguments_list_p) /**< original arguments list pointer */
+{
+  /* Any ecma value list must be aligned to 4 byte. */
+  JERRY_ASSERT ((((uintptr_t) arguments_list_p) & 0x3) == 0);
+
+  /* Currently it returns with the same pointer. When classes
+   * will be enabled, it will set the lowest bit. */
+  return arguments_list_p;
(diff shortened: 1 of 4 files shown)


LaszloLango:
Please add a test for this.

zherczeg:
Test case added.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -639,27 +682,28 @@ ecma_op_function_call (ecma_object_t *func_obj_p, /**< Function object */
  *         Returned value must be freed with ecma_free_value
  */
 static ecma_value_t
-ecma_op_function_construct_simple_or_external (ecma_object_t *func_obj_p, /**< Function object */
+ecma_op_function_construct_ecma_or_external (ecma_object_t *func_obj_p, /**< Function object */
                                                const ecma_value_t *arguments_list_p, /**< arguments list */
```
LaszloLango: wrong indentation

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -422,27 +463,32 @@ ecma_op_function_call (ecma_object_t *func_obj_p, /**< Function object */
...
+    ecma_object_type_t func_type = ecma_get_object_type (func_obj_p);
+
+    JERRY_ASSERT (func_type == ECMA_OBJECT_TYPE_FUNCTION
+                  || !ecma_op_function_has_construct_flag (arguments_list_p));
+
+    if (func_type == ECMA_OBJECT_TYPE_FUNCTION)
     {
+      if (JERRY_UNLIKELY (ecma_get_object_is_builtin (func_obj_p)))
+      {
+        return ecma_builtin_dispatch_call (func_obj_p,
```
rerobika: Built-in function objects **_should  never_** have a `construct_flag`, but it would be nice to see a `JERRY_ASSERT (!ecma_op_function_has_construct_flag (arguments_list_p))` before returning with `ecma_builtin_dispatch_call`.

This case currently seems impossible but the assertion could prevent the wrong usage of the `ecma_op_function_set_construct_flag` in the future.

zherczeg:
Thank you for the comments, patch is updated.

PR #2422, opened 2018-07-12
Added literal list loading feature to the snapshot tool.

LaszloLango:
It is needed to load literals and register them as magic strings
to be able to generate static snapshots. Also modified the list
format saving feature to save all of the literals not only the
identifiers.

### DIFF
--- a/jerry-core/lit/lit-magic-strings.c
+++ b/jerry-core/lit/lit-magic-strings.c
@@ -151,9 +151,7 @@ lit_magic_strings_ex_set (const lit_utf8_byte_t **ex_str_items, /**< character a
        id < JERRY_CONTEXT (lit_magic_string_ex_count);
        id = (lit_magic_string_ex_id_t) (id + 1))
   {
-    lit_utf8_size_t string_size = lit_zt_utf8_string_size (lit_get_magic_string_ex_utf8 (id));
-    JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] == string_size);
-    JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] <= LIT_MAGIC_STRING_LENGTH_LIMIT);
+    lit_utf8_size_t string_size = JERRY_CONTEXT (lit_magic_string_ex_sizes)[id];
 
     /**
      * Check whether the strings are sorted by size and lexicographically,
@@ -163,6 +161,8 @@ lit_magic_strings_ex_set (const lit_utf8_byte_t **ex_str_items, /**< character a
     {
       const lit_magic_string_ex_id_t prev_id = id - 1;
       const lit_utf8_size_t prev_string_size = lit_get_magic_string_ex_size (prev_id);
+      JERRY_ASSERT (lit_is_valid_cesu8_string (lit_get_magic_string_ex_utf8 (id),
+                                               string_size));
       JERRY_ASSERT (prev_string_size <= string_size);
 
       if (prev_string_size == string_size)
--- a/jerry-main/main-unix-snapshot.c
+++ b/jerry-main/main-unix-snapshot.c
@@ -14,6 +14,7 @@
  */
 
 #include <assert.h>
+#include <stdlib.h>
 #include <string.h>
 
 #include "jerryscript.h"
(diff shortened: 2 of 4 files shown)


jerry-core/lit/lit-magic-strings.c
```diff
@@ -151,9 +151,7 @@ lit_magic_strings_ex_set (const lit_utf8_byte_t **ex_str_items, /**< character a
        id < JERRY_CONTEXT (lit_magic_string_ex_count);
        id = (lit_magic_string_ex_id_t) (id + 1))
   {
-    lit_utf8_size_t string_size = lit_zt_utf8_string_size (lit_get_magic_string_ex_utf8 (id));
-    JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] == string_size);
-    JERRY_ASSERT (JERRY_CONTEXT (lit_magic_string_ex_sizes)[id] <= LIT_MAGIC_STRING_LENGTH_LIMIT);
+    lit_utf8_size_t string_size = JERRY_CONTEXT (lit_magic_string_ex_sizes)[id];
```
zherczeg: While the change is ok, I would keep the asserts in some form. They provide useful checks (maxlimit, cesu8 format).

jerry-main/main-unix-snapshot.c
```diff
@@ -306,8 +321,38 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
...
+    if (sp_buffer_size > 0)
+    {
+      const char* sp_buffer_p = (const char*) sp_buffer_start_p;
+      uint32_t num_of_lit = 0;
 
+      do
+      {
+        char* sp_buffer_end_p = NULL;
+        jerry_length_t mstr_size = (jerry_length_t) strtol (sp_buffer_p, &sp_buffer_end_p, 10);
+        if (mstr_size > 0)
```
zherczeg: Can it be 0?

jerry-main/main-unix-snapshot.c
```diff
@@ -306,8 +321,38 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
...
+        {
+          magic_string_items[num_of_lit] = (jerry_char_ptr_t) (sp_buffer_end_p + 1);
+          magic_string_lengths[num_of_lit] = mstr_size;
+          num_of_lit++;
+        }
+        sp_buffer_p = sp_buffer_end_p + mstr_size + 1;
+      } while ((size_t)(sp_buffer_p - (char*) sp_buffer_start_p) < sp_buffer_size);
```
zherczeg: Shall we do validation here?

jerry-main/main-unix-snapshot.c
```diff
@@ -230,17 +243,19 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
         snapshot_flags |= JERRY_SNAPSHOT_SAVE_STATIC;
         break;
       }
-      case OPT_GENERATE_LITERAL_LIST:
       case OPT_GENERATE_LITERAL_C:
+      case OPT_GENERATE_LITERAL_LIST:
+      case OPT_IMPORT_LITERAL_LIST:
       {
-        if (save_literals_file_name_p != NULL)
+        if (literals_file_name_p != NULL)
```
zherczeg: Cannot we both load and save literals?

jerry-core/lit/lit-magic-strings.c
zherczeg: Only 32? that is new to me. Btw it would be good to remove the whole macro then.
No, you can trust the size provided by the user, but you should check that the string is valid cesu8.

jerry-main/main-unix-snapshot.c
zherczeg: Shouldn't we detect that before calling strtol? What happens for invalid numbers?

jerry-main/main-unix-snapshot.c
zherczeg: Why? The loaded and saved list can be different. Although loading probably used only with static snapshots.

jerry-main/main-unix-snapshot.c
zherczeg: I mean the items are correct and correctly ordered. This is a tool, so even expensive validation is justified here.

jerry-main/main-unix-snapshot.c
LaszloLango: Do you mean something like this?
```
if ( '0' > *sp_buffer_p || *sp_buffer_p > '9')
{
  break;
}
```

jerry-main/main-unix-snapshot.c
LaszloLango: `jerry_register_magic_strings` does the verification

jerry-main/main-unix-snapshot.c
LaszloLango: I can change this, but it is not a valid use case. It will never happen to use them in the same command.

jerry-core/lit/lit-magic-strings.c
LaszloLango: I've added the check. Please check.

LaszloLango:
I've updated the PR.

PR #2425, opened 2018-07-13
Remove jerry_get_arg_value function.

zherczeg:
Remove automatic conversion of errors. Errors are primary values, just like numbers or strings.


tests/unit-ext/test-ext-module-empty.c
```diff
@@ -39,6 +39,7 @@ main (int argc, char **argv)
   TEST_ASSERT (jerry_value_is_error (module));
 
   /* Retrieve the error message. */
+  module = jerry_get_value_from_error(module, true);
```
LaszloLango: Missing space
> error: there should be exactly one space before left parentheses

PR #2426, opened 2018-07-16
Move low-level debugger connection handling into jerry-ext.

zherczeg:


zherczeg:
I think the key part is the naming of new files and functions.

PR #2427, opened 2018-07-16
Fix broken pipe error in Jerry Debugger

DanielBallaSZTE:
This patch fixes an error caused by trying to send data to a closed socket.

### DIFF
--- a/jerry-ext/debugger/debugger-tcp.c
+++ b/jerry-ext/debugger/debugger-tcp.c
@@ -37,9 +37,9 @@ typedef struct
  * Log tcp error message.
  */
 static void
-jerryx_debugger_tcp_log_error (void)
+jerryx_debugger_tcp_log_error (int err_val)
 {
-  JERRYX_ERROR_MSG ("Error: %s\n", strerror (errno));
+  JERRYX_ERROR_MSG ("TCP Error: %s\n", strerror (err_val));
 } /* jerryx_debugger_tcp_log_error */
 
 /**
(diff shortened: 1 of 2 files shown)


jerry-core/api/jerry-debugger-transport.c
```diff
@@ -161,6 +161,7 @@ jerry_debugger_transport_close (void)
   while (current_p != NULL);
 
   JERRY_CONTEXT (debugger_flags) = JERRY_DEBUGGER_VM_IGNORE;
+  JERRY_CONTEXT (debugger_flags) &= (uint32_t) ~JERRY_DEBUGGER_CONNECTED;
```
zherczeg: There is a set (assign) above. Since the flag is not set, there is no point removing it.

DanielBallaSZTE: Yes, right, fixed it. Thank you

jerry-core/debugger/debugger-tcp.c
```diff
@@ -84,8 +92,8 @@ jerry_debugger_tcp_send (jerry_debugger_transport_header_t *header_p, /**< tcp i
         continue;
       }
 
-      jerry_debugger_tcp_log_error ();
       jerry_debugger_transport_close ();
+      jerry_debugger_tcp_log_error ();
```
zherczeg: Isn't socket `close` will overwrite `errno`?

zherczeg: I get the point: don't send the message through the network. Then save `errno` into a local variable, and call log_error with that value.

PR #2428, opened 2018-07-18
Add total frame counter to backtrace in debugger

Yuyupo:
This was needed for the VScode extension so we know in total how many frames we have in the backtrace.


jerry-debugger/jerry-client-ws.py
```diff
@@ -1219,20 +1219,26 @@ def main():
...
+                        print("Frame %d: %s" % (frame_index, breakpoint[0]))
+                        buffer_pos += 6
+                        buffer_size -= 6
+                        frame_index += 1
+                    else:
+                        total_frames = breakpoint_data[0]
+                        get_total_frame = True
+                        buffer_pos += 4
+                        buffer_size -= 4
+                        print("Total Frames: %d" % total_frames)
```
LaszloLango: `print("Total number of frames: %d" % total_frames)`

BTW, do we really need to print this information in the python client? I think it would be better to ignore it.

Yuyupo: I thought it might make it easier since you'll know how many frames can you choose from.

LaszloLango: I don't mind to keep it. It is only my personal preference.

What do you think about this?
1) Omit the total number of frames when we get the full backtrace. 
```
> backtrace
frame 0: ...
frame 1: ...
....
frame N: ...
```
2) Print the remaining frames when we get the first `X` frames.
```
> backtrace X
Getting the first X frames. There are (N - X) frames left.
frame 0: ...
frame 1: ...
....
frame X: ...
```
3) Printing the number of total and got frames.
```
> backtrace X Y
Getting (Y - X) frame out of N.
frame X: ...
frame X + 1: ...
....
frame Y: ...
```

jerry-core/debugger/debugger.c
```diff
@@ -175,7 +185,7 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
 
   backtrace_p->type = JERRY_DEBUGGER_BACKTRACE_END;
 
-  jerry_debugger_send (sizeof (jerry_debugger_send_type_t) + message_size);
+  jerry_debugger_send (sizeof (jerry_debugger_send_type_t) + message_size + 4);
 } /* jerry_debugger_send_backtrace */
```
robertsipka: Please avoid magic contants, use sizeof (uint32_t) instead of 4 (maybe you should add it to the `message_size` earlier in the code. 
Furthermore, as I see, you are only checks the  `max_frame_count` size in the `max_message_size`, which is not correct. I think you should add the `total_frames` also to calculate the maximum number of items for this type which can be transmitted in one message.

jerry-core/debugger/debugger.h
```diff
@@ -354,6 +354,7 @@ typedef struct
 typedef struct
 {
   uint8_t type; /**< type of the message */
+  uint8_t total_frames[sizeof (uint32_t)]; /**< number of total frames */
   jerry_debugger_frame_t frames[]; /**< frames */
```
zherczeg: Is this part of all backtraces?

jerry-debugger/jerry-client-ws.py
zherczeg: I would add it at the end:
```
frame 0:
frame 1:
There are 7 frames left.
```

jerry-core/debugger/debugger.h
Yuyupo: Yes indeed it is part of every backtrace request. The total frames is at the beginning, so only the first transmitted message will contain this data.

jerry-debugger/jerry-client-ws.py
Yuyupo: @LaszloLango That does seem nice. Added it. :)

jerry-core/debugger/debugger.h
zherczeg: I would rather use a separate message type for sending this data. It makes the client side quite complex.

Yuyupo:
@robertsipka Changed the version number of the `JERRY_DEBUGGER_VERSION` and finished that minor adjustment.

PR #2431, opened 2018-07-18
Simplify source evaluation options.

rerobika:

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -969,12 +969,12 @@ Perform JavaScript `eval`.
 jerry_value_t
 jerry_eval (const jerry_char_t *source_p,
             size_t source_size,
-            bool is_strict);
+            uint32_t parse_opts);
 ```
 
 - `source_p` - source code to evaluate, it must be a valid utf8 string.
 - `source_size` - length of the source code
-- `is_strict` - perform `eval` as it is called from "strict mode" code.
+- `parse_opts` - any combination of [jerry_parse_opts_t](#jerry_parse_opts_t) flags.
 - return value - result of eval, may be error value.
 
 **Example**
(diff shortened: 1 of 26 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -969,12 +969,12 @@ Perform JavaScript `eval`.
 jerry_value_t
 jerry_eval (const jerry_char_t *source_p,
             size_t source_size,
-            bool is_strict);
+            uint32_t parse_opts);
```
zherczeg: If this is an api change please add it to #2213.

jerry-core/ecma/operations/ecma-eval.c
```diff
@@ -82,15 +80,19 @@ ecma_op_eval (ecma_string_t *code_p, /**< code string */
...
 {
 #ifndef JERRY_DISABLE_JS_PARSER
   JERRY_ASSERT (code_p != NULL);
 
   ecma_compiled_code_t *bytecode_data_p;
 
-  bool is_strict_call = (is_direct && is_called_from_strict_mode_code);
+  uint32_t is_strict_call = ECMA_PARSE_STRICT_MODE | ECMA_PARSE_DIRECT_EVAL;
+
+  if ((parse_opts & is_strict_call) != is_strict_call)
```
zherczeg: Are you sure this is correct?

rerobika: I think it's an equivalent conversion. In the previous version the `is_strict_call` variable was only true, if both options were true. So it means whether one of the `ECMA_PARSE_STRICT_MODE` or ` ECMA_PARSE_DIRECT_EVAL` flags are not enabled, the script must be executed in non-strict mode.

docs/02.API-REFERENCE.md
rerobika: Thanks I've added it.

jerry-core/ecma/operations/ecma-eval.c
zherczeg: Ok

PR #2432, opened 2018-07-20
[WIP] Support for shorthand object literal notation.

AnthonyCalandra:
This pull request adds support for object literal shorthand notation (properties and methods). In code:
```js
let foo = 1;
let bar = { foo }; // Equivalent to: { foo: foo }
let baz = {
  fn() {
    return 123;
  }
}; // Equivalent to: { fn: function() { return 123; } }
```
JerryScript-DCO-1.0-Signed-off-by: Anthony Calandra anthony@anthony-calandra.com

### DIFF
--- a/jerry-core/parser/js/js-lexer.c
+++ b/jerry-core/parser/js/js-lexer.c
@@ -448,6 +448,45 @@ static const keyword_string_t * const keyword_string_list[9] =
 #undef LEXER_KEYWORD
 #undef LEXER_KEYWORD_END
 
+#ifndef CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER
+/**
+ * Checks if current identifier is a keyword or reserved word.
+ *
+ * @return true - if the current identifier is a keyword or reserved word
+ *         false - otherwise
+ */
+bool
+lexer_is_identifier_keyword (parser_context_t *context_p) /**< context */
+{
+  lexer_lit_location_t *literal_p = &context_p->token.lit_location;
+
+  JERRY_ASSERT (literal_p->type == LEXER_IDENT_LITERAL);
+  JERRY_ASSERT (literal_p->length <= PARSER_MAXIMUM_IDENT_LENGTH);
+
+  if (literal_p->has_escape || literal_p->length < 2 || literal_p->length > 10)
+  {
+    return false;
+  }
+
+  const keyword_string_t *keyword_p = keyword_string_list[literal_p->length - 2];
+
+  do
+  {
+    if (literal_p->char_p[0] == keyword_p->keyword_p[0]
+        && literal_p->char_p[1] == keyword_p->keyword_p[1]
+        && memcmp (literal_p->char_p, keyword_p->keyword_p, literal_p->length) == 0)
+    {
+      return true;
+    }
+
+    keyword_p++;
+  }
+  while (keyword_p->type != LEXER_EOS);
+
+  return false;
+} /* lexer_is_identifier_keyword */
+#endif /* !CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
+
 /**
  * Parse identifier.
  */
@@ -1326,7 +1365,7 @@ lexer_check_colon (parser_context_t *context_p) /**< context */
           && context_p->source_p[0] == (uint8_t) LIT_CHAR_COLON);
 } /* lexer_check_colon */
 
-#ifndef CONFIG_DISABLE_ES2015_CLASS
+#if !defined (CONFIG_DISABLE_ES2015_CLASS) || !defined (CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER)
 /**
  * Checks whether the next token is a left parenthesis.
  *
@@ -1343,7 +1382,7 @@ lexer_check_left_paren (parser_context_t *context_p) /**< context */
   return (context_p->source_p < context_p->source_end_p
           && context_p->source_p[0] == (uint8_t) LIT_CHAR_LEFT_PAREN);
 } /* lexer_check_left_paren */
-#endif /* !CONFIG_DISABLE_ES2015_CLASS */
+#endif /* !CONFIG_DISABLE_ES2015_CLASS || !CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
 
 #ifndef CONFIG_DISABLE_ES2015_ARROW_FUNCTION
 
@@ -2271,6 +2310,7 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
 #ifndef CONFIG_DISABLE_ES2015_CLASS
   int is_class_method = ((ident_opts & LEXER_OBJ_IDENT_CLASS_METHOD)
                          && !(ident_opts & LEXER_OBJ_IDENT_ONLY_IDENTIFIERS)
+                         && !(ident_opts & LEXER_OBJ_IDENT_OBJ_METHOD)
                          && (context_p->token.type != LEXER_KEYW_STATIC));
 #endif /* !CONFIG_DISABLE_ES2015_CLASS */
 
@@ -2290,8 +2330,18 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
       {
         lexer_skip_spaces (context_p);
 
-        if (context_p->source_p < context_p->source_end_p
-            && context_p->source_p[0] != LIT_CHAR_COLON)
+        bool not_end_of_literal = (context_p->source_p < context_p->source_end_p
+                                   && context_p->source_p[0] != LIT_CHAR_COLON);
+
+#ifndef CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER
+        not_end_of_literal = (not_end_of_literal
+                               && context_p->source_p[0] != LIT_CHAR_RIGHT_BRACE
+                               && context_p->source_p[0] != LIT_CHAR_COMMA
+                               /* Shorthand notation allows methods named `get` and allows getters. */
+                               && !lexer_check_left_paren (context_p));
+#endif /* !CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
+
+        if (not_end_of_literal)
         {
           if (lexer_compare_raw_identifier_to_current (context_p, "get", 3))
           {
@@ -2346,6 +2396,12 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
       context_p->column++;
       return;
     }
+#ifndef CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER
+    else if (context_p->source_p[0] == LIT_CHAR_LEFT_PAREN)
+    {
+      create_literal_object = true;
+    }
+#endif /* !CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
     else
     {
       const uint8_t *char_p = context_p->source_p;
@@ -2364,6 +2420,16 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
         return;
       }
     }
+
+#ifndef CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER
+    if (!(ident_opts & LEXER_OBJ_IDENT_ONLY_IDENTIFIERS)
+        && (ident_opts & LEXER_OBJ_IDENT_OBJ_METHOD)
+        && context_p->source_p[0] == LIT_CHAR_LEFT_PAREN)
+    {
+      context_p->token.type = LEXER_PROPERTY_METHOD;
+      return;
+    }
+#endif /* !CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
 
     if (create_literal_object)
     {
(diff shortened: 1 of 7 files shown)


AnthonyCalandra:
Currently there is one bug that I'm aware of. The following will throw a `ReferenceError` when the correct behavior is to throw a `SyntaxError`:
```js
try {
  var a = { true, false, null };
  assert(false);
} catch (e) {
  assert(e instanceof SyntaxError);
}
```
One obvious way I can fix this is by checking to see if the literal is reserved/keyword when I know I have shorthand notation. Looks like there currently isn't a helper function available to do this, so I would probably pull out what is in `lexer_parse_identifier`. Any alternative ideas?

zherczeg:
`var a = { true:true };` is valid but `var a = { true }` is not? Is it restricted to what kind of identifiers?

AnthonyCalandra:
@zherczeg Yes that's correct. No reserved words or identifiers allowed.

jerry-core/parser/js/js-lexer.c
```diff
@@ -1343,7 +1342,6 @@ lexer_check_left_paren (parser_context_t *context_p) /**< context */
   return (context_p->source_p < context_p->source_end_p
           && context_p->source_p[0] == (uint8_t) LIT_CHAR_LEFT_PAREN);
 } /* lexer_check_left_paren */
-#endif /* !CONFIG_DISABLE_ES2015_CLASS */
```
zherczeg: Why do you remove this check? Isn't this an ES6 feature?

jerry-core/parser/js/js-lexer.c
```diff
@@ -2291,7 +2289,9 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
         lexer_skip_spaces (context_p);
 
         if (context_p->source_p < context_p->source_end_p
-            && context_p->source_p[0] != LIT_CHAR_COLON)
+            && context_p->source_p[0] != LIT_CHAR_COLON
+            && context_p->source_p[0] != LIT_CHAR_RIGHT_BRACE
+            && context_p->source_p[0] != LIT_CHAR_COMMA)
```
zherczeg: What about the pre scanner?

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -666,23 +666,58 @@ parser_parse_object_literal (parser_context_t *context_p) /**< context */
                                          literal_index,
                                          PARSER_OBJECT_PROPERTY_VALUE);
 
-      lexer_next_token (context_p);
```
zherczeg: Why do you remove `lexer_next_token`?

jerry-core/parser/js/js-lexer.c
AnthonyCalandra: It came from the ES6 classes pull request but I also use it for checking if we have a method (it checks to see if there is a left parenthesis after the identifier). If I keep the ifdef around it then if the user disables classes I won't have this function available.

jerry-core/parser/js/js-lexer.c
AnthonyCalandra: Sorry I'm not sure I understand this comment.

jerry-core/parser/js/js-parser-expr.c
AnthonyCalandra: I didn't really remove it, I just moved it down into the if statement. If we have the shorthand notation, I cant skip this early because I need to create a literal object from the current token.

jerry-core/parser/js/js-lexer.c
zherczeg: I mean this shorthand notation isn/t an es6 feature? Then we might need to change the ifdef.

jerry-core/parser/js/js-lexer.c
zherczeg: I mean the pre-scanner which checks JS code early. It is used by `switch`, `for-in`, etc.

jerry-core/parser/js/js-parser-expr.c
AnthonyCalandra: Currently that's what I'm doing, but I won't emit bytecode if I see what's after the identifier is ill-formed.

jerry-core/parser/js/js-lexer.c
AnthonyCalandra: Done. I'm planning on adding computed properties at some point, so all these new language features to object initializers could fit nicely into this macro.

AnthonyCalandra:
Update: Fixed the issue with keywords/reserved words as identifiers not throwing `SyntaxError`.

Also, I've noticed that running test262 with all ES2015 features enabled fail five tests (two of them are from these changes). Looking at the code, we only pull in from their ECMAScript 5.1 branch, which means we only test for that particular version? Should we fix the failures or leave them? For context on what my PR breaks, refer to my first comment^.

jerry-core/config.h
```diff
@@ -37,9 +37,10 @@
  */
 #ifdef CONFIG_DISABLE_ES2015
 # define CONFIG_DISABLE_ES2015_ARROW_FUNCTION
-# define CONFIG_DISABLE_ES2015_CLASS
 # define CONFIG_DISABLE_ES2015_BUILTIN
+# define CONFIG_DISABLE_ES2015_CLASS
 # define CONFIG_DISABLE_ES2015_PROMISE_BUILTIN
+# define CONFIG_DISABLE_ES2015_SHORTHAND_OBJ_LITERAL_NOTATION
```
zherczeg: I would prefer a single guard for all object literal extensions. Perhaps `CONFIG_DISABLE_ES2015_OBJ_LITERAL` is enough.

jerry-core/parser/js/js-lexer.c
```diff
@@ -448,6 +448,45 @@ static const keyword_string_t * const keyword_string_list[9] =
...
 
+#ifndef CONFIG_DISABLE_ES2015_SHORTHAND_OBJ_LITERAL_NOTATION
+/**
+ * Checks if current identifier is a keyword or reserved word.
+ *
+ * @return true - if the current identifier is a keyword or reserved word
+ *         false - otherwise
+ */
+bool
+lexer_is_identifier_keyword (parser_context_t *context_p) /**< context */
```
zherczeg: There is a similar code for this. It would be good to not duplicate that.

You need to be smart here, invent a way which can be effectively optimized by tail merging or just "reparse" the string or whatever.

jerry-core/parser/js/js-lexer.c
```diff
@@ -2290,8 +2329,17 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
       {
         lexer_skip_spaces (context_p);
 
-        if (context_p->source_p < context_p->source_end_p
-            && context_p->source_p[0] != LIT_CHAR_COLON)
+#ifndef CONFIG_DISABLE_ES2015_SHORTHAND_OBJ_LITERAL_NOTATION
+        bool not_end_of_literal = (context_p->source_p < context_p->source_end_p
+          && context_p->source_p[0] != LIT_CHAR_COLON
+          && context_p->source_p[0] != LIT_CHAR_RIGHT_BRACE
+          && context_p->source_p[0] != LIT_CHAR_COMMA);
```
zherczeg: Align these lines after the left bracket.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -666,6 +666,63 @@ parser_parse_object_literal (parser_context_t *context_p) /**< context */
...
+        }
+
+        context_p->last_cbc_opcode = CBC_SET_LITERAL_PROPERTY;
+        context_p->last_cbc.value = literal_index;
+      }
+      else
+      {
+        /* Else, we have something that isn't a valid object literal. */
+        parser_raise_error (context_p, PARSER_ERR_COLON_EXPECTED);
+      }
```
zherczeg: I don't really like this implementation. It is chaotic. You probably need a helper such as lexer_check_arrow, and add a switch.

jerry-core/parser/js/js-lexer.c
```diff
@@ -2290,8 +2329,17 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
...
+#else /* !CONFIG_DISABLE_ES2015_SHORTHAND_OBJ_LITERAL_NOTATION */
+        bool not_end_of_literal = (context_p->source_p < context_p->source_end_p
+          && context_p->source_p[0] != LIT_CHAR_COLON);
+#endif /* CONFIG_DISABLE_ES2015_SHORTHAND_OBJ_LITERAL_NOTATION */
+
+        if (not_end_of_literal)
```
zherczeg: Please add test for the pre-parser such as:
```
switch (0) { default:
  var a = { foo, foo() {}, bar:5 }
}
```

AnthonyCalandra:
Thanks @zherczeg. I will come back to this PR with a different approach.

jerry-core/parser/js/js-lexer.c
AnthonyCalandra: This should throw a `ReferenceError` because `foo` isn't defined initially. Is that intended?

rerobika: The main reason of this test to check, whether the scanner is able to pre-scan the new syntax element. As it seems it can so the thrown `Reference Error` is correct.

AnthonyCalandra:
@zherczeg I have tried to make the patch less chaotic, please take a look if my design looks reasonable this time. I still can't figure out why I'm failing those test262 tests (refer to my second post)... Any ideas?

jerry-core/parser/js/js-lexer.c
AnthonyCalandra: I still need to figure out what to do about this. My updated patch doesn't yet address this.

jerry-core/parser/js/js-parser-scanner.c
```diff
@@ -833,15 +833,25 @@ parser_scan_until (parser_context_t *context_p, /**< context */
...
         lexer_next_token (context_p);
-        if (context_p->token.type != LEXER_COLON)
+
+#ifndef CONFIG_DISABLE_ES2015_OBJ_LITERAL
+        bool not_end_of_property_name = (context_p->token.type != LEXER_COLON
+                                         && context_p->token.type != LEXER_RIGHT_BRACE
+                                         && context_p->token.type != LEXER_COMMA);
+#else /* !CONFIG_DISABLE_ES2015_OBJ_LITERAL */
+        bool not_end_of_property_name = context_p->token.type != LEXER_COLON;
+#endif /* CONFIG_DISABLE_ES2015_OBJ_LITERAL */
```
rerobika: Comment mismatch ditto.

jerry-core/parser/js/js-lexer.c
```diff
@@ -2290,8 +2330,19 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
...
+        bool not_end_of_literal = (context_p->source_p < context_p->source_end_p
+                                   && context_p->source_p[0] != LIT_CHAR_COLON
+                                   && context_p->source_p[0] != LIT_CHAR_RIGHT_BRACE
+                                   && context_p->source_p[0] != LIT_CHAR_COMMA
+                                   /* Shorthand notation allows methods named `get` and allows getters. */
+                                   && !lexer_check_left_paren (context_p));
+#else /* !CONFIG_DISABLE_ES2015_OBJ_LITERAL */
+        bool not_end_of_literal = (context_p->source_p < context_p->source_end_p
+                                   && context_p->source_p[0] != LIT_CHAR_COLON);
+#endif /* CONFIG_DISABLE_ES2015_OBJ_LITERAL */
```
rerobika: Comment mismatch, it should be:

```c
#ifndef CONFIG_DISABLE_ES2015_OBJ_LITERAL
...
#else /* CONFIG_DISABLE_ES2015_OBJ_LITERAL */
...
#endif /* !CONFIG_DISABLE_ES2015_OBJ_LITERAL */
```

zherczeg:
I would like to help you a bit @AnthonyCalandra, so I started to implement computed properties. This patch contains all around changes, which should simplify your work. The patch has not been finished yet.

zherczeg:
Forgot to add the patch number: it is #2481.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -260,7 +260,6 @@ parser_parse_array_literal (parser_context_t *context_p) /**< context */
   }
 } /* parser_parse_array_literal */
 
-#ifdef CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER
```
AnthonyCalandra: @zherczeg I noticed that everything in this macro were needed for my code so I removed the ifdefs here. I'm guessing this change in your computed properties PR was to save binary space?

AnthonyCalandra:
Rebased and resolved conflicts that arose due to computed properties addition.

zherczeg:
Since there are two things in this patch (function and keyword) maybe splitting it into two patches could help. When a patch is hard to maintain, I usually just put it away and use it for inspiration and redo the work from scratch.

Anyway let us know if you want to work on this, because we also need these features and we have to work on it if you don't have time. Sorry for the inconvenience. Unfortunately the parser is by far the most challenging part of the engine from development perspective.

PR #2434, opened 2018-07-24
[API] Save literals to a file while merging the snapshots

TamasZakor:
Add command line options to snapshot merge:
--save-literals-list-format
--save-literals-c-format

JerryScript-DCO-1.0-Signed-off-by: Tamas Zakor ztamas@inf.u-szeged.hu

### DIFF
--- a/jerry-core/api/jerry-snapshot.c
+++ b/jerry-core/api/jerry-snapshot.c
@@ -1198,157 +1198,6 @@ update_literal_offsets (uint8_t *buffer_p, /**< snapshot buffer start */
   while (buffer_p < buffer_end_p);
 } /* update_literal_offsets */
 
-#endif /* JERRY_ENABLE_SNAPSHOT_SAVE */
-
-/**
- * Merge multiple snapshots into a single buffer
- *
- * @return length of merged snapshot file
- *         0 on error
- */
-size_t
-jerry_merge_snapshots (const uint32_t **inp_buffers_p, /**< array of (pointers to start of) input buffers */
-                       size_t *inp_buffer_sizes_p, /**< array of input buffer sizes */
-                       size_t number_of_snapshots, /**< number of snapshots */
-                       uint32_t *out_buffer_p, /**< output buffer */
-                       size_t out_buffer_size, /**< output buffer size */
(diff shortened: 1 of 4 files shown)


zherczeg:
This patch looks like an API change. Is this flagged as one?

docs/02.API-REFERENCE.md
```diff
@@ -5317,15 +5317,23 @@ jerry_parse_and_save_literals (const jerry_char_t *source_p,
                                bool is_strict,
                                uint32_t *buffer_p,
                                size_t buffer_size,
-                               bool is_c_format);
+                               bool is_c_format,
+                               const char *save_literals_file_name_p,
```
zherczeg: JerryScript does not know about the concept of files.

jerry-core/api/jerry-snapshot.c
```diff
@@ -1198,157 +1198,6 @@ update_literal_offsets (uint8_t *buffer_p, /**< snapshot buffer start */
...
-#endif /* JERRY_ENABLE_SNAPSHOT_SAVE */
-
-/**
- * Merge multiple snapshots into a single buffer
- *
- * @return length of merged snapshot file
- *         0 on error
- */
-size_t
-jerry_merge_snapshots (const uint32_t **inp_buffers_p, /**< array of (pointers to start of) input buffers */
```
zherczeg: Why this function is reordered?

TamasZakor: Because some static functions is used in jerry_merge_snapshots.

docs/02.API-REFERENCE.md
TamasZakor: What do you mean by this?

zherczeg: You cannot use fopen/fclose/etc. functions in JerryScript. Theis provides the most compatibility with low-end systems.

jerry-core/api/jerry-snapshot.c
zherczeg: I suspect the logic of merging and snapshot saving is combined in this patch. It should not.

PR #2437, opened 2018-07-25
Add %TypedArray%.prototype.sort([ compareFunction ]) support.

AnthonyCalandra:
Add %TypedArray%.prototype.sort([ compareFunction ]) support.

JerryScript-DCO-1.0-Signed-off-by: Anthony Calandra anthony@anthony-calandra.com


AnthonyCalandra:
While testing the sorting functionality, I've noticed that if I do something like the following, I get unexpected output:
```javascript
jerry> [0, 1, 0].sort(function() { return NaN; })
0,0,1
jerry> [0, 1, 0].sort(function() { return 0; })
1,0,0
```
When I execute the above in V8, the arrays remain as they were, but JerryScript seems to incorrectly "sort" them.

jerry-core/ecma/base/ecma-helpers-sort.c
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Function used to reconstruct the ordered binary tree.
+ * Shifts 'index' down in the tree until it is in the correct position.
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_helpers_array_to_heap (ecma_value_t array[], /**< heap data array */
```
zherczeg: This looks like a builtin specific function, keep it in ecma/builtins. You can create a new file for them if needed though.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -1276,6 +1276,220 @@ ecma_builtin_typedarray_prototype_fill (ecma_value_t this_arg, /**< this argumen
...
+  ecma_length_t typedarray_length = ecma_typedarray_get_length (typedarray_p);
+
+  if (!typedarray_length)
+  {
+    return ecma_copy_value (this_arg);
+  }
+
+  ecma_value_t ret_value = ECMA_VALUE_EMPTY;
+
+  JMEM_DEFINE_LOCAL_ARRAY (values_buffer, typedarray_length, ecma_value_t);
```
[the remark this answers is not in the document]
AnthonyCalandra: I don't think it would work if I just gave the heap-sort helper an underlying buffer without having to change the implementation -- then that means my helper isn't generic anymore. The heap-sort seems to work on `ecma_value_t`s and I don't think I can just map variable-sized (ie. 8-bit integers, 32-bit floats, etc.) arraybuffer elements to it?

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-sort.c
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Function used to reconstruct the ordered binary tree.
+ * Shifts 'index' down in the tree until it is in the correct position.
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_helper_array_to_heap (ecma_value_t array[], /**< heap data array */
```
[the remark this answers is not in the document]
AnthonyCalandra: I just copied this over, but sure, I'll make the change.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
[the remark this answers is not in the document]
zherczeg: Ok, at least this is good for binary size. We can optimize the code later if needed, since heapsort is not exactly a complicated algorithm.

AnthonyCalandra:
I've addressed your remaining comments.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1218,10 +1079,12 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
   /* Sorting. */
   if (copied_num > 1 && ecma_is_value_empty (ret_value))
   {
+    const sort_compare_cb sortfn = &ecma_builtin_array_prototype_object_sort_compare_helper;
```
zherczeg: I might wasn't clear. Ok, here is an example from jerry-core.h:

```
/**
 * Function type for allocating buffer for JerryScript instance.
 */
typedef void *(*jerry_instance_alloc_t) (size_t size, void *cb_data_p);

/**
 * Type information of a native pointer.
 */
typedef struct
{
  jerry_object_native_free_callback_t free_cb; /**< the free callback of the native pointer */
} jerry_object_native_info_t;
```

So we put `_t` after types, including callbacks. However, for the names of the variables we put `_cb` for callbacks, `_p` for pointers. This was the rule way before I joined to the project. So the correct form looks like this:

`const sort_compare_t sort_cb = &ecma_builtin_array_prototype_object_sort_compare_helper;`

AnthonyCalandra: I understand now. Made the change.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-sort.c
```diff
@@ -0,0 +1,146 @@
...
+
+/**
+ * Function used to reconstruct the ordered binary tree.
+ * Shifts 'index' down in the tree until it is in the correct position.
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_helper_array_to_heap (ecma_value_t *array, /**< heap data array */
```
zherczeg: the `_p` postfix is still missing. All pointers require a `_p` in JerryScript. I know the original style is wrong, but lets fix it if we change it.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-sort.c
```diff
@@ -0,0 +1,146 @@
...
+                                   int index, /**< current item index */
+                                   int right, /**< right index is a maximum index */
+                                   ecma_value_t comparefn, /**< compare function */
```
zherczeg: compare_func

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-sort.c
```diff
@@ -0,0 +1,146 @@
...
+} /* ecma_builtin_helper_array_to_heap */
+
+/**
+ * Heapsort function
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+ecma_value_t
+ecma_builtin_helper_array_heap_sort_helper (ecma_value_t *array, /**< array to sort */
```
zherczeg: These two needs to be fixed as well.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -1276,6 +1276,220 @@ ecma_builtin_typedarray_prototype_fill (ecma_value_t this_arg, /**< this argumen
...
+ * See also:
+ *          ECMA-262 v5, 15.4.4.11
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_typedarray_prototype_sort_compare_helper (ecma_value_t j, /**< left value */
+                                                       ecma_value_t k, /**< right value */
+                                                       ecma_value_t comparefn) /**< compare function */
```
zherczeg: Fix all instances :)

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -1276,6 +1276,220 @@ ecma_builtin_typedarray_prototype_fill (ecma_value_t this_arg, /**< this argumen
...
+{
+  ecma_value_t ret_value = ECMA_VALUE_EMPTY;
+  ecma_number_t result = ECMA_NUMBER_ZERO;
+  bool j_is_undef = ecma_is_value_undefined (j);
+  bool k_is_undef = ecma_is_value_undefined (k);
```
zherczeg: How can this be undefined for a typed array?

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -1276,6 +1276,220 @@ ecma_builtin_typedarray_prototype_fill (ecma_value_t this_arg, /**< this argumen
...
+
+  if (j_is_undef)
+  {
+    if (k_is_undef)
+    {
+      result = ECMA_NUMBER_ZERO;
```
zherczeg: You can just use `return ecma_make_integer_value (0);`

If you use return, less else blocks and indentation is needed, which makes the code more readable.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: I don't know either... I thought I had left this in for a reason but I guess I just forgot to remove it.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
AnthonyCalandra: I've cleaned up the code a little bit since I don't need to check if `j` and `k` are undefined, see if that helps.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-sort.c
AnthonyCalandra: Sorry, not used to this style.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -1276,6 +1276,197 @@ ecma_builtin_typedarray_prototype_fill (ecma_value_t this_arg, /**< this argumen
...
+ * See also:
+ *          ECMA-262 v5, 15.4.4.11
+ *
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_typedarray_prototype_sort_compare_helper (ecma_value_t lhs, /**< left value */
+                                                       ecma_value_t rhs, /**< right value */
+                                                       ecma_value_t comparefn) /**< compare function */
```
zherczeg: Still there are some `comparefn` in the code. Please search trough the patch.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1218,10 +1079,12 @@ ecma_builtin_array_prototype_object_sort (ecma_value_t this_arg, /**< this argum
   /* Sorting. */
   if (copied_num > 1 && ecma_is_value_empty (ret_value))
   {
+    const sort_compare_t sort_cb = &ecma_builtin_array_prototype_object_sort_compare_helper;
     ECMA_TRY_CATCH (sort_value,
-                    ecma_builtin_array_prototype_object_array_heap_sort_helper (values_buffer,
-                                                                                (int)(copied_num - 1),
-                                                                                arg1),
+                    ecma_builtin_helper_array_heap_sort_helper (values_buffer,
+                                                                (int)(copied_num - 1),
```
zherczeg: There should be a space after casting. It is strange that style checkers don't catch it. See your own `(double)` castings below. There are multiple instances of this style issue, please fix all.

I am thinking about the int type here. ES allows the full uint32 range for array indicies. However we cannot allocate a local buffer that big, so it might be not possible to use that big numbers anyway. What is your opinion?

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -1276,6 +1276,197 @@ ecma_builtin_typedarray_prototype_fill (ecma_value_t this_arg, /**< this argumen
...
+  }
+
+  /*
+   * comparefn, if not undefined, will always contain a callable function object.
+   * We checked this previously, before this function was called.
+   */
+  JERRY_ASSERT (ecma_op_is_callable (comparefn));
+  ecma_object_t *comparefn_obj_p = ecma_get_object_from_value (comparefn);
+
+  ecma_value_t compare_args[] = {lhs, rhs};
```
zherczeg: Space before lfs, and after rhs.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
AnthonyCalandra: Wouldn't hurt to be consistent with ES and "allow" the full range even if we have hardware limitations, IMHO.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-sort.c
```diff
@@ -0,0 +1,146 @@
...
+ecma_value_t
+ecma_builtin_helper_array_heap_sort_helper (ecma_value_t *array_p, /**< array to sort */
+                                            uint32_t right, /**< right index */
+                                            ecma_value_t compare_func, /**< compare function */
+                                            const sort_compare_t sort_cb) /**< sorting callback */
+{
+  ecma_value_t ret_value = ECMA_VALUE_EMPTY;
+
+  /* First, construct the ordered binary tree from the array. */
+  for (uint32_t i = (right / 2) + 1; i > 0 && ecma_is_value_empty (ret_value); i--)
```
AnthonyCalandra: To allow having the type of the index a `uint32_t` I made a small change here where I added one more to `i` and changed the condition from `i >= 0` to `i > 0` because before you would decrement when `i = 0` (which would previously terminate the loop) but after making the value unsigned this would wrap to the max unsigned value and so the loop would never terminate.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-sort.c
```diff
@@ -0,0 +1,146 @@
...
+  {
+    ECMA_TRY_CATCH (value,
+                    ecma_builtin_helper_array_to_heap (array_p, i - 1, right, compare_func, sort_cb),
```
AnthonyCalandra: Also changed `i` to `i - 1`; see reason above.

zherczeg: Ok

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.h
```diff
@@ -174,6 +174,20 @@ ecma_value_t
 ecma_builtin_helper_error_dispatch_call (ecma_standard_error_t error_type, const ecma_value_t *arguments_list_p,
                                          ecma_length_t arguments_list_len);
 
+/* ecma-builtin-helpers-sort.c */
+
+/**
+ * Comparison callback function header for sorting helper routines.
+ */
+typedef ecma_value_t (*sort_compare_t)(ecma_value_t lhs, /**< left value */
```
LaszloLango: Please use the `ecma_builtin_helper_` prefix: `ecma_builtin_helper_sort_compare_fn_t`

AnthonyCalandra: @LaszloLango Done.

PR #2438, opened 2018-07-30
Improve the output format of the testrunner

pmarkee:

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -196,6 +196,13 @@ def get_arguments():
 
 BINARY_CACHE = {}
 
+TERM_NORMAL = '\033[0m'
+TERM_BLUE = '\033[1;34m'
+
+def report_command(cmd_type, cmd):
+    sys.stderr.write('%s%s%s\n' % (TERM_BLUE, cmd_type, TERM_NORMAL))
+    sys.stderr.write('%s%s%s\n' % (TERM_BLUE, (' \\%s\n\t%s' % (TERM_NORMAL, TERM_BLUE)).join(cmd), TERM_NORMAL))
+
 def create_binary(job, options):
     build_args = job.build_args[:]
     if options.buildoptions:
@@ -214,7 +221,7 @@ def create_binary(job, options):
     if options.toolchain:
         build_cmd.append('--toolchain=%s' % options.toolchain)
 
-    sys.stderr.write('Build command: %s\n' % ' '.join(build_cmd))
+    report_command('Build command:', build_cmd)
 
     binary_key = tuple(sorted(build_args))
     if binary_key in BINARY_CACHE:
@@ -273,7 +280,7 @@ def iterate_test_runner_jobs(jobs, options):
         yield job, ret_build, test_cmd
 
 def run_check(runnable):
-    sys.stderr.write('Test command: %s\n' % ' '.join(runnable))
+    report_command('Test command:', runnable)
 
     try:
         ret = subprocess.check_call(runnable)
@@ -391,11 +398,11 @@ def run_unittests(options):
         if ret_build:
             break
 
-        ret_test |= run_check([
-            settings.UNITTEST_RUNNER_SCRIPT,
-            os.path.join(build_dir_path, 'tests'),
-            "-q" if options.quiet else "",
-        ])
+        ret_test |= run_check(
+            [settings.UNITTEST_RUNNER_SCRIPT] +
+            [os.path.join(build_dir_path, 'tests')] +
+            (["-q"] if options.quiet else [])
+        )
 
     return ret_build | ret_test
 
--- a/tools/runners/run-debugger-test.sh
+++ b/tools/runners/run-debugger-test.sh
@@ -19,6 +19,10 @@ DEBUGGER_CLIENT=$2
 TEST_CASE=$3
 CLIENT_ARGS=""
 
+TERM_NORMAL='\033[0m'
+TERM_RED='\033[1;31m'
+TERM_GREEN='\033[1;32m'
+
 if [[ $TEST_CASE == *"client_source"* ]]; then
   START_DEBUG_SERVER="${JERRY} --start-debug-server --debugger-wait-source &"
   if [[ $TEST_CASE == *"client_source_multiple"* ]]; then
@@ -50,9 +54,9 @@ rm -f ${RESULT_TEMP}
 
 if [ ${STATUS_CODE} -ne 0 ]
 then
-  echo "${TEST_CASE} failed"
+  echo -e "${TERM_RED}FAIL: ${TEST_CASE}${TERM_NORMAL}\n"
 else
-  echo "${TEST_CASE} passed"
+  echo -e "${TERM_GREEN}PASS: ${TEST_CASE}${TERM_NORMAL}\n"
 fi
 
 exit ${STATUS_CODE}
--- a/tools/runners/run-test-suite.sh
+++ b/tools/runners/run-test-suite.sh
@@ -37,6 +37,10 @@ TEST_FILES=$OUTPUT_DIR/$TESTS_BASENAME.files
 TEST_FAILED=$OUTPUT_DIR/$TESTS_BASENAME.failed
 TEST_PASSED=$OUTPUT_DIR/$TESTS_BASENAME.passed
 
+TERM_NORMAL="\033[0m"
+TERM_RED="\033[1;31m"
+TERM_GREEN="\033[1;32m"
+
 VERBOSE=1
 if [[ "$1" == "-q" ]]
 then
@@ -148,29 +152,29 @@ do
         # Testing snapshot
         SNAPSHOT_TEMP=`mktemp $(basename -s .js $test).snapshot.XXXXXXXXXX`
 
-        cmd_line="$RUNTIME ${SNAPSHOT_TOOL#$ROOT_DIR} generate -o $SNAPSHOT_TEMP ${full_test#$ROOT_DIR}"
         $TIMEOUT_CMD $TIMEOUT $RUNTIME $SNAPSHOT_TOOL generate -o $SNAPSHOT_TEMP $full_test &> $ENGINE_TEMP
         status_code=$?
+        comment=" (generate snapshot)"
 
         if [ $status_code -eq 0 ]
         then
-            test $VERBOSE && echo "[$tested/$TOTAL] $cmd_line: PASS"
+            test $VERBOSE && printf "[%4d/%4d] %bPASS: %s%s%b\n" "$tested" "$TOTAL" "$TERM_GREEN" "${full_test#$ROOT_DIR}" "$comment" "$TERM_NORMAL"
 
-            cmd_line="$RUNTIME ${ENGINE#$ROOT_DIR} $ENGINE_ARGS --exec-snapshot $SNAPSHOT_TEMP"
             $TIMEOUT_CMD $TIMEOUT $RUNTIME $ENGINE $ENGINE_ARGS --exec-snapshot $SNAPSHOT_TEMP &> $ENGINE_TEMP
             status_code=$?
+            comment=" (execute snapshot)"
         fi
 
         rm -f $SNAPSHOT_TEMP
     else
-        cmd_line="$RUNTIME ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${full_test#$ROOT_DIR}"
         $TIMEOUT_CMD $TIMEOUT $RUNTIME $ENGINE $ENGINE_ARGS $full_test &> $ENGINE_TEMP
         status_code=$?
+        comment=""
     fi
 
     if [ $status_code -ne $error_code ]
     then
-        echo "[$tested/$TOTAL] $cmd_line: FAIL ($status_code)"
+        printf "[%4d/%4d] %bFAIL (%d): %s%s%b\n" "$tested" "$TOTAL" "$TERM_RED" "$status_code" "${full_test#$ROOT_DIR}" "$comment" "$TERM_NORMAL"
         cat $ENGINE_TEMP
 
         echo "$status_code: $test" >> $TEST_FAILED
@@ -182,8 +186,7 @@ do
 
         failed=$((failed+1))
     else
-        test $VERBOSE && echo "[$tested/$TOTAL] $cmd_line: $PASS"
-
+        test $VERBOSE && printf "[%4d/%4d] %b%s: %s%s%b\n" "$tested" "$TOTAL" "$TERM_GREEN" "$PASS" "${full_test#$ROOT_DIR}" "$comment" "$TERM_NORMAL"
         echo "$test" >> $TEST_PASSED
 
         passed=$((passed+1))
@@ -195,13 +198,23 @@ done
 rm -f $ENGINE_TEMP
 
 ratio=$(echo $passed*100/$TOTAL | bc)
+if [ $passed -eq $TOTAL ]
+then
+    success_color=$TERM_GREEN
+else
+    success_color=$TERM_RED
+fi
 
 if [ "$IS_SNAPSHOT" == true ]
 then
     ENGINE_ARGS="--snapshot $ENGINE_ARGS"
 fi
 
-echo "[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}: $passed PASS, $failed FAIL, $TOTAL total, $ratio% success"
+echo -e "\n[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}\n"
+echo -e "TOTAL: $TOTAL"
+echo -e "${TERM_GREEN}PASS: $passed${TERM_NORMAL}"
+echo -e "${TERM_RED}FAIL: $failed${TERM_NORMAL}\n"
+echo -e "${success_color}Success: $ratio%${TERM_NORMAL}\n"
 
 if [ $failed -ne 0 ]
 then
--- a/tools/runners/run-unittests.sh
+++ b/tools/runners/run-unittests.sh
@@ -63,6 +63,10 @@ passed=0
 
 UNITTEST_TEMP=`mktemp unittest-out.XXXXXXXXXX`
 
+TERM_NORMAL="\033[0m"
+TERM_RED="\033[1;31m"
+TERM_GREEN="\033[1;32m"
+
 for unit_test in $UNITTESTS
 do
     cmd_line="$RUNTIME ${unit_test#$ROOT_DIR}"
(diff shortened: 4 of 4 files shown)


tools/runners/run-test-suite.sh
```diff
@@ -154,7 +158,8 @@ do
 
         if [ $status_code -eq 0 ]
         then
-            test $VERBOSE && echo "[$tested/$TOTAL] $cmd_line: PASS"
+            test $VERBOSE && printf "[%4d/%4d] " "$tested" "$TOTAL" &&
```
akosthekiss: Why is it nicer to have numbers padded to 4 characters with spaces?

tools/runners/run-test-suite.sh
```diff
@@ -154,7 +158,8 @@ do
...
+                echo " PASS: $cmd_line"
```
akosthekiss: What is the reason for this line break? And why is this echo not merged into the printf after all?

tools/runners/run-test-suite.sh
```diff
@@ -168,9 +173,11 @@ do
         status_code=$?
     fi
 
+    testfile=$(echo $cmd_line | sed s/" "/"\n"/g | tail -1 | sed s/"\/"/"\n"/g | tail -1)
```
akosthekiss: What is this trying to achieve? Is this some kind of a reinvented basename?

tools/runners/run-test-suite.sh
akosthekiss: BTW, why is this not color-coded like the rest of the output?

tools/runners/run-test-suite.sh
```diff
@@ -201,7 +209,11 @@ then
     ENGINE_ARGS="--snapshot $ENGINE_ARGS"
 fi
 
-echo "[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}: $passed PASS, $failed FAIL, $TOTAL total, $ratio% success"
+echo -e "\n[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}\n"
+echo -e " TOTAL:  $TOTAL"
+echo -e "$TERM_GREEN PASS:   $passed"
```
akosthekiss: It will not cause problems here but in general it's better to clear the color settings once they are not needed.

tools/runners/run-test-suite.sh
```diff
@@ -201,7 +209,11 @@ then
     ENGINE_ARGS="--snapshot $ENGINE_ARGS"
 fi
 
-echo "[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}: $passed PASS, $failed FAIL, $TOTAL total, $ratio% success"
+echo -e "\n[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}\n"
+echo -e " TOTAL:  $TOTAL"
```
akosthekiss: In the macOS logs, the tabulation of this line seems to be off. (Not sure why.)

tools/runners/run-test-suite.sh
akosthekiss: In the string, is the space between the color code and the PASS intentional? It makes the line indented by one space. If it's only there to separate the variable expansion from the PASS string then it can be written as `"${TERM_GREEN}PASS`.

tools/runners/run-unittests.sh
```diff
@@ -97,7 +105,11 @@ rm -f $UNITTEST_TEMP
 
 ratio=$(echo $passed*100/$total | bc)
 
-echo "[summary] ${DIR#$ROOT_DIR}/unit-*: $passed PASS, $failed FAIL, $total total, $ratio% success"
+echo -e "\n[summary] ${DIR#$ROOT_DIR}/unit-*\n"
+echo -e " TOTAL: $total"
+echo -e "$TERM_GREEN PASS: $passed"
+echo -e "$TERM_RED FAIL: $failed $TERM_EMPTY\n"
```
akosthekiss: This summary is not tabulated to the same column positions as the summary printed by run-test-suite.sh.

tools/runners/run-test-suite.sh
pmarkee: I padded them to four spaces because this way so that the `PASS` and `FAIL` messages always begin in the same coloumn.
It's not colored because in this `if [ "$IS_SNAPSHOT" == true ]` block we report the passing/failing of `jerry-snapshot generate` and that command can get quite long, so it's maybe more readable if it's not green color like the rest of the output. I tried it but it wasn't really better in my opinion. But I can color it if you prefer it that way.

tools/runners/run-test-suite.sh
pmarkee: You're right, I will merge these two commands into one.

tools/runners/run-test-suite.sh
pmarkee: The first sed command is to extract only the filename from the `$cmd_line`. The second sed command is not needed, but I didn't know about basename.

pmarkee:
Thanks for the review @akosthekiss , I have updated the PR.

tools/runners/run-test-suite.sh
```diff
@@ -168,9 +172,11 @@ do
         status_code=$?
     fi
 
+    testfile=$(basename $(echo $cmd_line | sed s/" "/"\n"/g | tail -1))
```
akosthekiss: I still don't get this. Isn't `$full_test` already available here?

tools/runners/run-test-suite.sh
```diff
@@ -168,9 +172,11 @@ do
...
     if [ $status_code -ne $error_code ]
     then
-        echo "[$tested/$TOTAL] $cmd_line: FAIL ($status_code)"
+        printf "[%4d/%4d] " "$tested" "$TOTAL"
+        echo -e "${TERM_RED}FAIL ($status_code): $testfile $TERM_EMPTY"
```
akosthekiss: There is no need for mixing `printf`s and `echo`s. If `printf` is the tool of choice then let's print that log line with it. Note: `printf` can deal with escape sequences with the help of `%b%`. E.g., `printf "xxx%byyy%bzzz\n" "\e[31m" "\e[0m"` (the y's in the middle should appear red).

tools/runners/run-test-suite.sh
```diff
@@ -182,7 +188,8 @@ do
 
         failed=$((failed+1))
     else
-        test $VERBOSE && echo "[$tested/$TOTAL] $cmd_line: $PASS"
+        test $VERBOSE && printf "[%4d/%4d] " "$tested" "$TOTAL" &&
+            echo -e "${TERM_GREEN}$PASS: $testfile $TERM_EMPTY"
```
akosthekiss: Unnecessary line break + unnecessary mixing of printf and echo.

tools/runners/run-unittests.sh
```diff
@@ -63,15 +63,21 @@ passed=0
...
+TERM_RED="\033[1;31m"
+TERM_GREEN="\033[1;32m"
+
 for unit_test in $UNITTESTS
 do
     cmd_line="$RUNTIME ${unit_test#$ROOT_DIR}"
     $RUNTIME $unit_test &>$UNITTEST_TEMP
     status_code=$?
 
+    testfile=$(basename $cmd_line)
```
akosthekiss: basename'ing the whole `$cmd_line` looks very strange. Isn't it the `$unit_test` that's being looked for?

tools/run-tests.py
```diff
@@ -196,6 +196,15 @@ def get_arguments():
 
 BINARY_CACHE = {}
 
+TERM_EMPTY = '\033[0m'
+TERM_BLUE = '\033[1;34m'
+
+def report_command(cmd, cmd_type):
+    sys.stderr.write('%s%s%s' % (TERM_BLUE, cmd_type, TERM_EMPTY))
```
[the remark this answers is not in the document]
pmarkee: I thought it was intuitive and easy to read this way. I didn't think about the copying though. Maybe the best solution would be to put `\` characters at the end of each line, that would make it copyable. It was intentional to have the first item of the command on the same line, but I think I will break that into a new line as well, that will make copying easier. What do you think about this?

tools/runners/run-test-suite.sh
pmarkee: My bad, I didn't see the `$full_test` variable. But that variable contains the absolute path to the test file, which is really not necessary in my opinion. I think it would be best if we display only the name of the js file, or maybe the path relative to the `jerryscript` directory.

tools/runners/run-unittests.sh
pmarkee: Yes, same as above in one of your comments, I didn't see the `$unit_test` variable. But I think it's unnecessary to display the absolute path to the file. How about only the filename or a relative path with something like this?
```testfile=$(realpath --relative-to . $unit_test)```

File: tools/runners/run-test-suite.sh (hunk 7 above)
Comment (akosthekiss) [reply tools/runners/run-test-suite.sh]: I was most troubled by the echo+sed+tail combo. Basename might be OK. As well as the way the relative path is already being determined (see 5 lines above). So, either `$(basename $full_test)` or `${full_test#$ROOT_DIR}`.

File: tools/runners/run-unittests.sh (hunk 10 above)
Comment (akosthekiss) [reply tools/runners/run-unittests.sh]: realpath may also be an approach but there is already a way for shortening the filename. So, again, either `$(basename $unit_test)` or `${unit_test#$ROOT_DIR}`.

File: tools/run-tests.py (hunk 11 above)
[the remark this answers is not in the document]
Comment (akosthekiss) [reply tools/run-tests.py]: I'm not completely convinced about the necessity of the line breaks. But if we are going to add then (for some readability reasons -- but that's somewhat subjective), I'll propose to print it like:
```bash
command \
        --option1 \
        --option2=value2 \
        --option3
```

Note the tabulation of the command line options so that the command is easy to distinguish from its options. Also note the lack of trailing backslash at the end of the last option.

Python can solve all this simply with `sys.stderr.write('%s%s%s\n' % (TERM_BLUE, '\\\n\t'.join(cmd), TERM_EMPTY))`, if I'm not mistaken.

tools/runners/run-unittests.sh
```diff
@@ -97,7 +104,11 @@ rm -f $UNITTEST_TEMP
 
 ratio=$(echo $passed*100/$total | bc)
 
-echo "[summary] ${DIR#$ROOT_DIR}/unit-*: $passed PASS, $failed FAIL, $total total, $ratio% success"
+echo -e "\n [summary] ${DIR#$ROOT_DIR}/unit-*\n"
+echo -e " TOTAL: $total"
+echo -e "$TERM_GREEN PASS: $passed"
+echo -e "$TERM_RED FAIL: $failed $TERM_EMPTY\n"
+echo -e " Success: $ratio%\n"
```
akosthekiss: And why are all the diagnostics indented by _one_ space? It might be just me but that's very strange to my eyes.

tools/runners/run-test-suite.sh
```diff
@@ -154,7 +158,7 @@ do
 
         if [ $status_code -eq 0 ]
         then
-            test $VERBOSE && echo "[$tested/$TOTAL] $cmd_line: PASS"
+            test $VERBOSE && printf "[%4d/%4d] PASS: $cmd_line\n" "$tested" "$TOTAL"
```
akosthekiss: after this point, what is the use of `$cmd_line`? it is set at two following points, but as prints are rewritten to display the test name only, they are useless. however, the full command line is printed here, which makes the first assignment of cmd_line necessary.

note, this proposed approach makes the diagnostics somewhat inconsistent. the snapshot generation is printed in full command line, everything else just prints the js filename.

note 2, in case of a snapshot mode failure, it doesn't seem to be easy to tell whether generation or execution failed. the diagnostic message will be the same for both cases.

tools/runners/run-unittests.sh
```diff
@@ -96,8 +100,18 @@ done
...
+then
+    success_color=$TERM_GREEN
+else
+    success_color=$TERM_RED
+fi
 
-echo "[summary] ${DIR#$ROOT_DIR}/unit-*: $passed PASS, $failed FAIL, $total total, $ratio% success"
+echo -e "\n[summary] ${DIR#$ROOT_DIR}/unit-*\n"
+echo -e "TOTAL: $total"
+echo -e "${TERM_GREEN}PASS: $passed"
```
akosthekiss: missing color reset

tools/runners/run-test-suite.sh
pmarkee: It's a bit tricky because of the different cases of expected failures, for example if the snapshot generation should fail, or if the generation should pass but execution should fail, etc. How about something like this?
```sh
        if [ $status_code -eq 0 ]
        then
            test $VERBOSE && printf "[%4d/%4d] %bPASS: %s%b\n" "$tested" "$TOTAL" "$TERM_GREEN" "$SNAPSHOT_TEMP" "$TERM_NORMAL"

            $TIMEOUT_CMD $TIMEOUT $RUNTIME $ENGINE $ENGINE_ARGS --exec-snapshot $SNAPSHOT_TEMP &> $ENGINE_TEMP
            status_code=$?
        elif [ $error_code -eq 1 ]
        then
            test $VERBOSE && printf "[%4d/%4d] %bPASS (XFAIL): %s%b\n" "$tested" "$TOTAL" "$TERM_GREEN" "$SNAPSHOT_TEMP" "$TERM_NORMAL"
        else
            printf "[%4d/%4d] %bFAIL (%d): %s%b\n" "$tested" "$TOTAL" "$TERM_RED" "$status_code" "$SNAPSHOT_TEMP" "$TERM_NORMAL"
        fi
```
This way we will report both snapshot generation and execution (later in the other if block). `$cmd_line` is also deleted, so we will only display the name of the snapshot file, and now it's also colored.

tools/runners/run-test-suite.sh
```diff
@@ -182,8 +187,7 @@ do
 
         failed=$((failed+1))
     else
-        test $VERBOSE && echo "[$tested/$TOTAL] $cmd_line: $PASS"
-
+        test $VERBOSE && printf "[%4d/%4d] %bPASS: %s%b\n" "$tested" "$TOTAL" "$TERM_GREEN" "$testfile" "$TERM_NORMAL"
```
akosthekiss: I'm just realizing that you've rewritten `$PASS` to `PASS` here. That's definitely not right. `$PASS` may be `PASS` or `PASS (XFAIL)`, and that has been determined quite some lines above.

tools/runners/run-test-suite.sh
akosthekiss: I don't think that we have had any concept for XFAIL of generation. And I don't think that this patch should rewrite the decision logic of the test runner. According to the PR title, this is about formatting improvements. So, I'd suggest no logic changes.

What I could think of is to have a `$comment` variable (or something like that), that's appended to `[index/total] outcome: test.js` outputs. For the various snapshot-related steps, it could contain ' (generate snapshot)' or ' (execute snapshot)', while it could be empty for non-snapshot executions.

tools/run-tests.py
```diff
@@ -196,6 +196,13 @@ def get_arguments():
 
 BINARY_CACHE = {}
 
+TERM_NORMAL = '\033[0m'
+TERM_BLUE = '\033[1;34m'
+
+def report_command(cmd_type, cmd):
+    sys.stderr.write('%s%s%s\n' % (TERM_BLUE, cmd_type, TERM_NORMAL))
+    sys.stderr.write('%s%s%s\n\n' % (TERM_BLUE, ' \\\n\t'.join(cmd), TERM_NORMAL))
```
akosthekiss: It's weird, but it seems that Travis CI automatically resets the color at the end of every line of the log. This is non-standard and differs from the behaviour of desktop terminals. Still, it makes the logs look strange. The above snippet should be adjusted to make both terminals and Travis logs look nice.

tools/run-tests.py
```diff
@@ -214,7 +221,7 @@ def create_binary(job, options):
     if options.toolchain:
         build_cmd.append('--toolchain=%s' % options.toolchain)
 
-    sys.stderr.write('Build command: %s\n' % ' '.join(build_cmd))
+    report_command('Build command: ', build_cmd)
```
akosthekiss: trailing space within string not needed anymore

tools/run-tests.py
```diff
@@ -273,7 +280,7 @@ def iterate_test_runner_jobs(jobs, options):
         yield job, ret_build, test_cmd
 
 def run_check(runnable):
-    sys.stderr.write('Test command: %s\n' % ' '.join(runnable))
+    report_command('Test command: ', runnable)
```
akosthekiss: ditto

tools/runners/run-debugger-test.sh
```diff
@@ -50,9 +54,9 @@ rm -f ${RESULT_TEMP}
 
 if [ ${STATUS_CODE} -ne 0 ]
 then
-  echo "${TEST_CASE} failed"
+  echo -e "${TERM_RED}FAIL: ${TEST_CASE} ${TERM_NORMAL}\n"
```
akosthekiss: no space needed between ${TEST_CASE} and ${TERM_NORMAL}

tools/runners/run-debugger-test.sh
```diff
@@ -50,9 +54,9 @@ rm -f ${RESULT_TEMP}
...
 else
-  echo "${TEST_CASE} passed"
+  echo -e "${TERM_GREEN}PASS: ${TEST_CASE} ${TERM_NORMAL}\n"
```
akosthekiss: ditto

tools/runners/run-test-suite.sh
```diff
@@ -143,34 +147,34 @@ do
 
     full_test=$TESTS_DIR/${test#./}
 
+    comment=""
```
akosthekiss: move this into the else branch. that will also ensure that comment is set on all control paths and follows the logic of how cmd_line was set.

tools/runners/run-test-suite.sh
```diff
@@ -195,13 +198,23 @@ done
...
 if [ "$IS_SNAPSHOT" == true ]
 then
     ENGINE_ARGS="--snapshot $ENGINE_ARGS"
 fi
 
-echo "[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}: $passed PASS, $failed FAIL, $TOTAL total, $ratio% success"
+echo -e "\n[summary] ${ENGINE#$ROOT_DIR} $ENGINE_ARGS ${TESTS#$ROOT_DIR}\n"
+echo -e "TOTAL: $TOTAL"
+echo -e "${TERM_GREEN}PASS: $passed ${TERM_NORMAL}"
+echo -e "${TERM_RED}FAIL: $failed ${TERM_NORMAL}\n"
```
akosthekiss: no need for space between $passed/$failed and ${TERM_NORMAL}

tools/runners/run-unittests.sh
```diff
@@ -96,8 +99,18 @@ done
...
+    success_color=$TERM_GREEN
+else
+    success_color=$TERM_RED
+fi
 
-echo "[summary] ${DIR#$ROOT_DIR}/unit-*: $passed PASS, $failed FAIL, $total total, $ratio% success"
+echo -e "\n[summary] ${DIR#$ROOT_DIR}/unit-*\n"
+echo -e "TOTAL: $total"
+echo -e "${TERM_GREEN}PASS: $passed${TERM_NORMAL}"
+echo -e "${TERM_RED}FAIL: $failed ${TERM_NORMAL}\n"
```
akosthekiss: no need for space between $failed and ${TERM_NORMAL}

PR #2440, opened 2018-07-30
Fix for null pointer dereference in jmem_heap_free_block

### LINKED ISSUE #2435: Null pointer dereference in jmem-heap
- Jerry Version:
f86d745

- Build command:
`python tools/build.py --profile=es2015-subset --jerry-libc=OFF --clean --strip OFF --compile-flag="-g" --compile-flag="-fsanitize=address"`

- OS:
Ubuntu 16.04

- Test case:
`(new Int8Array(0)).filter(parseInt)`

- Result:
```ASAN:SIGSEGV
  =================================================================
  ==7815==ERROR: AddressSanitizer: SEGV on unknown address 0x000000000004 (pc 0x000000402bc1 bp 0x000000000193 sp 0x7fff3cb1ba80 T0)
      #0 0x402bc0 in jmem_heap_free_block jerry-core/jmem/jmem-heap.c:463
      #1 0x422215  (build/bin/jerry+0x422215)
      #2 0x428231 in ecma_builtin_dispatch_routine jerry-core/ecma/builtin-objects/ecma-builtins.c:906
      #3 0x428231 in ecma_builtin_dispatch_call jerry-core/ecma/builtin-objects/ecma-builtins.c:931
      #4 0x428231 in ecma_op_function_call jerry-core/ecma/operations/ecma-function-object.c:495
      #5 0x4613f5 in opfunc_call jerry-core/vm/vm.c:436
      #6 0x4613f5 in vm_execute jerry-core/vm/vm.c:3008
      #7 0x461fda in vm_run jerry-core/vm/vm.c:3092
      #8 0x424411 in vm_run_eval jerry-core/vm/vm.c:270
      #9 0x424411 in ecma_op_eval_chars_buffer jerry-core/ecma/operations/ecma-eval.c:111
      #10 0x40274f in jerry_eval jerry-core/api/jerry.c:570
      #11 0x40274f in main jerry-main/main-unix.c:810
      #12 0x7f749b68f82f in __libc_start_main (/lib/x86_64-linux-gnu/libc.so.6+0x2082f)
      #13 0x403148 in _start (build/bin/jerry+0x403148)

  AddressSanitizer can not provide additional info.
  SUMMARY: AddressSanitizer: SEGV jerry-core/jmem/jmem-heap.c:463 jmem_heap_free_block
  ==7815==ABORTING
```

- Immediate cause:
It seems that this bug is caused by triggering undefined behavior at jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c:598. The second passed argument is NULL. This causes NULL dereference and allows GCC to optimize out subsequent checks for NULL. This further causes writing to null pointer, which results in a segmentation fault. Note, that because this bug occurs due to gcc optimizing out a check for NULL, this may not be reproducible with a debug build.

Please confirm if you can reproduce this issue.

This bug was discovered by Marcin Dominiak and Wojciech Rauner.

DanielBallaSZTE:
Fixes #2435.

### DIFF
--- a/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
+++ b/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
@@ -553,6 +553,11 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
   ecma_object_t *func_object_p = ecma_get_object_from_value (cb_func_val);
   ecma_value_t ret_value = ECMA_VALUE_EMPTY;
 
+  if (len == 0)
+  {
+    return ecma_op_create_typedarray_with_type_and_length (obj_p, 0);
+  }
+
   JMEM_DEFINE_LOCAL_ARRAY (pass_value_list_p, len * element_size, lit_utf8_byte_t);
 
   lit_utf8_byte_t *pass_value_p = pass_value_list_p;
@@ -587,7 +592,6 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
     uint32_t pass_num = (uint32_t) ((pass_value_p - pass_value_list_p) >> shift);
 
     ret_value = ecma_op_create_typedarray_with_type_and_length (obj_p, pass_num);
-
 
     if (!ECMA_IS_VALUE_ERROR (ret_value))
     {
--- a/tests/jerry/es2015/regression-test-issue-2435.js
+++ b/tests/jerry/es2015/regression-test-issue-2435.js
@@ -1,0 +1,15 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+(new Int8Array(0)).filter(parseInt)


dominiakm:
Hi,
I'm the original poster of the issue you are fixing here.

While this change does prevent segfault, undefined behavior is still triggered at jerryscript/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c:598

This is because behavior of memcpy is undefined if the second argument is NULL.
This can be verified by using UBSan (undefined behavior sanitizer).
To do this, build jerryscript with
`python tools/build.py --profile=es2015-subset --jerry-libc=off --lto=off --compile-flag="-fsanitize=undefined"`

Running the example provided by me shows the following message:
`jerryscript/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c:598:7: runtime error: null pointer passed as argument 2, which is declared to never be null
/usr/include/x86_64-linux-gnu/bits/string3.h:53:10: runtime error: null pointer passed as argument 2, which is declared to never be null.`

dominiakm:
The implementation of `filter` (from jerryscript/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c) calls `memcpy` (in line 598) with the second argument (`pass_value_list_p`) being `NULL`.

Current implementation (without this commit) in JMEM_FINALIZE_LOCAL_ARRAY checks if `var_name` (`pass_value_list_p` in this case) is `NULL`. However, this check is optimized out, because of the call to `memcpy`. This is because gcc assumes, that that the second argument of memcpy is never NULL.

The default build for jerryscript enbales gcc optimisation `-fdelete-null-pointer-checks`, which removes the check from JMEM_FINALIZE_LOCAL_ARRAY  -- because of the call to memcpy.

DanielBallaSZTE:
I've added an early return to resolve this issue, and the test case as well.

dominiakm:
This version no longer triggers undefined behavior.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -588,6 +588,11 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
 
     ret_value = ecma_op_create_typedarray_with_type_and_length (obj_p, pass_num);
 
+    if (len == 0)
+    {
+      return ret_value;
+    }
+
```
zherczeg: No need an extra newline.

jerry-core/jmem/jmem.h
```diff
@@ -195,16 +195,11 @@ void jmem_run_free_unused_memory_callbacks (jmem_free_unused_memory_severity_t s
...
   { \
-    JERRY_ASSERT (var_name ## ___size != 0); \
-    \
+    JERRY_ASSERT (var_name != NULL); \
     jmem_heap_free_block (var_name, var_name ## ___size); \
   } \
-  else \
-  { \
-    JERRY_ASSERT (var_name ## ___size == 0); \
-  } \
```
zherczeg: This change should be removed.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -588,6 +588,10 @@ ecma_builtin_typedarray_prototype_filter (ecma_value_t this_arg, /**< this argum
 
     ret_value = ecma_op_create_typedarray_with_type_and_length (obj_p, pass_num);
 
+    if (len == 0)
+    {
+      return ret_value;
+    }
```
zherczeg: Hm, why this is not before the define local array?

DanielBallaSZTE: Because in some cases we can not return with `ECMA_VALUE_EMPTY`

zherczeg: so we return with a new typedarray?

zherczeg: actually we return from a macro enclosed block, and that is not necessarily healthy.

DanielBallaSZTE: Yes but we only return from there when the length is 0, therefore the enclosing macro would basically do nothing

DanielBallaSZTE: And yes, we have to give back a 0 length typedarray in that case

zherczeg: The problem is we might change silently the macro and we have a hidden bug. Creating and returning with an empty array early should be easy.

DanielBallaSZTE: Yes, that is definitely right, it would just mean a bit if a code duplication there.

zherczeg: The compiler should handle that. Since it is a single line, I would not even call it a duplication.

DanielBallaSZTE: Yeah you convinced me there, I'm updating the PR

PR #2441, opened 2018-07-31
Improve python debugger client.

zherczeg:
Replace DisplayData to DebuggerAction. The new class has only four type options. Furthermore several functions returning with DisplayData is changed to return with string.


jerry-debugger/jerry_client.py
```diff
@@ -60,54 +58,49 @@ def do_display(self, args):
 
     def do_break(self, args):
         """ Insert breakpoints on the given lines or functions """
-        result = ""
-        result = self.debugger.set_break(args)
-        if self.debugger.not_empty(result):
-            print(result.get_data())
+        sys.stdout.write(self.debugger.set_break(args))
```
akosthekiss: what's the reason for rewriting lots of `print`s to `sys.stdout.write`? `print` is a lot more pythonic, `sys` is too low-level. unless there is a good reason, it's generally better practice to use `print`. (even for printing to stderr, if that's needed somewhere.)

LaszloLango: +1 for keeping the print

zherczeg: Same problem as `putstr`, adds an unwanted newline. People recommended to use stdout.write.

akosthekiss: I'm not completely sure what the unwanted newline may be. This code is printing to the console and users typically expect info to be broken into lines. If this is just part of the output, it may be a signal that this function should not print anything but return a string, perhaps, that can be composed by the caller into a full line of text and printed there.

Anyway, if the only feature that's missing is the no-newline-at-the-end-of-printing, then consult the manual: https://docs.python.org/2/library/functions.html#print . (`print('text to be printed', end='')`.

zherczeg: Strings returned by the debugger client contains newlines, including terminating newlines. If you print them with `print` you get two newlines, and that is ugly. The problem with `end=''` that it is python 3. The current one is compatible with any kind of python.

I have a suggestion: I can create a utility, e.g `yield` function:
```
def yield(str):
    sys.stdout.write(str)
```

And use yield everywhere. What do you think?

akosthekiss: The link I've sent you is Python 2 docs. Did you import print function from future?

As for yield, don't do that. Yield is a keyword in Python 3, used for generators.

akosthekiss: Yes, print.

```
$ python2
Python 2.7.6 (default, Nov 23 2017, 15:49:48)
[GCC 4.8.4] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import print_function
>>> print('text to be printed', end='')
text to be printed>>>
```

zherczeg: I don't like adding `end=''` at the end of each print. I would prefer to introduce a new function.

akosthekiss: I don't like reinventing the wheel. `print` is there. `end` is there, and it's not Python 3, it is compatible with any kind of Python. Moreover, if you introduce a new function, it'll either be a member of the `DebuggerPrompt` class or a becomes a global function in this module. The first option will mandate a `self.` prefix to all print replacements, which is hardly any better than `end=''`. If it's a global function, then it's ugly and awkward to have a single module-level `print_without_end` utility function that's only used by the class.

I still +1 myself and the `end=''` approach.

zherczeg: This function would be in the main program, not in the Debugger module, which should print nothing at all (and no need to know anything about where the output goes). No need for `self` in this utility as well.

akosthekiss: I don't understand what this means. I see no Debugger python module, and 'main program' (if it means the `main` function) is in this file where we are discussing the the pros and cons of `print` vs utility functions.

If 'main program' stands for the whole jerry_client.py file, then I've already commented on that approach (see: module-level global function used by a single class only -- ugly and awkward).

zherczeg: The debugger client has two files at the moment: `jerry_client_ws.py` which is the Debugger Client module, and  `jerry_client.py` application which uses the module to provide the user interface. The user interface is not something reusable, just a simple application. Adding utility functions there is normal.

akosthekiss: Just because something can be technically done doesn't mean that it is good design. Even the whole class in the jerry_client.py could be dropped and replaced with global functions and some global state. But that also doesn't mean that it is the way to go.

zherczeg: The `jerry_client.py` is not a single class file. It contains a class because python command line support designed this way, but that class is just a utility class. It cannot be avoided unless we use another command line system. So this is not a design, it is just forced on us.

The core function is the `main()` function, which is a message loop processor.

Patch updated.

zherczeg:
Any more comments?

jerry-debugger/jerry_client.py
```diff
@@ -229,66 +218,41 @@ def main():
...
         prompt.stop = False
-        if prompt.debugger.check_empty_data(args.client_source):
-            prompt.debugger.mainloop()
-            result = prompt.debugger.smessage
-            print(result)
-            prompt.debugger.smessage = ''
-            prompt.cmdloop()
 
-    if  prompt.debugger.not_empty(args.exception):
+    if args.exception != None:
```
akosthekiss: In python, this should be written as `args.exception is not None`

jerry-debugger/jerry_client_ws.py
```diff
@@ -363,53 +365,53 @@ def _exec_command(self, command_id):
...
     def step(self):
+        self.prompt = False
         self._exec_command(JERRY_DEBUGGER_STEP)
 
     def memstats(self):
+        self.prompt = False
         self._exec_command(JERRY_DEBUGGER_MEMSTATS)
 
     def set_break(self, args):
         if args == "":
```
akosthekiss: This could also be `if not args` as at the beginning of `def delete`.

jerry-debugger/jerry_client_ws.py
```diff
@@ -744,7 +749,9 @@ def mainloop(self):
                                JERRY_DEBUGGER_SOURCE_CODE_NAME_END,
                                JERRY_DEBUGGER_FUNCTION_NAME,
                                JERRY_DEBUGGER_FUNCTION_NAME_END]:
-                _parse_source(self, data)
+                result = _parse_source(self, data)
+                if result != "":
```
akosthekiss: This could also simply be `if result:`

jerry-debugger/jerry_client_ws.py
```diff
@@ -1067,31 +1045,39 @@ def _parse_source(debugger, data):
...
-                    source_lines = 0
+                    break
 
             if breakpoint.line:
                 if breakpoint.line <= source_lines:
                     command = breakpoint.source_name + ":" + str(breakpoint.line)
-                    if _set_breakpoint(debugger, command, True):
+                    set_result = _set_breakpoint(debugger, command, True)
+
+                    if set_result != "":
```
akosthekiss: ditto, `if set_result:`

jerry-debugger/jerry_client_ws.py
```diff
@@ -1067,31 +1045,39 @@ def _parse_source(debugger, data):
...
+                        result += set_result
                         del bp_list[breakpoint_index]
             elif breakpoint.function:
                 command = breakpoint.function
-                if _set_breakpoint(debugger, command, True):
+                set_result = _set_breakpoint(debugger, command, True)
+
+                if set_result != "":
```
akosthekiss: ditto

jerry-debugger/jerry_client_ws.py
```diff
@@ -1147,16 +1134,14 @@ def _set_breakpoint(debugger, string, pending):
...
 
     else:
         for function in debugger.function_list.values():
             if function.name == string:
-                _enable_breakpoint(debugger, function.lines[function.first_breakpoint_line])
-                found = True
+                result += _enable_breakpoint(debugger, function.lines[function.first_breakpoint_line])
 
-    if not found and not pending:
+    if result == "" and not pending:
```
akosthekiss: `if not result and not pending:`

zherczeg:
Thank you for the review. Patch updated.

LaszloLango:
Why did you changed the file permissions of `jerry-debugger/jerry_client_ws.py`?

zherczeg:
It is a library, should never be executed directly.

PR #2442, opened 2018-07-31
Fix a few parser style issues.

zherczeg:


zherczeg:
The issue is that lexer_hex_to_character is used by some internal tests. I don't really like this, but we can keep it if needed.

zherczeg:
Any suggestion?

LaszloLango:
Let's keep it or declare as an extern function in the related unit test file.

akosthekiss:
My two cents: keep `lexer_hex_to_character` as non-static. There are quite some other non-API functions that are accessed from unit tests, too. (Even in test-lit-char-helpers.c, e.g., `lit_char_get_utf8_length`.) For now, we have to live with keeping them non-static, IMHO.

Some (distant?) future work may deal with solving this, i.e., how to make those functions static that are only published for testing reasons. But that will deal with all such functions then, not only this one.

PR #2324, opened 2018-05-11
Update Doxygen config file and fix Doxygen warnings

LaszloLango:
A lot of warnings remained hibben because 'EXTRACT_ALL' was previously set to YES.

JerryScript-DCO-1.0-Signed-off-by: László Langó llango.u-szeged@partner.samsung.com

PR #1540, opened 2017-01-24
Add doxygen checker

LaszloLango:
* Fixed Doxygen issues
 * Updated Doxygen config
 * Added new script to test whether doxygen generation was successful or not
 * Added to Travis CI

JerryScript-DCO-1.0-Signed-off-by: László Langó llango.u-szeged@partner.samsung.com

PR #2446, opened 2018-08-06
Fix FILE_PATTERNS of Doxyfile and some of the issues it was hiding

akosthekiss:
`FILE_PATTERNS` is a space-separated list, in contrary to what is
suggested by its documentation. The pattern `*.h, *.c` does not
match header files but files with `.h,` extension only. Rewriting
the current comma-separated pattern makes Doxygen actually process
header files. However, it also reveals several hitherto hidden
issues (mostly missing documentation) in the code. This patch fixes
some of these documentation problems (and lists the files with
outstanding issues in a 'backlog').

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -142,7 +142,9 @@ typedef int32_t ecma_integer_value_t;
  */
 #define ECMA_DIRECT_SHIFT 4
 
-/* ECMA make simple value */
+/**
+ * ECMA make simple value
+ */
 #define ECMA_MAKE_VALUE(value) \
   ((((ecma_value_t) (value)) << ECMA_DIRECT_SHIFT) | ECMA_DIRECT_TYPE_SIMPLE_VALUE)
 
@@ -171,25 +173,43 @@ enum
                                                                 *   a special "base" value for vm */
 };
 
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
 /**
  * Maximum integer number for an ecma value
  */
-#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
 #define ECMA_INTEGER_NUMBER_MAX         0x7fffff
+/**
+ * Maximum integer number for an ecma value (shifted left with ECMA_DIRECT_SHIFT)
+ */
 #define ECMA_INTEGER_NUMBER_MAX_SHIFTED 0x7fffff0
 #else /* CONFIG_ECMA_NUMBER_TYPE != CONFIG_ECMA_NUMBER_FLOAT32 */
+/**
+ * Maximum integer number for an ecma value
+ */
 #define ECMA_INTEGER_NUMBER_MAX         0x7ffffff
+/**
+ * Maximum integer number for an ecma value (shifted left with ECMA_DIRECT_SHIFT)
+ */
 #define ECMA_INTEGER_NUMBER_MAX_SHIFTED 0x7ffffff0
 #endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32 */
 
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
 /**
  * Minimum integer number for an ecma value
  */
-#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
 #define ECMA_INTEGER_NUMBER_MIN         -0x7fffff
+/**
+ * Minimum integer number for an ecma value (shifted left with ECMA_DIRECT_SHIFT)
+ */
 #define ECMA_INTEGER_NUMBER_MIN_SHIFTED -0x7fffff0
 #else /* CONFIG_ECMA_NUMBER_TYPE != CONFIG_ECMA_NUMBER_FLOAT32 */
+/**
+ * Minimum integer number for an ecma value
+ */
 #define ECMA_INTEGER_NUMBER_MIN         -0x8000000
+/**
+ * Minimum integer number for an ecma value (shifted left with ECMA_DIRECT_SHIFT)
+ */
 #define ECMA_INTEGER_NUMBER_MIN_SHIFTED (-0x7fffffff - 1) /* -0x80000000 */
 #endif /* CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32 */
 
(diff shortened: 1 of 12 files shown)


akosthekiss:
Note: Doxygen has not been processing headers since #1540 .

Note 2: The list of files with still missing documentation (i.e., not fixed by this PR) is:
  - jerry-core/ecma/base/ecma-globals.h 
  - jerry-core/ecma/base/ecma-helpers.h 
  - jerry-core/ecma/operations/ecma-exceptions.h 
  - jerry-core/include/jerryscript-debugger-transport.h 
  - jerry-core/jcontext/jcontext.h 
  - jerry-core/parser/js/byte-code.h 
  - jerry-core/parser/js/common.h 
  - jerry-core/parser/js/js-lexer.h 
  - jerry-core/parser/js/js-parser-internal.h 
  - jerry-core/parser/regexp/re-parser.h 
  - jerry-core/vm/vm-stack.h
 
I'm happy to break them out into an issue to track the progress of their fixes, should the PR be accepted and the issue considered valid. (But I cannot volunteer to document them.)

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -752,14 +770,14 @@ typedef struct
...
-    /*
+    /**
      * Description of pseudo array objects.
      */
     struct
     {
       uint8_t type; /**< pseudo array type, e.g. Arguments, TypedArray*/
-      uint8_t extra_info; /**< extra infomations about the object.
-                            *  e.g. element_width_shift for typed arrays */
+      uint8_t extra_info; /**< extra infomation about the object.
```
zherczeg: infromation

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -994,21 +1014,28 @@ typedef double ecma_number_t;
  */
 #define ECMA_NUMBER_MINUS_ONE ((ecma_number_t) -1)
 
+#if CONFIG_ECMA_NUMBER_TYPE == CONFIG_ECMA_NUMBER_FLOAT32
 /**
- * Minimum positive and maximum value of ecma-number
+ * Number.MIN_VALUE (i.e., the smallest positive value of ecma-number)
+ *
+ * See also: ECMA_262 v5, 15.7.3.3
```
zherczeg: I don't really like these comment duplications. Is there a better way to do this?

jerry-core/include/jerryscript-compiler.h
```diff
@@ -73,61 +73,111 @@ void * __cdecl _alloca (size_t _Size);
 /*
  * Attributes
  */
+
+/**
+ * Function attribute to direct compiler to align function to given number of
```
zherczeg: I would remove these "to direct compiler" parts.

jerry-core/parser/js/common.h
```diff
@@ -57,21 +57,21 @@ typedef enum
 
 /* Flags for status_flags. */
 
-/* Local identifier (var, function arg). */
+/** Local identifier (var, function arg). */
```
zherczeg: Hm, would be better to convert these into an enum.

jerry-core/parser/js/js-parser.h
```diff
@@ -122,7 +122,7 @@ typedef enum
   PARSER_ERR_NON_STRICT_ARG_DEFINITION                /**< non-strict argument definition */
 } parser_error_t;
 
-/* Source code line counter type. */
+/** Source code line counter type. */
```
zherczeg: We should use the multiline form here.

jerry-core/ecma/base/ecma-globals.h
akosthekiss: information

jerry-core/parser/js/common.h
akosthekiss: I won't be pursuing that in this PR, which is solely about documentation fixes.

jerry-core/ecma/base/ecma-globals.h
akosthekiss: There are no easy ways and/or this is the way that's compatible with the rest of ecma-globals.h. (E.g., see the documentation of `ecma_number_t`, `ECMA_NUMBER_MAX_DIGITS`, `ECMA_NUMBER_SIGN_WIDTH`, `ECMA_NUMBER_BIASED_EXP_WIDTH`, and `ECMA_NUMBER_FRACTION_WIDTH`.) I've followed that approach here, too.

FYI, there may be one way to reduce duplication by using `@def MACRONAME(args)` doxygen syntax to break out documentation from before the macro definitions, but that might necessitate a bigger rewrite throughout the code base (to apply that approach everywhere where macros (or typedefs, or functions) are conditionally defined multiple times). Moreover, that approach should be discussed further as it splits the documentation from the documented entity, making maintenance somewhat harder. So, I'd have that separate from this PR.

(To restate: this PR tries to fix the problem in the Doxyfile with the least friction, fixing only the fallout that can be easily fixed, leaving the rest for follow-up.)

akosthekiss:
PR updated, actioned most of the review comments

jerry-core/ecma/base/ecma-globals.h
LaszloLango: You can use `@copydoc` or groups to avoid duplication in comments. Maybe my previous PR can help: https://github.com/jerryscript-project/jerryscript/pull/2324

akosthekiss: But you cannot copy documentation when an entity is defined multiple times because of conditional compilation, can you? I guess you cannot copydoc self.

```c
#if COND
/** documentation of X */
#define X
/** documentation of Y */
#define Y
#else
/** @copydoc X */   // <-- THIS SHOULDN'T WORK
#define X
/** @copydoc Y */   // <-- THIS SHOULDN'T WORK
#define Y
#endif
```

jerry-core/include/jerryscript-compiler.h
```diff
@@ -73,61 +73,110 @@ void * __cdecl _alloca (size_t _Size);
...
+ * Function attribute to trigger warning if function's caller doesn't use (e.g.,
+ * check) the return value.
+ */
 #ifndef JERRY_ATTR_WARN_UNUSED_RESULT
 #define JERRY_ATTR_WARN_UNUSED_RESULT
 #endif /* !JERRY_ATTR_WARN_UNUSED_RESULT */
 
 /*
  * Condition likeliness, unlikeliness
  */
```
zherczeg: Shall we keep this comment? It seems unnecessary now.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -142,7 +142,7 @@ typedef int32_t ecma_integer_value_t;
  */
 #define ECMA_DIRECT_SHIFT 4
 
-/* ECMA make simple value */
+/** ECMA make simple value */
```
zherczeg: I would turn this a 3 line form.

jerry-core/include/jerryscript-compiler.h
akosthekiss: It is/was only a grouping comment by I can remove it. But then I'll remove the other grouping comment ("Attributes"), too.

jerry-core/ecma/base/ecma-globals.h
LaszloLango: IHMO the default macro conditions should be documented. The Doxygen docs will be generated with the default config. BTW you can use a group comment before the `#if` which should work independently from the configuration. Although this can be useful in simple cases only (like your example in the prev. comment).

PR #2447, opened 2018-08-06
Implement the core of the map object.

zherczeg:
Heavily under construction. set/get/has and gc marking is working with Map. Next step is delete. WeakMap is there as well, but I realized the gc marking for WeakMaps can be too costly.


zherczeg:
Patch ready for review.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+
+      JERRY_ASSERT (!ecma_is_value_pointer (*item_p));
+      continue;
+    }
+
+    if (is_key)
+    {
+      if (JERRY_LIKELY (is_direct))
+      {
+        if (key_arg == item)
```
rerobika: These `if` statements can be merged.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+        {
+          break;
+        }
+      }
+      else if (key_str_p != NULL)
+      {
+        if (ecma_is_value_non_direct_string (item)
```
rerobika: Ditto.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+
+    if (last_key == ECMA_VALUE_ARRAY_HOLE)
+    {
+      last_key = item;
+    }
+    else
+    {
+      if (JERRY_LIKELY (is_direct))
+      {
+        if (key == last_key)
```
rerobika: These `if` statements can be merged.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+        {
+          return item_p - 1;
+        }
+      }
+      else if (key_str_p != NULL)
+      {
+        if (ecma_is_value_non_direct_string (last_key)
```
rerobika: Ditto.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+ *         Returned value must be freed with ecma_free_value.
+ */
+ecma_value_t
+ecma_op_map_set (ecma_value_t this_arg, /**< this argument */
+                 ecma_value_t key_arg, /**< key argument */
+                 ecma_value_t value_arg, /**< value argument */
+                 bool is_weak_map) /**< map is weak map */
+{
+  JERRY_UNUSED (key_arg);
+  JERRY_UNUSED (value_arg);
```
rerobika: Both `JERRY_UNUSED` are unnecessary.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+
+  if (value_p == NULL)
+  {
+    ecma_value_t *key_p = NULL;
+    ecma_map_object_chunk_t *last_chunk_p = ECMA_GET_POINTER (ecma_map_object_chunk_t,
+                                                              map_object_p->last_chunk_cp);
+
+    if (last_chunk_p != NULL)
+    {
+      if (last_chunk_p->items[2] == ECMA_VALUE_ARRAY_HOLE)
```
rerobika: These `if` statements can be merged.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+      }
+      else
+      {
+        last_chunk_p->items[ECMA_MAP_OBJECT_ITEM_COUNT] = ecma_make_pointer_value (new_chunk_p);
+      }
+
+      if (key_p == NULL)
+      {
+        JERRY_ASSERT (value_p == NULL);
+        key_p = new_chunk_p->items + 0;
```
rerobika: Does the `+0` is needed?

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -0,0 +1,757 @@
...
+        value_p = new_chunk_p->items + 1;
+      }
+      else
+      {
+        value_p = new_chunk_p->items + 0;
```
rerobika: Ditto?

jerry-core/ecma/operations/ecma-map-object.c
zherczeg: I don't think so, because if `is_direct` is true, but the `key != last_key`, we fall to the `ecma_is_value_float_number (last_key)` which we shouldn't.

jerry-core/ecma/operations/ecma-map-object.c
zherczeg: Nope, but it does not hurt as well. I wanted the emphasize the zero value here, but it can be removed.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -207,6 +210,87 @@ ecma_gc_mark_property (ecma_property_pair_t *property_pair_p, /**< property pair
...
+
+  ecma_value_t *item_p = ECMA_GET_NON_NULL_POINTER (ecma_map_object_chunk_t, first_chunk_cp)->items;
+
+  while (true)
+  {
+    ecma_value_t item = *item_p++;
+
+    if (!ecma_is_value_pointer (item))
+    {
+      if (ecma_is_value_object (item))
```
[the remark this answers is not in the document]
zherczeg: We cannot because we would enter into the `else` branch when value is pointer.

zherczeg:

PR #2455, opened 2018-08-08
Don't use a shell variable to store error output of doxygen

akosthekiss:
Storing the error output of doxygen into a shell variable first and
echoing it later is error prone. In case of a long error output,
the whole shell will crash before being able to print the doxygen
diagnostics at all. This patch rewrites the check script to tee the
diagnostics (pipe them to console and store them to a file at the
same time) and check the output file size at the end.

### DIFF
--- a/tools/check-doxygen.sh
+++ b/tools/check-doxygen.sh
@@ -14,12 +14,12 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-echo -n "Generating documentation with doxygen ..."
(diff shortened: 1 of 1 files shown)


akosthekiss:
Note: I've actually hit this issue when a configuration caused the version of doxygen available on Travis CI to print long (probably infinite) diagnostics. But the diagnostics did _not_ appear in the logs as the whole check-doxygen.sh crashed with only an error code signalling some trouble. Seeing the actual diagnostics would have helped to find the root of the problem (as my local doxgen version was newer and did not exhibit the issue).

PR #2458, opened 2018-08-08
Make string constants `const char * const`

akosthekiss:

### DIFF
--- a/jerry-core/api/jerry-snapshot.c
+++ b/jerry-core/api/jerry-snapshot.c
@@ -151,7 +151,7 @@ snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled
 
   if (globals_p->snapshot_buffer_write_offset > JERRY_SNAPSHOT_MAXIMUM_WRITE_OFFSET)
   {
-    const char *error_message_p = "Maximum snapshot size reached.";
+    const char * const error_message_p = "Maximum snapshot size reached.";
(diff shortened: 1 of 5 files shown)


akosthekiss:
Binary sizes (bytes)
29f6ffc35b:130908
f57cd9b75d:130820

88 bytes gain

jerry-core/api/jerry-snapshot.c
```diff
@@ -151,7 +151,7 @@ snapshot_add_compiled_code (ecma_compiled_code_t *compiled_code_p, /**< compiled
 
   if (globals_p->snapshot_buffer_write_offset > JERRY_SNAPSHOT_MAXIMUM_WRITE_OFFSET)
   {
-    const char *error_message_p = "Maximum snapshot size reached.";
+    static const char * const error_message_p = "Maximum snapshot size reached.";
```
zherczeg: While I understand the point of `const`, what is the point of static? It makes a pointer on the data space rather than directly loading the pointer (extra indirection).

akosthekiss: According to measurements, approaches with and without `static` give the exact same results. So, I'll just drop `static`. The only place I'll keep them is at arrays, where static guarantees that the array will be initialised only once.

BTW, in the update I'll also _remove_ `static` from before to string constants.

akosthekiss:
PR updated. Results unchanged:

Binary sizes (bytes)
17eb78aa63:130908
2bdf535c30:130820

PR #2459, opened 2018-08-09
Modify header file contains.

yangfasheng:
Using the custom header file "config.h", the header file directory does not contain the "jerry-core" directory.

JerryScript-DCO-1.0-Signed-off-by: YangFaSheng yangfasheng@live.com

### DIFF
--- a/jerry-core/lit/lit-char-helpers.c
+++ b/jerry-core/lit/lit-char-helpers.c
@@ -14,7 +14,7 @@
  */
 
 #include "lit-char-helpers.h"
-#include "lit/lit-unicode-ranges.inc.h"
+#include "lit-unicode-ranges.inc.h"
 #include "lit-strings.h"
 
 #ifndef CONFIG_DISABLE_UNICODE_CASE_CONVERSION


akosthekiss:
Nice catch. Thanks for the patch.

However, could you please read https://github.com/jerryscript-project/jerryscript/wiki/JerryScript-Developer's-Certificate-of-Origin and update your commit accordingly (with `git commit --amend`, adding `JerryScript-DCO-1.0-Signed-off-by: ...` as the last line of the commit message?)

akosthekiss:
@yangfasheng I see you've edited the PR's comment but that's not the commit. You have to edit the git commit's message locally on your computer (using `git commit --amend`) and then push the new commit to your github repo (using `git push -f YOURREPONAME master` -- I can only guess that you see your jerryscript repo on github as `origin`, so you may try that for `YOURREPONAME`; if that doesn't work, you have to check what remotes you have set using `git remote -v`). Pushing the updated commit to your repo will update the PR at the same time.

Two notes:
- Working in master is not the best practice. You may want to look up docs on git branching model. Creating and disposing branches are cheap in git and help a lot when the `master` of your upstream project progresses further.
- Your newly added text in the comment made me suspicious a bit: what do you mean by "Using the custom header file "config.h""? If that means `jerry-core/config.h`, then, well, (a) it's not custom, and (b) you should not be including it or any other header from the directory `jerry-core` except for the public headers in `jerry-core/include`.

akosthekiss:
And one more thing: please, check your commit on your local computer using `tools/check-signed-off.sh` before pushing to github. If that passes then you are fine, but if the script prints errors, then Travis CI will fail too and the PR will _not_ be mergeable.

PR #2461, opened 2018-08-09
Split string-sending debugger API into output- and log-sending functions

akosthekiss:
This helps to avoid the use of non-public headers and
protocol-internal constants in external code (e.g., in jerry-port
and jerry-ext).

The patch also cleans up the necessary includes in jerry-core public
headers, and the include order in jerry-port/default public headers.

### DIFF
--- a/jerry-core/debugger/debugger.h
+++ b/jerry-core/debugger/debugger.h
@@ -212,6 +212,10 @@ typedef enum
 
 /**
  * Subtypes of output_result.
+ *
+ * Note:
+ *      This enum has to be kept in sync with jerry_log_level_t with an offset
+ *      of +2.
  */
 typedef enum
 {
--- a/docs/07.DEBUGGER.md
+++ b/docs/07.DEBUGGER.md
@@ -359,20 +359,51 @@ Sends the program's output to the debugger client.
 **Prototype**
 
 ```c
-  void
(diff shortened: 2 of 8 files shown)


jerry-core/debugger/debugger.h
```diff
@@ -212,6 +212,10 @@ typedef enum
 
 /**
  * Subtypes of output_result.
+ *
+ * Note:
+ *      This enum has to be kept in sync with jerry_log_level_t with an offset
+ *      of +2.
```
akosthekiss: Note: it would be even simpler to get rid of `jerry_debugger_output_subtype_t` completely and use `jerry_log_level_t` as sub-type. But that would probably necessitate the introduction of a new message type `JERRY_DEBUGGER_LOG_RESULT` (i.e., `JERRY_DEBUGGER_OUTPUT_RESULT` + `JERRY_DEBUGGER_NO_SUBTYPE` to be used for user output and `JERRY_DEBUGGER_LOG_RESULT` + `jerry_log_level_t` to be used for log messages.). Or the situation may be even more intricate, as I see a `JERRY_DEBUGGER_OUTPUT_RESULT_END` message type as well. So, I rather didn't dare to touch the protocol itself, only the public API.

zherczeg: I would keep the JERRY_DEBUGGER_OUTPUT_ as private myself.

PR #2471, opened 2018-08-14
Update debugger documentation

akosthekiss:
The documentation has not been updated when the HTML client got
removed nor when the new transport layer replaced the previous
debug server initialization process.

### DIFF
--- a/docs/07.DEBUGGER.md
+++ b/docs/07.DEBUGGER.md
@@ -3,11 +3,11 @@
 JerryScript provides a remote debugger which allows debugging
 JavaScript programs. The debugger has two main components:
 a server which is part of the JerryScript binary and a
-separate client application. Currently two debugger clients
-are available in the /jerry-debugger subdirectory: an HTML
-and a Python application. These simple applications demonstrate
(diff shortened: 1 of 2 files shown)


docs/07.DEBUGGER.md
```diff
@@ -65,9 +66,10 @@ All available commands of the client can be queried by the
 
 ## Integrating debugger support into applications using JerryScript
 
-The debugger can be enabled by calling the `jerry_debugger_init (uint16_t port)`
-function after the `jerry_init ()` function. It initializes the debugger
-and blocks until a client connects.
+The debugger can be enabled by calling the `jerryx_debugger_after_connect
+(jerryx_debugger_tcp_create (debug_port) && jerryx_debugger_ws_create ())`
+after the `jerry_init ()` function. It initializes the debugger and blocks
+until a client connects.
```
zherczeg: It would be good to emphasize that users can choose from the available transport layer implementations provided by their own environments.

akosthekiss:
PR updated. I've added some extra words on transport layer customisability.

PR #2474, opened 2018-08-15
Change literal status flags to enum.

zherczeg:
Also fix a typo.

### DIFF
--- a/jerry-core/parser/js/common.h
+++ b/jerry-core/parser/js/common.h
@@ -55,24 +55,22 @@ typedef enum
                                          used by the byte code generator. */
 } lexer_literal_type_t;
 
-/* Flags for status_flags. */
-
-/** Local identifier (var, function arg). */
-#define LEXER_FLAG_VAR 0x01
-/** This local identifier cannot be stored in register. */
-#define LEXER_FLAG_NO_REG_STORE 0x02
-/** This local identifier is initialized with a value. */
-#define LEXER_FLAG_INITIALIZED 0x04
-/** This local identifier has a reference to the function itself. */
-#define LEXER_FLAG_FUNCTION_NAME 0x08
-/** This local identifier is a function argument. */
-#define LEXER_FLAG_FUNCTION_ARGUMENT 0x10
-/** This local identifier is not used in the current context. */
-#define LEXER_FLAG_UNUSED_IDENT 0x20
-/** No space is allocated for this character literal. */
-#define LEXER_FLAG_SOURCE_PTR 0x40
-/** Initialize this variable after the byte code is freed. */
-#define LEXER_FLAG_LATE_INIT 0x80
+/**
+ * Flag bits for status_flags member of lexer_literal_t.
+ */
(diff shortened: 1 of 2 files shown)


jerry-core/parser/js/js-parser-limits.h
```diff
@@ -77,7 +77,7 @@
 #endif /* !PARSER_MAXIMUM_CODE_SIZE */
 
 /**
- * Maximum number of values pushed onto the stack by a function.
+ * Maximum number of values pushed onto the stack by a function.
```
zherczeg: There is a 'C2 A0' character here, which a "NO-BREAK SPACE'. Change it to a normal space.

jerry-core/parser/js/common.h
```diff
@@ -55,24 +55,22 @@ typedef enum
...
-/** No space is allocated for this character literal. */
-#define LEXER_FLAG_SOURCE_PTR 0x40
-/** Initialize this variable after the byte code is freed. */
-#define LEXER_FLAG_LATE_INIT 0x80
+/**
+ * Flag bits for status_flags member of lexer_literal_t.
+ */
+typedef enum
+{
+  LEXER_FLAG_VAR = (1 << 0), /** local identifier (var, function arg) */
```
akosthekiss: All the enum value comments need to be `/**<` as they are after the entity they are documenting.

zherczeg:
Thank you for the review. Patch updated.

jerry-core/parser/js/common.h
```diff
@@ -55,24 +55,22 @@ typedef enum
...
+  LEXER_FLAG_NO_REG_STORE = (1 << 1), /**< this local identifier cannot be stored in register */
+  LEXER_FLAG_INITIALIZED = (1 << 2), /**< this local identifier is initialized with a value */
+  LEXER_FLAG_FUNCTION_NAME = (1 << 3), /**< this local identifier has a reference to the function itself */
+  LEXER_FLAG_FUNCTION_ARGUMENT = (1 << 4), /**< this local identifier is a function argument */
+  LEXER_FLAG_UNUSED_IDENT = (1 << 5), /**< this identifier is referenced by sub-functions,
+                                       *   but not referenced by the currently parsed function */
+  LEXER_FLAG_SOURCE_PTR = (1 << 6), /**< the literal is directly referenced in the source code
+                                     *   (no need to allocate memory) */
+  LEXER_FLAG_LATE_INIT = (1 << 7), /**< initialize this variable after the byte code is freed */
+} parser_lexer_literal_status_flags_t;
```
akosthekiss: Strange name for this type. It is used in `lexer_literal_t`, all the enum values have a `LEXER_` prefix, yet this type has a `parser_` prefix. I'd suggest going for `lexer_literal_status_flags_t` only.

zherczeg: Good catch. I haven't noticed that.

AnthonyCalandra:
LGTM (informal)

PR #2481, opened 2018-08-21
Implement computed properties for object literals.

zherczeg:


zherczeg:
Patch is doen, and ready for review.

jerry-core/vm/vm.h
```diff
@@ -121,6 +121,9 @@ typedef enum
   VM_OC_PUSH_LIT_NEG_BYTE,       /**< push literal and number between -1 and -256 */
   VM_OC_PUSH_OBJECT,             /**< push object */
   VM_OC_SET_PROPERTY,            /**< set property */
+#ifndef CONFIG_DISABLE_ES2015_OBJECT_LITERAL
```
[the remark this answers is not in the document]
zherczeg: This is reworked in the execution engine. Now the byte codes are translated to a valid VM opcode or a fatal error. This is much safer now, since we at least get an error rather than a random crash if we execute a snapshot with invalid byte codes. Normally this should be detected early, but we have a safety net now.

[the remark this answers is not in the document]
LaszloLango: OK

zherczeg:
Come to think of snapshots, I also increased the snapshot version number.

tests/jerry/es2015/object-computed.js
```diff
@@ -0,0 +1,103 @@
...
+
+function fxy() {
+  return "xy";
+}
+
+class C {
+  [fxy()] () {
+    return 6;
+  }
+
```
rerobika: Please add test cases for:

```js
class C {
  ["const" + "ructor"] { // 1.
    this.a = 5;
  }

  get ["constructor"]() { // 2.
    return this.a;
  }
  static ["prototype"] () { // 3.
    return 10;
  }
}
```

In the first case it must be a normal class method instead of the real constructor of the class.

The second one is a bit similar. In normal case the `get constructor() {}` is not allowed for classes (the real `constructor` cannot be an accessor), but as a computed property name it must be a simple accessor with `constructor` name.

Finally, the third one must throw a `TypeError` in runtime as the same way as `static prototype () {}` throws a `SyntaxError` during the script parsing.

jerry-core/vm/vm.c
```diff
@@ -1108,10 +1118,24 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
         case VM_OC_SET_SETTER:
         {
-          JERRY_ASSERT (byte_code_start_p[0] == CBC_EXT_OPCODE);
+          if (!ecma_is_value_string (left_value))
+          {
+            result = ecma_op_to_string (left_value);
+
+            if (ECMA_IS_VALUE_ERROR (result))
+            {
+              goto error;
```
rerobika: If the `result` is an error, should not the `left_value` be freed?

jerry-core/config.h
```diff
@@ -39,6 +39,7 @@
 # define CONFIG_DISABLE_ES2015_ARROW_FUNCTION
 # define CONFIG_DISABLE_ES2015_CLASS
 # define CONFIG_DISABLE_ES2015_BUILTIN
+# define CONFIG_DISABLE_ES2015_OBJECT_LITERAL
```
rerobika: It is a new feature, so please update the [profile](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/profiles/README.md) documentation.

jerry-core/vm/vm.c
rerobika: OK, thanks for clarifying it.

zherczeg:
Thank you for the review. Some bigger changes applied, @LaszloLango you might want to take another look (I changed literal to initializer, because es6 uses this term and reworked property setting in the vm). New tests added as well.

LaszloLango:
@zherczeg Ok, I will take a look on the changes.

LaszloLango:
still LGTM

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -356,8 +357,14 @@ parser_append_object_literal_item (parser_context_t *context_p, /**< context */
     context_p->stack_top_uint8 = PARSER_OBJECT_PROPERTY_BOTH_ACCESSORS;
   }
 } /* parser_append_object_literal_item */
+#endif /* CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
 
 #ifndef CONFIG_DISABLE_ES2015_CLASS
+
+#ifdef CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER
+#error "Class support requires ES2015 object literal support"
+#endif /* CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
```
AnthonyCalandra: If we're making this a requirement, should maybe this code be moved to config.h?

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -694,12 +780,14 @@ parser_parse_object_literal (parser_context_t *context_p) /**< context */
...
   }
 
+#ifdef CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER
   while (context_p->stack_top_uint8 != PARSER_OBJECT_PROPERTY_START)
   {
     parser_stack_pop (context_p, NULL, 3);
   }
 
   parser_stack_pop_uint8 (context_p);
+#endif /* !CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */
```
AnthonyCalandra: Shouldn't this be `/* CONFIG_DISABLE_ES2015_OBJECT_INITIALIZER */`?

jerry-core/parser/js/js-parser-expr.c
zherczeg: I searched through the code base and only a few `#error` is there. It seems there is a configuration check in the normal code base:
https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/jcontext/jcontext.h#L194

But config.h also have a check.
https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/config.h#L81

The truth is I don't know we have a policy for that. It depends config.h is mandatory, or user supplied. Anybody knows this?

zherczeg: (In most projects I know config.h is auto generated, and contains no dependency checks. In JerryScript config.h is not auto generated.)

PR #2482, opened 2018-08-22
Added missing documentation of JerryScript instances to the API reference

LaszloLango:


docs/02.API-REFERENCE.md
```diff
@@ -4973,6 +5007,97 @@ main (void)
...
+- `heap_size` - requested heap size of the JerryScript instance
+- `alloc` - function for allocation
+- `cb_data_p` - user data
+- return value
+  - pointer to the newly created JerryScript instance if success
+  - NULL otherwise.
+
+**Example**
+
+```c
```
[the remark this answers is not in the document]
LaszloLango: I wanted to do, but I ran into linking problems. I did not want to add `-pthread` to all doctest build. AFAIK doctest comments does not support additional build options.

jerry-core/include/jerryscript-core.h
```diff
@@ -529,7 +529,7 @@ jerry_value_t jerry_resolve_or_reject_promise (jerry_value_t promise, jerry_valu
 bool jerry_is_valid_utf8_string (const jerry_char_t *utf8_buf_p, jerry_size_t buf_size);
 bool jerry_is_valid_cesu8_string (const jerry_char_t *cesu8_buf_p, jerry_size_t buf_size);
 
-/*
+/**
  * External context functions.
```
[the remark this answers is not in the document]
LaszloLango: IMHO it is needed to make it consistent.

docs/02.API-REFERENCE.md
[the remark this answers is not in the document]
akosthekiss: That's correct, linking wont work. But you can instruct doctest to compile only (there are examples for that in the existing documentation). Compilation can already catch lots of potential errors.

jerry-core/include/jerryscript-core.h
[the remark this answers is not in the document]
akosthekiss: Well, if we want to be consistent _and_ correct, we should remove the extra asterisks from everywhere in this header, IMO. `/**` is for doctesting _one_ entity, not for marking up groups.

LaszloLango:
@akosthekiss I've updated the PR. Please check.

docs/02.API-REFERENCE.md
```diff
@@ -4973,6 +5001,99 @@ main (void)
...
+#include <stdlib.h>
+#include <pthread.h>
+
+#include "jerryscript.h"
+#include "jerryscript-port.h"
+
+/* A different Thread Local Storage variable for each jerry instance. */
+__thread jerry_instance_t* tls_instance;
```
akosthekiss: the asterisk seems to stick to the wrong side

docs/02.API-REFERENCE.md
```diff
@@ -4973,6 +5001,99 @@ main (void)
...
+  /* Returns the instance assigned to the thread. */
+  return tls_instance;
+}
+
+/* Allocate JerryScript heap for each thread. */
+static void *
+instance_alloc_fn (size_t size, void *cb_data)
+{
+  (void) cb_data;
+  return malloc(size);
```
akosthekiss: space is missing from between malloc and (

docs/02.API-REFERENCE.md
```diff
@@ -4973,6 +5001,99 @@ main (void)
...
+
+#define NUM_OF_THREADS 8
+
+int
+main (void)
+{
+  pthread_t threads[NUM_OF_THREADS];
+
+  /* Create the threads. */
+  for (int i = 0; i < NUM_OF_THREADS; i++) {
```
akosthekiss: open brace should be on its own line

docs/02.API-REFERENCE.md
```diff
@@ -4973,6 +5001,99 @@ main (void)
...
+    pthread_create (&threads[i], NULL, thread_function, (void *) (intptr_t) i);
+  }
+
+  /* Wait for the threads to complete, and release their resources. */
+  for (int i = 0; i < NUM_OF_THREADS; i++) {
```
akosthekiss: ditto

docs/02.API-REFERENCE.md
akosthekiss: and one more question: where will the here-allocated memory be freed?

docs/02.API-REFERENCE.md
```diff
@@ -4973,6 +5001,99 @@ main (void)
...
+  }
+
+  return 0;
+}
+```
+
+**See also**
+
+- [jerry_instance_t](#jerry_instance_t)
+- [jerry_instance_alloc_t](#jerry_instance_alloc_t)
```
akosthekiss: I don't think we have an example for cross referencing doc MD pages, but somehow the jerry_port_get_instance function should be mentioned here (even if without a link). Right now no text mentions it, it only appears in the example out of the blue, without any hints why it is needed.

LaszloLango: I was wondering on this too, but could not find a good solution which works both with MD and the generated `jerryscript.net` page.

PR #2483, opened 2018-08-23
Fix logging issues in the debugger.

zherczeg:
- Properly handle logging during transport close
 - Properly display data during parse

Furthermore hide global functions of the debugger.


zherczeg:
Good point. The internal function hiding is done in #2484.

akosthekiss:
Isn't this now pending on #2484? The current version of this PR still contains rewrites that are happening in the referenced PR.

zherczeg:
This should be landed after that PR.

LaszloLango:
#2484 is landed, please rebase the PR.

zherczeg:
Patch updated. Now debugger print works correctly with \0 characters.

jerry-ext/handler/handler-print.c
```diff
@@ -56,24 +56,24 @@ jerryx_handler_print (const jerry_value_t func_obj_val, /**< function object */
...
-      if (arg_index != 0)
-      {
-        jerryx_port_handler_print_char (' ');
-      }
-
+      const jerry_size_t substr_max_size = 256;
       jerry_size_t substr_size;
       jerry_length_t substr_pos = 0;
-      jerry_char_t substr_buf[256];
+      jerry_char_t substr_buf[substr_max_size];
```
LaszloLango: Plase use `JERRY_VLA` to fix the windows build.

zherczeg: But this is a fixed size array :(

LaszloLango: What about using #define instead of a const variable?

PR #2485, opened 2018-08-27
Introducing internal properties.

zherczeg:
Native handle and pointer are moved to internal properties.


robertsipka:
@zherczeg : Please rebase it to master.

PR #2496, opened 2018-08-29
Remove deprecated native handle support.

zherczeg:


zherczeg:
It turned out that there are no unit tests for the old api.

PR #2498, opened 2018-08-30
Merge lcache into context

akosthekiss:
It is superfluous to maintain multiple globals when the whole
reason of context is to keep them in a single place. It also
simplifies initialization and external context creation a bit.

Also removed unused lcache header includes.

JerryScript-DCO-1.0-Signed-off-by: Akos Kiss akiss@inf.u-szeged.hu

PR #2500, opened 2018-08-31
Reorganize heap context related elements

akosthekiss:
- Moved global context's `JMEM_HEAP_SIZE` to jcontext.h as external
  context's heap size is declared there, too.
- Moved the assert on `jmem_heap_t` vs `JMEM_HEAP_SIZE` to jcontext.c,
  as both entities are declared in the corresponding header.
- Removed superfluous checks on `JMEM_HEAP_SIZE` as it is not a
  publicly configurable macro (opposed to `CONFIG_MEM_HEAP_AREA_SIZE`).
- Ensured that all definitions of and references to `jmem_heap_t`,
  `JERRY_HEAP_CONTEXT`, `JERRY_HEAP_SIZE`, and `JERRY_HEAP_AREA_SIZE`
  are guarded by `#ifndef JERRY_SYSTEM_ALLOCATOR`, as they are
  meaningless if the system allocator is used.

The commit also contains some stylistic changes in jcontext.h

### DIFF
--- a/jerry-core/jcontext/jcontext.c
+++ b/jerry-core/jcontext/jcontext.c
@@ -20,10 +20,19 @@
  */
 
 #ifndef JERRY_ENABLE_EXTERNAL_CONTEXT
+
 /**
  * Global context.
  */
 jerry_context_t jerry_global_context;
+
+#ifndef JERRY_SYSTEM_ALLOCATOR
+
+/**
+ * Check size of heap is corresponding to configuration
+ */
+JERRY_STATIC_ASSERT (sizeof (jmem_heap_t) <= JMEM_HEAP_SIZE,
+                     size_of_mem_heap_must_be_less_than_or_equal_to_JMEM_HEAP_SIZE);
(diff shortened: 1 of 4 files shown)


akosthekiss:
Part of the context rework PR series, follow-up to #2498.

jerry-core/jcontext/jcontext.c
```diff
@@ -20,11 +20,20 @@
...
  */
 jerry_context_t jerry_global_context;
 
+#ifndef JERRY_SYSTEM_ALLOCATOR
+
+/**
+ * Check size of heap is corresponding to configuration
+ */
+JERRY_STATIC_ASSERT (sizeof (jmem_heap_t) <= JMEM_HEAP_SIZE,
+                     size_of_mem_heap_must_be_less_than_or_equal_to_JMEM_HEAP_SIZE);
```
zherczeg: I think we use lowercase everywhere.

PR #2501, opened 2018-08-31
Merge instance into context

akosthekiss:
There was quite some confusion about terminology around instances
and contexts. All the docs mentioned external contexts but
functions and types were referring to instances, and the relation
between these two concepts were not clear. This commit keeps
(external) context as the only surviving concept.

### DIFF
--- a/jerry-core/jcontext/jcontext.h
+++ b/jerry-core/jcontext/jcontext.h
@@ -14,7 +14,7 @@
  */
 
 /*
- * Memory context for JerryScript
+ * Engine context for JerryScript
  */
 #ifndef JCONTEXT_H
 #define JCONTEXT_H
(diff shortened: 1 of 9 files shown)


akosthekiss:
Part of the context rework PR series, follow-up to #2500 .

jerry-core/jcontext/jcontext.h
```diff
@@ -56,9 +72,17 @@ typedef struct jerry_context_data_header
...
  * all global variables for Jerry
  */
-typedef struct
+struct jerry_context_t
 {
-  /* Update JERRY_CONTEXT_FIRST_MEMBER if the first member changes */
+  /* The value of external context members must be preserved across initializations and cleanups. */
+#ifdef JERRY_ENABLE_EXTERNAL_CONTEXT
+#ifndef JERRY_SYSTEM_ALLOCATOR
+  jmem_heap_t *heap_p; /**< point to the heap aligned to JMEM_ALIGNMENT. */
```
zherczeg: I would put this before or after the context, probably after would be better, since the starting offset would be fixed. Then we don't need this pointer.

jerry-core/jcontext/jcontext.h
```diff
@@ -56,9 +72,17 @@ typedef struct jerry_context_data_header
  * The purpose of this header is storing
  * all global variables for Jerry
  */
-typedef struct
+struct jerry_context_t
```
zherczeg: Why do you remove the typedef?

akosthekiss: It is typedef'd in jerryscript-core.h. A typedef cannot be declared twice. `jerry_context_t` is taking over the place of `jerry_instance_t`, so it has to be opaquely forward declared in the public API header. But that means that only the struct can be declared here.

Note: `jerry_instance_t` was handled exactly the same way.

jerry-core/jcontext/jcontext.h
akosthekiss: The instance-based approach also contained this pointer, at the end of the instance struct, even if the heap always started right after the instance. I'd keep this pointer as it allows for future developments, e.g., one that allows placing context structures and heaps in different memory regions (which is already available for the builtin global context approach but is not available for external contexts right now).

akosthekiss:
@zherczeg Thanks for the review, your comments have been answered in-line. Please, let me know if you have further change requests (or still maintain the current ones).

PR #2503, opened 2018-09-01
Promote dynamic memory management from debugger transport to core API

### LINKED ISSUE #1805: Expose dynamic memory management via public API
It's been an issue several times that users of JerryScript don't have access to dynamic memory management (in the general case). jerry-libc doesn't provide malloc/free, partially because jerry-core doesn't necessarily need them. But that has some adverse effects on use cases:
- jerry-main cannot use dynamic memory for file input reading, or for command line argument handling,
- the default jobqueue port implementation has to access internal jmem functions to allocate memory for job queue items (in its current form -- until #1804 gets merged, if ever), and
- the recent context extension #1802 also needs some dynamic memory to work.

Wouldn't a pair of public API functions (`void *jerry_alloc ()` / `void jerry_free (void *)`) solve these issues? It is of course a somewhat risky thing to allow, because that may easily eat away the jerry heap. But is allowing access to jmem internals any better?

akosthekiss:
Under the cover of the debugger transport layer, allocation on the
engine's heap has been made available to the public. As there are
actually no restrictions on what the allocated memory can be used
for, the memory management functions better fit in the core part
of the API.

Closes #1805

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -4875,6 +4875,52 @@ main (void)
 - [jerry_substring_to_char_buffer](#jerry_substring_to_char_buffer)
 
 
+# Dynamic memory management functions
+
+## jerry_heap_alloc
+
+**Summary**
+
+Allocate memory on the engine's heap.
+
+*Note*: This function may take away memory from the executed JavaScript code.
+If any other dynamic memory allocation API is available (e.g., libc malloc), it
+should be used instead.
+
+**Prototype**
+
+```c
+void *jerry_heap_alloc (size_t size);
+```
+
+- `size`: size of the memory block.
+- return value: non-NULL pointer, if the memory is successfully allocated,
+                NULL otherwise.
+
+**See also**
+
+- [jerry_heap_free](#jerry_heap_free)
+
+## jerry_heap_free
+
+**Summary**
+
+Free memory allocated on the engine's heap.
+
+**Prototype**
+
+```c
+void jerry_heap_free (void *mem_p, size_t size);
+```
+
+- `mem_p`: value returned by `jerry_heap_alloc`.
+- `size`: same size as passed to `jerry_heap_alloc`.
+
+**See also**
+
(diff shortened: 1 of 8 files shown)


zherczeg:
I am not sure this is a good idea. The problem is that JerryScript allocator is specialized for JerryScript, and can be very slow for generic random allocations. Also would trigger a lot of GC when it is used frequently. Promoting this feature would be misleading.

akosthekiss:
How it is any more misleading than it is now? The alloc/free API is already exposed.  And it seems that it has been exposed exactly for the same reasons that were hypothesised in the old issue that suggested their introduction: there may be components that may need dynamic memory but cannot rely on the presence of libc `malloc`/`free`, like jerry-ext. And there it came, the tcp/ws transport layer implementation in jerry-ext, which needed dynamic memory, so a pair of `xxx_malloc`/`xxx_free` appeared. Why does this xxx have to be `jerry_debugger_transport`? I don't see anything debugging-specific in them.

zherczeg:
This is not a technical problem. We expect that only those will use the current API who creates a debugger transport layer, and we can expect them to have knowledge about the engine. Furthermore if debugger is disabled the malloc is not exposed, so the user application who uses this will fail.

Perhaps the name should be something which emphasize that this is not for general use. E.g. `jerry_internal_alloc_for_experienced_users`.

zherczeg:
The debugger api is not part of the core api.

akosthekiss:
It is a public API of the engine.

akosthekiss:
Note: short and meaningless answer intended as a response to a short and meaningless comment. I've given you a 5 paragraphs long explanation / justification, which expresses my view on the issue. All I got back is 10 words, not even clear which part of my comment they are reacting on.

zherczeg:
What about removing the transport malloc overall then?

akosthekiss:
PR updated. I've made the function names a *bit* longer to emphasise that allocation is happening on engine heap and added a note that it should only be used as a last resort.

PR #2505, opened 2018-09-03
Fix run-test-suite

Yuyupo:
You could not interrupt the test with Ctrl+C.


tools/runners/run-test-suite.sh
```diff
@@ -41,6 +41,11 @@ TERM_NORMAL="\033[0m"
 TERM_RED="\033[1;31m"
 TERM_GREEN="\033[1;32m"
 
+trap stop SIGINT
+stop() {
+    exit
+}
```
akosthekiss: Is a handler function needed for this? Isn't `trap exit SIGINT` or `trap 'exit' SIGINT` enough?

PR #2507, opened 2018-09-03
Improve API of literal save and the snapshot (command line) tool

LaszloLango:
Removed 'jerry_parse_and_save_literals' and introduced
'jerry_get_literals_from_snapshot' instead which works
on snapshot buffers rather than source code. Added literal
saving feature to snapshot merge in the snapshot command
line tool. Also added missing 'jerry_cleanup()' calls to the
snapshot tool. Improved the console messages of the snapshot
tool.

Based on previous work of Tamas Zakor <ztamas@inf.u-szeged.hu>


docs/02.API-REFERENCE.md
```diff
@@ -5303,30 +5303,28 @@ main (void)
...
 
 
-## jerry_parse_and_save_literals
+## jerry_get_literals_from_snapshot
 
 **Summary**
 
-Collect the used literals from the given source code and save them into a specific file in a list or C format.
-These literals are generated by the parser, they are valid identifiers and none of them are magic string.
+Collect the used literals from the given snapshot and save them into a buffer in a list or C format.
```
akosthekiss: ... in list or C format.

docs/02.API-REFERENCE.md
```diff
@@ -5303,30 +5303,28 @@ main (void)
...
+None of these literals are magic string. In C format only valid identifiers are collected.
```
akosthekiss: ... are magic strings.

docs/02.API-REFERENCE.md
```diff
@@ -5348,14 +5346,25 @@ main (void)
...
   const jerry_char_t *code_for_literal_save_p = (const jerry_char_t *) "var obj = { a:'aa', bb:'Bb' }";
 
-  size_t literal_sizes = jerry_parse_and_save_literals (code_for_literal_save_p,
-                                                        strlen ((const char *) code_for_literal_save_p),
-                                                        false,
-                                                        save_literal_buffer,
-                                                        sizeof (save_literal_buffer) / sizeof (uint32_t),
-                                                        true);
+  jerry_value_t generate_result;
+  generate_result = jerry_generate_snapshot (NULL,
```
akosthekiss: Is it needed to split this into two statements? I've checked, `jerry_value_t generate_result = jerry_generate_snapshot (...` fits in the line length limit.

jerry-core/api/jerry-snapshot.c
```diff
@@ -1582,48 +1582,42 @@ ecma_string_is_valid_identifier (const ecma_string_t *string_p)
...
 size_t
-jerry_parse_and_save_literals (const jerry_char_t *source_p, /**< script source */
-                               size_t source_size, /**< script source size */
-                               bool is_strict, /**< strict mode */
-                               uint32_t *buffer_p, /**< [out] buffer to save literals to */
-                               size_t buffer_size, /**< the buffer's size */
-                               bool is_c_format) /**< format-flag */
+jerry_get_literals_from_snapshot (const uint32_t *snapshot_p, /**< input snapshot buffer */
+                                  size_t snapshot_size, /**< size of the input snapshot buffer */
+                                  uint32_t *lit_buf_p, /**< [out] buffer to save literals to */
```
akosthekiss: Why is `lit_buf_p` an `uint32_t *` if it stores a string? It will be cast to `uint8_t *` in the code anyway. (Question nr.2: why not to `char *` or `jerry_char_t *`?)

jerry-main/main-unix-snapshot.c
```diff
@@ -391,13 +391,15 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
...
   if (literals_file_name_p != NULL && !is_import_literals)
   {
-    const size_t literal_buffer_size = jerry_parse_and_save_literals ((jerry_char_t *) source_p,
-                                                                      source_length,
-                                                                      false,
-                                                                      output_buffer,
-                                                                      sizeof (output_buffer) / sizeof (uint32_t),
-                                                                      is_save_literals_mode_in_c_format);
-    if (literal_buffer_size == 0)
+    uint32_t *lit_out_buf_p = output_buffer + snapshot_size;
```
akosthekiss: `output_buffer` is an `uint32_t` array, `snapshot_size` seems to be a expressed in bytes. Pointer arithmetics will cause the addition to be off by a factor of 4 than expected.

jerry-main/main-unix-snapshot.c
```diff
@@ -391,13 +391,15 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
...
+    size_t lit_out_buf_sz = (sizeof (output_buffer) / sizeof (uint32_t)) - snapshot_size;
```
akosthekiss: This expression also seems to be incorrect. `sizeof (output_buffer)` is already in bytes, isn't it?

akosthekiss: Extra note / question: What ensures that the output of `jerry_get_literals_from_snapshot` will fit in the buffer of size `lit_out_buf_sz`? And how could we detect if it didn't?

jerry-main/main-unix-snapshot.c
```diff
@@ -436,6 +440,12 @@ static const cli_opt_t merge_opts[] =
 {
   CLI_OPT_DEF (.id = OPT_MERGE_HELP, .opt = "h", .longopt = "help",
                .help = "print this help and exit"),
+  CLI_OPT_DEF (.id = OPT_MERGE_LITERAL_LIST, .longopt = "save-literals-list-format",
+               .meta = "FILE",
+               .help = "export literals found in parsed JS input (in list format)"),
+  CLI_OPT_DEF (.id = OPT_MERGE_LITERAL_C, .longopt = "save-literals-c-format",
+               .meta = "FILE",
+               .help = "export literals found in parsed JS input (in C source format)"),
```
akosthekiss: Wouldn't it be better to have a separate subcommand for extracting literals from a snapshot instead of adding export options to both generate and merge subcommands? That would reduce code duplications also.

docs/02.API-REFERENCE.md
LaszloLango: It is not needed indeed, but I think it looks a bit better. I can change if you want. Both solutions are fine to me.

jerry-main/main-unix-snapshot.c
LaszloLango: I prefer it this way, but we can change it. Do you mean something like this:
`jerry-snapshot literal --format=list -o literal.list FILE_1 ... FILE_N` ?

akosthekiss: Yes, I was thinking something like this. It would have several benefits, IMO. 

The subcommands of the snapshot tool would work as analogies to well known tools from the native domain, e.g., 'generate' is like a compiler, 'merge' resembles a linker, and this 'literal' (or however it is named) could be the counterpart of objdump. (Ofc, none of them are perfect matches, there are important differences everywhere.)

There is a reason why these build tools are separated. This way a user doesn't have to execute unnecessary functionalities to perform a given step. E.g., in this case, assumig that a snapshot is already present but the user forgot to ask for a literal dump at the time of its creation, they don't have to recreate the snapshot from source just to get that dump but it can be retrieved from the snapshot itself.

All this is also aligned with the linux command line tool philosophy: a tool (now, subcommand) should do one thing but do that right. (I.e., don't clone the same functionality into different tools but put it into a single one and leave the functionality composition via sequential execution or pipes or whatever to the user.)

docs/02.API-REFERENCE.md
akosthekiss: Variables should be initialized at the time of their declaration whenever possible. This may be called a personal preference. It's no blocker but wanted to make a note.

LaszloLango:
@akosthekiss thanks for your review. I've updated the PR.

jerry-main/main-unix-snapshot.c
```diff
@@ -389,36 +374,195 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
...
+                                                   snapshot_buffer_sizes[0],
+                                                   literal_buffer,
+                                                   JERRY_BUFFER_SIZE,
+                                                   is_c_format);
+  }
+  else
+  {
+    /* The input contains more than one input snapshot file, so we must merge them first. */
+    const char *error_p = NULL;
+    size_t merged_snapshot_size = jerry_merge_snapshots (snapshot_buffers,
```
akosthekiss: Is this necessary? Cannot it be specified that literal dump works on one snapshot only? Doing merge in this subcommand seems to go against the 'do one thing but do that right' idea.

jerry-main/main-unix-snapshot.c
```diff
@@ -389,36 +374,195 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
...
+  {
+    jerry_port_log (JERRY_LOG_LEVEL_ERROR,
+                    "Error: Literal saving failed! No literals were found in the input snapshot(s).\n");
+    jerry_cleanup ();
+    return JERRY_STANDALONE_EXIT_CODE_FAIL;
+  }
+
+  if (literals_file_name_p == NULL)
+  {
+    literals_file_name_p = is_c_format ? "literals.h" : "literals.list";
```
akosthekiss: cli option comment documents the c format case to write its output to literals.c not .h.

jerry-core/api/jerry-snapshot.c
```diff
@@ -1582,48 +1582,42 @@ ecma_string_is_valid_identifier (const ecma_string_t *string_p)
...
-                                      &bytecode_data_p);
+  const uint8_t *snapshot_data_p = (uint8_t *) snapshot_p;
 
-  if (ECMA_IS_VALUE_ERROR (parse_status))
+  if (snapshot_size <= sizeof (jerry_snapshot_header_t))
   {
-    ecma_free_value (JERRY_CONTEXT (error_value));
-    return 0;
+    ecma_raise_type_error ("Invalid snapshot format");
+    return ecma_create_error_reference_from_context ();
```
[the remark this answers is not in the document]
LaszloLango: Oops, it looks like it is messed up after the rebase. I'll fix it.

jerry-main/main-unix-snapshot.c
LaszloLango: IMHO, yes it is necessary. This is how we usually going to use it. E.g.: in IoT.js. It doesn't save the merged snapshot and does not modify the input, so I think it does only one thing (gather the literals). I don't see any problem with this.

jerry-main/main-unix-snapshot.c
LaszloLango: I've fixed the option comment.

LaszloLango:
@akosthekiss thanks for your review. I've updated the PR.

docs/02.API-REFERENCE.md
```diff
@@ -5303,30 +5303,28 @@ main (void)
...
 
-## jerry_parse_and_save_literals
+## jerry_get_literals_from_snapshot
 
 **Summary**
 
-Collect the used literals from the given source code and save them into a specific file in a list or C format.
-These literals are generated by the parser, they are valid identifiers and none of them are magic string.
+Collect the used literals from the given snapshot and save them into a buffer in list or C format.
+None of these literals are magic strings. In C format only valid identifiers are collected.
```
zherczeg: Is there a reason for the C limitation? It would not be bad to control this with a flag.

jerry-core/api/jerry-snapshot.c
```diff
@@ -1047,11 +1047,10 @@ jerry_exec_snapshot (const uint32_t *snapshot_p, /**< snapshot */
  */
 static void
 scan_snapshot_functions (const uint8_t *buffer_p, /**< snapshot buffer start */
-                         const uint8_t *buffer_end_p, /**< snapshot buffer end */
```
zherczeg: Hm, I don't really like removing this, but since this is a static function we can reintroduce it later if needed.

jerry-core/api/jerry-snapshot.c
```diff
@@ -1582,48 +1580,41 @@ ecma_string_is_valid_identifier (const ecma_string_t *string_p)
...
-                                      source_size,
-                                      is_strict,
-                                      &bytecode_data_p);
+  const uint8_t *snapshot_data_p = (uint8_t *) snapshot_p;
 
-  if (ECMA_IS_VALUE_ERROR (parse_status))
+  if (snapshot_size <= sizeof (jerry_snapshot_header_t))
   {
-    ecma_free_value (JERRY_CONTEXT (error_value));
+    /* Invalid snapshot format */
```
zherczeg: I would add a few sanity checks, for example the jerry magic string.

jerry-main/main-unix-snapshot.c
zherczeg: +1 for the auto-merge. Snapshot tool is not intended to run in low-end, so adding convenience functions should not hurt.

docs/02.API-REFERENCE.md
zherczeg: (Can be done later if backward incompatible API changes can be avoided)

LaszloLango: The reason is that special unicode characters breaks the C format. It is deficiency of the current version (latest master) as well. I'd like to work on it in a followup if it is good for you.

docs/02.API-REFERENCE.md
```diff
@@ -5393,20 +5391,31 @@ main (void)
...
+                                                                literal_buffer,
+                                                                256,
+                                                                true);
 
-  if (literal_sizes != 0)
+  if (literal_size != 0)
   {
-    FILE *literal_file_p = fopen ("literals.txt", "w");
-    fwrite (save_literal_buffer, sizeof (uint8_t), literal_sizes, literal_file_p);
+    FILE *literal_file_p = fopen ("literals.txt", "wb");
```
akosthekiss: note: `jerry_get_literals_from_snapshot` is called with `is_c_format=true`, so the .txt extension is a bit counterintuitive in the example. (this is nothing functional, just an observation)

jerry-core/api/jerry-snapshot.c
akosthekiss: Let me hm, too. I didn't want to suggest the removal of this argument, although I see how my comment has led here. I should dive a bit deeper into snapshot functions to check the semantics behind `buffer_end_p` and `literal_base_p`, where and when and how they are used, why they are the same at all/most call sites, how can the literal data start after the snapshot buffer, etc. If they are really always equal then that may hint at a conceptual problem (or that may be too harsh, something that could be optimised, or named better, etc).

So I +1 now to no removal and to switch back to the original redundant arguments. (But someone could put this on a todo list and take a closer look.)

jerry-main/main-unix-snapshot.c
akosthekiss: side note: this code could become simpler if `jerry_merge_snapshots` could deal with corner cases like `number_of_files=1`. I recall that there was some discussion about this back when the snapshot tool was first implemented. one more item on a todo list?

jerry-main/main-unix-snapshot.c
```diff
@@ -389,36 +374,199 @@ process_generate (cli_state_t *cli_state_p, /**< cli state */
...
+  {
+    fwrite (literal_buffer, sizeof (uint8_t), lit_buf_sz, file_p);
+    fclose (file_p);
+  }
+  else
+  {
+    jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Error: cannot open file: '%s'\n", literals_file_name_p);
+    jerry_cleanup ();
+    return JERRY_STANDALONE_EXIT_CODE_FAIL;
+  }
```
akosthekiss: the `if (...) { log and return } ...` (without an `else` branch) style is used all over the code, and could also be used here: `if (file_p == NULL) { bail out }`

LaszloLango: fixed

LaszloLango:
@zherczeg @akosthekiss the PR is updated.

PR #2512, opened 2018-09-06
Remove the JERRY_GET_CURRENT_CONTEXT macro

akosthekiss:
It was an unused and undocumented macro that enabled diverting the
`jerry_port_get_current_context` calls, an approach that was not
used anywhere else in the Port API.


akosthekiss:
It's been a week now and no one spoke up. Can we land this?

PR #2515, opened 2018-09-07
Send print output to debugger client from the port implementation

akosthekiss:
... not from the print handler extension. The sending of logs is
already in the port implementation.

### DIFF
--- a/jerry-port/default/defaultx-handler.c
+++ b/jerry-port/default/defaultx-handler.c
@@ -17,6 +17,14 @@
 
 #include "jerryscript-ext/handler.h"
 
+#ifdef JERRY_DEBUGGER
+
+#define DEBUG_BUFFER_SIZE (256)
+static char debug_buffer[DEBUG_BUFFER_SIZE];
+static int debug_buffer_index = 0;
+
+#endif /* JERRY_DEBUGGER */
+
 /**
  * Default implementation of jerryx_port_handler_print_char. Uses 'printf' to
  * print a single character to standard output.
@@ -25,4 +33,14 @@ void
 jerryx_port_handler_print_char (char c) /**< the character to print */
 {
   printf ("%c", c);
+
+#ifdef JERRY_DEBUGGER
+  debug_buffer[debug_buffer_index++] = c;
+
+  if ((debug_buffer_index == DEBUG_BUFFER_SIZE) || (c == '\n'))
+  {
+    jerry_debugger_send_output ((jerry_char_t *) debug_buffer, (jerry_size_t) debug_buffer_index);
+    debug_buffer_index = 0;
+  }
+#endif /* JERRY_DEBUGGER */
 } /* jerryx_port_handler_print_char */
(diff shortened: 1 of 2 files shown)


jerry-port/default/defaultx-handler.c
```diff
@@ -25,4 +33,14 @@ void
 jerryx_port_handler_print_char (char c) /**< the character to print */
 {
   printf ("%c", c);
+
+#ifdef JERRY_DEBUGGER
+  debug_buffer[debug_buffer_index++] = c;
+
+  if ((debug_buffer_index == DEBUG_BUFFER_SIZE) || (c == '\n'))
```
galpeter: What would happen if there is only one character to print out?

akosthekiss: Print is always terminated by a newline. This function is the port implementation helper of the print handler extension. We know how that handler works.

galpeter: Ah... that is true.

zherczeg:
Well, this would mean that without proper implementation by each port the output of print will never send to the debugger. I would like to hear the port maintainers opinion about this change.

akosthekiss:
Note that this is already true for the logs. If the port implementation of the log API doesn't support it, the debugger client will not get the logs either.

Moreover, we only 'guarantee' the print output to be sent to the debugger client if print is handled by the jerry-ext print handler `jerryx_handler_print `. But if it is handled by that implementation, we require that the port implements `jerryx_port_handler_print_char` anyway. If a port implements that, it can choose to add the debugger client support, too. (Just like it adds debugger client support to its `jerry_port_log` implementation.)

IMO, this patch unifies which component's responsibility is to support the debugger client (it moves all such responsibility to the port).

akosthekiss:
E.g., IoT.js doesn't even use the print handler. AFAIK, there is no `print` available in JS in that project. The closest thing is `console.log`. But the native code of the console module simply uses `fprintf` and doesn't send anything to a debugger client. So, the concept of 'print output' is very application specific and it's debugging support will always need some extra work from port and/or application developers.

jerry-port/default/defaultx-handler.c
```diff
@@ -25,4 +33,14 @@ void
 jerryx_port_handler_print_char (char c) /**< the character to print */
 {
   printf ("%c", c);
+
+#ifdef JERRY_DEBUGGER
```
LaszloLango: I have concerns of using defines in the default port implementation. The documentation does not say anything about the debugger related part. IoT.js uses it's own port implementation in certain platforms, so these changes should be applied there as well, but documentation does not contain information about the debugger output.

akosthekiss: Note 1: Please note that the default port already contains debugger-specific code and macro guards (see `default-io.c`).

Note 2: Please also note my comment above. AFAIK, IoT.js doesn't use the external print handler from `jerry-ext/handler/handler-print.c` at all, which means that nothing in this file (debugger-specific or not) is used by IoT.js or its platform adaptations. Or am I missing something?

LaszloLango: 1) Thanks for pointing it out. I did not know that.
2) My bad. You are right. It does not use the print handler of jerry-ext.

PR #2516, opened 2018-09-10
Add soft assert handler to jerry-ext

akosthekiss:
Allow developers to choose whether they want a hard or a soft
implementation behind `assert`, i.e., whether they want to kill
the whole engine in case of an assertion failure in the executed
script or just throw a JS error.

### DIFF
--- a/jerry-ext/handler/handler-assert.c
+++ b/jerry-ext/handler/handler-assert.c
@@ -17,16 +17,16 @@
 #include "jerryscript-port.h"
 
 /**
- * Assert for scripts. The routine calls jerry_port_fatal on assertion failure.
+ * Hard assert for scripts. The routine calls jerry_port_fatal on assertion failure.
  *
  * @return true - if only one argument was passed and that argument was a boolean true.
  *         Note that the function does not return otherwise.
  */
 jerry_value_t
-jerryx_handler_assert (const jerry_value_t func_obj_val, /**< function object */
-                       const jerry_value_t this_p, /**< this arg */
-                       const jerry_value_t args_p[], /**< function arguments */
-                       const jerry_length_t args_cnt) /**< number of function arguments */
+jerryx_handler_assert_fatal (const jerry_value_t func_obj_val, /**< function object */
+                             const jerry_value_t this_p, /**< this arg */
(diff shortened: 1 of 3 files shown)


jerry-ext/handler/handler-assert.c
```diff
@@ -17,16 +17,16 @@
...
  *
  * @return true - if only one argument was passed and that argument was a boolean true.
  *         Note that the function does not return otherwise.
  */
 jerry_value_t
-jerryx_handler_assert (const jerry_value_t func_obj_val, /**< function object */
-                       const jerry_value_t this_p, /**< this arg */
-                       const jerry_value_t args_p[], /**< function arguments */
-                       const jerry_length_t args_cnt) /**< number of function arguments */
+jerryx_handler_assert_fatal (const jerry_value_t func_obj_val, /**< function object */
```
zherczeg: Why this function is not a void function?

akosthekiss: Because it is an external handler function with a fixed signature that has to take a given number of arguments and must return a `jerry_value_t`. And it does indeed return with jerry-true if the assert passes. Just like it did from the introduction of this function.

zherczeg: I forgot to open the full code. Patch is ok but I dont like the if-return else-return constructs. There is no need an else after a return. Less indentation.

jerry-ext/handler/handler-assert.c
```diff
@@ -42,4 +42,46 @@ jerryx_handler_assert (const jerry_value_t func_obj_val, /**< function object */
...
+  (void) func_obj_val; /* unused */
+  (void) this_p; /* unused */
+
+  if (args_cnt == 1
+      && jerry_value_is_boolean (args_p[0])
+      && jerry_get_boolean_value (args_p[0]))
+  {
+    return jerry_create_boolean (true);
+  }
+  else
```
LaszloLango: Please remove the `else` as @zherczeg mentioned.

akosthekiss:
PR updated according to the reviews. (Both existing and new code restyled not to do if-return-else-return.)

PR #2519, opened 2018-09-11
Fix JERRY_CONTEXT (error_value) handling during ecma_builtin_promise_reject_abrupt

### LINKED ISSUE #2489: Assertion 'object_p->type_flags_refs >= ECMA_OBJECT_REF_ONE' failed in ecma_deref_object
###### Jerry version:

```
Checked revision: 300e40ba
Build command: tools/build.py --debug --logging=on --profile=es2015-subset
```

###### OS:

```
Ubuntu 18.04, x86_64
```

###### Test case:

```javascript
Object.defineProperty(Object.prototype, 0, {'get': function() { throw $ }});
Promise.all();
```

###### Backtrace:

```
ICE: Assertion 'object_p->type_flags_refs >= ECMA_OBJECT_REF_ONE' failed at jerryscript/jerry-core/ecma/base/ecma-gc.c(ecma_deref_object):149.

Program received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
51	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:51
#1  0x00007ffff7a24801 in __GI_abort () at abort.c:79
#2  0x000055555555f364 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-port/default/default-fatal.c:71
#3  0x000055555558ac1f in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#4  0x000055555558ac71 in jerry_assert_fail (assertion=0x5555555e2618 "object_p->type_flags_refs >= ECMA_OBJECT_REF_ONE", 
    file=0x5555555e2558 "jerryscript/jerry-core/ecma/base/ecma-gc.c", 
    function=0x5555555cca10 <__func__.4599.lto_priv.681> "ecma_deref_object", line=149)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#5  0x00005555555c31fd in ecma_deref_object (object_p=0x5555558f16d0 <jerry_global_heap+800>)
    at jerryscript/jerry-core/ecma/base/ecma-gc.c:149
#6  0x00005555555ae0d0 in ecma_free_value (value=803) at jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:921
#7  0x00005555555bfcd4 in jerry_release_value (value=803) at jerryscript/jerry-core/api/jerry.c:1193
#8  0x00005555555bdcec in main (argc=3, argv=0x7fffffffdc58) at jerryscript/jerry-main/main-unix.c:861
```

<sup>Found by [Fuzzinator](http://fuzzinator.readthedocs.io/) with [grammarinator](https://github.com/renatahodovan/grammarinator). </sup>

rerobika:
Fixes #2489.

Co-authored-by: Daniel Balla <dballa@inf.u-szeged.hu>

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
@@ -125,6 +125,7 @@ ecma_builtin_promise_reject_or_resolve (ecma_value_t this_arg, /**< "this" argum
 inline static ecma_value_t
 ecma_builtin_promise_reject_abrupt (ecma_value_t capability) /**< reject description */
 {
+  ecma_raise_type_error (ECMA_ERR_MSG ("Second argument is not an array."));
   ecma_value_t reason = JERRY_CONTEXT (error_value);
   ecma_string_t *reject_str_p = ecma_get_ecma_string_from_uint32 (ECMA_PROMISE_PROPERTY_REJECT);
   ecma_value_t reject = ecma_op_object_get (ecma_get_object_from_value (capability), reject_str_p);
@@ -134,6 +135,7 @@ ecma_builtin_promise_reject_abrupt (ecma_value_t capability) /**< reject descrip
                                                  &reason,
                                                  1);
   ecma_free_value (reject);
+  ecma_free_value (reason);
 
   if (ECMA_IS_VALUE_ERROR (call_ret))
   {
@@ -576,11 +578,8 @@ ecma_builtin_promise_race_or_all (ecma_value_t this_arg, /**< 'this' argument */
   if (!ecma_is_value_object (array)
       || ecma_get_object_type (ecma_get_object_from_value (array)) != ECMA_OBJECT_TYPE_ARRAY)
   {
-    ecma_raise_type_error (ECMA_ERR_MSG ("Second argument is not an array."));
     ret = ecma_builtin_promise_reject_abrupt (capability);
-    ecma_free_value (JERRY_CONTEXT (error_value));
     ecma_free_value (capability);
-
     return ret;
   }
 
--- a/tests/jerry/fail/regression-test-issue-2489.js
+++ b/tests/jerry/fail/regression-test-issue-2489.js
@@ -1,0 +1,16 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+Object.defineProperty(Object.prototype, 0, {'get': function() { throw $ }});
+Promise.all();


jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -577,10 +577,19 @@ ecma_builtin_promise_race_or_all (ecma_value_t this_arg, /**< 'this' argument */
       || ecma_get_object_type (ecma_get_object_from_value (array)) != ECMA_OBJECT_TYPE_ARRAY)
   {
     ecma_raise_type_error (ECMA_ERR_MSG ("Second argument is not an array."));
+    ecma_value_t current_error_value = JERRY_CONTEXT (error_value);
     ret = ecma_builtin_promise_reject_abrupt (capability);
-    ecma_free_value (JERRY_CONTEXT (error_value));
-    ecma_free_value (capability);
 
+    if (JERRY_CONTEXT (error_value) == current_error_value)
```
LaszloLango: Why is this if needed? If they are equal or not you must free `current_error_value`. Am I miss something?

rerobika: The reason of this `if` that `ecma_raise_type_error` always sets the `JERRY_CONTEXT(error_value)` (and this error must be freed like before) but `ecma_builtin_promise_reject_abrupt` can override this value. In this case the correct behavior is to keep the overriden error_value and free the previously raised TypeError.

zherczeg: Why dont we postpone the error construction after the `ecma_builtin_promise_reject_abrupt` call?

akosthekiss: On the necessity of the `if`, I go with @LaszloLango here. If `JERRY_CONTEXT (error_value) == current_error_value` (this is the predicate of the if) then `ecma_free_value (JERRY_CONTEXT (error_value))` (which is in the true branch of the if) should be rewritable to `ecma_free_value (current_error_value)`. But then the true and false branches of the if become equivalent, making all this simply a `ecma_free_value (current_error_value);`.

All which means that @zherczeg 's suggestion is most probably right: swapping `ecma_raise_type_error` and `ecma_builtin_promise_reject_abrupt` _should_ give an equivalent and much simpler solution.

rerobika: I @akosthekiss @LaszloLango Finally I've understood your thoughts, and this `if` in unnecessary, because the raised TypeError __must__ be freed anyway, but we have to make sure that we free the correct error. (That's why the we store the `JERRY_CONTEXT (error_value)` in a local variable, to prevent `ecma_builtin_promise_reject_abrupt` from override the `JERRY_CONTEXT (error_value)`.)

@zherczeg Swapping the `ecma_raise_type_error` and `ecma_builtin_promise_reject_abrupt` is not a way to go, because the raised TypeError is going to be the reason of the promise rejection, that's why this order is needed.

I've updated the PR, IMO this solution is much simpler then the previous one.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -577,10 +577,10 @@ ecma_builtin_promise_race_or_all (ecma_value_t this_arg, /**< 'this' argument */
       || ecma_get_object_type (ecma_get_object_from_value (array)) != ECMA_OBJECT_TYPE_ARRAY)
   {
     ecma_raise_type_error (ECMA_ERR_MSG ("Second argument is not an array."));
+    ecma_value_t current_error_value = JERRY_CONTEXT (error_value);
     ret = ecma_builtin_promise_reject_abrupt (capability);
-    ecma_free_value (JERRY_CONTEXT (error_value));
+    ecma_free_value (current_error_value);
```
akosthekiss: Shouldn't it be the responsibility of `ecma_builtin_promise_reject_abrupt` to free the error then? Between lines 135/136, between the `ecma_op_function_call` and `ecma_free_value (reject);`, `ecma_free_value (reason);` could achieve the same.

rerobika: Honestly, I'am not sure about that who's responsibility to free the `JERRY_CONTEXT (error_value)`.

Both solutions (I mean your suggestion and the PR's content) have advantages/disadvantages.
Your suggestion:
 - Pros: The `error_value` is always freed.
 - Cons: The `error_value` is also freed even if there wasn't a raised error before.

 PR's content:
 - Pros: Always the correct `error_value` is going to be freed because the developer raises and frees it.
 - Cons: The freeing part can be forgotten.

I do not insist to any of the proposed solutions but I'd ask @zherczeg and @LaszloLango about their opinions.

akosthekiss: > Cons: The error_value is also freed even if there wasn't a raised error before.

Note that `ecma_builtin_promise_reject_abrupt` is only called from here, so if it is called, then an error has been raised.

zherczeg: Yes, it is called from one place. I would move the error construction / free there then.

rerobika: You succeeded in persuading me :) The PR is updated.

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -578,9 +579,7 @@ ecma_builtin_promise_race_or_all (ecma_value_t this_arg, /**< 'this' argument */
   {
     ecma_raise_type_error (ECMA_ERR_MSG ("Second argument is not an array."));
```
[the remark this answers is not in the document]
rerobika: Anyway, it can be also moved. I've checked the [standard](https://www.ecma-international.org/ecma-262/6.0/#sec-ifabruptrejectpromise) and the raised error is the same in both cases (`Promise.rase/all`). OK, I move it as well.

PR #2525, opened 2018-09-13
Fix check-magic-strings.sh to return with non-zero if error happens

Achie72:
Running `tools/run_tests.py --check-magic-string` causes an IndexError,
but the script exits with 0, indicating error-less run.
Added some error checking to the files, and solution for the IndexError.

### DIFF
--- a/tools/check-magic-strings.sh
+++ b/tools/check-magic-strings.sh
@@ -20,13 +20,17 @@ MAGIC_STRINGS_TEMP=`mktemp lit-magic-strings.inc.h.XXXXXXXXXX`
 
 cp $MAGIC_STRINGS_INC_H $MAGIC_STRINGS_TEMP
 $MAGIC_STRINGS_GEN
-diff -q $MAGIC_STRINGS_INC_H $MAGIC_STRINGS_TEMP
 DIFF_RESULT=$?
-mv $MAGIC_STRINGS_TEMP $MAGIC_STRINGS_INC_H
 
-if [ $DIFF_RESULT -ne 0 ]
+if [ $DIFF_RESULT -eq 0 ]
 then
-  echo -e "\e[1;33m$MAGIC_STRINGS_INC_H must be re-generated. Run $MAGIC_STRINGS_GEN\e[0m"
+  diff -q $MAGIC_STRINGS_INC_H $MAGIC_STRINGS_TEMP
+  DIFF_RESULT=$?
+  if [ $DIFF_RESULT -ne 0 ]
+  then
+    echo -e "\e[1;33m$MAGIC_STRINGS_INC_H must be re-generated. Run $MAGIC_STRINGS_GEN\e[0m"
+  fi
 fi
+mv $MAGIC_STRINGS_TEMP $MAGIC_STRINGS_INC_H
 
 exit $DIFF_RESULT


LaszloLango:
Please fix the typo in the commit message: `runnnin` -> `running`

tools/check-magic-strings.sh
```diff
@@ -20,13 +20,18 @@ MAGIC_STRINGS_TEMP=`mktemp lit-magic-strings.inc.h.XXXXXXXXXX`
...
-DIFF_RESULT=$?
-mv $MAGIC_STRINGS_TEMP $MAGIC_STRINGS_INC_H
 
-if [ $DIFF_RESULT -ne 0 ]
-then
-  echo -e "\e[1;33m$MAGIC_STRINGS_INC_H must be re-generated. Run $MAGIC_STRINGS_GEN\e[0m"
+if [ $? -eq 0 ]; then
+  diff -q $MAGIC_STRINGS_INC_H $MAGIC_STRINGS_TEMP
+  DIFF_RESULT=$?
+  mv $MAGIC_STRINGS_TEMP $MAGIC_STRINGS_INC_H
```
akosthekiss: This `mv` should happen no matter what (as it restores the potentially scratched or overwritten header file), so it should be moved after `fi`.

tools/gen-magic-strings.py
```diff
@@ -133,7 +133,8 @@ def process_file(fname):
                 guards = guard_stack[-1]
                 guards[-1] = '!(%s)' % guards[-1]
             elif endif_match is not None:
-                guard_stack.pop()
+                if guard_stack:
+                    guard_stack.pop()
```
akosthekiss: I don't think that this should be fixed here / this way. If there is an `#endif` in the sources but the stack is empty then we have missed something, which should not go hidden.

I see 2-3 options:
1) Don't fix this.
2) Fix the problem of gen-magic-strings by updating the regexes in lines 122-125 (Right now they cannot recognize preproc directives if they don't start on column 0 of the line. The `'^#` patterns should be updated as `'^ *#`.)
3) Don't fix this in this PR but in a separate one, as mentioned in option 2.

I'd suggest option 3. With check-magic-strings.sh already being fixed. should any error be raised in gen-magic-strings.py, that will not go unnoticed anymore. So the primary objective, which is quite critical, is already solved.

Achie72: For a moment i was confused about the regex change, but then i realized that, the ' is just the opening quote for the regex string.
3, Seems like a good idea. Will make a PR for that in a moment.

tools/check-magic-strings.sh
```diff
@@ -20,13 +20,17 @@ MAGIC_STRINGS_TEMP=`mktemp lit-magic-strings.inc.h.XXXXXXXXXX`
...
 cp $MAGIC_STRINGS_INC_H $MAGIC_STRINGS_TEMP
 $MAGIC_STRINGS_GEN
-diff -q $MAGIC_STRINGS_INC_H $MAGIC_STRINGS_TEMP
-DIFF_RESULT=$?
-mv $MAGIC_STRINGS_TEMP $MAGIC_STRINGS_INC_H
 
-if [ $DIFF_RESULT -ne 0 ]
-then
-  echo -e "\e[1;33m$MAGIC_STRINGS_INC_H must be re-generated. Run $MAGIC_STRINGS_GEN\e[0m"
+if [ $? -eq 0 ]; then
```
akosthekiss: If I recall correctly, we usually write `if` and `then`on separate lines without `;`. It's better to use a consistent style.

tools/check-magic-strings.sh
```diff
@@ -20,13 +20,17 @@ MAGIC_STRINGS_TEMP=`mktemp lit-magic-strings.inc.h.XXXXXXXXXX`
...
+  diff -q $MAGIC_STRINGS_INC_H $MAGIC_STRINGS_TEMP
+  DIFF_RESULT=$?
+  if [ $DIFF_RESULT -ne 0 ]
+  then
+    echo -e "\e[1;33m$MAGIC_STRINGS_INC_H must be re-generated. Run $MAGIC_STRINGS_GEN\e[0m"
+  fi
+else
+  DIFF_RESULT=$?
```
akosthekiss: Bash is a tricky beast. Here, `$?` is not the result of `$MAGIC_STRINGS_GEN` anymore but that of `[ $? -eq 0 ]`.

Try:
```sh
$ false; echo $?
1
$ false; if [ $? -eq 0 ]; then echo; else echo $?; fi
1
$ false; if [ $? -ne 0 ]; then echo $?; fi
0
```

So, it should be better to write:
```sh
$MAGIC_STRINGS_GEN
DIFF_RESULT=$?

if [ $DIFF_RESULT -eq 0 ]
then
  diff ...
  ...
fi
```

PR #2535, opened 2018-09-20
Fix double literal free.

### LINKED ISSUE #2531: The JS engine may free a literal twice if error occurs at function "parser_post_processing"
The test code:
[regexp-simple-atom-and-iterations.js.txt](https://github.com/jerryscript-project/jerryscript/files/2393003/regexp-simple-atom-and-iterations.js.txt)

Compile like this:
python tools/build.py  --error-messages=ON --line-info=ON --mem-heap 9 --debug

Then there will be an assertion fail to indicate the pointer to be freed is not pointing to the heap.  

I think it's the double free of literal causing such a isuue.

First free at "js-parser.c" line 206,which is called by function "parser_post_processing" at line 1496

After error occurs at line 1726 for "out of memroy" error,the second free will happen at line 2416-->line 2413-->"common.c" line 42

Of course,when the second free happens,the address passed to function "jmem_heap_free_block" is totally wrong as it was changed at "js-parser.c" line 201,which is the reason why the assertion fails.

zherczeg:
Fixes #2531.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -75,7 +75,7 @@ matrix:
 
     - env:
         - JOBNAME="ASAN Tests"
-        - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list=parser-oom.js --buildoptions=--compile-flag=-fsanitize=address,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--compile-flag=-O2,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
+        - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list=parser-oom.js,parser-oom2.js --buildoptions=--compile-flag=-fsanitize=address,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--compile-flag=-O2,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
         - ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=true:strict_init_order=true
         - TIMEOUT=600
       compiler: gcc-5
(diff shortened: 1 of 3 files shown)


zherczeg:
Bots are green! Please review the patch.

.travis.yml
```diff
@@ -76,7 +76,7 @@ matrix:
 
     - env:
         - JOBNAME="ASAN Tests"
-        - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list=parser-oom.js --buildoptions=--compile-flag=-fsanitize=address,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--compile-flag=-O2,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
+        - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list="parser-oom.js,parser-oom2.js" --buildoptions=--compile-flag=-fsanitize=address,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--compile-flag=-O2,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
```
akosthekiss: It's strange that double quotes work within a double quoted string. And actually, they aren't. Travis CI logs show that the following env vars will get set:

```
$ export JOBNAME="ASAN Tests"
$ export OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list="
$ export buildoptions=--compile-flag=-fsanitize=address,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--compile-flag=-O2,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold
$ export ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=true:strict_init_order=true
$ export TIMEOUT=600
```

This is not as intended, I guess.

.travis.yml
```diff
@@ -87,7 +87,7 @@ matrix:
 
     - env:
         - JOBNAME="UBSAN Tests"
-        - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list=parser-oom.js --buildoptions=--compile-flag=-fsanitize=undefined,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
+        - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list="parser-oom.js,parser-oom2.js" --buildoptions=--compile-flag=-fsanitize=undefined,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
```
akosthekiss: ditto

zherczeg:
Patch updated,. please check.

PR #2540, opened 2018-09-25
Change jerry_port interface to allow for a correct implementation of timezones.

### LINKED ISSUE #1661: Calculation of daylight saving time is wrong
After the PST -> PDT change some of the test262 tests are failed. The reason is the wrong daylight saving time. `ecma_date_daylight_saving_ta` in `ecma-builtin-helpers-date.c` is get the daylight saving adjustment from the system, but it gets it for the current date and not for the date we must set (e.g.: `new Date(1969, 12)`).

```
=== Summary ===
 - Ran 11552 tests
 - Passed 11546 tests (99.9%)
 - Failed 6 tests (0.1%)

Failed tests
  ch15/15.9/15.9.3/S15.9.3.1_A5_T1 in non-strict mode
  ch15/15.9/15.9.3/S15.9.3.1_A5_T2 in non-strict mode
  ch15/15.9/15.9.3/S15.9.3.1_A5_T3 in non-strict mode
  ch15/15.9/15.9.3/S15.9.3.1_A5_T4 in non-strict mode
  ch15/15.9/15.9.3/S15.9.3.1_A5_T5 in non-strict mode
  ch15/15.9/15.9.3/S15.9.3.1_A5_T6 in non-strict mode
```

crazy2be:
The previous jerry_port interface did not allow timezones to be handled correctly,
even if the host system was up to the task. This PR changes the jerry_port interface
to allow a completely correct implementation to exist, and provides one (for Linux/BSD
systems) in default-date.c.

There is some additional background in my comment on the original issue https://github.com/jerryscript-project/jerryscript/issues/1661#issuecomment-420782958, but essentially, this solution resolves the original ticket by allowing the jerry_port interface to from the time-invariant 
```c
jerry_port_get_time_zone(utc_offset_out*, is_dst_out*)
```
to the time-dependent
```c
jerry_port_get_local_tza(time, is_utc)
```
As dst state and timezone rules in general change over time, it's necessary to give the time as an argument to correctly determine the local time offset at that time.

This interface also allows the platform to figure out the local time adjustment for a given date in whichever way it pleases. On most Linux / BSD systems, the `tm_gmtoff` field of `struct tm` can be used. On windows, `GetTimeZoneInformationForYear` can be used. On most embedded platforms, UTC can always be assumed, or some other local timezone.

The alternative is to have jerryscript link in some external timezone library, with all of the logic for handling timezones. However, I can't see any big advantage to this approach, and the libraries are often quite large. Most systems that run JerryScript simply don't need the complexity (they can just run UTC), and the ones that do need the complexity probably already have some other system which will be fairly easy to just link into the provided jerry_port interface (Linux and Windows are both quite easy, about 10 lines, to hook into this interface).

Fixes #1661

JerryScript-DCO-1.0-Signed-off-by: crazy2be crazy1be@gmail.com

### DIFF
--- a/docs/05.PORT-API.md
+++ b/docs/05.PORT-API.md
@@ -76,27 +76,37 @@ void jerry_port_log (jerry_log_level_t level, const char *fmt, ...);
 
 ```c
 /**
- * Jerry time zone structure
- */
-typedef struct
-{
-  int offset;                /**< minutes from west */
-  int daylight_saving_time;  /**< daylight saving time (1 - DST applies, 0 - not on DST) */
-} jerry_time_zone_t;
-
-/**
- * Get timezone and daylight saving data
+ * Get local time zone adjustment, in milliseconds, for the given timestamp.
+ * The timestamp can be specified in either UTC or local time, depending on
+ * the value of is_utc. Adding the value returned from this function to
+ * a timestamp in UTC time should result in local time for the current time
+ * zone, and subtracting it from a timestamp in local time should result in
+ * UTC time.
+ *
+ * Ideally, this function should satisfy the stipulations applied to LocalTZA
+ * in section 20.3.1.7 of the ECMAScript version 9.0 spec.
+ *
+ * See Also:
+ *          ECMA-262 v9, 20.3.1.7
  *
  * Note:
  *      This port function is called by jerry-core when
  *      CONFIG_DISABLE_DATE_BUILTIN is _not_ defined. Otherwise this function is
  *      not used.
  *
- * @param[out] tz_p time zone structure to fill.
- * @return true  - if success
- *         false - otherwise
+ * @param unix_ms The unix timestamp we want an offset for, given in
+ *                millisecond precision (could be now, in the future,
+ *                or in the past). As with all unix timestamps, 0 refers to
+ *                1970-01-01, a day is exactly 86 400 000 milliseconds, and
+ *                leap seconds cause the same second to occur twice.
+ * @param is_utc Is the given timestamp in UTC time? If false, it is in local
+ *               time.
+ *
+ * @return milliseconds between local time and UTC for the given timestamp,
+ *         if available
+ *.        0 if not available / we are in UTC.
  */
-bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p);
+double jerry_port_get_local_time_zone_adjustment (double unix_ms, bool is_utc);
 
 /**
  * Get system time
@@ -194,31 +204,26 @@ jerry_port_log (jerry_log_level_t level, /**< log level */
 ## Date
 
 ```c
+#include <time.h>
 #include <sys/time.h>
 #include "jerryscript-port.h"
 
 /**
- * Default implementation of jerry_port_get_time_zone.
+ * Default implementation of jerry_port_get_local_time_zone_adjustment.
  */
-bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p)
+double jerry_port_get_local_time_zone_adjustment (double unix_ms, /**< ms since unix epoch */
+                                                  bool is_utc)    /**< is the time above in UTC? */
 {
-  struct timeval tv;
-  struct timezone tz;
-
-  /* gettimeofday may not fill tz, so zero-initializing */
-  tz.tz_minuteswest = 0;
-  tz.tz_dsttime = 0;
-
-  if (gettimeofday (&tv, &tz) != 0)
+  struct tm tm;
+  time_t now = (time_t) (unix_ms / 1000);
+  localtime_r (&now, &tm);
+  if (!is_utc)
   {
-    return false;
+    now -= tm.tm_gmtoff;
+    localtime_r (&now, &tm);
   }
-
-  tz_p->offset = tz.tz_minuteswest;
-  tz_p->daylight_saving_time = tz.tz_dsttime > 0 ? 1 : 0;
-
-  return true;
-} /* jerry_port_get_time_zone */
+  return ((double) tm.tm_gmtoff) * 1000;
+} /* jerry_port_get_local_time_zone_adjustment */
 
 /**
  * Default implementation of jerry_port_get_current_time.
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
@@ -302,83 +302,32 @@ ecma_date_week_day (ecma_number_t time) /**< time value */
 } /* ecma_date_week_day */
 
 /**
- * Helper function to get local time zone adjustment.
+ * Helper function to get the local time zone offset at a given UTC timestamp.
+ * You can add this number to the given UTC timestamp to get local time.
  *
  * See also:
- *          ECMA-262 v5, 15.9.1.7
+ *          ECMA-262 v5, 15.9.1.9
  *
  * @return local time zone adjustment
  */
-static inline ecma_number_t JERRY_ATTR_ALWAYS_INLINE
-ecma_date_local_tza (jerry_time_zone_t *tz) /**< time zone information */
-{
-  return tz->offset * -ECMA_DATE_MS_PER_MINUTE;
-} /* ecma_date_local_tza */
-
-/**
- * Helper function to get the daylight saving time adjustment.
- *
- * See also:
- *          ECMA-262 v5, 15.9.1.8
- *
- * @return daylight saving time adjustment
- */
-static inline ecma_number_t JERRY_ATTR_ALWAYS_INLINE
-ecma_date_daylight_saving_ta (jerry_time_zone_t *tz, /**< time zone information */
-                              ecma_number_t time) /**< time value */
-{
-  JERRY_ASSERT (!ecma_number_is_nan (time));
-
-  /*
-   * TODO: Fix daylight saving calculation.
-   *       https://github.com/jerryscript-project/jerryscript/issues/1661
-   */
-  return tz->daylight_saving_time * ECMA_DATE_MS_PER_HOUR;
-} /* ecma_date_daylight_saving_ta */
-
-/**
- * Helper function to get local time from UTC.
- *
- * See also:
- *          ECMA-262 v5, 15.9.1.9
- *
- * @return local time
- */
 inline ecma_number_t JERRY_ATTR_ALWAYS_INLINE
-ecma_date_local_time_zone (ecma_number_t time) /**< time value */
+ecma_date_local_time_zone_adjustment (ecma_number_t time) /**< time value */
 {
-  jerry_time_zone_t tz;
-
-  if (ecma_number_is_nan (time)
-      || !jerry_port_get_time_zone (&tz))
-  {
-    return ecma_number_make_nan ();
-  }
-
-  return ecma_date_local_tza (&tz) + ecma_date_daylight_saving_ta (&tz, time);
-} /* ecma_date_local_time_zone */
+  return jerry_port_get_local_time_zone_adjustment (time, true);
+} /* ecma_date_local_time_zone_adjustment */
 
 /**
- * Helper function to get utc from local time.
+ * Helper function to get UTC time from local time.
  *
  * See also:
  *          ECMA-262 v5, 15.9.1.9
  *
- * @return utc time
+ * @return UTC time
  */
 ecma_number_t
 ecma_date_utc (ecma_number_t time) /**< time value */
 {
-  jerry_time_zone_t tz;
-
-  if (ecma_number_is_nan (time)
-      || !jerry_port_get_time_zone (&tz))
-  {
-    return ecma_number_make_nan ();
-  }
-
-  ecma_number_t simple_utc_time = time - ecma_date_local_tza (&tz);
-  return simple_utc_time - ecma_date_daylight_saving_ta (&tz, simple_utc_time);
+  return time - jerry_port_get_local_time_zone_adjustment (time, false);
 } /* ecma_date_utc */
 
 /**
@@ -615,7 +564,7 @@ ecma_date_timezone_offset (ecma_number_t time) /**< time value */
 {
   JERRY_ASSERT (!ecma_number_is_nan (time));
 
-  return (-ecma_date_local_time_zone (time)) / ECMA_DATE_MS_PER_MINUTE;
+  return (-ecma_date_local_time_zone_adjustment (time)) / ECMA_DATE_MS_PER_MINUTE;
 } /* ecma_date_timezone_offset */
 
 /**
@@ -724,7 +673,7 @@ ecma_date_to_string_format (ecma_number_t datetime_number, /**< datetime */
       }
       case LIT_CHAR_LOWERCASE_Z: /* Time zone minutes part. */
       {
-        int32_t time_zone = (int32_t) ecma_date_local_time_zone (datetime_number);
+        int32_t time_zone = (int32_t) ecma_date_local_time_zone_adjustment (datetime_number);
 
         if (time_zone >= 0)
         {
@@ -744,7 +693,7 @@ ecma_date_to_string_format (ecma_number_t datetime_number, /**< datetime */
       {
         JERRY_ASSERT (*format_p == LIT_CHAR_UPPERCASE_Z); /* Time zone seconds part. */
 
-        int32_t time_zone = (int32_t) ecma_date_local_time_zone (datetime_number);
+        int32_t time_zone = (int32_t) ecma_date_local_time_zone_adjustment (datetime_number);
 
         if (time_zone < 0)
         {
@@ -805,7 +754,7 @@ ecma_date_to_string_format (ecma_number_t datetime_number, /**< datetime */
 ecma_value_t
 ecma_date_value_to_string (ecma_number_t datetime_number) /**< datetime */
 {
-  datetime_number += ecma_date_local_time_zone (datetime_number);
+  datetime_number += ecma_date_local_time_zone_adjustment (datetime_number);
   return ecma_date_to_string_format (datetime_number, "$W $M $D $Y $h:$m:$s GMT$z:$Z");
 } /* ecma_date_value_to_string */
 
--- a/jerry-core/include/jerryscript-port.h
+++ b/jerry-core/include/jerryscript-port.h
@@ -111,27 +111,37 @@ void JERRY_ATTR_FORMAT (printf, 2, 3) jerry_port_log (jerry_log_level_t level, c
  */
 
 /**
- * Jerry time zone structure
- */
-typedef struct
-{
-  int offset;                /**< minutes from west */
-  int daylight_saving_time;  /**< daylight saving time (1 - DST applies, 0 - not on DST) */
-} jerry_time_zone_t;
-
-/**
- * Get timezone and daylight saving data
+ * Get local time zone adjustment, in milliseconds, for the given timestamp.
+ * The timestamp can be specified in either UTC or local time, depending on
+ * the value of is_utc. Adding the value returned from this function to
+ * a timestamp in UTC time should result in local time for the current time
+ * zone, and subtracting it from a timestamp in local time should result in
+ * UTC time.
+ *
+ * Ideally, this function should satisfy the stipulations applied to LocalTZA
+ * in section 20.3.1.7 of the ECMAScript version 9.0 spec.
+ *
+ * See Also:
+ *          ECMA-262 v9, 20.3.1.7
  *
  * Note:
  *      This port function is called by jerry-core when
  *      CONFIG_DISABLE_DATE_BUILTIN is _not_ defined. Otherwise this function is
  *      not used.
  *
- * @param[out] tz_p time zone structure to fill.
- * @return true  - if success
- *         false - otherwise
+ * @param unix_ms The unix timestamp we want an offset for, given in
+ *                millisecond precision (could be now, in the future,
+ *                or in the past). As with all unix timestamps, 0 refers to
+ *                1970-01-01, a day is exactly 86 400 000 milliseconds, and
+ *                leap seconds cause the same second to occur twice.
+ * @param is_utc Is the given timestamp in UTC time? If false, it is in local
+ *               time.
+ *
+ * @return milliseconds between local time and UTC for the given timestamp,
+ *         if available
+ *.        0 if not available / we are in UTC.
  */
-bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p);
+double jerry_port_get_local_time_zone_adjustment (double unix_ms, bool is_utc);
 
 /**
  * Get system time
--- a/jerry-port/default/CMakeLists.txt
+++ b/jerry-port/default/CMakeLists.txt
@@ -26,6 +26,19 @@ file(GLOB SOURCE_PORT_DEFAULT *.c)
 # (should only be necessary if we used compiler default libc but not checking that)
 set(DEFINES_PORT_DEFAULT _BSD_SOURCE _DEFAULT_SOURCE)
 
+INCLUDE (CheckStructHasMember)
+# CHECK_STRUCT_HAS_MEMBER works by trying to compile some C code that accesses the
+# given field of the given struct. However, our default compiler options break this
+# C code, so turn a couple of them off for this.
+set(CMAKE_REQUIRED_FLAGS "-Wno-error=strict-prototypes -Wno-error=old-style-definition")
+# tm.tm_gmtoff is non-standard, so glibc doesn't expose it in c99 mode
+# (our default). Define some macros to expose it anyway.
+set(CMAKE_REQUIRED_DEFINITIONS "-D_BSD_SOURCE -D_DEFAULT_SOURCE")
+CHECK_STRUCT_HAS_MEMBER ("struct tm" tm_gmtoff time.h HAVE_TM_GMTOFF)
+if(HAVE_TM_GMTOFF)
+  set(DEFINES_PORT_DEFAULT ${DEFINES_PORT_DEFAULT} HAVE_TM_GMTOFF)
+endif()
+
 # Sleep function availability check
 INCLUDE (CheckIncludeFiles)
 CHECK_INCLUDE_FILES (time.h HAVE_TIME_H)
--- a/jerry-port/default/default-date.c
+++ b/jerry-port/default/default-date.c
@@ -13,41 +13,43 @@
  * limitations under the License.
  */
 
+#ifdef HAVE_TM_GMTOFF
+#include <time.h>
+#endif /* HAVE_TM_GMTOFF */
 #ifdef __GNUC__
 #include <sys/time.h>
-#endif
+#endif /* __GNUC__ */
 
 #include "jerryscript-port.h"
 #include "jerryscript-port-default.h"
 
 /**
- * Default implementation of jerry_port_get_time_zone. Uses 'gettimeofday' if
- * available on the system, does nothing otherwise.
+ * Default implementation of jerry_port_get_local_time_zone_adjustment. Uses the 'tm_gmtoff' field
+ * of 'struct tm' (a GNU extension) filled by 'localtime_r' if available on the
+ * system, does nothing otherwise.
  *
- * @return true - if 'gettimeofday' is available and executed successfully,
- *         false - otherwise.
+ * @return offset between UTC and local time at the given unix timestamp, if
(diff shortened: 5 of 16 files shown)


crazy2be:
Can one of the maintainers take a look? Maybe @zherczeg?

zherczeg:
I don't know much about times but the patch looks reasonable. Does somebody knows more about time zones here?

LaszloLango:
@crazy2be Sorry for the late response. I'll take a look on this patch later this week.

crazy2be:
Those tests do pass for me, but it took some doing. Initially, they did not pass, and after investigating a bit, they always appeared to be off by 3 hours. After quite some digging, I discovered this is because the author lives in California, and the version of the tests we run assumes the California timezone. This was fixed in [1], but the `es5-tests` branch that `run-test-suite-test262.sh` checks out doesn't have that change. However, if I change my system timezone to California time, all of the tests pass:

```
=== Summary ===
 - Ran 11552 tests
 - All tests succeeded
```
```
ch15/15.9/15.9.3/S15.9.3.1_A5_T1 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T2 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T3 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T4 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T5 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T6 passed in non-strict mode
```
Given this caveat, do we still want these tests enabled by default? Or should I just change the TODO? Ideally, it seems like [1] should be cherry-picked onto the `es5-tests` branch, but that branch seems fairly dead - last commit was more than 3 years ago, in July 2015.

[1] https://github.com/tc39/test262/pull/573/files

jerry-core/include/jerryscript-port.h
```diff
@@ -111,27 +111,19 @@ void JERRY_ATTR_FORMAT (printf, 2, 3) jerry_port_log (jerry_log_level_t level, c
...
-  int offset;                /**< minutes from west */
-  int daylight_saving_time;  /**< daylight saving time (1 - DST applies, 0 - not on DST) */
-} jerry_time_zone_t;
-
-/**
- * Get timezone and daylight saving data
+ * Get local timezone adjustment, in milliseconds. This value, added to a timestamp in UTC
+ * time, should yield local time. Ideally, this function should satisfy the stipulations
+ * applied to LocalTZA in section 20.3.1.7 of ecmascript version 9.0 [1].
+ * [1] https://www.ecma-international.org/ecma-262/9.0/#sec-local-time-zone-adjustment
```
akosthekiss: IIRC, the code base doesn't use url references to external documents, it just mentions the standard name, the version, and the section. Like:
```c
  * See also:
  *          ECMA-262 v5, 15.9.1.9
```
(Taken from another code block, needs to be adapted.)

akosthekiss:
re: the time zone issue during test262#es5-tests testing  
The Travis CI job already ensures the proper time zone by explicitly setting `TZ=America/Los_Angeles` when running `tools/run-tests.py $OPTS` and `OPTS="--test262"`. However, this doesn't ensure the proper TZ during local testing, so it might be a better way to incorporate the environment fixup in `tools/run-tests.py`.

jerry-port/default/default-date.c
```diff
@@ -21,33 +24,32 @@
...
-
-  /* gettimeofday may not fill tz, so zero-initializing */
-  tz.tz_minuteswest = 0;
-  tz.tz_dsttime = 0;
-
-  if (gettimeofday (&tv, &tz) == 0)
+#ifdef HAVE_TM_GMTOFF
+  struct tm tm;
+  time_t now = (time_t) (js_time_ms / 1000);
+  localtime_r (&now, &tm);
```
akosthekiss: A little bit of thinking aloud: `localtime_r` is not C99 but an extension defined by the POSIX standard. Which means that this code will not necessarily compile on all targets. Unless the existence of the `tm.tm_gmtoff` GNU extension implies the POSIX extension as well.

Another approach could be to use `timegm(&tm) - t` instead of `tm.tm_gmtoff` but `timegm` is also a GNU/BSD extension. (Which could be emulated if not present on the system, but I'm not sure it would be worth the effort.)

So, it seems to me that we need to rely on at least two extensions ot get this work. I'm not sure yet which way to go. (That's way I was thinking aloud. I'm open to inputs.)

jerry-port/default/default-date.c
```diff
@@ -21,33 +24,32 @@
...
+ * of 'struct tm' (a GNU extension) filled by 'localtime_r' if available on the
+ * system, does nothing otherwise.
  *
- * @return true - if 'gettimeofday' is available and executed successfully,
- *         false - otherwise.
+ * @return offset between UTC and local time at the given unix timestamp, if
+ *         available. Otherwise, returns 0, assuming UTC time.
  */
-bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p) /**< [out] time zone structure to fill */
+double jerry_port_get_local_tza (double js_time_ms,  /**< ms since unix epoch */
```
akosthekiss: I'm not sure about the `js_` prefix for this param. It's time in millisecs, which is quite well expressed by `time_ms`. There is not much JS-specific in that concept.

jerry-core/include/jerryscript-port.h
```diff
@@ -111,27 +111,25 @@ void JERRY_ATTR_FORMAT (printf, 2, 3) jerry_port_log (jerry_log_level_t level, c
...
-{
-  int offset;                /**< minutes from west */
-  int daylight_saving_time;  /**< daylight saving time (1 - DST applies, 0 - not on DST) */
-} jerry_time_zone_t;
-
-/**
- * Get timezone and daylight saving data
+ * Get local timezone adjustment, in milliseconds. This value, added to a timestamp in UTC
+ * time, should yield local time. Ideally, this function should satisfy the stipulations
+ * applied to LocalTZA in section 20.3.1.7 of ecmascript version 9.0.
```
akosthekiss: Timezone might better be time zone, but I'm not completely sure, even the spec uses timezone at least once (although it uses time zone more often). About ECMAScript instead of ecmascript, I'm positive.

And I also think that the documentation of all other port functions are kept in sync in the docs/05.PORT-API.md and in jerry-core/include/jerryscript-port.h. This should also happen here. Perhaps the two texts should be merged into a single description and used at both places. (Especially as the current text here is a little bit vague about the case when `is_utc == false`.)

jerry-core/include/jerryscript-port.h
```diff
@@ -111,27 +111,25 @@ void JERRY_ATTR_FORMAT (printf, 2, 3) jerry_port_log (jerry_log_level_t level, c
...
+ *
+ * See Also:
+ *          ECMA-262 v9, 20.3.1.7
  *
  * Note:
  *      This port function is called by jerry-core when
- *      CONFIG_DISABLE_DATE_BUILTIN is _not_ defined. Otherwise this function is
- *      not used.
+ *      CONFIG_DISABLE_DATE_BUILTIN is _not_ defined. Otherwise this function
+ *      is not used.
```
akosthekiss: Nitpicking: moving the "is" back to the end of the previous line would give a bit less churn.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
```diff
@@ -301,41 +301,6 @@ ecma_date_week_day (ecma_number_t time) /**< time value */
...
-
-  /*
-   * TODO: Fix daylight saving calculation.
-   *       https://github.com/jerryscript-project/jerryscript/issues/1661
-   */
-  return tz->daylight_saving_time * ECMA_DATE_MS_PER_HOUR;
-} /* ecma_date_daylight_saving_ta */
-
 /**
  * Helper function to get local time from UTC.
```
akosthekiss: The comment of this function (`ecma_date_local_time_zone`) became (was?) misleading. Especially when considered together with the other changed function, `ecma_date_utc`. That latter one is documented as "get utc from local time" which is aligned to the code, but this current one has an inverse doc ("get local time from UTC") although it does not implement the inverse operation. Could be updated in this patch.

jerry-port/default/default-date.c
crazy2be: I'd only done this because it was a `double`, but I agree, I think `unix_ms` is probably a better name. (or do you prefer `time_ms`? Both seem better than `js_time_ms`).

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
crazy2be: I noticed that, but I wasn't sure if you guys would want it fixed. I renamed it to `ecma_date_local_time_zone_adjustment` and updated the docs as well.

jerry-port/default/default-date.c
akosthekiss: `unix_ms` is fine by me

tools/runners/run-test-suite-test262.sh
```diff
@@ -52,23 +52,30 @@ fi
...
+  (>&2 echo)
+  while read line; do
+    FRAC=$((NUM_LINES_GOTTEN * 100 / NUM_LINES_EXPECTED))
+    (>&2 echo -ne "\rEstimated ${FRAC}% complete (${NUM_LINES_GOTTEN}/${NUM_LINES_EXPECTED})...")
+    echo "$line"
+    NUM_LINES_GOTTEN=$((NUM_LINES_GOTTEN+1))
+  done
+  (>&2 echo)
+  (>&2 echo)
+}
```
akosthekiss: Now, this is something that I'd like to see in a different PR. This is a feature/improvement that has nothing to do with the goals of this current PR. This change could work no matter what port API is in place.

tools/runners/run-test-suite-test262.sh
```diff
@@ -52,23 +52,30 @@ fi
...
-rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T2.js"
-rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T3.js"
-rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T4.js"
-rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T5.js"
-rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T6.js"
+echo "Starting test262 testing for ${ENGINE}. Running test262 may take several minutes."
 
-echo "Starting test262 testing for ${ENGINE}. Running test262 may take a several minutes."
+function progress_monitor() {
+  NUM_LINES_EXPECTED=11558
```
akosthekiss: This is too hard-wired. How did you get this number? Why can't it be computed here by this script? (BTW, summary prints me 11552 executed tests.)

tools/runners/run-test-suite-test262.sh
```diff
@@ -52,23 +52,30 @@ fi
...
+  NUM_LINES_GOTTEN=0
+  (>&2 echo)
+  while read line; do
+    FRAC=$((NUM_LINES_GOTTEN * 100 / NUM_LINES_EXPECTED))
+    (>&2 echo -ne "\rEstimated ${FRAC}% complete (${NUM_LINES_GOTTEN}/${NUM_LINES_EXPECTED})...")
```
akosthekiss: This will print 10k lines. On a live console, that's no problem as lines end in CR only without LF. However, it still blows up logs if output is redirected (see Travis CI raw logs). Printing progress (the same way, with CRs) only when the percentage increases, that would be the proper solution.

jerry-core/include/jerryscript-port.h
```diff
@@ -111,27 +111,34 @@ void JERRY_ATTR_FORMAT (printf, 2, 3) jerry_port_log (jerry_log_level_t level, c
...
+ *                or in the past)
+ * @param is_utc Is the given timestamp in UTC time? If false, it is in local
+ *               time.
+ *
+ * @return milliseconds between local time and UTC for the given timestamp,
+ *         if available
+ *.        0 if not available / we are in UTC.
  */
-bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p);
+double jerry_port_get_local_time_zone_adjustment (double unix_ms, bool is_utc);
```
akosthekiss: Question to other reviewers (again, ping @zherczeg @LaszloLango ): do we want this long identifiers / port API identifiers? `jerry_port_get_local_time_zone_adjustment` or `jerry_port_get_local_tza` or ...?

docs/05.PORT-API.md
```diff
@@ -198,27 +205,21 @@ jerry_port_log (jerry_log_level_t level, /**< log level */
 #include "jerryscript-port.h"
 
 /**
- * Default implementation of jerry_port_get_time_zone.
+ * Default implementation of jerry_port_get_local_time_zone_adjustment.
  */
-bool jerry_port_get_time_zone (jerry_time_zone_t *tz_p)
+double jerry_port_get_local_time_zone_adjustment (double unix_ms, /**< ms since unix epoch */
+                                                  bool is_utc)    /**< is the time above in utc? */
```
akosthekiss: Suggestion: UTC

akosthekiss:
Now that #2551 has landed, it might be worth rebasing this PR onto latest master.

jerry-port/default/default-date.c
crazy2be: My point was mainly that if a given platform has `tm.tm_gmtoff` then it _most probably_ has `localtime_r` as well. But maybe there is a better way that has support on more platforms, I'm open to suggestions. We can always change / improve this later, however.

tools/runners/run-test-suite-test262.sh
crazy2be: This is number of lines, not number of tests, so it is slightly more than the number of tests. Previously it just ran for several minutes without giving any feedback, slightly wrong feedback (like if the test suite changes) seems a lot better than no feedback to me. I think the right place to put this is probably down a level in `test262.py`, but that is a considerably more involved change that I'm not about to attempt.

I suppose you could always just print number of tests executed, and not the percentage... at least then you can see that it's progressing. It still seems worse than giving the percentage, even if it's slightly off, but maybe it's an approach that will age better.

crazy2be:
Just looked into the build failures - they seem to be due to my CMake changes not having the desired effect (on Ubuntu, the target is getting built as though there is no `tm_gmtoff` support, when there most certainly is). The rest of the changes should be reviewable, however.

jerry-port/default/default-date.c
zherczeg: Btw what about windows? Does gcc on windows support this?

docs/05.PORT-API.md
```diff
@@ -76,27 +76,34 @@ void jerry_port_log (jerry_log_level_t level, const char *fmt, ...);
...
  *      This port function is called by jerry-core when
  *      CONFIG_DISABLE_DATE_BUILTIN is _not_ defined. Otherwise this function is
  *      not used.
  *
- * @param[out] tz_p time zone structure to fill.
- * @return true  - if success
- *         false - otherwise
+ * @param unix_ms The unix timestamp we want an offset for, given in
+ *                millisecond precision (could be now, in the future,
+ *                or in the past)
```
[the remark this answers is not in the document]
crazy2be: Correct, the base is 1970.01.01. The given `unix_ms` is assumed to be unix time - i.e. every day is exactly 86400 seconds, leap seconds cause the same second to occur twice. This seems to be the same set of assumptions made by the date object according to MDN [1]

> The JavaScript date is based on a time value that is milliseconds since midnight January 1, 1970, UTC. A day holds 86,400,000 milliseconds. The JavaScript Date object range is -100,000,000 days to 100,000,000 days relative to January 1, 1970 UTC.

This API only cares about handling them correctly to the extent that it allows it to most correctly determine if dst is in effect, other than that, they should not affect the returned `local_time_zone_adjustment`, since leap seconds occur at the same time around the world, regardless of local timezone.

[1] https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date

docs/05.PORT-API.md
```diff
@@ -76,27 +76,34 @@ void jerry_port_log (jerry_log_level_t level, const char *fmt, ...);
...
-  int daylight_saving_time;  /**< daylight saving time (1 - DST applies, 0 - not on DST) */
-} jerry_time_zone_t;
-
-/**
- * Get timezone and daylight saving data
+ * Get local time zone adjustment, in milliseconds, for the given timestamp.
+ * The timestamp can be specified in either UTC or local time, depending on
+ * the value of is_utc. Adding the value returned from this function to
+ * a timestamp in UTC time should result in local time for the current time
+ * zone.
```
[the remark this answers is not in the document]
crazy2be: Correct, assuming you specify the timestamp in local time (`is_utc` is false). Should I make this explicit in the docs?

jerry-port/default/default-date.c
crazy2be: Correct, this is the default implementation. Most embedded targets won't even use this, or they will just use the `return 0;` default implementation, which should have the same "assume UTC" behaviour that they always had.

Windows does not support `tm.tm_gmtoff`. It's also not supported by mingw. However, it's fairly straightforward to shim it by using `GetTimeZoneInformationForYear` if someone wants to do this at some point. I'd rather not do it in this PR, as it's already gotten quite big (and it's a PITA for me to test on windows...)

jerry-port/default/CMakeLists.txt
```diff
@@ -26,6 +26,12 @@ file(GLOB SOURCE_PORT_DEFAULT *.c)
 # (should only be necessary if we used compiler default libc but not checking that)
 set(DEFINES_PORT_DEFAULT _BSD_SOURCE _DEFAULT_SOURCE)
 
+INCLUDE (CheckStructHasMember)
+CHECK_STRUCT_HAS_MEMBER ("struct tm" tm_gmtoff time.h HAVE_TM_GMTOFF)
+if(HAVE_TM_GMTOFF)
+  set(DEFINES_PORT_DEFAULT ${DEFINES_PORT_DEFAULT} HAVE_TM_GMTOFF)
+endif()
```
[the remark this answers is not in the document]
akosthekiss: I haven't investigated all the details of the problem but my guess is that the root of the problem is that `tm_gmtoff` is non-standard. See, for example, in glibc:
- `tm_gmtoff` is only declared if `__USE_MISC` is defined: https://sourceware.org/git/?p=glibc.git;a=blob;f=time/bits/types/struct_tm.h;h=b13b631228d0ec36691b25db2e1f9b1d66b54bb0;hb=HEAD#l19
- and it only defines `__USE_MISC` if `_DEFAULT_SOURCE` is defined: https://sourceware.org/git/?p=glibc.git;a=blob;f=include/features.h;h=5bed0a499605a3a26d55443f3c8b7e67de152f74;hb=HEAD#l368
- but when a compiler is instructed to compile in strict C99 mode then it will not pass extra macros to the preprocessor and thus will not "enable" such non-standard libc parts.

So, if `CHECK_STRUCT_HAS_MEMBER` calls the compiler with default settings, i.e., with no extra defines, then the compilation during the check will not see `tm_gmtoff`. Perhaps setting `CMAKE_REQUIRED_DEFINITIONS` to `_DEFAULT_SOURCE` before `CHECK_STRUCT_HAS_MEMBER` may solve the problem (see: https://cmake.org/cmake/help/v3.0/module/CheckStructHasMember.html). But even if this works, it may be glibc-specific.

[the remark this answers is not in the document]
crazy2be: I think it is something to do with my defines not getting passed. If I put a message() inside the if, it gets printed (at least, if I run `(cd jerry-port/default && cmake .)`). And compiling outside of cmake or any build system compiles fine without errors:
```c
#include <stdio.h>
#include <time.h>
#include <stdbool.h>

double jerry_port_get_local_time_zone_adjustment (double unix_ms,  /**< ms since unix epoch */
                                                  bool is_utc)  /**< is the time above in UTC? */
{
  struct tm tm;
  time_t now = (time_t) (unix_ms / 1000);
  localtime_r (&now, &tm);
  if (!is_utc)
  {
    now -= tm.tm_gmtoff;
    localtime_r (&now, &tm);
  }
  return ((double) tm.tm_gmtoff) * 1000;
} /* jerry_port_get_local_time_zone_adjustment */

int main() {
	double adj = jerry_port_get_local_time_zone_adjustment(0, 0);
	printf("%lf\n", adj);
}
```
(the above compiles without any warnings or errors with just `gcc test_gmtoff.c`)

I verified that it isn't trying to use `tm.tm_gmtoff` (and getting, say, bogus values) by putting a printf in `default-date.c` before each return statement. The one we hit is the one immediately before `return 0;`, so it's not even attempting to use the `tm.tm_gmtoff` path.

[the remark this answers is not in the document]
crazy2be: Hmm... looking at `CMakeCache.txt` after a clean build, maybe I'm wrong, and for some reason, when run as part of the larger project, `tm.tm_gmtoff` isn't found. 
image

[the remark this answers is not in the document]
crazy2be: Confirmed that when running `cmake .` in `jerry-port/default`, my `CMakeCache.txt` is generated with `HAVE_TM_GMTOFF:INTERNAL=1`, but when run as part of `tools/build.py`, the generated `CMakeCache.txt` (in `build/tests/test262_tests/CMakeCache.txt`) has `HAVE_TM_GMTOFF:INTERNAL=`, as above.

Looking at the two different `CMakeCache.txt` files, no glaring differences in build configuration stick out at me.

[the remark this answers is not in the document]
crazy2be: Ah, I see now what you were saying. I didn't realize jerryscript compiled in strict c99 mode (which, as you said, makes `tm.tm_gmtoff` invisible). `gcc -std=c99 test_gmtoff.c` does indeed fail as you predicted.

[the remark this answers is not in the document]
crazy2be: Huh, so the default source code that `check_struct_has_member` uses to check if the given struct has the given member actually _fails to compile_ with the default jerryscript compiler options... Ugh.
```
Performing C SOURCE FILE Test HAVE_TM_GMTOFF failed with the following output:
Change Dir: /home/fbuser/src/jerryscript/build/tests/test262_tests/CMakeFiles/CMakeTmp

Run Build Command:"/usr/bin/make" "cmTC_a80df/fast"
/usr/bin/make -f CMakeFiles/cmTC_a80df.dir/build.make CMakeFiles/cmTC_a80df.dir/build
make[1]: Entering directory '/home/fbuser/src/jerryscript/build/tests/test262_tests/CMakeFiles/CMakeTmp'
Building C object CMakeFiles/cmTC_a80df.dir/src.c.o
/usr/bin/cc  -D_DEFAULT_SOURCE  -flto -fno-fat-lto-objects -std=c99 -pedantic -fno-builtin -fno-stack-protector -Wall -Werror=all -Wextra -Werror=extra -Wformat-nonliteral -Werror=format-nonliteral -Winit-self -Werror=init-self -Wconversion -Werror=conversion -Wsign-conversion -Werror=sign-conversion -Wformat-security -Werror=format-security -Wmissing-declarations -Werror=missing-declarations -Wshadow -Werror=shadow -Wstrict-prototypes -Werror=strict-prototypes -Wundef -Werror=undef -Wold-style-definition -Werror=old-style-definition -Wno-stack-protector -Wno-attributes -Werror -Wlogical-op -Werror=logical-op -DHAVE_TM_GMTOFF   -o CMakeFiles/cmTC_a80df.dir/src.c.o   -c /home/fbuser/src/jerryscript/build/tests/test262_tests/CMakeFiles/CMakeTmp/src.c
/home/fbuser/src/jerryscript/build/tests/test262_tests/CMakeFiles/CMakeTmp/src.c:4:5: error: function declaration isn’t a prototype [-Werror=strict-prototypes]
 int main()
     ^
/home/fbuser/src/jerryscript/build/tests/test262_tests/CMakeFiles/CMakeTmp/src.c: In function ‘main’:
/home/fbuser/src/jerryscript/build/tests/test262_tests/CMakeFiles/CMakeTmp/src.c:4:5: error: old-style function definition [-Werror=old-style-definition]
cc1: all warnings being treated as errors
CMakeFiles/cmTC_a80df.dir/build.make:65: recipe for target 'CMakeFiles/cmTC_a80df.dir/src.c.o' failed
make[1]: *** [CMakeFiles/cmTC_a80df.dir/src.c.o] Error 1
make[1]: Leaving directory '/home/fbuser/src/jerryscript/build/tests/test262_tests/CMakeFiles/CMakeTmp'
Makefile:126: recipe for target 'cmTC_a80df/fast' failed
make: *** [cmTC_a80df/fast] Error 2

Source file was:

#include <time.h>

int main()
{
   (void)sizeof(((struct tm *)0)->tm_gmtoff);
   return 0;
}
```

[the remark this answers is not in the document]
crazy2be: Finally got it all to build. On the plus side, I'm a lot more familiar with cmake now :).

docs/05.PORT-API.md
```diff
@@ -198,27 +205,21 @@ jerry_port_log (jerry_log_level_t level, /**< log level */
...
-  struct timezone tz;
-
-  /* gettimeofday may not fill tz, so zero-initializing */
-  tz.tz_minuteswest = 0;
-  tz.tz_dsttime = 0;
-
-  if (gettimeofday (&tv, &tz) != 0)
+  struct tm tm;
+  time_t now = (time_t) (unix_ms / 1000);
+  localtime_r (&now, &tm);
```
akosthekiss: both `struct tm` and `localtime_r` are declared in `time.h`, which should be listed as `#include`d at the beginning of the code example

crazy2be: Done.

crazy2be:
@rerobika: Those tests have been enabled, and pass.
@akosthekiss: I addressed your nit.
@zherczeg: I updated the docs to make the answers to your questions more explicit.

The travis-ci build seems to have hung on the signed-off-by check, I re-triggered it which should make the build go green. Everything else is passing.

jerry-port/default/default-date.c
```diff
@@ -13,6 +13,9 @@
  * limitations under the License.
  */
 
+#ifdef HAVE_TM_GMTOFF
+#include <time.h>
+#endif
```
rerobika: Please add a closing comment for `#endif`.

```
#endif /* HAVE_TM_GMTOFF */
```

jerry-port/default/default-date.c
```diff
@@ -13,6 +13,9 @@
...
 #ifdef __GNUC__
 #include <sys/time.h>
 #endif
```
rerobika: This patch does not affect this code path, but I would be great to unify the closing `#endif` comments as well.

jerry-port/default/default-date.c
```diff
@@ -21,33 +24,32 @@
...
-    return true;
+    now -= tm.tm_gmtoff;
+    localtime_r (&now, &tm);
   }
-#endif /* __GNUC__ */
-
-  return false;
-} /* jerry_port_get_time_zone */
+  return ((double) tm.tm_gmtoff) * 1000;
+#else
```
rerobika: Please add a comment here as well.

```
#else /* !HAVE_TM_GMTOFF */
```

tests/jerry/date-setters.js
```diff
@@ -82,25 +82,24 @@ assert (d.getUTCSeconds() == 1);
...
-//assert (d.getHours() == 1);
-//assert (d.getMinutes() == 1);
-//assert (d.getSeconds() == 1);
-//d.setTime(0);
-//assert (d.setHours(1, 1, 1, 1) == hour + min + sec + ms + d.getTimezoneOffset() * 60000);
-//assert (d.getHours() == 1);
-//assert (d.getMinutes() == 1);
-//assert (d.getSeconds() == 1);
-//assert (d.getMilliseconds() == 1);
+d.setTime(0);
```
rerobika: There are several `FIXME:`s in [date-getters.js](https://github.com/jerryscript-project/jerryscript/blob/master/tests/jerry/date-getters.js#L28)  as well due to the timezone adjustment. Could you check them please?

crazy2be: Done, and they pass (with some changes, the original tests were slightly incorrect).

PR #2547, opened 2018-10-01
Rework op_construct.

zherczeg:
The new form should be more suitable for ES5 class implementation. No binary / perf / mem change.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -700,86 +700,6 @@ ecma_op_function_call (ecma_object_t *func_obj_p, /**< Function object */
 } /* ecma_op_function_call */
 
 /**
- * [[Construct]] implementation for Function objects (13.2.2),
- * created through 13.2 (ECMA_OBJECT_TYPE_FUNCTION) and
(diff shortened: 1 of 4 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -874,13 +762,117 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
...
+
+    ecma_free_value (prototype_prop_value);
+
+    this_arg_value = ecma_make_object_value (new_this_obj_p);
+  }
+
+  /* 8. */
+  ecma_value_t ret_value;
+
+  switch (type)
```
LaszloLango: An `if-else` statement fits better here than a `switch`

rerobika: Currently it fits better but for further [cases](https://github.com/jerryscript-project/jerryscript/pull/2439/files#diff-c790ca6e2b99859fd40e7cce2885c2fbR1068) switch is more suitable.

LaszloLango: OK, leave as it is.

LaszloLango:
LGTM

PR #2550, opened 2018-10-04
Remove the JERRY_SNAPSHOT_FOUR_BYTE_CPOINTER snapshot flag.

rtakacs:
Removed the `JERRY_SNAPSHOT_FOUR_BYTE_CPOINTER` flag because that is not necessary. Also updated the snapshot version.

### DIFF
--- a/jerry-core/api/jerry-snapshot.h
+++ b/jerry-core/api/jerry-snapshot.h
@@ -52,7 +52,7 @@ typedef enum
   JERRY_SNAPSHOT_HAS_REGEX_LITERAL = (1u << 0), /**< byte code has regex literal */
   JERRY_SNAPSHOT_HAS_CLASS_LITERAL = (1u << 1), /**< byte code has class literal */
   /* 24 bits are reserved for compile time features */
-  JERRY_SNAPSHOT_FOUR_BYTE_CPOINTER = (1u << 8) /**< compressed pointers are four byte long */
+  JERRY_SNAPSHOT_FOUR_BYTE_CPOINTER = (1u << 8) /**< deprecated, an unused placeholder now */
 } jerry_snapshot_global_flags_t;
 
 #endif /* !JERRY_SNAPSHOT_H */
--- a/jerry-core/api/jerry-snapshot.c
+++ b/jerry-core/api/jerry-snapshot.c
@@ -42,9 +42,6 @@ snapshot_get_global_flags (bool has_regex, /**< regex literal is present */
 
   uint32_t flags = 0;
 
-#ifdef JERRY_CPOINTER_32_BIT
-  flags |= JERRY_SNAPSHOT_FOUR_BYTE_CPOINTER;
-#endif /* JERRY_CPOINTER_32_BIT */
 #ifndef CONFIG_DISABLE_REGEXP_BUILTIN
   flags |= (has_regex ? JERRY_SNAPSHOT_HAS_REGEX_LITERAL : 0);
 #endif /* !CONFIG_DISABLE_REGEXP_BUILTIN */


galpeter:
Shouldn't there be a CMakeLists modification also?

jerry-core/api/jerry-snapshot.h
```diff
@@ -41,7 +41,7 @@ typedef struct
 /**
  * Jerry snapshot format version.
  */
-#define JERRY_SNAPSHOT_VERSION (18u)
+#define JERRY_SNAPSHOT_VERSION (19u)
```
rerobika: Since the snapshot version number has increased, [this](https://github.com/jerryscript-project/jerryscript/blob/master/tests/unit-core/test-snapshot.c#L226) expected output should be modified as well.

jerry-core/api/jerry-snapshot.h
```diff
@@ -51,8 +51,6 @@ typedef enum
   /* 8 bits are reserved for dynamic features */
   JERRY_SNAPSHOT_HAS_REGEX_LITERAL = (1u << 0), /**< byte code has regex literal */
   JERRY_SNAPSHOT_HAS_CLASS_LITERAL = (1u << 1), /**< byte code has class literal */
-  /* 24 bits are reserved for compile time features */
-  JERRY_SNAPSHOT_FOUR_BYTE_CPOINTER = (1u << 8) /**< compressed pointers are four byte long */
```
akosthekiss: Is it OK to simply drop this enumerator? In other public API headers (e.g., jerryscript-core.h), there are examples to keep deprecated enumerators with original values, with a comment stating that it's "deprecated, an unused placeholder now" (see `jerry_init_flag_t`).

jerry-core/api/jerry-snapshot.h
akosthekiss: Is this really a change in the snapshot version at all?

jerry-core/api/jerry-snapshot.h
rtakacs: Thank you for the clarification. I've modified the patch and added a deprecated marker to this entry.

jerry-core/api/jerry-snapshot.h
rtakacs: @rerobika @akiss You are right, the expected output of the snapshot test also should be modified. Thanks.

rtakacs:
@galpeter There is nothing in the CMake files for the `JERRY_SNAPSHOT_FOUR_BYTE_CPOINTER`. So the CMake files should not be modified.

jerry-core/api/jerry-snapshot.h
LaszloLango: The format of snapshots are not changed in this patch. I agree with @akosthekiss, no need to change the snapshot version.

akosthekiss: @rtakacs two notes:
- the intent of my original comment was to suggest that the version should _not_ be changed (as @LaszloLango also pointed out)
- on these forums, `@akiss` is _not_ me: it's a project/tool ran by INRIA (France). here, I'm `@akosthekiss` ;)

akosthekiss:
Note to ourselves in the future: `JERRY_SNAPSHOT_VERSION`, the 4th byte of the snapshot format, is increasing somewhat fast. Which is not a bad thing per se. But it is "already" around `0x12` ~ `0x13`. At least 237 (236?) increments later, i.e., when we reach `0xFF`, we'll have to change the snapshot format and stick that 4th byte to `0xFF` so that it doesn't flip back to 0 at the next increment and cause potential problems; and then we'll have to introduce some other byte(s) in the format to store the file format version from that point on. I'd suggest we don't wait for that so long but do the format change when we reach `0x53` (at the latest) and make that byte and the following 3 bytes also part of the magic string of the format, fixating them to `0x4E 0x41 0x50`, thus making the first 8 bytes read `'JRRYSNAP'`. (And add a non-1-byte version field after the magic. 4 bytes could then last "forever".)

zherczeg:
I don't really understand this. We only need to worry if version 4 billion is reached, since we have 4 bytes for version number.

akosthekiss:
Oh, do we? I must have misinterpreted something. My bad. Scratch the above then.

zherczeg:
Maybe the little endian format is the root of confusion. So the first byte is the last byte of the version number. The little/big endian can be detected from the magic, which is JRRY of little and YRRJ for big endian systems. The version number bytes are ordered accordingly.

akosthekiss:
Not to be missed because of the noise generated by me: @rtakacs I still think that this patch should not increase the snapshot version number.

rtakacs:
@akosthekiss I've changed back the snapshot version to 18.

PR #2551, opened 2018-10-04
Ensure that tests are always executed in a proper time zone

akosthekiss:
The America/Los_Angeles time zone is already enforced for Travis CI
jobs but that doesn't guarantee the correctness of locally executed
tests. So, this patch moves the setting of the `TZ` environment
variable from `.travis.yml` to `run-tests.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,6 @@ matrix:
     - env:
         - JOBNAME="Conformance Tests"
         - OPTS="--test262"
-        - TZ=America/Los_Angeles
 
     - env:
         - JOBNAME="ASAN Tests"
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -199,8 +199,11 @@ BINARY_CACHE = {}
 TERM_NORMAL = '\033[0m'
 TERM_BLUE = '\033[1;34m'
 
-def report_command(cmd_type, cmd):
+def report_command(cmd_type, cmd, env=None):
     sys.stderr.write('%s%s%s\n' % (TERM_BLUE, cmd_type, TERM_NORMAL))
+    if env is not None:
+        sys.stderr.write(''.join('%s%s=%r \\%s\n' % (TERM_BLUE, var, val, TERM_NORMAL)
+                                 for var, val in sorted(env.items())))
     sys.stderr.write('%s%s%s\n' % (TERM_BLUE, (' \\%s\n\t%s' % (TERM_NORMAL, TERM_BLUE)).join(cmd), TERM_NORMAL))
 
 def create_binary(job, options):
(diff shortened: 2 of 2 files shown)


crazy2be:
Just curious - it looks like several of the jerryscript date tests (`python tools/run-tests.py --jerry-tests`) also rely on being in a particular timezone. Running them in my local timezone, all 4 date related tests fail. I was hoping this patch would fix that, but I still have the same 4 tests failing. Changing my computer timezone to California time, the same 4 tests still fail. Changing my computer timezone to GMT, I only fail the `date-setters.js` test. However, presumably there is _some_ configuration that causes these tests to all pass, since they appear to pass on travis-ci.

akosthekiss:
@crazy2be What's your setup? I just tried `TZ='Europe/London' tools/run-tests.py --jerry-tests` on macOS 10.13.6 (Darwin 17.7.0 x86_64) and on Ubuntu 18.04.1 LTS (Linux 4.15.0 x86_64), and got all tests passing.

And BTW, did the test262 tests pass for you with this patch?

crazy2be:
Hmm, interesting. All tests appear to pass for me as well on `master`, but `date_setters.js` definitely fails for me on my timezone fixing branch (#2540). I'm on OSX 10.13.2 (Darwin 17.3.0 x86_64). It isn't surprising this test would start failing, but I'm surprised it's not failing on travis-ci...

The relevant thought I had with regards to this PR is that we should make the test-runner set the correct timezone for the jerry-tests, since we are now doing it for the test262 tests (as of this PR!). Then people in other timezones (like me!) can just run the tests without expecting failures, or having to remember the correct timezone for each suite of tests.

akosthekiss:
I think it should be possible to lock the TZ for the other tests as well. I'll update this patch later tonight. I'll try to set TZ to Europe/Budapest. One is allowed to be just a bit patriotic, right? :)

crazy2be:
Ok, I figured out the `date-setters.js` problem. The following command makes all tests execute successfully (I can even uncomment the dst ones, and they work too!):
```
TZ='UTC' tools/run-tests.py --jerry-tests
```
However, now that we have proper timezone/dst support, any other timezone causes failures. This makes sense, since the part that is failing is `Date.setFullYear`, which sets the year (in local time, as per the spec), and then out tests assert that the returned ms value (which is in UTC time) is a particular value:
```
assert (d.setFullYear(0, 0) == -62167219200000);
```
This shouldn't be expected to work in any timezone other than the one used to retrieve the UTC ms value, which clearly was UTC, given
```
assert (d.setUTCFullYear(0, 0) == -62167219200000);
```

Thus, we can (and should!) make sure these tests are always run in a particular timezone. However, if that timezone is anything other than `UTC`, it will require some additional adjustments to the tests in #2540 in order to retrieve the correct values for the given timezone.

crazy2be:
I realize I never answered your second question. Yes, this patch does in fact make it so that the test262 tests pass on my system, without having to change my timezone manually.
```
ch15/15.9/15.9.3/S15.9.3.1_A5_T1 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T2 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T3 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T4 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T5 passed in non-strict mode
ch15/15.9/15.9.3/S15.9.3.1_A5_T6 passed in non-strict mode
```
```
=== Summary ===
 - Ran 11552 tests
 - All tests succeeded
```
With just the command line
```
tools/run-tests.py --test262
```
Awesome!

akosthekiss:
PR updated. Proper time zone is enforced for both test262 and jerry tests.

PR #2556, opened 2018-10-08
Move version numbers to public headers

Achie72:
JERRY_SNAPSHOT_VERSION and JERRY_DEBUGGER_VERSION were moved into public headers, to grant access to them.

### DIFF
--- a/jerry-core/api/jerry-snapshot.h
+++ b/jerry-core/api/jerry-snapshot.h
@@ -39,11 +39,6 @@ typedef struct
 #define JERRY_SNAPSHOT_MAGIC (0x5952524Au)
 
 /**
- * Jerry snapshot format version.
- */
-#define JERRY_SNAPSHOT_VERSION (18u)
-
-/**
  * Snapshot configuration flags.
  */
 typedef enum
--- a/jerry-core/debugger/debugger.h
+++ b/jerry-core/debugger/debugger.h
@@ -24,11 +24,6 @@
 /* JerryScript debugger protocol is a simplified version of RFC-6455 (WebSockets). */
 
 /**
- * JerryScript debugger protocol version.
- */
-#define JERRY_DEBUGGER_VERSION (6)
-
-/**
  * Frequency of calling jerry_debugger_receive() by the VM.
  */
 #define JERRY_DEBUGGER_MESSAGE_FREQUENCY 5
--- a/jerry-core/include/jerryscript-debugger.h
+++ b/jerry-core/include/jerryscript-debugger.h
@@ -29,6 +29,11 @@ extern "C"
  */
 
 /**
+ * JerryScript debugger protocol version.
+ */
+#define JERRY_DEBUGGER_VERSION (6)
+
+/**
  * Types for the client source wait and run method.
  */
 typedef enum
--- a/jerry-core/include/jerryscript-snapshot.h
+++ b/jerry-core/include/jerryscript-snapshot.h
@@ -28,6 +28,11 @@ extern "C"
  */
 
 /**
+ * Jerry snapshot format version.
+ */
+#define JERRY_SNAPSHOT_VERSION (18u)
+
+/**
  * Flags for jerry_generate_snapshot and jerry_generate_function_snapshot.
  */
 typedef enum


docs/02.API-REFERENCE.md
```diff
@@ -474,6 +474,14 @@ Possible values:
 API functions can return the `JERRY_TYPEDARRAY_INVALID` value if the
 TypedArray support is not in the engine.
 
+## jerry_versions_t
+
+Enum that contains the following elements:
+
+  - JERRY_VERSION_API_MAJOR - signals the major API version
```
galpeter: What if we would change the "signals" word, to "represents" / "describes"?

docs/02.API-REFERENCE.md
```diff
@@ -474,6 +474,14 @@ Possible values:
...
+  - JERRY_VERSION_API_MINOR - signals the minor API version
+  - JERRY_VERSION_DEBUGGER - signals the debugger version
+  - JERRY_VERSION_SNAPSHOT - signals the snapshot version
```
galpeter: As these values are only used for one method we should add a "see also" block to refer the method which uses this.

docs/02.API-REFERENCE.md
```diff
@@ -5523,6 +5531,63 @@ jerry_get_backtrace (uint32_t max_depth);
...
+
+```c
+uint32_t
+jerry_get_api_minor_version(void);
+```
+
+- return value
+  - API minor version
+
+## jerry_get_component_version
```
galpeter: If we have this method, why do we need the previous two? (get_api_minor & get_api_major)?

docs/02.API-REFERENCE.md
```diff
@@ -5523,6 +5531,63 @@ jerry_get_backtrace (uint32_t max_depth);
...
+
+**Summary**
+
+Get the component version described in parameter. If the defined components
+asked is not available, then it returns 0, as well as a default value.
```
galpeter: What do you mean by: `..it returns 0, as well as a default value. `? This will return a zero and another value?

docs/02.API-REFERENCE.md
```diff
@@ -5523,6 +5531,63 @@ jerry_get_backtrace (uint32_t max_depth);
...
+
+**Prototype**
+
+```c
+uint32_t
+jerry_get_component_version(jerry_versions_t component);
+```
+- `component` - A jerry_versions_t signaling desired component.
+- return value
+  - API major version
```
galpeter: AFAIK: this is a typo here: the return value is the version of the requested component.

docs/02.API-REFERENCE.md
```diff
@@ -5523,6 +5531,63 @@ jerry_get_backtrace (uint32_t max_depth);
...
+Get the component version described in parameter. If the defined components
+asked is not available, then it returns 0, as well as a default value.
+
+**Prototype**
+
+```c
+uint32_t
+jerry_get_component_version(jerry_versions_t component);
+```
+- `component` - A jerry_versions_t signaling desired component.
```
galpeter: Maybe a better text would be: "The component for which the version info is requested". What do you think?

docs/02.API-REFERENCE.md
zherczeg: comment: returns the...
Please also modify the others.

docs/02.API-REFERENCE.md
```diff
@@ -5523,6 +5531,63 @@ jerry_get_backtrace (uint32_t max_depth);
...
+
+**Summary**
+
+Get the major API version.
+
+**Prototype**
+
+```c
+uint32_t
+jerry_get_api_major_version(void);
```
zherczeg: These functions are redundant, aren't they?

docs/02.API-REFERENCE.md
```diff
@@ -5523,6 +5531,63 @@ jerry_get_backtrace (uint32_t max_depth);
...
+```
+- `component` - A jerry_versions_t signaling desired component.
+- return value
+  - API major version
+
+**Example**
+
+```c
+uint32_t api_version;
+api_version = jerry_get_component_version(API_MAJOR_VERSION);
```
galpeter: There is no `API_MAJOR_VERSION` described in the `jerry_version_t` enum.

docs/02.API-REFERENCE.md
```diff
@@ -5558,6 +5623,7 @@ jerry_get_arraybuffer_byte_length (const jerry_value_t value);
 ```
 
 **See also**
+
```
galpeter: Is this empty line needed here?

jerry-core/api/jerry.c
```diff
@@ -2759,6 +2759,81 @@ jerry_get_backtrace (uint32_t max_depth) /**< depth limit of the backtrace */
...
+jerry_get_api_minor_version (void)
+{
+  return JERRY_API_MINOR_VERSION;
+} /* jerry_get_api_minor_version */
+
+/**
+ * Get the component version described in parameter.
+ *
+ * Note:
+ *    * Debugger and snapshot versions are set to undefined, unless they are turned on.
```
galpeter: The comment states that the "debugger/snapshot version are set to undefined" if they are turned off. But in the code the return value will be 0. That looks pretty defined for me....

jerry-core/api/jerry.c
```diff
@@ -2759,6 +2759,81 @@ jerry_get_backtrace (uint32_t max_depth) /**< depth limit of the backtrace */
...
+ *
+ * @return a uint32_t containing:
+ *         API major, API minor, debugger and snapshot version.
```
galpeter: So.. this will return all of these values? in one uint32_t?

jerry-core/api/jerry.c
```diff
@@ -2759,6 +2759,81 @@ jerry_get_backtrace (uint32_t max_depth) /**< depth limit of the backtrace */
...
+ */
+uint32_t
+jerry_get_component_version (jerry_version_t component)  /**< the component, the version of which we want */
```
galpeter: This comment reads a bit awkward.

Achie72:
It could help, if we only have an binary file, to confirm, which version of jerry is it exactly, to avoid confusion.

zherczeg:
He is probably talking about dynamic modules, where you load a module and try to confirm the module is what you expected.

akosthekiss:
Dynamic libraries are also a tricky thing. If one knows the API version of the library (the major version at least) then one may know what functions can be expected to exist (at least the minimum subset). But that means that one already has to know the value of `JERRY_API_MAJOR_VERSION` even to call `jerry_get_component_version (JERRY_VERSION_API_MAJOR)` (because only at a given major version would it be guaranteed to exist). Almost a deadlock.

Moreover, although using jerry-core as a dynamic library is certainly possible, the device class where that's possible is pretty much not the main application area of the engine.

Achie72:
The moving of `JERRY_SNAPSHOT_VERSION` and `JERRY_DEBUGGER_VERSION` indeed seems a more simple solution, than the getter functions. If we don't want to complicate the code by introducing these changes, then i can modify the PR with your suggestion, if the other reviewers have no problem with it. @zherczeg @galpeter ?

zherczeg:
I have no strong opinion about this, so whatever is good for me.

PR #2557, opened 2018-10-08
List scope chain levels and their variables to the selected stack frame.

robertsipka:
It supports to list the scope chain of the current execution context and see which variables are available.

Please note that this is not a complete list, virtual properties are not included. I plan to expand them in a follow-up patch.

An example run should look like this:

test.js
```javascript

1 var c = 4;
2 var z = 5;
3 
4 function addX(x) {
5  return function(n) {
6     var b = 2;
7     return n + x;
8   }
9 }
10 addThree = addX(3);
11 d = addThree(c+z);
12 
13 function f() {
14   var x = true;
15   var g = 0;
16   (function() {
17     var a = [1,2,3]
18     a.y = "abc";
20     with (a) {
21       var h = [4,5,6]
22       with (h) {
23         h.d = "dfg"
24       }
25       a.d = x + g;
26     }
27   })();
28 }
29 
30 f();
```

```bash
$ ./build/bin/jerry --start-debug-server test.js
```
```bash
$ ./jerry-debugger/jerry_client.py
```

```bash
Connecting to: localhost:5001
Stopped at test.js:1
(jerry-debugger) b test.js:21
Breakpoint 1 at test.js:21 (in function() at line:16, col:4)
(jerry-debugger) c
Press enter to stop JavaScript execution.
Stopped at breakpoint:1 test.js:21 (in function() at line:16, col:4)
(jerry-debugger) scopes
level | type    
0     | with    
1     | local   
2     | closure 
3     | global  
(jerry-debugger) variables 0
name | type   | value 
y    | String | abc   
2    | Number | 3     
1    | Number | 2     
0    | Number | 1     
(jerry-debugger) variables 1
name | type  | value   
h    | Array | [4,5,6] 
a    | Array | [1,2,3] 
(jerry-debugger) variables 2
name | type    | value 
g    | Number  | 0     
x    | Boolean | true  
(jerry-debugger) variables 3
name     | type     | value                        
d        | Number   | 12                           
addThree | Function | function(){/* ecmascript */} 
f        | Function | function(){/* ecmascript */} 
addX     | Function | function(){/* ecmascript */} 
z        | Number   | 5                            
c        | Number   | 4                            
print    | Function | function(){/* ecmascript */} 
gc       | Function | function(){/* ecmascript */} 
assert   | Function | function(){/* ecmascript */} 
(jerry-debugger)
```

### DIFF
--- a/jerry-core/debugger/debugger.c
+++ b/jerry-core/debugger/debugger.c
@@ -18,6 +18,7 @@
 #include "ecma-builtin-helpers.h"
 #include "ecma-conversion.h"
 #include "ecma-eval.h"
+#include "ecma-function-object.h"
 #include "ecma-objects.h"
 #include "jcontext.h"
 #include "jerryscript-port.h"
@@ -37,9 +38,9 @@ typedef struct
  * The number of message types in the debugger should reflect the
  * debugger versioning.
  */
-JERRY_STATIC_ASSERT (JERRY_DEBUGGER_MESSAGES_OUT_MAX_COUNT == 28
-                     && JERRY_DEBUGGER_MESSAGES_IN_MAX_COUNT == 19
-                     && JERRY_DEBUGGER_VERSION == 6,
+JERRY_STATIC_ASSERT (JERRY_DEBUGGER_MESSAGES_OUT_MAX_COUNT == 32
+                     && JERRY_DEBUGGER_MESSAGES_IN_MAX_COUNT == 21
+                     && JERRY_DEBUGGER_VERSION == 7,
                      debugger_version_correlates_to_message_type_count);
 
 /**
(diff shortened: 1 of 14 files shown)


jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,305 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+        level_type = JERRY_DEBUGGER_SCOPE_GLOBAL;
+      }
+      else
+      {
+        level_type = JERRY_DEBUGGER_SCOPE_WITH;
+        next_func_is_local = true;
+      }
+    }
+
+    memcpy (message_type_p->string + buffer_pos++, &level_type, 1);
```
zherczeg: You don't need memcpy to copy 1 byte. To prepare classes ignore non-declarative and non-object-bound lex envs.

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,305 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+static bool
+copy_scope_variables_to_string_message (jerry_debugger_scope_variable_type_t variable_type, /**< variable type */
+                                        ecma_string_t *value_str, /**< property name or value string */
+                                        jerry_debugger_send_string_t *message_string_p, /**< message pointer */
+                                        size_t *buffer_pos) /**< string data position of the message */
+{
+  const size_t max_byte_count = JERRY_DEBUGGER_SEND_MAX (uint8_t);
+  const size_t max_message_size = JERRY_DEBUGGER_SEND_SIZE (max_byte_count, uint8_t);
+
+  ECMA_STRING_TO_UTF8_STRING (value_str, str_buff, str_buff_size);
```
zherczeg: Is this a cesu8 string or some costly conversion?

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,305 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+  size_t str_size = 0;
+  size_t str_limit = 255;
+
+  while (copied_values > 0)
+  {
+    if (*buffer_pos == max_byte_count)
+    {
+      if (!jerry_debugger_send (max_message_size))
+      {
+        ECMA_FINALIZE_UTF8_STRING (str_buff, str_buff_size);
```
zherczeg: I have never seen such a macro used for early return. They should behave block like.

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,305 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+  while (chain_index != 0)
+  {
+    lex_env_p = ecma_get_lex_env_outer_reference (lex_env_p);
+
+    if (lex_env_p == NULL)
+    {
+      jerry_debugger_send_type (JERRY_DEBUGGER_SCOPE_VARIABLES_END);
+      return;
+    }
+    chain_index--;
```
zherczeg: Ignore non this object or declarative lex envs as well.

tests/debugger/do_variables.expected
```diff
@@ -0,0 +1,122 @@
...
+(jerry-debugger) scopes
+level | type   
+0     | local  
+1     | global 
+(jerry-debugger) variables 1
+name     | type     | value                        
+d        | Number   | 12                           
+addThree | Function | function(){/* ecmascript */} 
+f        | Function | function(){/* ecmascript */} 
+addX     | Function | function(){/* ecmascript */}
```
zherczeg: We should send no values for functions.

tests/debugger/do_variables.expected
```diff
@@ -0,0 +1,122 @@
...
+Stopped at tests/debugger/do_variables.js:15
+(jerry-debugger) scopes
+level | type   
+0     | global 
+(jerry-debugger) variables
+name   | type      | value                        
+f      | Function  | function(){/* ecmascript */} 
+addX   | Function  | function(){/* ecmascript */} 
+z      | undefined | undefined                    
+c      | undefined | undefined
```
zherczeg: What about objects?

jerry-core/debugger/debugger.c
robertsipka: It is the name or value of the property, So, it is always returns with the cesu8 character array of a string.

jerry-core/debugger/debugger.c
robertsipka: I do not understand exactly what you mean. It will steps through the outer references of the current lexical environment until the `chain_index` arguments reach zero. It is always touch a declarative or this object bound lexical environment. If not, it will returns with an empty message with `JERRY_DEBUGGER_SCOPE_VARIABLES_END` type (e.g.: the `chain_index` argument is invalid / bigger than the scope chain).

tests/debugger/do_variables.expected
robertsipka: You are right, it makes sense. I've updated it accordingly.

tests/debugger/do_variables.expected
robertsipka: This really missed. I've added.

jerry-core/debugger/debugger.c
zherczeg: In the future classes will introduces a new scope type. This type should not be shared by the debugger client side, and when encountered it should be skipped. Therefore only the two currently supported types should be handled by this loops.

robertsipka: Oh I see, thanks. I've updated it.

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,333 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+      JERRY_ASSERT (ecma_get_lex_env_type (lex_env_p) == ECMA_LEXICAL_ENVIRONMENT_THIS_OBJECT_BOUND);
+
+      if (ecma_get_lex_env_outer_reference (lex_env_p) == NULL)
+      {
+        level_type = JERRY_DEBUGGER_SCOPE_GLOBAL;
+      }
+      else
+      {
+        level_type = JERRY_DEBUGGER_SCOPE_WITH;
+        next_func_is_local = true;
```
zherczeg: Why this value is set to true here? I might misunderstood its purpose.

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,333 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+        next_func_is_local = false;
+      }
+    }
+    else
+    {
+      JERRY_ASSERT (ecma_get_lex_env_type (lex_env_p) == ECMA_LEXICAL_ENVIRONMENT_THIS_OBJECT_BOUND);
+
+      if (ecma_get_lex_env_outer_reference (lex_env_p) == NULL)
+      {
+        level_type = JERRY_DEBUGGER_SCOPE_GLOBAL;
```
zherczeg: A break could be added here, and change the loop to while(true)

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,333 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+ * Get type of the scope variable property.
+ */
+static jerry_debugger_scope_variable_type_t
+ecma_get_typeof_scope_variable (ecma_value_t value) /**< input ecma value */
+{
+  jerry_debugger_scope_variable_type_t ret_value = JERRY_DEBUGGER_VARIABLE_NONE;
+
+  if (ecma_is_value_undefined (value))
+  {
+    ret_value = JERRY_DEBUGGER_VARIABLE_UNDEFINED;
```
zherczeg: It might be nitpicking, but I think VALUE is better than VARIABLE, since the value of a variable is undefined, rather than the variable, which would mean it does not exist.

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,333 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+      if (!jerry_debugger_send (max_message_size))
+      {
+        result = false;
+        break;
+      }
+
+      *buffer_pos = 0;
+    }
+
+    if (copied_values == 3)
```
zherczeg: This sounds like a state for a switch. Furthermore the last case is handled specially, I would take it out from this loop and done separately, since it already has a copy loop.

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,333 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+  memcpy (&chain_index, get_scope_variables_p->chain_index, sizeof (uint32_t));
+
+  vm_frame_ctx_t *iter_frame_ctx_p = JERRY_CONTEXT (vm_top_context_p);
+  ecma_object_t *lex_env_p = iter_frame_ctx_p->lex_env_p;
+
+  while (chain_index != 0)
+  {
+    lex_env_p = ecma_get_lex_env_outer_reference (lex_env_p);
+
+    if (lex_env_p == NULL)
```
zherczeg: JERRY_UNLIKELY

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,333 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+
+  while (prop_iter_p != NULL)
+  {
+    JERRY_ASSERT (ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
+
+    ecma_property_pair_t *prop_pair_p = (ecma_property_pair_t *) prop_iter_p;
+
+    for (int i = 0; i < ECMA_PROPERTY_PAIR_ITEM_COUNT; i++)
+    {
+      if (ECMA_PROPERTY_IS_NAMED_PROPERTY (prop_iter_p->types[i]))
```
zherczeg: I am not sure all names are printable. I think Promises has some internal properties with special magic string names. They should be ignored.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -656,6 +656,11 @@ typedef enum
  */
 #define ECMA_OBJECT_FLAG_EXTENSIBLE 0x20
 
+/**
+ * Catch clause.
+ */
+#define ECMA_OBJECT_FLAG_CATCH_CLAUSE 0x20
```
zherczeg: I would call it local scope, since we might add let/const at some day and they would be marked this flag as well.

robertsipka:
@zherczeg : Thanks for the review, I've updated the patch based on your suggestions.

jerry-core/vm/vm.c
```diff
@@ -2953,6 +2953,7 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
           }
 
           ecma_object_t *catch_env_p = ecma_create_decl_lex_env (frame_ctx_p->lex_env_p);
+          catch_env_p->type_flags_refs = (uint16_t) (catch_env_p->type_flags_refs | ECMA_OBJECT_FLAG_NON_CLOSURE);
```
akosthekiss: Two questions:
1. Is this code debugger-specific? If so, it should be wrapped into `#ifdef JERRY_DEBUGGER` / `#endif /* JERRY_DEBUGGER */`.
2. Isn't this equivalent to `catch_env_p->type_flags_refs = (uint16_t) ECMA_OBJECT_FLAG_NON_CLOSURE;`?

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,332 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+    else
+    {
+      ret_value = ecma_op_is_callable (value) ? JERRY_DEBUGGER_VALUE_FUNCTION : JERRY_DEBUGGER_VALUE_OBJECT;
+    }
+  }
+
+  JERRY_ASSERT (ret_value != JERRY_DEBUGGER_VALUE_NONE);
+
+  return ret_value;
+} /* ecma_get_typeof_scope_variable */
```
akosthekiss: The names of the new static functions are somewhat inconsistent. Most are prefixed with `jerry_debugger_` but this one here is `ecma_get_typeof_scope_variable ` while the next one is simply `copy_scope_variables_to_string_message` without any obvious prefix. Is this intentional?

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,332 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+      }
+
+      buffer_pos = 0;
+    }
+
+    if (ecma_get_lex_env_type (lex_env_p) == ECMA_LEXICAL_ENVIRONMENT_DECLARATIVE)
+    {
+      if ((lex_env_p->type_flags_refs & ECMA_OBJECT_FLAG_NON_CLOSURE) != 0)
+      {
+        message_type_p->string[buffer_pos++] = JERRY_DEBUGGER_SCOPE_CATCH;
```
zherczeg: Please don\t call this catch.

jerry-core/debugger/debugger.c
```diff
@@ -195,6 +196,332 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
...
+    {
+      memcpy (message_string_p->string + *buffer_pos, string_p, str_size);
+      *buffer_pos += str_size;
+    }
+  }
+
+  ECMA_FINALIZE_UTF8_STRING (str_buff, str_buff_size);
+
+  return result;
+} /* jerry_debugger_copy_variables_to_string_message */
```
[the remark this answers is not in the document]
robertsipka: Thanks, I've updated it, Travis should be green now.

PR #2558, opened 2018-10-08
Skip `-m32` build-test on non-Linux non-x86 platforms

renatahodovan:
The `-m32/64` compiler flags are x86/64-specific (mostly; at least in the
context of JerryScript-supported architectures) and macOS has deprecated 32-bit
support, so it's better to keep the build-testing of `-m32` to x86[-64]/Linux
only.

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -20,14 +20,18 @@ import argparse
 import collections
 import hashlib
 import os
+import platform
 import subprocess
 import sys
 import settings
 
 OUTPUT_DIR = os.path.join(settings.PROJECT_DIR, 'build', 'tests')
 
-Options = collections.namedtuple('Options', ['name', 'build_args', 'test_args'])
-Options.__new__.__defaults__ = ([], [])
+Options = collections.namedtuple('Options', ['name', 'build_args', 'test_args', 'skip'])
+Options.__new__.__defaults__ = ([], [], False)
+
+def skip_if(condition, desc):
+    return desc if condition else False
 
 OPTIONS_PROFILE_MIN = ['--profile=minimal']
 OPTIONS_PROFILE_ES51 = [] # NOTE: same as ['--profile=es5.1']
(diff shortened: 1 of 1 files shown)


renatahodovan:
On Darwin 17.7.0 x86_64 with gcc mimicked by clang Apple LLVM version 10.0.0 (clang-1000.11.45.2), I get the following error:

```
Build command:
tools/build.py \
	--compile-flag=-m32 \
	--cpointer-32bit=on \
	--system-allocator=on \
	--builddir=build/tests/buildoption_test-cpointer_32bit \
	--install=build/tests/buildoption_test-cpointer_32bit/local
ld: warning: The i386 architecture is deprecated for macOS (remove from the Xcode build setting: ARCHS)
```

tools/run-tests.py
```diff
@@ -29,6 +30,11 @@
 Options = collections.namedtuple('Options', ['name', 'build_args', 'test_args'])
 Options.__new__.__defaults__ = ([], [])
 
+SkippedOptions = collections.namedtuple('SkippedOptions', ['name', 'build_args', 'test_args', 'skip'])
```
galpeter: What if we do something like this?
```python
Options = collections.namedtuple('Options', ['name', 'build_args', 'test_args', 'skip'])
Options.__new__.__defaults__ = ([], [], False)

def skip_if(condition, desc):
    return desc if condition else False
```

Then we don't need the extra `SkippedOptions` and the new `Options` could be used like this:
```python
Options('buildoption_test-cpointer_32bit',
        ['--compile-flag=-m32', '--cpointer-32bit=on', '--system-allocator=on']),
        skip=skip_if(
            platform.system() != 'Linux' or (platform.machine() != 'i386' and platform.machine() != 'x86_64'),
            '-m32 is only supported on x86[-64]-linux')
        )
)
```
Of course then we need to checking the `skip` with `hasattr` is not enough and the skip description should be a valid non-empty string, but that can be checked (at least the length) in the `skip_if` method to ensure the developers put there meaningful text about by this option is skipped.

What do you think?

zherczeg:
Please fix the requested change or close the bug.

renatahodovan:
@galpeter thanks for the review. I've updated the PR accordingly. Except for the description check. I think it's the developer responsibility to give a decent description and throwing any kind of script error wouldn't be suitable here.

akosthekiss:
Still LGTM

PR #2559, opened 2018-10-09
Add some progress printing to the test262 tests

crazy2be:
Previously, these would run for several minutes without printing
any progress. Now, at least print the number of tests executed.

### DIFF
--- a/tools/runners/run-test-suite-test262.sh
+++ b/tools/runners/run-test-suite-test262.sh
@@ -62,9 +62,26 @@ rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T6.js"
 
 echo "Starting test262 testing for ${ENGINE}. Running test262 may take a several minutes."
 
+function progress_monitor() {
+  NUM_LINES_GOTTEN=0
+  (>&2 echo)
+  while read line
+  do
+    if [[ $((NUM_LINES_GOTTEN % 100)) == 0 ]]
+    then
+      (>&2 echo -ne "\rExecuted approx ${NUM_LINES_GOTTEN} tests...")
+    fi
+    echo "$line"
+    NUM_LINES_GOTTEN=$((NUM_LINES_GOTTEN + 1))
+  done
+  (>&2 echo)
+  (>&2 echo)
+}
+
 python2 "${PATH_TO_TEST262}"/tools/packaging/test262.py --command "${COMMAND}" \
                                                         --tests="${PATH_TO_TEST262}" --summary \
-                                                        &> "${REPORT_PATH}"
+                                                        | progress_monitor > "${REPORT_PATH}"
+
 TEST262_EXIT_CODE=$?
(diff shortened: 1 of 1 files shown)


LaszloLango:
I like this patch, but time to time we ran into the log limit on Travis CI. Fortunately it looks like this change does not reach the limit, because the bots are green. However it would be great to add a `quiet` option to easily turn off unnecessary logs when it is needed.

akosthekiss:
The commit is not properly signed, which makes the Travis CI fail. As long as the PR is red (i.e., some checks fail), it may get less attention from reviewers because they know that it will still certainly change. So, it's better to fix CI issues as early as possible.

crazy2be:
@akosthekiss: Oops, my bad! I didn't realize the CI builds were failing, thanks for the heads up :).

@LaszloLango: If the logs become too long, you can also adjust the space between logs. Right now it is 100, which should print about 115 lines. If that became too many, you can make it 200, which would only print 58 lines. Otherwise I am not sure the best way to add such a quiet option to the bash script (do you already have such a system in place for other scripts?).

tools/runners/run-test-suite-test262.sh
```diff
@@ -62,9 +62,24 @@ rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T6.js"
 
 echo "Starting test262 testing for ${ENGINE}. Running test262 may take a several minutes."
 
+function progress_monitor() {
+  NUM_LINES_GOTTEN=0
+  (>&2 echo)
+  while read line; do
```
LaszloLango: Minor style issue: put the `do` on separate line
```sh
while read line
do
```

tools/runners/run-test-suite-test262.sh
```diff
@@ -62,9 +62,24 @@ rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T6.js"
...
+    if [[ $((NUM_LINES_GOTTEN % 100)) == 0 ]]; then
```
LaszloLango: Minor style issue: put the `then` on a separate line
```
if [[ $((NUM_LINES_GOTTEN % 100)) == 0 ]]
then
```

tools/runners/run-test-suite-test262.sh
```diff
@@ -62,9 +62,24 @@ rm -f "${PATH_TO_TEST262}/test/suite/ch15/15.9/15.9.3/S15.9.3.1_A5_T6.js"
...
+      (>&2 echo -ne "\rExecuted approx ${NUM_LINES_GOTTEN} tests...")
+    fi
+    echo "$line"
+    NUM_LINES_GOTTEN=$((NUM_LINES_GOTTEN+1))
```
LaszloLango: Missing spaces around `+`:
`NUM_LINES_GOTTEN=$((NUM_LINES_GOTTEN + 1))`

PR #2560, opened 2018-10-12
Calculate string hashes only when needed

mmatyas:
It is not necessary to calculate the hash values for non-direct
strings on every operation. This patch aims to improve performance
by delaying the hash computations until the value is actually
required. This patch was done in cooperation with Tamás Kéri.


mmatyas:
This patch speeds up the string-related tests of Sunspider, however slows down for some reason on others. Here are our measurements:

measure1

measure2

zherczeg:
Probably caching, since if hash is not calculated, everything is bound to entry 0.

akosthekiss:
Although the goal of the change is performance improvement, half of the benchmarks degrade. I would not land this commit unless it can be enhanced to show a clear speedup.

zherczeg:
This patch is inactive, and I will close this bug until a perf gain is achieved.

PR #2563, opened 2018-10-18
Fix object initializer prescanning issues.

zherczeg:


zherczeg:
Thnak you. More description added.

PR #2570, opened 2018-10-24
Enable the already fixed test cases

matedabis:
Uncomment already fixed asserts.

### DIFF
--- a/tests/jerry/equality.js
+++ b/tests/jerry/equality.js
@@ -16,17 +16,17 @@ assert((5 == 5) == true);
 assert((7 != 2) == true);
 
 var num = 0;
-//var obj = new String("0");
+var obj = new String("0");
 var str = "0";
 var b = false;
 
 assert(num === num);
-//assert(obj === obj);
+assert(obj === obj);
 assert(str === str);
 
-//assert((num === obj) == false);
+assert((num === obj) == false);
 assert((num === str) == false);
-//assert((obj === str) == false);
-//assert((null === undefined) == false);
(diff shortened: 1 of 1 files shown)


akosthekiss:
I see the improvements to equality.js but what's the point of all the other "negative tests" in the other files? What test coverage does it improve?

matedabis:
@akosthekiss It improves the test coverage because it checks if these tests also work in the opposite way. There was not so many negative tests in the tests/jerry before, and with these test, I test cases that was not really tested before.

akosthekiss:
@matedabis I'd like to see this coverage increase supported by numbers. Could you please share coverage results (and how they were measured)?

My doubts come from the following observations:
- The tests already contain `if (true && false && fail())`
- The new addition is `if ((true && false && fail()) != false)`
- Assuming that `true && false && fail()` behaves consistently, this new code only tests `true != false`
- But that's already tested in test262/.../S8.3_A3.js.

This is just an example but there are similar constructs everywhere.

akosthekiss:
@matedabis Self correction. The new code only tests `false != false`, which is tested in .../S11.9.2_A3.1.js and S11.9.5_A3.js

matedabis:
@akosthekiss Okay, I will split it.

matedabis:
@akiss Now this PR only contains the changes of equality.js. I will put the other changes on @repasics's branch.

akosthekiss:
@matedabis Please, tag me (@akosthekiss) in communications.

PR #2572, opened 2018-10-26
Fix type conversion errors in case of TizenRT.

rtakacs:
There are some conversion errors when building JerryScript for TizenRT. This patch fixes that.


rtakacs:
@akosthekiss There are other `? true : false` places in the code and I use that solution for the hotfix. The explicit type case conversion indeed looks better.
The conversion errors are found by JSRemoteTest (when building for TizenRT). It is interesting why JerryScript's Travis jobs could not catch these errors.

zherczeg:
http://port70.net/~nsz/c/c11/n1570.html#6.3.1.2
This should be followed by C compilers. If a compiler complains, it does not follow the standard, so we should be careful.

akosthekiss:
The referred section of the standard only mandates _how_ a value must be converted to bool, not _when_. I.e., the standard does _not_ require the compiler to allow implicit conversion from int to bool.

PR #2574, opened 2018-10-26
Parser style fixes after Class part II patch.

zherczeg:


jerry-core/parser/js/js-parser-internal.h
```diff
@@ -39,41 +39,41 @@
...
-  PARSER_DEBUGGER_BREAKPOINT_APPENDED = (1u << 17), /**< pending (unsent) breakpoint
-                                                     *   info is available */
+  PARSER_IS_STRICT = (1u << 0),                      /**< strict mode code */
+  PARSER_IS_FUNCTION = (1u << 1),                    /**< function body is parsed */
+  PARSER_IS_CLOSURE = (1u << 2),                     /**< function body is encapsulated in {} block */
+  PARSER_IS_FUNC_EXPRESSION = (1u << 3),             /**< a function expression is parsed */
+  PARSER_IS_PROPERTY_GETTER = (1u << 4),             /**< a property getter function is parsed */
+  PARSER_IS_PROPERTY_SETTER = (1u << 5),             /**< a property setter function is parsed */
+  PARSER_NAMED_FUNCTION_EXP = (1u << 6),             /**< a function expression has a name binding */
+  PARSER_HAS_NON_STRICT_ARG = (1u << 7),                    /**< the function has arguments which
```
akosthekiss: This comment seems to be unaligned.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -1504,19 +1500,14 @@ parser_process_unary_expression (parser_context_t *context_p) /**< context */
 
         if (is_eval)
         {
+          parser_emit_cbc (context_p, CBC_EVAL);
```
akosthekiss: Isn't this causing extra work for the `#ifndef CONFIG_DISABLE_ES2015_CLASS` && `context_p->status_flags & PARSER_CLASS_HAS_SUPER` case? A `CBC_EVAL` is emitted just to be overwritten (below) with `context_p->last_cbc_opcode =`.

zherczeg: This is how CBC works in general. You start with an opcode and mutate it to others. While this could be done a bit more efficiently, I think the perf gain is negligible, since eval is not that frequent. And looks much better.

akosthekiss: Yes, bytecode rewrite happens, when a different step realizes that it can do something more efficiently than the previous step. But this is different as CBC rewrite happens right after the CBC is written. I don't think that we should write intentionally inefficient code only to avoid an `#ifndef`.

akosthekiss: I you mean by "manually" that it writes `last_cbc_opcode` and `last_cbc.value` directly, then that's still happening. To avoid that, the proper approach would be to add a new `parser_emit_cbc_class_eval` (or something like that) to js-parser-util.c and call that from the `if (context_p->status_flags & PARSER_CLASS_HAS_SUPER)` branch (and keep calling `parser_emit_cbc (context_p, CBC_EVAL);` otherwise).

LaszloLango:
@zherczeg Please rebase to the current master.

jerry-core/parser/js/js-parser-internal.h
zherczeg: It seems github does not close these automatically anymore.

jerry-core/parser/js/js-parser-expr.c
zherczeg: Code rewritten. I don't like these "Outdated" labels.

zherczeg:
Please check the patch again.

jerry-core/parser/js/js-lexer.c
```diff
@@ -319,7 +319,7 @@ static const keyword_string_t keywords_with_length_2[] =
 {
   LEXER_KEYWORD ("do", LEXER_KEYW_DO),
   LEXER_KEYWORD ("if", LEXER_KEYW_IF),
-  LEXER_KEYWORD ("in", LEXER_KEYW_IN),
+  LEXER_KEYWORD ("in", LEXER_KEYW_IN)
```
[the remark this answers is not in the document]
zherczeg: I think such change is rare because the items are in alphabetic order. And we usually don't put a comma after the last item in other enums as well.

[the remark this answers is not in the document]
akosthekiss: These are the arrays and enums where the use-trailing-comma-everywhere principle is used:

- arrays:
jerry_typedarray_mappings, ecma_builtin_routines, ecma_builtin_call_functions, ecma_builtin_construct_functions, ecma_builtin_property_list_references, ecma_error_mappings, cbc_flags, cbc_ext_flags, cbc_names, cbc_ext_names, statement_lengths
- enums:
jerry_debugger_flags_t, jerry_debugger_header_type_t, jerry_debugger_configuration_flags_t, jerry_debugger_eval_type_t, jerry_debugger_eval_result_type_t, jerry_debugger_output_subtype_t, ecma_init_flag_t, ecma_status_flag_t, ecma_parse_opts_t, ecma_property_types_t, ecma_list_properties_options_t, ecma_property_flags_t, ecma_collection_flag_t, ecma_direct_string_type_t, ecma_arraybuffer_extra_flag_t, ecma_property_hashmap_delete_status, ecma_builtin_property_type_t, ecma_builtin_number_type_t, ecma_array_object_set_length_flags_t, jerry_init_flag_t, jerry_type_t, jerry_typedarray_type_t, jerry_debugger_wait_for_source_status_t, jerry_generate_snapshot_opts_t, jerry_exec_snapshot_opts_t, jmem_free_unused_memory_severity_t, cbc_code_flags, cbc_opcode_t, cbc_ext_opcode_t, lexer_literal_type_t, lexer_literal_status_flags_t, skip_mode_t, lexer_token_type_t, lexer_obj_ident_opts_t, lexer_literal_object_type_t, lexer_number_type_t, parser_object_literal_item_types_t, parser_general_flags_t, parser_expression_flags_t, scan_modes_t, scan_stack_modes_t, parser_statement_type_t, parser_try_block_type_t, re_token_type_t, number_arithmetic_op, number_bitwise_logic_op, vm_stack_context_type_t, vm_oc_get_types, vm_oc_types, vm_oc_put_types, vm_call_operation

That's quite some, I think. But not universal, for sure. The style is mixed in the code base. But if I had to choose between future proof and not future proof, I'd go for the first option. And definitely would not change _back_ from the first style to the latter one.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -1375,6 +1371,24 @@ parser_parse_unary_expression (parser_context_t *context_p, /**< context */
   lexer_next_token (context_p);
 } /* parser_parse_unary_expression */
 
+/**
+ * Push an eval opcode.
+ */
+static inline void
+parser_emit_eval (parser_context_t *context_p) /**< context */
```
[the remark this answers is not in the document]
akosthekiss: Well, I was not referring to rules as such but to coding patterns, especially related to `parser_emit_` helper functions. Such functions appear in two files: js-parser-expr.c and js-parser-util.c. In js-parser-util.c, all `parser_emit_` functions are grouped to the beginning of the file, just like the single `parser_emit_unary_lvalue_opcode` helper in js-parser-expr.c (this file). Moreover `parser_emit_unary_lvalue_opcode` is first used at line 1693 but already defined at line 86 (i.e., not before the use).

LaszloLango:
@zherczeg please rebase the PR

zherczeg:
I may open it in the future.

PR #2575, opened 2018-10-26
Get rid of `? true : false`

akosthekiss:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
@@ -650,7 +650,7 @@ ecma_builtin_number_prototype_object_to_fixed (ecma_value_t this_arg, /**< this 
                                                                    num_digits + 1,
                                                                    exponent + frac_digits,
                                                                    &exponent,
-                                                                   ecma_number_is_zero (this_num) ? true : false);
+                                                                   ecma_number_is_zero (this_num));
 
           /* Buffer that is used to construct the string. */
           int buffer_size = (exponent > 0) ? exponent + frac_digits + 2 : frac_digits + 3;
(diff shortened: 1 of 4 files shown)


zherczeg:
I would like to see the effect on some exotic compiler environments. This has never caused any issues on Linux.

akosthekiss:
@zherczeg Please, name those exotic compiler environments and also those from whom you expect the feedback. Right now, I see no setup where `? true : false` would be a sensible construct.

LaszloLango:
LGTM if builds with in the environment @rtakacs use. @rtakacs please validate the patch.

rtakacs:
I've tried out JSRemoteTest with this branch:

```
js-remote-test/deps/jerryscript/jerry-core/parser/js/js-lexer.c: In function 'lexer_construct_regexp_object':
js-remote-test/deps/jerryscript/jerry-core/ecma/base/ecma-globals.h:246:3: error: conversion to '_Bool8' from 'long int' may alter its value [-Werror=conversion]
   (JERRY_UNLIKELY ((value) == ECMA_VALUE_ERROR))
   ^
js-remote-test/deps/jerryscript/jerry-core/parser/js/js-lexer.c:2178:19: note: in expansion of macro 'ECMA_IS_VALUE_ERROR'
   bool is_throw = ECMA_IS_VALUE_ERROR (completion_value);
```

rtakacs:
More detailed output: https://gist.github.com/rtakacs/0de9c0fa1b9a1b4b80162e435dd8f947

akosthekiss:
@rtakacs I've updated the PR. I did not expect `JERRY_UNLIKELY` (the expression behind `ECMA_IS_VALUE_ERROR`) to yield result incompatible with `bool`. Now, that result is also cast explicitly. Could you please run a new round of tests?

rtakacs:
@akosthekiss Okay, it works now :)

LGTM (informal)

akosthekiss:
Thanks, @rtakacs

PR #2577, opened 2018-10-29
Improve .travis.yml

akosthekiss:
- Use builtin support of Travis CI for job names instead of using
  a custom environment variable for the same purpose.
- Use `skip` in `script` and `install` steps instead of `true`.


akosthekiss:
I think this makes the Travis CI page nicer but that's highly subjective. Please, let me know what you think.

PR #2578, opened 2018-10-29
Add new test

repasics:
Added unusual-javascript-constructs.js to test corner cases

### DIFF
--- a/tests/jerry-test-suite/11/11.08/11.08.06/11.08.06-008.js
+++ b/tests/jerry-test-suite/11/11.08/11.08.06/11.08.06-008.js
@@ -12,19 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-var big = 2147483646;
-
-big++;
-assert(big == 2147483647);
-
-big += 1;
-assert(big == 2147483648); // overflow on 32bit numbers
-
-big++;
-assert(big == 2147483649); // overflow on 32bit numbers
-
-assert ((1152921504606846976).toString() === "1152921504606847000")
-
-assert (1.797693134862315808e+308 === Infinity);
-
-assert (9999999999999999 == 10000000000000000);
+var a = null;
+var c = a instanceof Object;
+assert (!c);
--- a/tests/jerry/arithmetics-2.js
+++ b/tests/jerry/arithmetics-2.js
@@ -51,3 +51,5 @@ try {
 catch (e) {
   assert (e instanceof ReferenceError);
 }
+
+assert (0.1 + 0.2 != 0.3);
--- a/tests/jerry/unusual.js
+++ b/tests/jerry/unusual.js
@@ -12,19 +12,16 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-var big = 2147483646;
+var x = [0];
+assert (x == x);
+assert (x == !x);
+assert (Array(3) == ",,");
 
-big++;
-assert(big == 2147483647);
+assert ([] + [] == "");
+assert ([] + {} == "[object Object]");
+assert (eval ("{} + []") == 0);
+assert (isNaN (eval ("{} + {}")));
+assert ({} + [] == "[object Object]");
+assert ({} + {} == "[object Object][object Object]");
 
-big += 1;
-assert(big == 2147483648); // overflow on 32bit numbers
-
-big++;
-assert(big == 2147483649); // overflow on 32bit numbers
-
-assert ((1152921504606846976).toString() === "1152921504606847000")
-
-assert (1.797693134862315808e+308 === Infinity);
-
-assert (9999999999999999 == 10000000000000000);
+assert ((! + [] + [] + ![]) === "truefalse");


tests/jerry/unusual-javascript-constructs.js
```diff
@@ -0,0 +1,39 @@
...
+
+assert (9999999999999999 == 10000000000000000);
+assert (0.1 + 0.2 != 0.3);
+assert (!(Math.max() > Math.min()));
+assert (Math.max() < Math.min());
+
+assert ([] + [] == "");
+assert ([] + {} == "[object Object]");
+//assert({} + [] == 0); // FIXME: assertion failed, it should be 0, but it is [object Object]
+//assert({} + {} == NaN); // // FIXME: assertion failed, it should be NaN, but it is [object Object][object Object]
```
loki04: Please create new issues for these two ones! It probably worth to add the link of the issues here as well.

rerobika: There is **no need** to add new issues and also the _assertions are wrong_.

The correct asserts would look like:

```js
assert (eval ("{} + []") == 0)
assert (isNaN (eval ("{} + {}"))); // isNaN should be used instead of == NaN
```
The reason behind this is how `{}` is being parsed.
- As a block (That's why `{} + []` evaluates to 0, since `{}` evaluates to an **empty block**, and `+[]` gives **0** as a result)
- As an object literal (putting this check in an assert the `{}` becomes a part of an **expression** so it's interpreted as an **object**)

Also these test cases should be added with a bit modification:

```js
assert ({} + [] == "[object Object]")
assert ({} + {} == "[object Object][object Object]");
```

akosthekiss: This is strange. I've just built jerry and the interactive shell gives me:
```
jerry> {} + {}
NaN
```

That's not `[object Object][object Object]`.

Moreover, the assert couldn't be passing anyway as comparing anything to NaN (NaN included) gives false.

tests/jerry/unusual-javascript-constructs.js
```diff
@@ -0,0 +1,39 @@
...
+assert (!("string" instanceof String));
+
+assert (9999999999999999 == 10000000000000000);
+assert (0.1 + 0.2 != 0.3);
+assert (!(Math.max() > Math.min()));
+assert (Math.max() < Math.min());
+
+assert ([] + [] == "");
+assert ([] + {} == "[object Object]");
+//assert({} + [] == 0); // FIXME: assertion failed, it should be 0, but it is [object Object]
```
akosthekiss: This also seems to work for me in the interactive shell:
```
jerry> {} + []
0
```

tests/jerry/unusual-javascript-constructs.js
akosthekiss: (To clarify: my above comment "This is strange" is not a response to @rerobika but a reaction to the commit itself. We must have been typing our reviews in parallel.)

rerobika: @akosthekiss That's the expected behavior. Since you are using the CLI, it evaluates the given line, so the output is correct. (same for the other test case)

``` js
jerry> {} + [] // parse {} as an empty block
0
jerry> print ( {} + [] ) // parse {} as an object literal ( now {} is a part of an expression)
[object Object]
undefined
```

tests/jerry/unusual-javascript-constructs.js
```diff
@@ -0,0 +1,39 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var x = [0];
+assert (x == x);
+assert (x == !x);
+assert (Array(3) == ",,");
+
+assert (typeof NaN == "number");
```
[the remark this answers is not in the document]
loki04: I agree with you. If the test case exists elsewhere in the repository, there is no reason to duplicate it.

tests/jerry/unusual-javascript-constructs.js
```diff
@@ -0,0 +1,32 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var x = [0];
+assert (x == x);
+assert (x == !x);
+assert (Array(3) == ",,");
+
+assert (!(null instanceof Object));
```
akosthekiss: Something like this is tested in tests/test262/test/suite/ch15/15.3/15.3.5/S15.3.5.3_A1_T6.js (testing [[HasInstance]]) but I agree that it's not completely the same. But then, this might deserve its own test case in tests/jerry-test-suite/11/11.08/11.08.06/ (tests of instanceof), a new 11.08.06-008.js test file.

tests/jerry/unusual-javascript-constructs.js
```diff
@@ -0,0 +1,32 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
```
akosthekiss: "javascript" in the file name doesn't carry too much information (considering that this is the test suite of a javascript engine). I think tests/jerry/unusual.js would be enough.

tests/jerry-test-suite/11/11.08/11.08.06/11.08.06-008.js
```diff
@@ -0,0 +1,17 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var a = null;
+var c = a instanceof Object;
+assert (!c);
```
rerobika: Isn't this a bit convoluted way of saying
```js
assert (!(null instanceof Object)
```
?

repasics: I used the other files from that directory as sample.

rerobika: Thanks for the clarification, now it's fine for me as well.

tests/jerry/arithmetics-2.js
```diff
@@ -28,6 +28,8 @@ assert(c == 210);
 c = a / b;
 assert(c >= 2.1 - 0.000001 && c <= 2.1 + 0.000001);
 
+assert (0.1 + 0.2 != 0.3);
```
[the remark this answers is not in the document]
repasics: I've inserted it there because there is the only floating point check in the file. I tought it fits there.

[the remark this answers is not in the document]
akosthekiss: I'd suggest moving it to the end.

PR #2583, opened 2018-10-31
Improve the evaluate requests

robertsipka:
Currently it evaluates the given expression in the context of the top most stack frame.
The expression should access to any variables and arguments that are in the scope chain.
Implement the eval_at request with the level of the scope chain as a further argument.

### DIFF
--- a/jerry-core/debugger/debugger.c
+++ b/jerry-core/debugger/debugger.c
@@ -40,7 +40,7 @@ typedef struct
  */
 JERRY_STATIC_ASSERT (JERRY_DEBUGGER_MESSAGES_OUT_MAX_COUNT == 32
                      && JERRY_DEBUGGER_MESSAGES_IN_MAX_COUNT == 21
-                     && JERRY_DEBUGGER_VERSION == 7,
+                     && JERRY_DEBUGGER_VERSION == 8,
                      debugger_version_correlates_to_message_type_count);
 
 /**
@@ -536,12 +536,17 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
   JERRY_ASSERT (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_VM_IGNORE));
 
   JERRY_DEBUGGER_SET_FLAGS (JERRY_DEBUGGER_VM_IGNORE);
-  ecma_value_t result = ecma_op_eval_chars_buffer (eval_string_p + 1, eval_string_size - 1, ECMA_PARSE_DIRECT_EVAL);
+
+  uint32_t chain_index;
+  memcpy (&chain_index, eval_string_p, sizeof (uint32_t));
+  uint32_t parse_opts = ECMA_PARSE_DIRECT_EVAL | (chain_index << ECMA_PARSE_CHAIN_INDEX_SHIFT);
+
+  ecma_value_t result = ecma_op_eval_chars_buffer (eval_string_p + 5, eval_string_size - 5, parse_opts);
   JERRY_DEBUGGER_CLEAR_FLAGS (JERRY_DEBUGGER_VM_IGNORE);
 
   if (!ECMA_IS_VALUE_ERROR (result))
   {
-    if (eval_string_p[0] != JERRY_DEBUGGER_EVAL_EVAL)
+    if (eval_string_p[4] != JERRY_DEBUGGER_EVAL_EVAL)
     {
       JERRY_DEBUGGER_SET_FLAGS (JERRY_DEBUGGER_VM_EXCEPTION_THROWN);
       JERRY_CONTEXT (error_value) = result;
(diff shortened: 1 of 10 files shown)


jerry-core/debugger/debugger.c
```diff
@@ -955,15 +968,23 @@ jerry_debugger_process_message (const uint8_t *recv_buffer_p, /**< pointer to th
       }
 
       jerry_debugger_uint8_data_t *eval_uint8_data_p;
-      size_t eval_data_size = sizeof (jerry_debugger_uint8_data_t) + eval_size;
+
+      size_t eval_data_size = sizeof (jerry_debugger_uint8_data_t) + eval_size + sizeof (uint32_t);
```
zherczeg: Why don't you simply copy all data? There is no need to copy chain_index separately, since it is also part of the message payload. This would simplify the code here a lot.

jerry-core/vm/vm.c
```diff
@@ -252,6 +253,23 @@ vm_run_eval (ecma_compiled_code_t *bytecode_data_p, /**< byte-code data */
...
     this_binding = ecma_copy_value (JERRY_CONTEXT (vm_top_context_p)->this_binding);
     lex_env_p = JERRY_CONTEXT (vm_top_context_p)->lex_env_p;
+
+    while (chain_index != 0)
+    {
+      lex_env_p = ecma_get_lex_env_outer_reference (lex_env_p);
+
+      if (JERRY_UNLIKELY (lex_env_p == NULL))
+      {
+        lex_env_p = JERRY_CONTEXT (vm_top_context_p)->lex_env_p;
```
zherczeg: Perhaps we could use the global lex_env instead, which is always the last. Something like:
```
if (JERRY_UNLIKELY (lex_env_p == jerry_get_global_lex_env())) {
  break;
}
```

The global lex env could be stored in a local variable to speed things up. We should enter to this loop if `chain_index > 0` which should be guarded with JERRY_UNLIKELY.

robertsipka:
@zherczeg : Thanks for the review, I've updated the PR.

tests/debugger/do_eval_at.cmd
```diff
@@ -0,0 +1,16 @@
...
+scopes
+eval_at 0 b
+eval_at 1 b
+eval_at 0 b=20
+eval_at 1 b=100
+n
+eval_at 0 a
+scopes
+eval_at 0 b
+c
```
LaszloLango: Please add negative test cases, like negative index, index over 65535, index over the max scope, etc.

robertsipka: I've added some test cases.

tests/debugger/do_eval_at.expected
```diff
@@ -0,0 +1,46 @@
...
+100
+(jerry-debugger) eval_at -1 b
+Error: Invalid scope chain index: -1 (must be between 0 and 65535)
+(jerry-debugger) eval_at 65536 b
+Error: Invalid scope chain index: 65536 (must be between 0 and 65535)
+(jerry-debugger) eval_at 65535 b
+100
+(jerry-debugger) eval_at b
+Error: invalid literal for int() with base 10: 'b'
+(jerry-debugger) eval_at 200
```
LaszloLango: IMHO it would be better to throw an error with a descriptive message in such cases instead of returning with `undefined`

robertsipka: @LaszloLango : `eval` and `eval_at` will return with `undefined` if the given expression is empty. It is an expected behavior. But I also understand what you requested, I modified the patch to return with an error if the chain index is too high.
If the error messages is enabled, it will show the detailed message.

zherczeg: Should be RangeError

robertsipka: I've modified to raise RangeError.

jerry-core/debugger/debugger.c
```diff
@@ -536,12 +536,17 @@ jerry_debugger_send_eval (const lit_utf8_byte_t *eval_string_p, /**< evaluated s
   JERRY_ASSERT (!(JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_VM_IGNORE));
 
   JERRY_DEBUGGER_SET_FLAGS (JERRY_DEBUGGER_VM_IGNORE);
-  ecma_value_t result = ecma_op_eval_chars_buffer (eval_string_p + 1, eval_string_size - 1, ECMA_PARSE_DIRECT_EVAL);
+
+  uint32_t chain_index;
+  memcpy (&chain_index, eval_string_p, sizeof (uint16_t));
```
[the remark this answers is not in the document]
robertsipka: Indeed, thanks. I have considered your request and modified the PR accordingly.

jerry-core/vm/vm.c
```diff
@@ -252,6 +252,26 @@ vm_run_eval (ecma_compiled_code_t *bytecode_data_p, /**< byte-code data */
...
+#ifdef JERRY_DEBUGGER
+    uint32_t chain_index = parse_opts >> ECMA_PARSE_CHAIN_INDEX_SHIFT;
+
+    while (chain_index != 0)
+    {
+      lex_env_p = ecma_get_lex_env_outer_reference (lex_env_p);
+
+      if (JERRY_UNLIKELY (lex_env_p == NULL))
+      {
+        return ecma_raise_range_error (ECMA_ERR_MSG ("The given chain index is too high"));
```
zherczeg: `ECMA_ERR_MSG ("Invalid scope chain index for eval")`

robertsipka: Okay, I've just changed it.

PR #2588, opened 2018-11-08
Temporarily disable class Array/%TypedArray%.prototype.{slice, map, concat, filter, splice} related tests

### LINKED ISSUE #2587: Missing prototype functions in ES2015
There is regression after the Class implementation landed.

**Build command**:
```bash
tools/build.py --clean --debug --profile=es2015-subset
```

**Test case**:
```js
function f(a,b,c) {
  var args = Array.prototype.slice.call(arguments, 3);
  print(typeof args.splice);
}
f();
```

**Output**:
```
undefined
```

**Expected output**:
```
function
```

rerobika:
This patch temporarily fixes #2587.

The reason of disabling these tests is that the current implementation slightly differs from the related part of the standard (ECMA-262 v6, 9.4.2.3.6.d.1).
This part requires the hidden Symbol.@@species property, hence this functionality has not been implemented yet in the project.
Also add the related test case to prevent further errors.

### DIFF
--- a/jerry-core/ecma/operations/ecma-array-object.c
+++ b/jerry-core/ecma/operations/ecma-array-object.c
@@ -154,41 +154,12 @@ ecma_op_create_array_object_by_constructor (const ecma_value_t *arguments_list_p
                                             ecma_object_t *object_p) /**< The object from whom the new array object
                                                                       *   is being created */
 {
-  ecma_value_t constructor_value = ecma_op_object_get_by_magic_id (object_p, LIT_MAGIC_STRING_CONSTRUCTOR);
+  /* TODO: Use @@species after Symbol has been implemented */
+  JERRY_UNUSED (object_p);
 
-  if (ECMA_IS_VALUE_ERROR (constructor_value)
-      || !ecma_is_value_object (constructor_value)
-      || !ecma_is_constructor (constructor_value))
-  {
-    ecma_free_value (constructor_value);
-    return ecma_raise_type_error (ECMA_ERR_MSG ("object.constructor is not a constructor."));
-  }
-
-  ecma_object_t *constructor_object_p = ecma_get_object_from_value (constructor_value);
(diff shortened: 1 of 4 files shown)


rerobika:
cc: @LaszloLango @zherczeg

jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -165,6 +166,8 @@ ecma_op_create_array_object_by_constructor (const ecma_value_t *arguments_list_p
   }
 
   ecma_object_t *constructor_object_p = ecma_get_object_from_value (constructor_value);
+
+  //FIXME: Use @@species after Symbol has been implemented
```
zherczeg: I think we only use `/* */` comments.

rerobika: True, fixed.

jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -165,6 +166,8 @@ ecma_op_create_array_object_by_constructor (const ecma_value_t *arguments_list_p
   }
 
   ecma_object_t *constructor_object_p = ecma_get_object_from_value (constructor_value);
+
+  /* FIXME: Use @@species after Symbol has been implemented */
```
LaszloLango: Please  write `TODO` instead of `FIXME`. We are using only `TODO` comments since #990

jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -154,6 +154,7 @@ ecma_op_create_array_object_by_constructor (const ecma_value_t *arguments_list_p
                                             ecma_object_t *object_p) /**< The object from whom the new array object
                                                                       *   is being created */
 {
+#ifdef CONFIG_DISABLE_ES2015_SYMBOL
```
LaszloLango: `#ifndef CONFIG_DISABLE_ES2015_SYMBOL`

PR #2589, opened 2018-11-12
Improve the vm_loop suspending for exec operations

rerobika:
Exec operations{call, construct, super_call} related bytecode sequences no longer executed twice.
The execution continues with the next opcode or a specific bytecode sequence if an error occurs during the operation.

### DIFF
--- a/jerry-core/jcontext/jcontext.h
+++ b/jerry-core/jcontext/jcontext.h
@@ -139,6 +139,8 @@ struct jerry_context_t
   jerry_debugger_transport_header_t *debugger_transport_header_p; /**< head of transport protocol chain */
   uint8_t *debugger_send_buffer_payload_p; /**< start where the outgoing message can be written */
   vm_frame_ctx_t *debugger_stop_context; /**< stop only if the current context is equal to this context */
+  uint8_t *debugger_exception_byte_code_p; /**< Location of the currently executed byte code if an
+                                            *   error occours while the vm_loop is suspended */
   jmem_cpointer_t debugger_byte_code_free_head; /**< head of byte code free linked list */
   jmem_cpointer_t debugger_byte_code_free_tail; /**< tail of byte code free linked list */
   uint32_t debugger_flags; /**< debugger flags */
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -401,6 +401,15 @@ vm_get_implicit_this_value (ecma_value_t *this_value_p) /**< [in,out] this value
   return false;
 } /* vm_get_implicit_this_value */
 
+/**
+ * Special bytecode sequence for error handling while the vm_loop
+ * is preserved for an execute operation
+ */
+static const uint8_t vm_error_byte_code_p[] =
+{
+  CBC_EXT_OPCODE, CBC_EXT_ERROR
+};
+
 #ifndef CONFIG_DISABLE_ES2015_CLASS
 /**
  * 'super(...)' function call handler.
@@ -411,7 +420,9 @@ vm_super_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
   JERRY_ASSERT (frame_ctx_p->call_operation == VM_EXEC_SUPER_CALL);
   JERRY_ASSERT (frame_ctx_p->byte_code_p[0] == CBC_EXT_OPCODE);
 
-  uint32_t arguments_list_len = frame_ctx_p->byte_code_p[2];
+  uint8_t *byte_code_p = frame_ctx_p->byte_code_p + 3;
+  uint8_t opcode = byte_code_p[-2];
+  uint32_t arguments_list_len = byte_code_p[-1];
 
   ecma_value_t *stack_top_p = frame_ctx_p->stack_top_p - arguments_list_len;
 
@@ -444,13 +455,35 @@ vm_super_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
     ecma_fast_free_value (stack_top_p[i]);
   }
 
-  ecma_free_value (stack_top_p[-1]);
-  stack_top_p[-1] = completion_value;
+  if (JERRY_UNLIKELY (ECMA_IS_VALUE_ERROR (completion_value)))
+  {
+#ifdef JERRY_DEBUGGER
+    JERRY_CONTEXT (debugger_exception_byte_code_p) = frame_ctx_p->byte_code_p;
+#endif /* JERRY_DEBUGGER */
+    frame_ctx_p->byte_code_p = (uint8_t *) vm_error_byte_code_p;
+  }
+  else
+  {
+    frame_ctx_p->byte_code_p = byte_code_p;
+    ecma_free_value (*(--stack_top_p));
+    uint32_t opcode_data = vm_decode_table[(CBC_END + 1) + opcode];
+
+    if (!(opcode_data & (VM_OC_PUT_STACK | VM_OC_PUT_BLOCK)))
+    {
+      ecma_fast_free_value (completion_value);
+    }
+    else if (opcode_data & VM_OC_PUT_STACK)
+    {
+      *stack_top_p++ = completion_value;
+    }
+    else
+    {
+      ecma_fast_free_value (frame_ctx_p->block_result);
+      frame_ctx_p->block_result = completion_value;
+    }
+  }
 
   frame_ctx_p->stack_top_p = stack_top_p;
-
-  /* 'super (...)' call continues as a normal 'Function call' */
-  frame_ctx_p->call_operation = VM_EXEC_CALL;
 } /* vm_super_call */
 #endif /* !CONFIG_DISABLE_ES2015_CLASS */
 
@@ -462,7 +495,8 @@ vm_super_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
 static void
 opfunc_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
 {
-  uint8_t opcode = frame_ctx_p->byte_code_p[0];
+  uint8_t *byte_code_p = frame_ctx_p->byte_code_p + 1;
+  uint8_t opcode = byte_code_p[-1];
   uint32_t arguments_list_len;
 
   if (opcode >= CBC_CALL0)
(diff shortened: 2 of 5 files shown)


rerobika:
Performance results:

Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 0.796 -> 0.792 : +0.573% | 
3d-raytrace.js | 1.031 -> 1.025 : +0.581% | 
access-binary-trees.js | 0.565 -> 0.548 : +3.131% | 
access-fannkuch.js | 2.066 -> 2.030 : +1.762% | 
access-nbody.js | 1.081 -> 1.080 : +0.078% | 
bitops-3bit-bits-in-byte.js | 0.492 -> 0.480 : +2.446% | 
bitops-bits-in-byte.js | 0.641 -> 0.642 : -0.121% | 
bitops-bitwise-and.js | 0.917 -> 0.916 : +0.093% | 
bitops-nsieve-bits.js | 1.110 -> 1.116 : -0.532% | 
controlflow-recursive.js | 0.373 -> 0.347 : **+7.104%** | 
crypto-aes.js | 0.874 -> 0.861 : +1.489% | 
crypto-md5.js | 0.592 -> 0.599 : -1.283% | 
crypto-sha1.js | 0.586 -> 0.588 : -0.256% | 
date-format-tofte.js | 0.746 -> 0.736 : +1.331% | 
date-format-xparb.js | 0.529 -> 0.521 : +1.358% | 
math-cordic.js | 1.159 -> 1.153 : +0.510% | 
math-partial-sums.js | 0.726 -> 0.714 : +1.598% | 
math-spectral-norm.js | 0.535 -> 0.530 : +0.941% | 
string-base64.js | 1.400 -> 1.428 : -2.035% | 
string-fasta.js | 1.341 -> 1.340 : +0.066% | 
Geometric mean: |**+0.959%** | 

Binary sizes (bytes)
74d474248d:132664
8ca7ca8f3a:132664

jerry-core/vm/vm.c
```diff
@@ -382,6 +382,22 @@ vm_get_implicit_this_value (ecma_value_t *this_value_p) /**< [in,out] this value
...
+static const uint8_t vm_error_byte_code_p[] =
+{
+  CBC_EXT_OPCODE, CBC_EXT_ERROR
+};
+
+/**
+ * Location of the currently executed byte code if an
+ * error occours while the vm_loop is preserved
+ */
+static uint8_t *vm_error_byte_code_location_p = NULL;
```
zherczeg: Why do we exactly need this? Could you show me an example?

rerobika: [This commit ](https://github.com/rerobika/jerryscript/compare/75b778c905e5...c8bdc4da53d9)is the same as the PR's content but without the usage of this static variable.

It fails on the `debugger/do_exception.js`. See the details [here](https://travis-ci.org/rerobika/jerryscript/jobs/453856622#L564).

zherczeg: This is good news! So we only need this variable when the debugger is enabled to determine the precise location of the exception. This could be moved inside the debugger part of the JCONTEXT.

rerobika:
@zherczeg I've applied your suggestion!

jerry-core/vm/vm.c
zherczeg: One more thing: when the debugger stops because of an exception, the byte code pointer should point the offending byte code. If the debugger sees the byte code after the call, it might be part of a different JavaScript expression.

jerry-core/jcontext/jcontext.h
```diff
@@ -146,6 +146,8 @@ struct jerry_context_t
   uint8_t debugger_message_delay; /**< call receive message when reaches zero */
   uint8_t debugger_max_send_size; /**< maximum amount of data that can be sent */
   uint8_t debugger_max_receive_size; /**< maximum amount of data that can be received */
+  uint8_t *debugger_exception_byte_code_p; /**< Location of the currently executed byte code if an
```
zherczeg: I would put this after `debugger_stop_context` to reduce alignment holes.

jerry-core/vm/vm.c
```diff
@@ -508,9 +543,35 @@ opfunc_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+      *stack_top_p++ = completion_value;
+    }
+    else
+    {
+      ecma_fast_free_value (frame_ctx_p->block_result);
+      frame_ctx_p->block_result = completion_value;
+    }
+  }
 
+  frame_ctx_p->call_operation = VM_NO_EXEC_OP;
```
zherczeg: Is this needed?

jerry-core/vm/vm.c
```diff
@@ -3416,28 +3416,35 @@ vm_execute (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
...
-    }
-
-    if (frame_ctx_p->call_operation == VM_EXEC_CALL)
+    switch (frame_ctx_p->call_operation)
     {
-      opfunc_call (frame_ctx_p);
-    }
+      case VM_NO_EXEC_OP:
+      {
+        goto exec_end;
```
zherczeg: What about moving the code after exec_end here? We would not need any goto then.

rerobika:
@zherczeg The PR is updated. See the fresh performance results:

Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 0.807 -> 0.788 : +2.369% | 
3d-raytrace.js | 1.037 -> 1.018 : +1.828% | 
access-binary-trees.js | 0.560 -> 0.541 : +3.305% | 
access-fannkuch.js | 2.095 -> 2.038 : +2.740% | 
access-nbody.js | 1.087 -> 1.068 : +1.817% | 
bitops-3bit-bits-in-byte.js | 0.506 -> 0.482 : +4.859% | 
bitops-bits-in-byte.js | 0.665 -> 0.646 : +2.894% | 
bitops-bitwise-and.js | 0.926 -> 0.918 : +0.888% | 
bitops-nsieve-bits.js | 1.132 -> 1.106 : +2.328% | 
controlflow-recursive.js | 0.366 -> 0.362 : +0.971% | 
crypto-aes.js | 0.878 -> 0.858 : +2.302% | 
crypto-md5.js | 0.600 -> 0.582 : +2.931% | 
crypto-sha1.js | 0.595 -> 0.573 : +3.613% | 
date-format-tofte.js | 0.748 -> 0.734 : +1.888% | 
date-format-xparb.js | 0.526 -> 0.518 : +1.534% | 
math-cordic.js | 1.174 -> 1.142 : +2.752% | 
math-partial-sums.js | 0.737 -> 0.715 : +3.047% | 
math-spectral-norm.js | 0.541 -> 0.536 : +1.024% | 
string-base64.js | 1.407 -> 1.394 : +0.916% | 
string-fasta.js | 1.341 -> 1.328 : +1.002% | 
Geometric mean: | **+2.256%** | 

Binary sizes (bytes)
22b08518c7:132664
1d5b907c66:132664

jerry-core/vm/vm.c
```diff
@@ -473,6 +506,7 @@ opfunc_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
   else
   {
     arguments_list_len = frame_ctx_p->byte_code_p[1];
+    byte_code_length++;
```
zherczeg: byte_code_length = 2.

This somehow could be done by directly increasing the byte code, since it is not used afterwards. From the opcode you can get the length, which is only needed when debugger is enabled.

rerobika: Unfortunately I'm not sure that we can do it directly.

As I've applied your suggestion it caused an error in `debugger_do_step.js` while sending the backtrace info.

```diff
--- tests/debugger/do_step.expected     2018-10-26 15:50:52.387344484 +0200
+++ tests/debugger/do_step.out.ufKJQwcUB4       2018-11-21 11:21:23.083086725 +0100
@@ -10 +10 @@
-Frame 2: tests/debugger/do_step.js:32
+Frame 2: tests/debugger/do_step.js:33
@@ -15 +15 @@
-Frame 1: tests/debugger/do_step.js:32
+Frame 1: tests/debugger/do_step.js:33
```

Reason:
 - It tries to send to actual bytecode position [here](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/debugger/debugger.c#L182), but this value is increased by the actual bytecode length.

So as I see we can't increase the actual byte_code pointer since the ecma_op_function_call (...) has not finished the real call operation.

zherczeg: It looks like this is again something debugger specific. 

What about changing the code to:
`arguments_list_len = *(++frame_ctx_p->byte_code_p);`

This way we are not fully reading the byte code, so not moving to the next instruction.

Or replace `byte_code_length` to save `byte_code_p` locally:
`uint8_t *byte_code_p = frame_ctx_p->byte_code_p;`

And update it accordingly. 

Setting a value is faster than += operation, since that needs reading the original value.

rerobika: +1 for your second suggestion, it looks promising.

jerry-core/vm/vm.c
```diff
@@ -411,7 +420,9 @@ vm_super_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
   JERRY_ASSERT (frame_ctx_p->call_operation == VM_EXEC_SUPER_CALL);
   JERRY_ASSERT (frame_ctx_p->byte_code_p[0] == CBC_EXT_OPCODE);
 
-  uint32_t arguments_list_len = frame_ctx_p->byte_code_p[2];
+  uint8_t *byte_code_p = frame_ctx_p->byte_code_p + 1;
+  uint8_t opcode = *byte_code_p++;
```
zherczeg:
```
  uint8_t *byte_code_p = frame_ctx_p->byte_code_p + 3;
  uint8_t opcode = byte_code_p[-2];
  uint32_t arguments_list_len = byte_code_p[-1];
```

Negative fixed address is cheap.

jerry-core/vm/vm.c
```diff
@@ -462,7 +495,8 @@ vm_super_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
 static void
 opfunc_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
 {
-  uint8_t opcode = frame_ctx_p->byte_code_p[0];
+  uint8_t *byte_code_p = frame_ctx_p->byte_code_p;
+  uint8_t opcode = *byte_code_p++;
```
zherczeg: Ditto here.

PR #2591, opened 2018-11-12
Calculate string lengths only when needed

mmatyas:
It is not necessary to calculate the length values for strings on
every operation. This patch aims to improve performance by delaying
the length computations until the value is actually required.
This patch was done in cooperation with Tamás Kéri.

### DIFF
--- a/jerry-core/api/jerry-snapshot.c
+++ b/jerry-core/api/jerry-snapshot.c
@@ -1550,7 +1550,7 @@ jerry_append_number_to_buffer (uint8_t *buffer_p, /**< buffer */
  *         false - otherwise
  */
 static bool
-ecma_string_is_valid_identifier (const ecma_string_t *string_p)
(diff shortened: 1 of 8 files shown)


mmatyas:
Measurement results:

image

LaszloLango:
I do not like the removal of `const` modifiers from getter functions. It can lead to unexpected behavior later. IMHO the performance gain is too small, so it is not worth it.

rerobika:
@mmatyas Could you update the patch to the latest master and remeasure the performance gain?

mmatyas:
Sure, here you go:

image

akosthekiss:
Are there any specialized benchmarks that show gain with this change? The attached table only shows statistical error / noise, i.e., no performance gain.

zherczeg:
Is there still work no this?

LaszloLango:
Closing due to long time inactivity.

PR #2597, opened 2018-11-14
Support identifier references for object initializers.

zherczeg:


zherczeg:
Hi @AnthonyCalandra, I found a simple solution for supporting identifier references.

PR #2601, opened 2018-11-16
Implement the Symbol builtin object part I.

rerobika:
This patch contains the base functionalities that the new builtin object requires.

Currently unavailable:
 - print (Symbol('')) - this features requires the refactor of the print handler function
 - Special builtin object accesors for override the builtin symbols (@@species, @@toString, etc.)


jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -383,6 +400,7 @@ ecma_check_value_type_is_spec_defined (ecma_value_t value) /**< ecma value */
                 || ecma_is_value_boolean (value)
                 || ecma_is_value_number (value)
                 || ecma_is_value_string (value)
+                || ECMA_ASSERT_VALUE_IS_SYMBOl (value)
```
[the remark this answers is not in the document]
rerobika: Due to the macro guards IMO we cant.

Since `ecma_is_value_symbol` is only available the the the Symbol builtin is enabled, the proper guard for the assert would be:

```c
JERRY_ASSERT (ecma_is_value_undefined (value)
                || ecma_is_value_null (value)
                || ecma_is_value_boolean (value)
                || ecma_is_value_number (value)
                || ecma_is_value_string (value)
#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
                || ecma_is_value_symbol(value)
#endif /* !CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN */
                || ecma_is_value_object (value));
```

but is result the following error:

```
home/fancsikrobi/harmony/jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:403:1: error: embedding a directive within macro arguments is not portable [-Werror]
 #ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
 ^
/home/fancsikrobi/harmony/jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:405:1: error: embedding a directive within macro arguments is not portable [-Werror]
 #endif /* !CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN */
 ^
```

Since JERRY_ASSSERT is also a macro.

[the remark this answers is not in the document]
legendecas: Does it make sense that make `ecma_is_value_symbol` available whether `CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN` defined or nor?
If `CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN` is set `ecma_is_value_symbol` could return false directly.

tests/jerry/es2015/object-get_own_property_symbols.js
```diff
@@ -0,0 +1,84 @@
...
+obj[b] = 'b';
+obj[c] = 'c';
+obj[d] = 'd';
+props = Object.getOwnPropertySymbols(obj);
+// props should not contain: a, b, c, d and the order is not defined!
+assert (props.indexOf(a) !== -1);
+assert (props.indexOf(b) !== -1);
+assert (props.indexOf(c) !== -1);
+assert (props.indexOf(d) !== -1);
+assert (props.length === 4);
```
legendecas: If assertion shall be made on properties keyed with same description symbols? Since symbols with same descriptions shall not be treated as same key.

rerobika: Well, that's a good idea! I'll add a new test case for it.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -137,6 +137,18 @@ typedef enum
  */
 #define ECMA_BOOL_TO_BITFIELD(x) ((x) ? 1 : 0)
 
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+/**
+ * JERRY_ASSERT compatible macro for checking whether the given ecma-value is symbol
+ */
+#define ECMA_ASSERT_VALUE_IS_SYMBOl(value) (ecma_is_value_symbol ((value)))
```
legendecas: Typo: `ECMA_ASSERT_VALUE_IS_SYMBOl` -> `ECMA_ASSERT_VALUE_IS_SYMBOL`

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -198,6 +203,26 @@ ecma_new_ecma_string_from_magic_string_ex_id (lit_magic_string_ex_id_t id) /**<
...
+ */
+ecma_string_t *
+ecma_new_symbol_from_descriptor_string (ecma_value_t string_desc) /**< ecma-string */
+{
+  JERRY_ASSERT (!ecma_is_value_symbol (string_desc));
+
+  ecma_string_t *symbol_p = ecma_alloc_string ();
+  symbol_p->refs_and_container = ECMA_STRING_REF_ONE | ECMA_STRING_CONTAINER_HEAP_UTF8_SYMBOL;
+  symbol_p->u.symbol_descriptor = string_desc;
+  symbol_p->hash = string_desc & ECMA_SYMBOL_HASH_MASK;
```
legendecas: Could symbols with identical descriptions being treated as unique keys be achieved by performing a bit-and an incremental counter here?

jerry-core/ecma/base/ecma-helpers.h
rerobika: Good catch, I've added to the TODO list. I'm going to apply your suggestions after the PR got an internal review.

jerry-core/ecma/base/ecma-helpers-string.c
rerobika: Not quite so. Each symbol is unique since we allocate a new string for it. The address of the new string will be the unique key. The hash property only stands for performance reasons.

legendecas: > Each symbol is unique since we allocate a new string for it.

I'm wondering if JerryScript would reuse string value `'foo'` in following case?

```js
var symbol1 = Symbol('foo')
var symbol2 = Symbol('foo')
```

If it does, then both `symbol1` and `symbol2` has same hash value, which ended with same property entry:

```js
var obj = {}
obj[symbol1] = 'foo'
obj[symbol2] = 'bar' // overrides obj[symbol1]
```

rerobika: 1. Yes, the engine reuses the `foo`. 
`symbol1` is a new `ecma-string` which container is `ECMA_STRING_CONTAINER_HEAP_UTF8_SYMBOL`, and the `foo` description string is stored in the `symbol_p->u.symbol_descriptor`. So it only puts and extra reference to `foo`.

In case of `symbol2` exactly happens the same, so `foo` now has +2 extra reference.

2. That test case is captures a real issue, due to the same hash. Now I see what you were trying to point on. Thanks for the reporting, I'm gonna fix it.

legendecas:
It seems that `ecma_builtin_object_object_define_property` asserts property name to be string type, which fails `Object.defineProperty(obj, symbol, descriptor)`.

rerobika:
Good catch, thanks for the reporting. Added to the TODO list.

jerry-core/ecma/base/ecma-helpers-value.c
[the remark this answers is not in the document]
zherczeg: I am ok with the macro.

jerry-core/ecma/builtin-objects/ecma-builtin-symbol-prototype.inc.h
```diff
@@ -0,0 +1,41 @@
...
+              ECMA_BUILTIN_ID_SYMBOL,
+              ECMA_PROPERTY_CONFIGURABLE_WRITABLE)
+
+NUMBER_VALUE (LIT_MAGIC_STRING_LENGTH,
+              0,
+              ECMA_PROPERTY_FLAG_WRITABLE)
+
+ROUTINE (LIT_MAGIC_STRING_TO_STRING_UL, ecma_builtin_symbol_prototype_object_to_string, 0, 0)
+ROUTINE (LIT_MAGIC_STRING_VALUE_OF_UL, ecma_builtin_symbol_prototype_object_value_of, 0, 0)
+
```
zherczeg: One newline.

jerry-core/ecma/builtin-objects/ecma-builtin-symbol.c
```diff
@@ -0,0 +1,240 @@
...
+    new_item_p->values[i] = JMEM_CP_NULL;
+  }
+
+  JMEM_CP_SET_POINTER (new_item_p->next_cp, JERRY_CONTEXT (symbol_list_first_p));
+  JERRY_CONTEXT (symbol_list_first_p) = new_item_p;
+
+  return ecma_copy_value (ecma_make_symbol_value (new_symbol_p));
+} /* ecma_builtin_symbol_for_helper */
+
+
```
zherczeg: One newline.

jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -224,6 +225,13 @@ ecma_op_to_boolean (ecma_value_t value) /**< ecma value */
     return !ecma_string_is_empty (str_p);
   }
 
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+  if (ecma_is_value_symbol (value))
+  {
+    return true;
+  }
+#endif /* !CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN */
+
```
zherczeg: Combine this with the code below.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -937,6 +941,14 @@ ecma_op_object_put (ecma_object_t *object_p, /**< the object */
         && ecma_get_object_extensible (object_p))
     {
       const ecma_object_type_t obj_type = ecma_get_object_type (object_p);
+      uint8_t prop_opts = ECMA_PROPERTY_CONFIGURABLE_ENUMERABLE_WRITABLE;
+
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+      if (JERRY_UNLIKELY (ecma_prop_name_is_symbol (property_name_p)))
+      {
+        prop_opts &= (uint8_t) ~ECMA_PROPERTY_FLAG_ENUMERABLE;
```
zherczeg: This is wrong!

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -198,6 +198,26 @@ ecma_new_ecma_string_from_magic_string_ex_id (lit_magic_string_ex_id_t id) /**<
...
+ */
+ecma_string_t *
+ecma_new_symbol_from_descriptor_string (ecma_value_t string_desc) /**< ecma-string */
+{
+  JERRY_ASSERT (!ecma_is_value_symbol (string_desc));
+
+  ecma_string_t *symbol_p = ecma_alloc_string ();
+  symbol_p->refs_and_container = ECMA_STRING_REF_ONE | ECMA_STRING_CONTAINER_HEAP_UTF8_SYMBOL;
+  symbol_p->u.symbol_descriptor = string_desc;
+  symbol_p->hash = (uint16_t) (((uintptr_t) symbol_p) >> 3);
```
zherczeg: I think we discussed 1 bit for built-ins. Is this changed? Also a define would be good for this constant.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -540,6 +558,21 @@ ecma_make_string_value (const ecma_string_t *ecma_string_p) /**< string to refer
...
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+/**
+ * Symbol value constructor
+ *
+ * @return ecma-value representation of the string argument
+ */
+inline ecma_value_t JERRY_ATTR_PURE JERRY_ATTR_ALWAYS_INLINE
+ecma_make_symbol_value (const ecma_string_t *ecma_symbol_p) /**< symbol to reference in value */
+{
+  JERRY_ASSERT (ecma_symbol_p != NULL);
```
zherczeg: Please add an assert that string type is symbol. Also add a check for ecma_make_string_value to exclude symbols.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -669,16 +698,31 @@ ecma_builtin_object_object_get_own_property_descriptor (ecma_value_t this_arg, /
...
+
+    if (ECMA_IS_VALUE_ERROR (name_str_value))
+    {
+      return name_str_value;
+    }
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+  }
+#endif /* !CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN */
 
   ecma_string_t *name_str_p = ecma_get_string_from_value (name_str_value);
```
zherczeg: Maybe a funcion `ecma_get_string_or_symbol_from_value` could be introduced to avoid possible confusion.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -886,38 +930,52 @@ ecma_builtin_object_object_define_property (ecma_value_t this_arg, /**< 'this' a
...
+  }
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (arg1);
+
+  ecma_value_t name_str_value;
+
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+  if (ecma_is_value_symbol (arg2))
+  {
+    name_str_value = ecma_copy_value (ECMA_CONVERT_SYMBOL_TO_STRING (arg2));
```
zherczeg: This looks like an ugly hack.

jerry-core/ecma/builtin-objects/ecma-builtin-symbol.c
```diff
@@ -0,0 +1,239 @@
...
+ */
+
+/**
+ * Handle calling [[Call]] of built-in Symbol object.
+ *
+ * @return ecma value
+ */
+ecma_value_t
+ecma_builtin_symbol_dispatch_call (const ecma_value_t *arguments_list_p, /**< arguments list */
+                                    ecma_length_t arguments_list_len) /**< number of arguments */
```
zherczeg: misaligned line.

jerry-core/ecma/builtin-objects/ecma-builtin-symbol.c
```diff
@@ -0,0 +1,239 @@
...
+ * Symbol constructor is not intended to be used
+ * with the new operator or to be subclassed.
+ *
+ * See also:
+ *          ECMA-262 v6, 19.4.1
+ * @return ecma value
+ */
+ecma_value_t
+ecma_builtin_symbol_dispatch_construct (const ecma_value_t *arguments_list_p, /**< arguments list */
+                                         ecma_length_t arguments_list_len) /**< number of arguments */
```
zherczeg: Ditto.

tests/jerry/es2015/symbol.js
```diff
@@ -0,0 +1,111 @@
...
+
+assert (a === b);
+assert (a == b);
+
+assert (Symbol.keyFor (a) === "foo")
+assert (Symbol.keyFor (b) === "foo")
+
+// Test built-in symbols
+var a = ["hasInstance",
+        "isConcatSpreadable",
```
zherczeg: Misaligned lines.

tests/jerry/es2015/object-get_own_property_symbols.js
```diff
@@ -0,0 +1,103 @@
...
+Parent.prototype.inheritedMethod = function() {};
+
+function Child() {
+  this[a] = 5;
+  this[b] = function() {};
+}
+Child.prototype = new Parent;
+Child.prototype.prototypeMethod = function() {};
+
+props = Object.getOwnPropertySymbols (new Child());
```
zherczeg: What about symbols and enumeration? It would be good to add a for-in test for enumerable symbols (not included), and also check the effect of enumerable/not-enumerable symbols on getOwnPropertySymbols.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -361,6 +362,9 @@ typedef enum
                                        *   that are not indices */
   ECMA_LIST_ENUMERABLE = (1 << 1), /**< exclude non-enumerable properties */
   ECMA_LIST_PROTOTYPE = (1 << 2), /**< list properties from prototype chain */
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+  ECMA_LIST_SYMBOLS = (1 << 3), /**< exclude non-symbol properties */
```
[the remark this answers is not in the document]
rerobika: I think the description is correct. If this flag is set the non-symbol properties should be skipped during listing. (I followed the naming conventions of this enum, see `ECMA_LIST_ARRAY_INDICES` and `ECMA_LIST_ENUMERABLE`.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1328,6 +1344,8 @@ typedef enum
                                              stored locally in the string's descriptor */
   ECMA_STRING_CONTAINER_MAGIC_STRING_EX, /**< the ecma-string is equal to one of external magic strings */
 
+  ECMA_STRING_CONTAINER_HEAP_UTF8_SYMBOL, /**< symbol data is a reference to an already existing ecma-string */
```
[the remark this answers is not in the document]
rerobika: Exactly.

jerry-core/ecma/base/ecma-helpers-value.c
rerobika: The first part is OK for me, but the second is not. Since all symbol is a special type `ecma-string` and the whole property management system depends on that the prop-name is `ecma-string` and these strings are passed via `ecma_values` this assert prevents to use symbols as property names.

tests/jerry/regression-test-issue-116.js
```diff
@@ -19,5 +19,17 @@ try
 }
 catch (e)
 {
-  assert (e instanceof ReferenceError);
+  var error_type = ReferenceError;
```
zherczeg: Move this test to es5.1 directory.

rerobika:
@zherczeg I've applied your suggestions, also rebased the PR to the latest master.

galpeter:
Should we also add api for symbols? At least for checking if a value is a symbol or not?

rerobika:
@galpeter Yes we should. This patch is quite a big right now, it will be part the follow-up patches.

tests/jerry/es2015/symbol.js
```diff
@@ -0,0 +1,109 @@
...
+assert (keys.toString() === 'c,d')
+assert (keys.length === 2)
+
+var c = Symbol ('bar');
+var o2 = {};
+o2[b] = 'foo'
+o2[c] = 'bar'
+
+assert (o2[b] == 'foo')
+assert (o2[c] == 'bar')
```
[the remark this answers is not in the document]
galpeter: Also it would be good to have a test like:
```js
sym_a = Symbol('AA');
obj = { AA: 55 };
obj[sym_a] = 77;
```
This would check if the properties does not overwrite each other.

jerry-core/lit/lit-magic-strings.inc.h
```diff
@@ -286,6 +304,10 @@ LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_PROMISE_UL, "Promise")
 #if !defined (CONFIG_DISABLE_MATH_BUILTIN)
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_SQRT1_2_U, "SQRT1_2")
 #endif
+#if !defined (CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN)
+LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_SYMBOL_LEFT_PAREN_UL, "Symbol(")
+LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_SYMBOL_DOT_UL, "Symbol.")
```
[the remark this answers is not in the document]
rerobika: The reason behind the 2 extra magic strings that these string should be created run-time anyway. It reduces the number of memory allocations, also if for some reason a `Symbol.` string would be constructed somehow during the execution it does not requires to allocate memory for it, since it is a magic string. But in the other hand, yes it consumes + 14B memory, but I think it's an acceptable deal.

jerry-core/ecma/builtin-objects/ecma-builtin-symbol.c
```diff
@@ -0,0 +1,239 @@
...
+        ecma_string_t *value_p = JMEM_CP_GET_NON_NULL_POINTER (ecma_string_t,
+                                                               symbol_list_p->values[i]);
+
+        if (is_for)
+        {
+          ecma_string_t *symbol_desc_p = ecma_get_symbol_description (value_p);
+
+          if (ecma_compare_ecma_strings (symbol_desc_p, string_p))
+          {
+            ecma_deref_ecma_string (symbol_desc_p);
```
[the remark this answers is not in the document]
rerobika: The reason is [this](https://github.com/rerobika/jerryscript/blob/symbols/jerry-core/ecma/builtin-objects/ecma-builtin-symbol.c#L194) line. Since `ecma_op_to_string` increases the strings ref count or allocates a new string for non-string values. Also this `free` step could be done in the end of `ecma_builtin_symbol_for` but during the comparison it's easier to do it, since the string is not wrapped to an `ecma_value_t`.

tests/jerry/es2015/object-get-own-property-symbols.js
```diff
@@ -0,0 +1,118 @@
...
+assert (props.length === 0);
+
+// Test object
+var obj = {};
+obj[a] = 'a';
+obj[b] = 'b';
+obj[c] = 'c';
+obj[d] = 'd';
+props = Object.getOwnPropertySymbols (obj);
+// props should not contain: a, b, c, d and the order is not defined!
```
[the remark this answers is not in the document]
rerobika: Good catch, that was a typo in the description!

tests/jerry/es2015/object-get-own-property-symbols.js
```diff
@@ -0,0 +1,118 @@
...
+var c = Symbol ('foo')
+var d = Symbol ('foo')
+
+var obj = {}
+obj[a] = 'foo';
+obj[b] = 'bar';
+obj[c] = 'foobar';
+obj[d] = 'barfoo';
+
+props = Object.getOwnPropertySymbols (obj);
```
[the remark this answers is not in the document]
rerobika: The difference is that in this case all Symbol property keys have the same descriptor string, to show even the descriptors are the same, the symbols are unique.

rerobika:

tests/jerry/es2015/object-get-own-property-symbols.js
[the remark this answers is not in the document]
galpeter: Ohh I see, the `a,b,c,d` vars have new value here. It's a bit akward to have the same variables with multiple `var`. Would it be possible to use different names? just to make any future debugging less error prone.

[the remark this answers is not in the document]
rerobika: Of course!

rerobika:
The conflicts are resolved and the patch is updated to the latest master.

akosthekiss:
Let me chime in. I've seen that the PR touches non-ES2015 code paths. A quick comparison has given me this on the binary size:

Binary (bytes) | master 325c426c1e | PR b7818c4905 | Gain |
.text | 115604 | 116282 | -0.586% |
.rodata | 11256 | 11248 | +0.071% |

That's +0.6K code in the ES5.1 config for an ES2015-only feature. Could you comment on this?

rerobika:
@akosthekiss The reason of this binary size change is that this PR does not contain #2646, since it has just landed. I'll rebase the PR and probably this difference will disappear.

rerobika:
@akosthekiss The PR is rebased, these are the new results:

Binary (bytes) | 08c7183ef8 | f2abf61709 | Gain |
.text | 115604 | 115588 | +0.014% |
.rodata | 11256 | 11256 | 0.000% |

PR #2606, opened 2018-11-21
Fix an issue while debugging with static snapshots

DanielBallaSZTE:
The issue was present when the executed code would have thrown an error.

Co-authored-by: Robert Fancsik frobert@inf.u-szeged.hu

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -3255,7 +3255,8 @@ error:
                                   | JERRY_DEBUGGER_VM_EXCEPTION_THROWN);
 
       if ((JERRY_CONTEXT (debugger_flags) & JERRY_DEBUGGER_CONNECTED)
-          && !(frame_ctx_p->bytecode_header_p->status_flags & CBC_CODE_FLAGS_DEBUGGER_IGNORE)
+          && !(frame_ctx_p->bytecode_header_p->status_flags
+               & (CBC_CODE_FLAGS_DEBUGGER_IGNORE | CBC_CODE_FLAGS_STATIC_FUNCTION))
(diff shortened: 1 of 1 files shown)


DanielBallaSZTE:
@akosthekiss Sorry for not adding it to the PR description, going to amend it to make sure it appears there as well.
So the failure was found when debugging IoT.js with static snapshots turned on, if an error was thrown an assertion failure was caused. The reason is the frame context's `bytecode_header_p` pointer was not a heap pointer. The `jerry_debugger_breakpoint_hit` function tries to set a compressed pointer which points to `bytecode_header_p` with the `JMEM_CP_SET_NON_NULL_POINTER` macro, which has an assertion that requires the above mentioned `bytecode_header_p` to be a heap pointer, which is obviously not.

akosthekiss:
@DanielBallaSZTE Is this issue IoT.js-specific? Or can it be reproduced somehow in a JerryScript-only environment?

DanielBallaSZTE:
@akosthekiss I haven't tried it yet, I could give it a shot, and if it is reproducible try and create a test for it. However, in that case we would need a static snapshot in the test environment to be debugged, while running debugger-tests.

zherczeg:
Not a trivial task. You need a unit test which creates such an environment, and you also need the python debugger to do debugging. Hence it needs a new infrastructure basically, where instead of a .js, a binary is run by the debugger tester driver. @DanielBallaSZTE please check how difficult would be to do this.

LaszloLango:
Any update on this? IMHO the test environment improvement should be done in a separate PR. It is not trivial and might take a while to implement it, but this bug is quite a blocker issue for us.

PR #2608, opened 2018-11-21
Fix possible failure in backtrace info

DanielBallaSZTE:
Make sure to ignore static snapshots when sending backtrace information to the debugger.

### DIFF
--- a/jerry-core/debugger/debugger.c
+++ b/jerry-core/debugger/debugger.c
@@ -160,7 +160,8 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
 
     while (frame_ctx_p != NULL && min_depth_offset++ < max_depth)
     {
-      if (frame_ctx_p->bytecode_header_p->status_flags & CBC_CODE_FLAGS_DEBUGGER_IGNORE)
+      if (frame_ctx_p->bytecode_header_p->status_flags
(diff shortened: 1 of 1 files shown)


akosthekiss:
Related / follow-up to #2606 .

jerry-core/debugger/debugger.c
```diff
@@ -160,7 +160,8 @@ jerry_debugger_send_backtrace (const uint8_t *recv_buffer_p) /**< pointer to the
 
     while (frame_ctx_p != NULL && min_depth_offset++ < max_depth)
     {
-      if (frame_ctx_p->bytecode_header_p->status_flags & CBC_CODE_FLAGS_DEBUGGER_IGNORE)
+      if (frame_ctx_p->bytecode_header_p->status_flags
+          & (CBC_CODE_FLAGS_DEBUGGER_IGNORE | CBC_CODE_FLAGS_STATIC_FUNCTION))
```
zherczeg: There is the total frame count computing above, that is probably invalid as well. Maybe @robertsipka scope chain code is affected as well, since parts of the scope chain should be hidden as well.

akosthekiss:
@DanielBallaSZTE ping. According to @zherczeg 's comment, it seems that this fix may need further changes. Will those be worked on?

DanielBallaSZTE:
@akosthekiss I think this one can be closed now, as this is not a blocker issue.
This PR needs some more work, however, it would be nice to also add a test environment for these bugs. I will update later with a new Pull Request, or reopen this one with some changes.

PR #2610, opened 2018-11-21
Fix JerryScript build with clang-6.0

DanielBallaSZTE:
Clang-6.0 would throw the following error while building

```
jerryscript/jerry-core/api/jerry.c:1527:71: error: implicit conversion loses integer precision:
      'jerry_regexp_flags_t' to 'uint16_t' (aka 'unsigned short') [-Werror,-Wconversion]
  jerry_value_t ret_val = ecma_op_create_regexp_object (ecma_pattern, flags);
```

This patch fixes the issue.
Also change the `jerry_create_regexp` and `jerry_create_regexp_sz` functions' `flags` parameter to `uint16_t` since the values created with `bitwise inclusive OR` are not part of the enum.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -3421,16 +3421,17 @@ jerry_create_string_sz (const jerry_char_t *str_p,
 **Summary**
 
 Returns a jerry_value_t RegExp object or an error, if the construction of the object fails.
-Optional flags can be set using [jerry_regexp_flags_t](#jerry_regexp_flags_t);
+Optional flags can be set using [jerry_regexp_flags_t](#jerry_regexp_flags_t).
+These flags can be combined together with the binary OR operator or used on their own as enum values.
 
 **Prototype**
 ```c
 jerry_value_t
-jerry_create_regexp (const jerry_char_t *pattern_p, jerry_regexp_flags_t flags);
+jerry_create_regexp (const jerry_char_t *pattern_p, uint16_t flags);
 ```
 
 - `pattern_p` - the RegExp pattern as a zero-terminated UTF-8 string
-- `flags` - optional flags for the RegExp object
+- `flags` - optional flags for the RegExp object, see [jerry_regexp_flags_t](#jerry_regexp_flags_t)
 - return value - the RegExp object as a `jerry_value_t`
 
 **Example**
@@ -3438,7 +3439,7 @@ jerry_create_regexp (const jerry_char_t *pattern_p, jerry_regexp_flags_t flags);
 ```c
 {
   jerry_char_t pattern_p = "[cgt]gggtaaa|tttaccc[acg]";
-  jerry_regexp_flags_t pattern_flags = JERRY_REGEXP_FLAG_IGNORE_CASE;
+  uint16_t pattern_flags = JERRY_REGEXP_FLAG_IGNORE_CASE;
 
   jerry_value_t regexp = jerry_create_regexp (pattern_p, pattern_flags);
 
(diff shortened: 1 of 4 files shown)


jerry-core/api/jerry.c
```diff
@@ -1524,7 +1524,7 @@ jerry_create_regexp_sz (const jerry_char_t *pattern_p, /**< zero-terminated UTF-
 
   ecma_string_t *ecma_pattern = ecma_new_ecma_string_from_utf8 (pattern_p, pattern_size);
 
-  jerry_value_t ret_val = ecma_op_create_regexp_object (ecma_pattern, flags);
+  jerry_value_t ret_val = ecma_op_create_regexp_object (ecma_pattern, (uint16_t) flags);
```
galpeter: I see that there is a `re_flags_t` enum in the `ecma-regexp-object.h` should be use that type here instead (and possibly in the `ecma_op_create_regexp_object` and other relevant methods)?

zherczeg: The problem is we use the combination of flags, rather than a single flag. enum -> int conversions has strange rules, some compilers use the smallest type, some compilers use int all the time.

galpeter: then the usage of the `jerry_regexp_flags_t` enum type in the function is also not correct, or not?

DanielBallaSZTE: good point

docs/02.API-REFERENCE.md
```diff
@@ -3426,7 +3426,7 @@ Optional flags can be set using [jerry_regexp_flags_t](#jerry_regexp_flags_t);
 **Prototype**
 ```c
 jerry_value_t
-jerry_create_regexp (const jerry_char_t *pattern_p, jerry_regexp_flags_t flags);
+jerry_create_regexp (const jerry_char_t *pattern_p, uint16_t flags);
```
akosthekiss: Now, there is no obvious connection between the `flags` argument of `jerry_create_regexp` and the enum type `jerry_regexp_flags_t` anymore. So, the doc of `flags` should be updated to mention that it can/has to be a combination of values from the `jerry_regexp_flags_t` enum.

docs/02.API-REFERENCE.md
```diff
@@ -3459,7 +3459,7 @@ Optional flags can be set using [jerry_regexp_flags_t](#jerry_regexp_flags_t);
 **Prototype**
 ```c
 jerry_value_t
-jerry_create_regexp_sz (const jerry_char_t *pattern_p, jerry_size_t pattern_size, jerry_regexp_flags_t flags);
+jerry_create_regexp_sz (const jerry_char_t *pattern_p, jerry_size_t pattern_size, uint16_t flags);
```
akosthekiss: same here

docs/02.API-REFERENCE.md
DanielBallaSZTE: There actually is already a reference in the `summary` of the function, however I added the same reference to the prototype description to make sure no one misses it.

docs/02.API-REFERENCE.md
DanielBallaSZTE: See my comment above.

jerry-core/include/jerryscript-core.h
```diff
@@ -458,9 +458,8 @@ jerry_value_t jerry_create_number_nan (void);
 jerry_value_t jerry_create_null (void);
 jerry_value_t jerry_create_object (void);
 jerry_value_t jerry_create_promise (void);
-jerry_value_t jerry_create_regexp (const jerry_char_t *pattern, jerry_regexp_flags_t flags);
-jerry_value_t jerry_create_regexp_sz (const jerry_char_t *pattern, jerry_size_t pattern_size,
-                                      jerry_regexp_flags_t flags);
+jerry_value_t jerry_create_regexp (const jerry_char_t *pattern, uint16_t flags);
```
zherczeg: Is there a reasion why this is limited to 16 bits? Usually this takes a register regardless of type. I would use a 32 bit value.

DanielBallaSZTE: We could use a 32 bit value, however that would go way out of the scope of this patch.
In that case we must change everything related to `ecma_op_create_regexp_object`. I would like to leave it as it is in this patch, and create a follow up, that changes every related function's parameters to 32 bit values.

docs/02.API-REFERENCE.md
akosthekiss: The summary description reads "Optional flags can be set using jerry_regexp_flags_t". This is a bit vague about _how_ to use that type, just like the newly added text. It is not clear whether only single enum values can be used or a combination of them, too.

Also, please fix the semicolon at the end of the summary text. (I cannot make a comment there.)

docs/02.API-REFERENCE.md
akosthekiss: See my comment above.

jerry-core/include/jerryscript-core.h
akosthekiss: I think @DanielBallaSZTE is right here. The API mirrors the internals.

On the other hand, if the goal is to break away from the internals, then I'd suggest _not_ to use a bitwidth-fixed (e.g., 32 bit) type, but a simple (`unsigned`?) `int` instead. Actually, that should be the fastest and easiest and most future-proof type on any machine.

DanielBallaSZTE: @akosthekiss Using an `unsigned int` requires you to cast it to `unsigned short int` (`uint16_t`) when calling the `ecma_op_create_regexp_object`. I already have a patch prepared that changes all the internal parameters to `uint32_t`, so I would like to leave it as an `uint16_t` for now and submit the patch if this one lands.

akosthekiss: Actually, I was not pushing for a change here. I was only reflecting to @zherczeg 's proposal to change this to `uint32_t`. And my comment was that _if_ we wanted a change (i.e., to have an API type that was different from the internal type) _then_ we should go for a non-sized type.

PR #2616, opened 2018-11-23
Reduce ecma_{ref, deref}_object calls while using ecma_builtin_get

rerobika:
The following stucture was highly frequented in the code base:

 - Get a builtin object // This operation increases the reference count of the object
 - Use it for create a new object
 - Deref the builtin object

After a builtin has been instantiated there is always at least one reference to "keep it alive",
so increase/decrease the reference count for getting the value only is unnecessary.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.c
@@ -155,6 +155,7 @@ static const ecma_builtin_property_list_reference_t ecma_builtin_property_list_r
 static size_t
 ecma_builtin_get_property_count (ecma_builtin_id_t builtin_id) /**< built-in ID */
 {
+  JERRY_ASSERT (builtin_id < ECMA_BUILTIN_ID__COUNT);
   const ecma_builtin_property_descriptor_t *property_list_p = ecma_builtin_property_list_references[builtin_id];
 
   const ecma_builtin_property_descriptor_t *curr_property_p = property_list_p;
@@ -180,20 +181,16 @@ ecma_builtin_is (ecma_object_t *obj_p, /**< pointer to an object */
   JERRY_ASSERT (obj_p != NULL && !ecma_is_lexical_environment (obj_p));
   JERRY_ASSERT (builtin_id < ECMA_BUILTIN_ID__COUNT);
 
-  if (JERRY_CONTEXT (ecma_builtin_objects)[builtin_id] == NULL)
-  {
-    /* If a built-in object is not instantiated,
-     * the specified object cannot be the built-in object */
-    return false;
-  }
-  else
-  {
-    return (obj_p == JERRY_CONTEXT (ecma_builtin_objects)[builtin_id]);
-  }
+  /* If a built-in object is not instantiated, its value is NULL,
+     hence it cannot be equal to a valid object. */
+  return (obj_p == JERRY_CONTEXT (ecma_builtin_objects)[builtin_id]);
 } /* ecma_builtin_is */
 
 /**
  * Get reference to specified built-in object
+ *
+ * Note:
+ *   Does not increase the reference counter.
  *
  * @return pointer to the object's instance
  */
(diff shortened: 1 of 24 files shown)


rerobika:
Performance results:

Benchmark | Stack (bytes) | Perf (sec) |
----: | ----: | ----: | 
3d-cube.js | 15896 -> 15888 : +0.050% | 0.809 -> 0.791 : +2.221% | 
3d-raytrace.js | 2848 -> 2840 : +0.281% | 1.039 -> 1.028 : +1.029% | 
access-binary-trees.js | 3312 -> 3304 : +0.242% | 0.558 -> 0.563 : -0.851% | 
access-fannkuch.js | 1592 -> 1584 : +0.503% | 2.096 -> 2.055 : +1.945% | 
access-nbody.js | 1568 -> 1560 : +0.510% | 1.086 -> 1.073 : +1.207% | 
bitops-3bit-bits-in-byte.js | 1536 -> 1528 : +0.521% | 0.506 -> 0.489 : +3.505% | 
bitops-bits-in-byte.js | 1536 -> 1528 : +0.521% | 0.665 -> 0.636 : +4.296% | 
bitops-bitwise-and.js | 1312 -> 1304 : +0.610% | 0.926 -> 0.905 : +2.231% | 
bitops-nsieve-bits.js | 1536 -> 1528 : +0.521% | 1.132 -> 1.103 : +2.540% | 
controlflow-recursive.js | 69792 -> 69784 : +0.011% | 0.367 -> 0.363 : +1.302% | 
crypto-aes.js | 2424 -> 2416 : +0.330% | 0.876 -> 0.863 : +1.464% | 
crypto-md5.js | 1932 -> 1924 : +0.414% | 0.599 -> 0.599 : -0.111% | 
crypto-sha1.js | 1648 -> 1640 : +0.485% | 0.594 -> 0.593 : +0.208% | 
date-format-tofte.js | 2064 -> 2056 : +0.388% | 0.747 -> 0.748 : -0.178% | 
date-format-xparb.js | 2504 -> 2496 : +0.319% | 0.522 -> 0.513 : +1.637% | 
math-cordic.js | 1536 -> 1528 : +0.521% | 1.174 -> 1.152 : +1.891% | 
math-partial-sums.js | 1496 -> 1488 : +0.535% | 0.736 -> 0.731 : +0.639% | 
math-spectral-norm.js | 1604 -> 1596 : +0.499% | 0.541 -> 0.535 : +1.199% | 
string-base64.js | 1568 -> 1560 : +0.510% | 1.403 -> 1.414 : -0.794% | 
string-fasta.js | 1520 -> 1512 : +0.526% | 1.342 -> 1.348 : -0.517% | 
Geometric mean: | **+0.415%** | **+1.252%** | 

Binary sizes (bytes)
5533b8ba9b:132664
bfd9c7be6c:132664

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -180,16 +181,9 @@ ecma_builtin_is (ecma_object_t *obj_p, /**< pointer to an object */
...
-    /* If a built-in object is not instantiated,
-     * the specified object cannot be the built-in object */
-    return false;
-  }
-  else
-  {
-    return (obj_p == JERRY_CONTEXT (ecma_builtin_objects)[builtin_id]);
-  }
+  /* If a built-in object is not instantiated,
+   * the specified object cannot be the built-in object */
```
zherczeg: I would change this comment:

If a built-in object is not instantiated, its value is NULL and that cannot be equal to a valid object.

rerobika: Good point! I've modified it.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -207,8 +201,6 @@ ecma_builtin_get (ecma_builtin_id_t builtin_id) /**< id of built-in to check on
     ecma_instantiate_builtin (builtin_id);
   }
 
-  ecma_ref_object (JERRY_CONTEXT (ecma_builtin_objects)[builtin_id]);
```
galpeter: We should add a note to the method's description that the object's ref count is not increased just like in the case of the `ecma_builtin_get_global` call.

rerobika: Good catch!

PR #2497, opened 2018-08-29
targets: zephyr: Update for Zephyr 1.13

pfalcon:
Zephyr 1.13 made some refactorings to its CMake infrastructure which must
be propagated to applications.

Also, some config options were removed/refactored, so cleanup prj.conf.

JerryScript-DCO-1.0-Signed-off-by: Paul Sokolovsky paul.sokolovsky@linaro.org

PR #2617, opened 2018-11-23
String construction buffer

mmatyas:
These patches implement a string construction buffer, which can be used to greatly improve the
memory usage and performance of string concatenation and JSON serialization.


mmatyas:
Some numbers:

strbuf__wip02_arm

Note that the high heap usage in the case of `string-fasta` is due to the garbage collection. The test runs perfectly with as low as 8KB of heap size.

akosthekiss:
Non-technical feedbacks:
- There is a typo in the PR/commit title (contruction).
- The PR contains 2 commits, and the second commit refers to a "realloc support recently added to jmem". I'm not aware of such support, except for the first commit also in this PR. This should be cleared up: either another PR should be opened for the first commit on its own, to be reviewed separately, or this PR should have a single squashed commit with a self-contained commit message.

mmatyas:
Yes, the second commit is based on the first one, but on its own that is not used anywhere else yet, which is why they are together in this issue. Would you prefer separate pull requests?

akosthekiss:
@mmatyas from https://github.com/jerryscript-project/jerryscript/pull/2497#issuecomment-420271553 :
> project custom is one commit per PR -- if changes are independent enough to be broken into separate commits then separate PRs should also be opened. And each PR/commit shall pass on the CI on its own

Whether the changes should be separated to two commits or merged in one, I leave that decision up to you. But _if_ you decide that the `realloc` functionality is a standalone change then you should open a PR for it and get it reviewed on its own, then you can reference it from another (e.g., this) PR. On the other hand, if the decision is to have one PR only then these two commits must be squashed and the commit message has to be self-contained.

mmatyas:
Ok, I've moved the jmem changes to a separate PR (#2619), I'll rebase this branch then when it gets merged.

PR #2618, opened 2018-11-23
Improve branch prediction in vm_loop

rerobika:
There are several instructions in the vm  which create boolean result and put it onto the stack. (equality, relation, push_{true,false}, etc..).
Until now, these results were popped from the stack, then got converted to boolean (unnecessary step) and
after the branch prediction they have been freed (also superflous, since they are simple values).

The patch contains a pre-check before emitting a branch-if-equal opcode for whether the last opcode produces boolean result.
If this new byte-code has not been emmitted the possible jump can be executed while decoding the branch offset.

Advantages:
 - No need to convert the value to boolean
 - No need to free it
 - No need to switch the group index of the current opcode

Disadvantages:
 - In special cases it increases the generated bytecode size by 1 byte/branch instruction


rerobika:
Performance results:

Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 0.809 -> 0.788 : +2.598% | 
3d-raytrace.js | 1.039 -> 1.017 : +2.086% | 
access-binary-trees.js | 0.558 -> 0.551 : +1.395% | 
access-fannkuch.js | 2.096 -> 2.032 : +3.044% | 
access-nbody.js | 1.086 -> 1.066 : +1.821% | 
bitops-3bit-bits-in-byte.js | 0.506 -> 0.485 : +4.287% | 
bitops-bits-in-byte.js | 0.665 -> 0.653 : +1.740% | 
bitops-bitwise-and.js | 0.926 -> 0.916 : +1.084% | 
bitops-nsieve-bits.js | 1.132 -> 1.105 : +2.364% | 
controlflow-recursive.js | 0.367 -> 0.352 : +4.209% | 
crypto-aes.js | 0.876 -> 0.862 : +1.542% | 
crypto-md5.js | 0.599 -> 0.587 : +1.876% | 
crypto-sha1.js | 0.594 -> 0.574 : +3.441% | 
date-format-tofte.js | 0.747 -> 0.730 : +2.269% | 
date-format-xparb.js | 0.522 -> 0.516 : +1.218% | 
math-cordic.js | 1.174 -> 1.139 : +2.919% | 
math-partial-sums.js | 0.736 -> 0.714 : +2.979% | 
math-spectral-norm.js | 0.541 -> 0.528 : +2.420% | 
string-base64.js | 1.403 -> 1.388 : +1.050% | 
string-fasta.js | 1.342 -> 1.321 : +1.496% | 
Geometric mean: | **+2.296%** | 

Binary sizes (bytes)
5533b8ba9b:132664
06e2b02d58:132664

PR #2619, opened 2018-11-26
Add realloc function to jmem

mmatyas:
This patch extends jmem functionality by adding a realloc function.
This opens up other paths of optimization which can result in smaller
peak memory usage and faster execution times, due to not having to
duplicate memory when we need to extend blocks.

### DIFF
--- a/jerry-core/jmem/jmem-heap.c
+++ b/jerry-core/jmem/jmem-heap.c
@@ -382,26 +382,16 @@ jmem_heap_alloc_block_null_on_error (const size_t size) /**< required memory siz
   return jmem_heap_gc_and_alloc_block (size, true);
 } /* jmem_heap_alloc_block_null_on_error */
 
-/**
- * Free the memory block.
- */
-void JERRY_ATTR_HOT
-jmem_heap_free_block (void *ptr, /**< pointer to beginning of data space of the block */
-                      const size_t size) /**< size of allocated region */
-{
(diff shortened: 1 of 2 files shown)


mmatyas:
The feature this patch implements will be used for the string construction buffer optimizations -- see #2617.

akosthekiss:
PR author and commit author are quite different. Interestingly, CI is not unhappy about that. Should this get approved and landed (via squash&merge on the web interface), I'm not completely sure what will happen. If I recall correctly, GH will rewrite at least the committer (making it the person who has pushed the green button). Whether the author gets also rewritten, I'm not sure. But either @mmatyas or @dbatyai will lose credit. Just sayin'.

DanielBallaSZTE:
As far as I know the last line of the commit should be the author of the commit.
A good feature to use in this case is the `co-author` feature.
You can read more about it here:
https://help.github.com/articles/creating-a-commit-with-multiple-authors/

DanielBallaSZTE:
@akosthekiss 
I used it several times, and put it right before the sign-off message. It didn't conflict, and marked @rerobika  as the co-author in the commit.
(Reference: https://github.com/jerryscript-project/jerryscript/pull/2514)

zherczeg:
How this is compatible with system allocator? I am not sure all embedded systems has realloc.

akosthekiss:
@DanielBallaSZTE I cannot find the referenced commit among @rerobika 's contributions.

PR #2622, opened 2018-11-29
Remove unnecessary check from VM_OC_NOT

rerobika:
This patch removes the ECMA_IS_VALUE_ERROR check from VM_OC_NOT, since the general toBoolean operation cannot throw an exception.

### DIFF
--- a/jerry-core/vm/opcodes.c
+++ b/jerry-core/vm/opcodes.c
@@ -68,20 +68,6 @@ vm_var_decl (vm_frame_ctx_t *frame_ctx_p, /**< interpreter context */
 } /* vm_var_decl */
 
 /**
- * 'Logical NOT Operator' opcode handler.
- *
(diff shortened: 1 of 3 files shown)


jerry-core/vm/opcodes.c
```diff
@@ -75,7 +75,7 @@ vm_var_decl (vm_frame_ctx_t *frame_ctx_p, /**< interpreter context */
  * @return ecma value
  *         Returned value must be freed with ecma_free_value
  */
-ecma_value_t
+inline ecma_value_t
```
LaszloLango: `JERRY_ATTR_ALWAYS_INLINE` is missing

zherczeg: Hm, we could just remove this function.

rerobika: Well, since it's only a "jumper" function, may worth to remove it.

PR #2629, opened 2018-11-30
Use Thumb instructions on TizenRT.

rtakacs:
In order to decrease the binary size of JerryScript, enabled the Thumb instruction set that is supported on ARM Cortex-R series processors.

The following table shows the binary size reduction by this patch:

|          | commit: 8410570 (2018-11-28) | this patch|
|--------|-----------|--------|
text     |  199708 B |  130440 B|
bss	   |  74088 B |  74088 B|
data    |  0 B     |  0 B|
rodata |  12341 B |  12342 B|

### DIFF
--- a/cmake/toolchain_mcu_artik053.cmake
+++ b/cmake/toolchain_mcu_artik053.cmake
@@ -16,7 +16,9 @@ set(CMAKE_SYSTEM_NAME TizenRT)
 set(CMAKE_SYSTEM_PROCESSOR armv7l)
 set(CMAKE_SYSTEM_VERSION ARTIK053)
 
-set(FLAGS_COMMON_ARCH -mcpu=cortex-r4 -mfpu=vfpv3 -fno-builtin -fno-strict-aliasing -fomit-frame-pointer -fno-strength-reduce -Wall -Werror -Wshadow -Wno-error=conversion)
+set(FLAGS_COMMON_ARCH -mcpu=cortex-r4 -mthumb -mfpu=vfpv3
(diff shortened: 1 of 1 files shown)


cmake/toolchain_mcu_artik053.cmake
```diff
@@ -16,7 +16,7 @@ set(CMAKE_SYSTEM_NAME TizenRT)
 set(CMAKE_SYSTEM_PROCESSOR armv7l)
 set(CMAKE_SYSTEM_VERSION ARTIK053)
 
-set(FLAGS_COMMON_ARCH -mcpu=cortex-r4 -mfpu=vfpv3 -fno-builtin -fno-strict-aliasing -fomit-frame-pointer -fno-strength-reduce -Wall -Werror -Wshadow -Wno-error=conversion)
+set(FLAGS_COMMON_ARCH -mcpu=cortex-r4 -mthumb -mfpu=vfpv3 -fno-builtin -fno-strict-aliasing -fomit-frame-pointer -fno-strength-reduce -Wall -Werror -Wshadow -Wno-error=conversion)
```
LaszloLango: This line is way too long. It would be better to make it shorter, e.g:
```
set(FLAGS_COMMON_ARCH -mcpu=cortex-r4 -mthumb -mfpu=vfpv3
                      -fno-builtin -fno-strict-aliasing -fomit-frame-pointer -fno-strength-reduce
                      -Wall -Werror -Wshadow -Wno-error=conversion)
```

rtakacs:
@LaszloLango Thanks! I've updated the PR.

PR #2635, opened 2018-12-11
Add subdir-level gitignore to the nuttx target directory

pmarkee:
JerryScript-DCO-1.0-Signed-off-by: marpeter@inf.u-szeged.hu

### DIFF
--- a/targets/nuttx-stm32f4/.gitignore
+++ b/targets/nuttx-stm32f4/.gitignore
@@ -1,0 +1,7 @@
+# Files generated / copied by the NuttX build
+.built
+.depend
+Make.dep
+*.c.obj
+*.o
+*.a


pmarkee:
The nuttx build creates a bunch of `.obj` and `.a` files in the `targets/nuttx-stm32f4` directory. By adding an `all` rule to the `Makefile` in that directory, we can automatically clean these files.

targets/nuttx-stm32f4/Makefile
```diff
@@ -52,10 +52,12 @@ updateobjs:
 
 .PHONY: cleanlibs
 cleanlibs: updateobjs
-	rm -f $(OBJS)
+	rm -f $(OBJS) $(LIBS)
 
 clean: cleanlibs
 
+all: cleanlibs
```
rtakacs: This line should be after the `include $(APPDIR)/Application.mk` line. First, the `Application.mk::all` rule should be executed, and the cleanup should only happen after that.

akosthekiss: I'm not sure I can agree to this approach, i.e., to make the `all` target depend on `clean` or `cleanlibs`. This goes against the traditional roles of these rules. By convention, `all` is the default rule of a Makefile and it creates all 'final' artefacts, plus intermediate results alongside if needed. Re-execution of `make all`, again by convention, re-creates these artefacts, preferably by an incremental build, re-using existing intermediates which are results of unchanged sources. To be able to achieve this, it doesn't clean after itself but keeps all build results. If a developer wants to remove both final and intermediate results, then is it time for a `make clean`. I'd suggest keeping these behaviours as this is what developers expect.

N.B.: If `clean` doesn't clean well, then it should be fixed, of course. (But do _not_ mix `all` and `clean`.)

rtakacs: Please note that not the NuttX builds the JerryScript (by this makefile). JerryScript is built separately for the NuttX target ([2nd step](https://github.com/jerryscript-project/jerryscript/tree/master/targets/nuttx-stm32f4#2-build-jerryscript-for-nuttx) in the JerryScript build guide). This Makefile just copies all the static libraries from JerryScript and extract them to have the created object files. After that, these object files are collected and copied into a NuttX specific `libapps.a` library.

When you use `simlink` (instead of `copy`) to put the JerryScript application files into NuttX ([3rd step](https://github.com/jerryscript-project/jerryscript/tree/master/targets/nuttx-stm32f4#3-copy-jerryscripts-application-files-to-nuttx) in the JerryScript build guide), all the static libs and their object files will be copied into the Makefile's folder, that's in this case the `jerryscript/targets/nuttx-stm32f4/`. So JerryScript will have a lot of untracked files if they are not cleaned after the NuttX build.

@akosthekiss I agree with you that this is a bit strange to modify the `all` rule to clean the artifacts as well. I've just checked, that this Makefile always (clean/incremental build) copies the resources from JerryScript, extract them, and put them into the `libapps.a` library.

akosthekiss: Some random thoughts:
- Having libjerry-port-default.a and libjerry-port-default-minimal.a in LIBS is strange and suspicious.
- Extracting the objects from the libs is also somewhat strange. The build system is already working with libs. So why extract them when they will be re-packaged right after that?
- If untracked files are the problem, then .gitignore should be updated, not the build system.

So, I still think that this PR shouldn't mix `all` and `clean` targets. But it should update `clean` if it does not clean up things properly. And it may add items to top level gitignore (or add a subdir level gitifgnore) if there are build artefacts that should be hidden from git status.

rtakacs: The subdir-level gitignore file looks reasonable for me :)

akosthekiss:
Pinging this PR. Any update? It is both failing and disapproved.

pmarkee:
Sorry, I forgot about it. I'm going to update asap.

pmarkee:
I have updated the PR @akosthekiss @rtakacs , thanks for the ping.

PR #2636, opened 2018-12-11
Fixed connection lost of debugging sessions

LaszloLango:

### DIFF
--- a/jerry-ext/debugger/debugger-tcp.c
+++ b/jerry-ext/debugger/debugger-tcp.c
@@ -77,6 +77,7 @@ jerryx_debugger_tcp_send (jerry_debugger_transport_header_t *header_p, /**< tcp 
 
   do
   {
+#ifdef __linux__
     ssize_t is_err = recv (tcp_p->tcp_socket, NULL, 0, MSG_PEEK);
 
     if (is_err == 0 && errno != EWOULDBLOCK)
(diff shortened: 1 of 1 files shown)


jerry-ext/debugger/debugger-tcp.c
```diff
@@ -86,6 +87,8 @@ jerryx_debugger_tcp_send (jerry_debugger_transport_header_t *header_p, /**< tcp
       jerryx_debugger_tcp_log_error (err_val);
       return false;
     }
+#endif /* __linux__ */
```
akosthekiss: - Mind sharing why the lookahead (aka peek) causes connection loss on TizenRT?
- If this issue is TizenRT-specific, it might be better to guard it off with some TizenRT macro (since "TizenRT" != "not Linux").
- Why is the peek necessary on Linux, btw?

LaszloLango: This peek code was landed in #2427. It is only Linux specific fix for the issue of broken pipe errors. The commit message might be a bit misleading. This peek won't work on nuttx too (probably not even on Tizen). I'll update the commit message to be more precise.

akosthekiss: But what is Linux-specific in the approach? Does peeking not work elsewhere?

LaszloLango: Exactly, other platforms handles it differently. The trick is to read zero bytes from the socket and Linux systems will set the errno if the socket is open. With this peek code path TizenRT will close the connection if the client does not send data on the socket, but only listening, because it does not set the errno as we expect.

LaszloLango:
@akosthekiss I've updated the commit message.

jerry-ext/debugger/debugger-tcp.c
akosthekiss: After some offline discussion we have concluded that the now guarded-off code seems to be a Linux-specific hack. Actually, the hack caused problems on Darwin, so guarding it off not only helps embedded but macOS, too.

PR #2642, opened 2018-12-14
Removed ECMA_TRY_CATCH macros from RegExp built-in

LaszloLango:


jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -63,156 +63,157 @@ ecma_builtin_regexp_prototype_compile (ecma_value_t this_arg, /**< this argument
...
-    ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("Incomplete RegExp type"));
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Incomplete RegExp type"));
   }
-  else
-  {
-    uint16_t flags = 0;
 
-    if (ecma_is_value_object (pattern_arg)
-        && ecma_object_class_is (ecma_get_object_from_value (pattern_arg), LIT_MAGIC_STRING_REGEXP_UL))
+  ecma_value_t ret_value = ECMA_VALUE_EMPTY;
```
akosthekiss: This declaration is way too far away from the first use. (And because the first access to this variable is actually an assignment, this initialisation is superfluous.)

LaszloLango:
@akosthekiss thanks for the review. I've updated the PR.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -63,156 +63,156 @@ ecma_builtin_regexp_prototype_compile (ecma_value_t this_arg, /**< this argument
...
+  ecma_string_t *pattern_string_p = NULL;
 
-        ret_value = ECMA_VALUE_UNDEFINED;
+  /* Get source string. */
+  ecma_value_t ret_value = ecma_regexp_read_pattern_str_helper (pattern_arg, &pattern_string_p);
+  if (ECMA_IS_VALUE_ERROR (ret_value))
+  {
+    JERRY_ASSERT (pattern_string_p == NULL);
+    return ret_value;
+  }
```
akosthekiss: I'm not completely sure but my gut feeling tells me that an assert would be good to have here. `ret_value` is not used later on anywhere. However, it is an `ecma_value_t`, which is usually something that has to be released unless it is known that is of type that needs no release. That is, if `ret_value` is not an error then it must be undefined(?)

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -84,54 +83,44 @@ ecma_builtin_regexp_dispatch_construct (const ecma_value_t *arguments_list_p, /*
...
+  }
 
-      flags_string_p = ecma_get_string_from_value (flags_str_value);
-      ecma_ref_ecma_string (flags_string_p);
+  ecma_string_t *pattern_string_p = NULL;
+  ecma_value_t ret_value = ecma_regexp_read_pattern_str_helper (pattern_value, &pattern_string_p);
+  if (ECMA_IS_VALUE_ERROR (ret_value))
+  {
+    return ret_value;
+  }
```
akosthekiss: As the value of `ret_value` is overwritten later on, an assert on `ret_value` being undefined (or empty?) should be used here, too.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -84,54 +83,44 @@ ecma_builtin_regexp_dispatch_construct (const ecma_value_t *arguments_list_p, /*
...
+    if (ECMA_IS_VALUE_ERROR (ret_value))
     {
       ecma_deref_ecma_string (pattern_string_p);
-    }
-
-    if (flags_string_p != NULL)
-    {
-      ecma_deref_ecma_string (flags_string_p);
+      return ret_value;
     }
```
akosthekiss: Same here.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
LaszloLango: It must be empty if not an error. I can add an assertion for that.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -264,22 +263,20 @@ ecma_op_create_regexp_object (ecma_string_t *pattern_p, /**< input pattern */
...
-
+  ecma_value_t ret_value = re_compile_bytecode (&bc_p, pattern_p, flags);
   if (ECMA_IS_VALUE_ERROR (ret_value))
   {
     ecma_deref_object (object_p);
+    return ret_value;
   }
 
-  return ret_value;
+  JERRY_ASSERT (ret_value == ECMA_VALUE_EMPTY);
```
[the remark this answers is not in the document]
LaszloLango: Good point. I fixed it.

PR #2643, opened 2018-12-14
Fix possible failure in backtrace info

DanielBallaSZTE:
Make sure to ignore static snapshots when sending backtrace information to the debugger.
The `JMEM_SET_NON_NULL_POINTER` macro requires the frame context's `bytecode_header_p` pointer to be a heap pointer, but due to the static snapshot it might not be, causing an assertion failure.

This PR is identical to #2608 which I sadly couldn't reopen.
Test environment is not added yet, however this one might be a blocker as well since our VSCode extension uses backtrace and total frame count requests.


DanielBallaSZTE:
CC @akosthekiss

PR #2644, opened 2018-12-14
Fix MbedOS build

akosthekiss:
Since recently, the newly released mbed-host-tests 1.4.4 gets
installed during the python requirement installation step. That
package requires pyocd>=0.14.0, which cannot be satisfied for
whatever reason. This quick fix pins mbed-host-tests to version
1.4.2, which does not have unsatisfiable requirements.


galpeter:
lgtm

PR #2646, opened 2018-12-18
Reduce the memory footprint of 'ecma_instantiate_builtin'

rerobika:
This patch introduces a new builtin descriptor table to substitute the generated switch-case structure in `ecma_instantiate_builtin`, also removes the corresponding helpers functions to simplify the instantiation process.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.c
@@ -52,6 +52,99 @@ typedef ecma_value_t (*ecma_builtin_dispatch_routine_t)(uint16_t builtin_routine
 typedef ecma_value_t (*ecma_builtin_dispatch_call_t)(const ecma_value_t arguments_list[],
                                                      ecma_length_t arguments_number);
 /**
+ * Definition of a builtin descriptor which contains the builtin object's:
+ * - prototype objects's id (13-bits)
+ * - type (3-bits)
+ *
+ * Layout:
+ *
+ * |----------------------|---------------|
+ *     prototype_id(13)      obj_type(3)
+ */
+typedef uint16_t ecma_builtin_descriptor_t;
+
+/**
+ * Bitshift index for get the prototype object's id from a builtin descriptor
+ */
+#define ECMA_BUILTIN_PROTOTYPE_ID_SHIFT 3
+
+/**
+ * Bitmask for get the object's type from a builtin descriptor
+ */
+#define ECMA_BUILTIN_OBJECT_TYPE_MASK ((1 << ECMA_BUILTIN_PROTOTYPE_ID_SHIFT) - 1)
+
+/**
+ * Create a builtin descriptor value
+ */
+#define ECMA_MAKE_BUILTIN_DESCRIPTOR(type, proto_id) \
+  (((proto_id) << ECMA_BUILTIN_PROTOTYPE_ID_SHIFT) | (type))
+
+/**
+ * List of the built-in descriptors.
+ */
+static const ecma_builtin_descriptor_t ecma_builtin_descriptors[] =
+{
+/** @cond doxygen_suppress */
+#define BUILTIN(a, b, c, d, e)
+#define BUILTIN_ROUTINE(builtin_id, \
+                        object_type, \
+                        object_prototype_builtin_id, \
+                        is_extensible, \
+                        lowercase_name) \
+  ECMA_MAKE_BUILTIN_DESCRIPTOR (object_type, object_prototype_builtin_id),
+#include "ecma-builtins.inc.h"
+#undef BUILTIN
+#undef BUILTIN_ROUTINE
+#define BUILTIN_ROUTINE(a, b, c, d, e)
+#define BUILTIN(builtin_id, \
+                object_type, \
+                object_prototype_builtin_id, \
+                is_extensible, \
+                lowercase_name) \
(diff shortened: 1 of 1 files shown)


rerobika:
Perf/binary size changes:

Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 0.779 -> 0.778 : +0.145% | 
3d-raytrace.js | 1.013 -> 1.004 : +0.898% | 
access-binary-trees.js | 0.556 -> 0.538 : +3.298% | 
access-fannkuch.js | 2.020 -> 2.007 : +0.603% | 
access-nbody.js | 1.061 -> 1.057 : +0.376% | 
bitops-3bit-bits-in-byte.js | 0.483 -> 0.482 : +0.162% | 
bitops-bits-in-byte.js | 0.642 -> 0.643 : -0.103% | 
bitops-bitwise-and.js | 0.906 -> 0.906 : -0.034% | 
bitops-nsieve-bits.js | 1.117 -> 1.113 : +0.361% | 
controlflow-recursive.js | 0.350 -> 0.357 : -2.073% | 
crypto-aes.js | 0.838 -> 0.831 : +0.823% | 
crypto-md5.js | 0.578 -> 0.580 : -0.180% | 
crypto-sha1.js | 0.566 -> 0.567 : -0.122% | 
date-format-tofte.js | 0.736 -> 0.734 : +0.277% | 
date-format-xparb.js | 0.507 -> 0.507 : -0.033% | 
math-cordic.js | 1.141 -> 1.144 : -0.341% | 
math-partial-sums.js | 0.716 -> 0.712 : +0.526% | 
math-spectral-norm.js | 0.524 -> 0.523 : +0.244% | 
string-base64.js | 1.384 -> 1.360 : +1.716% | 
string-fasta.js | 1.312 -> 1.316 : -0.305% | 
Geometric mean: | +0.317% | 

Binary (bytes) | a1595fa23a | 1f55392f1a | Gain |
.text | 116532 | 116156 | +0.323% |
.rodata | 11364 | 11442 | -0.686% |

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -51,6 +51,74 @@ typedef ecma_value_t (*ecma_builtin_dispatch_routine_t)(uint16_t builtin_routine
...
+/**
+ * Definition of a builtin descriptor which contains the builtin object's:
+ * - prototype objects's id (8-bits)
+ * - type (3-bits)
+ * - [[Extensible]] property's value (1-bit)
+ *
+ * Layout:
+ *
+ * |---------------------|-------------|---------------|--------------------|
+ *     prototype_id(8)      unused(4)     obj_type(3)     is_extensible(1)
```
zherczeg: I would put the obj_type to last, and the flag before it. Are you sure 8 bit for prototype id is enough? I would use the 4 unused bit for it as well, and if we need more bits we can figure something. Btw is there an object which is not extensible by default?

rerobika: OK, let's swap them.

I think 8 bit is more than enough for the prototype id. Currently we have 60 builtin objects (with es2015 profile) and only 7 of these used as a prototype.

Yes, the [ [[ThrowTypeError]] ](https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/ecma/builtin-objects/ecma-builtins.inc.h#L265) object is not extensible.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -51,6 +57,74 @@ typedef ecma_value_t (*ecma_builtin_dispatch_routine_t)(uint16_t builtin_routine
...
+#define BUILTIN(builtin_id, \
+                object_type, \
+                object_prototype_builtin_id, \
+                is_extensible, \
+                lowercase_name) \
+  ECMA_MAKE_BUILTIN_DESCRIPTOR (object_type, object_prototype_builtin_id, is_extensible),
+#include "ecma-builtins.inc.h"
+#undef BUILTIN
+#undef BUILTIN_ROUTINE
+/** @endcond */
```
zherczeg:
```
#ifndef NDEBUG

#define BUILTIN_ROUTINE() \\\
  (extensible != 0 || id = ERROR) &&

enum (
  assert_value = (
#include "ecma-builtins.inc.h"
  1);
)
STATIC_ASSERT(assert_value == 1);

#endif
```

rerobika:
@zherczeg The patch is updated.
@LaszloLango can take a look at it?

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -51,6 +51,99 @@ typedef ecma_value_t (*ecma_builtin_dispatch_routine_t)(uint16_t builtin_routine
  */
 typedef ecma_value_t (*ecma_builtin_dispatch_call_t)(const ecma_value_t arguments_list[],
                                                      ecma_length_t arguments_number);
+/**
+ * Definition of a builtin descriptor which contains the builtin object's:
+ * - prototype objects's id (12-bits)
+ * - type (3-bits)
```
LaszloLango: These are only 15 bits. Is the 16th bit unused?

rerobika: Oops! Good catch! All the remaining bits(13) are used for the prototype objects's id.

PR #2674, opened 2019-01-04
Increase test coverage: Array.prototype.sort

repasics:
Added new test cases to array-prototype-sort.js to improve branch coverage.
Branch coverage:
sort:
     - before: 44/50
     - after:  48/48

sort compare helper:
     - before: 18/24
     - after:  24/24

Also removed an unnecessary condition from the while loop what counts properties
with lower index than len.

### DIFF
--- a/tests/jerry/array-prototype-sort.js
+++ b/tests/jerry/array-prototype-sort.js
@@ -92,3 +92,123 @@ try {
   assert(e.message === "foo");
   assert(e instanceof ReferenceError);
 }
+
+// Checking behavior when this value is undefined
+var obj = { sort : Array.prototype.sort };
+
+try {
+  obj.sort.call(undefined, function () { });
+  assert(false);
+} catch (e) {
+  assert(e instanceof TypeError);
+}
+
+// Checking behavior when length's valueOf throws exception
+var len = { };
+Object.defineProperty(len, 'valueOf', { 'get' : function () { throw new ReferenceError ("foo"); } });
+var obj = { sort : Array.prototype.sort, length : len };
+
+try {
+  obj.sort();
+  assert(false);
+} catch (e) {
+  assert(e.message === 'foo');
+  assert(e instanceof ReferenceError);
+}
+
+// Checking behavior when unable to get elements
+var obj = { sort : Array.prototype.sort, length : 2};
+Object.defineProperty(obj, '0', { 'get' : function () { throw new ReferenceError ("foo"); } });
+Object.defineProperty(obj, '1', { 'get' : function () { throw new ReferenceError ("bar"); } });
+
+try {
+  obj.sort();
+  assert(false);
+} catch (e) {
+  assert(e.message === "foo");
+  assert(e instanceof ReferenceError);
+}
+
+// Checking behavior when array is non-extensible while sorting
+var arr = [1, 0];
+
+try {
+  arr.sort(function () { Object.freeze(arr) });
+  assert(false);
+} catch (e) {
+  assert(e instanceof TypeError);
+}
+
+// Checking behavior when unable to delete property
+var obj = {sort : Array.prototype.sort, '0' : 2, '1' : 1, length : 4};
+Object.defineProperty(obj, '3', function () {});
+
+try {
+  obj.sort();
+  assert(false);
+}
+catch (e) {
+  assert(e instanceof TypeError);
+}
+
+// Checking behavior when unable to get the last element
+var arr = [1, 2, ];
+Object.defineProperty(arr, '2', { 'get' : function () { throw new ReferenceError ("foo"); } });
+
+try {
+  arr.sort();
+  assert(false);
+}
+catch (e) {
+  assert(e.message === 'foo');
+  assert(e instanceof ReferenceError);
+}
+
+// Checking behavior when lhs_value throws exception at comparefn
+f = function () { throw new ReferenceError('foo'); };
+obj = { 'toString' : f };
+arr = [obj, 1];
+
+try {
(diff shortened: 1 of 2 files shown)


akosthekiss:
@repasics: How did you measure branch coverage? Which tool did you use?

matedabis:
@akosthekiss 
We used LCOV to measure the coverage.

tests/jerry/array-prototype-sort.js
```diff
@@ -92,3 +92,109 @@ try {
...
+}
+
+// Checking behavior when the element's toString throws error
+var a = { };
+Object.defineProperty(a, 'toString', { 'get' : function () { throw new ReferenceError ("foo"); } });
+var arr = [1, a];
+
+try {
+arr.sort();
+assert(false);
```
rerobika: These two lines are misaligned.

akosthekiss:
@repasics @matedabis Could you please elaborate on the whole measurement setup a bit more? How did you build jerryscript, how did you invoke lcov, which test suite(s) did you execute?

matedabis:
To set up LCOV I used the help of the following page:
http://ltp.sourceforge.net/coverage/lcov.php
There is also an example for how the result looks like.

We used this script I made to measure coverage:

JerryScript build command:
https://github.com/matedabis/jerryscript/blob/6a2b9cc3ce2e6621b133fc5003309ad160cc0a0e/tests/gcov-tests/gcovtester.py#L47

Coverage information capture command:
https://github.com/matedabis/jerryscript/blob/6a2b9cc3ce2e6621b133fc5003309ad160cc0a0e/tests/gcov-tests/gcovtester.py#L75

Generate HTML command:
https://github.com/matedabis/jerryscript/blob/6a2b9cc3ce2e6621b133fc5003309ad160cc0a0e/tests/gcov-tests/gcovtester.py#L88

It is also important, that in `ECMA_TRY_CATCH` and `ECMA_OP_TO_NUMBER_TRY_CATCH` we had to disable the following line:
`JERRY_ASSERT (return_value == ECMA_VALUE_EMPTY);`
Because of more than one test cases, if it is enabled, we may can not reach some branches.

akosthekiss:
@matedabis Thanks for the link. The script and your note raise some further questions:
- The script runs only a single test file in the engine (via the command line shell). Coverage (be that line or branch coverage) is usually measured by running all tests in a test suite or by running all available test suites. So, what are the (branch) coverage results when all test suites are executed?
- What is 48? Total number of branches of a function? Of a file? Which function? Which file? The script you linked or the commit message above don't answer that.
- The removal of an assert is a red flag to me. What branches can only be reached if an assert is removed? Asserts should not make any difference (theoretically, at least).

matedabis:
@akosthekiss 

1. I don't know that yet, we only measured the coverage of functions each with its test file in the `/tests/jerry` folder. For example in this case we test `ecma_builtin_array_prototype_object_sort` function only with `/tests/jerry/array-prototype-sort.js`. But if it is better, we will test the coverage by running all the test suites.

2. The total number is the number of branches of the function we are actually testing, in this case, the mentioned above, excluding `JERRY_ASSERT`s.

3. The reason we do not test `JERRY_ASSERT`s and we removed 1-1 of them in `ECMA_TRY_CATCH` and `ECMA_OP_TO_NUMBER_TRY_CATCH` is that, there are many test cases in a file, and if we trigger an assert, the engine stops running the js code at that point, so the test cases below are not executed. This way if we trigger one of the try-catches, the code does not stop running, we can reach those branches without disabling the asserts, but we would have this issue. This way I think we have two options. Either splitting the test file to many small parts to avoid this problem, or do not trigger asserts, and disable asserts in try-catches.

akosthekiss:
1) I'm a bit puzzled as I recall we've had similar discussions before when the test suite was planned to get extended with additional code. Then, I've pointed at both the jerry-test-suite and at test262. Picking out single test files from a suite can be quite misleading about the coverage. At the least, /tests/jerry-test-suite and /tests/jerry should be executed before drawing any conclusions, as they are project internal suites. But I'd be interested about combining them with test262, too, since that's a submodule, too.

3) I cannot get my head around two things. First, in a well-behaving engine, asserts must not happen in the executed test files. No matter how many test cases are written in a single .js file, they will all run. I they didn't execute properly, Travis CI were read. Second, what's the connection between asserts in .js test files and asserts in C macros? I cannot see how they affect each other.

matedabis:
Sorry I realized I misunderstood a thing, I did not want to mislead you. `JERRY_ASSERT`s are only disabled to see the real branch coverage, without the asserts. Of course we can reach all branches with asserts enabled, except 1 branch of each asserts. It is only for better visibility.

matedabis:
I will create a modified version of the coverage tester script and test the coverage with the test suites.

matedabis:
That will help for sure, thank you.

repasics:
I checked the coverage including jerry-test-suite and test262 tests, then removed the unnecessary test cases.
The new stat is:
     - before: 44/50
     - after: 48/48

repasics:
I added new test cases what covers the sort_compare_helper function too.

repasics:
Summary:
The following script is made for testing branch coverage with all the test suites (--jerry-test-suite --test262 --unittests --jerry-tests), or with only one .js file.

While measuring the branch coverage we don't count JERRY_ASSERT s.

Branch coverage (covered branch/all branch):
   sort function:
	- before: 44/50
	- after: 48/48

   sort compare helper:
	- before: 18/24
	- after: 24/24

repasics:
Closed PR because of moving into a unified PR.

PR #2683, opened 2019-01-08
Travis CI config maintenance

akosthekiss:
- `sudo: false` has been deprecated.
- `install` step is skipped by default for C projects, so there is
  no need setting it explicitly.
- Homebrew addon can be used to install dependencies for OSX jobs
  (and, by experience, it is a lot faster than manually invoking
  brew).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,8 @@
 language: c
 
-# Default environment: Container-based (sudo-less) Ubuntu Trusty 14.04.
+# Default environment: Ubuntu Trusty 14.04.
 os: linux
 dist: trusty
-sudo: false
-
-# Default dependency installation step: nop intentionally.
-# Jobs can add their own dependencies by redefinig the 'install' stage in the matrix below.
-install: skip
 
 # Default job task: run tests as defined in the $OPT environment variable.
 # Jobs can redefine the 'script' stage in the matrix below.
@@ -48,7 +43,9 @@ matrix:
       env:
         - OPTS="--quiet --jerry-tests --jerry-test-suite --unittests"
       os: osx
-      install: tools/brew-install-deps.sh
+      addons:
+        homebrew:
+          packages: [cmake, cppcheck, vera++]
 
     - name: "Build Tests"
       env:


yichoi:
I am sorry, I did not tell you precisely.
Could you please change ``jerryscript-project`` to ``pando-project``  ?
```
114       addons:
115         sonarcloud:
116           organization: "jerryscript-project"
117           token:
```

akosthekiss:
Could we break the fixup to SonarCloud (and perhaps also to Coverity Scan) out into a separate PR? The reason behind that is that I think the project is not ready for the suggested change:
- In `.travis.yml`, the `addons.sonarcloud.organization` key has to be set to `pando-project`, true. But it seems that on sonarcloud.io, the [`pando-project`](https://sonarcloud.io/organizations/pando-project/projects) organization doesn't have  a jerryscript project. At least, when I search for jerryscript, the website only finds jerryscript under the `jerryscript-project` org (+ under two personal orgs).
- When a user with appropriate privileges creates a project under the `pando-project` org, the project key will most likely become `pando-project_jerryscript`, so `sonar.projectKey` in `sonar-project.properties` will have to be updated, too.
- (The `sonar.organization` key can most probably be removed from `sonar-project.properties`. The addon in the Travis CI conf takes care of this now, if I'm not mistaken. The [documentation](https://docs.sonarqube.org/display/SCAN/Analyzing+with+SonarQube+Scanner) doesn't mention this key anymore. Plus, IoT.js doesn't use this key either. On the other hand, should it still be needed, then the name is outdated.)
- Furthermore, the sonar-related links in `README.md` should also be updated, once they are alive.

So, it seems to me that this change should be done by a user who can set up the project under pando-project in sonarcloud (or, at least, once that user has set up the project), and in a separate step that touches all sonar-related settings/files.

Note that the project probably also needs both setup activities and config maintenance related to the Coverity Scan service. However, Coverity Scan has been [completely down since Jan 7](https://community.synopsys.com/s/article/Coverity-Scan-Update):
> our Coverity Scan hosting service provider unexpectedly ceased operations ... Beginning immediately, Coverity Scan will be in extended maintenance mode and unreachable by users ... all open projects are frozen and will be unreachable ... Github integration will be unavailable during this extended maintenance period

So, right now, nothing can be done for that service.

yichoi:
@akosthekiss I am fine with your idea

akosthekiss:
@yichoi Thanks. Could you please confirm the new project key on sonarcloud? (I guess you've created the project under the pando-project org key?)

yichoi:
``sonar.projectKey=pando-project_jerryscript``

akosthekiss:
SonarCloud service maintenance broken out to follow-up PR #2690

akosthekiss:
As the SonarCloud service maintenance is now in another PR, could I ask for a review on this one, as this is not pending on any external circumstances?

PR #2684, opened 2019-01-08
Update the debugger to work on Windows

galpeter:
Changed the debugger-tcp.c file to include Windows specific
socket handling code.

Added appveyor configuration to build with and without debugger.

### DIFF
--- a/jerry-ext/CMakeLists.txt
+++ b/jerry-ext/CMakeLists.txt
@@ -47,5 +47,9 @@ target_include_directories(${JERRY_EXT_NAME} PRIVATE ${INCLUDE_EXT_PRIVATE})
 target_compile_definitions(${JERRY_EXT_NAME} PUBLIC ${DEFINES_EXT})
 target_link_libraries(${JERRY_EXT_NAME} jerry-core)
 
+if(USING_MSVC AND FEATURE_DEBUGGER)
+  target_link_libraries(${JERRY_EXT_NAME} ws2_32)
+endif()
+
 install(TARGETS ${JERRY_EXT_NAME} DESTINATION lib)
 install(DIRECTORY ${INCLUDE_EXT_PUBLIC}/ DESTINATION include)
--- a/jerry-ext/debugger/debugger-tcp.c
+++ b/jerry-ext/debugger/debugger-tcp.c
@@ -19,11 +19,39 @@
 
 #ifdef JERRY_DEBUGGER
 
-#include <arpa/inet.h>
 #include <errno.h>
+
+#ifdef WIN32
+#include <BaseTsd.h>
+typedef SSIZE_T ssize_t;
+#include <WS2tcpip.h>
+#include <winsock2.h>
+
+/* On Windows the WSAEWOULDBLOCK value can be returned for non-blocking operations */
+#define JERRYX_EWOULDBLOCK WSAEWOULDBLOCK
+
+/* On Windows the invalid socket's value of INVALID_SOCKET */
+#define JERRYX_SOCKET_INVALID INVALID_SOCKET
+
+/* On Windows sockets have a SOCKET typedef */
+typedef SOCKET jerryx_socket;
+
+#else /* !WIN32 */
+
+#include <arpa/inet.h>
 #include <fcntl.h>
-#include <unistd.h>
 #include <sys/socket.h>
+#include <unistd.h>
+
+/* On *nix the EWOULDBLOCK errno value can be returned for non-blocking operations */
+#define JERRYX_EWOULDBLOCK EWOULDBLOCK
+
+/* On *nix the invalid socket has a value of -1 */
+#define JERRYX_SOCKET_INVALID (-1)
+
+/* On *nix the sockets are integer identifiers */
+typedef int jerryx_socket;
+#endif /* WIN32 */
 
 /**
  * Implementation of transport over tcp/ip.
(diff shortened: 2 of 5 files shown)


galpeter:
I've updated the PR.

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -19,28 +19,92 @@
...
+  FormatMessage (FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                 NULL,
+                 errno_value,
+                 MAKELANGID (LANG_NEUTRAL, SUBLANG_DEFAULT),
+                 (LPTSTR) &error_message,
+                 0,
+                 NULL);
+  jerry_port_log (JERRY_LOG_LEVEL_ERROR, "%s: %s\n", prefix, error_message);
+  LocalFree (error_message);
+#else
```
rerobika: The `/* !WIN32 */` comment is missing.

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -19,28 +19,92 @@
...
+  jerry_port_log (JERRY_LOG_LEVEL_ERROR, "%s: %s\n", prefix, strerror (errno_value));
+#endif
```
rerobika: The `/* WIN32 */` comment is missing. Could you please add these comments in the all related places below?

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -19,28 +19,92 @@
...
+
+#include <arpa/inet.h>
 #include <fcntl.h>
-#include <unistd.h>
 #include <sys/socket.h>
+#include <unistd.h>
+
+#define JERRYX_EWOULDBLOCK EWOULDBLOCK
+#define JERRYX_SOCKET_INVALID (-1)
+typedef int jerryx_socket;
```
rerobika: Could you add descriptions for all the defines and typedefs above?

galpeter:

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -80,11 +154,11 @@ jerryx_debugger_tcp_send (jerry_debugger_transport_header_t *header_p, /**< tcp
 #ifdef __linux__
     ssize_t is_err = recv (tcp_p->tcp_socket, NULL, 0, MSG_PEEK);
 
-    if (is_err == 0 && errno != EWOULDBLOCK)
+    if (is_err == 0 && errno != JERRYX_EWOULDBLOCK)
```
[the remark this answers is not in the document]
galpeter: any suggestions? AFAIK this is the most descriptive as bot the win and non-win version uses the EWOULDBLOCK suffix

galpeter:
@zherczeg the biggest change is the creation of the `jerryx_debugger_tcp_configure_socket` method which was extracted from the `jerryx_debugger_tcp_create` method and this new method is the same on all platform. Almost all of the original methods only have one or two minor modifications (aside from the previously mentioned one).

jerry-ext/CMakeLists.txt
```diff
@@ -47,5 +47,9 @@ target_include_directories(${JERRY_EXT_NAME} PRIVATE ${INCLUDE_EXT_PRIVATE})
 target_compile_definitions(${JERRY_EXT_NAME} PUBLIC ${DEFINES_EXT})
 target_link_libraries(${JERRY_EXT_NAME} jerry-core)
 
+if(USING_MSVC AND FEATURE_DEBUGGER)
+  target_link_libraries(${JERRY_EXT_NAME} ws2_32.lib)
```
akosthekiss: Let' s state it upfront, I'm not completely familiar with cmake on windows (although I had to play with it a bit): but is the `.lib` suffix/extension needed for `target_link_libraries`? Isn't that function working with "library names" and appends/prepends whatever is mandated by the platform/os? (E.g., name "xxx" gets converted to `libxxx.a`, `libxxx.so`, `xxx.lib`, `xxx.dll`, etc. as needed?)

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -19,28 +19,102 @@
...
+  close (socket_id);
+#endif /* WIN32 */
+} /* jerryx_debugger_tcp_close_socket */
+
 /**
  * Log tcp error message.
  */
 static void
-jerryx_debugger_tcp_log_error (int err_val)
+jerryx_debugger_tcp_log_error (const char *prefix, /**< message prefix */
```
akosthekiss: minor: is the `prefix` really necessary? Are not all errors TCP-related so that it shouldn't be hardcoded? This is the TCP transport implementation after all.

jerry-ext/debugger/debugger-tcp.c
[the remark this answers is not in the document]
akosthekiss: I second `JERRYX_EWOULDBLOCK`

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -150,64 +226,105 @@ jerryx_debugger_tcp_receive (jerry_debugger_transport_header_t *header_p, /**< t
...
+  int opt_value = 1;
+
+  if (setsockopt (server_socket, SOL_SOCKET, SO_REUSEADDR, &opt_value, sizeof (int)) != 0)
   {
-    JERRYX_ERROR_MSG ("Error: %s\n", strerror (errno));
     return false;
   }
 
-  int opt_value = 1;
+  if (bind (server_socket, (struct sockaddr *)&addr, sizeof (struct sockaddr)) != 0)
```
akosthekiss: According to the docs of `bind` the 3rd arg should be the size of the 2nd arg. But `addr` is of type `struct sockaddr_in`, not `struct sockaddr`.

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -150,64 +226,105 @@ jerryx_debugger_tcp_receive (jerry_debugger_transport_header_t *header_p, /**< t
...
 
-  close (server_socket);
+  jerryx_debugger_tcp_close_socket (server_socket);
 
-  if (tcp_socket == -1)
+  if (tcp_socket == JERRYX_SOCKET_INVALID)
   {
-    JERRYX_ERROR_MSG ("Error: %s\n", strerror (errno));
+    int error = jerryx_debugger_tcp_get_errno ();
+    jerryx_debugger_tcp_log_error ("Error", error);
```
akosthekiss: These two lines can be merged. There is no need for a temp var.

jerry-ext/debugger/debugger-tcp.c
galpeter: Ahh yeah. Currently it works as the `sockaddr_in` can be considered as a "subclass" of the `sockaddr` and they have the same size (at least on linux). But will change it to `sizeof (struct sockaddr_in)`

jerry-ext/debugger/debugger-tcp.c
galpeter: Initially I wanted to use the simple `TCP error` hard-coded prefix, but in this variant I've followed the original `Error/TCP Error` prefixes. I'm happy to change this to the "TCP Error" prefix.

akosthekiss: I see. Well, then, it's no big deal. I vote for a single prefix but whatever majority votes for should prevail.

jerry-ext/CMakeLists.txt
galpeter: True, there is no need for the `.lib` extension.

PR #2687, opened 2019-01-09
Normalize routine's arguments before dispatch

rerobika:
Binary size gain:
 - Intel: ~860B (gcc-7.3)
 - Arm: ~660B (arm-linux-gnueabi-gcc-7.3)

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
@@ -579,13 +579,10 @@ ecma_builtin_date_prototype_dispatch_routine (uint16_t builtin_routine_id, /**< 
 
   if (builtin_routine_id == ECMA_DATE_PROTOTYPE_SET_TIME)
   {
-    ecma_value_t time = (arguments_number >= 1 ? arguments_list[0]
-                                               : ECMA_VALUE_UNDEFINED);
-
     ecma_value_t ret_value = ECMA_VALUE_EMPTY;
 
     /* 1. */
-    ECMA_OP_TO_NUMBER_TRY_CATCH (time_num, time, ret_value);
+    ECMA_OP_TO_NUMBER_TRY_CATCH (time_num, arguments_list[0], ret_value);
     *prim_value_p = ecma_date_time_clip (time_num);
 
     ret_value = ecma_make_number_value (time_num);
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.c
@@ -949,6 +949,32 @@ ecma_builtin_dispatch_routine (ecma_builtin_id_t builtin_object_id, /**< built-i
                                ecma_length_t arguments_list_len) /**< length of arguments' list */
 {
   JERRY_ASSERT (builtin_object_id < ECMA_BUILTIN_ID__COUNT);
+
+  ecma_value_t padded_arguments_list_p[3] = { ECMA_VALUE_UNDEFINED, ECMA_VALUE_UNDEFINED, ECMA_VALUE_UNDEFINED };
+
+  if (arguments_list_len <= 2)
+  {
+    switch (arguments_list_len)
+    {
+      case 2:
+      {
+        padded_arguments_list_p[1] = arguments_list_p[1];
+        /* FALLTHRU */
+      }
+      case 1:
+      {
+        padded_arguments_list_p[0] = arguments_list_p[0];
+        break;
+      }
+      default:
+      {
+        JERRY_ASSERT (arguments_list_len == 0);
+      }
+    }
+
+    arguments_list_p = padded_arguments_list_p;
+  }
+
   return ecma_builtin_routines[builtin_object_id] (builtin_routine_id,
                                                    this_arg_value,
                                                    arguments_list_p,
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-internal-routines-template.inc.h
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-internal-routines-template.inc.h
@@ -172,8 +172,7 @@ DISPATCH_ROUTINE_ROUTINE_NAME (uint16_t builtin_routine_id, /**< built-in wide r
 
   switch (builtin_routine_id)
   {
-#define ROUTINE_ARG(n) (arguments_number >= n ? arguments_list[n - 1] \
-                                              : ECMA_VALUE_UNDEFINED)
+#define ROUTINE_ARG(n) (arguments_list[n - 1])
 #define ROUTINE_ARG_LIST_0
 #define ROUTINE_ARG_LIST_1 , ROUTINE_ARG(1)
 #define ROUTINE_ARG_LIST_2 ROUTINE_ARG_LIST_1, ROUTINE_ARG(2)


rerobika:
Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 0.772 -> 0.775 : -0.413% | 
3d-raytrace.js | 1.001 -> 0.998 : +0.266% | 
access-binary-trees.js | 0.542 -> 0.543 : -0.068% | 
access-fannkuch.js | 1.983 -> 1.986 : -0.147% | 
access-nbody.js | 1.048 -> 1.046 : +0.249% | 
bitops-3bit-bits-in-byte.js | 0.478 -> 0.479 : -0.123% | 
bitops-bits-in-byte.js | 0.638 -> 0.638 : -0.001% | 
bitops-bitwise-and.js | 0.903 -> 0.911 : -0.939% | 
bitops-nsieve-bits.js | 1.110 -> 1.118 : -0.677% | 
controlflow-recursive.js | 0.363 -> 0.357 : +1.603% | 
crypto-aes.js | 0.830 -> 0.830 : +0.014% | 
crypto-md5.js | 0.581 -> 0.577 : +0.691% | 
crypto-sha1.js | 0.572 -> 0.565 : +1.260% | 
date-format-tofte.js | 0.732 -> 0.729 : +0.423% | 
date-format-xparb.js | 0.514 -> 0.518 : -0.692% | 
math-cordic.js | 1.142 -> 1.135 : +0.643% | 
math-partial-sums.js | 0.729 -> 0.722 : +1.001% | 
math-spectral-norm.js | 0.515 -> 0.512 : +0.639% | 
string-base64.js | 1.398 -> 1.422 : -1.709% | 
string-fasta.js | 1.332 -> 1.328 : +0.278% | 
Geometric mean: | +0.118% | 

The performance change is mostly noise due to the code layout change.

Binary (bytes) | 3ee771655f | 01646943d8 | Gain |
.text | 116116 | 115444 | +0.579% |
.rodata | 11446 | 11446 | 0.000% |

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
```diff
@@ -551,6 +551,9 @@ ecma_value_t
 ecma_builtin_date_prototype_dispatch_routine (uint16_t builtin_routine_id, /**< built-in wide routine
                                                                             *   identifier */
                                               ecma_value_t this_arg, /**< 'this' argument value */
+                                              const ecma_value_t normalized_arguments_list[3], /**< list of normalized
```
zherczeg: I think we can do this without introducing a new argument. Just reuse arguments_list.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -949,8 +950,17 @@ ecma_builtin_dispatch_routine (ecma_builtin_id_t builtin_object_id, /**< built-i
...
 {
   JERRY_ASSERT (builtin_object_id < ECMA_BUILTIN_ID__COUNT);
+
+  const ecma_value_t normalized_argumets_list_p[3] =
+  {
+    arguments_list_len >= 1 ? arguments_list_p[0] : ECMA_VALUE_UNDEFINED,
+    arguments_list_len >= 2 ? arguments_list_p[1] : ECMA_VALUE_UNDEFINED,
+    arguments_list_len >= 3 ? arguments_list_p[2] : ECMA_VALUE_UNDEFINED
+  };
+
```
akosthekiss: Wouldn't it be possible to use
```c
if (arguments_list_len <= 2)
{
  ecma_value_t padded_argumets_list_p[3] = { ECMA_VALUE_UNDEFINED, ECMA_VALUE_UNDEFINED, ECMA_VALUE_UNDEFINED };

  switch (arguments_list_len)
  {
    case 2: padded_argumets_list_p[1] = arguments_list_p[1];
    case 1: padded_argumets_list_p[0] = arguments_list_p[0];
    default: JERRY_ASSERT (arguments_list_len == 0);
  }

  return ecma_builtin_routines[builtin_object_id] (builtin_routine_id,
                                                   this_arg_value,
                                                   padded_arguments_list_p,
                                                   arguments_list_len);
}

return ecma_builtin_routines[builtin_object_id] (builtin_routine_id,
                                                 this_arg_value,
                                                 arguments_list_p,
                                                 arguments_list_len);
```
This way there would be no need for an additional parameter to `ecma_builtin_dispatch_routine_t` functions but they could always be sure that `arguments_list` had at least 3 elements, even if `arguments_number` was less than 3.

I think this is quite close to what @zherczeg suggested.

rerobika: Ohh! Thank you for the example code, thanks to it I see what you were thinking for.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -949,6 +949,40 @@ ecma_builtin_dispatch_routine (ecma_builtin_id_t builtin_object_id, /**< built-i
...
+      case 1:
+      {
+        padded_arguments_list_p[0] = arguments_list_p[0];
+        break;
+      }
+      default:
+      {
+        JERRY_ASSERT (arguments_list_len == 0);
+      }
+    }
```
zherczeg: `arguments_list_p = padded_arguments_list_p` isn't missing from here?

rerobika: I think is not missing. `padded_arguments_list_p` is a local array inside the `if (arguments_list_len <= 3)` condition.

zherczeg: But if you do that you don't need two `ecma_builtin_routines[builtin_object_id] (...)` calls, only one at the end of the function.

rerobika: Okay but for this case the `padded_arguments_list_p` must be in the function scope not in the `if ()`'s scope since it is a local array. I am not insist to the current solution, @akosthekiss what your opinion?

zherczeg: From technical perspective: it is always at func scope because messing with function stack allocation is a costly which compilers try to avoid. If you really want special casing this move to a never inline helper function, which can be called by a tail call, or create two functions.

rerobika: Okay, you convinced me!

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -949,6 +949,37 @@ ecma_builtin_dispatch_routine (ecma_builtin_id_t builtin_object_id, /**< built-i
...
+
+  if (arguments_list_len <= 3)
+  {
+    switch (arguments_list_len)
+    {
+      case 3:
+      {
+        padded_arguments_list_p[2] = arguments_list_p[2];
+        /* FALLTHRU */
+      }
```
akosthekiss: I think that this case is unnecessary and the outer `if` should only check for `arguments_list_len <= 2`. If `arguments_list_len == 3`, then there is no need for `padded_arguments_list_p` as `arguments_list_p` already has 3 elements.

rerobika: I cannot agree with that. We have routines with 3 arguments like `Object.defineProperty`.

```c
/**
 * The Object object's 'defineProperty' routine
 */
static ecma_value_t
ecma_builtin_object_object_define_property (ecma_value_t this_arg, /**< 'this' argument */
                                            ecma_value_t arg1, /**< routine's first argument */
                                            ecma_value_t arg2, /**< routine's second argument */
                                            ecma_value_t arg3) /**< routine's third argument */
```

These 3 arguments must came from the `padded_arguments_list_p` which arguments must be `undefined` or `arguments_list_p[i]` where `i>= 0 , i<3`

akosthekiss: Let's break down to cases:
- `arguments_list_len == 0`: we need `padded_arguments_list_p` and all its elements should be `ECMA_VALUE_UNDEFINED`. `arguments_list_p` will be replaced by `padded_arguments_list_p` and `ecma_builtin_routines[builtin_object_id]` will be called.
- `arguments_list_len == 1`: we need `padded_arguments_list_p`, its 0th element should be copied from `arguments_list_p` and the following two elements should be `ECMA_VALUE_UNDEFINED`. `arguments_list_p` will be replaced by `padded_arguments_list_p` and `ecma_builtin_routines[builtin_object_id]` will be called.
- `arguments_list_len == 2`: we need `padded_arguments_list_p`, its 0th and 1st elements should be copied from `arguments_list_p` and the last element should be `ECMA_VALUE_UNDEFINED`. `arguments_list_p` will be replaced by `padded_arguments_list_p` and `ecma_builtin_routines[builtin_object_id]` will be called.
- `arguments_list_len == 3`: we _do not_ need `padded_arguments_list_p` as `arguments_list_p` already has 3 elements so it can be used as is to call `ecma_builtin_routines[builtin_object_id]`.

rerobika: Indeed, the most trivial case tricked me a bit! :) Thanks for the clarification.

PR #2696, opened 2019-01-11
Bump NuttX RTOS target to 7.27

akosthekiss:
Both documentation and travis makefile were referring to NuttX 7.22,
which was released in Sep 2017. This commit bumps the version to the
latest release (7.27, from Nov 2018).

### DIFF
--- a/targets/nuttx-stm32f4/Makefile
+++ b/targets/nuttx-stm32f4/Makefile
@@ -54,7 +54,7 @@ updateobjs:
 cleanlibs: updateobjs
 	rm -f $(OBJS)
 
-clean: cleanlibs
+clean:: cleanlibs
 
 .built: $(LIBS) updateobjs
 
--- a/targets/nuttx-stm32f4/Makefile.travis
+++ b/targets/nuttx-stm32f4/Makefile.travis
@@ -39,8 +39,8 @@ install-kconfig:
 
 # Fetch nuttx/{apps,nuttx} repositories.
 install-clone-nuttx:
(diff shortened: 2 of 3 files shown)


targets/nuttx-stm32f4/Makefile
```diff
@@ -54,7 +54,7 @@ updateobjs:
 cleanlibs: updateobjs
 	rm -f $(OBJS)
 
-clean: cleanlibs
+clean:: cleanlibs
```
LaszloLango: Is it intentional? This looks accidental to me.

akosthekiss: No, this is intentional. NuttX [makefiles](https://bitbucket.org/nuttx/apps/src/b52aed887817a3a6374742e9d5d6298ebe300b08/Application.mk?at=nuttx-7.27&fileviewer=file-view-default#Application.mk-250) have changed and the `clean` target became a [double-colon rule](https://www.gnu.org/software/make/manual/make.html#Double_002dColon). From the docs:

> When a target appears in multiple rules, all the rules must be the same type: all ordinary, or all double-colon.

So, I had to update the `clean` rule here, too.

rtakacs:
LGTM (informally).  

I've tested this PR and it works fine until the `0c20f8e` JerryScript commit (Jan 8 - Fix broken links in README). After that commit (from `99b968d`), JerryScript doesn't work well neither with NuttX-7.27 nor with earlier NuttX versions. But this bug is not from this PR, the `99b968d` commit should be checked.

PR #2704, opened 2019-01-16
Fix %TypedArray% object instances symbol properties

rerobika:
Add missing check for during property handling for %TypedArray% object since until now only string property names were possible.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -141,6 +141,13 @@ ecma_op_object_get_own_property (ecma_object_t *object_p, /**< the object */
       /* ES2015 9.4.5.1 */
       if (ecma_is_typedarray (ecma_make_object_value (object_p)))
       {
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+        if (ecma_prop_name_is_symbol (property_name_p))
+        {
+          break;
+        }
+#endif /* !CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN */
+
         uint32_t array_index = ecma_string_get_array_index (property_name_p);
 
         if (array_index != ECMA_STRING_NOT_ARRAY_INDEX)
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -291,8 +292,15 @@ ecma_raise_standard_error_with_format (ecma_standard_error_t error_type, /**< er
       }
       else
       {
-        ecma_value_t str_val = ecma_op_to_string (arg_val);
-        arg_string_p = ecma_get_string_from_value (str_val);
+        arg_string_p = ecma_op_to_prop_name (arg_val);
```
galpeter: AFAIK this modification should be on it's own it is not related to the TypedArray.

tests/jerry/es2015/symbol.js
```diff
@@ -120,3 +120,32 @@ var obj = { foobar : 55 };
...
+
+try {
+  undefined[Symbol ("foo")];
+  assert (false);
+} catch (e) {
+  /* Error message: Cannot read property 'Symbol(foo)' of undefined */
+  assert (e instanceof TypeError);
+}
+
+var typedArrayTypes = [Int8Array,
```
galpeter: This test-case should be on it's own file as this requires the TypedArray and Symbol support.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -164,17 +164,24 @@ ecma_op_object_get_own_property (ecma_object_t *object_p, /**< the object */
           return ECMA_PROPERTY_TYPE_NOT_FOUND_AND_STOP;
         }
 
-        ecma_number_t num = ecma_string_to_number (property_name_p);
-        ecma_string_t *num_to_str = ecma_new_ecma_string_from_number (num);
-
-        if (ecma_compare_ecma_strings (property_name_p, num_to_str))
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+        if (!ecma_prop_name_is_symbol (property_name_p))
```
galpeter: So if I understand correctly this should not be done in case of a symbol property? Symbol properties are handled somewhere else?

Would it make sense to do this?
```
if (is_symbol(..)) { break } 
...
```
this way we would not need to put macros before and after the code.

rerobika: Yeah, good idea!

rerobika: Sorry, let me answer for your question. These related code path is only for faster the way of defining/getting properties. For symbol properties we should use the "general" way.

rerobika:

tests/jerry/es2015/typedarray-symbol-properties.js
```diff
@@ -0,0 +1,35 @@
...
+                       Uint8Array,
+                       Uint8ClampedArray,
+                       Int16Array,
+                       Uint16Array,
+                       Int32Array,
+                       Uint32Array,
+                       Float32Array,
+                       Float64Array];
+
+typedArrayTypes.forEach (function (e) {
```
akosthekiss: Strange name for the argument. I first thought it stood for error or something like that. How about t, ty, TypedArray, ...?

PR #2707, opened 2019-01-17
Use logical operators for bool converions.

rtakacs:
Replaced the newly introduced explicit boolean conversions to logical operators.

Reasons:
  * to be the boolean type conversion consistent in JerryScript
  * fix object-get-own-property-symbols test on NuttX and TizenRT


rerobika:
Side note for the object-get-own-property-symbols fail:

The original code assumed that after `const bool is_symbols_only = (const bool) (opts & ECMA_LIST_SYMBOLS);` initialization the value of `is_symbols_only` is `0/1` but in NuttX and TizenRT the  value of `is_symbols_only` is `ECMA_LIST_SYMBOLS(8)`. Due to that [this](https://github.com/pando-project/jerryscript/blob/master/jerry-core/ecma/operations/ecma-objects.c#L1524) comparison failed.

PR #2708, opened 2019-01-17
Substitute symbols with their descriptive string in ecma_raise_standard_error_with_format

rerobika:

### DIFF
--- a/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/jerry-core/ecma/operations/ecma-exceptions.c
@@ -21,6 +21,7 @@
 #include "ecma-globals.h"
 #include "ecma-helpers.h"
 #include "ecma-objects.h"
+#include "ecma-symbol-object.h"
 #include "jcontext.h"
 #include "jrt.h"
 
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -291,8 +292,15 @@ ecma_raise_standard_error_with_format (ecma_standard_error_t error_type, /**< er
...
-        ecma_value_t str_val = ecma_op_to_string (arg_val);
-        arg_string_p = ecma_get_string_from_value (str_val);
+        arg_string_p = ecma_op_to_prop_name (arg_val);
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+        if (ecma_prop_name_is_symbol (arg_string_p))
+        {
+          ecma_deref_ecma_string (arg_string_p);
+          ecma_value_t symbol_desc_value = ecma_get_symbol_descriptive_string (arg_val);
+          arg_string_p = ecma_get_string_from_value (symbol_desc_value);
+        }
```
akosthekiss: This approach looks suboptimal to me. `arg_string_p` is assigned first, then it is discarded if we have a symbol, and gets reassigned. It's possible to decide earlier based on `ecma_is_value_symbol (arg_val)` and assign `ecma_op_to_prop_name (arg_val)` to arg_string_p only if it is _not_ a symbol.

rerobika: Good catch! I've found a much easier solution for it.

PR #2715, opened 2019-01-18
[Proposal] Introduce new jerry-ext methods to ease method/property registration

galpeter:
New methods
* jerryx_set_functions: Allows multiple method registration for via a single call
* jerryx_set_property_str: Allows property registration without the need to
    create the property name JS value. The property name can be directly passed
    as a `const char*` value and must be zero terminated.

### DIFF
--- a/docs/10.EXT-REFERENCE-HANDLER.md
+++ b/docs/10.EXT-REFERENCE-HANDLER.md
@@ -1,3 +1,418 @@
+# Common methods to handle properties
+
+The `jerryscript-ext/handler.h` header defines a set of convenience methods
+which makes the property access a bit straightforward.
+
+## jerryx_set_property_str
+
+**Summary**
+
+Set a property on a target object with a given name.
+
+*Note*:
+- The property name must be a zero terminated UTF-8 string.
+- There should be no '\0' (NULL) character in the name excluding the string terminator.
+- Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
+  is no longer needed.
+
+
+**Prototype**
+
+```c
+jerry_value_t
+jerryx_set_property_str (const jerry_value_t target_object,
+                         const char *name,
+                         const jerry_value_t value);
+```
+
+- `target_object` - the object where the property should be set
+- `name` - name of the property
+- `value` - property value to be set
+- return value
+  - JS true value, if success
+  - thrown error, if there was a problem setting the property
+
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+#include "jerryscript-ext/handler.h"
+
+int
+main (int argc, char **argv)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t global = jerry_get_global_object ();
+
+  jerry_value_t value = jerry_create_number (3.3);
+  jerry_value_t result = jerryx_set_property_str (global, "value", value);
+  if (jerry_value_is_error (result))
+  {
+    /* The error type/reason can be extracted via the `jerry_get_value_from_error` method */
+    printf ("Error during property configuration\r\n");
+  }
+
+  jerry_release_value (result);
+  jerry_release_value (value);
+  jerry_release_value (global);
+  jerry_cleanup();
+
+  return 0;
+}
+```
+
+## jerryx_get_property_str
+
+**Summary**
+
+Get the value of a property from the specified object with the given name.
+
+*Notes*:
+- The property name must be a zero terminated UTF-8 string.
+- There should be no '\0' (NULL) character in the name excluding the string terminator.
+- Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
+  is no longer needed.
+
+
+**Prototype**
+
+```
+jerry_value_t
+jerryx_get_property_str (const jerry_value_t target_object,
+                         const char *name);
+```
+
+- `target_object` - object on which the property name is accessed
+- `name` - property name as an UTF-8 `char*`
+- return value
+  - value of property, if success
+  - thrown error, if there was a problem accessing the property
+
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+#include "jerryscript-ext/handler.h"
+
+int
+main (int argc, char **argv)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t global = jerry_get_global_object ();
+
+  jerry_value_t math_object = jerryx_get_property_str (global, "Math");
+
+  /* use math_object */
+
+  jerry_release_value (math_object);
+  jerry_release_value (global);
+  jerry_cleanup();
+
+  return 0;
+}
+```
+
+## jerryx_has_property_str
+
+**Summary**
+
+Check if a property exists on an object.
+
+*Notes*:
+- The operation performed is the same as what the `jerry_has_property` method.
+- The property name must be a zero terminated UTF-8 string.
+- There should be no '\0' (NULL) character in the name excluding the string terminator.
+
+
+**Prototype**
+
+```
+bool
+jerryx_has_property_str (const jerry_value_t target_object,
+                         const char *name);
+```
+
+- `target_object` - object on which the property name is accessed
+- `name` - property name as an UTF-8 `char*`
+- return value
+  - true, if the given property name exsits on the object
+  - false, if there is no such property name or there was an error accessing the property
+
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+#include "jerryscript-ext/handler.h"
+
+int
+main (int argc, char **argv)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t global = jerry_get_global_object ();
+
+  bool have_math = jerryx_has_property_str (global, "Math");
+
+  jerry_release_value (global);
+  jerry_cleanup();
+
+  return have_math ? 0 : 1;
+}
+```
+
+# Utility to register multiple properties in bulk
+
+In some cases it is useful to register multiple properties for a given object
+for this the following utility structures and methods are provided.
+
+## jerryx_property_entry
+
+**Summary**
+
+Structure to define an array of properties with `name` and `value` fields which
+can be registered to a target object.
+
+The engine must be initialied before specifying the `jerry_value_t` in the struct.
+
+
+**Prototype**
+
+```c
+typedef struct {
+  const char *name;
+  jerry_value_t value;
+} jerryx_function_list_entry;
+```
+
+**See also**
+
+- [jerryx_set_properties](#jerryx_set_properties)
+
+
+## jerryx_register_result
+
+**Summary**
+
+Structure returned as the result of the [jerryx_set_properties](#jerryx_set_properties) operation.
+The `result` field will either be a JavaScript undefined value or an error object.
+In every case the `registered` field is used to indicated the number of
+successfully registered methods.
+
+This must be passed for the [jerryx_release_property_entry](#jerryx_release_property_entry) method
+after the property registration.
+
+If any error occured during the property registration the `result` field of the structure
+must be manually released after processing the error value.
+
+**Prototype**
+
+```c
+typedef struct {
+  jerry_value_t result;
+  uint32_t registered;
+} jerryx_register_result;
+```
+
+**See also**
+
+- [jerryx_set_properties](#jerryx_set_properties)
+- [jerryx_release_property_entry](#jerryx_release_property_entry)
+
+
+## jerryx_set_properties
+
+**Summary**
+
+Set multiple properties on a target object.
+
+The properties are an array of (name, jerry_value_t) pairs and
+this list must end with a (NULL, 0) entry.
+
+Important notes:
+* Each property value in the input array is released after a successful property registration.
+* The method [jerryx_release_property_entry](#jerryx_release_property_entry) must be called if there is any failed registration
+  to release the values in the entries array.
+  It is safe to call this cleanup method in every case not just in case of failure.
+* If the error value is reported via the result it must be freed manually.
+
+**Prototype**
+
+```c
+jerryx_register_result
+jerryx_set_properties (const jerry_value_t target_object,
+                       const jerryx_property_entry entries[]);
+```
+
+- `target_object` - object on which the entries will be set.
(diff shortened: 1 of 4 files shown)


akosthekiss:
I wonder whether this proposal could be merged into jerry-ext/handler somehow? That extension module contains oft-used native function implementations and a helper to register them into the _global_ object. Couldn't that be extended with helper(s) to register functions 'elsewhere', too? (Property registration would be a bit of an outlier in jerry-ext/handler, I know. At least the name "handler" doesn't seem to fit perfectly.)

zherczeg:
I have no objection to have this function.

docs/14.EXT-REFERENCE-REG-FUNCTIONS.md
```diff
@@ -0,0 +1,184 @@
...
+can be registered to a target object. When setting the `function` value
+a native C method must be used.
+
+**Prototype**
+
+```c
+typedef struct {
+  const char *name;
+  jerry_external_handler_t function;
+} jerryx_function_list_entry;
```
rerobika: My only problem with this list is that if you have a list for registering functions, later you will need a list for registering objects, strings, symbols numbers as well. Since the list contains not more than a key-value pairs I think all the mentioned types can be handled in one list. Then the structure would look like somehow like this:

```c
typedef struct {
  const char* name;
  void* data;
  jerry_value_type_t type;
}
```

I think it would be really useful to have a single universal structure for the API values registration what do you think about it?

jerry-ext/register-functions/register-functions.c
```diff
@@ -0,0 +1,80 @@
...
+ * Set multiple functions on a target object.
+ *
+ * The functions are an array of (name, function) pairs and
+ * this list must end with a (NULL, 0) entry.
+ *
+ * @return jerryx_register_result struct - if everything is ok with the (undefined, function count) values.
+ *         In case of error the (error object, registered method count) pair.
+ */
+jerryx_register_result
+jerryx_set_functions (const jerry_value_t target_object, /**< target object */
```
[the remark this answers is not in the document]
galpeter: I would rather not return a list of failed "registers" as the user would need to free the resulting array which could be error prone (someone can easily forget that).

docs/14.EXT-REFERENCE-REG-FUNCTIONS.md
galpeter: The idea is good. However, there is (at least) one problem, take the following example:
```c
struct {
  const char* name;
  void* data;
  jerry_value_type_t type;
} entries[] = {
 { "name", 3.2, JERRY_TYPE_NUMBER},
 ...
};
```
In this case the user would want to store the `3.2` as a double value which is a 64 bit data. However, on a 32 bit system the `void*` has a size of 32 bit thus we can not put a 64 bit data in it.

Another thing I was wondering is this:
```c
struct {
  const char* name;
  jerry_type_t type;
  union {
    double number;
    void* function;
    char* str;
    bool boolean;
    jerry_value_t object;
  }
} entries[] = {
  { "name", JERRY_TYPE_NUMBER, { .number = 3.2 } },
 ..
};
```
However this form of union initialization can not be used in C++ ( if I'm correct the C++17 standard or C++20 supports this now).

So.... I don't see a universally usable method allowing all kind of property value registration :(.

rerobika: However, I think we can use a workaround for the `double` variables.

```c
double my_double_variable = 3.2;

struct {
  const char* name;
  void* data;
  jerry_value_type_t type;
} entries[] = {
 { "name", &my_double_variable, JERRY_TYPE_NUMBER},
 ...
};
```

I think it could work on 32 and 64 bit as well. The only requirement is to this solution is that the double variable must be declared and initialized before the it gets part of the initialization list. The only disadvantage is that `JERRY_TYPE_NUMBER` could not be used since integer variables can be passed by value, but IMO we can introduce a new enum for the property registration. What do you think about it?

galpeter: That could work, yes. I'm not sure how useful it is in practice to have an extra variable for doubles. In addition if we go this way the initialization of the entries should be the same, thus all other data should be also initialized the same way (eg.: jerry_value, bool, ) to have a common interface.

rerobika: Okay that was only a suggestion, here come the second:

```c
/* entry registration for non-number types */
#define JERRY_EXT_GENERAL_ENTRY(name, value, type) \
  { name, (void *) value, type}

/* entry registration for number types */
#define JERRY_EXT_NUMBER_ENTRY(name, value) \
  { name, (void *) (jerry_create_number (value)), JERRY_TYPE_NUMBER}

struct {
  const char* name;
  void* data;
  jerry_value_type_t type;
} entries[] = {
 JERRY_EXT_GENERAL_ENTRY ("foo1", "string value", JERRY_TYPE_STRING),
 JERRY_EXT_GENERAL_ENTRY ("foo2", true, JERRY_TYPE_BOOLEAN),
 JERRY_EXT_NUMBER_ENTRY ("foo3", 4),
 JERRY_EXT_NUMBER_ENTRY ("foo4", 3.2),
 ...
};
```

Truly the number registration is a bit problematic, but I think it's acceptable to use an other macro for it. What's your opinion about that?

galpeter: Going further we could even use `jerry_value_t` as the data type and expect the user to correctly create all values and in addition we'll provide the macros.

rerobika: I support this idea!

docs/10.EXT-REFERENCE-HANDLER.md
```diff
@@ -1,3 +1,329 @@
+# Common methods to handle properties
+
+## jerryx_property_entry
+
+**Summary**
+
+Structure to define an array of methods with `name` and `function` fields which
+can be registered to a target object. When setting the `function` value
+a native C method must be used.
```
rerobika: The description is outdated, since not only external functions can be registered now.

docs/10.EXT-REFERENCE-HANDLER.md
```diff
@@ -1,3 +1,329 @@
...
+
+int
+main (int argc, char **argv)
+{
+  jerryx_property_entry methods[] =
+  {
+    JERRYX_PROPERTY_FUNCTION ("demo", handler),
+    JERRYX_PROPERTY_NUMBER ("test_num", 2.3),
+    JERRYX_PROPERTY_UNDEFINED ("this_is_undefined"),
+    { NULL, 0 },
```
rerobika: Probably the `{ NULL, 0 }` last element could be hidden in a `JERRYX_PROPERTY_LIST_END` macro.

jerry-ext/handler/handler-register.c
```diff
@@ -40,3 +40,113 @@ jerryx_handler_register_global (const jerry_char_t *name_p, /**< name of the fun
...
+    if (!jerry_value_is_boolean (result)) {
+      return (jerryx_register_result) { result, idx };
+    }
+
+    jerry_release_value (entry->value);
+    jerry_release_value (result);
+  }
+
+  return (jerryx_register_result) { jerry_create_undefined (), idx };
+
```
rerobika: Unnecessary new line.

jerry-ext/handler/handler-register.c
```diff
@@ -40,3 +40,113 @@ jerryx_handler_register_global (const jerry_char_t *name_p, /**< name of the fun
...
+ * @return true value - if the operation was successful
+ *         thrown error - otherwise
+ */
+jerry_value_t
+jerryx_set_property_str (const jerry_value_t target_object, /**< target object */
+                         const char *name, /**< property name */
+                         const jerry_value_t value) /**< value to set */
+{
+  jerry_value_t function_name_val = jerry_create_string ((const jerry_char_t *) name);
+  jerry_value_t result_val = jerry_set_property (target_object, function_name_val, value);
```
rerobika: What happens if the `target object` or `value` is an error? This should be checked.

jerry-ext/handler/handler-register.c
```diff
@@ -40,3 +40,113 @@ jerryx_handler_register_global (const jerry_char_t *name_p, /**< name of the fun
...
+ *    to release the values in the entries array.
+ *
+ * @return jerryx_register_result struct - if everything is ok with the (undefined, property entry count) values.
+ *         In case of error the (error object, registered property count) pair.
+ */
+jerryx_register_result
+jerryx_set_properties (const jerry_value_t target_object, /**< target object */
+                       const jerryx_property_entry entries[]) /**< array of method entries */
+{
+  uint32_t idx = 0;
```
rerobika: The `target_object` and each entry's value should be check whether they are error values or not.

jerry-ext/handler/handler-register.c
```diff
@@ -40,3 +40,113 @@ jerryx_handler_register_global (const jerry_char_t *name_p, /**< name of the fun
...
+ *    * the property name must be zero terminated.
+ *    * returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return jerry_value_t - the property value
+ */
+jerry_value_t
+jerryx_get_property_str (const jerry_value_t target_object, /**< target object */
+                         const char *name) /**< property name */
+{
+  jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) name);
```
rerobika: Ditto (for target_object).

jerry-ext/include/jerryscript-ext/handler.h
```diff
@@ -45,6 +45,46 @@ jerry_value_t jerryx_handler_gc (const jerry_value_t func_obj_val, const jerry_v
...
+#define JERRYX_PROPERTY_STRING(NAME, STR)  (jerryx_property_entry) { NAME, jerry_create_string_from_utf8 (STR) }
+#define JERRYX_PROPERTY_BOOLEAN(NAME, VALUE) (jerryx_property_entry) { NAME, jerry_create_boolean (VALUE) }
+#define JERRYX_PROPERTY_FUNCTION(NAME, FUNC) (jerryx_property_entry) { NAME, jerry_create_external_function (FUNC) }
+#define JERRYX_PROPERTY_UNDEFINED(NAME) (jerryx_property_entry) {NAME, jerry_create_undefined() }
+
+/**
+ * Stores the result of function register operation.
+ */
+typedef struct {
+  jerry_value_t result; /**< result of function registering (undefined or error object) */
```
rerobika: `result of function registering` -> `result of property registering`

jerry-ext/include/jerryscript-ext/handler.h
```diff
@@ -45,6 +45,46 @@ jerry_value_t jerryx_handler_gc (const jerry_value_t func_obj_val, const jerry_v
...
+} jerryx_property_entry;
+
+#define JERRYX_PROPERTY_NUMBER(NAME, NUMBER) (jerryx_property_entry) { NAME, jerry_create_number (NUMBER) }
+#define JERRYX_PROPERTY_STRING(NAME, STR)  (jerryx_property_entry) { NAME, jerry_create_string_from_utf8 (STR) }
+#define JERRYX_PROPERTY_BOOLEAN(NAME, VALUE) (jerryx_property_entry) { NAME, jerry_create_boolean (VALUE) }
+#define JERRYX_PROPERTY_FUNCTION(NAME, FUNC) (jerryx_property_entry) { NAME, jerry_create_external_function (FUNC) }
+#define JERRYX_PROPERTY_UNDEFINED(NAME) (jerryx_property_entry) {NAME, jerry_create_undefined() }
+
+/**
+ * Stores the result of function register operation.
```
rerobika: `function` -> `property`

tests/unit-ext/test-ext-method-register.c
```diff
@@ -0,0 +1,244 @@
...
+  jerry_init (JERRY_INIT_EMPTY);
+
+  const char *prop_ok = "prop_ok";
+  const char *prop_err = "prop_err";
+  const char *prop_not = "prop_not";
+  jerry_value_t target_object = jerry_create_object ();
+  freeze_property (target_object, prop_err);
+
+  jerryx_property_entry methods[] = {
+    { prop_ok,  jerry_create_external_function (method_hello) }, // This registration is ok
```
rerobika: Could you update all the tests to use the recommended registration macros and broaden the tests with number, string, boolean registrations?

docs/10.EXT-REFERENCE-HANDLER.md
```diff
@@ -1,3 +1,329 @@
...
+}
+```
+
+**Convenience macros**
+
+To make property registration convenient, there are a set of macros to use
+when setting a property entry:
+
+* `JERRYX_PROPERTY_NUMBER(NAME, NUMBER)` - creates a number entry.
+* `JERRYX_PROPERTY_STRING(NAME, STR)` - creates an utf8 string entry.
```
rerobika: What do you think about adding a new macro for string registration with size? I think it could be useful as well.

* `JERRYX_PROPERTY_STRING_WITH_SIZE(NAME, STR, SIZE)` - creates an utf8 string entry with the given size

jerry-ext/handler/handler-register.c
galpeter: The `jerry_set_property` method already checks if there is an error object as an argument and it'll return another error. I'm not sure if we should add another check for these functions.

rerobika: The only advantage of the early error check, that the `function_name_val` should not be created in case of error. But I'm okay with your reasoning as well, so probably we can leave it in the current status.

jerry-ext/handler/handler-register.c
```diff
@@ -40,3 +40,148 @@ jerryx_handler_register_global (const jerry_char_t *name_p, /**< name of the fun
...
+{
+  jerry_value_t property_name_val = jerry_create_string_from_utf8 ((const jerry_char_t *) name);
+  jerry_value_t result_val = jerry_set_property (target_object, property_name_val, value);
+
+  jerry_release_value (property_name_val);
+
+  return result_val;
+} /* jerryx_set_property_str */
+
+
```
dbatyai: Please fix the double newlines inbetween the added functions, there are several occurrences.

jerry-ext/handler/handler-register.c
galpeter: AFAIK, that would be too much, we'll rely on the `jerry_set_property` to generate an error on such cases and mainly on the user to not pass errors.

PR #2716, opened 2019-01-18
Increase test coverage: Array.prototype.toLocaleString

matedabis:
The coverage is measured by the following script, using `--jerry-test-suite --test262 --unittests --jerry-tests`:

Branch coverage:
Before: 15/18
After: 18/18

### DIFF
--- a/tests/jerry/array-prototype-tolocalestring.js
+++ b/tests/jerry/array-prototype-tolocalestring.js
@@ -54,3 +54,38 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+/* ES v5.1 15.4.4.3.1.
+   Checking behavior when the function's this_argument is null */
+try {
+  Array.prototype.toLocaleString.call(null);
+  assert(false);
+} catch (e) {
+  assert(e instanceof TypeError);
+}
+
+/* ES v5.1 15.4.4.3.2.
+   Checking behavior when length throws error */
+try {
+  var obj = {};
+  Object.defineProperty(obj, 'length', { 'get' : function () { throw new ReferenceError("foo"); } });
+  Array.prototype.toLocaleString.call(obj);
+  assert(false);
+} catch (e) {
+  assert(e instanceof ReferenceError);
+  assert(e.message == "foo");
+}
+
+/* ES v5.1 15.4.4.3.3.
+   Checking behavior when length is an object, which has a valueOf method that throws an error */
+var len = { };
+Object.defineProperty(len, 'valueOf', { 'get' : function () { throw new ReferenceError("vOf"); } });
+var obj = { toLocaleString : Array.prototype.toLocaleString, length : len };
+
+try {
+  obj.toLocaleString();
+  assert(false);
+} catch (e) {
+  assert(e.message === 'vOf');
+  assert(e instanceof ReferenceError);
+}


tests/jerry/array-prototype-tolocalestring.js
```diff
@@ -54,3 +54,33 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+// Checking behavior when null is passed to the function
+try {
+  Array.prototype.toLocaleString.call(null);
```
rerobika: Add an `assert (false)` after this call.

tests/jerry/array-prototype-tolocalestring.js
```diff
@@ -54,3 +54,33 @@ try {
...
+}
+
+// Checking behavior when null is passed to the function
+try {
+  var obj = {};
+  Object.defineProperty(obj, 'length', { 'get' : function () { throw new ReferenceError ("foo"); } });
+  Array.prototype.toLocaleString.call(obj);
+  assert(false);
+} catch (e) {
+  assert(e instanceof ReferenceError);
```
rerobika: Please check the message as well! (And use unique messages)

tests/jerry/array-prototype-tolocalestring.js
```diff
@@ -54,3 +54,33 @@ try {
...
+}
+
+// Checking behavior when length is an object, wichs valueOf throws error
```
rerobika: typo: `wichs`

matedabis:
Corrected according to the reviews.
Replying to this comment: https://github.com/pando-project/jerryscript/pull/2716#discussion_r249092528
in the get function, and which is being catched are not the same `ReferenceError`s, as I check the message, I get assertion fail.

tests/jerry/array-prototype-tolocalestring.js
```diff
@@ -54,3 +54,34 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+// Checking behavior when null is passed to the function
```
rerobika: This description is a bit misleading. I'd suggest the same as in #2717.

matedabis:
I updated the patch according to the review.

tests/jerry/array-prototype-tolocalestring.js
matedabis: The thrown error and the one in the function are not the same `ReferenceError`s, so as I check the message, I get assertion fail.

akosthekiss: Something is not right then. Q1: Why set the error message then, if it cannot be checked? Q2: Does the error object get replaced somewhere by the builtin implementation? Q3: If yes, why? Q4: If yes, is it necessary to throw a reference error within the anonymous function or any kind of error can be thrown?

matedabis:
Updated the patch:
- Added comments for each case containing which part of the ES 5.1 it tests.
- Ran the test code in Gecko, V8, SpiderMonkey engines and got the same result.

tests/jerry/array-prototype-tolocalestring.js
matedabis: Sorry I probably made a typo, thats why it did not work. It works with the message check.

matedabis: Sorry I made a typo, that is why it did not work.

matedabis:
Corrected the issue above.

matedabis:
Closing PR because of moving to another branch.

PR #2717, opened 2019-01-18
Increase test coverage: Array.prototype.toString

matedabis:
The coverage is measured by the following script, using `--jerry-test-suite --test262 --unittests --jerry-tests`:

Branch coverage:
Before: 5/6
After:  6/6

### DIFF
--- a/tests/jerry/array-prototype-tostring.js
+++ b/tests/jerry/array-prototype-tostring.js
@@ -61,3 +61,12 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+/* ES v5.1 15.4.4.2.1.
+   Checking behavior when the function's this_argument is undefined */
+try {
+  Array.prototype.toString.call(undefined);
+  assert(false);
+} catch (e) {
+  assert(e instanceof TypeError);
+}


tests/jerry/array-prototype-tostring.js
```diff
@@ -61,3 +61,10 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+// Checking behavior when undefined is passed to the function
+try {
+  Array.prototype.toString.call(undefined);
```
rerobika: Add an `assert(false)` after this line.

tests/jerry/array-prototype-tostring.js
```diff
@@ -61,3 +61,10 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+// Checking behavior when undefined is passed to the function
```
rerobika: At this point not an `undefined` passed to the function. The function's `this_argument` is `undefined`.

matedabis:
I updated the patch according to the review.

matedabis:
Updated the patch:
- Added comment for the case containing which part of the ES5.1 it tests.
- Ran the test code in Gecko, V8, SpiderMonkey engines and got the same result.

tests/jerry/array-prototype-tostring.js
```diff
@@ -61,3 +61,12 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+/* ES 5.1 15.4.4.2 1. */
+// Checking behavior when the function's this_argument is undefined
```
rerobika: If the related part of the standard is present, than use the following format:

```
/* ES v5.1, 15.4.4.2.1. 
   Checking behavior when the function's this_argument is undefined */
```

matedabis:
Changed comment style.

matedabis:
Closing PR because of moving to another branch.

PR #2723, opened 2019-01-22
Added new 'jerry_value_instanceof' API function.

LaszloLango:

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -1084,6 +1084,84 @@ jerry_get_global_object (void);
 
 Functions to check the type of an API value ([jerry_value_t](#jerry_value_t)).
 
+
+## jerry_instanceof
+
+**Summary**
+
+Returns whether the given `jerry_value_t` is an instance of the given prototype.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_instanceof (const jerry_value_t value, const jerry_value_t prototype);
+```
+
+- `value` - api value of instance object
+- `prototype` - api value of prototype
+- return value
+  - error, if value argument is not an object or prototype argument is not a function
+  - true, if the specified value is an instance of the given prototype
+  - false, otherwise
+
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+
+static jerry_value_t
+my_constructor (const jerry_value_t func_val,
+                const jerry_value_t this_val,
+                const jerry_value_t argv[],
+                const jerry_length_t argc)
+{
+  return jerry_create_undefined ();
+}
+
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t base_obj = jerry_create_object ();
+  jerry_value_t constructor = jerry_create_external_function (my_constructor);
+
+  /* External functions does not have a prototype by default, so we need to create one */
+  jerry_value_t prototype_str = jerry_create_string ((const jerry_char_t *) ("prototype"));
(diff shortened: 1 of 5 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -1084,6 +1084,84 @@ jerry_get_global_object (void);
...
+
+```c
+jerry_value_t
+jerry_value_instanceof (const jerry_value_t value, const jerry_value_t prototype);
+```
+
+- `value` - api value of instance object
+- `prototype` - api value of prototype
+- return value
+  - error, if value argument is not an object or prototype argument is not a function
```
galpeter: There should be a test where the input `value` (and also for the `prototype`) is not an object to trigger this case.

docs/02.API-REFERENCE.md
```diff
@@ -1084,6 +1084,84 @@ jerry_get_global_object (void);
 
 Functions to check the type of an API value ([jerry_value_t](#jerry_value_t)).
 
+
+## jerry_value_instanceof
+
+**Summary**
+
+Returns whether the given `jerry_value_t` is an instance of the given prototype.
```
galpeter: Isn't there any difference between the es5.1 vs es2015 `instanceof` internal workings? Could that affect the api workings?

LaszloLango: It should not affect the api workings, but may affect the internal workings (e.g.: instance of classes).

rerobika: Classes are simple functions, the only specializations they have is a bytecode status flag, so I think it couldn't have any affect on it.

jerry-core/api/jerry.c
```diff
@@ -601,6 +601,34 @@ jerry_get_global_object (void)
...
+                        const jerry_value_t prototype) /**< prototpye function */
+{
+  jerry_assert_api_available ();
+
+  if (ecma_is_value_error_reference (value))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+  }
+
+  ecma_object_t *proto_obj_p = ecma_get_object_from_value (prototype);
```
rerobika: Couldn't it cause an assertion fail if this function is called like `jerry_value_instanceof (jerry_create_number (5), jerry_create_number (5))`, due to the second argument is not an object?

tests/unit-core/test-api-instaceof.c
```diff
@@ -0,0 +1,104 @@
...
+{
+  (void) func_val;
+  (void) this_val;
+  (void) argv;
+  (void) argc;
+  return jerry_create_undefined ();
+} /* my_constructor */
+
+int
+main (void)
```
galpeter: There should be more test cases, for example:
* string, string
* string, number
* array, array
* array, object
* object, array
* null, object
* object, string
* typedarray, object
* arraybuffer, object
* symbols, object

jerry-core/api/jerry.c
```diff
@@ -601,6 +601,40 @@ jerry_get_global_object (void)
...
+ * Note:
+ *      returned value must be freed with jerry_release_value, when it is no longer needed.
+ *
+ * @return true  - if the specified value is an instance of the given prototype
+ *         false - otherwise
+ *         error - if value argument is not an object or prototype argument is not a function.
+ *
+ */
+jerry_value_t
+jerry_value_instanceof (const jerry_value_t value, /**< object value */
```
galpeter: I'm not sure if the `value` should be in the method name.

LaszloLango:
Closing PR to prevent merging. Let's discuss and make decision on #2746 first.

PR #2725, opened 2019-01-22
Added new 'jerry_value_strict_equal' API function.

LaszloLango:
Added new API function to perform strict equality comparison
between 'jerry_value_t's from C code.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -1630,6 +1630,63 @@ jerry_value_is_undefined (const jerry_value_t value)
 
 - [jerry_release_value](#jerry_release_value)
 
+
+## jerry_strict_equal
+
+**Summary**
+
+Perform strict equality comparison on the given operands.
+
+**Prototype**
+
+```c
+jerry_value_t jerry_strict_equal (const jerry_value_t lhs, const jerry_value_t rhs);
+```
+
+- `lhs` - left-hand side operand
+- `rhs` - right-hand side operand
+- return value
+  - error, if argument has an error flag
+  - true, if the two operands are strictly equal
+  - false, otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t value1;
+  jerry_value_t value2;
+  ... // create or acquire value
+
+  jerry_value_t result = jerry_strict_equal (value1, value2)
+
+  if (!jerry_value_is_error (result))
+  {
+    if (jerry_get_boolean_value (result))
+    {
+       // value1 and value2 are equal
+    }
+    else
+    {
+      // value1 and value2 are NOT equal
+    }
+  }
+  else
+  {
+    ... // handle error
+  }
+
+  jerry_release_value (value1);
+  jerry_release_value (value2);
+  jerry_release_value (result);
+}
+```
+
+**See also**
+
+- [jerry_value_t](#jerry_value_t)
+
+
 ## jerry_value_get_type
 
 **Summary**
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -21,6 +21,7 @@
 #include "ecma-arraybuffer-object.h"
 #include "ecma-builtin-helpers.h"
 #include "ecma-builtins.h"
+#include "ecma-comparison.h"
 #include "ecma-exceptions.h"
 #include "ecma-eval.h"
 #include "ecma-function-object.h"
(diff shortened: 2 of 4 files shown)


jerry-core/api/jerry.c
```diff
@@ -782,6 +783,23 @@ jerry_value_is_undefined (const jerry_value_t value) /**< api value */
...
+ *
+ * @return true  - if the two operands are strictly equal,
+ *         false - otherwise.
+ */
+bool
+jerry_value_strict_equal (const jerry_value_t lhs, /**< first operand */
+                          const jerry_value_t rhs) /**< second operand */
+{
+  jerry_assert_api_available ();
+
```
rerobika: If either of the arguments is an error reference, is it able to cause any side effect? Maybe a test would be great for it.

tests/unit-core/test-api-strict-equal.c
```diff
@@ -0,0 +1,63 @@
...
+
+int
+main (void)
+{
+  TEST_INIT ();
+
+  jerry_init (JERRY_INIT_EMPTY);
+
+  /* Test objects */
+  jerry_value_t obj_val = jerry_eval ((const jerry_char_t *) "o={x:1};o", 9, JERRY_PARSE_NO_OPTS);
```
galpeter: There should be way more tests for this api. To make it easier we could do something like this:
```c
typedef struct {
  jerry_value_t lhs;
  jerry_value_t rhs;
  bool should_be_equal; 
} test_entry;

test_entry tests[] = {
 { jerry_create_number(3.1), jerry_create_number(10), true },
 ....
};

for (int idx = 0; idx < sizeof(tests) / sizeof(test_entry); idx++) {
  bool result = jerry_value_strict_equal (tests[idx].lhs, tests[idx].rhs);
  if (result !=  tests[idx].should_be_equal) {
    fprintf (stderr, "Test case failed, idx: %d\n", idx);
    TEST_ASSERT (0);
  }
  jerry_release_value (tests[idx].lhs);
  jerry_release_value (tests[idx].rhs);
}
```
or something like this.

A few missing test cases:
* string vs any other string (which are different)
* array vs any other thing
* typedarray vs any other thing
* is this true?  `jerry_value_strict_equal(a,b)` <=> `jerry_value_strict_equal(b,a)`
* error objects
* arraybuffers
* symbols

docs/02.API-REFERENCE.md
```diff
@@ -1630,6 +1630,41 @@ jerry_value_is_undefined (const jerry_value_t value)
 
 - [jerry_release_value](#jerry_release_value)
 
+
+## jerry_value_strict_equal
```
galpeter: If we gave strict equal check api, then maybe we should also have the simple logic operators also (`<, >, ==`). However that would be another PR if we decide to implement it.

jerry-core/api/jerry.c
galpeter: Also if this is a special case the documentation should mention it.

LaszloLango:
I've updated the PR.

jerry-core/include/jerryscript-core.h
```diff
@@ -356,6 +356,11 @@ bool jerry_value_is_promise (const jerry_value_t value);
 bool jerry_value_is_string (const jerry_value_t value);
 bool jerry_value_is_undefined (const jerry_value_t value);
 
+/*
+ * Comparison functions
+ */
+jerry_value_t jerry_strict_equal (const jerry_value_t lhs, const jerry_value_t rhs);
```
akosthekiss: If strict comparison is made part of the public API, shouldn't `ecma_op_abstract_equality_compare` have its API counterpart, too?

akosthekiss: Hm, I've missed @galpeter 's suggestion. I'm not sure I completely agree with the one function in this PR though. The < and > operators aren't necessarily needed here, but === and == are quite related, so publishing them via API should happen together, IMO.

galpeter: @akosthekiss To have other comparison api functions was suggested in my post above, but I had another "idea" which was not written down yet: As we introduce more and more api functions for comparisons, does it make sense to have a single function for each of them or is it possible to have a single api function which can be configured to perform different kind of comparisons?

For example something like this (do not take the naming into account just the concept):
```c
enum jerry_binary_operation {
    JERRY_OP_STRICT_EQUALS, // lhs === rhs
    JERRY_OP_EQUALS, // lhs == rhs
    JERRY_OP_NOT_EQUAL, // lhs != rhs
    JERRY_OP_LESS,  // lhs < rhs
    ...
    JERRY_OP_ADD, // lhs + rhs
};

jerry_value_t
jerry_op (jerry_binary_operation op, jerry_value_t lhs, jerry_value_t rhs);
```

I'm not sure which approach is better: having api functions for each operation or having a single one. (Maybe this discussion needs its own issue.)

zherczeg: I like this idea.

LaszloLango: @galpeter I like your idea, I will do it.

PR #2737, opened 2019-01-29
Add recursion limit for VM

rerobika:
This patch adds posibility to supervise the VM call stack to avoid aborts/crashes due to the recursion calls.

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -42,6 +42,13 @@
  * @{
  */
 
+/*
+ * Check VM recursion depth limit
+ */
+#ifdef VM_RECURSION_LIMIT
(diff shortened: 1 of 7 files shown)


jerry-core/vm/vm.c
```diff
@@ -3535,6 +3546,13 @@ vm_run (const ecma_compiled_code_t *bytecode_header_p, /**< byte-code data heade
         const ecma_value_t *arg_list_p, /**< arguments list */
         ecma_length_t arg_list_len) /**< length of arguments list */
 {
+#ifdef VM_RECURSION_LIMIT
+  if (JERRY_UNLIKELY (JERRY_CONTEXT (vm_recursion_counter)-- == 0))
```
zherczeg: You should throw an error forever if you reach 0.

rerobika: Good catch! Fixed.

PR #2741, opened 2019-01-30
Add new symbol related API functions.

rerobika:
New functions:
 - ecma_create_symbol
 - ecma_value_is_symbol
 - ecma_get_symbol_descriptive_string

Also improve the `jerryx_handler_print` to be able to print symbol values via using the `ecma_get_symbol_descriptive_string` API function for symbol values.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -24,6 +24,7 @@ Enum that contains JerryScript API value types:
  - JERRY_TYPE_OBJECT - object type
  - JERRY_TYPE_FUNCTION - function type
  - JERRY_TYPE_ERROR - error/abort type
+ - JERRY_TYPE_SYMBOL - symbol type
 
 ## jerry_error_t
 
@@ -61,6 +62,7 @@ Possible compile time enabled feature types:
  - JERRY_FEATURE_REGEXP - RegExp support
  - JERRY_FEATURE_LINE_INFO - line info available
  - JERRY_FEATURE_LOGGING - logging
+ - JERRY_FEATURE_SYMBOL - symbol support
 
 ## jerry_regexp_flags_t
 
@@ -1567,6 +1569,57 @@ jerry_value_is_string (const jerry_value_t value)
 - [jerry_release_value](#jerry_release_value)
 
 
+## jerry_value_is_symbol
+
+**Summary**
+
+Returns whether the given `jerry_value_t` is a symbol value.
+
+**Prototype**
+
+```c
+bool
+jerry_value_is_symbol (const jerry_value_t value)
+```
+
+- `value` - API value
+- return value
+  - true, if the given `jerry_value_t` is a symbol
+  - false, otherwise
+
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t string_value = jerry_create_string ((const jerry_char_t *) "Symbol description string");
+  jerry_value_t symbol_value = jerry_create_symbol (string_value);
+
+  jerry_release_value (string_value);
+
+  if (jerry_value_is_symbol (symbol_value))
+  {
+    // usage of symbol_value
+  }
+
+  jerry_release_value (symbol_value);
+
+  jerry_cleanup ();
+}
+```
+
+**See also**
+
+- [jerry_release_value](#jerry_release_value)
+
+
 ## jerry_value_is_typedarray
 
 **Summary**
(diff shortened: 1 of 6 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -3416,6 +3502,48 @@ jerry_create_string_sz (const jerry_char_t *str_p,
...
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_symbol (const jerry_value_t value)
+```
+
+- `value` - api value
+- return value - value of the created symbol
```
zherczeg: ` - thrown error, otherwise`

jerry-core/include/jerryscript-core.h
```diff
@@ -367,6 +369,7 @@ typedef enum
   JERRY_TYPE_BOOLEAN,   /**< boolean type */
   JERRY_TYPE_NUMBER,    /**< number type */
   JERRY_TYPE_STRING,    /**< string type */
+  JERRY_TYPE_SYMBOL,    /**< symbol type */
```
zherczeg: These should go to the end.

rerobika:
@zherczeg Thanks for the review, I've applied your suggestions.

docs/02.API-REFERENCE.md
```diff
@@ -2691,6 +2732,51 @@ jerry_resolve_or_reject_promise (jerry_value_t promise,
...
+Call the SymbolDescriptiveString ecma builtin operation on the api value.
+
+*Note*: Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
+is no longer needed.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_symbol_to_string (const jerry_value_t value);
```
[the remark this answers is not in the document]
rerobika: I think it is necessary.
Reasoning:
```
/**
 * Call the ToString ecma builtin operation on the api value.
 */
jerry_value_t
jerry_value_to_string (const jerry_value_t value) /**< input value */
...
```
But the builtin ToString operation throws error for symbol values. If this function could deal with symbols the description would be misleading.

jerry-core/api/jerry.c
```diff
@@ -1809,13 +1863,13 @@ jerry_has_property (const jerry_value_t obj_val, /**< object value */
   jerry_assert_api_available ();
 
   if (!ecma_is_value_object (obj_val)
-      || !ecma_is_value_string (prop_name_val))
+      || !ecma_is_value_prop_name (prop_name_val))
```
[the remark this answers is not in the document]
rerobika: Yes they are. Since property names can be symbols now, these API functions should support symbol property keys as well.

docs/02.API-REFERENCE.md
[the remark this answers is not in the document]
LaszloLango: @rerobika I still prefer one function for API value to string conversion, so I vote for the description update, but I'd like to hear the opinion of others. cc: @zherczeg @akosthekiss @galpeter

tests/unit-core/test-symbol.c
```diff
@@ -0,0 +1,197 @@
...
+
+/* Symbol(bar) desciptive string */
+#define SYMBOL_DESCIPTIVE_STRING_BAR "Symbol(bar)"
+
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  if (!jerry_is_feature_enabled (JERRY_FEATURE_SYMBOL))
```
galpeter: I've just realized that there is no need to call `jerry_init` before this as the feature enable check does not depend on any runtime information. Maybe we should update the docs in a different PR to state when it is possible to call the `jerry_is_feature_enabled` function.

docs/02.API-REFERENCE.md
```diff
@@ -2691,6 +2732,51 @@ jerry_resolve_or_reject_promise (jerry_value_t promise,
...
+```
+
+- `value` - symbol value
+- return value
+  - converted string value, if success
+  - thrown error, otherwise
+
+**Example**
```
galpeter: Would it be possible to have fully compilable, runnable examples? We have doc testing in place. That way we can catch if the API changes but the docs do not. (Not just here but in every case in this PR).

docs/02.API-REFERENCE.md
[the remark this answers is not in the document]
galpeter: This is a good question. The base `jerry_value_to_string` should perform the abstract toString operation, that is: the conversion to string. However the `jerry_symbol_to_string` performs the `Symbol('ds').toString()` (like) operation which is a bit different.
It makes sense to have two different operation, but maybe the naming is a bit off.

docs/02.API-REFERENCE.md
```diff
@@ -3416,6 +3502,50 @@ jerry_create_string_sz (const jerry_char_t *str_p,
...
+
+```c
+jerry_value_t
+jerry_create_symbol (const jerry_value_t value)
+```
+
+- `value` - api value
+- return value
+  - value of the created symbol, if success
+  - thrown error, otherwise
```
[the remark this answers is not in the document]
rerobika: Yes there [is](https://github.com/pando-project/jerryscript/pull/2741/files#diff-62b25ad3af984fc2434d2760c6a02b3fR150).

jerry-core/api/jerry.c
```diff
@@ -768,6 +769,25 @@ jerry_value_is_string (const jerry_value_t value) /**< api value */
...
+ *         false - otherwise
+ */
+bool
+jerry_value_is_symbol (const jerry_value_t value) /**< api value */
+{
+  jerry_assert_api_available ();
+
+#ifndef CONFIG_DISABLE_ES2015_SYMBOL_BUILTIN
+  return ecma_is_value_symbol (value);
+#else /* CONFIG_DISABLE_ES2015_PROMISE_BUILTIN */
```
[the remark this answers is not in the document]
rerobika: Good catch!

rerobika:
@LaszloLango @galpeter Thanks for the reviews, I've updated the PR.

rerobika:
@galpeter Do you have any other thoughts/comments on this PR?

tests/unit-core/test-symbol.c
```diff
@@ -0,0 +1,196 @@
...
+  jerry_value_t bar_symbol = jerry_create_symbol (bar_symbol_desc);
+  TEST_ASSERT (!jerry_value_is_error (bar_symbol));
+  TEST_ASSERT (jerry_value_is_symbol (bar_symbol));
+
+  jerry_release_value (bar_symbol_desc);
+
+  jerry_value_t bar_symbol_string = jerry_get_symbol_descriptive_string (bar_symbol);
+  TEST_ASSERT (jerry_value_is_string (bar_symbol_string));
+
+  jerry_size_t bar_symbol_string_size = jerry_get_string_size (bar_symbol_string);
```
galpeter: The `bar_symbol_string_size` should also be checked if it is of the expected size.

rerobika: Good idea!

docs/02.API-REFERENCE.md
```diff
@@ -2766,6 +2819,59 @@ jerry_resolve_or_reject_promise (jerry_value_t promise,
...
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t string_value = jerry_create_string ((const jerry_char_t *) "foo");
+  jerry_value_t symbol_value = jerry_create_symbol (string_value);
+
+  jerry_release_value (string_value);
+
+  jerry_value_t symbol_desc_string = jerry_get_symbol_descriptive_string (symbol_value);
+
+  // usage of symbol_to_string
```
galpeter: AFAIK, this should be: `symbol_desc_string` an not `symbol_to_string`

rerobika: Good catch!

PR #2746, opened 2019-02-01
Added new 'jerry_binary_operation' API function

LaszloLango:

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -276,6 +276,18 @@ An opaque declaration of the JerryScript context structure.
 typedef struct jerry_context_t jerry_context_t;
 ```
 
+
+## jerry_binary_operation_t
+
+Enum that contains the supported binary operation types
+ - JERRY_BIN_OP_EQUAL - equal comparison (==)
+ - JERRY_BIN_OP_STRICT_EQUAL - strict equal comparison (===)
+ - JERRY_BIN_OP_LESS - less relation (<)
+ - JERRY_BIN_OP_LESS_EQUAL - less or equal relation (<=)
+ - JERRY_BIN_OP_GREATER - greater relation (>)
+ - JERRY_BIN_OP_GREATER_EQUAL - greater or equal relation (>=)
+
+
 ## jerry_property_descriptor_t
 
 **Summary**
@@ -1700,6 +1712,69 @@ jerry_is_feature_enabled (const jerry_feature_t feature);
 
 }
 ```
+
+**See also**
+- [jerry_feature_t](#jerry_feature_t)
+
+
+# Binary operations
+
+## jerry_binary_operation
+
+**Summary**
+
+Perform binary operation on the given operands (==, ===, <, >, etc.).
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_binary_operation (jerry_binary_operation_t op,
+                        const jerry_value_t lhs,
+                        const jerry_value_t rhs);
+```
+
+- `op` - binary operation
+- `lhs` - left-hand side operand
+- `rhs` - right-hand side operand
+- return value
+  - error, if argument has an error flag or operation is unsuccessful or unsupported
+  - true/false, the result of the binary operation on the given operands otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t value1;
+  jerry_value_t value2;
+  ... // create or acquire value
+  jerry_value_t result = jerry_binary_operation (JERRY_BIN_OP_EQUAL, value1, value2)
+
+  if (!jerry_value_is_error (result))
+  {
+    if (jerry_get_boolean_value (result))
+    {
+       // value1 and value2 are equal
+    }
+    else
+    {
+      // value1 and value2 are NOT equal
+    }
+  }
+  else
+  {
+    ... // handle error
+  }
+
+  jerry_release_value (value1);
+  jerry_release_value (value2);
+  jerry_release_value (result);
+}
+```
+
+**See also**
+- [jerry_binary_operation_t](#jerry_binary_operation_t)
+
 
 # Error manipulation functions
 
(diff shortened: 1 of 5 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -1701,6 +1713,70 @@ jerry_is_feature_enabled (const jerry_feature_t feature);
...
+jerry_value_t
+jerry_binary_operation (jerry_binary_operation_t op,
+                        const jerry_value_t lhs,
+                        const jerry_value_t rhs);
+```
+
+- `op` - binary operation
+- `lhs` - left-hand side operand
+- `rhs` - right-hand side operand
+- return value
```
galpeter: AFAIK the return value description is not correct.

jerry-core/api/jerry.c
```diff
@@ -901,6 +903,58 @@ jerry_is_feature_enabled (const jerry_feature_t feature) /**< feature to check *
           );
 } /* jerry_is_feature_enabled */
 
+/**
+ * Perform binary operation on the given operands (==, ===, <, >, etc.).
+ *
+ * @return true  - if the two operands are strictly equal
```
galpeter: Incorrect doc comments.

jerry-core/api/jerry.c
```diff
@@ -901,6 +903,58 @@ jerry_is_feature_enabled (const jerry_feature_t feature) /**< feature to check *
...
+  if (ecma_is_value_error_reference (lhs) || ecma_is_value_error_reference (rhs))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+  }
+
+  switch (op)
+  {
+    case JERRY_BIN_OP_EQUAL:
+    {
+      return  ecma_op_abstract_equality_compare (lhs, rhs);
```
galpeter: double space after the `return` keyword.

jerry-core/api/jerry.c
```diff
@@ -901,6 +903,58 @@ jerry_is_feature_enabled (const jerry_feature_t feature) /**< feature to check *
...
+    }
+    case JERRY_BIN_OP_STRICT_EQUAL:
+    {
+      return  ecma_make_boolean_value (ecma_op_strict_equality_compare (lhs, rhs));
```
galpeter: double space after the `return` keyword.

tests/unit-core/test-api-binary-operations.c
```diff
@@ -0,0 +1,299 @@
...
+    T (JERRY_BIN_OP_GREATER_EQUAL, jerry_create_boolean (true), jerry_create_boolean (true), true),
+    T (JERRY_BIN_OP_GREATER_EQUAL, jerry_create_boolean (true), jerry_create_boolean (false), true),
+    T (JERRY_BIN_OP_GREATER_EQUAL, jerry_create_boolean (false), jerry_create_boolean (true), false),
+    T (JERRY_BIN_OP_GREATER_EQUAL, jerry_create_boolean (false), jerry_create_boolean (false), true),
+  };
+
+  for (uint32_t idx = 0; idx < sizeof (tests) / sizeof (test_entry_t); idx++)
+  {
+    jerry_value_t result = jerry_binary_operation (tests[idx].op, tests[idx].lhs, tests[idx].rhs);
+    fprintf (stderr, "%d\n", idx);
```
galpeter: Is the `fprint` here intentional?

LaszloLango:
@galpeter thanks for the review, I've updated the PR.

jerry-core/api/jerry.c
```diff
@@ -901,6 +903,57 @@ jerry_is_feature_enabled (const jerry_feature_t feature) /**< feature to check *
...
+  if (ecma_is_value_error_reference (lhs) || ecma_is_value_error_reference (rhs))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+  }
+
+  switch (op)
+  {
+    case JERRY_BIN_OP_EQUAL:
+    {
+      return ecma_op_abstract_equality_compare (lhs, rhs);
```
rerobika: This function can throw an error [here](https://github.com/pando-project/jerryscript/blob/master/jerry-core/ecma/operations/ecma-comparison.c#L161), so I think ` jerry_return` should be used there.

docs/02.API-REFERENCE.md
```diff
@@ -1701,6 +1713,69 @@ jerry_is_feature_enabled (const jerry_feature_t feature);
...
+  - error, if argument has an error flag or operation is unsupported
+  - true/false, the result of the binary operation on the given operands otherwise
```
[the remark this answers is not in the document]
rerobika: I think there are 3 ways to get error:
 - operation, if either of the argument has error flag - mentioned
 - operation, if the given operation is unsupported - mentioned
 - the result of the executed binary-operation also could be an error, but the description only mentions true/false.

jerry-core/ecma/operations/ecma-comparison.c
```diff
@@ -31,6 +31,10 @@
  *
  * See also: ECMA-262 v5, 11.9.3
  *
+ * Note:
+ *      This function might raise an exception, so the
+ *      returned value must be freed with ecma_free_value.
```
galpeter: So... then the  `@return` entry below is incorrect. Please update it accordingly.

LaszloLango: I thought this would be an appropriate fix for the wrong `return` comment. Could you suggest something which makes it more clearer? Remove the `Note` and put the sentence into the `@return` block?

galpeter: We can keep the note and modify the return like this:
```
return true - if values are equal,
       false - if values are not equal,
       error - in case of any problems
```

LaszloLango:
@galpeter thanks for the review, I've updated the PR

jerry-core/include/jerryscript-core.h
```diff
@@ -308,6 +308,19 @@ typedef struct
  */
 typedef struct jerry_context_t jerry_context_t;
 
+/**
+  * Enum that contains the supported binary operation types
```
galpeter: Nitpick: the `*` is not in the correctly column.

LaszloLango: Good catch, fixed it.

PR #2748, opened 2019-02-01
Fix instanceof operator for implicit class constructors

rerobika:
Implicit class constructor functions should not be handled as bound functions during [[HasInstance]] check.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -324,6 +324,70 @@ ecma_op_arrow_function_get_compiled_code (ecma_arrow_function_t *arrow_function_
 
 #endif /* !CONFIG_DISABLE_ES2015_ARROW_FUNCTION */
 
+#ifndef CONFIG_DISABLE_ES2015_CLASS
+/**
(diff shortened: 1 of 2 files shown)


rerobika:
Note: The bug was found via this test case on the latest master.

``` js
class myArray extends Array { }

print ([] instanceof myArray) // true, which is incorrect
```

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -324,6 +324,70 @@ ecma_op_arrow_function_get_compiled_code (ecma_arrow_function_t *arrow_function_
 
 #endif /* !CONFIG_DISABLE_ES2015_ARROW_FUNCTION */
 
+#ifndef CONFIG_DISABLE_ES2015_CLASS
+/**
+ * Helper function for implicit class constructors [[HasInstance]] check.
```
galpeter: Is it possible to add reference to the standard here?

rerobika: I think it's not. This implementation is not part of the standard directly, it's only needed due special implementation of the implicit class constructors.

PR #2751, opened 2019-02-07
Added 'instanceof' binary operation to the API

LaszloLango:


docs/02.API-REFERENCE.md
```diff
@@ -1092,6 +1093,7 @@ jerry_get_global_object (void);
 - [jerry_release_value](#jerry_release_value)
 - [jerry_define_own_property](#jerry_define_own_property)
 
+
```
rerobika: This new line is unnecessary.

LaszloLango: IMHO it is necessary. We put two spaces before top level headers (`#`) to make it more readable in raw format. It helps to notice where a new chapter starts. This style is used in the whole MD file.

LaszloLango:
@galpeter good catch, the commit message also wrong. I'll fix it.

jerry-core/include/jerryscript-core.h
```diff
@@ -313,12 +313,13 @@ typedef struct jerry_context_t jerry_context_t;
...
-  JERRY_BIN_OP_LESS_EQUAL,   /**< less or equal relation (<=) */
-  JERRY_BIN_OP_GREATER,      /**< greater relation (>) */
-  JERRY_BIN_OP_GREATER_EQUAL /**< greater or equal relation (>=)*/
+  JERRY_BIN_OP_EQUAL = 0u,    /**< equal comparison (==) */
+  JERRY_BIN_OP_STRICT_EQUAL,  /**< strict equal comparison (===) */
+  JERRY_BIN_OP_LESS,          /**< less relation (<) */
+  JERRY_BIN_OP_LESS_EQUAL,    /**< less or equal relation (<=) */
+  JERRY_BIN_OP_GREATER,       /**< greater relation (>) */
+  JERRY_BIN_OP_GREATER_EQUAL, /**< greater or equal relation (>=)*/
+  JERRY_BIN_OP_INSTANCEOF     /**< instanceof operation */
```
galpeter: At other places there is a comma (`,`) after the last value. It is good to follow that as if new operations are added only one line needs changing.

PR #2754, opened 2019-02-11
Compile with clang-7.0

wille-io:
Next try. `jerry_debugger_scope_variable_type_t` was not explicitly sized by the compiler as 8bit.
Instead the fn could accept a `uint8_t`. You still can call the fn with a jerry_debugger_scope_variable_type_t as parameter.

### DIFF
--- a/jerry-core/debugger/debugger.c
+++ b/jerry-core/debugger/debugger.c
@@ -271,11 +271,12 @@ jerry_debugger_send_scope_chain (void)
 
 /**
  * Get type of the scope variable property.
+ * @return (jerry_debugger_scope_variable_type_t)
  */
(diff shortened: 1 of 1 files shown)


akosthekiss:
@wille-io You don't have to close a PR just because a fix didn't work out. (Unless you give up on the fix completely.) You can just keep working on the same branch and push / force push it to GitHub. It will update the PR automatically. It is a better approach than opening a new PR for every try. Please see the [GitHub Flow intro (section "Discuss and review your code")](https://guides.github.com/introduction/flow/).

wille-io:
@akosthekiss Can we please have a talk about me failing git @ irc? Sadly nobody was answering there for several days.

wille-io:
reopen

PR #2759, opened 2019-02-14
Fix JSON.stringify

imiklos:
There was a wrong check on the property types, it did not check that the
property was a named accessor. Also fixed a test case.


LaszloLango:
@rerobika as you see we have had a test for it already, but the assertion in the testcase was wrong. Fixing the wrong test case is good enough to me. Of course I am not against adding more tests if they test uncovered behaviour and code path.

Issue #2745, opened 2019-01-31
Are there any plans to support Set() defined by ES6?


acoinfo:
Are there any plans to support Set() defined by ES6?
And report a bug: Map() did not support forEach() method.

LaszloLango:
CC: @rerobika

acoinfo:
And 'let' is a very necessary keyword, I recommend implementation ^_^

akosthekiss:
@acoinfo Issues should be focused. Right now, this issue seems to contain
- a question ("Are there any plans to support Set() defined by ES6?"),
- a bug report (?) ("Map() did not support forEach() method."), and
- a feature request ("'let' is a very necessary keyword, I recommend implementation").

It's a lot easier to conduct discussions and reach conclusions if an issue is about a single topic. So, I'd suggest you to open separate issues for unrelated questions/feedbacks.

Moreover,
- for bug reports, it is strongly preferred to include various details, like the version (git hash) the bug is reported against; the way the project is configured, built, and executed; the platform used to build and/or run the engine; a reproduction case (js file); and how the bug manifests itself; while
- for feature requests, it is good practice to provide motivation and/or use cases that justify the proposal.

(Note: Actually, the bug report here is more like a feature request. The lack of support for part of the ES6/2015 standard is not a bug, as the project explicitly states that it implements only a subset of the ES6/2015 standard at the moment.)

rerobika:
@acoinfo Despite of having a separate issue for each question would be great let me give you a quick answer.

- The Set builtin probably will be supported, but first we need to figure out how to combine the implementation with the Map builtin to reduce the code duplication.
- It is not a bug, it's a feature request. Please open an issue for it.
- `let` and `const` is more harder nut to crack. I'm working on it but it won't be published until their performance/memory usage produce similar results to what can be achieved via using `var`s.

rerobika:
@acoinfo https://github.com/pando-project/jerryscript/pull/2841 has landed, so now you can use the Set builtin object, also the missing Map.prototype.foreach routine as well.

cup:
https://github.com/jerryscript-project/jerryscript/issues/3169

PR #2760, opened 2019-02-15
Rework Map object

rerobika:
This patch reworks the core of the builtin Map object.

Advantages:
 - Provide sublinear access time for the elements via Lcache and property hashmap
 - This implementation is suitable for the builtin Set object as well

Also add the missing 'forEach' routine for the builtin object as well.


rerobika:
Related issue: #2745

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -234,6 +235,41 @@ ecma_prop_name_is_symbol (ecma_string_t *string_p) /**< ecma-string */
...
+ */
+ecma_string_t *
+ecma_new_map_key_string (ecma_value_t value) /**< non prop-name ecma-value */
+{
+  JERRY_ASSERT (!ecma_is_value_prop_name (value));
+
+  ecma_string_t *string_p = ecma_alloc_string ();
+  string_p->refs_and_container = ECMA_STRING_REF_ONE | ECMA_STRING_CONTAINER_MAP_KEY;
+  string_p->u.value = ecma_copy_value_if_not_object (value);
+  string_p->hash = ecma_is_value_simple (value) ? (uint16_t) value : 0;
```
zherczeg: This hash is too simple.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1862,6 +1905,13 @@ ecma_compare_ecma_non_direct_strings (const ecma_string_t *string1_p, /**< ecma-
     return false;
   }
 
+#ifndef CONFIG_DISABLE_ES2015_MAP_BUILTIN
+  if (string1_container == ECMA_STRING_CONTAINER_MAP_KEY)
+  {
+    return ecma_op_same_value_zero (string1_p->u.value, string2_p->u.value);
```
zherczeg: This looks like a quite costly operation.

rerobika: Indeed. It's a bit costly but it's required, since with the current implementation the following tests are failing.

```js
var m = new Map();
m.set(NaN, "not a number");
assert(m.size === 1);
assert(m.get(NaN) === "not a number"); // Currently: m.get(NaN) is undefined
assert(m.get(Number("foo")) === "not a number") // Currently: m.get(Number("foo") is undefined

m.set(0, "PosZero");
assert(m.size === 2); // Currently: m.size === 2
m.set(-0, "NegZero");
assert(m.size === 2); // Currently: m.size === 3
assert(m.get (0) === "NegZero");
assert(m.get (-0) === "NegZero");

```

rerobika:
Note: This patch depends on #2766.

rerobika:
The patch is done and ready for the review.

jerry-core/ecma/operations/ecma-map-object.c
```diff
@@ -87,113 +102,79 @@ ecma_op_map_get_object (ecma_value_t this_arg) /**< this argument */
 } /* ecma_op_map_get_object */
 
 /**
- * Returns with the size of the map object.
+ * Create a property key for the internal object from the given argument
```
LaszloLango: `Creates ...`

PR #2764, opened 2019-02-19
Separate the debugger-client to support the later usage of other communication protocols

robertsipka:

### DIFF
--- a/jerry-debugger/jerry_client_tcp.py
+++ b/jerry-debugger/jerry_client_tcp.py
@@ -1,0 +1,48 @@
+#!/usr/bin/env python
+
+# Copyright JS Foundation and other contributors, http://js.foundation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import socket
+import select
+
+class Socket(object):
+    """ Create a new socket using the given address family, socket type and protocol number. """
+    def __init__(self, socket_family=socket.AF_INET, socket_type=socket.SOCK_STREAM, proto=0, fileno=None):
+        self.socket = socket.socket(socket_family, socket_type, proto, fileno)
+
+    def connect(self, address):
+        """
+        Connect to a remote socket at address (host, port).
+        The format of address depends on the address family.
+        """
+        self.socket.connect(address)
+
+    def close(self):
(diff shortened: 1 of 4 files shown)


jerry-debugger/jerry_client_tcp.py
```diff
@@ -0,0 +1,50 @@
...
+    def receive_data(self, max_size=1024):
+        """ The maximum amount of data to be received at once is specified by max_size. """
+        return self.socket.recv(max_size)
+
+    def send_data(self, data):
+        """ Send data to the socket. The socket must be connected to a remote socket. """
+        return self.socket.send(data)
+
+    def ready(self):
+        """ Monitor the file descriptor, return weither it became ready for the I/O operations or not. """
```
zherczeg: I think this should only be used with receive_data, so the comment is not exactly right.

jerry-debugger/jerry_client_tcp.py
```diff
@@ -0,0 +1,50 @@
...
+        result = select.select([self.socket], [], [], 0)[0]
+        if self.socket not in result:
+            return False
+
+        return True
```
rerobika: I think you can use `return self.socket in result` here.

robertsipka: Indeed, good catch. :)

robertsipka:
@zherczeg @rerobika : Thanks, I've modified the patch accordingly.

jerry-debugger/jerry_client_websocket.py
```diff
@@ -0,0 +1,135 @@
...
+
+import struct
+from jerry_client_tcp import Socket
+
+MAX_BUFFER_SIZE = 128
+WEBSOCKET_BINARY_FRAME = 2
+WEBSOCKET_FIN_BIT = 0x80
+
+class WebSocket(object):
+    def __init__(self, address, protocol=Socket()):
```
galpeter: Using the `Socket()` as default argument is not correct. The `protocol` will have a reference to the created object and if the WebSocket is constructed multiple times the same socket will be used in both cases.
example.:
```
a = WebSocket()
b = WebSocket()
a.protocol == b.protocol
```

We should use `None` as default and add an check in the constructor to create the socket object.

jerry-debugger/jerry_client_websocket.py
```diff
@@ -0,0 +1,135 @@
...
+                    b"Connection: Upgrade\r\n" +
+                    b"Sec-WebSocket-Accept: s3pPLMBiTxaQ9kYGzzhZRbK+xOo=\r\n\r\n")
+
+        len_expected = len(expected)
+
+        while len(self.data_buffer) < len_expected:
+            self.data_buffer += self.protocol.receive_data()
+
+        if self.data_buffer[0:len_expected] != expected:
+            raise Exception("Unexpected handshake")
```
galpeter: In the future it would be good to have specialized Exception classes so we can catch our module's exceptions in a uniform fashion and would make it possible to distinguish between the "base"  exceptions ans our exceptions.

galpeter: This applies to all exceptions in this PR.

PR #2766, opened 2019-02-20
Merge external direct magic strings into direct magic strings

rerobika:


rerobika:
Instead of the a 1 free bit gain for other direct string types there is some performance gain, also the binary size is decreased as well:

Benchmark | Perf (sec) |
----: | ----: | 
3d-cube.js | 0.798 -> 0.795 : +0.391% | 
3d-raytrace.js | 0.965 -> 0.957 : +0.807% | 
access-binary-trees.js | 0.528 -> 0.524 : +0.775% | 
access-fannkuch.js | 1.920 -> 1.898 : +1.154% | 
access-nbody.js | 1.013 -> 1.009 : +0.380% | 
bitops-3bit-bits-in-byte.js | 0.470 -> 0.470 : -0.044% | 
bitops-bits-in-byte.js | 0.644 -> 0.645 : -0.214% | 
bitops-bitwise-and.js | 0.891 -> 0.896 : -0.614% | 
bitops-nsieve-bits.js | 1.126 -> 1.101 : +2.201% | 
controlflow-recursive.js | 0.342 -> 0.326 : **+4.476%** | 
crypto-aes.js | 0.808 -> 0.809 : -0.027% | 
crypto-md5.js | 0.538 -> 0.536 : +0.460% | 
crypto-sha1.js | 0.539 -> 0.539 : +0.059% | 
date-format-tofte.js | 0.725 -> 0.724 : +0.120% | 
date-format-xparb.js | 0.502 -> 0.501 : +0.269% | 
math-cordic.js | 1.122 -> 1.123 : -0.019% | 
math-partial-sums.js | 0.695 -> 0.702 : -1.029% | 
math-spectral-norm.js | 0.472 -> 0.471 : +0.087% | 
string-base64.js | 1.338 -> 1.345 : -0.509% | 
string-fasta.js | 1.290 -> 1.269 : +1.619% | 
Geometric mean: | **+0.524%** | 

Binary (bytes) | master(a1fdcf06cb) | patch(88dcb5d264) | Diff |
size | 132664 | 132664 | 0 bytes |
.rodata | 11255 | 11255 | 0 bytes |
.dynstr | 262 | 262 | 0 bytes |
.rel.plt | 176 | 176 | 0 bytes |
.interp | 25 | 25 | 0 bytes |
.dynsym | 384 | 384 | 0 bytes |
.gnu.hash | 188 | 188 | 0 bytes |
.text | 114404 | 114268 | **-136 bytes** |
.comment | 59 | 59 | 0 bytes |
.shstrtab | 250 | 250 | 0 bytes |
.data | 8 | 8 | 0 bytes |
.ARM.exidx | 8 | 8 | 0 bytes |
.rel.dyn | 16 | 16 | 0 bytes |
.init | 12 | 12 | 0 bytes |
.got | 104 | 104 | 0 bytes |
.plt | 288 | 288 | 0 bytes |
.note.gnu.build-id | 36 | 36 | 0 bytes |
.note.ABI-tag | 32 | 32 | 0 bytes |
.gnu.version_r | 64 | 64 | 0 bytes |
.bss | 1575168 | 1575168 | 0 bytes |
.fini | 8 | 8 | 0 bytes |
.gnu.version | 48 | 48 | 0 bytes |
.fini_array | 4 | 4 | 0 bytes |
.init_array | 4 | 4 | 0 bytes |
.dynamic | 232 | 232 | 0 bytes |
.eh_frame | 4 | 4 | 0 bytes |
.ARM.attributes | 51 | 51 | 0 bytes |
.jcr | 4 | 4 | 0 bytes |

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -135,18 +134,19 @@ ecma_string_get_chars_fast (const ecma_string_t *string_p, /**< ecma-string */
   {
     if (ECMA_GET_DIRECT_STRING_TYPE (string_p) == ECMA_DIRECT_STRING_MAGIC)
     {
-      lit_magic_string_id_t id = (lit_magic_string_id_t) ECMA_GET_DIRECT_STRING_VALUE (string_p);
+      uint32_t id = (uint32_t) ECMA_GET_DIRECT_STRING_VALUE (string_p);
```
zherczeg: We need to document somewhere that we don't support UINT32_MAX number of magic strings anymore. It is impractical anyway, but we need to document it. We could say at least `UINT32_MAX / 2` is supported, maybe more. Also, if set external magic strings is called, and the number of magic strings is greater than the limit, we should "crop" the rest:
```
if (number_of_ext_magic_string > UINT32_MAX - MAGIC_STRINGS)
   number_of_ext_magic_string = UINT32_MAX - MAGIC_STRINGS
```

LaszloLango:
The jerry-main and sunspider test suite does not really use external magic strings. It would be great to use a custom example which highly depends on external magic strings.

jerry-core/ecma/base/ecma-helpers-string.c
rerobika: Good idea!

rerobika:
@zherczeg I've added the missing documentation.
@LaszloLango You can find the performance results [here](https://gist.github.com/rerobika/48d6bb580bd57f3fd0ad32774d941bae). This file contains the runtime change of the sunspider tests. These results were measured on RPi2 via using snapshots.

zherczeg:
I don't see the doc update. Does this affect snapshots btw?

jerry-core/ecma/base/ecma-helpers-string.c
rerobika: I'd vote for the `UINT32_MAX/2` since `MAGIC_STRING_COUNT` is not a constant value.

rerobika:
@zherczeg I've added the missing API doc, I think it may affect the snapshot so I increased the snapshot version number.

LaszloLango:
@rerobika IoT.js does not work for me with this PR. Could you investiage it? It highly depends on external magic strings, so it might reveal a bug.

docs/02.API-REFERENCE.md
```diff
@@ -654,6 +654,8 @@ someplace_in_the_code (void)
 Registers an external magic string array.
 
 *Note*: The strings in the array must be sorted by size at first, then lexicographically.
+*Note*: The maximum number of external magic strings is limited to 2147483648 (UINT32_MAX / 2).
+        If there are more than 2147483648 external magic strings the overplus is cropped.
```
zherczeg: "the extra is cropped."

rerobika:
@LaszloLango Thanks for the review I'll take a look at it.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1559,7 +1559,8 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
           ecma_property_pair_t *prop_pair_p = (ecma_property_pair_t *) prop_iter_p;
 
           if (ECMA_PROPERTY_GET_NAME_TYPE (*property_p) == ECMA_DIRECT_STRING_MAGIC
-              && prop_pair_p->names_cp[i] >= LIT_NON_INTERNAL_MAGIC_STRING__COUNT)
+              && prop_pair_p->names_cp[i] >= LIT_NON_INTERNAL_MAGIC_STRING__COUNT
+              && prop_pair_p->names_cp[i] < LIT_MAGIC_STRING__COUNT)
```
rerobika: @LaszloLango After several hours of debugging I found the cause of the error. I think it's working now with IoT.js, but could you please check it again?

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -1462,19 +1469,35 @@ ecma_string_get_chars (const ecma_string_t *string_p, /**< ecma-string */
...
-        length = size;
+        if (id >= LIT_MAGIC_STRING__COUNT)
+        {
+          id -= LIT_MAGIC_STRING__COUNT;
+          size = lit_get_magic_string_ex_size (id);
+          result_p = lit_get_magic_string_ex_utf8 (id);
+          length = 0;
 
-        result_p = lit_get_magic_string_utf8 (id);
+          if (JERRY_UNLIKELY (*flags_p & ECMA_STRING_FLAG_IS_ASCII))
```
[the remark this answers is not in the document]
rerobika: Yes, it is intentional, since ext magic string can be non-ascii strings. The length for these kind of strings is calculated later.

PR #2789, opened 2019-03-05
Make jerryx_port_handler_print_char truly a port function

galpeter:
In the previous approach `jerryx_port_handler_print_char` was implemented
in by the jerry-default-port. This implementation however required the
jerry-ext handler header file which created a requirement for the
jerry-default-port for the jerry-ext headers.


jerry-port/default/default-io.c
```diff
@@ -98,3 +98,32 @@ jerry_port_log (jerry_log_level_t level, /**< message log level */
...
+#endif /* JERRY_DEBUGGER */
+
+/**
+ * Default implementation of jerry_port_print_char. Uses 'printf' to
+ * print a single character to standard output.
+ */
+void
+jerry_port_print_char (char c) /**< the character to print */
+{
+  printf ("%c", c);
```
dbatyai: Since the purpose of this function is to print a single character, would it make sense to use `putchar`? Could that make a difference?

galpeter: I'm not against it, currently I've just moved the original implementation. AFAIK if there is a `printf` on a port, the `putchar` should also be there. So it should not change anything. Should I change it to `putchar`?

LaszloLango: Both `printf` and `putchar` are good to me.

dbatyai: I think `putchar` would be preferable here, but if you feel that it's outside of the scope of this change, then I won't insist.

galpeter:
@dbatyai I've updated the PR. Now the putchar is used in the default port

dbatyai:
lgtm

PR #2790, opened 2019-03-05
 Add a "single" source/header file generation and build mode

galpeter:
Introduce a new way of building: before any file is compiled
all source files are combined into a single C file and all
header files into a single H file (per subdir).

This new approach makes it possible to quickly integrate JerryScript
into other projects:

```
$ gcc -o demo demo.c jerryscript.c jerryscript-port-default.c -lm
```

To use the source generator run:
```
$ cmake -Bbuild_dir -H. -DENABLE_ALL_IN_ONE_SOURCE=ON
$ make -C build_dir generate-single-source
```

This will create the following files in the `build_dir`:
* jerryscript.c
* jerryscript.h
* jerryscript-config.h
* jerryscript-port-default.c
* jerryscript-port-default.h

### DIFF
--- a/jerry-core/CMakeLists.txt
+++ b/jerry-core/CMakeLists.txt
@@ -150,6 +150,56 @@ if(ENABLE_ALL_IN_ONE)
   set(SOURCE_CORE_FILES ${ALL_IN_FILE})
 endif()
 
+# "Single" JerryScript source/header build.
+#  The process will create the following files:
+#   * jerryscript.c
+#   * jerryscript.h
+#   * jerryscript-config.h
+if(ENABLE_ALL_IN_ONE_SOURCE)
+  # Create a default configuration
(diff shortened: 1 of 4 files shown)


galpeter:
This PR currently depends on the #2789 PR. (That commit is also included in this PR)

LaszloLango:
This PR contains #2789 too. Does this PR depend on #2789?

galpeter:
@LaszloLango I would not add the docs yet, as the `jerryscript-config.h` usage and generation is not finalized at this moment. That would be another PR to make it stable. If I include that here then it would be too big imho.

LaszloLango:
@galpeter follow up PR is good for me.

galpeter:
@LaszloLango I've fixed the alignment problem and rebased to master.

Issue #2520, opened 2018-09-11
Consolidate configuration macros


akosthekiss:
The engine's code base has lots of macro guards, making it well-configurable. However, these macros grew somewhat organically and got named as `CONFIG_{EN,DIS}ABLE_xxx`, `CONFIG_xxx`, `{EN,DIS}ABLE_xxx`, `JERRY_{EN,DIS}ABLE_xxx`, `JERRY_xxx`, or `modulename_xxx`. Quite a mix. Moreover, right now it is encoded in some of the names of the macros whether they are enabled/disabled by default (but not in all of the names). And boolean configuration macros make their effect based on whether they are defined or not, while the rest carry their meaning via their content.

I propose to consolidate the configuration macros name and value wise, as shown in the table below:

<sub>

| Current macro | Proposed macro | Proposed value |
|-|-|-|
| **ES5.1 features** |
| CONFIG_DISABLE_BUILTINS                  | JERRY_BUILTINS                  | {0,*1*} |
| CONFIG_DISABLE_ANNEXB_BUILTIN            | JERRY_ANNEXB_BUILTIN            | {0,*1*} |
| CONFIG_DISABLE_ARRAY_BUILTIN             | JERRY_ARRAY_BUILTIN             | {0,*1*} |
| CONFIG_DISABLE_BOOLEAN_BUILTIN           | JERRY_BOOLEAN_BUILTIN           | {0,*1*} |
| CONFIG_DISABLE_DATE_BUILTIN              | JERRY_DATE_BUILTIN              | {0,*1*} |
| CONFIG_DISABLE_ERROR_BUILTINS            | JERRY_ERROR_BUILTINS            | {0,*1*} |
| CONFIG_DISABLE_JSON_BUILTIN              | JERRY_JSON_BUILTIN              | {0,*1*} |
| CONFIG_DISABLE_MATH_BUILTIN              | JERRY_MATH_BUILTIN              | {0,*1*} |
| CONFIG_DISABLE_NUMBER_BUILTIN            | JERRY_NUMBER_BUILTIN            | {0,*1*} |
| CONFIG_DISABLE_REGEXP_BUILTIN            | JERRY_REGEXP_BUILTIN            | {0,*1*} |
| CONFIG_DISABLE_STRING_BUILTIN            | JERRY_STRING_BUILTIN            | {0,*1*} |
| **ES2015 features** |
[...]

PR #2793, opened 2019-03-08
Rework usages/naming of configuration macros

galpeter:
There are quite a few configuration macros in the project.
As discussed in the #2520 issue there are a few awkward constructs.

Main changes:

* Renamed all `CONFIG_DISABLE_<name>_BUILTIN` macro to `JERRY_BUILTIN_<name>` format.
* The special `JERRY_BUILTINS` macro specifies the basic config for all es5.1 builtins.
* Renamed all `CONFIG_DISABLE_ES2015_<name>` to `JERRY_ES2015_<name>` format.
* The special `JERRY_ES2015` macro specifies the basic config for all es2015 builtins.
* Renamed `UNICODE_CASE_CONVERSION` to `JERRY_UNICODE_CASE_CONVERSION`.
* Renamed `ENABLE_REGEXP_STRICT_MODE` to `JERRY_REGEXP_STRICT_MODE`.
* All options (in this change) can have a value of 0 or 1.
* Renamed `ENABLE_REGEXP_STRICT_MODE` to `JERRY_REGEXP_STRICT_MODE`.
  `JERRY_REGEXP_STRICT_MODE` is set to 0 by default.
* Reworked `CONFIG_ECMA_NUMBER_TYPE` macro to `JERRY_NUMBER_TYPE_FLOAT64` name and now
  it uses the value 1 for 64 bit floating point numbers and 0 for 32 bit floating point
  number.
  By default the 64-bit floating point number mode is enabled.
* All new `JERRY_` defines can be used wit the `#if ENABLED (JERRY_...)` construct to
  test if the feature is enabled or not.
* Added/replaced a few config.h includes to correctly propagate the macro values.
* Added sanity checks for each macro to avoid incorrectly set values.
* Updated profile documentation.
* The CMake feature names are not updated at this point.


jerry-core/api/jerry-snapshot.c
```diff
@@ -42,12 +42,12 @@ snapshot_get_global_flags (bool has_regex, /**< regex literal is present */
 
   uint32_t flags = 0;
 
-#ifndef CONFIG_DISABLE_REGEXP_BUILTIN
+#if defined (JERRY_BUILTIN_REGEXP) && (JERRY_BUILTIN_REGEXP == 1)
```
rerobika: As I see there is a sanity check macro guard for the incorrect configuration:

```c
#if !defined (JERRY_BUILTIN_REGEXP) \
|| ((JERRY_BUILTIN_REGEXP != 0) && (JERRY_BUILTIN_REGEXP != 1))
# error "Invalid value for JERRY_BUILTIN_REGEXP macro."
#endif
```
Therefore wouldn't be much easier to use only the

```c
#if JERRY_BUILTIN_REGEXP == 1
//code
#endif
```

guard all over the codebase? I think these double checks could make the code much harder to read especially if two or more guards are combined [e.g.](https://github.com/pando-project/jerryscript/pull/2793/files#diff-8ffb0036b085bcd4039b9adcb34f9ad2R1049).

LaszloLango:
Please rebase to the current master

dbatyai:
I agree with @rerobika, we should aim to make the code as readable as we can.

Since each guard is checked to have well defined values, I would even go as far as to say that we should skip the comparison checks as well (in case of boolean guards).

For example, instead of
```
#if (JERRY_A  == 1) && (JERRY_B == 0)
```
we could simply write
```
#if JERRY_A && !JERRY_B
```

What do you think?

galpeter:
@dbatyai @rerobika I'm fine with simplifying (the sanity check would catch all problems).
Another thing I was considering is the format: `#if ENABLED(JERRY_FEATURE)` which could hide all defined and comparison checks.

Which one would you prefer?

dbatyai:
I think `#if ENABLED(JERRY_FEATURE)`  is fine, it's more descriptive while remaining well readable

rerobika:
The `#if ENABLED(JERRY_FEATURE)` is fine for me as well.

galpeter:
@rerobika @dbatyai I've just updated with the `ENABLED` version :)

jerry-core/config.h
```diff
@@ -107,4 +189,128 @@
  */
 #define CONFIG_ECMA_GC_NEW_OBJECTS_SHARE_TO_START_GC (16)
 
+
+/**
+ * Sanity check for macros to see if the values are 0 or 1
```
rerobika: Are these sanity checks are generated somehow, or they should be written by hand? If the second, it would be good to add a note for add these checks whenever a new feature guard is introduced.

rerobika: Also, as the feature guards are grouped by categories (es5.1, es2015, other), I think these checks should be separated as well for the easier maintenance. (Probably a comment is enough)

galpeter: I'll add comments for these.

PR #2795, opened 2019-03-12
Fix the vm stop checking bug

wangzhikang123:
when the vm executes VM_OC_LESS operation, the lookahead optimization will cause the by-passing of vm stop checking.
Thus we need to disable the optimization here if JERRY_VM_EXEC_STOP is defined.

JerryScript-DCO-1.0-Signed-off-by: Wang Zhikang wzk0406@mail.ustc.edu.cn


zherczeg:
Are there other similar cases?

wangzhikang123:
> LGTM

wangzhikang123:
> Are there other similar cases?

not found yet

Issue #2733, opened 2019-01-28
 Assertion 'id < LIT_NON_INTERNAL_MAGIC_STRING__COUNT' failed in lit_get_magic_string_size


renatahodovan:
###### Jerry version:

```
Checked revision: 924f4bbd

Build command: ./tools/build.py --clean --debug --compile-flag=-m32 --profile=es2015-subset --system-allocator=on --error-messages=on --logging=on
```

###### OS:

```
Linux-4.15.0-43-generic-x86_64-with-Ubuntu-18.04-bionic
```

###### Test case:

```javascript
var fz_globalObject = this

function fz_is_primitive(value) {
    var value_type = typeof value
    if (value_type !== "function" && value_type !== "object")
        return value_type
}

function fz_starts_with(str, pattern) {
    for (var i = 0; i < pattern.length; i++)
        if (str[i] !== pattern[i])
            return
    return true
}

function fz_collect_values(value) {
    var primitive = fz_is_primitive(value)
    if (primitive)
        return
    var prop_names = Object.getOwnPropertyNames(value)
    for (var i = 0; i < prop_names.length; i++) {
        var prop_name = prop_names[i]
        if (!fz_starts_with(prop_name, "fz_")) {
            fz_collect_values(value[prop_name])
        }
    }
}

fz_collect_values(fz_globalObject)
```

###### Backtrace:

```
ICE: Assertion 'id < LIT_NON_INTERNAL_MAGIC_STRING__COUNT' failed at jerryscript/jerry-core/lit/lit-magic-strings.c(lit_get_magic_string_size):77.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0xf7fd5059 in __kernel_vsyscall ()
(gdb) bt
#0  0xf7fd5059 in __kernel_vsyscall ()
#1  0xf7de1832 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf7de2cc1 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0x5655ca1e in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x565a009a in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION)
    at jerryscript/jerry-core/jrt/jrt-fatals.c:58
[...]

PR #2804, opened 2019-03-26
Implement DataView builtin

rerobika:
New API functions:
 - jerry_create_dataview
 - jerry_value_is_dataview
 - jerry_get_dataview_buffer


tests/jerry/es2015/regression-test-issue-2733.js
```diff
@@ -41,7 +41,12 @@ function fz_collect_values(value) {
     for (var i = 0; i < prop_names.length; i++) {
         var prop_name = prop_names[i]
         if (!fz_starts_with(prop_name, "fz_")) {
-            fz_collect_values(value[prop_name])
+            try {
```
LaszloLango: It is not so good we have to touch a regression test for this new feature. How does this effect the original issue (#2733)?

jerry-core/config.h
```diff
@@ -39,6 +39,7 @@
 # define CONFIG_DISABLE_ES2015_ARROW_FUNCTION
 # define CONFIG_DISABLE_ES2015_BUILTIN
 # define CONFIG_DISABLE_ES2015_CLASS
+# define CONFIG_DISABLE_ES2015_DATAVIEW_BUILTIN
```
LaszloLango: Please update the `jerry-core/profiles/README.md` file with this new define.

jerry-core/ecma/operations/ecma-dataview-object.c
```diff
@@ -0,0 +1,323 @@
...
+  return NULL;
+} /* ecma_op_dataview_get_object */
+
+/**
+ * Helper union to specify the system's endiannes
+ */
+typedef union
+{
+  uint32_t number; /**< for write numeric data */
+  char data[sizeof (void *)]; /**< for read numeric data */
```
galpeter: Why not `sizeof (uint32_t)` ?

tests/jerry/es2015/regression-test-issue-2733.js
rerobika: After a bit of investigation I realized that the original test case was invalid, and the only reason why it worked was the "luckiness" of the property enumeration order. Therefore I deleted this file.

rerobika:
@galpeter @LaszloLango Thanks for the reviews I've updated the patch according to them.

rerobika:
Rebased after #2793 has landed.

jerry-core/config.h
LaszloLango: This haven't been fixed yet.

rerobika: You can find the new define's description [here](https://github.com/pando-project/jerryscript/pull/2804/files#diff-52116e59bc8cbc847a29fc0adf4a41f3R34).

PR #2801, opened 2019-03-22
Remove outer loop from ecma_op_function_call

rerobika:
JerryScript-DCO-1.0-Signed-off-by: Robert Fancsik frobert@inf.u-szeged.hu

PR #2805, opened 2019-03-27
Fix bound functions with no arguments after #2801

rerobika:


LaszloLango:
I have a few questions about this fix:
1) Could you add a regression test for it?
2) What are the new performance results? Is this fix better then reverting #2801?

rerobika:
@LaszloLango Thanks for the remarks I've added a test case. The performance/binary size is unchanged after the fix.

PR #2808, opened 2019-03-28
Update for build

lidija-b:
I was looking at mips builds and got across failing tests (around 90 test-suite, 4 unittests, 30 jerry-tests) because it was not set for mips to use __LITTLE_ENDIAN macro.
Happened both for cross compile and native build with and without toolchain file. This would effect any build without  toolchain file.

JerryScript-DCO-1.0-Signed-off-by: Lidija Besker lidija.besker@rt-rk.com

### DIFF
--- a/jerry-libm/jerry-libm-internal.h
+++ b/jerry-libm/jerry-libm-internal.h
@@ -30,14 +30,22 @@
 /* Sometimes it's necessary to define __LITTLE_ENDIAN explicitly
    but these catch some common cases. */
 
+
+#ifndef __LITTLE_ENDIAN
+/* Check if compiler has byte order macro. Some older versions do not.
+ * If byte order is supported and set to little or target is among common
+ * cases checked define __LITTLE_ENDIAN.
+ */
 #if (defined (i386) || defined (__i386) || defined (__i386__) || \
      defined (i486) || defined (__i486) || defined (__i486__) || \
      defined (intel) || defined (x86) || defined (i86pc) || \
      defined (__alpha) || defined (__osf__) || \
      defined (__x86_64__) || defined (__arm__) || defined (__aarch64__) || \
-     defined (__xtensa__))
+     defined (__xtensa__) || defined (__MIPSEL)) || \
+(defined (__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))
 #define __LITTLE_ENDIAN
 #endif
+#endif /* !__LITTLE_ENDIAN */
 
 #ifdef __LITTLE_ENDIAN
 #define __HI(x) *(1 + (int *) &x)


jerry-libm/jerry-libm-internal.h
```diff
@@ -38,6 +39,7 @@
      defined (__xtensa__))
```
galpeter: In case of GCC/Clang we could use the compiler's `__BYTE_ORDER__` macro to check if it is equal to `__ORDER_LITTLE_ENDIAN__` and then defined the `__LITTLE_ENDIAN` macro here. That way we don't need to add extra defines to the toolchain/build files. Would that solve this problem?

lidija-b: If it doesn't make problem for other architectures, it is simpler option.
I should work for mips. I will test it.

galpeter: or I think it is also possible to simply check the `MIPSEL` or `_MIPSEL` macro definitions.

lidija-b: True, `MIPSEL` macros are defined and would solve this specific problem but maybe generic option would be a good idea (in case of new architectures)? Is there reason each target is checked separately instead of checking endians?

galpeter: This libm is from the https://www.netlib.org/fdlibm/ library and the macros were used that way. At that time (of import) we did not changed the architecture checks. However, I'm fine with changing this to use the byte order macro.

galpeter: ahh.. I can think one problem: if someone uses an older compiler which does not support the byte order macro, I think pre gcc 4.6 is one such compiler.

galpeter: Maybe the best way would be to check the MIPSEL macro if it is defined, and also check if the byte order macro is defined and set to little endian mode. That way all sides are covered.

lidija-b: Updated so that if there is `__BYTE_ORDER__` endians are checked there and if not gets to checking targets with `MIPSEL` added.

jerry-libm/jerry-libm-internal.h
```diff
@@ -30,14 +30,25 @@
...
+#if (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__)
+#define __LITTLE_ENDIAN
+/* Common cases are checked when byte order macro is not supported. */
+#elif (defined (i386) || defined (__i386) || defined (__i386__) || \
      defined (i486) || defined (__i486) || defined (__i486__) || \
      defined (intel) || defined (x86) || defined (i86pc) || \
      defined (__alpha) || defined (__osf__) || \
      defined (__x86_64__) || defined (__arm__) || defined (__aarch64__) || \
-     defined (__xtensa__))
+     defined (__xtensa__) || defined (__MIPSEL))
```
galpeter: Can't we just add the `(defined (__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))` after the MIPSEL macro? In addition: do we need to check the `__LITTLE_ENDIAN` at the top?

lidija-b: `(defined (__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))` after MIPSEL sure, no problem. I have just put the broadest check first. 
`#ifndef __LITTLE_ENDIAN` check was needed with previous version, now it is not necessary but if macro is defined anywhere before (like command line) build will fail so I left it.

lidija-b: I have changed order of checks and removed update of mipsel toolchain but left` #ifndef __LITTLE_ENDIAN`.

lidija-b: Hi, is there anything that needs to be changed in code for it to be merged?

LaszloLango: First of all the CI must be green, please check the logs: https://travis-ci.org/pando-project/jerryscript/jobs/514555761

lidija-b: Warnings are for something this pull request doesn't change. I'll see to it. Thank you.

lidija-b: Files with error are changed 2 years ago and errors are:
```
jerry-libm/sqrt.c:237: error(uninitvar): Uninitialized variable: z
jerry-libm/trig.c:736: error(uninitvar): Uninitialized variable: qx
```
but this is same for all other checks but no errors are reported.
I will correct it but I am little confused by this.

LaszloLango: I will take a look at on this later. @galpeter any idea why did these issues appear?

galpeter: My first guess would be that the cppcheck incorrectly detects the macro expansion. I'll check it on my pc.

galpeter: yeah... cppcheck can't properly process the new byte_order check and will incorrectly assume that the __LO/__HI calls are functions....

galpeter: @lidija-b by reworking the parentheses a bit the cppcheck now correctly processes the files, could you also check it?

Here's the diff (note the closing parenthesis after the MIPSEL check) :
```diff
 #if (defined (i386) || defined (__i386) || defined (__i386__) || \
      defined (i486) || defined (__i486) || defined (__i486__) || \
      defined (intel) || defined (x86) || defined (i86pc) || \
      defined (__alpha) || defined (__osf__) || \
      defined (__x86_64__) || defined (__arm__) || defined (__aarch64__) || \
-     defined (__xtensa__))
+     defined (__xtensa__) || defined (__MIPSEL)) || \
+     (defined (__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))
 #define __LITTLE_ENDIAN
 #endif
```

lidija-b: Looks good to me, pushed with suggested change

galpeter: looks good, but sadly the vera checker complains that the byteorder check line does have an indentation. We need to remove the whitespaces before that line.

lidija-b: I rushed to push, sorry. Corrected, pushed and passed checks.

galpeter:
LGTM

jerry-libm/jerry-libm-internal.h
```diff
@@ -30,14 +30,22 @@
...
      defined (i486) || defined (__i486) || defined (__i486__) || \
      defined (intel) || defined (x86) || defined (i86pc) || \
      defined (__alpha) || defined (__osf__) || \
      defined (__x86_64__) || defined (__arm__) || defined (__aarch64__) || \
-     defined (__xtensa__))
+     defined (__xtensa__) || defined (__MIPSEL)) || \
+(defined (__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))
 #define __LITTLE_ENDIAN
 #endif
+#endif
```
LaszloLango: missing comment: `#endif __LITTLE_ENDIAN`

lidija-b: Done

PR #2814, opened 2019-04-08
Allow the JS objects to have more than one native pointer data

rerobika:
Currently JS objects can only have one native pointer data which could be a limitation in special cases. This patch allows to register multiple native infos, which can be accessed/associated with the corresponding `jerry_object_native_info_t`. 

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -4756,23 +4756,11 @@ jerry_set_prototype (const jerry_value_t obj_val,
 
 **Summary**
 
-Get native pointer and its type information.
+Get native pointer by the given type information.
 The pointer and the type information are previously associated with the object by jerry_set_object_native_pointer.
 
-*Note*: It is recommended to ensure that the `out_native_info_p` value pointer
-        is equal to the native info pointer that is expected, before casting
-        and accessing the `out_native_pointer_p`.
-        An example of when this is important: external functions that expect
-        `this` to have a native pointer of a certain C type.
-        It is possible in JavaScript to change `this` at will – using `call()`,
-        `apply()` or `bind()`. Therefore, it is possible that the native pointer
-        of `this` is not of the expected C type. To handle this safely and
-        securely, one must always add type checks to make sure that the
-        `out_native_pointer_p` is of the expected type, before casting
-        and dereferencing `out_native_pointer_p`.
-
-*Note*: `out_native_pointer_p` and `out_native_info_p` can be NULL, and it means the
-        caller doesn't want to get the native_pointer or type information.
+*Note*: `out_native_pointer_p` can be NULL, and it means the
+        caller doesn't want to get the native_pointer.
 
 **Prototype**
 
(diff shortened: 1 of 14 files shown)


rerobika:
Note: The patch contains API breaking changes so if the reviewers support this idea this should be land before the 2.0 release. I highly recommend to check [this](https://github.com/rerobika/jerryscript/blob/native_pointers/docs/02.API-REFERENCE.md#jerry_get_object_native_pointer) example to understand the motivation of the PR.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -294,10 +294,11 @@ typedef struct
 /**
  * Representation for native pointer data.
  */
-typedef struct
+typedef struct ecma_native_pointer_t
 {
   void *data_p; /**< points to the data of the object */
   ecma_object_native_info_t *info_p; /**< native info */
+  struct ecma_native_pointer_t *next_p; /**< points to the next ecma_native_pointer_t element */
```
zherczeg: This could be changed later but adding a pointer doubles the size of this structure.

LaszloLango:
I like this proposal. I have two comments on this:
1) a `jerry_delete_object_native_pointer` function is also needed
2) `jerry_get_object_native_pointer` should return with `void *` instead of `bool`, where `NULL` means there is no such pointer.

legendecas:
> 2. `jerry_get_object_native_pointer` should return with `void *` instead of `bool`, where `NULL` means there is no such pointer.

There might be cases that the native pointer might be NULL yet the JS objects been bound with gc-callbacks.

rerobika:
> There might be cases that the native pointer might be NULL yet the JS objects been bound with gc-callbacks.

@legendecas Is this a frequently used structure? Could you show me an example for it?

LaszloLango:
@rerobika @legendecas I don't mind to keep the NULL value as a valid native pointer.

legendecas:
> Is this a frequently used structure? Could you show me an example for it?

I don't see the frequent usage of this behavior nor any example for it, sorry for that. Yet if the API doesn't force the native pointer to be non-NULL, then that is a valid usage anyway.

docs/02.API-REFERENCE.md
```diff
@@ -4897,6 +4997,40 @@ best-practice example.
...
+## jerry_delete_object_native_pointer
+
+**Summary**
+
+Delete the native pointer of the specified object associated with the given native type info.
+You can get them by calling jerry_get_object_native_pointer later.
+
+*Note*: If the specified object has no matching native pointer for the given native type info the operation has no effect.
+
+*Note*: This operation cannot throw an exception.
```
LaszloLango: Merge these notes like this:

```
*Note*:
  - If the specified object has no matching native pointer for the given native type info the operation has no effect.
  - This operation cannot throw an exception.
```

jerry-core/api/jerry.c
```diff
@@ -2605,20 +2604,19 @@ jerry_objects_foreach_by_native_info (const jerry_object_native_info_t *native_i
 } /* jerry_objects_foreach_by_native_info */
 
 /**
- * Get native pointer and its type information, associated with specified object.
+ * Get native pointer and its type information, associated with the given native type info
```
LaszloLango: Missing period character at the end of the sentence.

rerobika:
@LaszloLango Thanks for the review, I updated the PR.

PR #2818, opened 2019-04-12
Set constructor to prototype for ES2015 Classes

legendecas:
Add missing step 18 of https://www.ecma-international.org/ecma-262/6.0/#sec-runtime-semantics-classdefinitionevaluation 


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -607,6 +612,12 @@ parser_parse_class (parser_context_t *context_p, /**< context */
 
   JERRY_ASSERT (context_p->token.type == LEXER_RIGHT_BRACE);
 
+  parser_emit_cbc_literal (context_p, CBC_PUSH_LITERAL, class_ident_index);
```
rerobika: Good catch! I think we can use the new code block like that:

```c
if (!is_class_env)  {
// code here
}
```

Since if the inheritance is present, the constructor set [here](https://github.com/pando-project/jerryscript/blob/master/jerry-core/vm/vm.c#L1398).

LaszloLango:
@legendecas don't forget to change the state of the PR from draft when it is ready to review.

legendecas:
> @legendecas don't forget to change the state of the PR from draft when it is ready to review.

Yes, I would update the code when I got free and change the PR status :)

jerry-core/parser/js/js-parser-expr.c
legendecas: After several tries, I found that changing byte codes might not be the elegant way to implement the feature. I've update the PR, could you have another review? :D

jerry-core/vm/vm.c
```diff
@@ -1393,8 +1393,16 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
 
           ecma_object_t *child_class_p = ecma_get_object_from_value (child_value);
           ecma_object_t *child_prototype_class_p = ecma_get_object_from_value (child_prototype_value);
-          ecma_property_value_t *prop_value_p;
 
+          ecma_property_t *property_p = ecma_find_named_property (child_prototype_class_p,
+                                                                  ecma_get_magic_string (LIT_MAGIC_STRING_CONSTRUCTOR));
+          if (property_p != NULL)
```
rerobika: If the constructor property is set in `VM_OC_PUSH_CLASS_CONSTRUCTOR_AND_PROTOTYPE` is it necessary to delete it? After this patch, I think the delete/create operations become unnecessary here.

legendecas: Yes, it's unnecessary here. I've remove it.

jerry-core/vm/vm.c
```diff
@@ -1440,7 +1432,20 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
           ecma_extended_object_t *ext_func_obj_p = (ecma_extended_object_t *) function_obj_p;
           ext_func_obj_p->u.external_handler_cb = ecma_op_function_implicit_constructor_handler_cb;
 
-          *stack_top_p++ = ecma_make_object_value (function_obj_p);
+          ecma_value_t function_obj_value = ecma_make_object_value (function_obj_p);
+          *stack_top_p++ = function_obj_value;
+
+          ecma_object_t *prototype_class_p = ecma_create_object (ecma_builtin_get (ECMA_BUILTIN_ID_OBJECT_PROTOTYPE),
+                                                     0,
```
LaszloLango: Wrong indentation

jerry-core/vm/vm.c
```diff
@@ -1440,7 +1432,20 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+                                                     ECMA_OBJECT_TYPE_GENERAL);
+          *stack_top_p++ = ecma_make_object_value (prototype_class_p);
+
+          /* 14.5.14.18 Set constructor to prototype */
+          ecma_property_value_t *prop_value_p = ecma_create_named_data_property (prototype_class_p,
+                                                          ecma_get_magic_string (LIT_MAGIC_STRING_CONSTRUCTOR),
```
LaszloLango: wrong indentation

legendecas: Update :)

PR #2836, opened 2019-04-18
Make the underlying buffer of the ArrayBuffer accessible.

LaszloLango:

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -3198,34 +3198,29 @@ jerry_get_arraybuffer_byte_length (const jerry_value_t value) /**< ArrayBuffer *
  * Get a pointer for the start of the ArrayBuffer.
  *
  * Note:
- *    * Only valid for ArrayBuffers created with jerry_create_arraybuffer_external.
  *    * This is a high-risk operation as the bounds are not checked
  *      when accessing the pointer elements.
- *    * jerry_release_value must be called on the ArrayBuffer when the pointer is no longer needed.
  *
  * @return pointer to the back-buffer of the ArrayBuffer.
- *         pointer is NULL if the parameter is not an ArrayBuffer with external memory
-             or it is not an ArrayBuffer at all.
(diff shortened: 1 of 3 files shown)


tests/unit-core/test-arraybuffer.c
```diff
@@ -326,8 +326,6 @@ main (void)
       sum += data[i];
     }
 
-    jerry_release_value (buffer);
```
galpeter: Why was this removed?

LaszloLango: I feel this kind of implicit value acquire leads to unexpected memory leaks.

LaszloLango: Furthermore it cannot stop users to try to access the pointer after the buffer is freed.

rerobika: +1 for removing the extra reference. That's totally the user's responsibility to manage the buffer correctly.

galpeter: ahh true

jerry-core/api/jerry.c
```diff
@@ -3198,34 +3198,29 @@ jerry_get_arraybuffer_byte_length (const jerry_value_t value) /**< ArrayBuffer *
...
 uint8_t *
-jerry_get_arraybuffer_pointer (const jerry_value_t value) /**< Array Buffer to use */
+jerry_get_arraybuffer_pointer (const jerry_value_t array_buffer) /**< Array Buffer to use */
 {
   jerry_assert_api_available ();
+
 #if ENABLED (JERRY_ES2015_BUILTIN_TYPEDARRAY)
-  if (!ecma_is_arraybuffer (value))
+  if (ecma_is_value_error_reference (array_buffer)
+      || !ecma_is_arraybuffer (array_buffer))
```
galpeter: Doesn't the ecma_is_arraybuffer already checks error references?

LaszloLango: No it doesn't.

galpeter:
lgtm

PR #2841, opened 2019-04-26
Implement the Set builtin object

rerobika:
Also implement the missing iterator initializer part from the Map builtin object constructor.


jerry-core/ecma/operations/ecma-iterator-object.c
```diff
@@ -161,6 +161,241 @@ ecma_op_create_iterator_object (ecma_value_t iterated_value, /**< value from cre
...
+ *
+ * Note:
+ *      Returned value must be freed with ecma_free_value.
+ *
+ * @return ietartor result object - if success
+ *         raised error - otherwise
+ */
+static ecma_value_t
+ecma_op_iterator_next (ecma_value_t iterator, /**< iterator value */
+                       ecma_value_t value)
```
LaszloLango: missing comment

jerry-core/config.h
```diff
@@ -274,6 +278,10 @@
 || ((JERRY_ES2015_BUILTIN_MAP != 0) && (JERRY_ES2015_BUILTIN_MAP != 1))
 # error "Invalid value for JERRY_ES2015_BUILTIN_MAP macro."
 #endif
+#if !defined (JERRY_ES2015_BUILTIN_SET) \
+|| ((JERRY_ES2015_BUILTIN_SET != 0) && (JERRY_ES2015_BUILTIN_SET != 1))
+# error "Invalid value for JERRY_ES2015_BUILTIN_SET macro."
+#endif
```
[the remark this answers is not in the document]
rerobika: All the other configuration checks above do not have comment either so I think it's not necessary to add it.

rerobika:
@LaszloLango Thanks for the review, I've updated the PR.

jerry-core/ecma/builtin-objects/ecma-builtin-map-prototype.c
```diff
@@ -35,7 +35,7 @@
  */
 
 /**
- * The Map.prototype object's 'clear' routine
+ * The Set.prototype object's 'clear' routine
```
LaszloLango: I think this meant to be `Map/Set`. It is odd the comment changed for this function, but not for the other similar functions in this file (e.g.: `ecma_builtin_map_prototype_object_delete`).

jerry-core/ecma/builtin-objects/ecma-builtin-set-prototype.c
```diff
@@ -0,0 +1,143 @@
...
+ * @return ecma value
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_builtin_set_prototype_object_foreach (ecma_value_t this_arg, /**< this argument */
+                                           ecma_value_t predicate, /**< callback function */
+                                           ecma_value_t predicate_this_arg) /**< this argument for
+                                                                             *   invoke predicate */
+{
+  return ecma_op_map_foreach (this_arg, predicate, predicate_this_arg, true);
```
LaszloLango: I would rename these `ecma_op_map_` functions to `ecma_op_container_`.

jerry-core/ecma/builtin-objects/ecma-builtin-map-prototype.c
rerobika: Thanks, this change was unnecessary.

tests/jerry/es2015/set.js
```diff
@@ -0,0 +1,80 @@
...
+assert (set.size === 4);
+assert (set.delete (number));
+assert (set.size === 3);
+assert (set.delete (object));
+assert (set.size === 2);
+assert (set.delete (func));
+assert (set.size === 1);
+assert (set.delete (symbol));
+assert (set.size === 0);
+
```
dbatyai: Please add a test for adding the same value twice, and for the 'clear' method as well.

rerobika: Thanks for the suggestion, I've added these test cases as well.

PR #2845, opened 2019-04-30
Add core API jerry_value_instanceof

legendecas:


LaszloLango:
`instanceof` operation is accessible via the API sincs https://github.com/pando-project/jerryscript/pull/2751

PR #2861, opened 2019-05-06
Implement path normalization in port-default

dbatyai:
Updated the default port so that it now provides path normalization for
common platforms.

### DIFF
--- a/jerry-port/default/default-io.c
+++ b/jerry-port/default/default-io.c
@@ -13,6 +13,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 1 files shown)


rerobika:
The patch seems OK for me, but could be please add some test cases to test the new behavior?

dbatyai:
As far as I can tell it's not possible to test how many times a source code was evaluated with the testing infrastructure we currently have in place.

Even if it would be possible, this is entirely port specific, so I'd rather not force an expected behavior.

PR #2872, opened 2019-05-13
Fix heap limit calculation when using the system allocator.

dbatyai:


dbatyai:
Depends on #2874.

dbatyai:
Depends on #2875.

PR #2875, opened 2019-05-15
Properly release property name collection in for-in opcode handler

dbatyai:

### DIFF
--- a/jerry-core/vm/opcodes.c
+++ b/jerry-core/vm/opcodes.c
@@ -252,7 +252,9 @@ opfunc_for_in (ecma_value_t left_value, /**< left value */
     *result_obj_p = ecma_make_object_value (obj_p);
   }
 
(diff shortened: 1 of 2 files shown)


dbatyai:
Just realized this is not the correct fix, I'll have to double check this.

dbatyai:
Here's the thing: the for-in implementation directly manages the ecma collection structure, instead of relying on the iterator to do so, which also means that it releases the collection on it's own way.

This feels wrong, but unfortunately it's not an easy fix. We can't store the iterator pointer as a compressed pointer on the stack, since the iterator may be 4 byte aligned.

So instead of fixing the entire problem I resorted to keeping it a quick fix, and simply release the header the appropriate way, through the pool allocator.

LaszloLango:
@dbatyai what about adding a TODO or NOTE comment with this information you mentioned above?

dbatyai:
Done.

PR #2878, opened 2019-05-17
jerryscript-ext/module.h: extern and symbol macro for modules

IngwiePhoenix:
The two macros `JERRYX_MODULE_EXTERN(module_name)` and `JERRYX_MODULE_SYMBOL(module_name)` have been added.

- `_EXTERN(module_name)`: Returns an `extern jerryx_native_module_t` declaration.
- `_SYMBOL(module_name)`: Returns the identifier that references the `jerryx_native_module_t` symbol.

### DIFF
--- a/jerry-ext/include/jerryscript-ext/module.h
+++ b/jerry-ext/include/jerryscript-ext/module.h
@@ -83,6 +83,12 @@ typedef struct jerryx_native_module_t
   {                                                                    \
     jerryx_native_module_unregister(&_##module_name##_definition);     \
   }
+  
+#define JERRYX_NATIVE_MODULE_EXTERN(module_name) \
+  extern jerryx_native_module_t _ ## module_name ## _definition
+
+#define JERRYX_NATIVE_MODULE_SYMBOL(module_name) \
+  _ ## module_name ## _definition
 
 /**
  * Register a native module. This makes it available for loading via jerryx_module_resolve, when
--- a/jerry-ext/include/jerryscript-ext/module.h
+++ b/jerry-ext/include/jerryscript-ext/module.h
@@ -59,36 +59,108 @@ typedef struct jerryx_native_module_t
 #define JERRYX_NATIVE_MODULE(module_name, on_resolve_cb)  \
   JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb)
 
-#define JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb)        \
-  static jerryx_native_module_t _ ## module_name ## _definition =      \
-  {                                                                    \
-    .name_p = (jerry_char_t *) #module_name,                           \
-    .on_resolve_p = (on_resolve_cb),                                   \
-    .next_p = NULL                                                     \
-  };                                                                   \
-                                                                       \
+/**
+ * @brief  Defines the symbol name. This is used in any place where the generated symbol name
+ *         might be needed.
+ * 
+ * @param module_name: Name of the module.
+ * @returns: A concatenated identifier, representing the symbol used for the native module's structure.
+ */
+#define JERRYX_NATIVE_MODULE_SYMBOL(module_name) \
+  _ ## module_name ## _definition
+
+/**
+ * @brief  Generates a structure of `jerryx_native_module_t`.
+ * 
+ * @param module_name: Name of the module.
+ * @param on_resolve_cb: The function to be called once the module is resolved.
+ * @returns: A generated structure, representing a native module.
+ * 
+ * @see: jerryx_native_module_register, jerryx_native_module_unregister
+ * 
+ * @note: `const` is used to indicate that this structure shall not change during runtime.
+ */
+#define JERRYX_NATIVE_MODULE_STRUCTURE(module_name, on_resolve_cb)    \
+  const jerryx_native_module_t _ ## module_name ## _definition =      \
+  {                                                                   \
+    .name_p = (jerry_char_t *) #module_name,                          \
+    .on_resolve_p = (on_resolve_cb),                                  \
+    .next_p = NULL                                                    \
+  };
+
+/**
+ * @brief  Generates the definition AND function body for the initial constructor.
+ * 
+ * @param module_name: Name of the module.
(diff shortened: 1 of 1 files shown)


jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -83,6 +83,12 @@ typedef struct jerryx_native_module_t
   {                                                                    \
     jerryx_native_module_unregister(&_##module_name##_definition);     \
   }
+  
+#define JERRYX_NATIVE_MODULE_EXTERN(module_name) \
+  extern jerryx_native_module_t _ ## module_name ## _definition
```
gabrielschulhof: IINM this is insufficient, because `_ ## module_name ## _definition` is declared static, so declaring it extern is not going to help find the symbol.

You may want something like
```C
#define JERRYX_NATIVE_MODULE_STRUCTURE(qualifier, module_name, on_resolve_cb) \
  qualifier jerryx_native_module_t _ ## module_name ## _definition =      \
  {                                                                    \
    .name_p = (jerry_char_t *) #module_name,                           \
    .on_resolve_p = (on_resolve_cb),                                   \
    .next_p = NULL                                                     \
};
```

and then start `JERRYX_NATIVE_MODULE_IMPLEM()` with
```C
  JERRYX_NATIVE_MODULE_STRUCTURE(static, module_name, (on_resolve_cb))
  ...
```

You'll probably want to store the rest of the definition of `JERRYX_NATIVE_MODULE_IMPLEM()` in its own macro, like `JERRYX_NATIVE_MODULE_ENTRY_POINTS()` and then define `JERRYX_NATIVE_MODULE_EXTERN()` as
```C
#define JERRYX_NATIVE_MODULE_EXTERN(module_name, on_resolve_cb) \
  JERRYX_NATIVE_MODULE_EXTERN_IMPLEM(module_name, (on_resolve_cb))
#define JERRYX_NATIVE_MODULE_EXTERN_IMPLEM(module_name, on_resolve_cb) \
  JERRYX_NATIVE_MODULE_STRUCTURE(, module_name, on_resolve_cb) \
  JERRYX_NATIVE_MODULE_ENTRY_POINTS(module_name)
```

IngwiePhoenix: I tested with the following:
```c
// top.h
extern int i;

// left.c
static int i = 0;

// right.c
int main() { return i; }
```

And as you might have expected, it did fail. And later, I learned why: https://stackoverflow.com/a/2841771/2423150

```
$ gcc left.c right.c -o middle
left.c:3:12: error: static declaration of 'i' follows non-static declaration
static int i = 0;
           ^
./top.h:1:12: note: previous declaration is here
extern int i;
           ^
1 error generated.
```

So using `static` in the context of this structure is very likely a bad idea... What do you think about using `const`, though?

IngwiePhoenix: @gabrielschulhof I made a second attempt at the macros. Hope this one is better! :)

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -59,31 +59,109 @@ typedef struct jerryx_native_module_t
 #define JERRYX_NATIVE_MODULE(module_name, on_resolve_cb)  \
   JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb)
 
-#define JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb)        \
-  static jerryx_native_module_t _ ## module_name ## _definition =      \
```
gabrielschulhof: `static` here needs to be passed in as a macro parameter, because `extern` will not be able to find a `static` across two compilation units. Then, if you want to declare a module reachable from another compilation unit, you must call `JERRYX_NATIVE_MODULE_IMPLEM()` without `static`.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -59,31 +59,109 @@ typedef struct jerryx_native_module_t
...
+
+/**
+ * @brief  Generates the definition AND function body for the initial constructor.
+ * 
+ * @param module_name: Name of the module.
+ * @returns: The generated, initial constructor.
+ * 
+ * @note: When compiled with ENABLE_INIT_FINI, you may - dependent on platform - enable
+ *        the initial constructuion when this module has been loaded as a shared object
+ *        or library. However, this does **not** apply to statically "compiled in" modules.
```
gabrielschulhof: Compiled-in modules can also have constructors. They will be executed before `main()`.

gabrielschulhof: Look at the tests for modules. Each test consists of a single executable, in which the modules auto-register on startup by way of the constructors. It even handles the case where constructors are disabled: https://github.com/pando-project/jerryscript/blob/master/tests/unit-ext/module/jerry-module-test.c#L192

IngwiePhoenix: Oh! I honestly wasn't exactly ware of how that macro worked, so I looked up the attributes and used what I found to describe it in the doc. But - good to know! I will fix that.

gabrielschulhof:
Exposing a structure as a global symbol is a bad idea. We had it in Node.js in 0.10 and we moved away from it. The only good ways to register a dynamically loaded addon are

1.
    1. `dlopen()`
    2. constructor calls public API and passes module definition into core
    3. core initializes module immediately or later, as needed
2.
    1. `dlopen()`
    2. core looks for well-known module initialization symbol inside the handle returned by `dlopen()` using `dlsym()`
    3. core calls the symbol immediately or later, as needed.

The first method works with both static and dynamic libraries and, as can be seen from the [test](https://github.com/pando-project/jerryscript/blob/master/tests/unit-ext/module/jerry-module-test.c#L205), can be worked around in the absence of library constructor support if the module is statically loaded.

The second method requires an implementation of `dlopen()` and `dlsym()` in order to work.

IngwiePhoenix:
I am sorry for my lack of activity here. I have been occupied with writing a lot of job applications and thus have not had a lot of time to actually work on OSS stuff…I am very sorry for that! You are free to close the PR, from my point of view. :)

PR #2882, opened 2019-05-21
Implement the Map/Set iterator operations

rerobika:


jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -614,12 +620,203 @@ ecma_op_container_delete (ecma_value_t this_arg, /**< this argument */
...
+    {
+      index = (uint32_t) (ecma_get_number_from_value (index_value) + 1);
+    }
+
+    ecma_value_t put_result = ecma_op_object_put (obj_p,
+                                                  prop_name_p,
+                                                  ecma_make_uint32_value (index),
+                                                  true);
+
+    JERRY_ASSERT (ecma_is_value_true (put_result));
```
LaszloLango: Are you sure an assert is enough here? It could be an error value as well which should be returned from the function.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -614,12 +620,203 @@ ecma_op_container_delete (ecma_value_t this_arg, /**< this argument */
...
+    }
+
+    ecma_free_value (value);
+    break;
+  }
+
+  ecma_free_values_collection (props_p, 0);
+
+  return ret_value;
+} /* ecma_op_container_iterator_next */
```
LaszloLango: I think the guards of this file are not correct. What happens if only `JERRY_ES2015_BUILTIN_SET` is enabled? What happens if `JERRY_ES2015_BUILTIN_MAP` and `JERRY_ES2015_BUILTIN_SET` are enabled, but `JERRY_ES2015_BUILTIN_ITERATOR` is disabled?

rerobika: `JERRY_ES2015_BUILTIN_SET` requires the `JERRY_ES2015_BUILTIN_MAP` so it cannot cause any problem. However, your second thought is currently a valid problem I'll check it.

jerry-core/ecma/operations/ecma-container-object.c
rerobika: Good catch!

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -614,15 +647,219 @@ ecma_op_container_delete (ecma_value_t this_arg, /**< this argument */
...
+
+      if (ecma_value_p == NULL)
+      {
+        ret_value = ecma_create_iter_result_object (ECMA_VALUE_UNDEFINED, ECMA_VALUE_TRUE);
+      }
+      continue;
+    }
+
+    ecma_value_t value = ecma_copy_value (ECMA_PROPERTY_VALUE_PTR (property_p)->value);
+    ecma_value_t key_arg;
```
galpeter: If I read this correctly the `key_arg` is used only if the `is_set` is false. Would it be possible to only "calculate" the key_arg at the required places? Maybe move `key_arg` "load" into a method?

galpeter: In addition this same logic seems to be already used in this file.

jerry-core/ecma/operations/ecma-container-object.c
rerobika: Fixed. Now the set is independent from the map builtin, also the corresponding iterator related routines are guarded correctly.

jerry-core/ecma/operations/ecma-container-object.c
rerobika: Thanks, I've applied your suggestion.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -614,15 +629,216 @@ ecma_op_container_delete (ecma_value_t this_arg, /**< this argument */
...
+    }
+
+    ecma_value_t put_result = ecma_op_object_put (obj_p,
+                                                  prop_name_p,
+                                                  ecma_make_uint32_value (index),
+                                                  true);
+
+    ecma_free_value (index_value);
+
+    if (!ecma_is_value_true (put_result))
```
dbatyai: Since this is an internal property, should this actually be able to throw an error/fail? I might be mistaken, but as far as I can tell internal slots should always be accessible. If that is the case, we should use something else instead of `ecma_op_object_put`.

rerobika: Internal properties behave like 'average' properties, only you cannot access it from the JS context. However, an object with an internal property can be freezed, which prevents the extension of the object, therefore IMO this operation can throw an error.

dbatyai: From what I can gather from the [standard](https://www.ecma-international.org/ecma-262/6.0/#sec-object-internal-methods-and-internal-slots): 

> Internal slots are not object properties and they are not inherited.

> ... internal slots are allocated as part of the process of creating an object and may not be dynamically added to an object.

Also Object.freeze seems to only affect property keys for the data and accessor properties of an object. I think it shouldn't have any impact on internal slots.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -207,21 +205,34 @@ ecma_op_container_create (const ecma_value_t *arguments_list_p, /**< arguments l
...
+  char *msg_p = "Expected a Map object.";
+
+#if ENABLED (JERRY_ES2015_BUILTIN_SET)
+  if (lit_id == LIT_MAGIC_STRING_SET_UL)
+  {
+    msg_p = "Expected a Set object.";
+  }
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_SET) */
+
+  ecma_raise_type_error (ECMA_ERR_MSG (msg_p));
```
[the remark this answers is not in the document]
rerobika: I think it would not be less complicated so I would leave as it if it's okay for you.

jerry-core/ecma/operations/ecma-container-object.c
rerobika: @dbatyai I updated the PR with a new approach, which uses lower level property management function. Therefore any kind property attribute settings cannot have an affect in this code path.

PR #2890, opened 2019-05-30
Fix rounding issue in ecma_utf8_string_to_number

### LINKED ISSUE #2802: ecma_utf8_string_to_number error for "2e23"
Function `ecma_utf8_string_to_number（"2e23",4） `  result is  200000000000000016777216
''. It is not correct.  The correct result is 199999999999999983222784

Yuyupo:
Fixes #2802


tests/jerry/arithmetics-bignums.js
```diff
@@ -23,8 +23,18 @@ assert(big == 2147483648); // overflow on 32bit numbers
...
 assert(big == 2147483649); // overflow on 32bit numbers
 
-assert ((1152921504606846976).toString() === "1152921504606847000")
+assert ((1152921504606846976).toString() === "1152921504606847000");
 
 assert (1.797693134862315808e+308 === Infinity);
 
 assert (9999999999999999 == 10000000000000000);
+
+assert(9007199254740993 == 9007199254740992);
```
dbatyai: These should also be checked as a string, similarly to line 26.

tests/jerry/arithmetics-bignums.js
```diff
@@ -23,8 +23,18 @@ assert(big == 2147483648); // overflow on 32bit numbers
...
+
+assert(9007199254740992 == 9007199254740992);
+
+assert(9007199254740994 == 9007199254740994);
+
```
dbatyai: I would also add asserts for the following:
```
9007199254740995 -> 9007199254740996
18014398509481989 -> 18014398509481988
18014398509481990 -> 18014398509481992
18014398509481991 -> 18014398509481992
18014398509481993 -> 18014398509481992
18014398509481994 -> 18014398509481996
```
I think these would cover all of the possible cases.

jerry-core/ecma/base/ecma-helpers-conversion.c
```diff
@@ -46,9 +46,28 @@ typedef struct
...
  */
-#define ECMA_UINT128_ROUND_HIGH_TO_UINT64(name) \
-  (name.hi + (name.lo >> 63u))
+static uint64_t
+ecma_round_high_to_uint64 (ecma_uint128_t *num_p)
+{
+  uint64_t masked_lo = num_p->lo & ~(1ULL << 63u);
+  uint64_t masked_hi = num_p->hi & 0x1;
+
+  if (num_p->lo >> 63u != 0)
```
dbatyai: Seems like all of these if's could be merged together into a single condition
```
if ((num_p->lo >> 63u != 0)
     && (masked_lo > 0 || masked_hi != 0))
```
What do you think?

Yuyupo: I totally overlooked that I could merge those. That's a neat idea indeed. :)

PR #2892, opened 2019-05-30
Throw error on deleting prop from non-object-coercible

### LINKED ISSUE #2891: Can Jerry delete a undefined object?
When running the following code, I find Jerry cannot output error information.

```
var f = function(name){
    delete this.keywords[name];
    print("Hello Jerry");
}
f("jerry");
```
Outputs: Hello Jerry
I think jerry does not execute the code "delete this.keywords[name];". I don't know why this happens

legendecas:
Fixes #2891
Ref: ecma-262#12.5.4.2.1

### DIFF
--- a/jerry-core/vm/opcodes.c
+++ b/jerry-core/vm/opcodes.c
@@ -150,11 +150,6 @@ vm_op_delete_prop (ecma_value_t object, /**< base object */
                    ecma_value_t property, /**< property name */
                    bool is_strict) /**< strict mode */
 {
-  if (ecma_is_value_undefined (object))
-  {
-    return ECMA_VALUE_TRUE;
-  }
-
   ecma_value_t check_coercible = ecma_op_check_object_coercible (object);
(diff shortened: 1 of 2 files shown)


jerry-core/vm/opcodes.c
```diff
@@ -150,11 +150,6 @@ vm_op_delete_prop (ecma_value_t object, /**< base object */
                    ecma_value_t property, /**< property name */
                    bool is_strict) /**< strict mode */
 {
-  if (ecma_is_value_undefined (object))
```
rerobika: I think it cannot be removed, however it can be guarded to be present only in ES2015 profile.

legendecas: I found that this code was used in the following cases:
```js
delete 'test';
delete 0;
```
And return value of those cases are `true` in both es5.1 and es2015.

I'm wondering the issue would requiring a bytecode generation minor refactoring. Working on it.

legendecas:
Closed for now.

PR #2898, opened 2019-06-05
Integrate LLVM's libfuzzer engine with JerryScript

renatahodovan:


LaszloLango:
Could you add some documentation about the usage of this feature?

renatahodovan:
@LaszloLango sure! Where should it be placed?
Note: as a result of this patch, a new binary named jerry-libfuzzer will be created, which has the cli interface provided by the underlying llvm version (documented here: https://llvm.org/docs/LibFuzzer.html). So our documentation should be at most a reference to the original description (IMHO).

LaszloLango:
I'd suggest to put it into `docs/01.GETTING-STARTED.md` if a reference is enough.

renatahodovan:
@LaszloLango Updated and rebased.

PR #2915, opened 2019-06-18
Improve JS object native pointer unit test

gabrielschulhof:
After setting a second native pointer, check that both the pointer that
was first set and the second pointer that was set are there.


rerobika:
@gabrielschulhof The patch looks good to me, but could you please force push it to trigger the Travis CI?

gabrielschulhof:
@rerobika done.

PR #2927, opened 2019-06-26
Rework usages/naming of configuration macros [part 3]

galpeter:
Reworked the JERRY_DEBUGGER macro to be a 0/1 switch.

### DIFF
--- a/jerry-core/config.h
+++ b/jerry-core/config.h
@@ -160,6 +160,17 @@
 #endif /* !defined (JERRY_CPOINTER_32_BIT) */
 
 /**
+ * Enable/Disable the engine's JavaScript debugger interface
+ *
+ * Allowed values:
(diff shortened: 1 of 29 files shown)


jerry-core/config.h
```diff
@@ -159,6 +159,17 @@
...
 
+/**
+ * Enable/Disable the engine's JavaScript debugger interface
+ *
+ * Allowed values:
+ *  0: Disable the debugger parts.
+ *  1: Enable the debugger.
+ */
+#ifndef JERRY_DEBUGGER
+# define JERRY_DEBUGGER 0
```
rerobika: After line 495 there are checks for the configuration macros. Could you please add it for this macro as well?

galpeter:

PR #2928, opened 2019-06-26
Fix implicit integer conversion reported by Clang 8

akosthekiss:
When building on Ubuntu 18.04.2 LTS (Linux 4.15.0 x86_64) with Clang 8, the compilation fails:

```sh
$ CC=clang-8 ./tools/build.py --clean --debug --profile=es2015-subset --lto=off
jerry-core/ecma/operations/ecma-container-object.c:72:45: error: implicit conversion loses integer precision: 'lit_magic_string_id_t' to 'uint16_t'
      (aka 'unsigned short') [-Werror,-Wimplicit-int-conversion]
  map_obj_p->header.u.class_prop.class_id = lit_id;
                                          ~ ^~~~~~
jerry-core/ecma/operations/ecma-container-object.c:672:42: error: implicit conversion loses integer precision: 'ecma_pseudo_array_type_t' to 'uint8_t'
      (aka 'unsigned char') [-Werror,-Wimplicit-int-conversion]
                                         iterator_type,
                                         ^~~~~~~~~~~~~
```

This commit fixes the issue by adding explicit conversions. After the change, the build succeeds.

### DIFF
--- a/jerry-core/ecma/operations/ecma-container-object.c
+++ b/jerry-core/ecma/operations/ecma-container-object.c
@@ -69,7 +69,7 @@ ecma_op_container_create (const ecma_value_t *arguments_list_p, /**< arguments l
                                                 ECMA_OBJECT_TYPE_CLASS);
 
   ecma_map_object_t *map_obj_p = (ecma_map_object_t *) object_p;
-  map_obj_p->header.u.class_prop.class_id = lit_id;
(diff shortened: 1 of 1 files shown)


galpeter:
lgtm. In the future imho we would need to add a static assert check that both type's values can fit into its given type (uint16/uint8) range.

PR #2932, opened 2019-06-28
Make API unavailable while processing native free callbacks

### LINKED ISSUE #2889: Question about native pointer.
I used the object's properties and call object's method in native pointer `free_cb` callback, I found that there might be a memory leak or crash. test code as following, the native pointer `free_cb` will call `finalize()` method. 

Whether the native pointer does not allow use this object method or any property?

``` javascript
class Test {
	constructor() {
		this.a = 1;
	}
	
	finalize() {
		console.log('foo');
		this.a = 0;
	}
}

var t = new Test();
gc(true);

var t = new Test();
gc(true);

var t = new Test();
gc(true);

var t = new Test();
gc(true);
```

rerobika:
This patch fixes #2889 and extends the API documentation of the native free callbacks.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -5179,7 +5179,8 @@ You can get them by calling jerry_get_object_native_pointer later.
 
 *Note*: If a non-NULL free callback is specified in the native type information,
         it will be called by the garbage collector when the object is freed.
-        The type info is always overwrites the previous value, so passing
+        This callback **must not** invoke API functions.
+        The type info always overwrites the previous value, so passing
         a NULL value deletes the current type info.
 
 **Prototype**
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -527,6 +527,10 @@ ecma_gc_free_native_pointer (ecma_property_t *property_p) /**< property */
   native_pointer_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_native_pointer_t,
                                                       value_p->value);
 
+#ifndef JERRY_NDEBUG
+  JERRY_CONTEXT (status_flags) &= (uint32_t) ~ECMA_STATUS_API_AVAILABLE;
+#endif /* !JERRY_NDEBUG */
+
   while (native_pointer_p != NULL)
   {
     if (native_pointer_p->info_p != NULL)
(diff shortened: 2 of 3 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -5178,7 +5178,8 @@ You can get them by calling jerry_get_object_native_pointer later.
 *Note*: If native pointer was already set for the object, its value is updated.
 
 *Note*: If a non-NULL free callback is specified in the native type information,
-        it will be called by the garbage collector when the object is freed.
+        it will be called by the garbage collector when the object is freed and
+        this callback **can not** invoke API functions.
```
akosthekiss: Some remarks:
- Appending this note to the existing sentence with an "and" sounds a bit weird. I'd suggest to simply start a new sentence. "when the object is freed. This callback"
- "can not" -> cannot (alternatively: "must not"?!)
- The following unrelated sentence has a glitch. It *is* unrelated but perhaps couldn't hurt fixing now? "type info ~~is~~ always overwrites"

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -545,6 +547,8 @@ ecma_gc_free_native_pointer (ecma_property_t *property_p) /**< property */
 
     native_pointer_p = next_p;
   }
+
+  JERRY_CONTEXT (status_flags) |= ECMA_STATUS_API_AVAILABLE;
```
akosthekiss: These are equivalents of the `jerry_make_api_unavailable` and `jerry_make_api_available` static inline functions of jerry.c. But there, they are only called from `jerry_cleanup` and `jerry_init`. Here, they are called on every invocation of `ecma_gc_free_native_pointer`. Wont be that a performance hit?

(Note: that whole flag is strange. It is unconditionally set/cleared by `jerry_init` and `jerry_cleanup` but only checked in debug asserts. Seems like setting/clearing it should also happen in debug builds only. And then the preformance hit question would also go away.)

rerobika: I support the idea to toggle it only in debug builds. However it still does not solve the possible release build crashes, but IMO after a crash with a debug build the issue can be figured out easily.

akosthekiss: Note that toggling the flag in release builds would not rule out release crashes either as nobody would check the flag (because the checks are debug-only asserts).

docs/02.API-REFERENCE.md
```diff
@@ -5179,7 +5179,8 @@ You can get them by calling jerry_get_object_native_pointer later.
 
 *Note*: If a non-NULL free callback is specified in the native type information,
         it will be called by the garbage collector when the object is freed.
-        The type info is always overwrites the previous value, so passing
+        This callback **must not** invoke API functions.
+        The type info always overwrites the previous value, so passing
```
akosthekiss: Just spotted: the doc comment of `jerry_set_object_native_pointer` in jerry.c is not updated accordingly.

rerobika: Thanks, I've forgotten it.

PR #2935, opened 2019-06-28
Limit the call stack size for native/builtin functions as well

### LINKED ISSUE #2905: The try-catch block is unable to catch the exception?
The try-catch block is unable to catch the error when I execute the code shows below, but throw  `ERR_REF_COUNT_LIMIT` error .
````
var func = function () {
	foo.prototype = new Array(1, 2, 3);
	function foo() {}
	var f = new foo();
	f.length = f;
	try {
		print("run to try");
		print("Using f will give an error: " + f);
		print("run to 2");
	} catch (e) {
		print("run to catch!")
	}
};
func();
````
The output is:
````
run to try
Error: ERR_REF_COUNT_LIMIT
````

rerobika:
VM_RECURSION_LIMIT only prevented the recursion of interpreted codeblocks but
non-interpeted codeblocks can also create stack overflow due to the too deep recursion.

This patch fixes #2905.

Co-authored-by: Gabor Loki loki@inf.u-szeged.hu


rerobika:
Note: The basics of this PR are the same, only the name changed and the check positions have been extended.

PR #2939, opened 2019-07-01
CBC_PUSH_THIS_LITERAL must be converted back to CBC_PUSH_THIS during emitting unary lvalue opcode

### LINKED ISSUE #2937: Expression c++ is not recognized correctly
### Testcase:
````
1. var foo = function () {
2. 	var c = 0;
3. 	var p1 = this[c];
4. 	print(p1);
5. 	var p2 = this[c++];
6. 	print(p2);
7. };
8. foo();
````
### Command:
````
./jerryscript/build/bin/jerry testcase.js
````
### Result:
````
undefined
Exception backtrace:
  0: 20190612/373821.js:5
  1: 20190612/373821.js:9
Script Error: ReferenceError: Undefined reference.
````
### Comment:
Both variable p1 and variable p2 should be `undefined`, but jerry throws an exception on line 5. Other more, other engines do not throw exception on line 5, such as v8, spiderMonkey, chakra, javascriptCore.

### PR DESCRIPTION
**Pull request by rerobika**
This patch fixes #2937.

Co-authored-by: Gabor Loki loki@inf.u-szeged.hu

### DIFF
--- a/tests/jerry/regression-test-issue-2937.js
+++ b/tests/jerry/regression-test-issue-2937.js
@@ -1,0 +1,30 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: tests/jerry/regression-test-issue-2937.js [hunk 1]
```diff
@@ -0,0 +1,31 @@
...
+assert (c === 0);
+
+var p3 = this[++c];
+assert (p3 === undefined);
+assert (c === 1);
+
+var p4 = this[--c];
+assert (p4 === undefined);
+assert (c === 0);
+
```
dbatyai: Please remove this trailing empty line.

rerobika: Thanks, I didn't notice that.

PR #2940, opened 2019-07-01
Add issue template for bug reports

akosthekiss:
Help both the community and ourselves by describing what
information is required in a bug report.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,77 @@
+---
+name: Bug report
+about: Create a report to help us improve
+
+---
+
+First of all, when reporting a bug, give the issue a descriptive title.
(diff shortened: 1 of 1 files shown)


akosthekiss:
I know that the template is a bit verbose, but hopefully not frighteningly long. I hope that it will save us from repeating the same info requirements in issue discussions over and over again by showing those requirements upfront.

rerobika:
@akosthekiss I highly support this idea. In my opinion the fact that the issue reporter *wants / is planning* to submit a PR for the reported issue not a rare thing. So I would suggest to include this "willingness to contribute" somehow to the template. What do you think about it?

akosthekiss:
According to our history, I think that issue reporters fixing the reported bug is actually a rare thing. (Feature requests might be statistically different, but this PR wants to add a template for bugs only.)

PR #2941, opened 2019-07-02
Include file path in Syntax error messages

dbatyai:
When using ES6 modules it was not possible to identify which module
an error originates from.

This PR changes the error message to also include the file path using
the file:line:column format, and updates the source context printing for
unhandled exceptions to use the correct file.


jerry-core/ecma/builtin-objects/ecma-builtin-function.c
```diff
@@ -160,7 +160,7 @@ ecma_builtin_function_dispatch_construct (const ecma_value_t *arguments_list_p,
   ECMA_STRING_TO_UTF8_STRING (function_body_str_p, function_body_buffer_p, function_body_buffer_size);
 
 #if ENABLED (JERRY_LINE_INFO)
-  JERRY_CONTEXT (resource_name) = ecma_make_magic_string_value (LIT_MAGIC_STRING__EMPTY);
```
rerobika: Is it intended that the `|| ENABLED (JERRY_ERROR_MESSAGES)` guard is missing from here?

jerry-core/api/jerry.c
```diff
@@ -396,10 +396,17 @@ jerry_parse (const jerry_char_t *resource_name_p, /**< resource name (usually a
 #if ENABLED (JERRY_PARSER)
   jerry_assert_api_available ();
 
-#if ENABLED (JERRY_LINE_INFO)
-  JERRY_CONTEXT (resource_name) = ecma_find_or_create_literal_string (resource_name_p,
-                                                                      (lit_utf8_size_t) resource_name_length);
-#endif /* ENABLED (JERRY_LINE_INFO) */
+#if ENABLED (JERRY_LINE_INFO) || ENABLED (JERRY_ERROR_MESSAGES)
+  if (!resource_name_length)
```
rerobika: Probably it's only my personal preference, but could you please use `resource_name_length == 0` instead?

jerry-core/ecma/builtin-objects/ecma-builtin-function.c
dbatyai: Nope, I forgot to add it. Thanks.

jerry-core/api/jerry.c
dbatyai: Sure.

jerry-core/jcontext/jcontext.h
```diff
@@ -191,7 +191,7 @@ struct jerry_context_t
   uint8_t debugger_max_receive_size; /**< maximum amount of data that can be received */
 #endif /* ENABLED (JERRY_DEBUGGER) */
 
-#if ENABLED (JERRY_LINE_INFO)
+#if ENABLED (JERRY_LINE_INFO) || ENABLED (JERRY_ERROR_MESSAGES)
   ecma_value_t resource_name; /**< resource name (usually a file name) */
 #endif /* ENABLED (JERRY_LINE_INFO) */
```
galpeter: The comment for `#endif` should also be updated.

targets/tizenrt-artik053/apps/jerryscript/jerry_main.c
```diff
@@ -155,40 +154,46 @@ print_unhandled_exception (jerry_value_t error_value, /**< error value */
...
+
+          while (i < sz && err_str_buf[i] != ':')
           {
-            break; /* wrong position info format */
+            num_str[j++] = (char) err_str_buf[i++];
           }
+          num_str[j] = '\0';
 
-          i += 10;
+          err_line = (unsigned int) strtol (num_str, NULL, 10);
```
galpeter: As the strtol is used, we could leverage it to in every case where the str->number conversion is required. With it we would not need to manually copy the chars into the num_str array, just point the start of the string where the number should be and it will return the the end pointer where the value is not convertible. This would reduce a bit of code here.

tests/unit-core/test-api.c
```diff
@@ -827,7 +827,7 @@ main (void)
     jerry_release_value (err_str_val);
     jerry_release_value (parsed_code_val);
     TEST_ASSERT (!strcmp ((char *) err_str_buf,
-                          "SyntaxError: Primary expression expected. [line: 2, column: 10]"));
+                          "SyntaxError: Primary expression expected. [<anonymous>:2:10]"));
```
galpeter: A test for the normal case ( having a file path ) would be good.

rerobika:
@dbatyai Please rebase the PR.

PR #2956, opened 2019-07-09
Add extra details for the api example page

galpeter:
Add simple examples and details on how to execute them.

### DIFF
--- a/docs/03.API-EXAMPLE.md
+++ b/docs/03.API-EXAMPLE.md
@@ -1,8 +1,43 @@
 JerryScript Engine can be embedded into any application, providing the way to run JavaScript in a large range of environments - from desktops to low-memory microcontrollers.
 
-This guide is intended to introduce you to JerryScript embedding API through creation of simple JavaScript shell.
+This guide is intended to introduce you to JerryScript embedding API and to create a minimal JavaScript shell.
+The examples are not using all API methods please also check out the API reference document which contains additional examples.
 
-## Step 1. Execute JavaScript from your application
+
+## Before trying out the examples: Get and build JerryScript library
+
+Before getting started using the JerryScript library it should be cloned and built for a target os/device.
+
+There are quite a few configuration options but for these examples the JerryScript is built
+**with default configuration** and installed to a user directory on a Linux system.
+This is done by the following commands:
+
+```sh
+$ mkdir jerry
+$ cd jerry
+$ git clone https://github.com/jerryscript-project/jerryscript.git
+$ jerryscript/tools/build.py --builddir=$(pwd)/example_build --cmake-param="-DCMAKE_INSTALL_PREFIX=$(pwd)/example_install/"
+$ make -C $(pwd)/example_build install
+```
+
+With this the JerryScript library is installed into the `$(pwd)/example_install/{include,lib}` directories.
+
+In this guide we will use `pkg-config` to ease the usage of headers and libraries.
+In order to do so, the following export is required (executed from the jerry directory):
+
+```sh
+$ export PKG_CONFIG_PATH=$(pwd)/example_install/lib/pkgconfig/
+```
+
+Test if the `pkg-config` works for JerryScript:
+
+```sh
+$ pkg-config --cflags --libs libjerry-core libjerry-port-default libjerry-ext libjerry-libm
+```
+
+## Example 1. Execute JavaScript from your application
+
+The most basic example to test the engine is to create an `api-example-1.c` file containing the following code:
 
 [doctest]: # ()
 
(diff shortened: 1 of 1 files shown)


docs/03.API-EXAMPLE.md
```diff
@@ -147,8 +503,8 @@ main (void)
   jerry_value_t global_object = jerry_get_global_object ();
 
   /* Constructing strings */
-  jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) "s");
-  jerry_value_t prop_value = jerry_create_string (str);
+  jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) "my_var");
+  jerry_value_t prop_value = jerry_create_string ((const jerry_char_t *) "Hello from C!");
 
   /* Setting the string value as a property of the Global object */
   jerry_release_value (jerry_set_property (global_object, prop_name, prop_value));
```
rerobika: `jerry_set_property` may returns with error, so I'd suggest to add the error check to the example in the same way, as you did above.

docs/03.API-EXAMPLE.md
```diff
@@ -175,16 +531,46 @@ main (void)
...
-JerryScript value can be a boolean, number, null, object, string or undefined. The value has an error flag,
-that indicates whether is an error or not. Every type has an error flag not only objects. The error flag should
-be cleared before the value is passed as an argument, otherwise it can lead to a type error. The error objects
-created by API functions has the error flag set.
+```js
+throw "message";
+```
 
-The following example function will output a JavaScript value:
+To check what type is a given `jerry_value_t` the `jerry_value_is_*` methods or the `jerry_value_get_type` could be used.
```
rerobika: Suggestion:
To check what type a given `jerry_value_t` is the ...

docs/03.API-EXAMPLE.md
```diff
@@ -175,16 +531,46 @@ main (void)
...
+For example the following code snippet could print out there types (not all type is checked):
```
rerobika: I think you meant `these types` instead.

galpeter:

docs/03.API-EXAMPLE.md
```diff
@@ -1,8 +1,42 @@
...
+$ mkdir jerry
+$ git clone https://github.com/jerryscript-project/jerryscript.git
+$ jerryscript/tools/build.py --builddir=$(pwd)/example_build --cmake-param="-DCMAKE_INSTALL_PREFIX=$(pwd)/example_install/"
+$ make -C $(pwd)/example_build install
+```
+
+With this the JerryScript library is installed into the `$(pwd)/example_install/{include,lib}` directories.
+
+In order to ease the usage of headers and libraries (archives) the `pkg-config` will be used and
+for it to detect the JerryScript files the following export is done (which is executed in the `jerry` directory):
```
dbatyai: How about: "In this guide we will use `pkg-config` to ease the usage of headers and libraries. In order to do so, the following export is required (executed from the `jerry` directory)"?

docs/03.API-EXAMPLE.md
```diff
@@ -20,39 +54,211 @@ main (void)
...
 #include "jerryscript.h"
-#include "jerryscript-ext/handler.h"
 
 int
 main (void)
 {
-  const jerry_char_t script[] = "print ('Hello, World!');";
+  const jerry_char_t script[] = "var str = 'Hello, World!';";
+  const jerry_length_t script_size = sizeof (script) - 1;
+  /* Note: sizeof only usable as the compiler knows the static character array's size! */
```
dbatyai: "sizeof can be used here only because the compiler knows the static character arrays's size. If this is not the case, strlen should be used instead."?

docs/03.API-EXAMPLE.md
```diff
@@ -20,39 +54,211 @@ main (void)
...
 
-  /* Register 'print' function from the extensions */
-  jerryx_handler_register_global ((const jerry_char_t *) "print",
-                                  jerryx_handler_print);
+  /* Run the demo script with 'eval' */
+  jerry_value_t eval_ret = jerry_eval (script,
+                                       script_size,
+                                       JERRY_PARSE_NO_OPTS);
+
+  /* Check if there was any error (syntax or runtime */
```
dbatyai: Missing closing paren

docs/03.API-EXAMPLE.md
```diff
@@ -20,39 +54,211 @@ main (void)
...
+$ ./api-example-3
+$ echo $?
+```
+
+## Example 4. Adding a C method for JavaScript
+
+The previous examples were not that eye catching as there were no visual output by the JavaScript code
+and C program.
+
+In this example a very simple "print" method which prints out a static string.
```
dbatyai: This sentence is missing a verb.

docs/03.API-EXAMPLE.md
```diff
@@ -72,46 +278,184 @@ main (void)
...
+$ ./api-example-4
+```
+
+## Example 5. Passing and processing arguments for native C code
+
+In the previous example the `print_handler` simply wrote a static string to the standard output.
+However in most cases this is not useful, ideally the method's argument(s) should be printed.
+
+In this example the `print_handler` is extended to convert the first
+argument (which propably comes from a JavaScript source) to a JS string and prints out it to the standard output.
```
dbatyai: propably -> probably, "prints out it" -> "prints it out"

docs/03.API-EXAMPLE.md
```diff
@@ -120,9 +464,27 @@ main (void)
...
+
+If everything is correct the application should print out the string passed for the `print` method in the JS code:
+
+```
+$ ./api-example-6
+```
+
+## Example 7. Interaction with JavaScript environment - adding a string property
+
+Previously a C method was registered for the global object, now this examples show how can one add a string
```
dbatyai: "this examples show how can one add" -> "this example shows how one can add"

docs/03.API-EXAMPLE.md
```diff
@@ -1,8 +1,42 @@
...
+## Before trying out the examples: Get and build JerryScript library
+
+Before getting started using the JerryScript library it should be cloned and built for a target os/device.
+
+There are quite a few configuration options but for these examples the JerryScript is built
+**with default configuration** and installed to a user directory on a Linux system.
+This is done by the following commands:
+
+```sh
+$ mkdir jerry
```
rerobika: I'm not sure but based on the build commands below I think you meant `mkdir jerry && cd jerry` here.

PR #2962, opened 2019-07-11
Optional arguments should advance the iterator in jerryx_arg_transform_optional

### LINKED ISSUE #2288: jerryx_arg_transform_optional does not advance iterator in case the value is undefined
I would expect in [this line] to pop the iterator if `js_arg` is `undefined`.
Otherwise there can only be a single optional argument. Another consequence is that  `jerryx_arg_transform_optional` does not work together with `jerryx_arg_transform_object_properties` because the iterator does not get advanced when an optional arg is `undefined`...

It looks like the unit test doesn't have either of these scenarios as a test case.

@jiangzidong do you agree this is a bug?

[this line]: https://github.com/jerryscript-project/jerryscript/blob/4e7a9d2d53804c1af7f7f7c57e2f6640898d0807/jerry-ext/arg/arg-transform-functions.c#L37

rerobika:
This patch fixes #2288

### DIFF
--- a/tests/unit-ext/test-ext-arg.c
+++ b/tests/unit-ext/test-ext-arg.c
@@ -35,6 +35,10 @@ static const jerry_char_t test_source[] = TEST_STRING_LITERAL (
   "arg2 = new Number(10.5);"
   "test_validator1(arg1, arg2, arg3);"
   "test_validator1(arg1, 10.5, 'abcdef');"
+  "test_validator3(arg1, arg1);"
+  "test_validator3(arg1);"
+  "test_validator3();"
+  "test_validator3(undefined, undefined);"
   "var obj_a = new MyObjectA();"
   "var obj_b = new MyObjectB();"
   "test_validator2.call(obj_a, 5);"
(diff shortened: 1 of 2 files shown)


tests/unit-ext/test-ext-arg.c
```diff
@@ -242,6 +245,52 @@ test_validator2_handler (const jerry_value_t func_obj_val __attribute__((unused)
...
+{
+
+  bool arg1 = false;
+
+  jerryx_arg_t mapping[] =
+  {
+    /* ignore this */
+    jerryx_arg_ignore (),
+    /* 1th argument should be boolean, and it is optional */
+    jerryx_arg_boolean (&arg1, JERRYX_ARG_COERCE, JERRYX_ARG_OPTIONAL),
```
galpeter: AFAIK, there should be multiple optional args, to see if multiple optional arguments can be handled.

galpeter: Oh and what if an "undefined" value is passed?

rerobika: Good catch! I added tests for it.

rerobika:
@galpeter Thanks for the review! I've updated the PR.

PR #2976, opened 2019-07-16
Resolve module paths relative to the current module

dbatyai:
The current module implementation resolves module paths relative to the
current working directory, but paths should be resolved relative to the
currently evaluated module/source.

This requires a change in the jerry_port_normalize_path port API
function, so that it also takes the current module path as an argument.
On the engine side, we now also create a module object for the main
script, so that we can properly identify the base path for other
modules.


jerry-core/include/jerryscript-port.h
```diff
@@ -225,11 +225,16 @@ void jerry_port_release_source (uint8_t *buffer_p);
  * @param in_path_p Input path as a zero terminated string.
  * @param out_buf_p Pointer to the output buffer where the normalized path should be written.
  * @param out_buf_size Size of the output buffer.
+ * @param base_file_p A file path that 'in_path_p' is relative to, usually the current module file.
+ *                    A NULL value represents that 'in_path_p' is relative to the current woring directory.
```
rerobika: Typo: woring

jerry-port/default/default-io.c
```diff
@@ -203,22 +206,58 @@ jerry_port_release_source (uint8_t *buffer_p) /**< buffer to free */
  * @return length of the path written to the output buffer
  */
 size_t
-jerry_port_normalize_path (const char *in_path_p, /**< input file path */
-                           char *out_buf_p,       /**< output buffer */
-                           size_t out_buf_size)   /**< size of output buffer */
+jerry_port_normalize_path (const char *in_path_p,   /**< input file path */
+                           char *out_buf_p,         /**< output buffer */
+                           size_t out_buf_size,     /**< size of output buffer */
+                           char *base_file_p) /**< base file path */
```
rerobika: The comment is misaligned to the other parameter's comments

Issue #2402, opened 2018-06-19
const keyword in JavaScript


BernardXiong:
Could JerryScript support const keyword?

for example, 
```javascript
const ctx = wx.createCanvasContext('myCanvas')

// Create linear gradient
const grd = ctx.createLinearGradient(0, 0, 200, 0)
grd.addColorStop(0, 'red')
grd.addColorStop(1, 'white')

// Fill with gradient
ctx.setFillStyle(grd)
ctx.fillRect(10, 10, 150, 80)
ctx.draw()
```

zherczeg:
The lexical environment rules for const and let makes it very hard to implement it in an Abstract-Syntax-Tree less environments. At least a somewhat efficient way. Until somebody came with a clever idea how to do this it will probably won't happen.

BernardXiong:
or just simply let 'const' as 'var'? Or add a read-only flag in var?

zherczeg:
Possible but will work in a different way that people expect.

BernardXiong:
Yes. But better than `parse failed`.

How about the specification of 'const' key word in JavaScript?

LaszloLango:
I agree with @zherczeg. Non-standard behavior is something we should avoid. People usually expects standard behavior. You can transcript your ES2015 code to ES5.1 with babel before you run JerryScript. It does what you want (changes 'const' to 'var').

BernardXiong:
OK. I see. It's the feature of ES2015/ES6. Thank you!

PR #2980, opened 2019-07-23
Parse let and const as var

Yuyupo:
Co-authored-by: Tamas Keri tkeri@inf.u-szeged.hu


Yuyupo:
This is not a feature just a compatibility thing, so for example one can just copy code from StackOverflow. They do work as a `var`.

dbatyai:
As @Yuyupo said, this is a temporary "solution" for allowing let/const in the source, until we provide a proper implementation for them.

Nevertheless, as you pointed out this can create different results than what people would expect, therefore I would argue that this should not be enabled by default. It must be entirely optional, and users must be conscious about using it.

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -386,13 +394,22 @@ parser_parse_var_statement (parser_context_t *context_p) /**< context */
       parser_parse_expression (context_p,
                                PARSE_EXPR_STATEMENT | PARSE_EXPR_NO_COMMA | PARSE_EXPR_HAS_LITERAL);
     }
+#if ENABLED (JERRY_ES2015_LET_CONST_AS_VAR)
+    else
+    {
+      if (is_const)
```
rerobika: Please merge this if statement into an else if.

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -2300,6 +2336,8 @@ parser_parse_statements (parser_context_t *context_p) /**< context */
         && context_p->token.type != LEXER_LEFT_BRACE
         && context_p->token.type != LEXER_RIGHT_BRACE
         && context_p->token.type != LEXER_KEYW_VAR
+        && context_p->token.type != LEXER_KEYW_LET
```
rerobika: Missing guards.

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -2319,6 +2357,8 @@ parser_parse_statements (parser_context_t *context_p) /**< context */
...
```
rerobika: Ditto.

jerry-core/parser/js/js-parser-statm.c
rerobika: Also, this check can be incorrect inside a for in/of loop. Therefore, some kind of information need to be passed to this function ( as function argument/parser context status flag) to represent whether this function has been invoked during the for in/of statement initializer part parsing or not.

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -334,9 +334,17 @@ parser_parse_enclosed_expr (parser_context_t *context_p) /**< context */
  * Parse var statement.
  */
 static void
-parser_parse_var_statement (parser_context_t *context_p) /**< context */
+parser_parse_declaration_statement (parser_context_t *context_p) /**< context */
```
yorkie: The original `parser_parse_var_statement` contains the initialization parsing with the keyword `var`, so the changed name seems a little bit confusion to me :)

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -1053,8 +1072,14 @@ parser_parse_for_statement_start (parser_context_t *context_p) /**< context */
     parser_save_range (context_p, &range, context_p->source_end_p);
     parser_set_range (context_p, &start_range);
     lexer_next_token (context_p);
-
-    if (context_p->token.type == LEXER_KEYW_VAR)
+#if ENABLED (JERRY_ES2015_LET_CONST_AS_VAR)
+    bool is_var_let_const = ((context_p->token.type == LEXER_KEYW_VAR)
+                               || (context_p->token.type == LEXER_KEYW_LET)
+                               || (context_p->token.type == LEXER_KEYW_CONST));
```
yorkie: How about creating a MACRO for this kind of check?

yorkie:
> Nevertheless, as you pointed out this can create different results than what people would expect, therefore I would argue that this should not be enabled by default. It must be entirely optional, and users must be conscious about using it.

The PR attempts to implement this feature in wrong way IMHO, optional feature doesn't mean an incompatible behavior with ECMA.

Block scoping seems better to implement in this PR, a simple way to get started is to make the `lex_env_p` of `vm_frame_ctx_t` as a stack, then at `const/let` declaration or initialization, it SET_POINTER to the top `lex_env_p`.

dbatyai:
Unfortunately it's not that simple. If we were to implement block scoping on top of the current parser architecture, it would cause a significant increase in memory usage. We will first need to make adjustments in the parser, so that we can provide a more efficient implementation.

So, it is not the intentions of this PR to pave the way for block scoping support, instead we would like to remedy previous requests for parsing let/const, even if it's only with var semantics (#2402). As we see, this is a fairly common practice among similar engines, and since it doesn't take a lot of effort to implement, we at least wanted to have the option to allow it.

Altough I agree that non-conformant behaviour is not preferable as it can cause confusion, allowing people to optionally enable a feature like this doesn't cause any harm, as long as they are aware of the consequences. If people feel like it would cause problems in their code, they can stick to using the default configuration, and everything will work as expected.
 
Lastly, I just wanted to state that we _are_ working on implementing block-scoping in the proper way, as it is something we would like to have.

yorkie:
> Unfortunately it's not that simple. If we were to implement block scoping on top of the current parser architecture, it would cause a significant increase in memory usage.

Hi @dbatyai could you please point out which parts will be increased significantly in memory? What I can figure it out is the lexical environments only :)

> Altough I agree that non-conformant behaviour is not preferable as it can cause confusion, allowing people to optionally enable a feature like this doesn't cause any harm, as long as they are aware of the consequences.

How about writing a document about this MACRO and its current behavior, due to a feature that's incompatible with ECMA standard would hart user in anyway.

> Lastly, I just wanted to state that we are working on implementing block-scoping in the proper way, as it is something we would like to have.

It sounds great that your team are working on this feature! BTW, the reason that I did that comment is I really think a wrong behavior is not a feature indeed, and we should not implement a "feature" that just for convenient copy & paste from StackOverflow just like @Yuyupo said.

PR #2981, opened 2019-07-23
Reduce the size of the jerry_context_t structure

rerobika:
This patch substitutes several global pointer with compressed pointers to reduce the size of the structure.


rerobika:
Note: The binary size is reduces ~120B without performance effect.

PR #2996, opened 2019-07-30
Correctly handle call arguments when using 'super' calls

### LINKED ISSUE #2990: BUG of super call with argument.
em as follow, `super.getid.call(this, a)` error:

```js
class A {
	constructor() {
		this.id = 10;
	}

	getid(a) {
		if (typeof a === 'number') {
			this.id += a;
		}
		return this.id;
	}
}

class B extends A {
	constructor() {
		super();
		this.id = 100;
	}

	getid(a) {
		// 1. error
		return super.getid.call(this, a);

               // 2. ok
		return super.getid.call(this);

		// 3. ok
		var fn = super.getid.bind(this);
		return fn(a);
	}
}

var s = new B();
console.log(s.getid(1));
```

galpeter:
There are two interconnected changes here:
* Added missing argument number handling for `super.propname(..)` calls.
* The `super.propname.call(...)` is does not need extra 'this' binding helper.

Fixes #2990.

### DIFF
--- a/tests/jerry/es2015/class-super-access-direct.js
+++ b/tests/jerry/es2015/class-super-access-direct.js
@@ -1,0 +1,102 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+class Base {
+  constructor () {
+    this.parent_value = 100;
+  }
+
+  parent_value () {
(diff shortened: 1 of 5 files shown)


tests/jerry/es2015/class-super-access-direct.js
```diff
@@ -0,0 +1,101 @@
...
+
+var obj = new Target();
+
+assert (obj.parent_value_direct() === -10);
+assert (obj.parent_value_direct_arg(1) === -9);
+assert (obj.parent_value_direct_arg(1, 2) === -7);
+assert (obj.parent_value_direct_arg(1, 2, 3) === -4);
+
+try {
+  obj.parent_value();
```
rerobika: Missing assert(false)

tests/jerry/es2015/class-super-access-indirect.js
```diff
@@ -0,0 +1,113 @@
...
+
+var obj = new Target();
+
+assert (obj.parent_value_indirect() === -10);
+assert (obj.parent_value_indirect_arg(1) === -9);
+assert (obj.parent_value_indirect_arg(1, 2) === -7);
+assert (obj.parent_value_indirect_arg(1, 2, 3) === -4);
+
+try {
+  obj.parent_value();
```
rerobika: Ditto.

PR #2998, opened 2019-08-01
Add realloc function to jmem

dbatyai:
This patch extends jmem functionality by adding a realloc function.
This opens up other paths of optimization which can result in smaller
peak memory usage and faster execution times, due to not having to
duplicate memory when we need to extend blocks.

### DIFF
--- a/jerry-core/jmem/jmem-heap.c
+++ b/jerry-core/jmem/jmem-heap.c
@@ -353,25 +353,16 @@ jmem_heap_alloc_block_null_on_error (const size_t size) /**< required memory siz
   return block_p;
 } /* jmem_heap_alloc_block_null_on_error */
 
-/**
- * Internal method for freeing a memory block.
- */
-void JERRY_ATTR_HOT
-jmem_heap_free_block_internal (void *ptr, /**< pointer to beginning of data space of the block */
-                               const size_t size) /**< size of allocated region */
-{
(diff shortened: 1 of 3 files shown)


LaszloLango:
This was previously proposed by @mmatyas in #2619
1) Please refer his name as well.
2) It did not have any gain back then. Why do you think this patch is an improvement? Any measurements or future plans with it?

dbatyai:
@LaszloLango, the pull request was opened by mmatyas, but the patch was supplied by me.

I don't really know where do you get the assumption that it didn't have any gain back then, but it doesn't matter. There are several builtins that are bottlenecked by ecma_concat_strings: String.replace, Array.join, JSON.stringify just to name a few. These produce exponentially worse runtimes with larger inputs, due to how ecma_concat_strings copies the string data.

Coincidentally, these are also the builtins methods that are not very well tested by sunspider, or at least by the few tests we were running at the time, so these issues don't show up, but they are there. If we take a look at the v8 benchmarks, we see the same trend, among other things, but I'd rather not get into the details.

Going a bit further, there are engine internals that could be improved, for example ecma_collections, and regexp bytecode reallocation. There are also plans for optimizing Arrays.

Overall, I believe this to be sufficient to justify this change.

LaszloLango:
@dbatyai thanks for clarifying this for me. I don't know when I will have time to review it in details, but I don't want to block this PR. If any reviewer gives a second approval then it could be merged.

PR #2999, opened 2019-08-01
Add string builder helper

dbatyai:
This change adds a new string builder type, that can be used to construct
strings internally by appending various types of other strings.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -1381,6 +1381,34 @@ typedef struct
 } ecma_long_string_t;
 
 /**
+ * String builder header
+ */
+typedef struct
+{
+  lit_utf8_size_t current_size; /**< size of the data in the buffer */
+} ecma_stringbuilder_header_t;
+
+/**
+ * Get pointer to the beginning of the stored string in the string builder
+ */
+#define ECMA_STRINGBUILDER_STRING_PTR(header_p) \
+  ((lit_utf8_byte_t *) (((ecma_string_t *) header_p) + 1))
+
+/**
+ * Get the size of the stored string in the string builder
+ */
+#define ECMA_STRINGBUILDER_STRING_SIZE(header_p) \
+  ((lit_utf8_size_t) (header_p->current_size - sizeof (ecma_string_t)))
+
+/**
+ * String builder handle
+ */
+typedef struct
+{
+  ecma_stringbuilder_header_t *header_p; /**< pointer to header */
+} ecma_stringbuilder_t;
+
+/**
  * Abort flag for error reference.
  */
 #define ECMA_ERROR_REF_ABORT 0x1
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -54,6 +54,8 @@ JERRY_STATIC_ASSERT ((int) ECMA_DIRECT_STRING_UINT == (int) ECMA_STRING_CONTAINE
 JERRY_STATIC_ASSERT (ECMA_PROPERTY_NAME_TYPE_SHIFT > ECMA_VALUE_SHIFT,
                      ecma_property_name_type_shift_must_be_greater_than_ecma_value_shift);
 
+JERRY_STATIC_ASSERT (sizeof (ecma_stringbuilder_header_t) <= sizeof (ecma_string_t),
+                     ecma_stringbuilder_header_must_not_be_larger_than_ecma_string);
 
 /**
  * Convert a string to an unsigned 32 bit value if possible
@@ -273,17 +275,18 @@ ecma_prop_name_is_map_key (ecma_string_t *string_p) /**< ecma-string */
 #endif /* ENABLED (JERRY_ES2015_BUILTIN_MAP) || ENABLED (JERRY_ES2015_BUILTIN_SET) */
 
 /**
- * Allocate new ecma-string and fill it with characters from the utf8 string
+ * Checks whether a string has a special representation, that is, the string is either a magic string,
+ * an external magic string, or an uint32 number, and creates an ecma string using the special representation,
+ * if available.
  *
- * @return pointer to ecma-string descriptor
+ * @return pointer to ecma string with the special representation
+ *         NULL, if there is no special representation for the string
  */
-ecma_string_t *
-ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 string */
-                                lit_utf8_size_t string_size) /**< string size */
+static ecma_string_t *
+ecma_find_special_string (const lit_utf8_byte_t *string_p, /**< utf8 string */
+                          lit_utf8_size_t string_size) /**< string size */
 {
   JERRY_ASSERT (string_p != NULL || string_size == 0);
-  JERRY_ASSERT (lit_is_valid_cesu8_string (string_p, string_size));
-
   lit_magic_string_id_t magic_string_id = lit_is_utf8_string_magic (string_p, string_size);
 
   if (magic_string_id != LIT_MAGIC_STRING__COUNT)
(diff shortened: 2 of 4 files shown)


dbatyai:
Depends on #2998

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2482,6 +2507,164 @@ ecma_string_trim (const ecma_string_t *string_p) /**< pointer to an ecma string
...
+
+  if (JERRY_LIKELY (string_size < UINT16_MAX))
+  {
+    string_p->refs_and_container = ECMA_STRING_CONTAINER_HEAP_UTF8_STRING | ECMA_STRING_REF_ONE;
+    string_p->u.utf8_string.size = (uint16_t) string_size;
+    string_p->u.utf8_string.length = (uint16_t) length;
+    string_p->hash = hash;
+
+#if ENABLED (JERRY_MEM_STATS)
+    jmem_stats_allocate_string_bytes (header_p->current_size);
```
rerobika: Are you sure that data is still valid? At this point the memory area pointed by `header_p` now contains the ecma string header part if I'm not mistaken.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1380,6 +1380,28 @@ typedef struct
...
+ */
+typedef struct
+{
+  lit_utf8_size_t current_size; /**< size of the data in the buffer */
+} ecma_stringbuilder_header_t;
+
+#define ECMA_STRINGBUILDER_STRING_PTR(header_p) \
+  ((lit_utf8_byte_t *) (((ecma_string_t *) header_p) + 1))
+
+#define ECMA_STRINGBUILDER_STRING_SIZE(header_p) \
```
rerobika: Missing doc.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1380,6 +1380,28 @@ typedef struct
...
+  ((lit_utf8_size_t) (header_p->current_size - sizeof (ecma_string_t)))
+
+/**
+ * String builder handle
+ */
+typedef struct
+{
+  ecma_stringbuilder_header_t *header_p;
```
rerobika: Ditto.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2482,6 +2507,164 @@ ecma_string_trim (const ecma_string_t *string_p) /**< pointer to an ecma string
...
+{
+  const lit_utf8_size_t initial_size = sizeof (ecma_string_t);
+  ecma_stringbuilder_header_t *buffer_p = (ecma_stringbuilder_header_t *) jmem_heap_alloc_block (initial_size);
+  buffer_p->current_size = initial_size;
+
+  ecma_stringbuilder_t ret;
+  ret.header_p = buffer_p;
+
+  return ret;
+} /* ecma_stringbuilder_create */
```
rerobika: I'd suggest to have a function to create a string builder with an ecma-string argument. Therefore one less reallocation is needed. It could be useful in array/typedarray builtin routines.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2482,6 +2507,185 @@ ecma_string_trim (const ecma_string_t *string_p) /**< pointer to an ecma string
...
+  const lit_utf8_size_t size = (lit_utf8_size_t) lit_char_get_utf8_length (c);
+  lit_utf8_byte_t *dest_p = ecma_stringbuilder_grow (builder_p, size);
+
+  lit_char_to_utf8_bytes (dest_p, c);
+} /* ecma_stringbuilder_append_char */
+
+/**
+ * Finalize a string builder
+ *
+ * @return the created string from the underlying buffer
```
galpeter: It would be good to document that after this do not use the `builder_p` as it is released.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2482,6 +2507,185 @@ ecma_string_trim (const ecma_string_t *string_p) /**< pointer to an ecma string
...
+
+#if ENABLED (JERRY_MEM_STATS)
+  jmem_stats_allocate_string_bytes (required_size);
+#endif /* ENABLED (JERRY_MEM_STATS) */
+
+  return string_p;
+} /* ecma_stringbuilder_finalize */
+
+/**
+ * Destroy a string builder
```
galpeter: When should this be called? Should it be called after calling finalize?

tests/unit-core/test-stringbuilder.c
```diff
@@ -0,0 +1,244 @@
...
+  {
+    static const lit_utf8_byte_t string_data[] = "abc";
+    ecma_string_t *uint_str_p = ecma_new_ecma_string_from_uint32 (234);
+
+    ecma_stringbuilder_t builder = ecma_stringbuilder_create ();
+    ecma_stringbuilder_append_char (&builder, LIT_CHAR_1);
+    ecma_stringbuilder_append_raw (&builder, string_data, sizeof (string_data) - 1);
+    ecma_stringbuilder_append (&builder, uint_str_p);
+    ecma_stringbuilder_append_magic (&builder, LIT_MAGIC_STRING_STRING);
+    ecma_stringbuilder_destroy (&builder);
```
galpeter: What does this case tests?

tests/unit-core/test-stringbuilder.c
```diff
@@ -0,0 +1,244 @@
...
+  }
+
+  {
+    static const lit_utf8_byte_t string_data[] = "abcdefghijklmnop";
+    size_t count = UINT16_MAX / (sizeof (string_data) - 1) + 1;
```
galpeter: This division seems a bit off: `65535/(16-1) + 1`? Wouldn't this be one? Also this count can be a `const`.

jerry-core/ecma/base/ecma-helpers-string.c
dbatyai: This can be used to free a builder that is no longer needed, for example when we encounter an error. I will update the comment to be a bit more descriptive.

tests/unit-core/test-stringbuilder.c
dbatyai: This is just to test that we do not leak.

tests/unit-core/test-stringbuilder.c
dbatyai: `sizeof (string_data)` is 17, becasue there's an implicit `\0` at the end, which results in 4096 + 1. This would be how many times we need to concatenate the string, so that we need to store it in a long utf8 string.

galpeter: I don't know why did I calculated this as one :)....
Thanks

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -273,22 +275,23 @@ ecma_prop_name_is_map_key (ecma_string_t *string_p) /**< ecma-string */
...
+ * @return true, if string has a special representation
+ *         false, otherwise
  */
-ecma_string_t *
-ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 string */
-                                lit_utf8_size_t string_size) /**< string size */
+static bool
+ecma_is_string_special (const lit_utf8_byte_t *string_p, /**< utf8 string */
+                        lit_utf8_size_t string_size, /**< string size */
+                        ecma_string_t **out_p) /**< [out] found special string */
```
galpeter: There should be an assert check for `out_p` imho, or a check before each `*out_p` assignment.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -273,22 +275,23 @@ ecma_prop_name_is_map_key (ecma_string_t *string_p) /**< ecma-string */
...
  *
- * @return pointer to ecma-string descriptor
+ * @return true, if string has a special representation
+ *         false, otherwise
  */
-ecma_string_t *
-ecma_new_ecma_string_from_utf8 (const lit_utf8_byte_t *string_p, /**< utf-8 string */
-                                lit_utf8_size_t string_size) /**< string size */
+static bool
+ecma_is_string_special (const lit_utf8_byte_t *string_p, /**< utf8 string */
```
galpeter: This method checks and returns the given special string, but it is not documented above. Also would it make sense to change the name to `ecma_try_string_special` so it'll indicate that this is not a simple "type checker"?

PR #3000, opened 2019-08-01
Update API version number

galpeter:


rerobika:
I think the expected result buffer in the snapshot related unittest file should be updated as well.

galpeter:
@rerobika the snapshot has its own version number which is not changed.

PR #3005, opened 2019-08-05
Add build option for changing the heap limit

dbatyai:
This change adds a build option that allows adjusting the garbage
collection heap usage limit, which can be used to fine-tune how often
garbage collection should be triggered.

### DIFF
--- a/jerry-core/config.h
+++ b/jerry-core/config.h
@@ -208,6 +208,15 @@
 #endif /* !defined (JERRY_GLOBAL_HEAP_SIZE) */
 
 /**
+ * The allowed heap usage limit until next garbage collection, in bytes.
+ *
+ * If value is 0, the default is 1/32 of JERRY_HEAP_SIZE
+ */
+#ifndef JERRY_GC_LIMIT
+# define JERRY_GC_LIMIT 0
+#endif /* !defined (JERRY_GC_LIMIT) */
+
+/**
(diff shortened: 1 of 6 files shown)


jerry-core/config.h
```diff
@@ -207,6 +207,15 @@
...
 #endif /* !defined (JERRY_GLOBAL_HEAP_SIZE) */
 
+/**
+ * The allowed heap usage limit until next garbage collection, in bytes.
+ *
+ * If value is 0, the default is 1/32 of JERRY_HEAP_SIZE
+ */
+#ifndef JERRY_HEAP_LIMIT
+# define JERRY_HEAP_LIMIT 0
+#endif /* !defined (JERRY_HEAP_SIZE) */
```
rerobika: `JERRY_HEAP_SIZE` -> `JERRY_HEAP_LIMIT`

LaszloLango:
Although this is a good improvement of the configuration options, IMHO `JERRY_HEAP_SIZE` and `JERRY_HEAP_LIMIT` are too similar and this is confusing. Maybe `JERRY_GC_HEAP_LIMIT` or `JERRY_GC_LIMIT` would be better.

dbatyai:
@LaszloLango, I agree, it can be confusing. I changed it to `JERRY_GC_LIMIT`.

PR #3006, opened 2019-08-05
Eliminate ECMA_TRY_CATCH macros part I.

rerobika:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -27,7 +27,6 @@
 #include "ecma-iterator-object.h"
 #include "ecma-objects.h"
 #include "ecma-string-object.h"
-#include "ecma-try-catch-macro.h"
 #include "jrt.h"
 
 #if ENABLED (JERRY_BUILTIN_ARRAY)
@@ -96,21 +95,23 @@ enum
  * Helper function to set an object's length property
  *
  * @return ecma value (return value of the [[Put]] method)
- *         Returned value must be freed with ecma_free_value.
+ *         Calling ecma_free_value on the returned value is optional if it is not abrupt completion.
  */
 static ecma_value_t
 ecma_builtin_array_prototype_helper_set_length (ecma_object_t *object, /**< object*/
                                                 ecma_number_t length) /**< new length */
 {
-  ecma_value_t ret_value;
-
   ecma_value_t length_value = ecma_make_number_value (length);
-  ret_value = ecma_op_object_put (object,
-                                  ecma_get_magic_string (LIT_MAGIC_STRING_LENGTH),
-                                  length_value,
-                                  true);
+  ecma_value_t ret_value = ecma_op_object_put (object,
+                                               ecma_get_magic_string (LIT_MAGIC_STRING_LENGTH),
+                                               length_value,
+                                               true);
 
   ecma_free_value (length_value);
+
+  JERRY_ASSERT (ecma_is_value_boolean (ret_value)
+                || ecma_is_value_empty (ret_value)
+                || ECMA_IS_VALUE_ERROR (ret_value));
   return ret_value;
 } /* ecma_builtin_array_prototype_helper_set_length */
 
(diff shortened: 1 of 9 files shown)


rerobika:
Note: the patch reduces the binary size by 592B on ARM without performance loss.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -245,40 +246,40 @@ ecma_builtin_array_prototype_object_concat (const ecma_value_t args[], /**< argu
...
+    JERRY_ASSERT (ecma_is_value_empty (concat_value));
   }
-  else
+
+  ecma_value_t set_length_value = ecma_builtin_array_prototype_helper_set_length (new_array_p,
+                                                                                  ((ecma_number_t) new_length));
+  if (ECMA_IS_VALUE_ERROR (set_length_value))
   {
-    ecma_free_value (new_array);
+    return set_length_value;
```
galpeter: Freeing the `new_array_p` in case of error is not required?

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -492,40 +492,37 @@ ecma_builtin_array_prototype_object_push (const ecma_value_t *argument_list_p, /
...
   {
-    /* 5.a */
-    ecma_value_t e_value = argument_list_p[index];
-
     /* 5.b */
     ecma_string_t *n_str_p = ecma_new_ecma_string_from_number (n);
 
-    ECMA_TRY_CATCH (put_value, ecma_op_object_put (obj_p, n_str_p, e_value, true), ret_value);
-    ECMA_FINALIZE (put_value);
+    ecma_value_t put_value = ecma_op_object_put (obj_p, n_str_p, argument_list_p[index], true);
```
galpeter: Can't we use `ecma_op_object_put_by_index` here?

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
rerobika: Thanks! Good catch.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
rerobika: `ecma_op_object_put_by_index` is designed to handle only uint32_t indices, but at this point `ecma_number_t n = (ecma_number_t) length;` can overflow in the UINT32_MAX limit.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
galpeter: Is this also means that there is no test for this? :)

rerobika: Indeed we do not have test for it. The reason is that to trigger this branch we need to achieve that the newly constructed array's length should be greater than UINT32_MAX. We can create this kind of test but the routine iterates through the whole property chain to find the current indexed property which takes serious amount of time. So IMO we can assume that this deref is necessary but I'd not enclose a test case for this corner case.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -642,59 +677,66 @@ ecma_builtin_array_prototype_object_shift (ecma_object_t *obj_p, /**< array obje
...
+    }
+
+    /* 7.b */
+    uint32_t to = k - 1;
 
     if (ecma_is_value_found (curr_value))
     {
       /* 7.d.i, 7.d.ii */
-      ECMA_TRY_CATCH (put_value, ecma_op_object_put (obj_p, to_str_p, curr_value, true), ret_value);
+      ecma_value_t put_value = ecma_op_object_put_by_index (obj_p, to, curr_value, true);
```
galpeter: The "put" and "delete" parts are really similar, would it be possible to refactor it like this?
```c
ecma_value_t new_value;

if (ecma_value_is_found (curr_value))
{
    new_value = ecma_op_object_put_by_index (obj_p, to, curr_value, true);
    ecma_free_value (curr_value);
}
else
{
    new_value = ecma_op_object_delete_by_index (obj_p, to, true);
}

if (ECMA_IS_VALUE_ERROR (new_value))
{
    ecma_free_value (first_value);
    return new_value;
}
```

I'm okay having it in a followup PR.

rerobika: Good idea, let me do it in a follow up patch.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1912,86 +1948,78 @@ ecma_builtin_array_reduce_from (ecma_value_t callbackfn, /**< routine's 1st argu
...
+    while (!k_present && index < len)
     {
       /* 8.b.i */
-      ecma_string_t *index_str_p = ecma_new_ecma_string_from_uint32 (start_from_left ? index
-                                                                                     : last_index - index);
       k_present = true;
 
       /* 8.b.ii-iii */
-      ECMA_TRY_CATCH (current_value, ecma_op_object_find (obj_p, index_str_p), ret_value);
+      ecma_value_t current_value = ecma_op_object_find_by_index (obj_p, start_from_left ? index : last_index - index);
```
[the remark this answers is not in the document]
rerobika: Good catch!

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1319,64 +1375,68 @@ ecma_builtin_array_prototype_object_unshift (const ecma_value_t args[], /**< arg
...
     ecma_string_t *to_str_p = ecma_new_ecma_string_from_number (new_idx);
 
-    /* 6.c */
-    ECMA_TRY_CATCH (get_value, ecma_op_object_find (obj_p, from_str_p), ret_value);
-
     if (ecma_is_value_found (get_value))
     {
       /* 6.d.i, 6.d.ii */
-      ECMA_TRY_CATCH (put_value, ecma_op_object_put (obj_p, to_str_p, get_value, true), ret_value);
+      ecma_value_t put_value = ecma_op_object_put (obj_p, to_str_p, get_value, true);
```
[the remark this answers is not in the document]
rerobika: Indeed. I've applied your suggestion.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1319,64 +1375,68 @@ ecma_builtin_array_prototype_object_unshift (const ecma_value_t args[], /**< arg
...
+        return put_value;
+      }
     }
     else
     {
       /* 6.e.i */
-      ECMA_TRY_CATCH (del_value, ecma_op_object_delete (obj_p, to_str_p, true), ret_value);
-      ECMA_FINALIZE (del_value);
-    }
+      ecma_value_t del_value = ecma_op_object_delete (obj_p, to_str_p, true);
```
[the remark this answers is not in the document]
rerobika: ditto.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
rerobika: @galpeter I've already updated the PR with this idea.

PR #3013, opened 2019-08-09
Fix building with build.py on Windows

ossy-szeged:
Changes:
  - gen-doctest.py: Use slashes in paths to make cmake happy.
  - unit-doc/CMakeLists.txt: Don't add invalid arguments to MSVC and call gen-doctest.py properly.
  - build.py: Build and install with cmake calls on Windows.

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -25,6 +25,10 @@ import sys
 import settings
 
 def default_toolchain():
+    # We don't have default toolchain on Windows and os.uname() isn't supported.
+    if sys.platform == 'win32':
+        return None
+
     (sysname, _, _, _, machine) = os.uname()
     toolchain = os.path.join(settings.PROJECT_DIR,
                              'cmake',
@@ -60,7 +64,7 @@ def get_arguments():
     buildgrp.add_argument('--compile-flag', metavar='OPT', action='append', default=[],
                           help='add custom compile flag')
     buildgrp.add_argument('--debug', action='store_const', const='Debug', dest='build_type',
-                          help='debug build')
+                          default='MinSizeRel', help='debug build')
     buildgrp.add_argument('--install', metavar='DIR', nargs='?', default=None, const=False,
                           help='install after build (default: don\'t install; '
                                'default directory if install: OS-specific)')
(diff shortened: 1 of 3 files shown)


tools/build.py
```diff
@@ -244,11 +248,19 @@ def configure_jerry(arguments):
 
     return subprocess.call(cmake_cmd)
 
-def make_jerry(arguments, target=None):
-    make_cmd = ['make', '--no-print-directory', '-j', str(arguments.jobs), '-C', arguments.builddir]
+def make_jerry(arguments):
+    if sys.platform == 'win32':
+        make_cmd = ['cmake', '--build', arguments.builddir, '--config', arguments.build_type]
```
dbatyai: Is the build type argument really needed? Shouldn't the cmake configure phase cache it?

tools/build.py
```diff
@@ -244,11 +248,19 @@ def configure_jerry(arguments):
...
+    else:
+        make_cmd = ['make', '--no-print-directory', '-j', str(arguments.jobs), '-C', arguments.builddir]
```
dbatyai: I'd say we should build/install with cmake calls on all platforms. What do you think?

tools/build.py
ossy-szeged: Yes, because vcproj files don't have default build type, you can build all build type with the same vcproj file. (if the only differences are compiler options and cmake project doesn't do any other trick)

tools/build.py
ossy-szeged: It sounds good, I'll check if this idea works on other platforms too. I think it should, but let's try it.

tools/build.py
```diff
@@ -244,11 +248,19 @@ def configure_jerry(arguments):
...
 
-    if target:
-        make_cmd.append(target)
+    return subprocess.call(make_cmd)
+
+def install_jerry(arguments):
+    if sys.platform == 'win32':
+        make_cmd = ['cmake', '--install', arguments.builddir, '--config', arguments.build_type]
```
kaadam: Hi Ossy, I've just bumped into that '--install' switch was introduced in Cmake 3.15, it doesn't exist in previous versions (See: https://cmake.org/cmake/help/v3.15/release/3.15.html). I used the VS2019's internal version of CMake (VS2019 was shipped with 3.14). In order to get the built project installed 'cmake ... --target install' could be used in this case.
Do you have any plan for supporting former versions of CMake? Of course, I know JerryScript on Windows is experimental currently, that's why it's just a question. :)

tools/build.py
dbatyai: I see.

tools/build.py
ossy-szeged: Done in the latest updated commit.

tools/build.py
ossy-szeged: I found a workaround, which doesn't need --install option. The latest commit already works with very old cmake too.

PR #3017, opened 2019-08-15
Don't run Coverity as the last task on Travis

dbatyai:
Follow-up after #3008

Having Coverity as the last task is fine for pull requests, but on the
master branch, where the task actually performs the scans, it can have a
large delay on the builds.

This change moves the Coverity task back to the middle ground, and puts
the ESP8266 build as the last, which seems to always run fast. This saves a
few extra minutes of build time on the master branch.


LaszloLango:
LGTM

PR #3026, opened 2019-08-27
Make run-tests --unittests work on Windows too

ossy-szeged:


tools/run-tests.py
```diff
@@ -329,6 +337,7 @@ def run_jerry_debugger_tests(options):
     for job in DEBUGGER_TEST_OPTIONS:
         ret_build, build_dir_path = create_binary(job, options)
         if ret_build:
+            print ("\n%sBuild failed%s\n" % (TERM_RED, TERM_NORMAL))
```
galpeter: If I remember correctly there should be no space before `(` character here.

tools/runners/run-unittests.py
```diff
@@ -0,0 +1,98 @@
...
+            if sys.platform != 'win32' or testfile.endswith(".exe"):
+                unittests.append(testfile)
+    unittests.sort()
+    return unittests
+
+
+def main(args):
+    unittests = get_unittests(args.path)
+    total = len(unittests)
+    if total == 0:
```
galpeter: In python there is no need for length check: just: `if not unittests:` an empty list is false by default.

tools/runners/run-unittests.py
```diff
@@ -0,0 +1,98 @@
...
+    passed = 0
+    failed = 0
+
+    runtime = os.environ.get('RUNTIME')
+    test_cmd = [runtime] if runtime else []
+
+    for test in unittests:
+        tested += 1
+        testpath = os.path.relpath(test)
+        proc = subprocess.Popen(test_cmd + [test],
```
galpeter: There is a convenience method: `subprocess.check_output` it will return the output and throws an exception in case the execution failed (the return code is not zero). Maybe we could use that here?

tools/runners/run-unittests.py
```diff
@@ -0,0 +1,98 @@
...
+    print("%sPASS: %d%s" % (TERM_GREEN, passed, TERM_NORMAL))
+    print("%sFAIL: %d%s\n" % (TERM_RED, failed, TERM_NORMAL))
+
+    success_color = TERM_GREEN if passed == total else TERM_RED
+    print("%sSuccess: %d%%%s" % (success_color, passed*100/total, TERM_NORMAL))
+
+    if failed > 0:
+        return 1
+
+    return 0
```
galpeter: Do we need the return values at all? In this code the return value is not used at all.

tools/run-tests.py
ossy-szeged: Good point, fixed.

tools/runners/run-unittests.py
ossy-szeged: The length of unittests are used later, that's why I get it here and used for this check too.

tools/runners/run-unittests.py
ossy-szeged: Good catch, we need it, but I missed to return this rc later. Fixed.

tools/runners/run-unittests.py
ossy-szeged: Already replaced with subprocess.check_output, it really looks better.

PR #302, opened 2015-07-03
Fix element indexing for Object.keys/Object.getOwnPropertyNames

galpeter:
Previously the index was incremented even if an internal property
or a non enumerable property was found.

JerryScript-DCO-1.0-Signed-off-by: Peter Gal pgal.u-szeged@partner.samsung.com

PR #3029, opened 2019-08-29
Make run-tests --test262 work on Windows too

ossy-szeged:


tools/runners/run-unittests.py
```diff
@@ -0,0 +1,98 @@
...
+    print("%sPASS: %d%s" % (TERM_GREEN, passed, TERM_NORMAL))
+    print("%sFAIL: %d%s\n" % (TERM_RED, failed, TERM_NORMAL))
+
+    success_color = TERM_GREEN if passed == total else TERM_RED
+    print("%sSuccess: %d%%%s" % (success_color, passed*100/total, TERM_NORMAL))
+
+    if failed > 0:
+        return 1
+
+    return 0
```
galpeter: Do we need the return values at all? In this code the return value is not used at all.

tools/runners/run-unittests.py
```diff
@@ -0,0 +1,98 @@
...
+            if sys.platform != 'win32' or testfile.endswith(".exe"):
+                unittests.append(testfile)
+    unittests.sort()
+    return unittests
+
+
+def main(args):
+    unittests = get_unittests(args.path)
+    total = len(unittests)
+    if total == 0:
```
galpeter: In python there is no need for length check: just: `if not unittests:`  an empty list is false by default.

tools/runners/run-test-suite-test262.py
```diff
@@ -0,0 +1,96 @@
...
+    if os.path.isdir(path):
+        shutil.rmtree(path)
+
+    proc = subprocess.Popen(get_platform_cmd_prefix() +
+                            [os.path.join(path_to_test262, 'tools/packaging/test262.py'),
+                             '--command', (runtime + ' ' + engine).strip(),
+                             '--tests', path_to_test262,
+                             '--summary'],
+                            stdout=subprocess.PIPE)
+    output_file = open(os.path.join(os.path.dirname(engine), 'test262.report'), 'w+')
```
galpeter: * When opening files please use the `with open(..) as fp:` construct to make sure that the file is always closed.
* Do we need the `w+` flag? Isn't the `w` enough?

tools/runners/run-test-suite-test262.py
```diff
@@ -0,0 +1,96 @@
...
+    if not os.path.isdir(os.path.join(path_to_test262, '.git')):
+        print (path_to_test262)
+        proc = subprocess.Popen(['git', 'clone', 'https://github.com/tc39/test262.git',
+                                 '-b', 'es5-tests', path_to_test262])
+        proc.wait()
+        if proc.returncode:
+            print('Cloning test262 repository failed.')
+            return proc.returncode
+
+    path = os.path.join(path_to_test262, 'test/suite/bestPractice')
```
galpeter: If we are already using `os.path.join` then we should use it all the way:
`os.path.join(path_to_test262, 'test', 'suite', 'bestPractice')`

In addition the `path` variable name does not say anything. It should be a bit more descriptive.

tools/runners/run-unittests.py
ossy-szeged: Already fixed in #3026. That commit is the part of #302, I just included it to be able to test test262 change on CI.

tools/runners/run-unittests.py
ossy-szeged: Handled in #3026. That commit is the part of #302, I just included it to be able to test test262 change on CI.

tools/runners/run-test-suite-test262.py
ossy-szeged: - with open: Done.
- w+: no, w is enough. Fixed.

tools/runners/run-test-suite-test262.py
ossy-szeged: Fixed.

PR #3033, opened 2019-08-30
Skip fast access arrays during hashmap frees

rerobika:
Also this patch fixes some small issues during the fast array construct process.

### DIFF
--- a/jerry-core/ecma/operations/ecma-array-object.c
+++ b/jerry-core/ecma/operations/ecma-array-object.c
@@ -76,6 +76,19 @@ ecma_op_new_array_object (ecma_length_t length) /**< length of the new array */
 ecma_object_t *
 ecma_op_new_fast_array_object (ecma_length_t length) /**< length of the new fast access mode array */
 {
+  const uint32_t aligned_length = ECMA_FAST_ARRAY_ALIGN_LENGTH (length);
+  ecma_value_t *values_p = NULL;
+
+  if (length != 0)
+  {
+    values_p = (ecma_value_t *) jmem_heap_alloc_block_null_on_error (aligned_length * sizeof (ecma_value_t));
+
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -76,15 +76,12 @@ ecma_op_new_array_object (ecma_length_t length) /**< length of the new array */
 ecma_object_t *
 ecma_op_new_fast_array_object (ecma_length_t length) /**< length of the new fast access mode array */
 {
-  ecma_object_t *object_p = ecma_op_new_array_object (length);
```
galpeter: Would it make sense to reorder it like this?:
```c
const uint32_t aligned_length = ECMA_FAST_ARRAY_ALIGN_LENGTH (length);
ecma_value_t *values_p = NULL;

if (length != 0)
{
  values_p = (ecma_value_t *) jmem_heap_alloc_block_null_on_error (aligned_length * sizeof (ecma_value_t));

  if (JERRY_UNLIKELY (values_p == NULL))
  {
    return NULL;
  }
}

ecma_object_t *object_p = ecma_op_new_array_object (length);

ecma_extended_object_t *ext_obj_p = (ecma_extended_object_t *) object_p;
ext_obj_p->u.array.is_fast_mode = true;

JERRY_ASSERT (object_p->u1.property_list_cp == JMEM_CP_NULL);

for (uint32_t i = 0; i < aligned_length; i++)
{
  values_p[i] = ECMA_VALUE_ARRAY_HOLE;
}

ECMA_SET_POINTER (object_p->u1.property_list_cp, values_p);
return object_p;
```

This way the object construction and flag configuration are written down once

rerobika: Good idea!

PR #3037, opened 2019-09-02
Bump reference platform to Ubuntu 18.04 LTS

ossy-szeged:
Ubuntu 14.04 reached its end of life on April 30m 2019.
Let's bump the reference to the latest LTS, which is 18.04.

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-value.c
+++ b/jerry-core/ecma/base/ecma-helpers-value.c
@@ -40,6 +40,7 @@ JERRY_STATIC_ASSERT (sizeof (jmem_cpointer_t) <= sizeof (jmem_cpointer_tag_t),
 
 #ifdef ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY
 
+/* cppcheck-suppress zerodiv */
 JERRY_STATIC_ASSERT (sizeof (uintptr_t) <= sizeof (ecma_value_t),
                      uintptr_t_must_fit_in_ecma_value_t);
 
--- a/jerry-core/jmem/jmem-heap.c
+++ b/jerry-core/jmem/jmem-heap.c
@@ -288,6 +288,7 @@ jmem_heap_gc_and_alloc_block (const size_t size, /**< required memory size */
 
   void *data_space_p = jmem_heap_alloc (size);
 
+  /* cppcheck-suppress memleak */
   while (JERRY_UNLIKELY (data_space_p == NULL) && JERRY_LIKELY (pressure < max_pressure))
   {
     pressure++;
--- a/jerry-ext/handler/handler-register.c
+++ b/jerry-ext/handler/handler-register.c
@@ -63,7 +63,13 @@ jerryx_set_properties (const jerry_value_t target_object, /**< target object */
 {
 #define JERRYX_SET_PROPERTIES_RESULT(VALUE, IDX) ((jerryx_register_result) { VALUE, IDX })
   uint32_t idx = 0;
-  for (; ((entries + idx) != NULL) && (entries[idx].name != NULL); idx++)
+
+  if (entries == NULL)
+  {
+    return JERRYX_SET_PROPERTIES_RESULT (jerry_create_undefined (), 0);
+  }
+
+  for (; (entries[idx].name != NULL); idx++)
   {
     const jerryx_property_entry *entry = &entries[idx];
 
@@ -99,9 +105,12 @@ void
 jerryx_release_property_entry (const jerryx_property_entry entries[], /**< list of property entries */
                                const jerryx_register_result register_result) /**< previous result of registration */
 {
-  for (uint32_t idx = register_result.registered;
-       ((entries + idx) != NULL) && (entries[idx].name != NULL);
-       idx++)
+  if (entries == NULL)
+  {
+    return;
+  }
+
+  for (uint32_t idx = register_result.registered; entries[idx].name != NULL; idx++)
   {
     jerry_release_value (entries[idx].value);
   }
--- a/jerry-libm/atan2.c
+++ b/jerry-libm/atan2.c
@@ -82,7 +82,7 @@ atan2 (double y, double x)
   {
     return atan (y);
   }
-  m = ((hy >> 31) & 1) | ((hx >> 30) & 2); /* 2 * sign(x) + sign(y) */
+  m = ((hy < 0) ? 1 : 0) + ((hx < 0) ? 2 : 0); /* 2 * sign(x) + sign(y) */
 
   /* when y = 0 */
   if ((iy | ly) == 0)
--- a/tests/unit-core/test-arraybuffer.c
+++ b/tests/unit-core/test-arraybuffer.c
@@ -304,6 +304,7 @@ main (void)
   /* Test ArrayBuffer external memory map/unmap */
   {
     const uint32_t buffer_size = 20;
+    /* cppcheck-suppress variableScope */
     JERRY_VLA (uint8_t, buffer_p, buffer_size);
     {
       jerry_value_t input_buffer = jerry_create_arraybuffer_external (buffer_size, buffer_p, NULL);
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,8 @@
 language: c
 
-# Default environment: Ubuntu Trusty 14.04.
+# Default environment: Ubuntu Bionic 18.04.
 os: linux
-dist: trusty
+dist: bionic
 
 # Default job task: run tests as defined in the $OPT environment variable.
 # Jobs can redefine the 'script' stage in the matrix below.
@@ -12,13 +12,12 @@ script: tools/run-tests.py $OPTS
 matrix:
   include:
     - name: "Checks"
-      install: pip install --user pylint==1.6.5
       script:
         - tools/run-tests.py --check-signed-off=travis --check-doxygen --check-vera --check-license --check-magic-strings --check-pylint
         - travis_wait 40 tools/run-tests.py --check-cppcheck
       addons:
         apt:
-          packages: [doxygen, cppcheck, vera++]
+          packages: [doxygen, cppcheck, vera++, pylint]
 
     - name: "Linux/x86-64 Build, Correctness & Debugger Tests"
       env:
@@ -70,22 +69,20 @@ matrix:
         - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list=parser-oom.js,parser-oom2.js,stack-limit.js,regression-test-issue-2190.js,regression-test-issue-2258-2963.js,regression-test-issue-2448.js,regression-test-issue-2905.js --buildoptions=--stack-limit=0,--compile-flag=-fsanitize=address,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--compile-flag=-O2,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
         - ASAN_OPTIONS=detect_stack_use_after_return=1:check_initialization_order=true:strict_init_order=true
         - TIMEOUT=600
-      compiler: gcc-5
       addons:
         apt:
           sources: ubuntu-toolchain-r-test
-          packages: [gcc-5, gcc-5-multilib]
+          packages: [gcc-multilib]
 
     - name: "UBSAN Tests"
       env:
         - OPTS="--quiet --jerry-tests --jerry-test-suite --skip-list=parser-oom.js,parser-oom2.js --buildoptions=--compile-flag=-fsanitize=undefined,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold"
         - UBSAN_OPTIONS=print_stacktrace=1
         - TIMEOUT=600
-      compiler: gcc-5
       addons:
         apt:
           sources: ubuntu-toolchain-r-test
-          packages: [gcc-5, gcc-5-multilib]
+          packages: [gcc-multilib]
 
     - name: "Coverity Scan & SonarQube"
       env:
--- a/docs/00.GETTING-STARTED.md
+++ b/docs/00.GETTING-STARTED.md
@@ -1,6 +1,6 @@
 ## Setting up prerequisites
 
-Currently, only Ubuntu 14.04+ is officially supported as primary development environment.
+Currently, only Ubuntu 18.04+ is officially supported as primary development environment.
 
 There are several dependencies, that should be installed manually. The following list is the absolute minimum for building:
 
--- a/jerry-core/ecma/base/ecma-helpers-number.c
+++ b/jerry-core/ecma/base/ecma-helpers-number.c
@@ -697,7 +697,7 @@ ecma_number_parse_int (const lit_utf8_byte_t *string_buff, /**< routine's first 
                                                            *   string buffer's size */
                         ecma_value_t radix) /**< routine's second argument */
 {
-  if (string_buff_size <= 0)
+  if (string_buff_size == 0)
   {
     return ecma_make_nan_value ();
   }
@@ -874,7 +874,7 @@ ecma_number_parse_float (const lit_utf8_byte_t *string_buff, /**< routine's firs
                           lit_utf8_size_t string_buff_size) /**< routine's first argument's
                                                              *   string buffer's size */
 {
-  if (string_buff_size <= 0)
+  if (string_buff_size == 0)
   {
     return ecma_make_nan_value ();
   }
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.c
@@ -349,6 +349,7 @@ ecma_instantiate_builtin (ecma_builtin_id_t obj_builtin_id) /**< built-in id */
 
   ecma_object_t *prototype_obj_p;
 
+  /* cppcheck-suppress arrayIndexOutOfBoundsCond */
   if (JERRY_UNLIKELY (object_prototype_builtin_id == ECMA_BUILTIN_ID__COUNT))
   {
     prototype_obj_p = NULL;
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -1296,7 +1296,9 @@ ecma_op_lazy_instantiate_prototype_object (ecma_object_t *object_p) /**< the fun
   }
 #endif /* ENABLED (JERRY_ES2015) */
 
+#if ENABLED (JERRY_ES2015)
   if (proto_object_p == NULL)
+#endif /* ENABLED (JERRY_ES2015) */
   {
     proto_object_p = ecma_op_create_object_object_noarg ();
   }
--- a/jerry-debugger/jerry_client.py
+++ b/jerry-debugger/jerry_client.py
@@ -252,7 +252,7 @@ def src_check_args(args):
         print("Error: Non-negative integer number expected: %s" % (val_errno))
         return -1
 
-# pylint: disable=too-many-branches,too-many-locals,too-many-statements,redefined-variable-type
+# pylint: disable=too-many-branches,too-many-locals,too-many-statements
 def main():
     args = jerry_client_main.arguments_parse()
 
--- a/jerry-debugger/jerry_client_main.py
+++ b/jerry-debugger/jerry_client_main.py
@@ -267,7 +267,7 @@ class DebuggerAction(object):
 
 
 class JerryDebugger(object):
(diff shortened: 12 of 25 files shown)


ossy-szeged:
#3069 fixes the strange issue caught by the ASAN build.

ossy-szeged:
2.5 months have passed. Can we go ahead and bump the reference platform? 

It would be great to choose 16.04LTS or 18.04LTS (and 20.04LTS is coming in 5 months), 
because we have to fix many pylint and cppcheck issues at the same time with bumping.

ossy-szeged:
Just to have a detailed log about pylint and cppcheck warnings fixed or suppressed:
```
$ tools/run-tests.py --check-pylint
Test command:
/home/oszi/jerryscript/tools/check-pylint.sh
Using config file /home/oszi/jerryscript/tools/pylint/pylintrc
************* Module run-tests
C:301,14: Do not use `len(SEQUENCE)` to determine if a sequence is empty (len-as-condition)
************* Module jerry_client_main
E:270, 0: Bad option value 'redefined-variable-type' (bad-option-value)
R:483, 4: Either all return statements in a function should return an expression, or none of them should. (inconsistent-return-statements)
************* Module jerry_client
E:253, 0: Bad option value 'redefined-variable-type' (bad-option-value)

-------------------------------------------------------------------
Your code has been rated at 9.96/10 (previous run: 10.00/10, -0.04)
```

```
$ tools/run-tests.py --check-cppcheck
Test command:
/home/oszi/jerryscript/tools/check-cppcheck.sh
jerry-core/ecma/base/ecma-helpers-value.c:40: error(zerodiv): Division by zero.
jerry-core/ecma/builtin-objects/ecma-builtin-helpers-macro-defines.inc.h:28: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-helpers-macro-undefs.inc.h:19: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.inc.h:45: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.c:130: style(unsignedLessThanZero): Checking if unsigned variable 'string_buff_size' is less than zero.
jerry-core/ecma/builtin-objects/ecma-builtin-global.c:308: style(unsignedLessThanZero): Checking if unsigned variable 'string_buff_size' is less than zero.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:57: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:63: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:70: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:77: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:84: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:91: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:103: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:136: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:142: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:150: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:157: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:202: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:208: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:215: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:222: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:229: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-global.inc.h:249: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-object.inc.h:42: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtin-object.inc.h:57: information(ConfigurationNotChecked): Skipping configuration 'ENABLED' since the value of 'ENABLED' is unknown. Use -D if you want to check it. You can use -U to skip it explicitly.
jerry-core/ecma/builtin-objects/ecma-builtins.c:351: warning(arrayIndexOutOfBoundsCond): Either the condition 'object_prototype_builtin_id==ECMA_BUILTIN_ID__COUNT' is redundant or the array 'ecma_builtin_property_list_references[8]' is accessed at index 8, which is out of bounds.
jerry-core/ecma/builtin-objects/ecma-builtins.c:351: warning(arrayIndexOutOfBoundsCond): Either the condition 'object_prototype_builtin_id==ECMA_BUILTIN_ID__COUNT' is redundant or the array 'ecma_builtin_descriptors[8]' is accessed at index 8, which is out of bounds.
jerry-core/ecma/builtin-objects/ecma-builtins.c:351: warning(arrayIndexOutOfBoundsCond): Either the condition 'object_prototype_builtin_id==ECMA_BUILTIN_ID__COUNT' is redundant or the array 'jerry_global_context.ecma_builtin_objects[8]' is accessed at index 8, which is out of bounds.
jerry-core/jmem/jmem-heap.c:291: error(memleak): Memory leak: data_space_p
jerry-core/vm/vm.c:3778: style(variableScope): The scope of the variable 'completion_value' can be reduced.
jerry-ext/handler/handler-register.c:66: warning(pointerAdditionResultNotNull): Comparison is wrong. Result of 'entries+idx' can't be 0 unless there is pointer overflow, and pointer overflow is undefined behaviour.
jerry-ext/handler/handler-register.c:103: warning(pointerAdditionResultNotNull): Comparison is wrong. Result of 'entries+idx' can't be 0 unless there is pointer overflow, and pointer overflow is undefined behaviour.
jerry-libm/acos.c:91: style(duplicateExpression): Same expression on both sides of '/'.
jerry-libm/asin.c:92: style(duplicateExpression): Same expression on both sides of '/'.
jerry-libm/atan2.c:85: error(shiftTooManyBitsSigned): Shifting signed 32-bit value by 31 bits is undefined behaviour
jerry-libm/exp.c:200: portability(shiftNegativeLHS): Shifting a negative value is technically undefined behaviour
jerry-libm/fmod.c:56: style(duplicateExpression): Same expression on both sides of '/'.
jerry-libm/fmod.c:126: warning(shiftTooManyBitsSigned): Shifting signed 32-bit value by 31 bits is undefined behaviour. See condition at line 124.
jerry-libm/fmod.c:144: warning(shiftTooManyBitsSigned): Shifting signed 32-bit value by 31 bits is undefined behaviour. See condition at line 142.
jerry-libm/pow.c:262: style(duplicateExpression): Same expression on both sides of '/'.
jerry-libm/pow.c:257: error(shiftTooManyBitsSigned): Shifting signed 32-bit value by 31 bits is undefined behaviour
jerry-libm/sqrt.c:128: style(duplicateExpression): Same expression on both sides of '/'.
jerry-main/main-unix-snapshot.c:153: portability(invalidPrintfArgType_uint): %lu in format string (no. 2) requires 'unsigned long' but the argument type is 'size_t {aka unsigned long}'.
jerry-main/main-unix-snapshot.c:587: portability(invalidPrintfArgType_uint): %lu in format string (no. 1) requires 'unsigned long' but the argument type is 'size_t {aka unsigned long}'.
jerry-main/main-unix-snapshot.c:621: portability(invalidPrintfArgType_uint): %lu in format string (no. 2) requires 'unsigned long' but the argument type is 'size_t {aka unsigned long}'.
jerry-main/main-unix-snapshot.c:755: portability(invalidPrintfArgType_uint): %lu in format string (no. 2) requires 'unsigned long' but the argument type is 'size_t {aka unsigned long}'.
jerry-main/main-unix.c:57: warning(nullPointer): Possible null pointer dereference: file_name
tests/unit-core/test-arraybuffer.c:290: style(variableScope): The scope of the variable 'buffer_p' can be reduced.
tests/unit-core/test-typedarray.c:175: portability(invalidPointerCast): Casting between integer* and float* which have an incompatible binary data representation.
tests/unit-core/test-typedarray.c:176: portability(invalidPointerCast): Casting between integer* and double* which have an incompatible binary data representation.
tests/unit-core/test-typedarray.c:521: style(variableScope): The scope of the variable 'expected_value' can be reduced.
*:0: information(unmatchedSuppression): Unmatched suppression: commaSeparatedReturn
```

jerry-ext/handler/handler-register.c
```diff
@@ -63,6 +63,7 @@ jerryx_set_properties (const jerry_value_t target_object, /**< target object */
 {
 #define JERRYX_SET_PROPERTIES_RESULT(VALUE, IDX) ((jerryx_register_result) { VALUE, IDX })
   uint32_t idx = 0;
+  /* cppcheck-suppress pointerAdditionResultNotNull */
   for (; ((entries + idx) != NULL) && (entries[idx].name != NULL); idx++)
```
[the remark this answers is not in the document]
ossy-szeged: You're right, it only happens if entries is null. I would write an early return instead of moving the whole code to an if block.

jerry-ext/handler/handler-register.c
```diff
@@ -100,6 +101,7 @@ jerryx_release_property_entry (const jerryx_property_entry entries[], /**< list
                                const jerryx_register_result register_result) /**< previous result of registration */
 {
   for (uint32_t idx = register_result.registered;
+       /* cppcheck-suppress pointerAdditionResultNotNull */
        ((entries + idx) != NULL) && (entries[idx].name != NULL);
```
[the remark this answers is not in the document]
ossy-szeged: Yes.

jerry-core/ecma/base/ecma-helpers-value.c
```diff
@@ -37,6 +37,7 @@ JERRY_STATIC_ASSERT (sizeof (jmem_cpointer_t) <= sizeof (ecma_value_t),
 
 #ifdef ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY
 
+/* cppcheck-suppress zerodiv */
```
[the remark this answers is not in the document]
ossy-szeged: > To be honest, I'm not really fond of inline cppcheck-specific warning suppression comments in the C code. They either signal something bad that should be fixed, or they should be listed in the suppression list file if they are false positives. IMHO.

I think adding a suppression anywhere should always mean false positive warning. If it signals a real bug, it should be fixed instead of suppressing. In exceptional cases I can imagine suppressing a warning until fixing, but in this case I would add at least a FIXME comment and file a bug report.

I'd rather add inline suppress comments, because they aren't sensitive to linenumber changes. I'm afraid if we added line specific suppressions to cppcheck/suppression-list file, we should update line numbers in it too frequently.

jerry-libm/atan2.c
```diff
@@ -82,7 +82,7 @@ atan2 (double y, double x)
   {
     return atan (y);
   }
-  m = ((hy >> 31) & 1) | ((hx >> 30) & 2); /* 2 * sign(x) + sign(y) */
+  m = ((hy < 0) ? 1 : 0) + ((hx < 0) ? 2 : 0); /* 2 * sign(x) + sign(y) */
```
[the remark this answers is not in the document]
ossy-szeged: On all modern CPUs integer add and bitwise OR is executed in 1 CPU cycle. I can't see any reason why not to use the exact operation in the formula in the following comment.

jerry-core/ecma/base/ecma-helpers-value.c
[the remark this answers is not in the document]
ossy-szeged: Of course I can move the remaining 4 inline suppression to the cppcheck/suppression-list, if do you still prefer that solution. Or maybe we should add comment for the inline suppression to clarify that they are false warnings, for example:
/* cppcheck-suppress <warning-type> ; false cppcheck warning */

ossy-szeged: Some comments for this suppression. It's definitely false warning, suppressing is the only way to fix this issue.
Cppcheck assumes the types and type sizes are fixed ("By default the platform which was used to compile Cppcheck is used."), but additionally it checks all combinations of if/ifdef macros. Unfortunately it is an invalid assumption, because the true case is for x86, the false case is for x86_64. It's incorrect to check the true case with x86_64 type sizes, which makes the static assert fails ( which is 1 / 0 == division by zero)

Here is a small example to show the same false warning:
```
#include <stdint.h>
#if UINTPTR_MAX <= UINT32_MAX
/* CPU: x86 / ARM */
enum { akarmi = 1 / (!!(sizeof (uintptr_t) <= sizeof (uint32_t)  ))};
#else
/* CPU: x86_64*/
enum { akarmi = 1 / (!!(sizeof (uintptr_t) > sizeof (uint32_t)  ))};
#endif
```

jerry-core/jmem/jmem-heap.c
```diff
@@ -288,6 +288,7 @@ jmem_heap_gc_and_alloc_block (const size_t size, /**< required memory size */
 
   void *data_space_p = jmem_heap_alloc (size);
 
+  /* cppcheck-suppress memleak */
   while (JERRY_UNLIKELY (data_space_p == NULL) && JERRY_LIKELY (pressure < max_pressure))
   {
```
ossy-szeged: It is a nonsense, it can't be a memleak, data_space_p gets new value only if it is NULL.

tests/unit-core/test-arraybuffer.c
```diff
@@ -287,6 +287,7 @@ main (void)
   /* Test ArrayBuffer external memory map/unmap */
   {
     const uint32_t buffer_size = 20;
+    /* cppcheck-suppress variableScope */
```
ossy-szeged: False warning, buffer_p can't be moved to the next block, because it is used from JS code: "var array = new Uint8Array(input_buffer);". Of course cppcheck shouldn't know that jerry_create_arraybuffer_external  and register_js_value propagates buffer_p to a JS code. I can't see any better fix than suppressing the warning.

jerry-libm/atan2.c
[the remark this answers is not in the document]
akosthekiss: Fair enough, works with me

jerry-core/ecma/base/ecma-helpers-value.c
akosthekiss: Something is still itching me. There are other static asserts in the code that are mutually exclusive. E.g., in jerry-core/ecma/base/ecma-helpers-number.c:
```c
#if !ENABLED (JERRY_NUMBER_TYPE_FLOAT64)
JERRY_STATIC_ASSERT (sizeof (ecma_number_t) == sizeof (uint32_t),
                     size_of_ecma_number_t_must_be_equal_to_4_bytes);
#elif ENABLED (JERRY_NUMBER_TYPE_FLOAT64)
JERRY_STATIC_ASSERT (sizeof (ecma_number_t) == sizeof (uint64_t),
                     size_of_ecma_number_t_must_be_equal_to_8_bytes);
```
These don't trigger cppcheck warnings, and not because they are suppressed. What's the difference? Cannot we tweak the problematic code to get rid of the warning?

tests/unit-core/test-arraybuffer.c
akosthekiss: Well, what I'm not sure of in the first place is why is there a block for the 3 statements? It's not the body of an if/for/while/... . And by my very subjective opinion, having such "pointless" blocks is ugly. Getting rid of the superfluous curly braces could both improve code quality and make cppcheck happy.

(Unfortunately, tests have quite some of such blocks, it doesn't occur here alone. At least that used to be the case.)

jerry-core/ecma/base/ecma-helpers-value.c
akosthekiss: BTW, in this code:
```c
#ifdef ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY

JERRY_STATIC_ASSERT (sizeof (uintptr_t) <= sizeof (ecma_value_t),
                     uintptr_t_must_fit_in_ecma_value_t);

#else /* !ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY */

JERRY_STATIC_ASSERT (sizeof (uintptr_t) > sizeof (ecma_value_t),
                     uintptr_t_must_not_fit_in_ecma_value_t);

#endif /* ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY */
```
what is the else branch trying to express? "uintptr_t must not fit in ecma_value_t"? must not? What will make the program misbehave? (The if branch is clear: if uintptr_t doesn't fit in ecma_value_t, quite some code will not work in the ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY-guarded code paths. But it doesn't seem to be the case the other way around.)

ossy-szeged: > These don't trigger cppcheck warnings, and not because they are suppressed. What's the difference? Cannot we tweak the problematic code to get rid of the warning?

They are different. Both assert in the cited JERRY_NUMBER_TYPE_FLOAT64 example is true and aren't architecture (type size) dependent, the user can set JERRY_NUMBER_TYPE_FLOAT64.

The original issue is different, because the #if condition is a constant architecture dependent value. On x86 we get: #if true, on x86_64 we get #if false. The problem is that cppcheck shouldn't check the other assertion which will never be reached on the host architecture.

tests/unit-core/test-arraybuffer.c
ossy-szeged: I discussed with the author of the code. This block is added to ensure input_buffer isn't used later and VLA is used instead of malloc to test jerry_create_arraybuffer_external without free callback passed.

ossy-szeged:
Ping?

14.04 LTS reached EOL ~1 year ago and 20.04 LTS is coming soon.

PR #3064, opened 2019-09-05
Improve expected identifier checks.

zherczeg:
Checks for "of" or "from" does not accept quoted strings.

### DIFF
--- a/jerry-core/parser/js/js-parser-expr.c
+++ b/jerry-core/parser/js/js-parser-expr.c
@@ -435,7 +435,8 @@ parser_parse_class_literal (parser_context_t *context_p) /**< context */
         is_computed = true;
       }
       else if (!(status_flags & PARSER_CLASS_STATIC_FUNCTION)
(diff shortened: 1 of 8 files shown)


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -435,7 +435,8 @@ parser_parse_class_literal (parser_context_t *context_p) /**< context */
         is_computed = true;
       }
       else if (!(status_flags & PARSER_CLASS_STATIC_FUNCTION)
-               && lexer_compare_raw_identifier_to_current (context_p, "constructor", 11))
+               && LEXER_IS_IDENT_OR_STRING (context_p->token.lit_location.type)
```
galpeter: Wouldn't it make sense to have this `LEXER_IS_IDENT_OR_STRING` in the method? And not as an assert? The other variant of the method correctly checks the types without assert.

zherczeg: This is a pretty low level function and in the other cases it knows that the type is limited to the allowed types.

PR #3069, opened 2019-09-05
Get rid of strict aliasing rule violations from libm

ossy-szeged:
Math.exp() returns wrong results in many cases on Ubuntu 16.04 and
18.04 on a special build configuration (ASAN build on Travis CI).

The problem was that the high word of a double variable is modified
with an int pointer and then the original unmodified double variable
was returned. The compiler couldn't detect that the int pointer is
an alias for that double, that's why it returned the cached original
double value.

### DIFF
--- a/jerry-libm/acos.c
+++ b/jerry-libm/acos.c
@@ -70,7 +70,7 @@
 double
 acos (double x)
 {
-  double z, p, q, r, w, s, c, df;
+  double z, p, q, r, w, s, c;
   int hx, ix;
 
   hx = __HI (x);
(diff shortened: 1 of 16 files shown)


akosthekiss:
If I get it right, the problem is caused by the `__HI (y) +=` statements. The `__HI` macro is used at several places in the libm. Do they not cause problem? All the uses should also be strict aliasing violations. Is there no general solution for this?

ossy-szeged:
> If I get it right, the problem is caused by the `__HI (y) +=` statements. The `__HI` macro is used at several places in the libm. Do they not cause problem? All the uses should also be strict aliasing violations. Is there no general solution for this?

Yes, you're right, __HI(x) and __LO(x) macros are the problem. All uses of them on left side of any assignment operator are strict aliasing rule violation. Fortunately this one was the only one issue we ran into until now. But it doesn't mean that we shouldn't fix this issue everywhere in libm. Let's do it.

Unfortunately there is only one valid way to do type punning without strict aliasing violation, which is only valid in C99 (and above), but not necessarily valid in C++. But I don't think it is a big problem, since JerryScript has officially C99 source code.

typedef union
{
  double dbl;
  struct
  {
    int lo;
    int hi;
  } asInt;
} double_accessor;

I'll update my PR soon with fixing all possible places where HI(x) and LO(x) is used at left side of assignment operator and making HI and LO macros const not to be able to use with strict aliasing violations.

ossy-szeged:
Very strange ... cppcheck noticed an already existed fail. Why didn't notice it previously?
jerry-libm/pow.c:246: style(duplicateExpression): Same expression on both sides of '-'.

It seems I had to use NAN instead of the "(z - z) / (z - z)" hack now to make Travis CI green.

rerobika:
@ossy-szeged As I see you replaced the all the lvalue assignment `__HI(x)` and `__LOW(x)` macros. Could you tell me is there any plan/is it possible to get rid of these macros completely?

ossy-szeged:
> @ossy-szeged As I see you replaced the all the lvalue assignment `__HI(x)` and `__LOW(x)` macros. Could you tell me is there any plan/is it possible to get rid of these macros completely?

I'm planning to remove these macros in a separated PR. It will be a little bit smaller PR than this one.

Issue #3067, opened 2019-09-05
Assertion byte_code_start_p[-1] == CBC_EXT_PUSH_SUPER || byte_code_start_p[-1] == CBC_EXT_PUSH_CONSTRUCTOR_SUPER_PROP in vm_loop


renatahodovan:
###### JerryScript revision
086c4ebf

###### Build platform
Linux-4.15.0-54-generic-x86_64-with-Ubuntu-18.04-bionic

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```

###### Test case
```javascript
function f ( ) { }

var B = class extends f {
    constructor ( ) {
        super(0)
        super.$
        this.$= $
    }
}

C = class extends B {
    g ( ) {
        return function ( ) { }
    }
}

D = class extends C {
    constructor ( ) { super ( ) }
    g ( ) { }
}

new D
```

###### Output
```text
ICE: Assertion 'byte_code_start_p[-1] == CBC_EXT_PUSH_SUPER || byte_code_start_p[-1] == CBC_EXT_PUSH_CONSTRUCTOR_SUPER_PROP' failed at jerryscript/jerry-core/vm/vm.c(vm_loop):1621.
Error: ERR_FAILED_INTERNAL_ASSERTION
```

###### Backtrace
```text
bt
#0  0xf7fd5059 in __kernel_vsyscall ()
#1  0xf77fc832 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf77fdcc1 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0x5657ae47 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-port/default/default-fatal.c:71
#4  0x56617934 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
[...]

PR #3077, opened 2019-09-06
Correctly propagate super prop ref when parsing expressions

galpeter:
The "super prop ref" flag is incorrectly used at multiple expressions
as it was only cleared if there was an assignment operation.


galpeter:
Fixes: #3067

PR #3086, opened 2019-09-09
Rework ecma collection

rerobika:
After this patch the ecma value collection is a resizable buffer of ecma-values where the adjacent elements are allocated next to each other.


jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1216,57 +1215,37 @@ typedef double ecma_number_t;
...
+typedef struct
+{
+  uint32_t item_count; /**< number of items in the collection */
+  uint32_t capacity; /**< number of items can be stored in the underlying buffer */
+  ecma_value_t *buffer_p; /**< underlying data buffer */
+} ecma_collection_t;
 
 /**
- * Collection flags.
+ * TODO
```
dbatyai: It would seem there's still something left to do.

rerobika: Opps, thanks I've added the missing descriptions.

PR #3128, opened 2019-09-17
Add public API functions for internal property management

rerobika:
These 4 new API functions give possibility to perform [[Get]], [[Set]], [[Has]], [[Delete]] operations for properties
which are not accessible from the JavaScript context only from the public API.


docs/02.API-REFERENCE.md
```diff
@@ -4679,6 +4679,65 @@ main (void)
...
+
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t global_object = jerry_get_global_object ();
+  jerry_value_t prop_name = jerry_create_string ((const jerry_char_t *) "hidden_property");
+
+  jerry_value_t has_internal_js_prop = jerry_has_own_property (global_object, prop_name);
```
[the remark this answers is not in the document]
rerobika: Opps, typo.

docs/02.API-REFERENCE.md
```diff
@@ -4680,6 +4680,63 @@ main (void)
...
+
+  return 0;
+}
+```
+
+**See also**
+
+- [jerry_delete_internal_property](#jerry_delete_internal_property)
+- [jerry_get_internal_property](#jerry_get_internal_property)
+- [jerry_set_internal_property](#jerry_set_internal_property)
```
zherczeg: Just a thought: what about having a function, which returns with the object directly, and you can do the has/delete/enumerate etc. operations directly? I think they are rarely needed, but would be useful when needed.

rerobika: I think we should not propagate for API level how the internal properties as managed. I can accept that it'd reduce the number of new API functions,  but IMO publishing the internal object directly can be confusing, and also can cause reference count management problems, since the published object can be released/acquired by the user.

zherczeg: Ok.

PR #3146, opened 2019-09-23
Add toLength and toInteger operations

Yuyupo:


tests/unit-core/test-to-integer.c
```diff
@@ -0,0 +1,125 @@
...
+  TEST_ASSERT (num == 123);
+
+  /* 3 */
+  ecma_value_t nan = ecma_make_nan_value ();
+
+  result = ecma_op_to_integer (nan, &num);
+
+  ecma_free_value (nan);
+
+  TEST_ASSERT (!ECMA_IS_VALUE_ERROR (result));
```
rerobika: Can we add a testcase when the operation fails?

zherczeg:
Who is using these functions?

rerobika:
@zherczeg Unfortunately currently nobody uses these functions, but these conversion operations also present in ES5.1. To be more precise, in the ES5.1 profile instead of calling these conversion methods, the methods' steps are implemented in all occurrences which increases the binary size. So  we are planning to revise these conversion methods all over the code base to be consistent with the 5.1 and the 6.0 standard as well.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1092,6 +1092,11 @@ typedef double ecma_number_t;
 #define ECMA_NUMBER_FRACTION_WIDTH   (52)
 #endif /* !ENABLED (JERRY_NUMBER_TYPE_FLOAT64) */
 
+/**
+ * Value 'NaN' of ecma_number_t
+ */
+#define ECMA_NUMBER_NAN (ecma_number_t) (0.0 / 0.0)
```
ossy-szeged: It seems MSVC doesn't like 0.0/0.0 to generate NaN:
C:/projects/jerryscript/jerry-core/ecma/operations/ecma-conversion.c(935): error C2124: divide or mod by zero [C:\projects\jerryscript\build\jerry-core\jerry-core.vcxproj]

But you can use the ecma_number_make_nan() function which is used everywhere to generate NaN.

Yuyupo: Was trying to avoid using that function. Seems like I need to use that for now. Thank you. :)

jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -910,6 +912,122 @@ ecma_op_to_property_descriptor (ecma_value_t obj_value, /**< object value */
...
+
+  /* 5 */
+  if (ecma_number_is_infinity (num))
+  {
+    *length = UINT32_MAX;
+    return ECMA_VALUE_EMPTY;
+  }
+
+  /* 6 */
+  *length = (uint32_t) JERRY_MIN (num, UINT32_MAX);
```
dbatyai: `JERRY_MIN` is not necessary here, the `(uint32_t)` cast alone will correctly clamp the value to the [0, UINT32_MAX] range.

jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -910,6 +912,122 @@ ecma_op_to_property_descriptor (ecma_value_t obj_value, /**< object value */
...
+                   uint32_t *length) /**< [out] ecma number */
+{
+  /* 1 */
+  if (ECMA_IS_VALUE_ERROR (value))
+  {
+    return value;
+  }
+
+#if ENABLED (JERRY_ES2015)
+
```
dbatyai: Please remove this empty line.

jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -910,6 +912,119 @@ ecma_op_to_property_descriptor (ecma_value_t obj_value, /**< object value */
...
+  /* 4 */
+  if (num <= 0.0f)
+  {
+    *length = 0;
+    return ECMA_VALUE_EMPTY;
+  }
+
+  /* 5 */
+  if (ecma_number_is_infinity (num))
```
dbatyai: If we change this condition to `num >= (ecma_number_t) UINT32_MAX`, then the `<= 0.0` check above would no longer be necessary, the cast will handle negative numbers correctly.

Yuyupo: The 4th and 5th returns with a different values. the `<= 0.0` check returns with 0, while the `ecma_number_is_infinity` returns with `UINT32_MAX`.
I'm pretty sure those two are needed.

dbatyai: The `(uint32_t)` cast correctly clamps negative numbers and negative infinity to zero. It also correctly clamps numbers > UINT32_MAX to UINT32_MAX. The only problem is +Infinity, which results in 0. Since `ecma_number_is_infinity` is also true for -Infinity, we can check >= UINT32_MAX  and return UINT32_MAX, which would also handle +Infinity correctly. This would achieve the same result with one less condition.

dbatyai: My bad, apparently casting constant values work differently, than casting during runtime. It is necessray to check <= 0.

tests/unit-core/test-to-length.c
```diff
@@ -0,0 +1,121 @@
...
+  /* +infinity */
+  ecma_value_t positive_infinity = ecma_make_number_value (ecma_number_make_infinity (false));
+
+  result = ecma_op_to_length (positive_infinity, &num);
+
+  ecma_free_value (positive_infinity);
+
+  TEST_ASSERT (!ECMA_IS_VALUE_ERROR (result));
+#if ENABLED (JERRY_ES2015)
+  TEST_ASSERT (num == ecma_number_to_uint32 (num));
```
dbatyai: What does this test? `num` is `uint32_t`, so it seems to me this is always true.

PR #3159, opened 2019-09-24
Reduce code duplication in js-parser-expr.c

rerobika:


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -40,7 +50,8 @@
 static const uint8_t parser_binary_precedence_table[36] =
 {
   3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3, 3,
-  4, 5, 6, 7, 8, 9, 10, 10, 10, 10,
+  PARSER_RIGHT_TO_LEFT_ORDER_TERNARY_PRECEDENCE, 5, PARSER_RIGHT_TO_LEFT_ORDER_MAX_PRECEDENCE,
```
zherczeg: What is the precedence of 5? Do we actually need to insert these numbers into the array?

rerobika:
```
LEXER_QUESTION_MARK,           /**< "?" (prec: 4) */
LEXER_LOGICAL_OR,              /**< "||" (prec: 5) */
LEXER_LOGICAL_AND,             /**< "&&" (prec: 6) */
```

Actually no need to insert them, I'll update the PR.

PR #3163, opened 2019-09-26
Make ecma_number_make_nan more optimal and C99 conform

ossy-szeged:
This change makes ecma_number_make_nan and ecma_number_make_infinity
always return constant value without any function call. Previously we
relied on compiler optimizations.

The ecma_number_t_accessor union is introduced to be able to access
float values as float and uint32_t (and doubles as double and uint64_t)
properly, without violating strict aliasing rules. There were many
copies of it, all of them were replaced to this new union.

Additionally ecma_number_make_nan should return QNaN instead of SNaN,
same value as C99 nan(""). Unfortunately calling nan("") here isn't
always optimal, because compilers sometimes generate constant returns,
sometimes function calls.

Before this change ecma_number_make_nan returned SNaN:
- double: 0x7FF0 0000 0000 0001 (sign:0, exponent: all 1 bits, fraction: 0...01)
- float: 0x7F8 00001 (sign:0, exponent: all 1 bits, fraction: 0...01)

After this change ecma_number_make_nan returns QNaN:
- double: 0x7FF8 0000 0000 0000 (sign:0, exponent: all 1 bits, fraction: 10..0)
- float: 0x7FC0 0000 (sign:0, exponent: all 1 bits, fraction: 10...0)

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -1024,6 +1024,15 @@ typedef struct
  */
 typedef float ecma_number_t;
 
+/**
+ * It makes possible to read/write an ecma_number_t as uint32_t without strict aliasing rule violation.
+ */
+typedef union
+{
+  ecma_number_t as_ecma_number_t;
+  uint32_t as_uint32_t;
+} ecma_number_accessor_t;
+
 #define DOUBLE_TO_ECMA_NUMBER_T(value) (ecma_number_t) (value)
 
 /**
@@ -1059,6 +1068,15 @@ typedef float ecma_number_t;
  * Description of an ecma-number
  */
 typedef double ecma_number_t;
+
+/**
+ * It makes possible to read/write an ecma_number_t as uint64_t without strict aliasing rule violation.
+ */
+typedef union
+{
+  ecma_number_t as_ecma_number_t;
+  uint64_t as_uint64_t;
+} ecma_number_accessor_t;
 
 #define DOUBLE_TO_ECMA_NUMBER_T(value) value
 
--- a/jerry-core/ecma/base/ecma-helpers-number.c
+++ b/jerry-core/ecma/base/ecma-helpers-number.c
@@ -58,15 +58,9 @@ ecma_number_pack (bool sign, /**< sign */
                            (biased_exp << ECMA_NUMBER_FRACTION_WIDTH) |
                            ((uint32_t) fraction));
 
-  union
-  {
-    uint32_t u32_value;
-    ecma_number_t float_value;
-  } u;
-
-  u.u32_value = packed_value;
-
-  return u.float_value;
+  ecma_number_accessor_t u;
+  u.as_uint32_t = packed_value;
+  return u.as_ecma_number_t;
 } /* ecma_number_pack */
 
 /**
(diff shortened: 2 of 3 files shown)


jerry-core/ecma/base/ecma-helpers-number.c
```diff
@@ -267,9 +267,25 @@ ecma_number_is_nan (ecma_number_t num) /**< ecma-number */
...
 {
-  return ecma_number_pack (false,
-                           (1u << ECMA_NUMBER_BIASED_EXP_WIDTH) - 1u,
-                           1u);
+#if ENABLED (JERRY_NUMBER_TYPE_FLOAT64)
+  union
+  {
+    double dbl;
+    uint64_t ui;
+  } f;
```
rerobika: Can we reuse this union for making +-Infinity value? If we could I'd move this into ecma-globals.h.

jerry-core/ecma/base/ecma-helpers-number.c
```diff
@@ -267,9 +267,25 @@ ecma_number_is_nan (ecma_number_t num) /**< ecma-number */
...
+#else
+  union
+  {
+    float flt;
+    uint32_t ui;
+  } f;
+  f.ui = 0x7fc00000;
+  /* QNaN, same as the C99 nanf("") returns. sign bit: 0, exponent: all 1 bits, fraction: 1....0 */
+  return f.flt;
+#endif
```
rerobika: Missing comment.

jerry-core/ecma/base/ecma-helpers-number.c
```diff
@@ -267,9 +267,25 @@ ecma_number_is_nan (ecma_number_t num) /**< ecma-number */
...
-  return ecma_number_pack (false,
-                           (1u << ECMA_NUMBER_BIASED_EXP_WIDTH) - 1u,
-                           1u);
+#if ENABLED (JERRY_NUMBER_TYPE_FLOAT64)
+  union
+  {
+    double dbl;
+    uint64_t ui;
+  } f;
+  f.ui = 0x7ff8000000000000;
```
zherczeg: I would add an ull postfix after the number, or make it from two uint32 value.

jerry-core/ecma/base/ecma-helpers-number.c
ossy-szeged: Good idea. I did it. And removed all duplication of this union.

jerry-core/ecma/base/ecma-helpers-number.c
ossy-szeged: Done.

jerry-core/ecma/base/ecma-helpers-number.c
ossy-szeged: Fixed.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1024,6 +1024,12 @@ typedef struct
  */
 typedef float ecma_number_t;
 
+typedef union
+{
+  ecma_number_t as_ecma_number_t;
+  uint32_t as_uint32_t;
+} ecma_number_t_accessor;
```
rerobika: How about `ecma_number_accessor_t` rather? Also please add doxygen comments for the union/union members as well.

ossy-szeged: Fixed.

PR #3167, opened 2019-09-26
[WIP] Run tests on AppVeyor

ossy-szeged:
Run the following tests on AppVeyor (on x86 and x64 platforms)
- tools/run-tests.py --buildoption-tests (2 x 9 mins)
- tools/run-tests.py --unittests (2 x 25 mins)
- tools/run-tests.py --test262 (2 x 25 mins)
- tools/run-tests.py --jerry-tests (2 x 8 mins)
- tools/run-tests.py --jerry-test-suite (2 x 17 mins)
- tools/run-tests.py --kangax-tests (2 x 3.5 mins)

Still missing tests: (lack of script support on Windows)
- tools/run-tests.py --jerry-debugger
- tools/run-tests.py "all style checker"


dbatyai:
I'm completely in favor of running tests on windows, but I don't think that it would be sustainable in the current condition. I feel like waiting 2+ hours per pull request would be more of a hindrance than a benefit, which would only increase as the remaining test runners are updated.

I would argue that we should either cut tests so that it finishes in a reasonable time, and run only the bare necessities, or run everything as it currently is, but only on the master branch.

ossy-szeged:
Changed title to WIP, because more tests were added. We should measure their runtime, additionally we should decide which tests we would like to run on AppVeyor to avoid unacceptable long runtime.

ossy-szeged:
Unfortunately AppVeyor is very slow, we can't afford 1+ hour long runtime. It was a good experiment, but let's move forward to GitHub Actions.

PR #3175, opened 2019-09-30
Fix Date.prototype.toString() and toISOString()

ossy-szeged:
The implementation was incorrect for negative years and years bigger than 9999.
-1 was 000/ because the negative (year%10) was added to '0' character, years
bigger than 9999 was truncated to 4 digits.

ES5.1 15.9.1.15.1 defines extended years format with 6 digits, but toString()
and toISOString() sections don't mention anything about extended years. ES6
20.3.4.3 already clarifies that Date.prototype.toISOString() should use this
extended year format if it is necessary.

Changes:
- Date.prototype.toString() uses 4 digits for years by default, 5 or 6 if it
is necessary and put '-' sign for negative years, no sign for positive years.
Date.prototype.toString() was implementation dependent until ES9, but ES9
already specify exactly this format.
- Date.prototype.toISOString() uses fixed 4 digits for years 0 - 9999,
otherwise sign + 6 digits (extended years).
- Tests added for corner cases.

### DIFF
--- a/tests/jerry/date-tostring.js
+++ b/tests/jerry/date-tostring.js
@@ -142,3 +142,29 @@ assert (Date (2015, 1, 1) == (new Date ()).toString ());
 assert (Date (Number.NaN) == Date ());
 
 assert (new Date ("2015-07-08T11:29:05.023Z").toISOString() == "2015-07-08T11:29:05.023Z");
+
+// corner cases
+assert (new Date (-8640000000000001).toString() == "Invalid Date")
+assert (new Date (-8640000000000000).toString() == "Tue Apr 20 -271821 00:00:00 GMT+00:00")
+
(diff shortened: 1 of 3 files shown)


tests/jerry/date-tostring.js
```diff
@@ -142,3 +142,30 @@ assert (Date (2015, 1, 1) == (new Date ()).toString ());
 assert (Date (Number.NaN) == Date ());
 
 assert (new Date ("2015-07-08T11:29:05.023Z").toISOString() == "2015-07-08T11:29:05.023Z");
+
+
```
dbatyai: Extra newline

ossy-szeged: Thanks, fixed.

PR #3195, opened 2019-10-02
Make ecma_number_is_infinity more optimal

ossy-szeged:
We relied on compiler optimizations, and hope that the compiler inlines
ecma_number_get_biased_exponent_field and ecma_number_get_fraction_field
functions. Comparing to a constant number is always optimal.


LaszloLango:
Please share the measurement results of this change.

ossy-szeged:
> Please share the measurement results of this change.
I'm on it, will check with different architectures and configurations. 
Please don't merge it until we can confirm the results.

ossy-szeged:
After #3470 there is no significant performance effect with this PR. Additionally using compiler builtins is better direction to optimize this code path.

PR #3196, opened 2019-10-02
Make Date.parse handle 24:00:00.000 time properly

ossy-szeged:
ES5.1 15.9.1.15 Note1 defines that 24:00 is same as 0:00
of the next day. The spec explicitly doesn't mention that
24:01 should be invalid, but it should be self-evident.
(FireFox and Chrome also refuses times bigger than 24:00)


jerry-core/ecma/builtin-objects/ecma-builtin-date.c
```diff
@@ -324,6 +320,11 @@ ecma_builtin_date_parse (ecma_value_t this_arg, /**< this argument */
           }
         }
 
+        if (hours == 24 && (minutes !=0 || seconds !=0 || milliseconds !=0))
```
rerobika: Please add spaces around `!=`.

ossy-szeged: Fixed.

PR #3210, opened 2019-10-09
Refactor code to use ToLength

Yuyupo:

### DIFF
--- a/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
+++ b/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
@@ -933,18 +933,10 @@ ecma_builtin_typedarray_prototype_set (ecma_value_t this_arg, /**< this argument
                   ecma_op_object_get_by_magic_id (source_obj_p, LIT_MAGIC_STRING_LENGTH),
                   ret_val);
 
-  ECMA_OP_TO_NUMBER_TRY_CATCH (source_length_num, source_length, ret_val);
-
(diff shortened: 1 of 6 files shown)


jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -943,7 +943,8 @@ ecma_op_create_typedarray (const ecma_value_t *arguments_list_p, /**< the arg li
 
       ECMA_OP_TO_NUMBER_TRY_CATCH (num2, arg2, ret);
```
rerobika: The  `ECMA_OP_TO_NUMBER_TRY_CATCH` is unnecessary when using toLength operation.

Yuyupo: I made an error there the language specification says to use ToInteger there and not ToLength. So I've deleted my modifications there. On the other hand, I should have used ToLength somewhere else which I have added.

jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -940,7 +940,11 @@ ecma_builtin_typedarray_prototype_set (ecma_value_t this_arg, /**< this argument
     source_length_num = 0;
   }
 
-  uint32_t source_length_uint32 = ecma_number_to_uint32 (source_length_num);
```
DanielBallaSZTE: `ECMA_OP_TO_NUMBER_TRY_CATCH`(936) should be substituted with `ecma_op_to_length` .

Issue #3218, opened 2019-10-15
GC is not work for garbages generated by function called by `jerry_call_function()` in C code side


niklauslee:
Hello,

I have my-own implementation of `setInterval(callback, interval)` written in C. When I run the below code, heap memory usage is increasing every callback call and finally system is stuck due to the lac of memory.

```js
setInterval(function () {
  var t = millis();
  console.log ("Current time is ", t, "ms.");
}, 100);
```

In our system, there is the main even loop. In the event loop, check interval time and call the callback using `jerry_call_function()`. Every callback call results memory consumption, but garbages are not collected even if `jerry_gc()` is called.

Like below code, heap memory is not increased if the callback is not generates garbages (like without `var t = millis();`).

```js
setInterval(function () {
  console.log ("Current time is unknown");
}, 100);
```

Same phenomenon is observed in all cases that javascript callback function is called using `jerry_call_function()` from C code side. (Timer callback, UART data read callback, etc.)

For more info:
- [setInterval() implementation](https://github.com/kameleon-project/kameleon/blob/master/src/global.c#L317)
- [the code calling callback](https://github.com/kameleon-project/kameleon/blob/master/src/global.c#L282)

What I'm doing wrong? Any advices?

Thank you.

rerobika:
Hi @niklauslee.

After a bit of investigation I've realized that there is no problem with the GC, the leak in your code comes from [this](https://github.com/kameleon-project/kameleon/blob/7fd18349f952fb6c91582e1a85a1eadb8e8243d9/src/jerryxx.c#L91) line, since the `toString`-ed value must be released.

Please let me know whether this missing `jerry_release_value` solved your problem or not. :)

Best regards,
Robert

niklauslee:
Hi @rerobika. Now it's works well after fixing what you pointed.
Thank you so much! 😄

PR #3219, opened 2019-10-15
Static strings should be reference counted in debug mode

rerobika:
This patch helps to find out invalid reference count usage for ecma-strings.


rerobika:
No there isn't. This idea has came up after #3218, since after the string rework static strings are not ref counted anymore, which could lead to misuse the strings reference count. E.g. you give a function argument and you perform `toString` operation on it, but you forgot to release the `toString`ed value. If the argument was a static string there is no leak at all, otherwise you have left a reference on the argument. To be more precise the current code is more permissive, than it should be since it absorbs the wrong ref-count usage both in release/debug mode.

PR #3232, opened 2019-10-18
Fix issue about Number.prototype.toString

### LINKED ISSUE #3229: An issue about method Number.prototype.toString
According to ES5.1, for method [Number.prototype.toString ( [ radix ] )](http://www.ecma-international.org/ecma-262/5.1/index.html#sec-15.7.4.2), if ToInteger(radix) is not an integer between 2 and 36 inclusive throw a RangeError exception. But, Jerry did not throw a RangeError exception when running the following test case.

###### JerryScript revision
[7df87b7](https://github.com/jerryscript-project/jerryscript/commit/7df87b7778124af9b0dd309570c33864a67792f7)

###### Build platform
```
Ubuntu 16.04.6 LTS (Linux 4.15.0-55-generic x86_64)
```

###### Build steps
```
python tools/build.py --clean --debug --logging=on --line-info=on --error-messages=on --cpointer-32bit=on --mem-heap=1572864 --profile=es2015-subset
```

###### Test case
```
var NISLFuzzingFunc = function(e) {
    var tmp = Number(0).toString(e);
    print(tmp);
};
var NISLParameter4 = "hello jerry";
NISLFuzzingFunc(NISLParameter4);
```

###### Execution steps
```
build/bin/jerry testcase.js
```

###### Output
`0`

###### Expected behavior
Throw a RangeError exception

Yuyupo:
Fixes #3229

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
@@ -249,16 +249,6 @@ ecma_builtin_number_prototype_object_to_string (ecma_number_t this_arg_number, /
                                                 const ecma_value_t *arguments_list_p, /**< arguments list */
                                                 ecma_length_t arguments_list_len) /**< number of arguments */
 {
-
-  if (arguments_list_len == 0
-      || ecma_number_is_nan (this_arg_number)
-      || ecma_number_is_infinity (this_arg_number)
-      || ecma_number_is_zero (this_arg_number)
-      || (arguments_list_len > 0 && ecma_is_value_undefined (arguments_list_p[0])))
-  {
-    ecma_string_t *ret_str_p = ecma_new_ecma_string_from_number (this_arg_number);
-    return ecma_make_string_value (ret_str_p);
-  }
   static const lit_utf8_byte_t digit_chars[36] =
   {
     '0', '1', '2', '3', '4', '5', '6', '7', '8', '9',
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -253,7 +253,6 @@ ecma_builtin_number_prototype_object_to_string (ecma_number_t this_arg_number, /
   if (arguments_list_len == 0
       || ecma_number_is_nan (this_arg_number)
       || ecma_number_is_infinity (this_arg_number)
-      || ecma_number_is_zero (this_arg_number)
```
rerobika: How about `Number.prototype.toString.call (-Infinity, "hello jerry")`? This should also throw RangeError. IMO this check should not be modified, just should be placed after the radix validation part.

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -282,7 +281,7 @@ ecma_builtin_number_prototype_object_to_string (ecma_number_t this_arg_number, /
     return ecma_raise_range_error (ECMA_ERR_MSG ("Radix must be between 2 and 36."));
   }
 
-  if (radix == 10)
+  if (radix == 10 || ecma_number_is_zero (this_arg_number))
```
rerobika: IMO this check is not correct. What about other radix values? The source of the problem comes from `if (radix < 2 || radix > 36)` check does not handle `NaN` after the toNumber conversion. That's why toInteger should be used here.

PR #3236, opened 2019-10-21
Introduce jerry_get_resource_name API function

### LINKED ISSUE #2170: Keep resource name for function objects?
Hey JrS people!

I'm trying to implement a native binding much alike Node's [`require.resolve()`](https://nodejs.org/api/modules.html#modules_require_resolve_request_options)
The tricky thing about the `require.resolve` function is that when using a relative path, i.e. `require.resolve('./thing.js')` the look up happens relative to the source file in which the `require.resolve()` call was written.

I noticed that for the JerryScript debugger support, `jerry_parse_named_resource` was added, which seemed promising at a glance. Unfortunately, the "resource name" is merely passed off to the debugger client and not kept around.

Any suggestions on how to solve this?

I guess I want an API like `const jerry_char_t *jerry_get_current_resource_name(void);` that would return the currently executing resource name (or `NULL` if no named resource is executing / it's not available).

For reference, V8 has a [`GetScriptName()`](https://v8docs.nodesource.com/node-9.3/dd/d1e/classv8_1_1_stack_frame.html#a63d53e7637627027f0fe77e0e9dbeab7) method on the `StackFrame` class which is similar to what I'm looking for.

rerobika:
This new API function adds possibility to query the resource name of the currently executed script (including modules) or a function object.
This patch closes #2170.


rerobika:
Depends on #3238.

PR #3260, opened 2019-10-29
Make run-tests --jerry-tests --jerry-test-suite  work on Windows too

ossy-szeged:
Changes:
* Bash based runners/run-test-suite.sh runner replaced with a python runner
* Common util functions moved to util.py
* Fixed EOL issues in tests directory


tools/runners/run-test-suite.py
```diff
@@ -0,0 +1,214 @@
...
+
+def get_tests(test_dir, test_list, skip_list):
+    tests = []
+    if test_dir:
+        tests = []
+        for root, _, files in os.walk(test_dir):
+            tests.extend([os.path.join(root, test_file) for test_file in files if test_file.endswith('.js')])
+
+    if test_list:
+        with open(test_list, "r") as test_list_fd:
```
galpeter: Afaik this could be rewritten as:
```js
dirname = os.path.dirname(test_list)
with open(...) as test_list_fd:
  for entry in test_list_fd:
    tests.append(os.path.normpath(os.path.join(dirname, entry)))
```

tools/runners/run-test-suite.py
```diff
@@ -0,0 +1,214 @@
...
+
+    tests.sort()
+
+    def filter_tests(test):
+        for skipped in skip_list:
+            if skipped in test:
+                return False
+        return True
+
+    return [test for test in tests if filter_tests(test)]
```
galpeter: Could this be rewritten to use the a built-in `filter` method? Something like this:?
```python
return filter(filter_tests, tests)
```

tools/runners/run-test-suite.py
```diff
@@ -0,0 +1,214 @@
...
+        summary_list.append('--snapshot')
+    if args.test_dir:
+        summary_list.append(os.path.relpath(args.test_dir))
+    if args.test_list:
+        summary_list.append(os.path.relpath(args.test_list))
+    util.print_test_summary(' '.join(summary_list), total, passed, failed)
+
+    if failed > 0:
+        return 1
+    return 0
```
galpeter: Can't we just return with the `failed` value?

tools/runners/run-test-suite.py
```diff
@@ -0,0 +1,214 @@
...
+    execute_snapshot_cmd = get_platform_cmd_prefix()
+    generate_snapshot_cmd = get_platform_cmd_prefix()
+    if args.runtime:
+        execute_snapshot_cmd.append(args.runtime)
+        generate_snapshot_cmd.append(args.runtime)
+
+    execute_snapshot_cmd.extend([args.engine, '--exec-snapshot', 'js.snapshot'])
+
+    if sys.platform == 'win32':
+        generate_snapshot_cmd.append(args.engine[:-4] + '-snapshot.exe')
```
galpeter: Why is the `[:-4]` required?

tools/runners/util.py
```diff
@@ -0,0 +1,65 @@
...
+
+    success_color = TERM_GREEN if passed == total else TERM_RED
+    print("%sSuccess: %d%%%s" % (success_color, passed*100/total, TERM_NORMAL))
+
+
+def print_test_result(tested, total, is_passed, passed_string, test_path, is_snapshot_generation=None):
+    if is_snapshot_generation is None:
+        snapshot_string = ''
+    else:
+        snapshot_string = ' (generate snapshot)' if is_snapshot_generation else ' (execute snapshot)'
```
galpeter: Could we write this as:?
```python
if is_snapshot_generation is None:
   ...
elif is_snapshot_generation:
   ...
else:
   ....
```
this way it is more visible that there are three states.

tools/runners/run-test-suite.py
ossy-szeged: I checked, we can do it with an additional rstrip() for entry to strip newline character(s)

tools/runners/run-test-suite.py
ossy-szeged: First I tried with filter, but pylint warned me to use list comprehension instead of filter. Now I can't see this warning with filter (maybe I tested on a different machine with different pylint version.)

tools/runners/run-test-suite.py
ossy-szeged: We can, but ... exit code on Linux is between 0-255. If (failed%256) == 0, the exit could would be 0, which is not good. Of course, we can simplify it with bool case, but we can't return failed directly.

tools/runners/run-test-suite.py
ossy-szeged: Because the name of the binary is different on Windows and Linux (jerry.exe and jerry). But you're right, we can do a platform independent solution with os.path.splitext().

tools/runners/util.py
ossy-szeged: Yes, we can. Good point.

PR #3265, opened 2019-10-30
Implement the spread operator for array initialization

rerobika:


jerry-core/vm/opcodes.c
```diff
@@ -258,18 +261,123 @@ opfunc_for_in (ecma_value_t left_value, /**< left value */
...
+            goto clean_up;
+          }
+
+          while (true)
+          {
+            ecma_value_t next = ecma_op_iterator_step (iterator);
+
+            if (ECMA_IS_VALUE_ERROR (next))
+            {
+              goto clean_up;
```
galpeter: Do we really need the goto here? Can't we just `break` here?

jerry-core/vm/opcodes.c
```diff
@@ -258,18 +261,123 @@ opfunc_for_in (ecma_value_t left_value, /**< left value */
...
+              break;
+            }
+
+            ecma_value_t value = ecma_op_iterator_value (next);
+
+            ecma_free_value (next);
+
+            if (ECMA_IS_VALUE_ERROR (value))
+            {
+              goto clean_up;
```
galpeter: Do we really need the goto here? Can't we just break here?

jerry-core/vm/opcodes.c
```diff
@@ -258,18 +261,123 @@ opfunc_for_in (ecma_value_t left_value, /**< left value */
...
 
+#if ENABLED (JERRY_ES2015)
+  if (has_spread_element)
+  {
+    for (uint32_t i = 0, j = old_length; i < values_length; i++)
+    {
+      if (!ecma_is_value_array_hole (stack_top_p[i]))
+      {
+        ecma_value_t ret_value = ECMA_VALUE_ERROR;
+        ecma_value_t put_comp;
```
galpeter: This should be moved into the code part where the variable is truly used, no need to have this at the start.

jerry-core/vm/opcodes.c
```diff
@@ -258,18 +261,123 @@ opfunc_for_in (ecma_value_t left_value, /**< left value */
...
 
   ecma_extended_object_t *ext_array_obj_p = (ecma_extended_object_t *) array_obj_p;
   uint32_t old_length = ext_array_obj_p->u.array.length;
 
+#if ENABLED (JERRY_ES2015)
+  if (has_spread_element)
+  {
+    for (uint32_t i = 0, j = old_length; i < values_length; i++)
+    {
+      if (!ecma_is_value_array_hole (stack_top_p[i]))
```
galpeter: If possible please invert the check to reduce an indentation level.

jerry-core/vm/opcodes.c
```diff
@@ -258,18 +261,123 @@ opfunc_for_in (ecma_value_t left_value, /**< left value */
...
 
   JERRY_ASSERT (ecma_get_object_type (array_obj_p) == ECMA_OBJECT_TYPE_ARRAY);
 
   ecma_extended_object_t *ext_array_obj_p = (ecma_extended_object_t *) array_obj_p;
   uint32_t old_length = ext_array_obj_p->u.array.length;
 
+#if ENABLED (JERRY_ES2015)
+  if (has_spread_element)
+  {
+    for (uint32_t i = 0, j = old_length; i < values_length; i++)
```
galpeter: Could we move this big code block into it's own method? If it is it's own method, then it could be called after line 298 and return with the result there.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -222,7 +225,18 @@ parser_parse_array_literal (parser_context_t *context_p) /**< context */
...
+        {
+          parser_emit_cbc_call (context_p, PARSER_TO_EXT_OPCODE (CBC_EXT_SPREAD_ARRAY_APPEND), pushed_items);
+        }
+        else
+        {
+#endif /* ENABLED (JERRY_ES2015) */
+          parser_emit_cbc_call (context_p, CBC_ARRAY_APPEND, pushed_items);
+#if ENABLED (JERRY_ES2015)
+        }
+#endif /* ENABLED (JERRY_ES2015) */
```
galpeter: These #if guards are a bit awkward.  What if we do this?:
```c
#if ENABLED(..)
if (pushed_spread_element)
{
   ... 
   return;
}
#endif

parser_emit_cbc_call (....)
```

galpeter: Or maybe we could guard only the second argument?
```c
cbc_opcode_t cbc_opcode = CBC_ARRAY_APPEND;

#if ENABLED(..)
if (pushed_spread_element)
{
   cbc_opcode = PARSER_TO_EXT_OPCODE (CBC_EXT_SPREAD_ARRAY_APPEND);
}
#endif

parser_emit_cbc_call (context_p, cbc_opcode, pushed_items);
```

This solution could also be used below when the same append is invoked. Maybe this could even be extracted into an inline method?

jerry-core/vm/opcodes.c
```diff
@@ -277,18 +280,123 @@ opfunc_for_in (ecma_value_t left_value, /**< left value */
...
+        if (ecma_op_is_spread_object (stack_top_p[i]))
+        {
+          ecma_object_t *spread_object_p = ecma_get_object_from_value (stack_top_p[i]);
+          ecma_value_t spread_value = ecma_op_spread_object_get_spreaded_element (spread_object_p);
+
+          ecma_value_t iterator = ecma_op_get_iterator (spread_value, ECMA_VALUE_EMPTY);
+
+          if (ECMA_IS_VALUE_ERROR (iterator))
+          {
+            goto clean_up;
```
galpeter: It would be good to eliminate this goto if the are also removed.

rerobika:
@galpeter Thanks for the review, I applied you suggestions.

PR #3272, opened 2019-10-31
Implement Object.is routine from ES2015 specification

kaadam:
Related part of the standard: ECMA-262 v6, 19.1.2.10

JerryScript-DCO-1.0-Signed-off-by: Adam Kallai kadam@inf.u-szeged.hu

PR #3273, opened 2019-10-31
Fix the same Symbol value detection in SameValue function

kaadam:
'is_types_equal' check was missing for Symbol values.

This fix is necessary for this PR #3272.

### DIFF
--- a/jerry-core/ecma/operations/ecma-conversion.c
+++ b/jerry-core/ecma/operations/ecma-conversion.c
@@ -94,11 +94,19 @@ ecma_op_same_value (ecma_value_t x, /**< ecma value */
   const bool is_y_string = ecma_is_value_string (y);
   const bool is_y_object = ecma_is_value_object (y);
 
+#if ENABLED (JERRY_ES2015)
+  const bool is_x_symbol = ecma_is_value_symbol (x);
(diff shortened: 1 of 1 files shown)


jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -94,12 +94,21 @@ ecma_op_same_value (ecma_value_t x, /**< ecma value */
...
+#endif /* ENABLED (JERRY_ES2015) */
+
   const bool is_types_equal = ((is_x_undefined && is_y_undefined)
                                || (is_x_null && is_y_null)
                                || (is_x_boolean && is_y_boolean)
                                || (is_x_number && is_y_number)
                                || (is_x_string && is_y_string)
-                               || (is_x_object && is_y_object));
+                               || (is_x_object && is_y_object)
+#if ENABLED (JERRY_ES2015)
```
rerobika: Cannot we just simple insert the #ifdef -ed condition above the `|| (is_x_object && is_y_object));` line? Therefore no need to put the closing `);` into a new line.

kaadam: Thanks for the review. Updated.

PR #3279, opened 2019-11-04
Add missing GC limit check to realloc with system allocator

dbatyai:


ossy-szeged:
Good catch, informal LGTM.

PR #3291, opened 2019-11-08
Implement String.prototype.codePointAt method

szilagyiadam:
The algorithm is based on ECMA-262 v6, 21.1.3.3


jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -1858,6 +1860,56 @@ ecma_builtin_string_prototype_object_repeat (ecma_string_t *original_string_p, /
...
+{
+  ecma_number_t pos_num;
+  ecma_value_t error = ecma_op_to_integer (pos, &pos_num);
+
+  if (ECMA_IS_VALUE_ERROR (error))
+  {
+    return error;
+  }
+
+  ecma_free_value (error);
```
rerobika: If the result of `toInteger` operation call is not error then the result must be `ECMA_VALUE_EMPTY` direct value. Therefore, releasing this value is unnecessary.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -1858,6 +1860,56 @@ ecma_builtin_string_prototype_object_repeat (ecma_string_t *original_string_p, /
...
+
+  ecma_length_t size = ecma_string_get_length (this_string_p);
+
+  if (pos_num < 0 || pos_num >= size)
+  {
+    return ECMA_VALUE_UNDEFINED;
+  }
+
+  uint32_t index = ecma_number_to_uint32 (pos_num);
```
rerobika: The maximal length of an UTF-8 string only can be UINT32_MAX, so this conversion can be substituted with a single check.

PR #3293, opened 2019-11-08
Added missing logical operator in js-parser-expr

kisbg:
This fix is needed to build JerryScript 2.1 on a STM board. When building for STM board, the compiler reports a warning (conversion warning) by doing a bool check this can be resolved.


rtakacs:
LGTM (Informal)

PR #3310, opened 2019-11-13
Fix timezone in Date.prototype.toString()

ossy-szeged:
Timezone minutes part was always 0 instead of the proper value.
Additionally fixed minor typos in comments.


ossy-szeged:
> The patch looks good to be, but can we add a test case for it?
Unfortunately we can't, because jerry tests are intentionally run always in UTC timezone and there is no way to modify timezone from JS code.

PR #3314, opened 2019-11-13
Make Date.parse() ECMA-262 conform

### LINKED ISSUE #2946: Date.parse() should handle UTC date string format
var timestamp = Date.parse('Wed, 03 Jul 2019 03:15:44 GMT');
console.log(timestamp);
// Result: NaN
// Expected: 1562123744000

ossy-szeged:
Changes:
- Parse output of Date.prototype.toString() and Date.prototype.toUTCString()
- Date.prototype.toString() is ECMA-262 v9 conform now, only TZ part changed
  (Before ECMA-262 v9 it was implementation-dependent.)
- Reused day_names_p and month_names_p arrays (and made them more efficient)
- Tests updated and new tests added

Fixes #2946.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date.c
@@ -97,6 +97,103 @@ ecma_date_parse_special_char (const lit_utf8_byte_t **str_p, /**< pointer to the
 } /* ecma_date_parse_special_char */
 
 /**
+ * Helper function to try to parse a 4-5-6 digit year with optional negative sign in a date string
+ *
+ * Date.prototype.toString() and Date.prototype.toUTCString() emits year
+ * in this format and Date.parse() should parse this format too.
+ *
+ * @return the parsed year or NaN.
+ */
+static ecma_number_t
+ecma_date_parse_year (const lit_utf8_byte_t **str_p, /**< pointer to the cesu8 string */
+                      const lit_utf8_byte_t *str_end_p) /**< pointer to the end of the string */
+{
+  bool is_year_sign_negative = ecma_date_parse_special_char (str_p, str_end_p, '-');
+  const lit_utf8_byte_t *str_start_p = *str_p;
+  int32_t parsed_year = 0;
+
+  while (str_start_p - *str_p <= 6)
+  {
+    if (*str_p >= str_end_p || !lit_char_is_decimal_digit (*str_start_p))
+    {
+      break;
+    }
+
+    parsed_year = 10 * parsed_year + *str_start_p - LIT_CHAR_0;
+    str_start_p++;
+  }
+
+  if (str_start_p - *str_p >=4)
+  {
+    *str_p = str_start_p;
+    if (is_year_sign_negative)
+    {
+      return -parsed_year;
+    }
+    return parsed_year;
+  }
+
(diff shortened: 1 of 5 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-date.c
```diff
@@ -34,6 +34,9 @@
 #define BUILTIN_UNDERSCORED_ID date
 #include "ecma-builtin-internal-routines-template.inc.h"
 
+#define BREAK_IF_FALSE(cond) if (!(cond)) { break; }
+#define BREAK_IF_NAN(number) if (ecma_number_is_nan (number)) { break; }
```
zherczeg: I am not a fond of this define. What about using `if` and `&&`?

rerobika: I second this idea.

ossy-szeged: I can agree that these macros aren't so nice, but I can't find any good looking solution for this parsing. I tried the suggested "if and &&", but in my opinion it is very ugly, confusing, makes debugging hard. I wouldn't like to submit that huge and ugly if. :) You can see how does it look like: https://github.com/ossy-szeged/jerryscript/commit/2fdba92bee25adf119d4beaa0cc90ea8151d544a

Other possible solutions:
1.)  early returns instead of macros with breaks
```
if (!parse_call(...))
{
  return nan;
}
```
Now any parsing call is 1 line, it would make the code 4x longer and redundant.

2.)  nested ifs - soo ugly, because we never use else cases. Not to mention that we would run into coding style limits (maximum line length) and we had to break calls into more lines.
if (parse_call(...)
{
  if (parse_call(...)
  {
    if (parse_call(...)
    {
      ...
    }
  }
}

3.) keep the macros, but using return instead of break
It is essentially same as the 1.) solution, but makes the code as compact as my original idea.
Now I think is better than the original one. What do you think about it?

Do you have any better idea how can we make this parsing function good looking?

dbatyai: I don't think 'good looking' is the goal here. In my opinion the important things are keeping the code readable, and keeping it consistent.

In general I don't like the construct of conditionally doing something in a macro, this in my opinion makes it harder to understand what is happening and in what circumstances. Instead of `DO_X_IF_Y` I would prefer `if (y) { DO_X }` which is a lot clearer, but in that case, we would end up with a single break in the macro.

Speaking of breaks, having control statements in macros is also something I would strongly discourage, this is something I believe also takes away from readability.

In my opinion doing the early returns is the way to go.
Is it longer? Sure, but that is just how Date parsing works, it's a lot of code, with a lot of conditions. Wrapping it with macros to make it more 'compact' does not have an effect on code size, it only takes away from readability.
Is it redundant? I would not consider it any more redundant than what the current code looks like. In fact you would not need an extra condition at the and just to check if parsing was successful. Let the compiler optimize the rest as it sees fit.

ossy-szeged: I "preprocessed" these macros, it made the parse function three pages instead of one. I still don't like how it looks like ("if (!parse(...)) { return nan;"), but if want it, let's do it.

PR #3318, opened 2019-11-14
Introduce parser statement flags

rerobika:
These flags makes the code more readable also makes the process of introducing a new statement type easier.

### DIFF
--- a/jerry-core/parser/js/js-parser-statm.c
+++ b/jerry-core/parser/js/js-parser-statm.c
@@ -43,15 +43,9 @@
  * Parser statement types.
  *
  * When a new statement is added, the following
- * functions may need to be updated as well:
- *
- *  - parser_statement_length()
- *  - parser_parse_break_statement()
- *  - parser_parse_continue_statement()
- *  - parser_free_jumps()
- *  - 'case LEXER_RIGHT_BRACE:' in parser_parse_statements()
- *  - 'if (context_p->token.type == LEXER_RIGHT_BRACE)' in parser_parse_statements()
- *  - 'switch (context_p->stack_top_uint8)' in parser_parse_statements()
+ * arrays must be updated as well:
+ *  - statement_lengths[]
+ *  - parser_statement_flags[]
  */
 typedef enum
 {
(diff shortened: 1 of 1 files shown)


jerry-core/parser/js/js-parser-statm.c
```diff
@@ -84,6 +73,66 @@ typedef enum
...
+  PARSER_STATM_CONTINUE_TARGET = (1 << 2), /**< continue target statement */
+  PARSER_STATM_CONTEXT_BREAK = (1 << 3), /**< uses another instruction form when crosses their borders */
+} parser_statement_flags_t;
+
+/**
+ * Parser statement attributes.
+ * Note: the order of the attributes must be keep in sync with parser_statement_type_t
+ */
+static const uint8_t parser_statement_flags[] =
+{
```
ossy-szeged: I like the idea to remove redundancies and simplify the code base. But it would be great if we could make sure that the enum and this array are in synch.

If we can change the order of enum values, we can do it with macros, for example:

```
#define PARSER_STATEMENT_TYPE(F) \
  F(PARSER_STATEMENT_START, PARSER_STATM_NO_OPTS), \
  F(PARSER_STATEMENT_BLOCK, PARSER_STATM_NO_OPTS), \
...

#define PARSER_STATEMENT_TYPE_ES2015(F) ...

#define EXPAND1(A, B)  A
#define EXPAND2(A, B)  B

typedef enum
{
  PARSER_STATEMENT_TYPE(EXPAND1)
#if ENABLED (JERRY_ES2015)
  PARSER_STATEMENT_TYPE_ES2015(EXPAND1)
} parser_statement_type_t;

static const uint8_t parser_statement_flags[] =
{
  PARSER_STATEMENT_TYPE(EXPAND2)
#if ENABLED (JERRY_ES2015)
  PARSER_STATEMENT_TYPE_ES2015(EXPAND2)
}
```

rerobika: That's a good idea, probably it can be combined with `statement_lengths[]` as well.
@zherczeg What do you think about it?

zherczeg: Separate patch. I want to se how it looks like.

zherczeg: (lines in the length are quite long, and we have 120 char limit)

PR #3320, opened 2019-11-14
Fix the initialization of let/const patterns when block context is needed

rerobika:

### DIFF
--- a/jerry-core/parser/js/js-parser-expr.c
+++ b/jerry-core/parser/js/js-parser-expr.c
@@ -1820,6 +1820,109 @@ parser_process_unary_expression (parser_context_t *context_p, /**< context */
 } /* parser_process_unary_expression */
 
 /**
+ * Append a binary '=' token.
+ */
+static void
+parser_append_binary_single_assignment_token (parser_context_t *context_p, /**< context */
+                                              uint8_t assign_ident_opcode) /**< assign ident opcode */
+{
+  JERRY_ASSERT (assign_ident_opcode == CBC_ASSIGN_SET_IDENT
+                || assign_ident_opcode == CBC_ASSIGN_LET_CONST);
+
+  /* Unlike other tokens, the whole byte code is saved for binary
+   * assignment, since it has multiple forms depending on the
+   * previous instruction. */
+
+  if (PARSER_IS_PUSH_LITERAL (context_p->last_cbc_opcode)
+      && context_p->last_cbc.literal_type == LEXER_IDENT_LITERAL)
+  {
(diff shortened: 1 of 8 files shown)


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -1820,108 +1820,122 @@ parser_process_unary_expression (parser_context_t *context_p, /**< context */
 } /* parser_process_unary_expression */
 
 /**
- * Append a binary token.
+ * Append a binary '=' token.
  */
 static void
-parser_append_binary_token (parser_context_t *context_p) /**< context */
+parser_append_binary_assigment_token (parser_context_t *context_p, /**< context */
```
zherczeg: assignment

zherczeg:
Let me start with some general comments:

1) This should be redeclaration error:

`let a, { "x": a } = { x:4 };`
`Script Error: SyntaxError: Invalid destructuring assignment target.`

I suspect this check is too early here:

https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/parser/js/js-parser-expr.c#L642

2) We are not too consistent on error messages:

```
let a, { x: b.c } = {x:6};
~~~~~~~~~~~~~^
Script Error: SyntaxError: Expected ',' or '}' after a property definition.
```

Note: an assignment (=) is also allowed here

```
let a, [ b.c ] = [6];
~~~~~~~~~~^
Script Error: SyntaxError: Invalid destructuring assignment target.
```

3) We have a test:
`checkSyntax ("[()] = []");`

but I would also add tests for these:
`checkSyntax ("let [(a)] = [1]");`
`checkSyntax ("let {a:(a)} = {a:1}");`

These test cases would be useful as well:
```
var a = 6;
[((a))] = [7];
assert(a === 7);
```

This should work but we produce a syntax error:

```
var a = 6;
({"a":((a))} = {a:7});
~~~~~~^
Script Error: SyntaxError: Identifier expected.
```

rerobika:
@zherczeg Thanks for the review, all the related issues have been resolved.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -2097,7 +2116,8 @@ parser_pattern_get_target (parser_context_t *context_p, /**< context */
     scanner_location_t start_location;
 
     if (context_p->next_scanner_info_p->source_p != context_p->source_p
-        || context_p->next_scanner_info_p->type == SCANNER_TYPE_ERR_REDECLARED)
+        || context_p->next_scanner_info_p->type == SCANNER_TYPE_ERR_REDECLARED
+        || flags & PARSER_PATTERN_REST_ELEMENT)
```
zherczeg: put brackets around &

PR #3321, opened 2019-11-15
Implement IsRegExp operation

szilagyiadam:
Algorithm is based on ECMA-262 v6, 22.1.3.1.1

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
@@ -66,6 +66,10 @@ ecma_builtin_regexp_dispatch_construct (const ecma_value_t *arguments_list_p, /*
 {
   ecma_value_t pattern_value = ECMA_VALUE_UNDEFINED;
   ecma_value_t flags_value = ECMA_VALUE_UNDEFINED;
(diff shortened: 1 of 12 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -78,8 +78,20 @@ ecma_builtin_regexp_dispatch_construct (const ecma_value_t *arguments_list_p, /*
     }
   }
 
-  if (ecma_is_value_object (pattern_value)
-      && ecma_object_class_is (ecma_get_object_from_value (pattern_value), LIT_MAGIC_STRING_REGEXP_UL))
+#if ENABLED (JERRY_ES2015)
+  ecma_value_t regexp_value = ecma_op_is_regexp (pattern_value);
```
dbatyai: If we add this, then we should also implement handling RegExp-like objects as arguments, which means creating a builtin RegExp instance from the properties of the argument object, as described in 21.2.3.1/6.

rerobika:
Please rebase the PR to the latest master.

rerobika:
The Appveyor build failed due to an internal resource unavailable error in one job which is unrelated to the content of the PR.

PR #3322, opened 2019-11-15
Implement function destructuring argument support.

zherczeg:
Furthermore create unmapped arguments objects if a function has a non-simple argument. A few destructuring pattern issues were fixed as well.


zherczeg:
#3320 needs to land first.

PR #3327, opened 2019-11-18
Implement binding pattern support for rest argument and for statement.

zherczeg:


zherczeg:
@dbatyai @rerobika this patch was quite simple. Shall I merge it with #3322 ?

PR #3329, opened 2019-11-19
Implement spread operator for function call arguments

rerobika:


zherczeg:
I don't think spread objects are neededso optimize this later. Otherwise ok.

PR #3332, opened 2019-11-19
Fix fast array array hole initialization

rerobika:
This patch eliminates the valgrind error: "Conditional jump or move depends on uninitialised value(s)".


rerobika:
The CI [passed](https://travis-ci.org/jerryscript-project/jerryscript/builds/613977708?utm_source=github_status&utm_medium=notification) but the check status is still pending (probably a glitch).

PR #3334, opened 2019-11-19
Import Kangax ES5/ES6 compatibilty tests

ossy-szeged:
Changes:
- Import tests with actual expectations from https://github.com/kangax/compat-table
- Add --kangax-tests option to tools/run-tests.py test runner
- Run Kangax tests on Travis CI


rerobika:
Cannot we use these tests as a submodule instead of copying the whole test suite? IMO we do the same for test262 test as well.

ossy-szeged:
> 
> 
> Cannot we use these tests as a submodule instead of copying the whole test suite? IMO we do the same for test262 test as well.

It's not a simple copying of the test suite. I extracted the test suite and adapted to our test 
runner infrastructure. All kangax tests are in only one JS object with all expected results, see:
https://github.com/kangax/compat-table/blob/gh-pages/data-es6.js

Kangax has a node.js based runner to extract a test from that JS object, wrap it into a function, collect the results and compare with the expected results of the actual released version of the engine.

I would like to import these tests because it's the easiest way to maintain the expected results in synch with the master. If you implement a new ES6 feature, run tests, move new passign tests from es6/fail to es6 directory. Additionally it makes easy to catch regressions in time (before merging bugs).

Hexxeh:
The output of this seems like it'd be a good addition to the wiki to show how feature complete the engine is for ES6.

ossy-szeged:
JerryScript 2.3.0 already has 95% Kangax ES6 support, there is no need to add them to the repository anymore.

PR #3339, opened 2019-11-20
Improve RegExp compatibility with web browsers

galpeter:
The modification adds support to parse /A{/ like RegExps.
That is: if the iterator is invalid it should be treated as normal
characters.

This behaviour is defined in the ES2015 standard Annex B 1.4 point

This only works if the `JERRY_REGEXP_STRICT_MODE` is disabled
(set to zero/OFF).

### DIFF
--- a/jerry-core/parser/regexp/re-parser.c
+++ b/jerry-core/parser/regexp/re-parser.c
@@ -630,11 +630,22 @@ re_parse_next_token (re_parser_ctx_t *parser_ctx_p, /**< RegExp parser context *
     {
       return ecma_raise_syntax_error (ECMA_ERR_MSG ("Invalid RegExp token."));
     }
+    case LIT_CHAR_NULL:
+    {
+      out_token_p->type = RE_TOK_EOF;
+      break;
+    }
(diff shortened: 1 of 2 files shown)


jerry-core/parser/regexp/re-parser.c
```diff
@@ -630,11 +630,22 @@ re_parse_next_token (re_parser_ctx_t *parser_ctx_p, /**< RegExp parser context *
...
+    }
     case LIT_CHAR_LEFT_BRACE:
     {
 #if ENABLED (JERRY_REGEXP_STRICT_MODE)
       return ecma_raise_syntax_error (ECMA_ERR_MSG ("Invalid RegExp token."));
 #else /* !ENABLED (JERRY_REGEXP_STRICT_MODE) */
+
+      /* Make sure that the current '{' does not start an iterator.
+       *
+       * E.g: /\s+{3,4}/ should fail as there is noting to iterate.
```
rerobika: typo `noting`.

galpeter:
@rerobika Thanks! I've updated the PR (also added an extra test case).

PR #3340, opened 2019-11-21
Support get/set as property name in object literal

rerobika:

### DIFF
--- a/jerry-core/parser/js/js-lexer.c
+++ b/jerry-core/parser/js/js-lexer.c
@@ -2472,6 +2472,7 @@ lexer_expect_object_literal_id (parser_context_t *context_p, /**< context */
 #if ENABLED (JERRY_ES2015)
             && context_p->source_p[0] != LIT_CHAR_COMMA
             && context_p->source_p[0] != LIT_CHAR_RIGHT_BRACE
+            && context_p->source_p[0] != LIT_CHAR_LEFT_PAREN
 #endif /* ENABLED (JERRY_ES2015) */
             && context_p->source_p[0] != LIT_CHAR_COLON)
         {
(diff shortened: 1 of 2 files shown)


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -791,11 +791,21 @@ parser_parse_object_literal (parser_context_t *context_p) /**< context */
         uint32_t status_flags;
         cbc_ext_opcode_t opcode;
         uint16_t literal_index, function_literal_index;
+        bool is_getter = context_p->token.type == LEXER_PROPERTY_GETTER;
```
zherczeg: There is no need to save this value, `lexer_check_next_character` does not change it.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -791,11 +791,21 @@ parser_parse_object_literal (parser_context_t *context_p) /**< context */
...
 #if !ENABLED (JERRY_ES2015)
         parser_object_literal_item_types_t item_type;
+#else /* ENABLED (JERRY_ES2015) */
+        if (lexer_check_next_character (context_p, LIT_CHAR_LEFT_PAREN))
+        {
+          lexer_construct_literal_object (context_p,
+                                          (is_getter ? (lexer_lit_location_t *) &lexer_get_literal
+                                                     : (lexer_lit_location_t *) &lexer_set_literal),
+                                          LEXER_STRING_LITERAL);
+          goto parse_object_method;
```
zherczeg: Can we avoid this goto?

rerobika: Sure! I've reordered the switch cases to substitute the goto with a fallthru.

zherczeg:
Just a question: would it be better to add `(` here:
https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/parser/js/js-lexer.c#L2474

PR #3345, opened 2019-11-22
Implement Regexp.prototype[@@match] method

szilagyiadam:
The algorithm is based on ECMA-262 v6, 21.2.5.6

The following helper methods are also implemented:
  - RegExpExec: ECMA-262 v6, 21.2.5.2.1
  - AdvanceStringIndex: ECMA-262 v6, 21.2.5.2.3

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-string.c
+++ b/jerry-core/ecma/base/ecma-helpers-string.c
@@ -2799,6 +2799,55 @@ ecma_stringbuilder_destroy (ecma_stringbuilder_t *builder_p) /**< string builder
 #endif /* ENABLED (JERRY_MEM_STATS) */
 } /* ecma_stringbuilder_destroy */
 
+#if ENABLED (JERRY_ES2015)
+/**
+ * AdvanceStringIndex operation
+ *
+ * See also:
+ *          ECMA-262 v6.0, 21.2.5.2.3
+ *
+ * @return uint32_t - the proper character index based on the operation
+ */
+uint32_t
+ecma_op_advance_string_index (ecma_string_t *str_p, /**< input string */
+                              uint32_t index, /**< given character index */
+                              bool is_unicode) /**< true - if regexp object's "unicode" flag is set
+                                                    false - otherwise */
+{
+  JERRY_ASSERT (index < UINT32_MAX - 1);
+
+  uint32_t next_index = index + 1;
+
+  if (!is_unicode)
+  {
+    return next_index;
+  }
+
+  ecma_length_t str_len = ecma_string_get_length (str_p);
+
+  if (next_index >= str_len)
+  {
+    return next_index;
+  }
+
+  ecma_char_t first = ecma_string_get_char_at_pos (str_p, index);
+
+  if (first < LIT_UTF16_HIGH_SURROGATE_MIN || first > LIT_UTF16_HIGH_SURROGATE_MAX)
+  {
+    return next_index;
+  }
+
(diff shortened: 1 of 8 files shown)


jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2799,6 +2799,55 @@ ecma_stringbuilder_destroy (ecma_stringbuilder_t *builder_p) /**< string builder
...
 } /* ecma_stringbuilder_destroy */
 
+#if ENABLED (JERRY_ES2015)
+/**
+ * AdvanceStringIndex operation
+ *
+ * See also:
+ *          ECMA-262 v6.0, 21.2.5.2.3
+ *
+ * @return uint32_t
```
galpeter: What does the `uint32_t` represent here? Could we add a bit more info on it?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -2011,6 +2011,65 @@ ecma_regexp_replace_helper (ecma_value_t this_arg, /**< this argument */
...
+    ecma_object_t *function_p = ecma_get_object_from_value (exec);
+    ecma_value_t arguments[] = { ecma_make_string_value (str_p) };
+
+    ecma_value_t result = ecma_op_function_call (function_p, this_arg, arguments, 1);
+
+    ecma_deref_object (function_p);
+
+    if (ECMA_IS_VALUE_ERROR (result))
+    {
+      return result;
```
galpeter: Shouldn't the `exec` be freed/released before returning here?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -2011,6 +2011,65 @@ ecma_regexp_replace_helper (ecma_value_t this_arg, /**< this argument */
...
+    }
+
+    if (!ecma_is_value_object (result) && !ecma_is_value_null (result))
+    {
+      ecma_free_value (result);
+      return ecma_raise_type_error (ECMA_ERR_MSG ("Return value of 'exec' must be an Object or Null"));
```
galpeter: Shouldn't the `exec` be freed/released before returning here?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -2011,6 +2011,65 @@ ecma_regexp_replace_helper (ecma_value_t this_arg, /**< this argument */
...
+    }
+
+    return result;
```
galpeter: Shouldn't the `exec` be freed/released before returning here?

tests/jerry/es2015/regexp-prototype-match.js
```diff
@@ -0,0 +1,56 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var regexp = /[0-9]+/g;
+var str = '2016-01-02';
+var num = 12131233;
+
+// Test with string input
+var result = regexp[Symbol.match](str);
```
galpeter: Is this functionality only accessible via the `Symbol.match`?

galpeter: In addition this functionality means that if someone creates a "subclass" of RegExp where the `Symbol.match` or `exec`(?) is implemented the "new" method would be invoked instead of the `Regexp.prototype.exec`? If so would it be possible to have a test case for it?

Issue #3360, opened 2019-11-23
Assertion context_p->next_scanner_info_p->type == SCANNER_TYPE_FUNCTION in parser_parse_function_arguments


renatahodovan:
###### JerryScript revision
62356796

###### Build platform
Linux-4.15.0-62-generic-x86_64-with-Ubuntu-18.04-bionic

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```

###### Test case
```javascript
function $ (b = [ ].$) { }
var $ = { $: function ( ) { } } 
```

###### Output
```text
ICE: Assertion 'context_p->next_scanner_info_p->type == SCANNER_TYPE_FUNCTION' failed at jerryscript/jerry-core/parser/js/js-parser.c(parser_parse_function_arguments):1621.
Error: ERR_FAILED_INTERNAL_ASSERTION
```

###### Backtrace
```text
bt
#0  0xf7fd5079 in __kernel_vsyscall ()
#1  0xf77fc832 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf77fdcc1 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0x5657b78e in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-port/default/default-fatal.c:30
#4  0x56607f14 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x56607f55 in jerry_assert_fail (assertion=0x566d5400 "context_p->next_scanner_info_p->type == SCANNER_TYPE_FUNCTION", file=0x566d4ae0 "jerryscript/jerry-core/parser/js/js-parser.c", function=0x566ae580 <__func__.5658.lto_priv.506> "parser_parse_function_arguments", line=1621) at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x5661378e in parser_parse_function_arguments (context_p=0xffffc4c0, end_type=LEXER_RIGHT_PAREN) at jerryscript/jerry-core/parser/js/js-parser.c:1621
[...]

PR #3367, opened 2019-11-25
Fix the scanning of default array/object literals.

rerobika:
Scanning should continue with `SCAN_MODE_POST_PRIMARY_EXPRESSION` in all cases.


rerobika:
Fixes #3360 .

PR #3369, opened 2019-11-26
Fix the order of the function arguments for spread operation

rerobika:

### DIFF
--- a/jerry-core/parser/js/js-parser-expr.c
+++ b/jerry-core/parser/js/js-parser-expr.c
@@ -1662,7 +1662,7 @@ parser_process_unary_expression (parser_context_t *context_p, /**< context */
         lexer_next_token (context_p);
 
 #if ENABLED (JERRY_ES2015)
-        uint32_t spread_arguments = 0;
+        bool has_spread_element = false;
 #endif /* ENABLED (JERRY_ES2015) */
 
         if (context_p->token.type != LEXER_RIGHT_PAREN)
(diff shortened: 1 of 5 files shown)


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -1675,23 +1675,17 @@ parser_process_unary_expression (parser_context_t *context_p, /**< context */
             }
 
 #if ENABLED (JERRY_ES2015)
-            lexer_token_type_t type = context_p->token.type;
-            if (type == LEXER_THREE_DOTS)
+            if (context_p->token.type == LEXER_THREE_DOTS)
             {
-              spread_arguments++;
+              has_spread_element = true;
+              call_arguments++;
```
zherczeg: Does this divide the maximum number of arguments by 2?

rerobika: Currently (127 spreaded arguments) in the worst case it does. However the usage of spread operation in function argument list serves the purpose of reducing function arguments by packing them into iterable object. IMO it's fine for now.

zherczeg: What about adding another byte argument to the call instructions which contains the number of spread opcodes?

rerobika: IMO, currently only 1 byte number argument is available for CBC opcodes.

zherczeg: (I can work on it later if you prefer that)

rerobika: That could be useful later. Moreover, the maximum number of arguments is not defined in the standard.

PR #3373, opened 2019-11-28
Allow to create ArrayBuffer with empty external user data

rtakacs:
This patch helps to unify the behavior of `jerry_create_arraybuffer(0)` and `jerry_create_arraybuffer_external(0, NULL, NULL)`. The latter one is useful on systems where the ArrayBuffer creation is based on network communication. For example, if there is no received data (empty external user data), an empty ArrayBuffer could be created.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -3707,7 +3707,7 @@ jerry_create_arraybuffer_external (const jerry_length_t size
 - `free_cb` - the callback function called when the object is released
 - return value
   - the new ArrayBuffer as a `jerry_value_t`
-  - if the `size` is zero or `buffer_p` is a null pointer will return RangeError
+  - if the `size` is zero or `buffer_p` is a null pointer this will return an empty ArrayBuffer.
 
 *New in version 2.0*.
 
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -3446,7 +3446,7 @@ jerry_create_arraybuffer (const jerry_length_t size) /**< size of the ArrayBuffe
  *     * the size is specified in bytes.
  *     * the buffer passed should be at least the specified bytes big.
  *     * if the typed arrays are disabled this will return a TypeError.
- *     * if the size is zero or the buffer_p is a null pointer this will return a RangeError.
+ *     * if the size is zero or buffer_p is a null pointer this will return an empty ArrayBuffer.
  *
  * @return value of the construced ArrayBuffer object
  */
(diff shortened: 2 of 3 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -3707,7 +3707,6 @@ jerry_create_arraybuffer_external (const jerry_length_t size
 - `free_cb` - the callback function called when the object is released
 - return value
   - the new ArrayBuffer as a `jerry_value_t`
-  - if the `size` is zero or `buffer_p` is a null pointer will return RangeError
```
galpeter: This change should be documented here, that this was modified.

jerry-core/api/jerry.c
```diff
@@ -3458,14 +3457,19 @@ jerry_create_arraybuffer_external (const jerry_length_t size, /**< size of the b
   jerry_assert_api_available ();
 
 #if ENABLED (JERRY_ES2015_BUILTIN_TYPEDARRAY)
-  if (size == 0 || buffer_p == NULL)
+  ecma_object_t *arraybuffer;
+
+  if (JERRY_UNLIKELY (size == 0 || buffer_p == NULL))
+  {
+    arraybuffer = ecma_arraybuffer_new_object_external (0, NULL, NULL);
```
galpeter: Is it ok to not register the free callback?

rtakacs: You are right. It's good to know when the ArrayBuffer is garbage collected.

jerry-core/api/jerry.c
```diff
@@ -3458,14 +3458,19 @@ jerry_create_arraybuffer_external (const jerry_length_t size, /**< size of the b
...
+  if (JERRY_UNLIKELY (size == 0 || buffer_p == NULL))
+  {
+    arraybuffer = ecma_arraybuffer_new_object_external (0, NULL, free_cb);
+  }
+  else
   {
-    return jerry_throw (ecma_raise_range_error (ECMA_ERR_MSG ("invalid buffer size or storage reference")));
+    arraybuffer = ecma_arraybuffer_new_object_external (size,
+                                                        buffer_p,
+                                                        (ecma_object_native_free_callback_t) free_cb);
```
zherczeg: Why this casting is needed? I know it is in the old code, still :)

rtakacs: The type of the cb parameter is `jerry_object_native_free_callback_t` while `ecma_arraybuffer_new_object_external` expects `ecma_object_native_free_callback_t` typed cb. Both type is `void*`. Would you like to remove this casting?

zherczeg: A line above you wrote:
`arraybuffer = ecma_arraybuffer_new_object_external (0, NULL, free_cb);`
Based on this, the cast is unneeded.

rtakacs: I see. I've added missing conversion to the `arraybuffer = ecma_arraybuffer_new_object_external (0, NULL, free_cb);` line.

PR #3375, opened 2019-11-28
Update String.prototype.match to ECMA-262 v6

szilagyiadam:
The algorithm is based on ECMA-262 v6, 21.1.3.11


jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -317,180 +318,125 @@ static ecma_value_t
...
-  ret_value = ECMA_VALUE_ERROR;
+  ecma_value_t func_value = ecma_op_object_get_by_symbol_id (new_regexp_obj, LIT_MAGIC_STRING_MATCH);
 
-  /* 8.f. */
-  while (last_match)
+  if (ECMA_IS_VALUE_ERROR (func_value))
   {
-    /* 8.f.i. */
-    ecma_value_t exec_value = ecma_regexp_exec_helper (regexp_value, this_to_string_value, false);
+    goto cleanup;
```
galpeter: Would it be possible to avoid `goto`s here?

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -317,180 +318,125 @@ static ecma_value_t
...
+  }
 
-    if (ECMA_IS_VALUE_ERROR (exec_value))
-    {
-      break;
-    }
+  if (!ecma_op_is_callable (func_value))
+  {
+    result = ecma_raise_type_error (ECMA_ERR_MSG ("@@match is not callable."));
+    goto cleanup;
```
galpeter: and here also?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -2011,6 +2011,190 @@ ecma_regexp_replace_helper (ecma_value_t this_arg, /**< this argument */
...
+      uint32_t next_index = ecma_op_advance_string_index (str_p, index, full_unicode);
+
+      ecma_value_t next_set_status = ecma_op_object_put (obj_p,
+                                                         ecma_get_magic_string (LIT_MAGIC_STRING_LASTINDEX_UL),
+                                                         ecma_make_uint32_value (next_index),
+                                                         true);
+#else /* !ENABLED (JERRY_ES2015) */
+      ecma_value_t next_set_status = ecma_op_object_put (obj_p,
+                                                         ecma_get_magic_string (LIT_MAGIC_STRING_LASTINDEX_UL),
+                                                         ecma_make_number_value (this_index + 1),
```
dbatyai: `this_index` is an ecma_value_t, it should not be incremented.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -317,180 +318,129 @@ static ecma_value_t
...
-    /* 8.g. */
-    ret_value = ECMA_VALUE_NULL;
+    ecma_deref_ecma_string (pattern_p);
+    return new_regexp;
   }
-  else
+
+  ecma_value_t result = ecma_regexp_match_helper (new_regexp, this_to_string_value);
+
+  if (ECMA_IS_VALUE_ERROR (result))
```
dbatyai: I don't think we need an error check here, we would just return with the result anyway.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -317,180 +318,129 @@ static ecma_value_t
...
-                                                         ecma_get_magic_string (LIT_MAGIC_STRING_LASTINDEX_UL),
-                                                         ecma_make_number_value (this_index + 1),
-                                                         true);
-      if (ECMA_IS_VALUE_ERROR (index_put_value))
-      {
-        ecma_free_value (this_index_value);
-        ecma_free_number (this_index_number);
-        goto cleanup;
-      }
+  if (ecma_is_value_undefined (regexp_arg))
```
dbatyai: You could use `ecma_regexp_read_pattern_str_helper` which handles the toString conversion and the undefined argument.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -317,180 +318,129 @@ static ecma_value_t
...
-    ecma_deref_ecma_string (this_string_p);
-    ecma_deref_object (regexp_obj_p);
-    return ret_value;
+    return ECMA_VALUE_ERROR;
   }
 
-  /* 8.a. */
-  ecma_string_t *index_zero_string_p = ecma_get_ecma_string_from_uint32 (0);
+  ecma_value_t regexp_construct_argument[] = { regexp_arg };
+  ecma_value_t new_regexp = ecma_builtin_regexp_dispatch_construct (regexp_construct_argument, 1);
```
dbatyai: The RegExp construction code could be shared between ES5.1 and ES6. Could you change the guards so that code is not duplicated? Preferably by keeping the current ES5.1 code path.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -317,180 +318,88 @@ static ecma_value_t
...
+  }
 
-    ecma_free_value (match_string_value);
-    ecma_free_value (this_index_value);
-    ecma_free_number (this_index_number);
+  ecma_object_t *func_obj = ecma_get_object_from_value (func_value);
 
-    ecma_deref_object (exec_obj_p);
-  }
+  ecma_string_t *str_p = ecma_op_to_string (this_to_string_value);
```
dbatyai: We should check for error here.

PR #3378, opened 2019-11-28
Remove CBC_EXT_CONTINUE_EXEC opcode.

zherczeg:
#3372 should land first.


jerry-core/ecma/builtin-objects/ecma-builtin-generator-prototype.c
```diff
@@ -83,7 +109,14 @@ ecma_builtin_generator_prototype_object_do (ecma_value_t this_arg, /**< this arg
...
 
-  executable_object_p->frame_ctx.call_operation = yield_mode;
+  if (resume_mode == ECMA_GENERATOR_RETURN)
+  {
+    executable_object_p->frame_ctx.byte_code_p = ecma_builtin_generator_prototype_return;
+  }
+  else if (resume_mode == ECMA_GENERATOR_THROW)
+  {
+    executable_object_p->frame_ctx.byte_code_p = ecma_builtin_generator_prototype_throw;
+  }
```
galpeter: Is it guaranteed here that only these two cases can occur? Shouldn't we check the `NEXT` case also here?

zherczeg: No need to do anything in the next case. And the enum has no more values.

PR #1896, opened 2017-06-28
Rework argument parsing.

zherczeg:
Heavily simplify the interface of CLI.

PR #3384, opened 2019-11-29
Make REPL option parser handle "-" option properly

ossy-szeged:
REPL could read JS file from standard input when we passed "-" option,
but #1809 missed to add this function to the new option parser.

### DIFF
--- a/jerry-main/cli.c
+++ b/jerry-main/cli.c
@@ -112,7 +112,7 @@ cli_consume_option (cli_state_t *state_p) /**< state of the command line option 
 
   state_p->arg = arg;
 
-  if (arg[0] != '-')
+  if (arg[0] != '-' || (arg[0] == '-' && arg[1] == '\0'))
   {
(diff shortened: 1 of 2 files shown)


akosthekiss:
For the records, this code has been reworked after #1809 in e6832f5f276c1e1acc7c7111a30d526b65223d81 (cannot find the PR for that). The original CLI parser was prepared for "-", I'm not taking responsibility for this variant.

ossy-szeged:
> For the records, this code has been reworked after #1809 in [e6832f5](https://github.com/jerryscript-project/jerryscript/commit/e6832f5f276c1e1acc7c7111a30d526b65223d81) (cannot find the PR for that). The original CLI parser was prepared for "-", I'm not taking responsibility for this variant.

I wasn't checked the hisotry thorough enough, but you're right, #1896 broke this functionality. I updated the commit log accordingly.

PR #3401, opened 2019-12-02
Add iterator close support for for-of statement

rerobika:

### DIFF
--- a/jerry-core/vm/vm-stack.c
+++ b/jerry-core/vm/vm-stack.c
@@ -18,6 +18,7 @@
 #include "ecma-helpers.h"
 #include "vm-defines.h"
 #include "vm-stack.h"
+#include "ecma-iterator-object.h"
 
 /** \addtogroup vm Virtual machine
  * @{
@@ -79,8 +80,15 @@ vm_stack_context_abort (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
 #if ENABLED (JERRY_ES2015)
     case VM_CONTEXT_FOR_OF:
     {
+      ecma_value_t iterator = vm_stack_top_p[-3];
+
+      if (context_info & VM_CONTEXT_CLOSE_ITERATOR)
+      {
+        ecma_op_iterator_close (iterator);
+      }
+      ecma_free_value (iterator);
+
       ecma_free_value (vm_stack_top_p[-2]);
-      ecma_free_value (vm_stack_top_p[-3]);
       VM_MINUS_EQUAL_U16 (frame_ctx_p->context_depth, PARSER_FOR_OF_CONTEXT_STACK_ALLOCATION);
       vm_stack_top_p -= PARSER_FOR_OF_CONTEXT_STACK_ALLOCATION;
       break;
--- a/jerry-core/vm/vm-stack.h
+++ b/jerry-core/vm/vm-stack.h
@@ -28,13 +28,13 @@
 /**
  * Create context on the vm stack.
  */
-#define VM_CREATE_CONTEXT(type, end_offset) ((ecma_value_t) ((type) | ((end_offset) << 6)))
+#define VM_CREATE_CONTEXT(type, end_offset) ((ecma_value_t) ((type) | ((end_offset) << 7)))
 
 /**
  * Create context on the vm stack with environment.
  */
 #define VM_CREATE_CONTEXT_WITH_ENV(type, end_offset) \
-  ((ecma_value_t) ((type) | ((end_offset) << 6) | VM_CONTEXT_HAS_LEX_ENV))
+  (VM_CREATE_CONTEXT ((type),(end_offset)) | VM_CONTEXT_HAS_LEX_ENV)
 
 /**
  * Get type of a vm context.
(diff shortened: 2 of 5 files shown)


tests/jerry/es2015/for-of.js
```diff
@@ -119,3 +119,109 @@ var i = 0;
 for (var a of obj) {
   assert (a === i++);
 }
+
+function __createIterableObject(arr, methods) {
```
zherczeg: Why two underscore is needed?

tests/jerry/es2015/for-of.js
```diff
@@ -119,3 +119,109 @@ var i = 0;
...
+  methods = methods || {};
```
zherczeg: You can use a default parameter

jerry-core/vm/vm-stack.c
```diff
@@ -79,8 +80,11 @@ vm_stack_context_abort (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
 #if ENABLED (JERRY_ES2015)
     case VM_CONTEXT_FOR_OF:
     {
+      ecma_value_t iterator = vm_stack_top_p[-3];
       ecma_free_value (vm_stack_top_p[-2]);
-      ecma_free_value (vm_stack_top_p[-3]);
+
+      ecma_op_iterator_close (iterator);
```
zherczeg: What are the possible side effects?

tests/jerry/es2015/for-of.js
rerobika: This method comes from the `kangax` test suite, therefore I left the original source.

jerry-core/vm/vm-stack.c
rerobika: This operation may changes the `JERRY_CONTEXT (error_value)` and may applies the `ECMA_STATUS_EXCEPTION` status flags. At this point there is no way to handle exception, so the caller's responsibility to check whether the context abort raised error. See [the example](https://github.com/jerryscript-project/jerryscript/pull/3401/files#diff-4eae7d69004d99f61986f6e351bda007R3632) here.

tests/jerry/es2015/for-of.js
zherczeg: What about the license?

jerry-core/vm/vm-stack.c
zherczeg: What happens if multiple iterators are closed? Nested `for-if`-s. Are they working correctly if their closes throws errors?

rerobika: The iterator close operation may raises error but it never overwrites the original error which caused the context abort. Each context abort is called, see the example [here](https://github.com/jerryscript-project/jerryscript/pull/3401/files#diff-bf7b2eb18cb71828704e5cc5bad6a567R211).

tests/jerry/es2015/for-of.js
rerobika: [These methods](https://github.com/kangax/compat-table/blob/gh-pages/node.js#L29) only helper functions to run the test suite but useful for testing purposes. I see no license marker from the origin of the file.

akosthekiss: Beware! Here be dragons! The fact that the file has no license header doesn't mean that it is not covered by a license. (If it had no license at all, that would be the most dangerous situation, actually. That would disallow its reuse anywhere!)

Fortunately, the project repo does contain a [license](https://github.com/kangax/compat-table/blob/gh-pages/LICENSE), which covers all files, including the source reused here. One small problem remains: the git logs claim that this is the MIT license, but actually it isn't: it is some kind of MIT--. The MIT license would contain the condition "The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software", but that's missing from this license. 

Nevertheless, I think it is safe to assume the unmodified MIT license as that's more strict. But then, we should replicate the MIT license in our code. There is example for that, e.g., in jerry-libm.

Some background info on MIT license:
- https://opensource.org/licenses/MIT
- https://tldrlegal.com/license/mit-license

rerobika: @akosthekiss Thanks for the clarification, I've missed that part.

rerobika:
@zherczeg The patch is updated, please take an other look at it.

jerry-core/vm/vm-stack.h
```diff
@@ -44,13 +44,20 @@
...
  */
 #define VM_CONTEXT_HAS_LEX_ENV 0x20
 
+
+/**
+ * This flag is set if the iterator close operation should be invoked during a for-of context break.
+ */
+#define VM_CONTEXT_CLOSE_ITERATOR 0x40
+
+
```
zherczeg: One newline.

jerry-core/vm/vm-stack.h
```diff
@@ -28,13 +28,13 @@
 /**
  * Create context on the vm stack.
  */
-#define VM_CREATE_CONTEXT(type, end_offset) ((ecma_value_t) ((type) | ((end_offset) << 6)))
+#define VM_CREATE_CONTEXT(type, end_offset) ((ecma_value_t) ((type) | ((end_offset) << 8)))
```
zherczeg: Seems 7 is enough. I would not increase it to 8, because it might be misleading.

rerobika:
@zherczeg Thanks for the review, I've updated the patch.

tests/jerry/es2015/for-of-iterator-close.js
```diff
@@ -0,0 +1,244 @@
...
+  iterable['return'] = methods['return'];
+  iterable['throw'] = methods['throw'];
+
+  return iterable;
+};
+
+function close1() {
+  var closed = false;
+  var iter = createIterable([1, 2, 3], {
+    'return': function(){ closed = true; return {}; }
```
zherczeg: I would put space after `function()` here, and below.

rerobika: Sure! Resolved.

PR #3403, opened 2019-12-03
Implement ArrayBuffer.isView function

rtakacs:


jerry-core/ecma/builtin-objects/ecma-builtin-arraybuffer.c
```diff
@@ -55,11 +55,8 @@ ecma_builtin_arraybuffer_object_is_view (ecma_value_t this_arg, /**< 'this' argu
                                          ecma_value_t arg) /**< argument 1 */
 {
   JERRY_UNUSED (this_arg);
-  JERRY_UNUSED (arg);
 
-  /* TODO: if arg has [[ViewArrayBuffer]], return true */
-
-  return ECMA_VALUE_FALSE;
+  return jerry_create_boolean (jerry_value_is_typedarray (arg) || jerry_value_is_dataview (arg));
```
galpeter: This is a layering violation, `jerry_` methods should not be used inside the `ecma_` layer. There are `ecma` methods to achieve the same thing.

rtakacs: Yeah, that's true. Fixed.

PR #3405, opened 2019-12-03
Implement Symbol.unscopables

szilagyiadam:
Added a new property for Array.prototype based on ECMA-262 v6, 22.1.3.31
Also upgraded the HasBinding operation with ECMA-262 v6, 8.1.1.2.1 steps 7-9


dbatyai:
I believe the order is not quite right. It should be checked first whether the binding exists, and only if it does should we acces the unscopables property.

For example the result of the following should be a reference error, and not the value thrown by the unscopables accessor:
```
var o = {};
Object.defineProperty(o, Symbol.unscopables, { get: function () { throw 42; } });
with (o) {
  prop;
}
```

szilagyiadam:
> I believe the order is not quite right. It should be checked first whether the binding exists, and only if it does should we acces the unscopables property.
> 
> For example the result of the following should be a reference error, and not the value thrown by the unscopables accessor:
> 
> ```
> var o = {};
> Object.defineProperty(o, Symbol.unscopables, { get: function () { throw 42; } });
> with (o) {
>   prop;
> }
> ```
I've changed the order, now we first check if the binding exists, and then we acces the unscopables property, so the code above returns reference error.

jerry-core/ecma/operations/ecma-reference.c
```diff
@@ -132,38 +183,66 @@ ecma_op_resolve_reference_value (ecma_object_t *lex_env_p, /**< starting lexical
       ecma_object_t *binding_obj_p = ecma_get_lex_env_binding_object (lex_env_p);
 
 #if ENABLED (JERRY_LCACHE)
-      ecma_property_t *property_p = ecma_lcache_lookup (binding_obj_p, name_p);
-
-      if (property_p != NULL)
+#if ENABLED (JERRY_ES2015)
+      /* Global environment record does not have withEnvironment flag,
+         therefore the direct lcache lookup can be performed.  */
+      if (lex_env_p == ecma_get_global_environment ())
```
galpeter: If I understand correctly only the `if` check is for ES2015 any other thing is the same in all profiles?
If so what if we do something like this?:
```c
#if ENABLED(..)
bool lcache_lookup_allowed = (lex_env_p == ecma_get_global_environment ());
#else
bool lcache_lookup_allowed = true;
#endif

if (lcache_lookup_allowed)
{
```

This way we could avoid guarding the closing brackets.

What does the other reviewers think about this? @dbatyai @rerobika ?

jerry-core/ecma/operations/ecma-reference.c
```diff
@@ -132,38 +183,66 @@ ecma_op_resolve_reference_value (ecma_object_t *lex_env_p, /**< starting lexical
...
       }
+#endif /* ENABLED (JERRY_ES2015) */
 #endif /* ENABLED (JERRY_LCACHE) */
 
       ecma_value_t prop_value = ecma_op_object_find (binding_obj_p, name_p);
 
       if (ecma_is_value_found (prop_value))
       {
-        return prop_value;
+#if ENABLED (JERRY_ES2015)
```
galpeter: This is a "huge" code block to guard and only a single line is the same in both cases. I would suggest using duplicating that single line and have a #else block of the #if.

jerry-core/ecma/operations/ecma-reference.c
rerobika: I second this approach.

PR #3418, opened 2019-12-05
Implement Array.of method

szilagyiadam:
The algorithm is based on ECMA-262 v6, 22.1.2.3


jerry-core/ecma/builtin-objects/ecma-builtin-array.c
```diff
@@ -390,6 +390,79 @@ ecma_builtin_array_object_from (ecma_value_t this_arg, /**< 'this' argument */
...
+                                          ECMA_VALUE_UNDEFINED,
+                                          &len,
+                                          1);
+  }
+  else
+  {
+    ret_val = ecma_op_create_array_object (&len, 1, true);
+  }
+
+
```
dbatyai: Extra newline

zherczeg:
I think we havbe a constructor which constructs an array from a list of elements. Can't we use it?

rerobika:
@zherczeg The constructor always creates an array instance, however this method is generic.

dbatyai:
@rerobika, @zherczeg We could do something like:
```
if (!ecma_is_constructor (this_arg))
{
  return ecma_op_create_array_object (arguments_list_p, arguments_list_len, false);
}

/* Do generic construct. */
```

szilagyiadam:
I did the requested changes above.

PR #3428, opened 2019-12-06
Implement TypedArray.of method

szilagyiadam:
The algorithm is based on ECMA-262 v6, 22.2.2.2

### DIFF
--- a/tests/jerry/es2015/typedarray-of.js
+++ b/tests/jerry/es2015/typedarray-of.js
@@ -1,0 +1,131 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tests/jerry/es2015/typedarray-of.js
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+var int_typedarrays = [
+    Uint8ClampedArray,
```
rerobika: Please add tests for when you invoke %TypedArray%.of with non constructor/typedarray this value. (use Function.prototype.call for it.)

PR #3429, opened 2019-12-06
Keywords must not contain escape sequences.

zherczeg:
The ES5.1 standard is unclear about this rule.

### DIFF
--- a/jerry-core/parser/js/js-lexer.c
+++ b/jerry-core/parser/js/js-lexer.c
@@ -349,6 +349,16 @@ typedef struct
 /** @} */
 
 /**
+ * Length of the shortest keyword.
+ */
+#define LEXER_KEYWORD_MIN_LENGTH 2
+
+/**
+ * Length of the longest keyword.
+ */
+#define LEXER_KEYWORD_MAX_LENGTH 10
+
+/**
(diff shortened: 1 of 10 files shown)


akosthekiss:
Isn't the commit/PR title contradicting the commit message? unclear vs must not?

zherczeg:
This is how browsers work. In es2015, you have this sentence:

The ReservedWord definitions are specified as literal sequences of specific SourceCharacter elements. A code point in a ReservedWord cannot be expressed by a \ UnicodeEscapeSequence.

Strictly speaking, you cannot reverse this sentence. But it looks like they did. ES2019 is somewhat better:

The following tokens are ECMAScript keywords and may not be used as Identifiers in ECMAScript programs.

So keywords cannot be identifiers, regardless of escape sequences.

Es5.1 simply lists the keywords, without any comments. Since this PR passes, no es262 test covers identifiers with escape sequences.

zherczeg:
Ok I found this in ES5.1:
In string literals, regular expression literals, and identifiers, any character (code unit) may also be expressed as a Unicode escape sequence consisting of six characters, namely \u plus four hexadecimal digits.

So keywords are missing from this list, therefore they cannot contain escape sequences.

PR #3436, opened 2019-12-10
Implement and add support for RegExp.prototype[@@search]

dbatyai:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
@@ -274,37 +274,6 @@ ecma_builtin_string_prototype_object_locale_compare (ecma_string_t *this_string_
 } /* ecma_builtin_string_prototype_object_locale_compare */
 
 #if ENABLED (JERRY_BUILTIN_REGEXP)
-
-/**
- * The common preparation code for 'search' and 'match' functions
- * of the String prototype.
- *
- * @return empty value on success, error value otherwise
- *         Returned value must be freed with ecma_free_value.
- */
-static ecma_value_t
-ecma_builtin_string_prepare_search (ecma_value_t regexp_arg, /**< regex argument */
-                                    ecma_value_t *regexp_value) /**< [out] ptr to store the regexp object */
-{
-  /* 3. */
-  if (ecma_object_is_regexp_object (regexp_arg))
-  {
-    *regexp_value = ecma_copy_value (regexp_arg);
-    return ECMA_VALUE_EMPTY;
-  }
-
-  /* 4. */
-  ecma_value_t regexp_arguments[1] = { regexp_arg };
-  ecma_value_t new_regexp_value = ecma_builtin_regexp_dispatch_construct (regexp_arguments, 1);
-
-  if (!ECMA_IS_VALUE_ERROR (new_regexp_value))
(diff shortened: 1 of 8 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -600,62 +569,101 @@ ecma_builtin_string_prototype_object_replace (ecma_value_t this_value, /**< this
...
+      {
+        ecma_free_value (search_symbol);
+        return ecma_raise_type_error (ECMA_ERR_MSG ("@@search is not callable"));
+      }
+
+      ecma_object_t *search_method = ecma_get_object_from_value (search_symbol);
 
-  ecma_value_t ret_value = ECMA_VALUE_ERROR;
+      ecma_value_t arguments[] = { this_value };
+      ecma_value_t search_result = ecma_op_function_call (search_method, regexp_value, arguments, 1);
```
rerobika: How about `&this_value` instead of `arguments`?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1524,6 +1524,83 @@ ecma_regexp_read_pattern_str_helper (ecma_value_t pattern_arg) /**< the RegExp p
...
+  if (ECMA_IS_VALUE_ERROR (prev_last_index))
+  {
+    goto cleanup_string;
+  }
+
+  /* 7-8. */
+  ecma_value_t status = ecma_op_object_put (regexp_object_p, last_index_str_p, ecma_make_uint32_value (0), true);
+  if (ECMA_IS_VALUE_ERROR (status))
+  {
+    goto cleanup_string;
```
rerobika: IMO `prev_last_index`  is not feed at this point.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1524,6 +1524,83 @@ ecma_regexp_read_pattern_str_helper (ecma_value_t pattern_arg) /**< the RegExp p
...
+  }
+  ecma_free_value (status);
```
rerobika: If the result of the put operation only can be `ECMA_VALUE_{TRUE_FALSE_ERROR}` so this release is unnecessary here.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -600,62 +569,101 @@ ecma_builtin_string_prototype_object_replace (ecma_value_t this_value, /**< this
...
+  {
+    goto cleanup_string;
+  }
 
-    offset = ecma_get_number_from_value (index_value);
+  ecma_value_t new_regexp = ecma_op_create_regexp_object (pattern_p, 0);
+  ecma_deref_ecma_string (pattern_p);
+  if (ECMA_IS_VALUE_ERROR (new_regexp))
+  {
+    goto cleanup_string;
```
galpeter: Shouldn't we need to set `result` to `new_regexp` (the error)?

dbatyai: It's not necesary, `result` has been initialized to `ECMA_VALUE_ERROR` and has not been re-assigned since then.

PR #3447, opened 2019-12-12
Implement \u{hex} support.

zherczeg:
A large rework because surrogate pairs must be combined.

Currently only the 0x10C80..0x10CF2 is accepted as valid identifier character from the non-basic plane.

### DIFF
--- a/jerry-core/parser/js/js-lexer.c
+++ b/jerry-core/parser/js/js-lexer.c
@@ -53,14 +53,13 @@ align_column_to_tab (parser_line_counter_t column) /**< current column */
 /**
  * Parse hexadecimal character sequence
  *
- * @return character value
+ * @return character value or UINT32_MAX on error
  */
-ecma_char_t
-lexer_hex_to_character (parser_context_t *context_p, /**< context */
-                        const uint8_t *source_p, /**< current source position */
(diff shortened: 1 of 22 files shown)


tools/run-tests.py
```diff
@@ -284,6 +284,7 @@ def create_binary(job, options):
         subprocess.check_output(build_cmd)
         ret = 0
     except subprocess.CalledProcessError as err:
+        print(err.output)
```
zherczeg: I feel it is very bad that you cannot see any output on build fail. This line fixes this.

jerry-core/parser/js/js-lexer.c
```diff
@@ -2726,75 +2900,135 @@ lexer_scan_identifier (parser_context_t *context_p, /**< context */
...
+
+  uint8_t buf_p[64];
+
+  if (length <= 64)
+  {
+    lexer_convert_ident_to_cesu8 (buf_p, left_p->char_p, length);
+    return lexer_compare_identifier_to_chars (right_p->char_p, buf_p, length);
+  }
+
+  uint8_t *dynamic_buf_p =  parser_malloc (context_p, length);
```
rerobika: Only one space please.

PR #3449, opened 2019-12-13
Fix Math.round(x) for big integer numbers

ossy-szeged:
Math.round(x) should be x it x is already integer. (ES5 15.8.2.15)
The current implementation calculates with x +/- 0.5, which isn't
representable numbers if x >= 2^52.

A correct bug fix can be to follow the spec and simply
return the number itself if it is already integer.

### DIFF
--- a/tests/jerry/math-round.js
+++ b/tests/jerry/math-round.js
@@ -35,3 +35,17 @@ assert (Math['round'](-0.7) === -1.0);
 assert (Math['round'](-1.2) === -1.0);
 assert (Math['round'](-1.7) === -2.0);
 assert (Math['round'](-1.5) === -1.0);
+
+assert (Math['round'](1) === 1);
+assert (Math['round'](-1) === -1);
+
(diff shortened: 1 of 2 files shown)


tests/jerry/math-round.js
```diff
@@ -35,3 +35,17 @@ assert (Math['round'](-0.7) === -1.0);
 assert (Math['round'](-1.2) === -1.0);
 assert (Math['round'](-1.7) === -2.0);
 assert (Math['round'](-1.5) === -1.0);
+
+assert (Math['round'](1) === 1);
+assert (Math['round'](-1) === -1);
+
+for (var n=1; n<=53; n++)
```
LaszloLango: Please add spaces around the assignment and binary operands (`n = 1`, `n <= 53`, `x - 1`, etc.)

ossy-szeged: Done.

PR #3453, opened 2019-12-17
Implement and add support for Symbol.split

dbatyai:


dbatyai:
Benchmark | Score |
----: | ----: |
regexp.js | 49.8 -> 204 |

Binary (ES5.1) | master | patch | diff |
.text | 121396 | 121276 | -120 bytes |
.rodata | 13820 | 13816 | -4 bytes |

Binary (ES2015) | master | patch | diff |
.text | 172844 | 172980 | +136 bytes |
.rodata | 18174 | 18192 | +18 bytes |

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -1606,6 +1606,527 @@ ecma_regexp_search_helper (ecma_value_t regexp_arg, /**< regexp argument */
...
+cleanup_buffer:
+  if (string_flags & ECMA_STRING_FLAG_MUST_BE_FREED)
+  {
+    jmem_heap_free_block ((void *) string_buffer_p, string_size);
+  }
+cleanup_string:
+  ecma_deref_ecma_string (string_p);
+
+  return result;
+#endif /* ENABLED (JERRY_ES2015) */
```
zherczeg: Looks hard to maintain, isn't there a better way than maintaining two large code paths?

dbatyai: Unfortunately, ES6 uses an entirely different approach to RegExp based splitting, and the two are incompatible with eachother. ES6 requires us to reconstruct the RegExp with a sticky flag and split the string by adjusting the lastIndex property, while ES5 calls the builtin RegExp matching directly.

Maybe 10% of the code could be shared between the two implementations, but that would require a lot of guards to pull off, and in my opinion would just be messy.

PR #3460, opened 2019-12-19
Support parsing async modifiers for functions.

zherczeg:
Only parsing is implemented, so the async functions currently behave like normal function except they return with a resolved Promise object when the function is terminated correctly.


jerry-core/vm/opcodes.c
```diff
@@ -706,6 +707,78 @@ opfunc_resume_executable_object (vm_executable_object_t *executable_object_p, /*
   return result;
 } /* opfunc_resume_executable_object */
 
+/**
+ * Create a Promise object if needed and resolve it with a value
+ *
+ * @return Promise object
+ */
+ecma_value_t
+opfunc_return_promise (ecma_value_t value) /**< value */
```
rerobika: Cannot we simply call `ecma_builtin_promise_reject_or_resolve` ? I'd not prefer this code duplication even this code part is going to be changed the next steps of the implementation of the async functions.

zherczeg: That is a private function and the header which provides it is full of macros. If we move it to somewhere else we might need to move it back. What is your suggestion?

PR #3465, opened 2019-12-19
Make Math.min() and Math.max() ECMA262 conform

ossy-szeged:
Math.min() and Math.max() functions should apply ToNumber on each arguments
always, even if a NaN argument occured previously and the result will be NaN.

ECMA262 v5.1 15.8.2:
"Each of the following Math object functions applies the ToNumber abstract
operator to each of its arguments (in left-to-right order if there is more
than one) and then performs a computation on the resulting Number value(s).

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-math.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-math.c
@@ -103,6 +103,7 @@ ecma_builtin_math_object_max_min (bool is_max, /**< 'max' or 'min' operation */
                                   ecma_length_t args_number) /**< number of arguments */
 {
   ecma_number_t result_num = ecma_number_make_infinity (is_max);
+  bool nan_found = false;
 
   while (args_number > 0)
   {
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-math.c
```diff
@@ -154,6 +157,11 @@ ecma_builtin_math_object_max_min (bool is_max, /**< 'max' or 'min' operation */
     args_number--;
   }
 
+  if (JERRY_UNLIKELY (nan_found))
+  {
+    result_num = ecma_number_make_nan ();;
```
rerobika: Please remove the extra semicolon.

jerry-core/ecma/builtin-objects/ecma-builtin-math.c
```diff
@@ -126,10 +127,12 @@ ecma_builtin_math_object_max_min (bool is_max, /**< 'max' or 'min' operation */
       ecma_fast_free_value (value);
     }
 
-    if (JERRY_UNLIKELY (ecma_number_is_nan (arg_num)))
+    if (JERRY_UNLIKELY (nan_found || ecma_number_is_nan (arg_num)))
     {
-      result_num = arg_num;
-      break;
+      nan_found = true;
+      arg++;
```
rerobika: As i see the 
```
arg++;
args_number--;
```
can be performed before the `isnan` and `iszero` conditions so I'd move them up, therefore no duplication is required.

jerry-core/ecma/builtin-objects/ecma-builtin-math.c
ossy-szeged: Done.

jerry-core/ecma/builtin-objects/ecma-builtin-math.c
ossy-szeged: Good point, fixed.

PR #3469, opened 2019-12-19
Add support for new.target

galpeter:
Notable changes:
* Extracted the pure JS/builtin and external C method invocations
  into two new methods (`ecma_op_function_call_{simple, external}`).
* Updated parser/scanner to handle "new.target" correctly.
* Added JS test case.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -725,6 +725,162 @@ ecma_op_set_class_prototype (ecma_value_t completion_value, /**< completion_valu
 #endif /* ENABLED (JERRY_ES2015) */
 
 /**
+ * Perform a JavaScript function object method call.
+ *
+ * The input function object should be a pure JavaScript method or
+ * a builtin function implemented in C.
+ *
+ * In case of JERRY_ES2015 the arguments_list_p contains the information
+ * wheter if the function was invoked with "new" or not.
+ *
+ * @return the result of the function call.
+ */
+static ecma_value_t
+ecma_op_function_call_simple (ecma_object_t *func_obj_p, /**< Function object */
+                              ecma_value_t this_arg_value, /**< 'this' argument's value */
+                              const ecma_value_t *arguments_list_p, /**< arguments list */
+                              ecma_length_t arguments_list_len) /**< length of arguments list */
+{
+  JERRY_ASSERT (ecma_get_object_type (func_obj_p) == ECMA_OBJECT_TYPE_FUNCTION);
+
+  if (JERRY_UNLIKELY (ecma_get_object_is_builtin (func_obj_p)))
+  {
+    JERRY_ASSERT (!ecma_op_function_has_construct_flag (arguments_list_p));
+    ecma_value_t ret_value = ecma_builtin_dispatch_call (func_obj_p,
+                                                         this_arg_value,
+                                                         arguments_list_p,
+                                                         arguments_list_len);
+
+    return ret_value;
+  }
+
+  /* Entering Function Code (ECMA-262 v5, 10.4.3) */
+  ecma_extended_object_t *ext_func_p = (ecma_extended_object_t *) func_obj_p;
+
+  ecma_object_t *scope_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_object_t,
+                                                            ext_func_p->u.function.scope_cp);
+
+  /* 8. */
+  ecma_value_t this_binding = this_arg_value;
+  bool free_this_binding = false;
+
+  const ecma_compiled_code_t *bytecode_data_p = ecma_op_function_get_compiled_code (ext_func_p);
+  uint16_t status_flags = bytecode_data_p->status_flags;
+
+#if ENABLED (JERRY_ES2015)
+  if (JERRY_UNLIKELY (status_flags & (CBC_CODE_FLAGS_CONSTRUCTOR | CBC_CODE_FLAGS_GENERATOR)))
+  {
+    bool is_construct_call = ecma_op_function_has_construct_flag (arguments_list_p);
+
+    if (!is_construct_call && (status_flags & CBC_CODE_FLAGS_CONSTRUCTOR))
+    {
+      return ecma_raise_type_error (ECMA_ERR_MSG ("Class constructor cannot be invoked without 'new'."));
+    }
+    if (is_construct_call && (status_flags & CBC_CODE_FLAGS_GENERATOR))
+    {
+      return ecma_raise_type_error (ECMA_ERR_MSG ("Generator functions cannot be invoked with 'new'."));
+    }
+  }
+#endif /* ENABLED (JERRY_ES2015) */
+
+  /* 1. */
+  if (!(status_flags & CBC_CODE_FLAGS_STRICT_MODE))
+  {
+    if (ecma_is_value_undefined (this_binding)
+        || ecma_is_value_null (this_binding))
+    {
+      /* 2. */
+      this_binding = ecma_make_object_value (ecma_builtin_get_global ());
+    }
+    else if (!ecma_is_value_object (this_binding))
+    {
+      /* 3., 4. */
+      this_binding = ecma_op_to_object (this_binding);
+      free_this_binding = true;
+
+      JERRY_ASSERT (!ECMA_IS_VALUE_ERROR (this_binding));
+    }
+  }
+
+  arguments_list_p = ecma_op_function_clear_construct_flag (arguments_list_p);
+
+  /* 5. */
+  ecma_object_t *local_env_p;
+  if (status_flags & CBC_CODE_FLAGS_LEXICAL_ENV_NOT_NEEDED)
+  {
+    local_env_p = scope_p;
+  }
+  else
+  {
+    local_env_p = ecma_create_decl_lex_env (scope_p);
+    if (bytecode_data_p->status_flags & CBC_CODE_FLAGS_IS_ARGUMENTS_NEEDED)
+    {
+      ecma_op_create_arguments_object (func_obj_p,
+                                       local_env_p,
+                                       arguments_list_p,
+                                       arguments_list_len,
+                                       bytecode_data_p);
+    }
+#if ENABLED (JERRY_ES2015)
+    if (JERRY_UNLIKELY (status_flags & CBC_CODE_FLAGS_CONSTRUCTOR))
+    {
+      ecma_op_set_class_this_binding (local_env_p, this_binding);
+    }
+#endif /* ENABLED (JERRY_ES2015) */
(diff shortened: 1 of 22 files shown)


jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1781,6 +1781,42 @@ scanner_is_global_context_needed (parser_context_t *context_p) /**< context */
...
+ */
+bool
+scanner_try_scan_new_target (parser_context_t *context_p) /**< parser/scanner context */
+{
+  lexer_token_t original_token = context_p->token;
+  JERRY_ASSERT (original_token.type == LEXER_KEYW_NEW);
+
+  scanner_location_t location;
+  scanner_get_location (&location, context_p);
+  lexer_next_token (context_p);
```
zherczeg: This is too complex. There is a function to test the next character.

galpeter: I need to test the next character and next literal. Is there a method for it?

rerobika: How about this?

```
if (lexer_check_next_character (context_p, LIT_CHAR_DOT))
      {
        context_p->source_p++;
        context_p->token.flags &= (uint8_t) ~LEXER_NO_SKIP_SPACES;
        lexer_skip_spaces (context_p);
        /* 'new.target' must not contain escaped characters */
        if (context_p->source_end_p - context_p->source_p >= 6
            && memcmp (context_p->source_p, "target", 6) == 0)
        {
          // emit new.target opcode
        }
        /* new.<anything> is an invalid primary expression> */
        else
        {
          parser_raise_error (context_p, PARSER_ERR_PRIMARY_EXP_EXPECTED);
        }
      }
```

zherczeg: Use consumeCharacter because the column needs to be increased as well. As far as I know, there cannot be anything else after a `new.` except `target`, so you can use `lexer_next_token` after the dot and throw an error on fail. The prposed solution is also wrong because `new.targetX` is also accepted as well.

galpeter:
I'll do another update with corrected error messages (currently only one is used).

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -1326,6 +1326,21 @@ parser_abort_parsing_after_arrow (parser_context_t *context_p)
...
 } /* parser_abort_parsing_after_arrow */
 
+/**
+ * Check and throw an error if the assignment target is the "new.target".
+ */
+static void
+parser_check_new_target_assignment (parser_context_t *context_p) /**< parser context */
+{
+  JERRY_ASSERT (context_p->token.type == LEXER_ASSIGN
+    || LEXER_IS_BINARY_LVALUE_TOKEN (context_p->token.type));
```
rerobika: misaligned line.

tests/jerry/es2015/new-target.js
```diff
@@ -0,0 +1,111 @@
...
+    assert (ex instanceof SyntaxError);
+  }
+}
+
+expect_syntax_error ("function assign () { new.target = 3; }");
+expect_syntax_error ("function assign () { new.target += 3; }");
+expect_syntax_error ("function assign () { new.target *= 3; }");
+expect_syntax_error ("function assign () { new.target -= 3; }");
+expect_syntax_error ("function assign () { new.target |= 3; }");
+expect_syntax_error ("function assign () { new.target &= 3; }");
```
rerobika: How about `function assign () { new.target++ };`. This should be a syntax error, but currently it is a ReferenceError. (Hint: `CBC_EXT_THROW_REFERENCE_ERROR` is emitted.)

jerry-core/vm/vm.c
```diff
@@ -2137,6 +2137,20 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+        case VM_OC_PUSH_NEW_TARGET:
+        {
+          ecma_object_t *new_target_object = JERRY_CONTEXT (current_new_target);
+          if (new_target_object == NULL)
+          {
+            *stack_top_p++ = ECMA_VALUE_UNDEFINED;
+          }
+          else
+          {
+            ecma_ref_object (new_target_object);
```
rerobika: I'd assert that `JERRY_CONTEXT (current_new_target) != JERRY_CONTEXT_INVALID_NEW_TARGET `.

tests/jerry/es2015/new-target.js
galpeter: Ah... yes, then this will need a bit of work. If I see correctly the post/prefix operations should also be a syntax error, but unary operations and other binary operations does not trigger syntax errors.

jerry-core/vm/vm.c
galpeter: Yeah, good idea. Will add it.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -113,6 +113,30 @@ parser_check_invalid_assign (parser_context_t *context_p) /**< context */
   }
 } /* parser_check_invalid_assign */
 
+#if ENABLED (JERRY_ES2015)
+/**
+ * Check and throw an error if the "new.target" is invalid as a left-hand side expression.
+ */
+static void
+parser_check_new_target_lhs (parser_context_t *context_p,
```
rerobika: missing argument description,

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -113,6 +113,30 @@ parser_check_invalid_assign (parser_context_t *context_p) /**< context */
...
+                             cbc_opcode_t opcode) /**< parser context */
+{
+  JERRY_ASSERT ((opcode == CBC_ASSIGN
+                && (context_p->token.type == LEXER_ASSIGN
+                    || LEXER_IS_BINARY_LVALUE_TOKEN (context_p->token.type)))
+                || opcode == CBC_PRE_INCR
```
rerobika: How about a rage check for the opcode? `opcode >= PRE_INC & opcode <= POST_DECR`?

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -724,6 +724,163 @@ ecma_op_set_class_prototype (ecma_value_t completion_value, /**< completion_valu
...
+    }
+#if ENABLED (JERRY_ES2015)
+    if (JERRY_UNLIKELY (status_flags & CBC_CODE_FLAGS_CONSTRUCTOR))
+    {
+      ecma_op_set_class_this_binding (local_env_p, this_binding);
+    }
+#endif /* ENABLED (JERRY_ES2015) */
+  }
+
+
```
rerobika: One new line is enough.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1156,6 +1225,15 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
...
   }
 
+#if ENABLED (JERRY_ES2015)
+  ecma_object_t *old_new_target = JERRY_CONTEXT (current_new_target);
+  if (JERRY_LIKELY (new_this_obj_p != NULL))
+  {
+    /* This is not a super call so update the "new.target" */
+    JERRY_CONTEXT (current_new_target) = func_obj_p;
+  }
+#endif
```
rerobika: missing comment

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -3044,7 +3044,14 @@ parser_parse_statements (parser_context_t *context_p) /**< context */
           options |= PARSE_EXPR_HAS_LITERAL;
         }
 
-        if (context_p->status_flags & PARSER_IS_FUNCTION)
+#if ENABLED (JERRY_ES2015)
+        bool is_eval = (context_p->status_flags & PARSER_IS_EVAL) != 0;
+#else
```
rerobika: Ditto.

jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1752,6 +1752,37 @@ scanner_is_global_context_needed (parser_context_t *context_p) /**< context */
...
+ *          false if there is no "." after the new.
+ */
+bool
+scanner_try_scan_new_target (parser_context_t *context_p) /**< parser/scanner context */
+{
+  JERRY_ASSERT (context_p->token.type == LEXER_KEYW_NEW);
+
+  if (lexer_check_next_character (context_p, LIT_CHAR_DOT))
+  {
+    lexer_consume_next_character (context_p);
```
zherczeg: I think this might affect the column info in the error. So I would call `lexer_next_token` here and check that the dot is a single dot character. This way we could get a correct column when three dots are used. I think this is a rare case, so the overall perf overhead is small.

jerry-core/parser/js/js-parser-expr.c
galpeter: Fixed

jerry-core/parser/js/js-scanner-util.c
galpeter: Thanks, updated!

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -113,6 +113,27 @@ parser_check_invalid_assign (parser_context_t *context_p) /**< context */
...
+ * Check and throw an error if the "new.target" is invalid as a left-hand side expression.
+ */
+static void
+parser_check_new_target_lhs (parser_context_t *context_p, /**< parser context */
+                             cbc_opcode_t opcode) /**< current opcode under parsing */
+{
+  JERRY_ASSERT ((opcode == CBC_ASSIGN
+                && (context_p->token.type == LEXER_ASSIGN
+                    || LEXER_IS_BINARY_LVALUE_TOKEN (context_p->token.type)))
+                || (opcode >= CBC_PRE_INCR && opcode <= CBC_POST_DECR));
```
zherczeg: I think there is an indentation error here. I would put the last check to the first, and that would look better.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -113,6 +113,27 @@ parser_check_invalid_assign (parser_context_t *context_p) /**< context */
   }
 } /* parser_check_invalid_assign */
 
+#if ENABLED (JERRY_ES2015)
+/**
+ * Check and throw an error if the "new.target" is invalid as a left-hand side expression.
+ */
+static void
+parser_check_new_target_lhs (parser_context_t *context_p, /**< parser context */
```
zherczeg: This name should be in line with `parser_check_invalid_assign` somehow. You can rename any or both of them of course, but they have a similar function so they should have a similar name.

galpeter:
@zherczeg thanks I'v updated the PR.

PR #3472, opened 2019-12-20
Support all literals in 'jerry_get_literals_from_snapshot' when C format is used

LaszloLango:
The literals must be saved in hex format to support cases with special
characters, like `var s = 'hello",\n"world';`

JerryScript-DCO-1.0-Signed-off-by: László Langó lango@inf.u-szeged.hu

### DIFF
--- a/tests/unit-core/test-snapshot.c
+++ b/tests/unit-core/test-snapshot.c
@@ -400,7 +400,7 @@ main (void)
 
     static jerry_char_t literal_buffer_c[LITERAL_BUFFER_SIZE];
     static uint32_t literal_snapshot_buffer[SNAPSHOT_BUFFER_SIZE];
-    static const jerry_char_t code_for_c_format[] = "var object = { aa:'fo o', Bb:'max', aaa:'xzy0' };";
+    static const jerry_char_t code_for_c_format[] = "var object = { aa:'fo\" o\\n \\\\', Bb:'max', aaa:'xzy0' };";
 
     jerry_value_t generate_result;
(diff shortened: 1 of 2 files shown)


tests/unit-core/test-snapshot.c
```diff
@@ -423,22 +423,24 @@ main (void)
...
                                             "{\n"
-                                            "  \"Bb\",\n"
-                                            "  \"aa\",\n"
-                                            "  \"aaa\",\n"
-                                            "  \"xzy0\"\n"
+                                            "  \"\\x42\\x62\",\n"
+                                            "  \"\\x61\\x61\",\n"
+                                            "  \"\\x61\\x61\\x61\",\n"
+                                            "  \"\\x66\\x6F\\x20\\x6F\",\n"
+                                            "  \"\\x78\\x7A\\x79\\x30\"\n"
```
zherczeg: Please don't convert the ascii characters. Always use two hex chars after \x, so `a..f` letters will not be accidentally consumed by the compiler.

LaszloLango: I cannot follow you, could you explain this with a bit more detail? The double quote character is an ascii character, but it will cause problems in C format, like in my example in the commit message. It must be converted to hex or escaped. I thought this is an easier solution than escaping only some special characters.

zherczeg: The patch changed `"  \"Bb\",\n"` -> `\"\\x42\\x62\",\n`.

PR #3480, opened 2020-01-02
Add fast-path check to ecma_builtin_array_prototype_slice

pmarkee:
Fixes #3459

JerryScript-DCO-1.0-Signed-off-by: Peter Marki marpeter@inf.u-szeged.hu

PR #3481, opened 2020-01-02
Add length check to ecma_builtin_array_prototype_object_slice

### LINKED ISSUE #3479: Assertion old_length < new_length in ecma_fast_array_extend
###### JerryScript revision
518fcf2c

###### Build platform
Linux-4.15.0-72-generic-x86_64-with-Ubuntu-18.04-bionic

###### Build steps
```
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```

###### Test case
```javascript
class MyObservedArray extends Array {
    constructor() {
        super('"use strict"; var x = "\\411";')
    } [Symbol]() {}
}

new MyObservedArray().slice()
```

###### Output
```text
ICE: Assertion 'old_length < new_length' failed at jerryscript/jerry-core/ecma/operations/ecma-array-object.c(ecma_fast_array_extend):317.
```

###### Backtrace
```text
bt
#0  0xf7fd5079 in __kernel_vsyscall ()
#1  0xf77fc832 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf77fdcc1 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0x5657bd39 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-port/default/default-fatal.c:30
#4  0x5664e317 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x5664e358 in jerry_assert_fail (assertion=0x566f1a00 "old_length < new_length", file=0x566f16e0 "jerryscript/jerry-core/ecma/operations/ecma-array-object.c", function=0x566b8d20 <__func__.5035.lto_priv.702> "ecma_fast_array_extend", line=317) at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x5665cc17 in ecma_fast_array_extend (object_p=0xf5f03370, new_length=1) at jerryscript/jerry-core/ecma/operations/ecma-array-object.c:317
#7  0x565f2ed4 in ecma_builtin_array_prototype_object_slice (arg1=72, arg2=72, obj_p=0xf5f03460, len=1) at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c:857
#8  0x565f8616 in ecma_builtin_array_prototype_dispatch_routine (builtin_routine_id=86, this_arg=4126159971, arguments_list_p=0xffffc420, arguments_number=0) at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c:2625
#9  0x56698da7 in ecma_builtin_dispatch_routine (builtin_object_id=ECMA_BUILTIN_ID_ARRAY_PROTOTYPE, builtin_routine_id=86, this_arg_value=4126159971, arguments_list_p=0xffffc420, arguments_list_len=0) at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.c:1020
#10 0x56699008 in ecma_builtin_dispatch_call (obj_p=0xf5f01120, this_arg_value=4126159971, arguments_list_p=0xffffc6ec, arguments_list_len=0) at jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.c:1045
#11 0x5666b55d in ecma_op_function_call (func_obj_p=0xf5f01120, this_arg_value=4126159971, arguments_list_p=0xffffc6ec, arguments_list_len=0) at jerryscript/jerry-core/ecma/operations/ecma-function-object.c:762
#12 0x56631eba in opfunc_call.lto_priv.465 (frame_ctx_p=0xffffc690) at jerryscript/jerry-core/vm/vm.c:699
#13 0x565e5638 in vm_execute (frame_ctx_p=0xffffc690) at jerryscript/jerry-core/vm/vm.c:4117
#14 0x565e5c63 in vm_run (bytecode_header_p=0xf4d03780, this_binding_value=4126149459, lex_env_p=0xf5d007b0, arg_list_p=0x0, arg_list_len=0) at jerryscript/jerry-core/vm/vm.c:4240
#15 0x566307df in vm_run_global (bytecode_p=0xf4d03780) at jerryscript/jerry-core/vm/vm.c:286
#16 0x566a097b in jerry_run (func_val=4126152483) at jerryscript/jerry-core/api/jerry.c:595
#17 0x5669d0a8 in main (argc=2, argv=0xffffcb04) at jerryscript/jerry-main/main-unix.c:740
```

<sup>Found by [Fuzzinator](http://fuzzinator.readthedocs.io/) with [grammarinator](https://github.com/renatahodovan/grammarinator). </sup>

pmarkee:
Fixes #3479

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -828,8 +828,9 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
   JERRY_ASSERT (start <= len && end <= len);
 
   bool use_fast_path = ecma_op_object_is_fast_array (obj_p);
+  uint32_t copied_length = (end > start) ? end - start : 0;
 #if ENABLED (JERRY_ES2015)
-  ecma_value_t new_array = ecma_op_array_species_create (obj_p, 0);
+  ecma_value_t new_array = ecma_op_array_species_create (obj_p, copied_length);
 
   if (ECMA_IS_VALUE_ERROR (new_array))
   {
@@ -846,25 +847,45 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
   /* 9. */
   uint32_t n = 0;
 
-  if (use_fast_path)
+  if (use_fast_path && copied_length > 0)
   {
     ecma_extended_object_t *ext_from_obj_p = (ecma_extended_object_t *) obj_p;
 
-    if (ext_from_obj_p->u.array.u.hole_count < ECMA_FAST_ARRAY_HOLE_ONE
-        && len != 0
-        && start < end)
+    if (ext_from_obj_p->u.array.u.hole_count < ECMA_FAST_ARRAY_HOLE_ONE)
     {
-      uint32_t length = end - start;
       ecma_extended_object_t *ext_to_obj_p = (ecma_extended_object_t *) new_array_p;
-      ecma_value_t *to_buffer_p = ecma_fast_array_extend (new_array_p, length);
+
+#if ENABLED (JERRY_ES2015)
+      uint32_t target_length = ext_to_obj_p->u.array.length;
+      ecma_value_t *to_buffer_p;
+      if (copied_length == target_length)
+      {
+        to_buffer_p = ECMA_GET_NON_NULL_POINTER (ecma_value_t, new_array_p->u1.property_list_cp);
+      }
+      else if (copied_length > target_length)
+      {
+        to_buffer_p = ecma_fast_array_extend (new_array_p, copied_length);
+      }
+      else
+      {
+        ecma_delete_fast_array_properties (new_array_p, copied_length);
+        to_buffer_p = ECMA_GET_NON_NULL_POINTER (ecma_value_t, new_array_p->u1.property_list_cp);
+      }
+#else /* !ENABLED (JERRY_ES2015) */
+      ecma_value_t *to_buffer_p = ecma_fast_array_extend (new_array_p, copied_length);
+#endif /* ENABLED (JERRY_ES2015) */
+
       ecma_value_t *from_buffer_p = ECMA_GET_NON_NULL_POINTER (ecma_value_t, obj_p->u1.property_list_cp);
 
       for (uint32_t k = start; k < end; k++, n++)
       {
+#if ENABLED (JERRY_ES2015)
+        ecma_free_value_if_not_object (to_buffer_p[n]);
+#endif /* ENABLED (JERRY_ES2015) */
         to_buffer_p[n] = ecma_copy_value_if_not_object (from_buffer_p[k]);
       }
 
-      ext_to_obj_p->u.array.u.hole_count -= length * ECMA_FAST_ARRAY_HOLE_ONE;
+      ext_to_obj_p->u.array.u.hole_count &= ECMA_FAST_ARRAY_HOLE_ONE - 1;
 
       return new_array;
     }
--- a/tests/jerry/es2015/array-prototype-slice.js
+++ b/tests/jerry/es2015/array-prototype-slice.js
@@ -1,0 +1,55 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Constructor creates longer array than expected.
+class LongArray extends Array {
+    constructor(len) {
+        super (42);
+        this.fill ("foo");
+    }
+}
+
+var a = new LongArray (5);
+a.length = 5;
+var sliced = a.slice ();
+assert (sliced.length == 5);
+assert (JSON.stringify (sliced) == '["foo","foo","foo","foo","foo"]')
+
+// Constructor creates shorter array than expected.
+class ShortArray extends Array {
+    constructor(len) {
+        super (2);
+        this.fill ("bar");
+    }
+}
+
+var b = new ShortArray (8);
+b.length = 8;
+b.fill ("asd", 2);
+var sliced2 = b.slice ();
+assert (sliced2.length == 8);
+assert (JSON.stringify (sliced2) == '["bar","bar","asd","asd","asd","asd","asd","asd"]');
+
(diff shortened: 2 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -850,7 +850,8 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
 
     if (ext_from_obj_p->u.array.u.hole_count < ECMA_FAST_ARRAY_HOLE_ONE
         && len != 0
-        && start < end)
+        && start < end
+        && end - start > len)
```
dbatyai: Correct me if I'm wrong, but both `start` and `end` are in the [0, len] range, and we we know that `start < end` from the condition, which means `end - start` will allways be `<= len`.

So essentially this just disables the fast path entirely.

LaszloLango: Good point, this might be a wrong fix for the issue.

pmarkee: Right, I missed that. I will come up with another solution.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -854,11 +854,23 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
     {
       uint32_t length = end - start;
       ecma_extended_object_t *ext_to_obj_p = (ecma_extended_object_t *) new_array_p;
-      ecma_value_t *to_buffer_p = ecma_fast_array_extend (new_array_p, length);
+
+      ecma_value_t *to_buffer_p;
+      if (length == len)
+      {
+        to_buffer_p = ECMA_GET_NON_NULL_POINTER (ecma_value_t, new_array_p->u1.property_list_cp);
```
pmarkee: @dbatyai @LaszloLango I have updated the PR, but on this line we could have a nullpointer which will makes a few tests fail. This patch will have to be rebased once #3480 lands, and then it should work.

pmarkee: Yes, meanwhile I realized that my solution is not correct.
However it is not true that the target array can contain any number of elements. According to [standard](https://www.ecma-international.org/ecma-262/6.0/#sec-array.prototype.slice):
```
11. Let  count be max(final – k, 0).
12. Let A be ArraySpeciesCreate(O, count).
```
This way the target array is already created with the desired length, so if we follow the standard than there shouldn't be a need for either extending or shrinking in any case (with ES2015 profile). I think we should do it like this:
```c
#if ENABLED (JERRY_ES2015)
      ecma_value_t *to_buffer_p = ECMA_GET_NON_NULL_POINTER (ecma_value_t, new_array_p->u1.property_list_cp);
      ecma_value_t *to_buffer_p = ecma_fast_array_extend (new_array_p, length);
```
And of course the creation of `new_array` needs to be changed to follow the standard, because currently we create it with 0 length.
What do you think about this? I tried this locally and all tests passed with `tools/run-tests.py --jerry-tests`.

dbatyai: You are correct, the constructor is called with incorrect arguments. However, even if the construct argument is fixed, the custom constructor can just ignore it, and create an arbitrary array. Try this:

```
class CustomArray extends Array {
    constructor(len) {
        super(42);
        this.fill ("str");
    }
}

var a = new CustomArray(5);  // Array length will be 42, regardless of the argument
a.length = 5;  // Crop the source array;
a.slice();  // Number of copied elements will be 5, but ArraySpeciesCreate will create an array with length 42
```

This is what I meant when I asked for more tests. We should add tests for when the constructor creates smaller/same sized/larger arrays than what is expected. I would also prefer a test for when the constructor creates a non-fast array, for example by adding a non array index property in the constructor.

pmarkee: Good point, I did not consider the custom constructor.

pmarkee:
Thanks for the review @dbatyai , I have updated the PR.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -829,7 +829,7 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
 
   bool use_fast_path = ecma_op_object_is_fast_array (obj_p);
 #if ENABLED (JERRY_ES2015)
-  ecma_value_t new_array = ecma_op_array_species_create (obj_p, 0);
+  ecma_value_t new_array = ecma_op_array_species_create (obj_p, (uint32_t) JERRY_MAX ((int32_t) (end - start), 0));
```
dbatyai: Could we factor out the new length into a const variable, and use that here, and later on instead of `length`? We should probably give it a more descriptive name as well, maybe somethng like `copied_length`.

After that, we could do an early return if `copied_length == 0` right after the result array is constructed. I believe this would allow us to omit `len != 0 && start < end` from the condition on line 853. (`len == 0` or `start >= end` would result in a copied length of zero, which would be handled by the early return.)

pmarkee: Done.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -843,28 +844,63 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
 
   ecma_object_t *new_array_p = ecma_get_object_from_value (new_array);
 
+  if (copied_length == 0)
+  {
+#if ENABLED (JERRY_ES2015)
+    ecma_value_t set_length_value = ecma_builtin_array_prototype_helper_set_length (new_array_p, ((ecma_number_t) 0));
```
dbatyai: Hmm, I didn't realize we'd have to set the length, my bad. I'd rather not duplicate this, could you please remove the early return and do `if (use_fast_path && copied_length > 0)`? That should have the same result.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -828,8 +828,9 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
   JERRY_ASSERT (start <= len && end <= len);
 
   bool use_fast_path = ecma_op_object_is_fast_array (obj_p);
+  uint32_t copied_length = (uint32_t) JERRY_MAX ((int32_t) (end - start), 0);
```
dbatyai: This can result in an incorrect value in some cases. `end = 4294967295` and `start = 0` will be considered negative, and will result in zero. We should do `(end > start) ? end - start : 0` instead.

PR #3489, opened 2020-01-06
Throw error for generator function class constructor

rerobika:

### DIFF
--- a/jerry-core/parser/js/js-parser-expr.c
+++ b/jerry-core/parser/js/js-parser-expr.c
@@ -569,11 +569,20 @@ parser_parse_class_literal (parser_context_t *context_p) /**< context */
     {
       is_computed = true;
     }
-    else if ((status_flags & PARSER_CLASS_STATIC_FUNCTION)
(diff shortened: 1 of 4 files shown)


zherczeg:
I don't get why `class C { *["constructor"]() { } }` is supported

rerobika:
> I don't get why `class C { *["constructor"]() { } }` is supported

To create a generator object from a generator function you should **call** the function, but you cannot invoke a class constructor without **new** keyword. However, I can agree that's a really weird example but due the this restriction no any runtime validation is required, since the generator function is inaccessible.

zherczeg:
This works in firefox:
```
class C { ["constructor"]() { return 5; } }
(new C).constructor(); // 5
```

The code seems ok, but I would add two new testcases (with and without generator) for these.

rerobika:
@zherczeg I've added these testcases.

rerobika:
Note: The PR and the CI fail have nothing in common. The CI fail has been resolved in #3493.

PR #3490, opened 2020-01-07
Introduce builtin intrinsic object to share property values between builtin objects

rerobika:


jerry-core/ecma/operations/ecma-objects-arguments.c
```diff
@@ -140,6 +140,22 @@ ecma_op_create_arguments_object (ecma_object_t *func_obj_p, /**< callee function
...
+  /* ECMAScript v6, 9.4.4.6.7, 9.4.4.7.22 */
+  ecma_string_t *symbol_p = ecma_op_get_global_symbol (LIT_MAGIC_STRING_ITERATOR);
+
+  prop_value_p = ecma_create_named_data_property (obj_p,
+                                                  symbol_p,
+                                                  ECMA_PROPERTY_CONFIGURABLE_WRITABLE,
+                                                  NULL);
+
+  prop_value_p->value = ecma_op_object_get_by_magic_id (ecma_builtin_get (ECMA_BUILTIN_ID_ARRAY_PROTOTYPE),
+                                                        LIT_MAGIC_STRING_VALUES);
```
zherczeg: Is this always successfull?

rerobika: Ohh, good catch. According to the standard the arguments object creation operation never throws exception, however, this [[Get]] can, since the Array.prototype.values is function object is configurable and writable. However the standard specifies, that `Perform DefinePropertyOrThrow(obj, @@iterator, PropertyDescriptor {[[Value]]:%ArrayProto_values%, [[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true}).`. Therefore the `%ArrayProto_values%` function object need to be saved after the lazy instantiation or it should be a new builtin routine.

rerobika:
@zherczeg I've updated the PR, please take an other look at it.

jerry-core/lit/lit-magic-strings.h
```diff
@@ -32,6 +32,7 @@ typedef enum
 #undef LIT_MAGIC_STRING_FIRST_STRING_WITH_SIZE
 /** @endcond */
   LIT_NON_INTERNAL_MAGIC_STRING__COUNT, /**< number of non-internal magic strings */
+  LIT_INTERNAL_MAGIC_STRING_ARRAY_PROTOTYPE_VALUES, /**< %ArrayProto_values% intrinsic routine */
```
galpeter: Is it ok to put it here? The next internal magic string will have the previous value.

rerobika: ohh, good catch

PR #3496, opened 2020-01-07
Support iterable objects in Promise.all and Promise.race functions

DanielBallaSZTE:

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -2673,6 +2673,65 @@ ecma_op_species_constructor (ecma_object_t *this_value, /**< This Value */
 #endif /* ENABLED (JERRY_ES2015) */
 
 /**
+ * 7.3.18 Abstract operation Invoke when property name is a magic string
+ *
+ * @return ecma_value result of the invoked function or raised error
+ *         note: returned value must be freed with ecma_free_value
+ */
+inline ecma_value_t JERRY_ATTR_ALWAYS_INLINE
+ecma_op_invoke_by_magic_id (ecma_value_t object, /**< Object value */
+                            lit_magic_string_id_t magic_string_id, /**< Magic string ID */
+                            ecma_value_t *args_p, /**< Argument list */
+                            ecma_length_t args_len) /**< Argument list length */
+{
+  return ecma_op_invoke (object, ecma_get_magic_string (magic_string_id), args_p, args_len);
+} /* ecma_op_invoke_by_magic_id */
+
+/**
+ * 7.3.18 Abstract operation Invoke
+ *
+ * @return ecma_value result of the invoked function or raised error
+ *         note: returned value must be freed with ecma_free_value
+ */
+ecma_value_t
+ecma_op_invoke (ecma_value_t object, /**< Object value */
+                ecma_string_t *property_name_p, /**< Property name */
+                ecma_value_t *args_p, /**< Argument list */
+                ecma_length_t args_len) /**< Argument list length */
+{
+  /* 3. */
+  ecma_value_t object_value = ecma_op_to_object (object);
+  if (ECMA_IS_VALUE_ERROR (object_value))
+  {
+    return object_value;
+  }
+
+  ecma_object_t *object_p = ecma_get_object_from_value (object_value);
+  ecma_value_t func = ecma_op_object_get (object_p, property_name_p);
+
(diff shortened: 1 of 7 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2672,6 +2672,65 @@ ecma_op_species_constructor (ecma_object_t *this_value, /**< This Value */
 
 #endif /* ENABLED (JERRY_ES2015) */
 
+/**
+ * 7.3.18 Abstract operation Invoke when property name is a magic string
+ *
+ * @return ecma_value result of the invoked function or raised error
+ *         note: returned value must be freed with ecma_free_value
+ */
+extern inline ecma_value_t JERRY_ATTR_ALWAYS_INLINE
```
galpeter: Do we need the `extern` here?

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -409,6 +409,35 @@ ecma_call_builtin_executor (ecma_object_t *executor_p, /**< the executor object
...
 
+/**
+ * Helper function for PromiseCreateResovingFucntions.
+ *
+ * See also: ES2015 25.4.1.3 2. - 7.
+ *
+ * @return pointer to the resolving function
+ */
+static ecma_object_t *
+ecma_promise_create_resolving_functions_helper (ecma_object_t *obj_p,
```
galpeter: Should we have comments for the arguments here?

tests/jerry/es2015/regression-test-issue-2544.js
```diff
@@ -13,4 +13,4 @@
 // limitations under the License.
 
 Object.defineProperty(Array.prototype, 0, { get : function () { throw $; } });
-Promise.race([ , this]).then(Error);
+Promise.race([ , this]).then(Error).catch(function(err) { assert(err instanceof ReferenceError); });
```
galpeter: Would it possible to test if the Promise executed at all? Otherwise it can happen that the test reports OK but the Promise was not invoked at all.

tests/jerry/es2015/promise-all-iterator.js
```diff
@@ -0,0 +1,64 @@
...
+  iterable['return'] = methods['return'];
+  iterable['throw'] = methods['throw'];
+
+  return iterable;
+};
+
+// iterator next
+Promise.all({[Symbol.iterator]() { return {get next() { throw 5; }}}
+}).catch(err => {
+  assert(err === 5);
```
galpeter: It would be good to have also test that the `catch ` was correctly called in all cases.

jerry-core/ecma/operations/ecma-objects.c
rerobika: IMO that corresponds to some kind of Windows port issue. cc @ossy-szeged.

ossy-szeged: Extern is needed only if the inline function is called outside of jerry-core, typically from unit-core.
It seems it isn't that case, it should be safe to remove the extern keyword.

tests/jerry/es2015/regression-test-issue-2544.js
DanielBallaSZTE: There's a solution, I'll push it, feedback would be appreciated on it.

tests/jerry/es2015/promise-all-iterator.js
DanielBallaSZTE: ^same as above.

tests/jerry/es2015/regression-test-issue-2544.js
```diff
@@ -13,4 +13,8 @@
 // limitations under the License.
 
 Object.defineProperty(Array.prototype, 0, { get : function () { throw $; } });
-Promise.race([ , this]).then(Error);
+var asyncPassed = false;
+Promise.race([ , this]).then(Error).catch(function(err) { asyncPassed = (err instanceof ReferenceError); });
+new Promise(function() {
+  throw 5;
+}).then(Error).catch(function() { assert(asyncPassed); });
```
galpeter: Is the Promise execution/resolve order guaranteed in this test case? Could it happen that the race finishes after the Promise which is after it?

DanielBallaSZTE: It is guaranteed to happen in this order.

PR #3499, opened 2020-01-08
Add builtin GeneratorFunction support 

rerobika:


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -143,21 +144,168 @@ ecma_is_constructor (ecma_value_t value) /**< ecma value */
...
+  }
+
+  ecma_string_t *str_p = ecma_op_to_string (arguments_list_p[0]);
+
+  if (JERRY_UNLIKELY (str_p == NULL))
+  {
+    return str_p;
+  }
+
+  if (arguments_list_len == 2)
```
galpeter: Is there a test case for this `if`? For some reason it's bit awkward for me. I don't know at the moment which case is this.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -219,8 +367,36 @@ ecma_op_create_function_object (ecma_object_t *scope_p, /**< function's scope */
...
+ *
+ * @return pointer to newly created Function object
+ */
+ecma_object_t *
+ecma_op_create_generator_function_object (ecma_object_t *scope_p, /**< function's scope */
+                                          const ecma_compiled_code_t *bytecode_data_p) /**< byte-code array */
+{
+  return ecma_op_create_function_object (scope_p, bytecode_data_p, ECMA_BUILTIN_ID_GENERATOR);
+} /* ecma_op_create_generator_function_object */
+
```
galpeter: Do we need these methods at all? Can't we just use the `..._create_function_object` directly?

tests/jerry/es2015/generator-function.js
```diff
@@ -0,0 +1,65 @@
...
+
+  item = iterator.next();
+  assert(item.value === 3);
+  assert(item.done === false);
+
+  item = iterator.next();
+  assert(item.value === undefined);
+  assert(item.done === true);
+
+  assert (g.constructor === (function*(){}).constructor);
```
galpeter: There is a space before the parentheses however at other places there is no space. It would be good to have them in a consistent state.

jerry-core/ecma/operations/ecma-function-object.c
rerobika: Yes they can be called directly, but I thinks these helper functions are inlined by the compiler. I don't insist to this methods, but the used API has not changed only has been renamed (that was the institution).

jerry-core/ecma/operations/ecma-function-object.c
rerobika: if the `arguments_list_len == 2` it means a builtin function ctor is called like this: `Function('arg1', 'body')`, so no comma separator is needed after the first argument.  I see no corresponding test so I'll add a new tc for it.

PR #3502, opened 2020-01-08
Add Map, Set, WeakMap, WeakSet basic functionality to the API

DanielBallaSZTE:

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -71,8 +71,25 @@ Possible compile time enabled feature types:
  - JERRY_FEATURE_SYMBOL - symbol support
  - JERRY_FEATURE_DATAVIEW - DataView support
  - JERRY_FEATURE_PROXY - Proxy support
+ - JERRY_FEATURE_MAP - Map support
+ - JERRY_FEATURE_SET - Set support
+ - JERRY_FEATURE_WEAKMAP - WeakMap support
+ - JERRY_FEATURE_WEAKSET - WeakSet support
 
 *New in version 2.0*.
+*Changed in version 2.3* : Added `JERRY_FEATURE_WEAKMAP`, `JERRY_FEATURE_WEAKSET` values.
+
+## jerry_container_type_t
+
+Container object types:
+
+ - JERRY_CONTAINER_TYPE_INVALID - Invalid container
+ - JERRY_CONTAINER_TYPE_MAP - Map type
+ - JERRY_CONTAINER_TYPE_SET - Set type
+ - JERRY_CONTAINER_TYPE_WEAKMAP - WeakMap type
+ - JERRY_CONTAINER_TYPE_WEAKSET - WeakSet type
+
+ *New in version 2.3*.
 
 ## jerry_regexp_flags_t
 
@@ -2070,6 +2087,64 @@ main (void)
 **See also**
 
 - [jerry_create_typedarray](#jerry_create_typedarray)
+
+
+## jerry_get_container_type
+
+**Summary**
+
+Checks whether the given `jerry_value_t` is the given `jerry_container_type_t` type container object.
+
+*Notes*
+- This API function depends on a build option (`JERRY_ES2015_BUILTIN_CONTAINER`) and can be checked
+  runtime with the `JERRY_FEATURE_MAP, JERRY_FEATURE_SET, JERRY_FEATURE_WEAKMAP, JERRY_FEATURE_WEAKSET`
+  feature enum values.
+  see: [jerry_is_feature_enabled](#jerry_is_feature_enabled).
+- The ES2015-subset profile enables this by default.
+
+*New in version 2.3*.
+
+**Prototype**
+
+```c
+jerry_container_type_t
+jerry_get_container_type (const jerry_value_t value)
+```
+
+- `value` - Container object
+- return value
+  - The corresponding enum value of `jerry_container_type_t`, or `JERRY_CONTAINER_TYPE_INVALID` if the container
+    was not a valid container object.
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t value = jerry_create_container (JERRY_CONTAINER_TYPE_MAP, NULL, 0);
+
+  if (jerry_get_container_type (value) == JERRY_CONTAINER_TYPE_MAP)
+  {
+    /* "value" is a map. */
+  }
+
+  jerry_release_value (value);
+
+  jerry_cleanup ();
+
+  return 0;
+}
+```
+
+**See also**
+
+- [jerry_create_container](#jerry_create_container)
+- [jerry_container_type_t](#jerry_container_type_t)
 
 
 ## jerry_value_is_undefined
@@ -4800,6 +4875,70 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
 - [jerry_typedarray_type_t](#jerry_typedarray_type_t)
 - [jerry_value_is_typedarray](#jerry_value_is_typedarray)
 - [jerry_release_value](#jerry_release_value)
+
+
+## jerry_create_container
+
+**Summary**
+
+Create a jerry_value_t representing a given type container object.
+
+*Notes*:
+- This method is expected to work the same way as the JavaScript Map constructor.
+- Returned value must be freed with [jerry_release_value](#jerry_release_value)
+  when it is no longer needed.
+- This API depends on a build option (`JERRY_ES2015_BUILTIN_CONTAINER`) and can be checked
+  in runtime with the `JERRY_FEATURE_MAP, JERRY_FEATURE_SET, JERRY_FEATURE_WEAKMAP, JERRY_FEATURE_WEAKSET`
+  feature enum values.
+  see: [jerry_is_feature_enabled](#jerry_is_feature_enabled).
+- The ES2015-subset profile enables this by default.
+
+*New in version 2.3*.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_container (jerry_container_type_t container_type,
+                        const jerry_value_t *arguments_list_p,
+                        jerry_length_t arguments_list_len);
+```
+
+- `container_type` - Type of the container to be created, see `jerry_container_type_t`.
+- `arguments_list_p` - The arguments passed to the container constructor to be inserted to the container.
+- `arguments_list_len` - The length of the above arguments.
+- return value - the new container object as a `jerry_value_t`
+
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+  jerry_char_t src[] = "[1,2,3,4].entries()";
+  jerry_value_t iterable = jerry_eval (src, sizeof (src) - 1, JERRY_PARSE_NO_OPTS);
+
+  jerry_value_t map = jerry_create_container (JERRY_CONTAINER_TYPE_MAP, &iterable, 1);
+  jerry_release_value (iterable);
+
+  // use the Map
+
+  jerry_release_value (map);
+
+  jerry_cleanup ();
+  return 0;
+}
+```
+
+**See also**
+
+- [jerry_container_type_t](#jerry_container_type_t)
+- [jerry_get_container_type](#jerry_get_container_type)
 
 
 ## jerry_create_undefined
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -22,6 +22,7 @@
 #include "ecma-builtin-helpers.h"
 #include "ecma-builtins.h"
 #include "ecma-comparison.h"
+#include "ecma-container-object.h"
 #include "ecma-dataview-object.h"
 #include "ecma-exceptions.h"
 #include "ecma-eval.h"
@@ -1001,6 +1002,18 @@ jerry_is_feature_enabled (const jerry_feature_t feature) /**< feature to check *
 #if ENABLED (JERRY_LOGGING)
           || feature == JERRY_FEATURE_LOGGING
 #endif /* ENABLED (JERRY_LOGGING) */
+#if ENABLED (JERRY_ES2015_BUILTIN_MAP)
+          || feature == JERRY_FEATURE_MAP
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_MAP) */
+#if ENABLED (JERRY_ES2015_BUILTIN_SET)
+          || feature == JERRY_FEATURE_SET
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_SET) */
+#if ENABLED (JERRY_ES2015_BUILTIN_WEAKMAP)
+          || feature == JERRY_FEATURE_WEAKMAP
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_WEAKMAP) */
+#if ENABLED (JERRY_ES2015_BUILTIN_WEAKSET)
+          || feature == JERRY_FEATURE_WEAKSET
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_WEAKSET) */
           );
 } /* jerry_is_feature_enabled */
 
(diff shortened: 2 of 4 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -2021,6 +2035,65 @@ main (void)
...
+  runtime with the `JERRY_FEATURE_MAP, JERRY_FEATURE_SET, JERRY_FEATURE_WEAKMAP, JERRY_FEATURE_WEAKSET`
+  feature enum values.
+  see: [jerry_is_feature_enabled](#jerry_is_feature_enabled).
+- The ES2015-subset profile enables this by default.
+
+**Prototype**
+
+```c
+bool
+jerry_value_is_container_type(const jerry_value_t value,
```
rerobika: space before `(`

docs/02.API-REFERENCE.md
```diff
@@ -2021,6 +2035,65 @@ main (void)
...
+
+  return 0;
+}
+```
+
+**See also**
+
+- [jerry_create_container](#jerry_create_container)
+- [jerry_container_type_t](#jerry_container_type_t)
+- [jerry_release_value](#jerry_release_value)
```
rerobika: I think this reference can be removed.

docs/02.API-REFERENCE.md
```diff
@@ -4692,6 +4765,53 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
...
+
+**Summary**
+
+Create a jerry_value_t representing a given type container object.
+
+*Notes*:
+- Returned value must be freed with [jerry_release_value](#jerry_release_value)
+  when it is no longer needed.
+- This API depends on a build option (`JERRY_ES2015_BUILTIN_CONTAINER`) and can be checked
+  in runtime with the `JERRY_FEATURE_CONTAINER` feature enum value,
```
rerobika: `JERRY_FEATURE_CONTAINER` is not introduced at all.

docs/02.API-REFERENCE.md
```diff
@@ -4692,6 +4765,53 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
...
+
+  jerry_release_value (map);
+}
+```
+
+**See also**
+
+- [jerry_container_type_t](#jerry_container_type_t)
+- [jerry_value_is_container_type](#jerry_value_is_container_type)
+- [jerry_release_value](#jerry_release_value)
```
rerobika: ditto.

jerry-core/api/jerry.c
```diff
@@ -4197,6 +4210,130 @@ jerry_json_stringify (const jerry_value_t object_to_stringify) /**< a jerry_obje
...
+  {
+    if (ecma_is_value_error_reference (arguments_list_p[i]))
+    {
+      return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+    }
+  }
+
+  switch (container_type)
+  {
+    case JERRY_CONTAINER_TYPE_MAP:
```
rerobika: The cases should be guarded with the corresponding container guards.

jerry-core/api/jerry.c
```diff
@@ -4197,6 +4210,130 @@ jerry_json_stringify (const jerry_value_t object_to_stringify) /**< a jerry_obje
...
+ *
+ * @return true if the above conditions are satisfied, false otherwise.
+ */
+bool
+jerry_value_is_container_type (const jerry_value_t value, /**< the container object */
+                               jerry_container_type_t container_type) /**< Type of the container */
+{
+  jerry_assert_api_available ();
+
+#if ENABLED (JERRY_ES2015_BUILTIN_CONTAINER)
```
rerobika: Ditto.

jerry-core/api/jerry.c
```diff
@@ -4197,6 +4210,130 @@ jerry_json_stringify (const jerry_value_t object_to_stringify) /**< a jerry_obje
...
+    {
+      type = LIT_MAGIC_STRING_WEAKMAP_UL;
+      break;
+    }
+    case JERRY_CONTAINER_TYPE_WEAKSET:
+    {
+      type = LIT_MAGIC_STRING_WEAKSET_UL;
+      break;
+    }
+
```
rerobika: no new line is needed.

jerry-core/api/jerry.c
```diff
@@ -4197,6 +4210,130 @@ jerry_json_stringify (const jerry_value_t object_to_stringify) /**< a jerry_obje
...
+    default:
+    {
+      return false;
+    }
+  }
```
rerobika: one new line please.

tests/unit-core/test-container.c
```diff
@@ -0,0 +1,81 @@
...
+
+#include "jerryscript.h"
+#include "test-common.h"
+
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  if (!jerry_is_feature_enabled (JERRY_FEATURE_MAP) || !jerry_is_feature_enabled (JERRY_FEATURE_SET)
```
rerobika: all conditions must be places into new lines.

tests/unit-core/test-container.c
```diff
@@ -0,0 +1,81 @@
...
+      || !jerry_is_feature_enabled (JERRY_FEATURE_WEAKMAP) || !jerry_is_feature_enabled (JERRY_FEATURE_WEAKSET))
+  {
+    jerry_port_log (JERRY_LOG_LEVEL_ERROR, "Containers are disabled!\n");
+    jerry_cleanup ();
+    return 0;
+  }
+
+  // Should always be true, no need to be freed.
```
rerobika: However there is no allocated memory for the boolean direct value, in the ECMA layer releasing the value is unnecessary, but in the API layer is required.

DanielBallaSZTE: The `jerry_binary_operation` can return an error, and the API reference also warns the user to release the returned value.

docs/02.API-REFERENCE.md
```diff
@@ -2021,6 +2036,61 @@ main (void)
...
+**Summary**
+
+Checks whether the given `jerry_value_t` is the given `jerry_container_type_t` type container object.
+
+*Notes*
+- This API function depends on a build option (`JERRY_ES2015_BUILTIN_CONTAINER`) and can be checked
+  runtime with the `JERRY_FEATURE_MAP, JERRY_FEATURE_SET, JERRY_FEATURE_WEAKMAP, JERRY_FEATURE_WEAKSET`
+  feature enum values.
+  see: [jerry_is_feature_enabled](#jerry_is_feature_enabled).
+- The ES2015-subset profile enables this by default.
```
galpeter: The documentation should also contain in which version was this API method added. Please see other post-Jerry 1.0 method documentations.

docs/02.API-REFERENCE.md
```diff
@@ -2021,6 +2036,61 @@ main (void)
...
+**Prototype**
+
+```c
+jerry_container_type_t
+jerry_get_container_type (const jerry_value_t value)
+```
+
+- `value` - Container object
+- return value
+  - The corresponding enum value of `jerry_container_type_t`
```
galpeter: IMHO the doc should mention what happens will be returned if the input value is not a valid container type.

docs/02.API-REFERENCE.md
```diff
@@ -4692,6 +4762,53 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
...
+  in runtime with the `JERRY_FEATURE_MAP, JERRY_FEATURE_SET, JERRY_FEATURE_WEAKMAP, JERRY_FEATURE_WEAKSET`
+  feature enum values.
+  see: [jerry_is_feature_enabled](#jerry_is_feature_enabled).
+- The ES2015-subset profile enables this by default.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_container (const jerry_value_t *arguments_list_p,
```
galpeter: IMHO it would be better to have the container type as the first argument. Usually that is the most important one.

docs/02.API-REFERENCE.md
```diff
@@ -4692,6 +4762,53 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
...
+Create a jerry_value_t representing a given type container object.
+
+*Notes*:
+- Returned value must be freed with [jerry_release_value](#jerry_release_value)
+  when it is no longer needed.
+- This API depends on a build option (`JERRY_ES2015_BUILTIN_CONTAINER`) and can be checked
+  in runtime with the `JERRY_FEATURE_MAP, JERRY_FEATURE_SET, JERRY_FEATURE_WEAKMAP, JERRY_FEATURE_WEAKSET`
+  feature enum values.
+  see: [jerry_is_feature_enabled](#jerry_is_feature_enabled).
+- The ES2015-subset profile enables this by default.
```
galpeter: Same as above: please provide version information.

docs/02.API-REFERENCE.md
galpeter: In addition what will happen if the INVALID container type is used?

docs/02.API-REFERENCE.md
```diff
@@ -4692,6 +4762,53 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
...
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_container (const jerry_value_t *arguments_list_p,
+                        jerry_length_t arguments_list_len,
+                        jerry_container_type_t container_type);
+```
+
+- `arguments_list_p` - The arguments passed to the container constructor.
```
galpeter: What does these arguments mean? : Is this the size of the Map/set? The elements which would be inserted?

docs/02.API-REFERENCE.md
```diff
@@ -4692,6 +4762,53 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
...
+- `arguments_list_len` - The length of the above arguments.
+- `container_type` - Type of the container to be created, see `jerry_container_type_t`.
+- return value - the new container object as a `jerry_value_t`
+
+**Example**
+
+```c
+{
+  jerry_value_t map = jerry_create_container (NULL, 0, JERRY_CONTAINER_TYPE_MAP);
```
galpeter: It would be good to have a "real" example which uses some arguments for the map. For example: initialize the map with a set of key-values.

jerry-core/include/jerryscript-core.h
```diff
@@ -666,6 +670,19 @@ typedef enum
...
+/**
+ * Container types.
+ */
+typedef enum
+{
+  JERRY_CONTAINER_TYPE_MAP = 0, /**< Map type */
+  JERRY_CONTAINER_TYPE_SET, /**< Set type */
+  JERRY_CONTAINER_TYPE_WEAKMAP, /**< WeakMap type */
+  JERRY_CONTAINER_TYPE_WEAKSET, /**< WeakSet type */
+  JERRY_CONTAINER_TYPE_INVALID, /**< Invalid container */
```
galpeter: It would be better to have the INVALID as the first element (just like in case of typedarrays). That way one can do something like this:
```c
jerry_container_type_t type = jerry_get_container_type (value);
if (!type)
{
   // do some error handling
}
```

Having the invalid elements as a value of 0 would make it consistent. Of course it is best to write out the _INVALID enum value in the check.

tests/unit-core/test-container.c
```diff
@@ -0,0 +1,86 @@
...
+  jerry_value_t global_weakset = jerry_get_property (global, weakset_str);
+
+  jerry_release_value (global);
+
+  jerry_release_value (map_str);
+  jerry_release_value (set_str);
+  jerry_release_value (weakmap_str);
+  jerry_release_value (weakset_str);
+
+  jerry_value_t empty_map = jerry_create_container (NULL, 0, JERRY_CONTAINER_TYPE_MAP);
```
galpeter: All of these are empty containers, could we have tests for containers with contents? If so it would be good to have test functions for them. Part of the testing will be probably in JS as some of the container operations are not in the API.

docs/02.API-REFERENCE.md
```diff
@@ -70,9 +70,24 @@ Possible compile time enabled feature types:
...
+## jerry_container_type_t
+
+Container object types:
+
+ - JERRY_CONTAINER_TYPE_MAP - Map type
+ - JERRY_CONTAINER_TYPE_SET - Set type
+ - JERRY_CONTAINER_TYPE_WEAKMAP - WeakMap type
+ - JERRY_CONTAINER_TYPE_WEAKSET - WeakSet type
+ - JERRY_CONTAINER_TYPE_INVALID - Invalid container
+
```
galpeter: Missing version information. (see above)

docs/02.API-REFERENCE.md
```diff
@@ -70,9 +70,24 @@ Possible compile time enabled feature types:
  - JERRY_FEATURE_LOGGING - logging
  - JERRY_FEATURE_SYMBOL - symbol support
  - JERRY_FEATURE_DATAVIEW - DataView support
+ - JERRY_FEATURE_MAP - Map support
+ - JERRY_FEATURE_SET - Set support
+ - JERRY_FEATURE_WEAKMAP - WeakMap support
+ - JERRY_FEATURE_WEAKSET - WeakSet support
 
 *New in version 2.0*.
```
galpeter: Please describe the version changes also. ex.
```md
*New in version 2.0*
*Changed in version xxx* : Added JERRY_FEATURE_MAP, .... values.
```

docs/02.API-REFERENCE.md
```diff
@@ -2021,6 +2036,61 @@ main (void)
...
+  jerry_value_t value = jerry_create_container (NULL, 0, JERRY_CONTAINER_TYPE_MAP);
+
+  if (jerry_get_container_type (value) == JERRY_CONTAINER_TYPE_MAP)
+  {
+    /* "value" is a map. */
+  }
+
+  jerry_release_value (value);
+
+  jerry_cleanup();
```
galpeter: There should be a space before the opening parentheses.

docs/02.API-REFERENCE.md
```diff
@@ -2021,6 +2036,61 @@ main (void)
 - [jerry_create_typedarray](#jerry_create_typedarray)
 
 
+## jerry_get_container_type
```
galpeter: This doc entry should be placed after the `jerry_value_get_type` method (it is a bit similar to that).
In addition what if we use the name `jerry_value_get_container_type` ?

docs/02.API-REFERENCE.md
DanielBallaSZTE: An error is thrown with the message `Invalid container type.`

tests/unit-core/test-container.c
DanielBallaSZTE: The API function part is covered by these tests, calling them with initial elements would cover the same part as our JS tests do, since the builtin routines (e.g. `ecma_builtin_map_dispatch_construct`) are calling the same functions as these API functions do (e.g. `ecma_op_container_create`).

docs/02.API-REFERENCE.md
DanielBallaSZTE: The previous name was `jerry_value_get_container_type` however other API functions use the `jerry_get_...` (`jerry_get_error_type`, `jerry_get_array_length`, `jerry_get_string_size`) scheme without the `value`.

docs/02.API-REFERENCE.md
```diff
@@ -70,8 +70,25 @@ Possible compile time enabled feature types:
  - JERRY_FEATURE_LOGGING - logging
  - JERRY_FEATURE_SYMBOL - symbol support
  - JERRY_FEATURE_DATAVIEW - DataView support
+ - JERRY_FEATURE_MAP - Map support
+ - JERRY_FEATURE_SET - Set support
+ - JERRY_FEATURE_WEAKMAP - WeakMap support
+ - JERRY_FEATURE_WEAKSET - WeakSet support
 
 *New in version 2.0*.
+*Changed in version 2.3* : Added `JERRY_FEATURE_WEAKMAP`, `JERRY_FEATURE_WEAKSET` values.
```
galpeter: The next release version would be `2.2`.

docs/02.API-REFERENCE.md
```diff
@@ -4692,6 +4767,69 @@ jerry_create_typedarray_for_arraybuffer_sz (jerry_typedarray_type_t type_name,
...
+#include "jerryscript.h"
+
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+  jerry_char_t src[] = "[1,2,3,4].entries()";
+  jerry_value_t iterable = jerry_eval (src, sizeof (src) - 1, JERRY_PARSE_NO_OPTS);
+
+  jerry_value_t map = jerry_create_container (JERRY_CONTAINER_TYPE_MAP, &iterable, 1);
```
galpeter: What would happen if I pass an js array to the argument list?

DanielBallaSZTE: The same thing that would happen if you passed an Array to the new Map() constructor in JavaScript. It would throw a TypeError.
You can expect the API to work the same as the JavaScript, since it only wraps the builtin routines.

galpeter: I see, then it would be good to describe this in the docs above. That this method will work the same way as the JS constructor calls.

DanielBallaSZTE: Good idea!

docs/02.API-REFERENCE.md
```diff
@@ -70,8 +70,25 @@ Possible compile time enabled feature types:
  - JERRY_FEATURE_LOGGING - logging
  - JERRY_FEATURE_SYMBOL - symbol support
  - JERRY_FEATURE_DATAVIEW - DataView support
+ - JERRY_FEATURE_MAP - Map support
+ - JERRY_FEATURE_SET - Set support
+ - JERRY_FEATURE_WEAKMAP - WeakMap support
+ - JERRY_FEATURE_WEAKSET - WeakSet support
 
 *New in version 2.0*.
+*Changed in version 2.2* : Added `JERRY_FEATURE_WEAKMAP`, `JERRY_FEATURE_WEAKSET` values.
```
galpeter: All other entries refer to 2.3 should this be also 2.3?

PR #3505, opened 2020-01-10
Fix identifier parsing for ES5.1 profile

### LINKED ISSUE #3498: ICE: Assertion 'lit_is_valid_cesu8_string (string_p, string_size)'
###### JerryScript revision
It started to assert since 40d930d62c518361218e232d8600a9960472776b , but valid on ToT too.

###### Build platform
platform independent

###### Build steps
$ tools/build.py --debug --clean --error-messages=ON --logging=ON

###### Test case
tests/jerry/fail/regression-test-issue-1871-1.js
tests/jerry/fail/regression-test-issue-1871-2.js

###### Output
ICE: Assertion 'lit_is_valid_cesu8_string (string_p, string_size)' failed at /home/oszi/jerryscript/jerry-core/ecma/base/ecma-helpers-string.c(ecma_new_ecma_string_from_utf8):378.
Aborted (core dumped)

Unfortunately we didn't catch this assert previously, because it is an expected failing test and the test runner accepts crash/assert as normal fail.

rerobika:
Code points outside of the basic plane cannot be part of an identifier.
This patch fixes #3498.


jerry-core/parser/js/js-lexer.c
```diff
@@ -722,6 +722,8 @@ lexer_parse_identifier (parser_context_t *context_p, /**< context */
         decoded_length = 2 * 3;
         has_escape = true;
       }
+#else /* !ENABLED (JERRY_ES2015) */
+      parser_raise_error (context_p, PARSER_ERR_INVALID_IDENTIFIER_PART);
```
zherczeg: This should behave the same way as other invalid characters, So:
```
if (length == 0)
{
  return false;
}
break;
```

PR #3507, opened 2020-01-13
Add option to list symbols in Object.getOwnPropertyNames()

pmarkee:
Also a little refactoring of Object.getOwnPropertyNames()

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-object.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-object.c
@@ -290,7 +290,7 @@ ecma_builtin_object_object_get_own_property_names (ecma_object_t *obj_p) /**< ro
 static ecma_value_t
 ecma_builtin_object_object_get_own_property_symbols (ecma_object_t *obj_p) /**< routine's argument */
 {
-  return ecma_builtin_helper_object_get_properties (obj_p, ECMA_LIST_SYMBOLS);
+  return ecma_builtin_helper_object_get_properties (obj_p, ECMA_LIST_SYMBOLS_ONLY);
 } /* ecma_builtin_object_object_get_own_property_symbols */
 
 #endif /* ENABLED (JERRY_ES2015) */
@@ -781,15 +781,15 @@ ecma_builtin_object_object_assign (ecma_object_t *target_p, /**< target object *
     ecma_object_t *from_obj_p = ecma_get_object_from_value (from_value);
 
     /* 5.b.iii */
-    /* TODO: extends this collection if symbols will be supported */
     ecma_collection_t *props_p = ecma_op_object_get_property_names (from_obj_p, ECMA_LIST_CONVERT_FAST_ARRAYS
-                                                                                | ECMA_LIST_ENUMERABLE);
+                                                                                | ECMA_LIST_ENUMERABLE
+                                                                                | ECMA_LIST_SYMBOLS);
 
     ecma_value_t *buffer_p = props_p->buffer_p;
 
     for (uint32_t j = 0; (j < props_p->item_count) && ecma_is_value_empty (ret_value); j++)
     {
-      ecma_string_t *property_name_p = ecma_get_string_from_value (buffer_p[j]);
+      ecma_string_t *property_name_p = ecma_get_prop_name_from_value (buffer_p[j]);
 
       /* 5.c.i-ii */
       ecma_property_descriptor_t prop_desc;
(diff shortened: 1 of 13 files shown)


jerry-core/ecma/base/ecma-helpers-collection.c
```diff
@@ -144,6 +144,31 @@ ecma_collection_push_back (ecma_collection_t *collection_p, /**< value collectio
...
+                        uint32_t index /**< index of the element to be removed */)
+{
+  JERRY_ASSERT (collection_p != NULL);
+  JERRY_ASSERT (collection_p->item_count > 0);
+  JERRY_ASSERT (index < collection_p->item_count);
+
+  ecma_value_t *buffer_p = collection_p->buffer_p;
+  const uint32_t len = collection_p->item_count;
+
+  ecma_free_value (buffer_p[index]);
```
rerobika: That should be the caller's responsibility.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -783,13 +783,14 @@ ecma_builtin_object_object_assign (ecma_object_t *target_p, /**< target object *
     /* 5.b.iii */
     /* TODO: extends this collection if symbols will be supported */
     ecma_collection_t *props_p = ecma_op_object_get_property_names (from_obj_p, ECMA_LIST_CONVERT_FAST_ARRAYS
```
rerobika: IMO the Todo comment above can be removed.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2081,7 +2107,8 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
 
                  Also if is_symbols_only is true and prop_name is not symbol
                  we should skip the current property e.g. Object.getOwnPropertySymbols. */
-              if (JERRY_UNLIKELY (is_symbols_only != ecma_prop_name_is_symbol (name_p)))
+              bool is_symbol = ecma_prop_name_is_symbol (name_p);
```
rerobika: Please update the comment above.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2112,6 +2139,32 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
 
               if (is_add)
               {
+                uint32_t index = ecma_string_get_array_index (name_p);
+                if (index != ECMA_STRING_NOT_ARRAY_INDEX)
```
rerobika: NO local variable is needed.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1969,6 +1972,28 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
...
+      uint32_t index = ecma_string_get_array_index (name_p);
+
+      if (index != ECMA_STRING_NOT_ARRAY_INDEX)
+      {
+        array_index_named_properties_count++;
+      }
+      else if (is_array_indices_only)
+      {
+        ecma_free_value (buffer_p[i]);
+        ecma_collection_remove (prop_names_p, i);
```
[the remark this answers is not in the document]
pmarkee: Good idea, this is not a very big change in these functions and then we don't need to add `ecma_collection_remove` at all.

pmarkee:
Thanks for the review @rerobika , I updated the PR.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -891,18 +891,16 @@ ecma_builtin_try_to_instantiate_property (ecma_object_t *object_p, /**< object *
  */
 void
 ecma_builtin_list_lazy_property_names (ecma_object_t *object_p, /**< a built-in object */
-                                       bool separate_enumerable, /**< true -  list enumerable properties into
-                                                                  *           main collection, and non-enumerable
-                                                                  *           to collection of 'skipped non-enumerable'
-                                                                  *           properties,
-                                                                  *   false - list all properties into main collection.
-                                                                  */
+                                       uint32_t opts, /**< listing options */
```
galpeter: Could we update the comment to reference the `ecma_list_properties_options_t` enum?
like: `uint32_t opts, /**< listing options using flags from ecma_list_properties_options_t */`

or maybe we should just mention it at the method's comments? What do you think?
(btw: this also applies to the other places where the `opts` is used)

pmarkee:
Thanks for the review @galpeter . I have added a test case including symbols to the `Object.assign()` test, and a new test file to the `es2015` directory to check that `Object.getOwnPropertyNames()` does not include symbols by default.

PR #3512, opened 2020-01-14
Make run-test-suite.py handle crashes properly

ossy-szeged:
Changes:
- Crashing jerry / jerry-snapshot is always "FAIL",
only the 1 return code is accepted as expected fail.
- "FAIL (XPASS)" string is only used if the test marked
as expected fail, but the return code is 0.
- Simple "FAIL" string is used if the test marked as expected
fail, but the return code is not 0 or 1. (crash)


ossy-szeged:
Note: It is a regression, the old sh runner worked properly, but I missed to add crash handling to the new python runner. We realized this bug after #3498.

CI will be red now, can be landed after #3498 is fixed.

PR #3513, opened 2020-01-14
Add strict mode check to for-in and for-of header in parser_parse_for_statement_start

szilagyiadam:

### DIFF
--- a/jerry-core/parser/js/js-parser-statm.c
+++ b/jerry-core/parser/js/js-parser-statm.c
@@ -1168,6 +1168,11 @@ parser_parse_for_statement_start (parser_context_t *context_p) /**< context */
         || context_p->token.type == LEXER_KEYW_LET
         || context_p->token.type == LEXER_KEYW_CONST)
     {
+      if (context_p->status_flags & PARSER_IS_STRICT)
+      {
+        parser_raise_error (context_p, PARSER_ERR_FOR_IN_OF_DECLARATION);
+      }
+
       token_type = context_p->token.type;
       has_context = (context_p->token.type != LEXER_KEYW_VAR);
       scanner_get_location (&start_location, context_p);
--- a/jerry-core/parser/js/js-parser-util.c
+++ b/jerry-core/parser/js/js-parser-util.c
@@ -951,6 +951,10 @@ parser_error_to_string (parser_error_t error) /**< error code */
     {
       return "Incorrect use of yield keyword.";
     }
+    case PARSER_ERR_FOR_IN_OF_DECLARATION:
+    {
+      return "for in-of loop variable declaration may not have an initializer.";
+    }
 #endif /* ENABLED (JERRY_ES2015) */
     case PARSER_ERR_DELETE_IDENT_NOT_ALLOWED:
     {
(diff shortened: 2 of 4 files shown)


jerry-core/parser/js/js-parser-statm.c
```diff
@@ -1168,15 +1168,22 @@ parser_parse_for_statement_start (parser_context_t *context_p) /**< context */
...
-
-      /* TODO: remove this after the pre-scanner supports strict mode detection. */
-      if (context_p->next_scanner_info_p->source_p == context_p->source_p
-          && context_p->next_scanner_info_p->type == SCANNER_TYPE_LET_EXPRESSION)
+      if (context_p->status_flags & PARSER_IS_STRICT)
       {
-        scanner_release_next (context_p, sizeof (scanner_info_t));
+        parser_raise_error (context_p, PARSER_ERR_FOR_IN_DECLARATION);
+      }
+      else
```
rerobika: This else block is not needed.

tests/jerry/es2015/forin-header-strict.js
```diff
@@ -0,0 +1,22 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+try {
+  eval('for (var i = 0 in {}) {}');
+}
+catch(e) {
```
rerobika: Move the catch one line upwards.

jerry-core/parser/js/js-parser-util.c
```diff
@@ -951,6 +951,10 @@ parser_error_to_string (parser_error_t error) /**< error code */
     {
       return "Incorrect use of yield keyword.";
     }
+    case PARSER_ERR_FOR_IN_DECLARATION:
+    {
+      return "for-in loop variable declaration may not have an initializer.";
+    }
```
zherczeg: What about for-of? Is this rule ES2015 only?

tests/jerry/es2015/forin-header-strict.js
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+'use strict';
+try {
+  eval('for (var i = 0 in {}) {}');
```
galpeter: There should be an `assert(false)` after the eval to make sure that the eval did throw an error,

jerry-core/parser/js/js-parser-util.c
```diff
@@ -951,6 +951,14 @@ parser_error_to_string (parser_error_t error) /**< error code */
     {
       return "Incorrect use of yield keyword.";
     }
+    case PARSER_ERR_FOR_IN_DECLARATION:
+    {
+      return "for-in loop variable declaration may not have an initializer.";
```
rerobika: I`d use only one new error message like `for in-of loop ...`

PR #3517, opened 2020-01-14
Improve new.target handling in case of eval calls

galpeter:
Only direct eval calls propagate the "new.target" information.
In any other eval cases the "new.target" invocation is a syntax error.

Added test cases also.


galpeter:
Before landing this we'll need to add an api function (almost done, a bit of work needed).

galpeter:
I've updated the PR.

PR #3518, opened 2020-01-14
Add runtime option "--call-on-exit [FUNCNAME]" to main-unix.c

DanielBallaSZTE:
With this option you can call a function after the user script and promises have ran, to be able to do assertions that are executed just before the process would exit.

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -325,7 +325,8 @@ typedef enum
   OPT_EXEC_SNAP,
   OPT_EXEC_SNAP_FUNC,
   OPT_LOG_LEVEL,
-  OPT_NO_PROMPT
+  OPT_NO_PROMPT,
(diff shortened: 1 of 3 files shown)


jerry-main/main-unix.c
```diff
@@ -365,6 +366,8 @@ static const cli_opt_t main_opts[] =
                .help = "set log level (0-3)"),
   CLI_OPT_DEF (.id = OPT_NO_PROMPT, .longopt = "no-prompt",
                .help = "don't print prompt in REPL mode"),
+  CLI_OPT_DEF (.id = OPT_CALL_ON_EXIT, .longopt = "call-on-exit", .meta = "STRING",
+               .help = "Invokes the specified function when the process is just about to exit."),
```
zherczeg: The other descriptions don't use full sentences.

PR #3522, opened 2020-01-15
Introduce new.target C api

galpeter:
Added new "jerry_get_new_target" API function and updated the unit test.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -7654,6 +7654,138 @@ main (void)
 
 - [jerry_create_external_function](#jerry_create_external_function)
 
+## jerry_get_new_target
+
+**Summary**
+
+Returns the current "new.target" JavaScript function at the call site.
+
+If used outside of a native C function it will return "undefined" value.
+
+*Notes*:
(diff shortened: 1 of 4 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -7654,6 +7654,138 @@ main (void)
...
+
+Returns the current "new.target" JavaScript function at the call site.
+
+If used outside of a native C function it will return "undefined" value.
+
+*Notes*:
+- Returned value must be freed with [jerry_release_value](#jerry_release_value) when it
+is no longer needed.
+- This feature depends on build option (`JERRY_ES2015`) and can be checked
+  in runtime with the `JERRY_FEATURE_SYMBOL` feature enum value (as symbols are enabled in case of ES2015),
```
rerobika: Are you sure about this sentence?

galpeter: The ES2015 does not have its own feature enum value. If the `JERRY_ES2015` build option is set the symbol feature will be enabled (symbols does not have their own build flag). That's why we can check the symbol feature if the "base" es2015 feature set is required.

PR #3538, opened 2020-01-28
Update the RegExp constructor to ECMA-262 v6

szilagyiadam:
The following methods have been implemented:
- RegExpAlloc, based on ECMA-262 v6, 21.2.3.2.1
- RegExpInitialize, based on ECMA-262 v6, 22.2.3.2.2
- RegExpCreate, based on ECMA-262 v6, 21.2.3.2.3

Co-authored-by: Robert Fancsik frobert@inf.u-szeged.hu

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -31,6 +31,7 @@
 #include "ecma-helpers.h"
 #include "ecma-init-finalize.h"
 #include "ecma-lex-env.h"
+#include "lit-char-helpers.h"
 #include "ecma-literal-storage.h"
 #include "ecma-objects.h"
 #include "ecma-objects-general.h"
@@ -1773,11 +1774,20 @@ jerry_create_regexp_sz (const jerry_char_t *pattern_p, /**< zero-terminated UTF-
     return jerry_throw (ecma_raise_common_error (ECMA_ERR_MSG ("Input must be a valid utf8 string")));
   }
 
+  ecma_object_t *regexp_obj_p = ecma_op_regexp_alloc (NULL);
+
+  if (JERRY_UNLIKELY (regexp_obj_p == NULL))
+  {
+    return ECMA_VALUE_ERROR;
+  }
+
   ecma_string_t *ecma_pattern = ecma_new_ecma_string_from_utf8 (pattern_p, pattern_size);
 
-  jerry_value_t ret_val = ecma_op_create_regexp_object (ecma_pattern, flags);
-
+  jerry_value_t ret_val = ecma_op_create_regexp_with_flags (regexp_obj_p,
+                                                            ecma_make_string_value (ecma_pattern),
+                                                            flags);
   ecma_deref_ecma_string (ecma_pattern);
+
   return ret_val;
 
 #else /* !ENABLED (JERRY_BUILTIN_REGEXP) */
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
@@ -60,21 +60,23 @@ static ecma_value_t
 ecma_builtin_regexp_prototype_flags_helper (ecma_value_t this, /**< this value */
                                             uint16_t *flags_p) /**< [out] flags */
 {
-  if (!ecma_object_is_regexp_object (this))
+  ecma_extended_object_t *ext_obj_p = ecma_op_check_object_type_is_class (this);
+
+  if (ext_obj_p != NULL)
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Incompatible type"));
+    if (ext_obj_p->u.class_prop.class_id == LIT_MAGIC_STRING_REGEXP_UL
+        || ext_obj_p->u.class_prop.class_id == LIT_INTERNAL_MAGIC_STRING_REGEXP_PROTO)
+    {
+      re_compiled_code_t *bc_p = ECMA_GET_INTERNAL_VALUE_POINTER (re_compiled_code_t,
+                                                                  ext_obj_p->u.class_prop.u.value);
+
+      *flags_p = bc_p->header.status_flags;
+
+      return ECMA_VALUE_EMPTY;
+    }
   }
 
-  ecma_extended_object_t *re_obj_p = (ecma_extended_object_t *) ecma_get_object_from_value (this);
-  re_compiled_code_t *bc_p = ECMA_GET_INTERNAL_VALUE_ANY_POINTER (re_compiled_code_t,
-                                                                  re_obj_p->u.class_prop.u.value);
-
-  if (bc_p != NULL)
-  {
-    *flags_p = bc_p->header.status_flags;
-  }
-
-  return ECMA_VALUE_EMPTY;
+  return ecma_raise_type_error (ECMA_ERR_MSG ("Incompatible type"));
 } /* ecma_builtin_regexp_prototype_flags_helper */
 
 /**
@@ -220,21 +222,21 @@ ecma_op_escape_regexp_pattern (ecma_string_t *pattern_str_p) /**< RegExp pattern
 static ecma_value_t
 ecma_builtin_regexp_prototype_get_source (ecma_value_t this_arg) /**< this argument */
 {
-  if (!ecma_object_is_regexp_object (this_arg))
+  ecma_extended_object_t *ext_obj_p = ecma_op_check_object_type_is_class (this_arg);
+
+  if (ext_obj_p != NULL)
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Incompatible type"));
+    if (ext_obj_p->u.class_prop.class_id == LIT_MAGIC_STRING_REGEXP_UL
+        || ext_obj_p->u.class_prop.class_id == LIT_INTERNAL_MAGIC_STRING_REGEXP_PROTO)
+    {
+      re_compiled_code_t *bc_p = ECMA_GET_INTERNAL_VALUE_POINTER (re_compiled_code_t,
+                                                                  ext_obj_p->u.class_prop.u.value);
+
+      return ecma_op_escape_regexp_pattern (ecma_get_string_from_value (bc_p->source));
+    }
   }
 
-  ecma_extended_object_t *re_obj_p = (ecma_extended_object_t *) ecma_get_object_from_value (this_arg);
-  re_compiled_code_t *bc_p = ECMA_GET_INTERNAL_VALUE_ANY_POINTER (re_compiled_code_t,
-                                                                  re_obj_p->u.class_prop.u.value);
-
-  if (bc_p != NULL)
-  {
-    return ecma_op_escape_regexp_pattern (ecma_get_string_from_value (bc_p->source));
-  }
-
-  return ecma_make_string_value (ecma_get_magic_string (LIT_MAGIC_STRING_EMPTY_NON_CAPTURE_GROUP));
+  return ecma_raise_type_error (ECMA_ERR_MSG ("Incompatible type"));
 } /* ecma_builtin_regexp_prototype_get_source */
 
 /**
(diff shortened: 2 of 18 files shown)


jerry-core/lit/lit-magic-strings.inc.h
```diff
@@ -268,7 +268,10 @@ LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_LOG10E_U, "LOG10E")
 #endif
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_NUMBER_UL, "Number")
 LIT_MAGIC_STRING_DEF (LIT_MAGIC_STRING_OBJECT_UL, "Object")
+#if ENABLED (JERRY_BUILTIN_REGEXP) \
+|| ENABLED (JERRY_ES2015)
```
galpeter: Shouldn't this only depend on the builtin regexp?

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -60,21 +60,28 @@ static ecma_value_t
...
 
-  if (bc_p != NULL)
-  {
-    *flags_p = bc_p->header.status_flags;
+    if (ecma_get_object_type (obj_p) == ECMA_OBJECT_TYPE_CLASS)
+    {
+      ecma_extended_object_t *ext_obj_p = (ecma_extended_object_t *) obj_p;
+      if (ext_obj_p->u.class_prop.class_id == LIT_MAGIC_STRING_REGEXP_UL
+          || ext_obj_p->u.class_prop.class_id == LIT_MAGIC_STRING_REGEXP_MATCHER)
+      {
```
galpeter: Would it be possible to have a function which does this check? ( And optionally returns the extended object?)

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -60,21 +60,28 @@ static ecma_value_t
...
+        ecma_extended_object_t *re_obj_p = (ecma_extended_object_t *) ecma_get_object_from_value (this);
```
galpeter: At this point the `obj_p` and `ext_obj_p` would already have the object pointer

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
szilagyiadam: Do you mean like this?
```.js
static bool
ecma_op_check_object_type_is_class (ecma_value_t this, /**< this value */
                                    ecma_extended_object_t **ext_obj_p) /**< [out] pointer */
{
  if (ecma_is_value_object (this))
  {
    ecma_object_t *obj_p = ecma_get_object_from_value (this);

    if (ecma_get_object_type (obj_p) == ECMA_OBJECT_TYPE_CLASS)
    {
      *ext_obj_p = (ecma_extended_object_t *) obj_p;
      return true;
    }
  }

  return false;
}
```

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -238,86 +245,195 @@ ecma_regexp_unicode_advance (const lit_utf8_byte_t **str_p, /**< reference to st
...
+ecma_op_regexp_initialize (ecma_object_t *obj_p, /**< RegExp object */
+                           ecma_value_t pattern_value, /**< pattern of the RegExp object */
+                           ecma_value_t flags_value) /**< flags of the RegExp object */
 {
-  JERRY_ASSERT (pattern_p != NULL);
+  ecma_string_t *pattern_str_p = NULL;
+  uint16_t flags = 0;
+  const re_compiled_code_t *bc_p;
+
+  if (!ecma_is_value_empty (flags_value))
```
dbatyai: I don't really like this approach of creating a jack-of-all-trades function to initialize regexp instances. I especially don't like the fact that the bytecode is occasionally passed not-so-transparently as the pattern value.

I'd much prefer if we split this code up into logical parts, and not jumble it together. We could have separate functions to handle regexp allocation/construction, bytecode creation, and regexp initialization, and combine these as needed into two functions that create a regexp instance from a pattern or a bytecode respectively.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -125,87 +125,94 @@ ecma_regexp_parse_flags (ecma_string_t *flags_str_p, /**< Input string with flag
...
- * Helper function to assign a value to a property
+ * Update the properties of a RegExp instance.
  */
 static void
-ecma_regexp_helper_assign_prop (ecma_object_t *re_object_p, /**< RegExp object */
-                                lit_magic_string_id_t prop_id, /**< property name ide */
-                                ecma_value_t value) /**< value */
+ecma_regexp_update_props (ecma_object_t *re_object_p, /**< RegExp object */
+                          ecma_string_t *source_p, /**< source string */
+                          uint16_t flags) /**< flags */
```
dbatyai: Could you group the code in this function a bit logically? It's not very nice to read.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -238,86 +245,195 @@ ecma_regexp_unicode_advance (const lit_utf8_byte_t **str_p, /**< reference to st
...
+                       ecma_value_t flags_value) /**< flags of the RegExp object */
+{
   ecma_object_t *re_prototype_obj_p = ecma_builtin_get (ECMA_BUILTIN_ID_REGEXP_PROTOTYPE);
 
-  ecma_object_t *object_p = ecma_create_object (re_prototype_obj_p,
-                                                sizeof (ecma_extended_object_t),
-                                                ECMA_OBJECT_TYPE_CLASS);
-  ecma_extended_object_t *ext_object_p = (ecma_extended_object_t *) object_p;
-  ext_object_p->u.class_prop.class_id = LIT_MAGIC_STRING_UNDEFINED;
+  ecma_value_t ctr_value = ecma_op_object_get_by_magic_id (re_prototype_obj_p, LIT_MAGIC_STRING_CONSTRUCTOR);
```
dbatyai: Can we move this into `ecma_op_regexp_alloc`?

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -50,6 +50,30 @@
...
 
 #if ENABLED (JERRY_ES2015)
+
+/**
+ * Helper function to check if the given value is object
+ *
+ * @return bool
+ */
+static bool
+ecma_op_check_object_type_is_class (ecma_value_t this, /**< this value */
```
dbatyai: Could we do this without an output parameter? For example by returning with the extended object pointer or NULL.

jerry-core/api/jerry.c
```diff
@@ -1710,10 +1711,37 @@ jerry_create_regexp_sz (const jerry_char_t *pattern_p, /**< zero-terminated UTF-
...
+  if (flags & RE_FLAG_STICKY)
+  {
+    ecma_stringbuilder_append_char (&builder, LIT_CHAR_LOWERCASE_S);
+  }
+  if (flags & RE_FLAG_UNICODE)
+  {
+    ecma_stringbuilder_append_char (&builder, LIT_CHAR_LOWERCASE_U);
+  }
+
+  ecma_value_t flags_value = ecma_make_string_value (ecma_stringbuilder_finalize (&builder));
```
dbatyai: I'd really prefer if we didn't create this string just so that we can re-parse it later. There should be a way to create a regexp with already parsed flags.

szilagyiadam:
Made an update according to the suggestions. Now we have different methods for creating a RegExp object from pattern, bytecode, and with parsed flags.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -131,47 +162,66 @@ ecma_builtin_regexp_dispatch_construct (const ecma_value_t *arguments_list_p, /*
...
+  else
+  {
+#endif /* ENABLED (JERRY_ES2015) */
 
-    if (JERRY_UNLIKELY (flags_string_p == NULL))
-    {
-      ecma_deref_ecma_string (pattern_string_p);
-      goto cleanup;
-    }
+    ret_value = ecma_op_create_regexp_from_pattern (new_target_obj_p, pattern_value, flags_value);
```
dbatyai: I think `pattern_value` and `flags_value` can leak here if `pattern_is_regexp` is true.

jerry-core/lit/lit-magic-strings.h
```diff
@@ -77,6 +78,19 @@ typedef enum
...
+/**
+ * the [[RegExpMatcher]] internal slot handling in ECMA v6 and v5.1
+ */
+#if ENABLED (JERRY_ES2015)
+#define LIT_MAGIC_STRING_REGEXP_MATCHER LIT_INTERNAL_MAGIC_STRING_REGEXP_MATCHER
+#else /* !ENABLED (JERRY_ES2015) */
+#if ENABLED (JERRY_BUILTIN_REGEXP)
+#define LIT_MAGIC_STRING_REGEXP_MATCHER LIT_MAGIC_STRING_REGEXP_UL
+#else /*!ENABLED (JERRY_BUILTIN_REGEXP) */
+#define LIT_MAGIC_STRING_REGEXP_MATCHER LIT_MAGIC_STRING_OBJECT_UL
```
dbatyai: Why do we need to define this as `LIT_MAGIC_STRING_OBJECT_UL`? Cases with this magic string should not be reachable if the regexp builtins are disabled.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -90,36 +73,77 @@ ecma_builtin_regexp_dispatch_construct (const ecma_value_t *arguments_list_p, /*
...
+    {
+      create_regexp_from_bc = true;
+    }
+    else
+    {
+      create_regexp_from_bc = false;
+      pattern_value = bc_p->source;
+    }
+
+    pattern_is_regexp = false;
```
galpeter: Is this ok? The `pattern_is_regexp` is set to `false` if the pattern was a regexp object.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -90,36 +73,77 @@ ecma_builtin_regexp_dispatch_construct (const ecma_value_t *arguments_list_p, /*
...
+
+    if (ecma_is_value_undefined (flags_value))
+    {
+      create_regexp_from_bc = true;
+    }
+    else
+    {
+      create_regexp_from_bc = false;
+      pattern_value = bc_p->source;
+    }
```
galpeter: This could be simplified like:
```c
create_regexp_from_bc = ecma_is_value_undefined (flags_value);

if (!create_regexp_from_bc) {
  ecma_extended_object_t *pattern_obj_p = (ecma_extended_object_t *) ecma_get_object_from_value (pattern_value);
  bc_p = ECMA_GET_INTERNAL_VALUE_POINTER (re_compiled_code_t,
                                          pattern_obj_p->u.class_prop.u.value);

  pattern_value = bc_p->source;
}
```

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -238,86 +231,233 @@ ecma_regexp_unicode_advance (const lit_utf8_byte_t **str_p, /**< reference to st
 #endif /* ENABLED (JERRY_ES2015) */
 
 /**
- * RegExp object creation operation.
+ * RegExpAlloc method
  *
- * See also: ECMA-262 v5, 15.10.4.1
+ * See also: ECMA-262 v6, 21.2.3.2.1
```
galpeter: It would be good to keep the reference to the v5.1 standard.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -238,86 +231,233 @@ ecma_regexp_unicode_advance (const lit_utf8_byte_t **str_p, /**< reference to st
...
+    ecma_value_t parse_flags_value = ecma_regexp_parse_flags (flags_str_p, &flags);
+    ecma_deref_ecma_string (flags_str_p);
+
+    if (ECMA_IS_VALUE_ERROR (parse_flags_value))
+    {
+      ecma_deref_ecma_string (pattern_str_p);
+      return parse_flags_value;
+    }
+
+    ecma_free_value (parse_flags_value);
```
galpeter: If I understand correctly then `ecma_regexp_parse_flags` can only return `VALUE_EMPTY` or an error. If so then this `free_value` call could be turned into an assert to check if the parse result is `VALUE_EMTPY`. (Just like on line `396`.)

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
szilagyiadam: It is set to false, because in the standard (ECMA-262 v6, 21.2.3.1) step 6 is an else if statement, so if we done the previous step, we should skip the next statement. But becasue of the backwards compatibility, i cant't use an else if statement. I couldn't come up with a better solution.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -41,34 +43,15 @@
...
+ecma_builtin_regexp_dispatch_helper (const ecma_value_t *arguments_list_p, /**< arguments list */
+                                     ecma_length_t arguments_list_len, /**< number of arguments */
+                                     ecma_object_t *new_target_p) /**< pointer to the new target object */
 {
   ecma_value_t pattern_value = ECMA_VALUE_UNDEFINED;
   ecma_value_t flags_value = ECMA_VALUE_UNDEFINED;
 #if ENABLED (JERRY_ES2015)
-  ecma_value_t new_pattern = ECMA_VALUE_EMPTY;
-  ecma_value_t new_flags = ECMA_VALUE_EMPTY;
+  bool create_regexp_from_bc = false;
```
galpeter: This could be moved closer to the first actual usage.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
galpeter: I see, then maybe we could shuffle around things. Would this work?
```c
#if ENABLED(...) // this could be probably mered into the '#if' above
if (ecma_object_is_regexp_object (pattern_value))   // 21.2.3.1: 5. step
{
  create_regexp_from_bc = ecma_is_value_undefined (flags_value);
  if (!create_regexp_from_bc)
  { // Note that the   bc_p  assignment was moved inside this if
    ecma_extended_object_t *pattern_obj_p = (ecma_extended_object_t *) ecma_get_object_from_value (pattern_value);
    bc_p = ECMA_GET_INTERNAL_VALUE_POINTER (re_compiled_code_t,
                                          pattern_obj_p->u.class_prop.u.value);
    pattern_value = bc_p->source;
  }
}
else if (pattern_is_regexp) // 21.2.3.1: 6. step
{
  ecma_object_t *pattern_obj_p = ecma_get_object_from_value (pattern_value);
  pattern_value = ecma_op_object_get_by_magic_id (pattern_obj_p, LIT_MAGIC_STRING_SOURCE);

  if (ECMA_IS_VALUE_ERROR (pattern_value))
  {
    return pattern_value;
  }

  if (ecma_is_value_undefined (flags_value))
  {
    flags_value = ecma_op_object_get_by_magic_id (pattern_obj_p, LIT_MAGIC_STRING_FLAGS);

    if (ECMA_IS_VALUE_ERROR (flags_value))
    {
      return flags_value;
    }
  }
}
#else /* .. */
if (ecma_object_is_regexp_object (pattern_value))
{
  if (ecma_is_value_undefined (flags_value))
  {
    return ecma_copy_value (pattern_value);
  }

  return ecma_raise_type_error (ECMA_ERR_MSG ("Invalid argument of RegExp call."));
}
#endif /* .. */
```

The idea would be to clearly separate the 5.1 and 6.0 standard and it would a bit more readable. What do you thing?

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
galpeter: Or even better remove it and use the `bc_p != NULL` check.

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
szilagyiadam: This can work, hovewer the bc_p initialization can't be moved inside that if statement, because we need it at the end of the function to create the RegExp object, if the create_regexp_from_bc is true

PR #3541, opened 2020-01-30
For in-of declaration should only throw error after initialization.

rerobika:
This patch fixes the error introduced in #3513.

### DIFF
--- a/tests/jerry/es2015/forin-header-strict.js
+++ b/tests/jerry/es2015/forin-header-strict.js
@@ -27,3 +27,35 @@ try {
 } catch (e) {
   assert(e instanceof SyntaxError);
 }
+
+var reached = false;
+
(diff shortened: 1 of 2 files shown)


tests/jerry/es2015/forin-header-strict.js
```diff
@@ -27,3 +27,35 @@ try {
...
 }
+
+var reached = false;
+
+for (var i in {}) {
+  reached = true;
+}
+assert(!reached);
+
+for (var i of []) {
```
galpeter: The `reached` should be reseted to a `false` state, otherwise the first "true" will make all other assert-s pass.

galpeter: ahh.. I've misunderstood the check. Sorry.

PR #3546, opened 2020-02-06
Implement Object.prototype.__proto__ accessor property

szilagyiadam:
We are using the already existing Object.getPrototypeOf and
Object.setProtoypeOf methods


jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -835,6 +835,96 @@ ecma_builtin_helper_def_prop (ecma_object_t *obj_p, /**< object */
...
+  while (true)
+  {
+    if (iterator_p == NULL)
+    {
+      break;
+    }
+
+    if (obj_p == iterator_p)
+    {
+      return ecma_raise_type_error (ECMA_ERR_MSG ("Cyclic __proto__ value"));
```
dbatyai: We already have function for [Get/Set]PrototypeOf, can't we reuse those instead of duplicating the logic?

jerry-core/vm/vm.c
```diff
@@ -91,6 +92,21 @@ vm_op_get_value (ecma_value_t object, /**< base object */
     else if (ecma_is_value_string (property))
     {
       property_name_p = ecma_get_string_from_value (property);
+
+#if ENABLED (JERRY_ES2015)
+      if (JERRY_UNLIKELY (ecma_compare_ecma_string_to_magic_id (property_name_p, LIT_MAGIC_STRING__PROTO__)))
+      {
+        ecma_value_t to_obj_val = ecma_op_to_object (object);
```
rerobika: `object` is already a valid object so this should no be `toObject`ed.

jerry-core/vm/vm.c
```diff
@@ -1569,6 +1579,13 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
           }
 
           const int index = (int) (opcode_data >> VM_OC_NON_STATIC_SHIFT) - 2;
+
+          if (JERRY_UNLIKELY (ecma_compare_ecma_string_to_magic_id (prop_name_p, LIT_MAGIC_STRING__PROTO__)))
+          {
+            result = ecma_builtin_object_object_set_prototype_of (stack_top_p[index], left_value, true);
```
dbatyai: What if instead of adding an argument to the `set_prototype_of` function, we call it as we would normally and release the resulting value, and just set the vm result to undefined? Would that be feasible?
@rerobika What is your opinion?

rerobika: +1

szilagyiadam: I've made a new method based on ECMA-262 v6, B.2.2.1.2 to separate the different behaviours, so if we want to set the __proto__ accessor poperty, we use this new method.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -157,6 +157,7 @@ ecma_builtin_object_object_get_prototype_of (ecma_object_t *obj_p) /**< routine'
 } /* ecma_builtin_object_object_get_prototype_of */
 
 #if ENABLED (JERRY_ES2015)
+
```
dbatyai: Please revert this newline

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -216,11 +217,74 @@ ecma_builtin_object_object_set_prototype_of (ecma_value_t arg1, /**< routine's f
   }
 
   JERRY_ASSERT (ecma_is_value_true (status));
+
```
dbatyai: This as well

PR #3558, opened 2020-02-13
Fix lexical scoping between scripts

dbatyai:
This change fixes the handling of lexical blocks when executing multiple
scripts, and also fixes a few issues with module environments.

After this change, all script files will run in the same context and
will have access to lexically scoped global variables of previous
scripts, and module environments will no longer have a bound global 'this'
value.

The REPL implementation in main-unix is also fixed to correctly handle
lexically scoped variables.


jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1302,7 +1299,8 @@ scanner_detect_invalid_var (parser_context_t *context_p, /**< context */
     }
   }
 
-  if ((context_p->status_flags & PARSER_IS_EVAL)
+  if (((context_p->status_flags & PARSER_IS_EVAL)
+       || literal_pool_p->prev_p == NULL)
```
zherczeg: This can be kept in one line.

dbatyai:
In the standard parsing and execution cannot be done separately, each specified scipt should be parsed exactly once, and if no error occured, should be executed exactly once immediatly after, so normally this shouldn't be a problem.

However, since in our case parsing and exection can be done separately, we'll either have to do runtime checks for previous `let` declarations if we would like to allow executing a parsed script multiple times or in different order, or we''ll have to make sure that evaluation is done according to the standard.

jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1738,6 +1731,96 @@ scanner_create_unused_literal (parser_context_t *context_p, /**< context */
...
+  JERRY_ASSERT (info_p->type == SCANNER_TYPE_FUNCTION);
+
+  literal.char_p = info_p->source_p - 1;
+
+  while (next_data_p[0] != SCANNER_STREAM_TYPE_END)
+  {
+    uint32_t type = next_data_p[0] & SCANNER_STREAM_TYPE_MASK;
+    const uint8_t *data_p = next_data_p;
+
+    if (type == SCANNER_STREAM_TYPE_HOLE)
```
zherczeg: I think this type is not allowed in global scope. I would add an assert instead.

jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1738,6 +1731,96 @@ scanner_create_unused_literal (parser_context_t *context_p, /**< context */
...
+    {
+      next_data_p++;
+      continue;
+    }
+
+    JERRY_ASSERT (type != SCANNER_STREAM_TYPE_HOLE
```
zherczeg: The assert is here, just delete the lines above.

PR #3562, opened 2020-02-17
Implement the core of Proxy object

rerobika:
- Internal routines of the of the proxy object are unimplemented
 - For-in enumerate with proxy target is currently not supported


jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.c
```diff
@@ -210,11 +211,37 @@ ecma_builtin_function_prototype_object_bind (ecma_object_t *this_arg_obj_p , /**
   /* 4. 11. 18. */
   ecma_object_t *prototype_obj_p;
 
+#if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
```
galpeter: This code part has a bit too much marcos and bit hard to read (macro guarded {}-s). Would it make sense to move this to it's own function?
like:
```c
ecma_object_t *bind_get_proto (ecma_object_t *this_arg_obj_p)
{
#if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
  if (ecma_object_is_proxy (this_arg_obj_p))
  {
    ecma_value_t proto = ecma_proxy_object_get_prototype_of (this_arg_obj_p);

    if (ECMA_IS_VALUE_ERROR (proto))
    {
      return proto;
    }
    if (ecma_is_value_null (proto))
    {
      return NULL;
    }
    return ecma_get_object_from_value (proto);
  }


#if ENABLED (JERRY_ES2015)
  ecma_object_t *prototype_obj_p = NULL;
  jmem_cpointer_t proto_cp = ecma_op_ordinary_object_get_prototype_of (this_arg_obj_p);
  if (proto_cp != JMEM_CP_NULL)
  {
    prototype_obj_p = ECMA_GET_NON_NULL_POINTER (ecma_object_t, proto_cp);
    ecma_ref_object (prototype_obj_p);
  }
  return prototype_obj_p;
  return ecma_builtin_get (ECMA_BUILTIN_ID_FUNCTION_PROTOTYPE);
}
```

jerry-core/ecma/builtin-objects/ecma-builtin-object-prototype.c
```diff
@@ -143,7 +144,23 @@ static ecma_value_t
 ecma_builtin_object_prototype_object_has_own_property (ecma_object_t *obj_p, /**< this argument */
                                                        ecma_string_t *prop_name_p) /**< first argument */
 {
-  return ecma_make_boolean_value (ecma_op_object_has_own_property (obj_p, prop_name_p));
+#if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
+  if (ecma_object_is_proxy (obj_p))
+  {
+    ecma_property_descriptor_t prop_desc;
+
+    ecma_value_t status = ecma_proxy_object_get_own_property_descriptor (obj_p, prop_name_p, &prop_desc);
```
galpeter: Would it make sense to pass `NULL` as the third argument indicating that we don't want any data back? If so we could spare the free call below.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -251,11 +188,37 @@ ecma_builtin_object_object_set_prototype_of (ecma_value_t arg1, /**< routine's f
...
+  }
+  else
+  {
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
+    status = ecma_op_ordinary_object_set_prototype_of (obj_p, arg2);
+#if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
+  }
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
+
+  if (ecma_is_value_false (status))
```
galpeter: Isn't this only possible if the `obj_p` is a proxy?

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -251,11 +188,37 @@ ecma_builtin_object_object_set_prototype_of (ecma_value_t arg1, /**< routine's f
...
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Cannot set [[Prototype]]."));
+  }
+
+  JERRY_ASSERT (ecma_is_value_true (status));
+  ecma_ref_object (obj_p);
```
galpeter: Why do we need the `ref_object` here?

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -295,57 +258,181 @@ ecma_builtin_object_object_get_own_property_symbols (ecma_object_t *obj_p) /**<
...
+  if (props_p == NULL)
+  {
+    return ECMA_VALUE_ERROR;
+  }
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
+
   ecma_value_t *buffer_p = props_p->buffer_p;
 
-  for (uint32_t i = 0; i < props_p->item_count; i++)
+  if (is_seal)
```
galpeter: Both parts of this if is very similar (I've spotted only 4-5 lines of difference, a single if). Based on the standard however there should be more differences. Is this correct? If it is , can we merge the two sides of the if somehow?

docs/02.API-REFERENCE.md
```diff
@@ -4212,6 +4260,63 @@ jerry_create_promise (void)
...
+  - Returned value must be freed with [jerry_release_value](#jerry_release_value)
+    when it is no longer needed.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_create_proxy (const jerry_value_t target,
+                    const jerry_value_t handler)
+```
```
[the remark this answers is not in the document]
rerobika: Sure, however this feature does not require more API functions than {create/is}_proxy, since it only relates to internal methods.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -280,12 +280,19 @@ ecma_builtin_helper_object_get_properties (ecma_object_t *obj_p, /**< object */
...
   JERRY_ASSERT (obj_p != NULL);
 
+  ecma_collection_t *props_p = ecma_op_object_get_property_names (obj_p, opts);
+
+#if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
+  if (props_p == NULL)
+  {
+    return ECMA_VALUE_ERROR;
+  }
+#endif /* ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
```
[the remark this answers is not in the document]
rerobika: Good idea!

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -393,39 +410,43 @@ ecma_op_object_get_property (ecma_object_t *object_p, /**< the object */
...
-                                                              &property_ref,
-                                                              ECMA_PROPERTY_GET_NO_OPTIONS);
-
-  return property != ECMA_PROPERTY_TYPE_NOT_FOUND && property != ECMA_PROPERTY_TYPE_NOT_FOUND_AND_STOP;
-} /* ecma_op_object_has_own_property */
-
-/**
- * Checks whether an object (including prototypes) has a named property
+ * See also:
+ *          ECMASCRIPT v6, 9.1.7.1
```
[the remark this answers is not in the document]
rerobika: Indeed, that was a typo.

jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.c
rerobika: I'm not sure about that makes this codepath more readable. Also it the Proxy.[[GetPrototypeOf]] can also throw an error therefore the

```
if (ECMA_IS_VALUE_ERROR (proto))
    {
      return proto;
    }
```
return is incorrect, since its an ecma-value. We can do a workaround by returning with 0x01 when error happens, but this returned value also should be check in the caller site. So IMO separating this part into an other function does not worth it. However the code can slightly be simplified, I'll update soon.

jerry-core/ecma/builtin-objects/ecma-builtin-object-prototype.c
rerobika: To be honest, I think the users 95% do not know about the existence of this function (the remaining 5% are JS engine developers :) ). The `ecma_proxy_object_get_own_property_descriptor` is a frequently used function, where you should fill all the structure's field with the current property data if the property exists. So i'd not burden with extra null checks the general proxy method due to this case where the descriptor's data is not needed at all.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
rerobika: No, it can happen when the target object is not extensible, or the new prototype object would cause cyclical prototype chain.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
rerobika: Yes, we need since we return with the this value, which is not `toObject`ed, and the reference for the this value at the end of the function call is released.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
rerobika: According to the ES5 spec they are really similar, but in ES6 there are significant differences. This should be updated, but out of this patch scope.

jerry-core/api/jerry.c
```diff
@@ -954,9 +974,6 @@ jerry_is_feature_enabled (const jerry_feature_t feature) /**< feature to check *
 #if ENABLED (JERRY_BUILTIN_JSON)
           || feature == JERRY_FEATURE_JSON
 #endif /* ENABLED (JERRY_BUILTIN_JSON) */
-#if ENABLED (JERRY_ES2015_BUILTIN_PROMISE)
```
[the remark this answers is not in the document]
rerobika: Oops, fixed.

rerobika:
@galpeter @dbatyai Thanks for the reviews, the PR is updated.

docs/02.API-REFERENCE.md
```diff
@@ -1865,6 +1866,55 @@ jerry_value_is_promise (const jerry_value_t value)
...
+
+- `value` - api value
+- return value
+  - true, if the given `jerry_value_t` is a proxy object
+  - false, otherwise
+
+**Example**
+
+*New in version 2.3*.
```
zherczeg: How do you know the next version number?

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -1139,11 +1204,15 @@ ecma_gc_free_object (ecma_object_t *object_p) /**< object to free */
         default:
         {
           /* The undefined id represents an uninitialized class. */
-          JERRY_ASSERT (ext_object_p->u.class_prop.class_id == LIT_MAGIC_STRING_UNDEFINED
-                        || ext_object_p->u.class_prop.class_id == LIT_MAGIC_STRING_ARGUMENTS_UL
-                        || ext_object_p->u.class_prop.class_id == LIT_MAGIC_STRING_BOOLEAN_UL
-                        || ext_object_p->u.class_prop.class_id == LIT_MAGIC_STRING_ERROR_UL
-                        || ext_object_p->u.class_prop.class_id == LIT_INTERNAL_MAGIC_STRING_INTERNAL_OBJECT);
+          bool is_uninitialized = (ext_object_p->u.class_prop.class_id == LIT_MAGIC_STRING_UNDEFINED
```
zherczeg: I would put an NDEBUG guard around here.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -159,6 +159,17 @@ typedef enum
...
 
+/**
+ * Check whether the given object has [[ProxyHandler]] and [[ProxyTarger]] internal slots
+ *
+ * @param obj_p ecma-object
+ */
+#if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
+#define ECMA_OBJECT_IS_PROXY(obj_p) (JERRY_UNLIKELY (ecma_get_object_type ((obj_p)) == ECMA_OBJECT_TYPE_PROXY))
+#else /* !ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
+#define ECMA_OBJECT_IS_PROXY(obj_p) (false)
```
zherczeg: Is there a code where this is used?

rerobika: Yes, in ordinary-object methods I use this macro to assert that the object cannot be proxy.

PR #3563, opened 2020-02-18
Fix Array.find and Array.findIndex omitting empty values.

rwalczyna:
Empty values in array were omitted. Predicate should be called
for every element from 0 to length.


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2297,41 +2297,44 @@ ecma_builtin_array_prototype_object_find (ecma_value_t predicate, /**< callback
       return get_value;
     }
 
-    if (ecma_is_value_found (get_value))
+    /* Predicate should be called also for empty elements */
+    if (!ecma_is_value_found (get_value))
```
dbatyai: We could use `ecma_op_object_get_by_uint32_index` to get the value, then we won't need this check anymore.

rwalczyna: Done, thanks.

PR #3567, opened 2020-02-19
Add es5.1 specific directory in jerry-test-suite

clover2123:
since compatibility issues between es5.1 and es6 exist,
es5.1 specific directory is newly added which includes es5.1 specific test


zherczeg:
The `jerry-test-suite` has an `es2015` directory. Perhaps we should create an `es5.1` directory and move those tests which are `es5.1` specific there.

clover2123:
@zherczeg 
Thank you for the review. I added es5.1 specific test list which has the compatibility problem with es6 standard. Now `es2015-subset profile` runs the whole tests in jerry-suite-test except some included in the es5.1 specific list.

zherczeg:
The patch is good for me, but I have a question. Would it be possible to delete `tests/jerry-test-suite/es51-profile-list` since it seems redundant now.

clover2123:
@zherczeg 
default profile (es5.1) uses `es51-profile-list` in `jerry-test-suite`. To remove this list, I think that a new option like `--skip-dir` is necessary to exclude test paths in `es2015` directory.

ossy-szeged:
I think we don't need to add --exclude-list feature to the test runner. The --skip-list is more than enough to do what we should solve here. 

tools/runners/run-test-suite.py is used by --jerry-tests and --jerry-test-suite options of tools/run-tests.py and jerry tests work fine with ES5 and ES2015 too. I suggest jerry-test-suite should have similar directory structure to jerry tests. There is an es5.1 an es2015 directory. run-tests script adds es5.1 to skip-list if running es2015 profile built jerry, and es2015 in case of es5.1. We can do the same thing with jerry-test-suite, just es5.1 directory is missing (for tests pass only with es5.1, but fail with es2015), and run-tests script should fine-tuned a little bit.

ossy-szeged:
I think we should do this in nutshell:
- move 13/13.02/13.02-002.js to tests/jerry-test-suite/es5.1
- delete es51-profile-list
- fix run_jerry_test_suitie in run-tests.py:

Adding test dir should be unconditional:
```
test_cmd.append('--test-dir')
test_cmd.append(settings.JERRY_TEST_SUITE_DIR)
```

Reuse this snippet from run_jerry_tests
```
skip_list = []

if '--profile=es2015-subset' in job.build_args:
	skip_list.append(os.path.join('es5.1', ''))
else:
	skip_list.append(os.path.join('es2015', ''))

if options.skip_list:
	skip_list.append(options.skip_list)

if skip_list:
	test_cmd.append("--skip-list=" + ",".join(skip_list))
```

The difference between run_jerry_tests and run_jerry_test_suite will be only the --profile=minimal case.

clover2123:
@ossy-szeged 
Thank you for the review. I fixed this patch by adding `es5.1` directory which is similar to `jerry-tests` case.

clover2123:
Thanks for merging this patch.
But it seems that the author is changed at this commit (https://github.com/jerryscript-project/jerryscript/commit/9633e85fb3c59515f40686b12aa455b9faf8a62a).
Is there any reason for this?

zherczeg:
No. I have never seen such thing before. This is the top commit so we could cut it from the repository, and you could re-submit the patch with a mew PR.

zherczeg:
I sent a mail to GitHub support. Lets see what they say.

zherczeg:
They have admitted the problem is on their side:

This is something the team claims to be working on, the workaround for this is to either let the author of the PR squash and merge it. or a maintainer can use the git command line rather than using the GitHub UI to do that.

We need to figure out what to do with this and future PRs. Landing everything manually can be difficult since we may miss a step,

clover2123:
@zherczeg 
Thank you for your support.
I'll hold this patch until you make a decision.

zherczeg:
It turned out Github made an experiment and now they reverted it because it caused more harm than good:
https://github.com/isaacs/github/issues/1303#issuecomment-595231303

The revert will be done in 24 hours, until that, we don't land any PRs. However we fix your patch and force push it.

dbatyai:
@clover2123
We updated the commit, you should now be credited as the author.

clover2123:
@zherczeg @dbatyai 
Thanks a lot for your support!

PR #3569, opened 2020-02-19
Optimize Array.prototype.fill for fast-array cases

szilagyiadam:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -2257,6 +2257,28 @@ ecma_builtin_array_prototype_fill (ecma_value_t value, /**< value */
     }
   }
 
+  if (ecma_op_object_is_fast_array (obj_p))
+  {
+    ecma_extended_object_t *ext_obj_p = (ecma_extended_object_t *) obj_p;
+
+    if (ext_obj_p->u.array.u.hole_count < ECMA_FAST_ARRAY_HOLE_ONE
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2233,6 +2233,27 @@ ecma_builtin_array_prototype_fill (ecma_value_t value, /**< value */
...
+
+    if (ext_obj_p->u.array.u.hole_count < ECMA_FAST_ARRAY_HOLE_ONE
+        && len != 0
+        && ecma_get_object_extensible (obj_p))
+    {
+      ecma_value_t *buffer_p = ECMA_GET_NON_NULL_POINTER (ecma_value_t, obj_p->u1.property_list_cp);
+
+      while (k < final)
+      {
+        buffer_p[k] = value;
```
dbatyai: The value in the buffer should be released before reassigning. It seems we don't have a proper test for this, we should add one where we call `fill` on an array that contains strings for example.

rerobika:
The patch looks go to me, just please share performance results before merging.

szilagyiadam:
**Performance results**

Intel: 0m10,45s -> 0m0,71s
ARM: 2m3s -> 0m10s

**Test case**
```.js
var array = []
for (var i = 0; i < 10000; i++)
{
  array[i] = i;
}

{
  array.fill(1);
  array.fill(2);
  array.fill(3);
  array.fill(4);
  array.fill(5);
}

```

PR #3570, opened 2020-02-20
Fix attributes of length property for function objects

clover2123:
length property of function objects should be configurable in ES2015.
Instead of lazy initialization, length property is added directly
when allocating each function object.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -1658,6 +1658,41 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
 {
   JERRY_ASSERT (!ecma_get_object_is_builtin (object_p));
 
+#if ENABLED (JERRY_ES2015)
+  if (ecma_compare_ecma_string_to_magic_id (property_name_p, LIT_MAGIC_STRING_LENGTH))
+  {
+    ecma_extended_object_t *ext_func_p = (ecma_extended_object_t *) object_p;
+    if (!ECMA_GET_FIRST_BIT_FROM_POINTER_TAG (ext_func_p->u.function.scope_cp))
+    {
+      /* Initialize 'length' property */
+      const ecma_compiled_code_t *bytecode_data_p = ecma_op_function_get_compiled_code (ext_func_p);
+      uint32_t len;
+      if (bytecode_data_p->status_flags & CBC_CODE_FLAGS_UINT16_ARGUMENTS)
+      {
+        cbc_uint16_arguments_t *args_p = (cbc_uint16_arguments_t *) bytecode_data_p;
+        len = args_p->argument_end;
+      }
+      else
+      {
+        cbc_uint8_arguments_t *args_p = (cbc_uint8_arguments_t *) bytecode_data_p;
+        len = args_p->argument_end;
+      }
+
+      /* Set tag bit to represent initialized 'length' property */
+      ECMA_SET_FIRST_BIT_TO_POINTER_TAG (ext_func_p->u.function.scope_cp);
+      ecma_property_t *value_prop_p;
+      ecma_property_value_t *value_p = ecma_create_named_data_property (object_p,
+                                                                        property_name_p,
+                                                                        ECMA_PROPERTY_FLAG_CONFIGURABLE,
+                                                                        &value_prop_p);
+      value_p->value = ecma_make_uint32_value (len);
+      return value_prop_p;
+    }
+
+    return NULL;
+  }
+#endif /* ENABLED (JERRY_ES2015) */
+
   if (ecma_compare_ecma_string_to_magic_id (property_name_p, LIT_MAGIC_STRING_PROTOTYPE)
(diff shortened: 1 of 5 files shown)


clover2123:
#3567 should be resolved ahead of this patch.

zherczeg:
JerryScript is a memory optimized engine, so we try to not allocate memory when it is not necessary. The `length` property of a function is a rarely used pointer so we could save some memory in ES2015 if create it when it is used.

In ES5.1 this property was "virtual", which means it is immutable so we never allocated memory for it. Similar virtual properties are the characters of a string, so `str[i]` just return with the i-th character of the string.

In ES2015 it is configurable, so we need to create the property when it is used, but only at the first time. To do that we need a bit which represents that the property was created. Looking at the function descriptor we have `scope_cp` and `bytecode_cp` members, which are "pointers". That means we have at least 3 * 2 bits which we can use freely. 

The problem is we have not used these bits in pointer fields before, so we don't have macros to extract the pointer value excluding the 3 bits. The infrastructure should be created first. Maybe we could only use the `scope_cp` for having 3 free bits first.

clover2123:
@zherczeg Thank you for the good suggestion.
I have looked more about "pointer" and summed up the issue as follows.
Please correct me if I'm wrong.

- Free 3 bits when a pointer value is converted to ecma_value_t.
1) Direct pointer case (ECMA_VALUE_CAN_STORE_UINTPTR_VALUE_DIRECTLY)
Each pointer has 0 value for the 3 least significant bits (0xXXX...8)
We can exploit these 3 LSB freely.
2) Compressed pointer case
Pointer is first set to the difference between the pointer and the start of the heap.
After then, pointer value is right shifted about 3bits.
So, the 3 most significant bits can be used.

From the further research, I found that `scope_cp` is differently stored in normal function and arrow function each. (normal function : ecma_value, arrow function : jmem_cpointer_t).
Would it be possible to unify the form of `scope_cp` to implement the bit notation easily? for example, change the form of `scope_cp` in normal function to jmem_cpointer_t.

zherczeg:
For ecma values, this function encodes the pointer with keeping the lower 3 bits free:

https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/ecma/base/ecma-helpers-value.c#L78

Of course you can unify the arrow functions to use normal scope pointers as well.

zherczeg:
This could be related: #3582

clover2123:
Thank you for the review.
I fixed this patch that `length` property is lazily initialized same as before but now has configurable attribute as specified in [ES6 19.2.4.1](https://www.ecma-international.org/ecma-262/6.0/#sec-function-instances-length).

But #3567 should be merged first.

clover2123:
This pr is ready for review.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -360,7 +360,6 @@ ecma_op_create_function_object (ecma_object_t *scope_p, /**< function's scope */
    *
    * See also: ecma_op_function_try_to_lazy_instantiate_property
    */
-
```
zherczeg: Why this line is deleted?

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1322,10 +1326,19 @@ ecma_op_object_put_with_receiver (ecma_object_t *object_p, /**< the object */
...
     else if (type == ECMA_OBJECT_TYPE_FUNCTION)
     {
+#if ENABLED (JERRY_ES2015)
+      /* Uninitialized 'length' property is non-writable (ECMA-262 v6, 19.2.4.1) */
+      if ((ecma_string_is_length (property_name_p))
+          && (!ECMA_GET_FIRST_BIT_FROM_POINTER_TAG (((ecma_extended_object_t *) object_p)->u.function.scope_cp)))
+      {
+        return ecma_reject (is_throw);
+      }
+#else /* !ENABLED (JERRY_ES2015) */
```
zherczeg: Would it work if we put `#if !ENABLED (JERRY_ES2015)` here and remove the code above? The property would be instantiated (I suspect this is very rare case), and rejected because it is non-writable by default.

jerry-core/ecma/operations/ecma-function-object.c
clover2123: It is accidentally deleted while adding some codes here and reverting that updates.
Should I recover the line?

jerry-core/ecma/operations/ecma-objects.c
clover2123: Yes, It would work correctly even if the above code removed.
I put the check code here to avoid instantiation of 'length' property.

jerry-core/ecma/operations/ecma-function-object.c
zherczeg: Please revert this change.

clover2123: Fixed it. Thank you.

rerobika:
@clover2123 Please rebase the PR to the latest master.

clover2123:
@rerobika Rebase done. Thanks.

PR #3571, opened 2020-02-20
Fix object destructuring for empty pattern

rerobika:


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -2797,6 +2797,15 @@ parser_parse_object_initializer (parser_context_t *context_p, /**< context */
 {
   parser_pattern_end_marker_t end_pos = parser_pattern_get_target (context_p, flags);
 
+  /* 12.14.5.2:  ObjectAssignmentPattern : { } */
+  if (lexer_check_next_character (context_p, LIT_CHAR_RIGHT_BRACE))
+  {
+    parser_emit_cbc_ext (context_p, CBC_EXT_REQUIRE_OBJECT_COERCIBLE);
+    lexer_next_token (context_p);
```
zherczeg: There is another function which skips a single byte.

rerobika: Indeed, fixed.

PR #3582, opened 2020-02-24
Remove ECMA_OBJECT_TYPE_ARROW

rerobika:
Until now arrow functions had a separate type, but most of the time they can be handled similar to single functions.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -63,22 +63,6 @@ ecma_op_resource_name (const ecma_compiled_code_t *bytecode_header_p)
 #endif /* ENABLED (JERRY_LINE_INFO) || ENABLED (JERRY_ES2015_MODULE_SYSTEM) */
 
 /**
- * Checks whether the type is a normal or arrow function.
- *
- * @return true - if the type is a normal or arrow function;
- *         false - otherwise
- */
-inline bool JERRY_ATTR_ALWAYS_INLINE
-ecma_is_normal_or_arrow_function (ecma_object_type_t type) /**< object type */
-{
-#if ENABLED (JERRY_ES2015)
-  return (type == ECMA_OBJECT_TYPE_FUNCTION || type == ECMA_OBJECT_TYPE_ARROW_FUNCTION);
-#else /* !ENABLED (JERRY_ES2015) */
-  return (type == ECMA_OBJECT_TYPE_FUNCTION);
-#endif /* ENABLED (JERRY_ES2015) */
-} /* ecma_is_normal_or_arrow_function */
-
-/**
  * IsCallable operation.
  *
  * See also: ECMA-262 v5, 9.11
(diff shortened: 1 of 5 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -683,59 +639,62 @@ ecma_op_function_has_instance (ecma_object_t *func_obj_p, /**< Function object *
 
 #if ENABLED (JERRY_ES2015)
 /**
- * Indicates whether the class has been invoked with 'new'.
+ * Indicates whether the function has been invoked with 'new'.
+ */
+#define ECMA_CONSTRUCT_FLAG ((uintptr_t) 0x01u)
+/**
+ * Indicates whether the super is called.
```
zherczeg: This could be a bit more descriptive.

zherczeg:
I think this comment is unnecessary: "The freed object type can be reused later for Proxy objects."

zherczeg:
simila -> similar in commit log

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1456,6 +1405,10 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
     /* This is not a super call so update the "new.target" */
     JERRY_CONTEXT (current_new_target) = func_obj_p;
   }
+  else
+  {
+    arguments_list_p = ecma_op_function_args_set_flag (arguments_list_p, ECMA_FUNC_ARG_FLAG_SUPER);
```
galpeter: If I read this correctly then: the `JERRY_CONTEXT (current_new_target)` is set inside the `..._call_simple` method. If so, then the contents of the `if` above is not needed. In addition with this change the `current_new_target` is saved twice onto the stack (once here and once in the `..._call_simple`).

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1465,14 +1418,14 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
...
-      arguments_list_p = ecma_op_function_set_construct_flag (arguments_list_p);
+      arguments_list_p = ecma_op_function_args_set_flag (arguments_list_p, ECMA_FUNC_ARG_FLAG_CONSTRUCT);
       ret_value = ecma_op_function_call_simple (func_obj_p, this_arg_value, arguments_list_p, arguments_list_len);
       break;
     }
 #if ENABLED (JERRY_ES2015)
     case ECMA_OBJECT_TYPE_BOUND_FUNCTION:
     {
-      JERRY_ASSERT (!ecma_op_function_has_construct_flag (arguments_list_p));
+      JERRY_ASSERT (!ecma_op_function_args_has_flag (arguments_list_p, ECMA_FUNC_ARG_FLAG_CONSTRUCT));
```
galpeter: Is it okay to only check the `CONSTRUCT` flag? Shouldn't the `SUPER` flag be checked here?

jerry-core/ecma/operations/ecma-function-object.c
rerobika: good idea.

jerry-core/ecma/operations/ecma-function-object.c
rerobika: Sure!

PR #3599, opened 2020-03-04
Implement Proxy object [[isExtensible]] internal method

szilagyiadam:
The algorithm is based on ECMA-262 v6, 9.5.3

### DIFF
--- a/jerry-core/ecma/operations/ecma-proxy-object.c
+++ b/jerry-core/ecma/operations/ecma-proxy-object.c
@@ -15,6 +15,7 @@
 
 #include "ecma-alloc.h"
 #include "ecma-builtins.h"
+#include "ecma-builtin-object.h"
 #include "ecma-exceptions.h"
 #include "ecma-function-object.h"
 #include "ecma-gc.h"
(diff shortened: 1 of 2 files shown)


tests/jerry/es2015/proxy_is_extensible.js
```diff
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO: Update these tests when the internal routine has been implemented
```
galpeter: It would be good to add a few extra tests like this:
```js
var target = {};

var h1 = {
    isExtensible(target) {
        print('h1 ext')
        return true;
    },
}

var h2 = {
    isExtensible(target) {
        print('h2 ext')
        return false;
    },
}

var p1 = new Proxy(target, h1);
var p2 = new Proxy(target, h2);

print(Object.isExtensible(p1));     // true
try {
    print(Object.isExtensible(p2));
} catch(ex) {
    print("expected typeerror ok");
}

var p3 = new Proxy(p1, h1);
var p4 = new Proxy(p1, h2);

print(Object.isExtensible(p3));     // true

try {
    print(Object.isExtensible(p4));
} catch(ex) {
    print("expected typeerror ok");
}

var p5 = new Proxy(p2, h2);

try {
    print(Object.isExtensible(p5));
} catch(ex) {
    print("expected typeerror ok");
}

Object.preventExtensions(target);
print(Object.isExtensible(p5));     // false

```

tests/jerry/es2015/proxy_is_extensible.js
```diff
@@ -26,21 +25,77 @@ try {
...
+    target.prop1 = false;
+    return Object.isExtensible(target);
+  }
+};
+
+var proxy = new Proxy(target, handler);
+assert(Object.isExtensible(proxy) === true);
+assert(target.prop1 === false);
+
+// trap result is invalid
```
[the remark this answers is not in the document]
szilagyiadam: Because by default, the target object is extensible so invoking the Object.isExtensible method with that target should return true. However, in the handler, we just simply return with false.

tests/jerry/es2015/proxy_is_extensible.js
szilagyiadam: I've been updated the test file, can you check it if it is good now?

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -310,8 +310,89 @@ ecma_value_t
...
+  {
+    return trap;
+  }
+
+  ecma_object_t *target_obj_p = ecma_get_object_from_value (target);
+
+  /* 7. */
+  if (ecma_is_value_undefined (trap))
+  {
+    if (ECMA_OBJECT_IS_PROXY (target_obj_p))
```
galpeter: It would be better to use the "general" method for is extensible here: `ecma_builtin_object_object_is_extensible`

rerobika: +1 for this idea.

tests/jerry/es2015/proxy_is_extensible.js
```diff
@@ -12,8 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO: Update these tests when the internal routine has been implemented
-
+// error checks
```
rerobika: Please include the license.

Issue #3592, opened 2020-02-26
Map.forEach omits newly added elements


rwalczyna:
As stated in [ES6 23.1.3.5](https://www.ecma-international.org/ecma-262/6.0/#sec-map.prototype.foreach), elements added after the call to __forEach__  begins should be visited. 
So new element added in callback needs to be visited after _old_ elements.

###### Test code:
```js
var map = new Map;
map.set('a', 1);
map.set('b', 2);
var l = 0;
map.forEach(function(value, key) {
  print(key, value);
  if(l === 0) {
    map.set('c', 3);
  }
  l++;
});
print("Loops:", l);
```
###### Output:
```
a 1
b 2
Loops: 2
```
###### Expected Output:
```
a 1
b 2
c 3
Loops: 3
```

I have checked current implementation and it looks like list of the properties' names is gathered at the beginning of execution and used till the end.
I think there should be some check after each callback execution, which would update properties' list.
The problem is in optimal implementation of this function - properties are stored in reversed singly linked list - new elements are added at the beginning (these elements should be listed at the end of foreach).

Something like internal iterator for collection object would be helpful, but this requires invalidating properties and deleting them not directly but a bit later by GC. But this needs a bit of additional memory for each collection's property and will be time consuming.

Is there any special reason why property list is reversed? Is this for optimization of adding new properties or it affects also implementations of other functions?

rtakacs:
@rwalczyna yes, this is a well known issue why Map - Set - WeakSet and WeakMap are not ECMA6 compliant. I've started to rework the collections so that will be a solution for this issue.

PR #3600, opened 2020-03-04
Use array based storage in container objects

rtakacs:
Currently, collections use object based solutions for storing elements and iterating on them. If an element is deleted and re-inserted, the storage position is the same as before so the iteration order is wrong. Newly added elements - during iterations - are also skipped.
This patch replaces the object based storage with an array based solution that helps to store and iterate elements as expected.

Fixes https://github.com/jerryscript-project/jerryscript/issues/3592

### DIFF
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -295,73 +295,106 @@ ecma_gc_mark_promise_object (ecma_extended_object_t *ext_object_p) /**< extended
 
 #endif /* ENABLED (JERRY_ES2015_BUILTIN_PROMISE) */
 
-#if ENABLED (JERRY_ES2015_BUILTIN_CONTAINER)
-
+#if ENABLED (JERRY_ES2015_BUILTIN_MAP)
 /**
- * Mark objects referenced by Map/Set built-in.
+ * Mark objects referenced by Map built-in.
  */
 static void
-ecma_gc_mark_container_object (ecma_object_t *object_p) /**< object */
+ecma_gc_mark_map_object (ecma_object_t *object_p) /**< object */
 {
+  JERRY_ASSERT (object_p != NULL);
+
   ecma_extended_object_t *map_object_p = (ecma_extended_object_t *) object_p;
-  ecma_object_t *internal_obj_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_object_t,
-                                                                   map_object_p->u.class_prop.u.value);
-  ecma_extended_object_t *internal_ext_p = (ecma_extended_object_t *) internal_obj_p;
+  ecma_collection_t *container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t,
+                                                                    map_object_p->u.class_prop.u.value);
+  ecma_value_t *start_p = ECMA_CONTAINER_START (container_p);
+  uint32_t entry_count = ECMA_CONTAINER_ENTRY_COUNT (container_p);
 
-  ecma_gc_set_object_visited (internal_obj_p);
-
-  jmem_cpointer_t prop_iter_cp = internal_obj_p->u1.property_list_cp;
-
-#if ENABLED (JERRY_PROPRETY_HASHMAP)
-  if (prop_iter_cp != JMEM_CP_NULL)
+  for (uint32_t i = 0; i < entry_count; i+= ECMA_CONTAINER_PAIR_SIZE)
   {
-    ecma_property_header_t *prop_iter_p = ECMA_GET_NON_NULL_POINTER (ecma_property_header_t,
-                                                                     prop_iter_cp);
-    if (prop_iter_p->types[0] == ECMA_PROPERTY_TYPE_HASHMAP)
+    ecma_container_pair_t *entry_p = (ecma_container_pair_t *) (start_p + i);
+
+    if (ecma_is_value_empty (entry_p->key))
     {
-      prop_iter_cp = prop_iter_p->next_property_cp;
-    }
-  }
-#endif /* ENABLED (JERRY_PROPRETY_HASHMAP) */
-
-  while (prop_iter_cp != JMEM_CP_NULL)
-  {
-    ecma_property_header_t *prop_iter_p = ECMA_GET_NON_NULL_POINTER (ecma_property_header_t, prop_iter_cp);
-    JERRY_ASSERT (ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
-
-    ecma_gc_mark_properties ((ecma_property_pair_t *) prop_iter_p);
-
-    if ((internal_ext_p->u.class_prop.extra_info & ECMA_CONTAINER_FLAGS_WEAK) == 0)
-    {
-      ecma_property_pair_t *prop_pair_p = (ecma_property_pair_t *) prop_iter_p;
-
-      for (uint32_t i = 0; i < ECMA_PROPERTY_PAIR_ITEM_COUNT; i++)
-      {
-        ecma_property_t *property_p = (ecma_property_t *) (prop_iter_p->types + i);
-
-        if (ECMA_PROPERTY_GET_NAME_TYPE (*property_p) == ECMA_DIRECT_STRING_PTR)
-        {
-          jmem_cpointer_t name_cp = prop_pair_p->names_cp[i];
-          ecma_string_t *prop_name_p = ECMA_GET_NON_NULL_POINTER (ecma_string_t, name_cp);
-
-          if (ECMA_STRING_GET_CONTAINER (prop_name_p) == ECMA_STRING_CONTAINER_MAP_KEY)
-          {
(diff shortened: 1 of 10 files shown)


rwalczyna:
I found two issues when testing the code:
1) When we remove last element from map, there is unexpected TypeError

###### Code example
```js
var map = new Map();
var iterationCount = 0;

map.set(0, 'a');
map.set(1, 'b');
map.set(2, 'c');

for (var [key, value] of map) {
  print(key, value);
  if(key === 0){
    map.delete(2)
  }
}
```

###### Output
```
0 a
1 b
Script Error: TypeError
```

2. When we add - remove - add element to set, then we visit last element _n_ times:

###### Code example
```js
var set = new Set();

set.add(0);

for (var x of set) {
  print(x);
  if(x === 0) {
    set.add(2);
    set.delete(2);
    set.add(3);
    set.delete(3);
    set.add(1);
  }
}
```

###### Output
```
0
1
1
1
```

rtakacs:
@rwalczyna: thank you for your notices. I've fixed the issues that the test cases shows.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -304,63 +304,45 @@ static void
...
-  while (prop_iter_cp != JMEM_CP_NULL)
-  {
-    ecma_property_header_t *prop_iter_p = ECMA_GET_NON_NULL_POINTER (ecma_property_header_t, prop_iter_cp);
-    JERRY_ASSERT (ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
+  uint8_t entry_size = ecma_op_container_entry_size (lit_id);
 
-    ecma_gc_mark_properties ((ecma_property_pair_t *) prop_iter_p);
+  for (uint32_t i = 0; i < ECMA_CONTAINER_ENTRY_COUNT (container_p); i+= entry_size)
+  {
+    ecma_value_t *entry_p = ECMA_CONTAINER_START (container_p) + i;
```
rerobika: The container start pointer can be precalculated before the for loop.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -987,6 +987,50 @@ typedef enum
...
   ECMA_CONTAINER_FLAGS_WEAK = (1 << 0) /** container object is weak */
 } ecma_container_flags_t;
+
+/**
+ * Description of map collection.
+ */
+typedef struct
+{
+  ecma_value_t key;
+  ecma_value_t value;
```
rerobika: missing struct member descriptions.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -35,6 +35,196 @@
...
+                              ecma_value_t key_arg, /**< key argument */
+                              lit_magic_string_id_t lit_id) /**< class id */
+{
+  JERRY_ASSERT (container_p != NULL);
+
+  uint8_t entry_size = ecma_op_container_entry_size (lit_id);
+
+  for (uint32_t i = 0; i < ECMA_CONTAINER_ENTRY_COUNT (container_p); i += entry_size)
+  {
+    ecma_value_t *entry_p = ECMA_CONTAINER_START (container_p) + i;
```
rerobika: ditto.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -35,6 +35,196 @@
...
+                                                                    map_object_p->u.class_prop.u.value);
+  lit_magic_string_id_t lit_id = map_object_p->u.class_prop.class_id;
+
+  ECMA_CONTAINER_SET_SIZE (container_p, 0);
+
+  uint8_t entry_size = ecma_op_container_entry_size (lit_id);
+
+  for (uint32_t i = 0; i < ECMA_CONTAINER_ENTRY_COUNT (container_p); i += entry_size)
+  {
+    ecma_value_t *entry_p = ECMA_CONTAINER_START (container_p) + i;
```
rerobika: ditto.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -853,123 +980,65 @@ ecma_op_container_iterator_next (ecma_value_t this_val, /**< this argument */
...
 
-    if (ecma_is_value_empty (ECMA_PROPERTY_VALUE_PTR (property_p)->value))
+    if (ecma_is_value_empty (*entry_p))
     {
-      if (i == props_p->item_count - 1)
+      if (i == (ECMA_CONTAINER_ENTRY_COUNT (container_p) - entry_size))
       {
-        ret_value = ecma_create_iter_result_object (ECMA_VALUE_UNDEFINED, ECMA_VALUE_TRUE);
+        ret_value = ecma_create_iter_result_object (ECMA_VALUE_UNDEFINED, ECMA_VALUE_TRUE);;
```
rerobika: double semicolon.

rtakacs:
@rerobika Thanks, I've updated the PR according to your comments.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -304,63 +304,47 @@ static void
...
 
-    if ((internal_ext_p->u.class_prop.extra_info & ECMA_CONTAINER_FLAGS_WEAK) == 0)
+    if (ecma_is_value_empty (*entry_p))
     {
-      ecma_property_pair_t *prop_pair_p = (ecma_property_pair_t *) prop_iter_p;
+      continue;
+    }
 
-      for (uint32_t i = 0; i < ECMA_PROPERTY_PAIR_ITEM_COUNT; i++)
+    if (lit_id == LIT_MAGIC_STRING_SET_UL || lit_id == LIT_MAGIC_STRING_MAP_UL)
```
dbatyai: I'm not a fan of repeatedly doing these checks inside a for loop. What do you think about having a separate loop for the different container types? It would result in slightly larger code, but it would be more efficient performance wise.

rtakacs: @dbatyai I've updated the PR according to your notices.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -295,73 +295,110 @@ ecma_gc_mark_promise_object (ecma_extended_object_t *ext_object_p) /**< extended
...
-  ecma_extended_object_t *internal_ext_p = (ecma_extended_object_t *) internal_obj_p;
+  ecma_collection_t *container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t,
+                                                                    map_object_p->u.class_prop.u.value);
+  ecma_value_t *start_p = ECMA_CONTAINER_START (container_p);
+  uint32_t entry_count = ECMA_CONTAINER_ENTRY_COUNT (container_p);
 
-  ecma_gc_set_object_visited (internal_obj_p);
+  for (uint32_t i = 0; i < entry_count; i+= ECMA_CONTAINER_PAIR_SIZE)
+  {
+    ecma_value_t *entry_p = start_p + i;
```
dbatyai: How about changing this to an `ecma_container_pair_t *`?

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -295,73 +295,110 @@ ecma_gc_mark_promise_object (ecma_extended_object_t *ext_object_p) /**< extended
...
+  ecma_collection_t *container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t,
+                                                                    map_object_p->u.class_prop.u.value);
+  ecma_value_t *start_p = ECMA_CONTAINER_START (container_p);
+  uint32_t entry_count = ECMA_CONTAINER_ENTRY_COUNT (container_p);
+
+  for (uint32_t i = 0; i < entry_count; i+= ECMA_CONTAINER_PAIR_SIZE)
   {
-    ecma_property_header_t *prop_iter_p = ECMA_GET_NON_NULL_POINTER (ecma_property_header_t, prop_iter_cp);
-    JERRY_ASSERT (ECMA_PROPERTY_IS_PROPERTY_PAIR (prop_iter_p));
+    ecma_value_t *entry_p = start_p + i;
```
dbatyai: Maybe `ecma_container_pair_t *` here as well.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -35,6 +35,325 @@
...
+    ((ecma_container_pair_t *) entry_p)->value = ecma_copy_value_if_not_object (value_arg);
+  }
+} /* ecma_op_internal_buffer_update */
+
+/**
+ * Delete element from the internal buffer.
+ */
+static void
+ecma_op_internal_buffer_delete (ecma_collection_t *container_p, /**< internal container pointer */
+                                ecma_value_t *entry_p, /**< key argument pointer */
```
dbatyai: This could be an `ecma_container_pair_t *` as well.

jerry-core/ecma/base/ecma-gc.c
rtakacs: Yes, that would be great! :)

rtakacs:
@dbatyai I've updated the PR.

PR #3601, opened 2020-03-05
cmake: drop-down profile selection 

wille-io:
Get profile files, put them in a list for a drop-down selection for easier and fool-proof profile selection.

JerryScript-DCO-1.0-Signed-off-by: Mike Wille mike@wille.io


dbatyai:
Hi @wille-io,

We support custom profile files outside of the builtin profiles, which can be specified with an absolute file path. Would it also be possible to input a custom path instead of selecting from the predefined values when using the gui?

Also, please rebase to the current master.

rerobika:
Closing to to inactivity.

PR #3602, opened 2020-03-05
Fix setting integrity level, when object has Symbol properties

rwalczyna:
Symbol type properties were not included when integrity level were changed. 

As it is in ES6, [SetIntegrityLevel](https://www.ecma-international.org/ecma-262/6.0/#sec-setintegritylevel) uses [OwnPropertyKeys](https://www.ecma-international.org/ecma-262/6.0/#sec-ordinary-object-internal-methods-and-internal-slots-ownpropertykeys) that should also include Symbol type properties.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-object.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-object.c
@@ -291,7 +291,12 @@ ecma_builtin_object_set_integrity_level (ecma_object_t *obj_p, /**< object */
 #endif /* ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
 
   /* 6. */
-  ecma_collection_t *props_p = ecma_op_object_get_property_names (obj_p, ECMA_LIST_CONVERT_FAST_ARRAYS);
+  uint32_t opts = ECMA_LIST_CONVERT_FAST_ARRAYS;
+#if ENABLED (JERRY_ES2015)
+  opts |= ECMA_LIST_SYMBOLS;
+#endif /* ENABLED (JERRY_ES2015) */
+
+  ecma_collection_t *props_p = ecma_op_object_get_property_names (obj_p, opts);
 
 #if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
   if (props_p == NULL)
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -291,7 +291,12 @@ ecma_builtin_object_set_integrity_level (ecma_object_t *obj_p, /**< object */
 #endif /* ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
 
   /* 6. */
+#if ENABLED (JERRY_ES2015)
+  ecma_collection_t *props_p = ecma_op_object_get_property_names (obj_p, ECMA_LIST_CONVERT_FAST_ARRAYS
```
rerobika: I'd prefer guarding only the options bits rather than the whole statement. Other looks good to me.

rwalczyna: Thanks for the review, I put _ECMA_LIST_SYMBOLS_ option to if statement as you suggested. I tried also with putting both options, but it didn't look good. What are the guidelines for situation like this? I couldn't find anything similar in existing code.

rerobika: How about this?

```c
uint32_t opts = convert_fast_array
#if es2015
 opts |= list_symbols
#endif
```
Now you can pass `opts` as a second argument to the property name listing function.

rwalczyna: I thought about this, but can we be sure, that compilator would optimize it? It's another variable and I wanted to avoid using it. But if it's okay, then I agree that it would be the best solution :)

rerobika: It's only a temporary variable with a constant immediate value ( in both cases) so it cannot cause any problem :)

rwalczyna: I checked binary executable files, and there is no change in size on release build. Only 16 additional bytes on debug build. So you are right - it's fine ;)

Issue #3568, opened 2020-02-19
Implement missing Math functions from ES6 standard


rwalczyna:
To conform compatibility with ES6 some of the Math functions need to be implemented.

List of missing features:

- [x] Math.acosh [ES v6, 20.2.2.3](https://www.ecma-international.org/ecma-262/6.0/#sec-math.acosh)
- [x] Math.asinh [ES v6, 20.2.2.5](https://www.ecma-international.org/ecma-262/6.0/#sec-math.asinh)
- [x] Math.atanh [ES v6, 20.2.2.7](https://www.ecma-international.org/ecma-262/6.0/#sec-math.atanh)
- [x] Math.cbrt [ES v6, 20.2.2.9](https://www.ecma-international.org/ecma-262/6.0/#sec-math.cbrt)
- [x] Math.clz32 [ES v6, 20.2.2.11](https://www.ecma-international.org/ecma-262/6.0/#sec-math.clz32)
- [x] Math.cosh [ES v6, 20.2.2.13](https://www.ecma-international.org/ecma-262/6.0/#sec-math.cosh)
- [x] Math.expm1 [ES v6, 20.2.2.15](https://www.ecma-international.org/ecma-262/6.0/#sec-math.expm1)
- [x] Math.fround [ES v6, 20.2.2.17](https://www.ecma-international.org/ecma-262/6.0/#sec-math.fround)
- [x] Math.hypot [ES v6, 20.2.2.18](https://www.ecma-international.org/ecma-262/6.0/#sec-math.hypot)
- [x] Math.imul [ES v6, 20.2.2.19](https://www.ecma-international.org/ecma-262/6.0/#sec-math.imul)
- [x] Math.log1p [ES v6, 20.2.2.21](https://www.ecma-international.org/ecma-262/6.0/#sec-math.log1p)
- [x] Math.log10 [ES v6, 20.2.2.22](https://www.ecma-international.org/ecma-262/6.0/#sec-math.log10)
- [x] Math.log2 [ES v6, 20.2.2.23](https://www.ecma-international.org/ecma-262/6.0/#sec-math.log2)
- [x] Math.sinh [ES v6, 20.2.2.31](https://www.ecma-international.org/ecma-262/6.0/#sec-math.sinh)
- [x] Math.tanh [ES v6, 20.2.2.34](https://www.ecma-international.org/ecma-262/6.0/#sec-math.tanh)

rwalczyna:
[...]

PR #3617, opened 2020-03-16
Implement missing Math logarithm functions from ES6

rwalczyna:
Math.log2, Math.log10, Math.log1p, Math.expm1

C implementation ported from fdlibm

Part of Issue #3568


tests/jerry/es2015/math-log1p.js
```diff
@@ -0,0 +1,25 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var p_zero = 0.0;
+var n_zero = -p_zero;
+
+assert(isNaN(Math.log1p(NaN)));
+assert(isNaN(Math.log1p(-42)));
+assert(isNaN(Math.log1p(-3.0)));
+assert(Math.log1p(n_zero) === n_zero);
```
ossy-szeged: ditto

tests/jerry/es2015/math-expm1.js
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var p_zero = 0.0;
+var n_zero = -p_zero;
+
+assert(isNaN(Math.expm1(NaN)));
+assert(Math.expm1(p_zero) === p_zero);
+assert(Math.expm1(n_zero) === n_zero);
```
ossy-szeged: Comparing to -0.0 won't work with == or ===, because 0.0 will be equal to -0.0.

Previously I found and fixed a similar bug, you can see 
the proper compare in tests/jerry/es2015/math-trunc.js .

rwalczyna: Thanks, you are right, it helped me to find bug in source code ;)

jerry-core/ecma/builtin-objects/ecma-builtin-math.c
```diff
@@ -570,38 +570,32 @@ ecma_builtin_math_dispatch_routine (uint16_t builtin_routine_id, /**< built-in w
...
-#ifdef JERRY_LIBM_MATH_H
-        return ecma_raise_type_error (ECMA_ERR_MSG ("UNIMPLEMENTED: Math.expm1"));
-#else /* !JERRY_LIBM_MATH_H */
-        x = DOUBLE_TO_ECMA_NUMBER_T (expm1 (x));
-#endif /* JERRY_LIBM_MATH_H */
+        /* ES6 20.2.2.15 Math.expm1(-0.0) = -0.0 */
+        if ((x == 0.0) && (ecma_number_is_negative (x)))
+        {
+          break;
+        }
```
ossy-szeged: This check should be inside #ifdef JERRY_LIBM_MATH_H guard,
because exp in libm conforms to the spec without this workaround.

jerry-core/ecma/builtin-objects/ecma-builtin-math.c
```diff
@@ -570,38 +570,32 @@ ecma_builtin_math_dispatch_routine (uint16_t builtin_routine_id, /**< built-in w
...
-#ifdef JERRY_LIBM_MATH_H
-        return ecma_raise_type_error (ECMA_ERR_MSG ("UNIMPLEMENTED: Math.log1p"));
-#else /* !JERRY_LIBM_MATH_H */
-        x = DOUBLE_TO_ECMA_NUMBER_T (log1p (x));
-#endif /* JERRY_LIBM_MATH_H */
+        /* ES6 20.2.2.21 Math.log1p(-0.0) = -0.0 */
+        if ((x == 0.0) && (ecma_number_is_negative (x)))
+        {
+          break;
+        }
```
ossy-szeged: ditto

jerry-libm/log2.c
```diff
@@ -0,0 +1,160 @@
...
+ * notice:
+ *
+ *     Copyright (C) 1993 by Sun Microsystems, Inc. All rights reserved.
+ *
+ *     Developed at SunSoft, a Sun Microsystems, Inc. business.
+ *     Permission to use, copy, modify, and distribute this
+ *     software is freely granted, provided that this notice
+ *     is preserved.
+ *
+ *     @(#)e_log2.c 1.3 95/01/18
```
ossy-szeged: Do we know the origin of this file? I can't see e_log2.c here: https://www.netlib.org/fdlibm/

jerry-core/ecma/builtin-objects/ecma-builtin-math.c
rwalczyna: exp(x) function conforms to specification, but for (0-) "left" and (0+) "right" "0" - but in this case it should distinguish "left" and "right" "1", because we are subtracting 1 from exp(x) result.

I deleted this check, build with _tools/build.py --profile=es2015-subset --debug --show-opcodes=ON --jerry-libm=OFF --link-lib=m --clean_ and unfortunately test fails.

jerry-libm/log2.c
rwalczyna: Source: https://github.com/JuliaMath/openlibm/blob/master/src/e_log2.c
It looks exactly the same as files on netlib.org, but I couldn't find original place with all fdlibm files - something like "SunSoft repository"

jerry-core/ecma/builtin-objects/ecma-builtin-math.c
ossy-szeged: Ah, I got it. In this case we should use expm1 and log1p instead of exp and log if we build with (GNU) libm, the new tests pass with them.

rwalczyna: I will add native implementation (not reuse of exp()) to provide better accuracy for small numbers (both for expm1 and log1p).

rwalczyna:
I added native implementations of expm1() and logp1().
This way the accuracy is better and there is no need to add extra checks to _ecma-builtin-math.c_

ossy-szeged:
LGTM (informal)

rwalczyna:
Thanks, I rebased the commit and fixed those lines.

PR #3621, opened 2020-03-17
Remove ECMA_OBJECT_TYPE_PSEUDO_ARRAY

rerobika:
All represented object types can be stored as ECMA_OBJECT_TYPE_CLASS.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -128,6 +128,54 @@ ecma_op_object_get_own_property (ecma_object_t *object_p, /**< the object */
           }
         }
       }
+#if ENABLED (JERRY_ES2015_BUILTIN_TYPEDARRAY)
+      /* ES2015 9.4.5.1 */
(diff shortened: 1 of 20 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1701,23 +1609,34 @@ ecma_op_object_define_own_property (ecma_object_t *obj_p, /**< the object */
...
 
-        if (ecma_compare_ecma_strings (property_name_p, num_to_str))
+        if (is_equal)
         {
-          ecma_deref_ecma_string (num_to_str);
-
           return ecma_reject (property_desc_p->flags & ECMA_PROP_IS_THROW);
         }
 
         ecma_deref_ecma_string (num_to_str);
```
galpeter: Do we need the deref here? There is already one above.

rerobika: Good catch! Fixed.

rerobika:
Closing the PR, I'll come up with an other approach for `ECMA_OBJECT_TYPE_PSEUDO_ARRAY`.

PR #3623, opened 2020-03-18
Implement Proxy object [[getPrototypeOf]] internal method

szilagyiadam:
The algorithm is based on ECMA-262 v6, 9.5.1


jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -292,8 +292,92 @@ ecma_value_t
...
+  /* 15. */
+  if (ECMA_IS_VALUE_ERROR (target_proto))
+  {
+    return target_proto;
+  }
+
+  ecma_value_t ret_value = handler_proto;
+
+  /* 16. */
+  if (!ecma_op_same_value (handler_proto, target_proto))
```
rerobika: Both handler and target proto only can be an object or null, therefore a simple `!=` comparison is enough here.

PR #3632, opened 2020-03-20
Fix VLA error in strings unittest

galpeter:
In the test-api-string.c there was an error if the unittests are built in release.
Full error message:
```
tests/unit-core/test-api-strings.c:89:20: error: variable-length array bound is unknown [-Werror=vla-larger-than=]
89 |   JERRY_VLA (char, string_from_cesu8_string, cesu8_sz);
```

The problem is that the compiler does not know if the buffer size is always greater than zero.
Added an extra check to the TEST_ASSERT a line above to resolve the error.


ossy-szeged:
LGTM (informal)

Issue #3637, opened 2020-03-24
SEGV on ecma_builtin_array_prototype_object_slice


juckchang:
###### JerryScript revision
`655110a` commit

###### Build steps
```sh
python tools/build.py --clean --compile-flag=-fsanitize=address
```

###### Test case
```
a = []
for(var i=0; i<100; i++) a.push(i)
a.slice(0, {valueOf: function(){a.length=0; return 100;}})
```
###### Output
```
==34465==ERROR: AddressSanitizer: SEGV on unknown address 0x000173d7b4e0 (pc 0x00010e47486f bp 0x7ffee179a130 sp 0x7ffee179a130 T0)
==34465==The signal is caused by a READ memory access.
    #0 0x10e47486e in ecma_ref_ecma_string (jerry:x86_64+0x10000f86e)
    #1 0x10e478e93 in ecma_copy_value (jerry:x86_64+0x100013e93)
    #2 0x10e47ef83 in ecma_builtin_array_prototype_dispatch_routine (jerry:x86_64+0x100019f83)
    #3 0x10e496a37 in ecma_builtin_dispatch_call (jerry:x86_64+0x100031a37)
    #4 0x10e4d9672 in vm_execute (jerry:x86_64+0x100074672)
    #5 0x10e4d910c in vm_run (jerry:x86_64+0x10007410c)
    #6 0x10e46a81f in jerry_run (jerry:x86_64+0x10000581f)
    #7 0x10e467e49 in main (jerry:x86_64+0x100002e49)
    #8 0x7fff72b5c7fc in start (libdyld.dylib:x86_64+0x1a7fc)

==34465==Register values:
rax = 0x000000002e7af69c  rbx = 0x0000000065737361  rcx = 0x000010002e7af600  rdx = 0x0000100000000000
rdi = 0x0000000173d7b4e0  rsi = 0x0000000000000000  rbp = 0x00007ffee179a130  rsp = 0x00007ffee179a130
 r8 = 0x0000100000000000   r9 = 0x00000000000006e8  r10 = 0x000000010e644670  r11 = 0x00007fff72cccf00
r12 = 0x000000010e64470c  r13 = 0x000000010e6441bc  r14 = 0x0000100000000000  r15 = 0x000000000000000f
AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV (jerry:x86_64+0x10000f86e) in ecma_ref_ecma_string
==34465==ABORTING
[...]

PR #3649, opened 2020-03-27
Fix Array.slice when using fast arrays

galpeter:
When a fast array was used during the Array.slice call and
the input array's properties were removed the property
list was used incorrectly.

Fixes: #3637

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -869,6 +869,15 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
 
     if (ext_from_obj_p->u.array.u.hole_count < ECMA_FAST_ARRAY_HOLE_ONE)
     {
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -869,6 +869,17 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
 
     if (ext_from_obj_p->u.array.u.hole_count < ECMA_FAST_ARRAY_HOLE_ONE)
     {
+      ecma_value_t *from_buffer_p = ECMA_GET_POINTER (ecma_value_t, obj_p->u1.property_list_cp);
+
+      if (JERRY_UNLIKELY (from_buffer_p == NULL))
```
rerobika: How about check whether the `obj_p->u1.property_list_cp == JMEM_CP_NULL`, then leave the  `ECMA_GET_NON_NULL_POINTER `?

galpeter: True, the `..GET_POINTER` does a check already, we can eliminate the double check.

galpeter:
@rerobika I've updated the PR

PR #3657, opened 2020-03-29
Implement Proxy object [[Construct]] internal method

szilagyiadam:
The algorithm is based on ECMA-262 v6, 9.5.14


jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -1224,8 +1224,59 @@ ecma_proxy_object_construct (ecma_object_t *obj_p, /**< proxy object */
...
+  /* 10 .*/
+  if (ECMA_IS_VALUE_ERROR (new_obj))
+  {
+    return new_obj;
+  }
+
+  /* 11. */
+  if (!ecma_is_value_object (new_obj))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Trap returned non-object"));
```
rerobika: `new_obj` should be released at this point. Also please add a tc for it.

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -1224,8 +1224,59 @@ ecma_proxy_object_construct (ecma_object_t *obj_p, /**< proxy object */
...
+  /* 9. */
+  ecma_value_t new_obj = ecma_op_function_call (func_obj_p, handler, function_call_args, 3);
+
+  ecma_free_value (arg_array);
+  ecma_deref_object (func_obj_p);
+
+  /* 10 .*/
+  if (ECMA_IS_VALUE_ERROR (new_obj))
+  {
+    return new_obj;
```
galpeter: It would be good to have a test where the construct handler throws an error.

PR #3659, opened 2020-04-01
Fix attributes of length property for bound function

### LINKED ISSUE #3615: Adding 'length' property to bound function
Hello,
I'm going to implement 'length' property of bound function.
I think that there are 2 approaches as belows.

1) Adding 'length' property when each bound function generated
This approach is simple but all bound function objects always hold 'length' property whether it is really used or not.

2) Lazily insert 'length' property
Adding 'length' property when it is accessed for the first time. This approach could save memory, but we should hold the length of target function when each bound function is generated. 
From the standard [es6 19.2.3.2](https://www.ecma-international.org/ecma-262/6.0/#sec-function.prototype.bind), 'length' property of bound function is calculated using the length of target function and its number of arguments.
number of arguments is already stored in `ecma_extended_object_t::u::bound_function::args_len_or_this`, but the length of target function is not.
```js
var foo = function (x,y) {};
var bound1 = foo.bind(null);
delete foo.length;
var bound2 = foo.bind(null);
print(bound1.length); // 2 (length property is defined when bound function is generated)
print(bound2.length); // 0
```
From the above case, we even cannot calculate the length property lazily because the length value of target function could be changed before the moment.
To resolve this issue, I'm considering storing the length value of target function into additional space outside the `ecma_extended_object_t` by allocating a one-slot bigger ecma_extended_object_t using `ecma_create_object()` when creating a bound function object.
Intialization of length property could be marked by pointer with tag (jmem_cpointer_tag_t) which is similar to the case of function object.

Which one do you prefer?

clover2123:
fix length property of bound function objects to be configurable (ECMA-262 v6, 19.2.4.1)
resolve #3615 

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.c
@@ -244,46 +244,48 @@ ecma_builtin_function_prototype_object_bind (ecma_object_t *this_arg_obj_p , /**
 #endif /* !ENABLED (JERRY_ES2015) */
 
   ecma_object_t *function_p;
-  ecma_extended_object_t *ext_function_p;
+  ecma_bound_function_t *bound_func_p;
 
   if (arguments_number == 0
       || (arguments_number == 1 && !ecma_is_value_integer_number (arguments_list_p[0])))
   {
     function_p = ecma_create_object (prototype_obj_p,
-                                     sizeof (ecma_extended_object_t),
+                                     sizeof (ecma_bound_function_t),
                                      ECMA_OBJECT_TYPE_BOUND_FUNCTION);
 
     /* 8. */
-    ext_function_p = (ecma_extended_object_t *) function_p;
-    ECMA_SET_INTERNAL_VALUE_POINTER (ext_function_p->u.bound_function.target_function,
-                                     this_arg_obj_p);
+    bound_func_p = (ecma_bound_function_t *) function_p;
+    ECMA_SET_NON_NULL_POINTER_TAG (bound_func_p->header.u.bound_function.target_function,
+                                   this_arg_obj_p,
+                                   0);
 
-    ext_function_p->u.bound_function.args_len_or_this = ECMA_VALUE_UNDEFINED;
+    bound_func_p->header.u.bound_function.args_len_or_this = ECMA_VALUE_UNDEFINED;
 
     if (arguments_number != 0)
     {
-      ext_function_p->u.bound_function.args_len_or_this = ecma_copy_value_if_not_object (arguments_list_p[0]);
+      bound_func_p->header.u.bound_function.args_len_or_this = ecma_copy_value_if_not_object (arguments_list_p[0]);
     }
   }
   else
   {
     JERRY_ASSERT (arguments_number > 0);
 
-    size_t obj_size = sizeof (ecma_extended_object_t) + (arguments_number * sizeof (ecma_value_t));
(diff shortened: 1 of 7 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.c
```diff
@@ -244,57 +244,76 @@ ecma_builtin_function_prototype_object_bind (ecma_object_t *this_arg_obj_p , /**
...
     ecma_value_t args_len_or_this = ecma_make_integer_value ((ecma_integer_value_t) arguments_number);
-    ext_function_p->u.bound_function.args_len_or_this = args_len_or_this;
+    bound_func_p->header.u.bound_function.args_len_or_this = args_len_or_this;
   }
 
 #if ENABLED (JERRY_ES2015)
+  ecma_integer_value_t len = 0;
+  ecma_value_t length_value = ecma_op_object_find_own (ecma_make_object_value (this_arg_obj_p),
+                                                       this_arg_obj_p,
+                                                       ecma_get_magic_string (LIT_MAGIC_STRING_LENGTH));
```
clover2123, line 303: `Function.prototype.bind` based on [es6 19.2.3.2](https://www.ecma-international.org/ecma-262/6.0/#sec-function.prototype.bind)
Currently, this patch gets the length property of target (this_arg_obj_p) when it is an ordinary object case only.
For the case of Proxy, can I use `ecma_proxy_object_get_own_property_descriptor` to get the length property?

zherczeg:
I would add the following ES6 test cases:

In this case changing the length of f affects the bound function:
```
function f(a,b,c) {}
Object.defineProperty(f, "length", { value: 30 });
var g = f.bind(1,2)
g.length // 29
```
In this case changing the length of f does not affect the bound function:
```
function f(a,b,c) {}
var g = f.bind(1,2)
Object.defineProperty(f, "length", { value: 30 });
g.length // 2
```

jerry-core/ecma/builtin-objects/ecma-builtin-function-prototype.c
rerobika: Good remark, currently [[GetOwnProperty]] and [[DefineOwnProperty]] are the remaining unimplemented Proxy features. However the functions are exists so feel free to use them, but they will throw unimplemented error whenever they called. Nevertheless it's a good opportunity to ensure that the error handling is correct.

clover2123: Thank you for the review.
I fixed it to use `ecma_op_object_get_own_property_descriptor` to handle Proxy and ordinary object both.

clover2123:
@zherczeg Thank you for the great suggestion. I fixed the test case following your comment.

jerry-core/ecma/operations/ecma-function-object.h
```diff
@@ -99,10 +99,10 @@ ecma_op_external_function_list_lazy_property_names (uint32_t opts,
                                                     ecma_collection_t *non_enum_collection_p);
 
 void
-ecma_op_bound_function_list_lazy_property_names (uint32_t opts,
+ecma_op_bound_function_list_lazy_property_names (ecma_object_t *object_p,
+                                                 uint32_t opts,
                                                  ecma_collection_t *main_collection_p,
                                                  ecma_collection_t *non_enum_collection_p);
-
```
zherczeg, line 105: I would not remove this line.

clover2123: Fixed it.

PR #3663, opened 2020-04-01
Fix the types of builtin prototype objects

dbatyai:
In ES2015 many builtin prototypes are no longer valid instances of their
respective classes. This change updates affected prototypes to be
regular objects as required.


rerobika:
Please rebase the PR to the latest master.

PR #3666, opened 2020-04-06
Update the isArray operation to handle proxy objects

szilagyiadam:
Added step 3. from ECMA-262 v6, 7.2.2

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
@@ -378,7 +378,7 @@ ecma_builtin_helper_array_concat_value (ecma_object_t *array_obj_p, /**< array *
 
   bool spread_object = is_spreadable == ECMA_VALUE_TRUE;
 #else /* !ENABLED (JERRY_ES2015) */
-  bool spread_object = ecma_is_value_array (value);
+  bool spread_object = ecma_is_value_true (ecma_is_value_array (value));
 #endif /* ENABLED (JERRY_ES2015) */
 
   if (spread_object)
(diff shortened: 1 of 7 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -378,7 +378,7 @@ ecma_builtin_helper_array_concat_value (ecma_object_t *array_obj_p, /**< array *
 
   bool spread_object = is_spreadable == ECMA_VALUE_TRUE;
 #else /* !ENABLED (JERRY_ES2015) */
-  bool spread_object = ecma_is_value_array (value);
+  bool spread_object = ecma_is_value_true (ecma_is_value_array (value));
```
rerobika, line 381: This may throws an error.

jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -683,7 +683,7 @@ ecma_op_array_species_create (ecma_object_t *original_array_p, /**< The object f
   ecma_value_t constructor = ECMA_VALUE_UNDEFINED;
   ecma_value_t original_array = ecma_make_object_value (original_array_p);
 
-  if (ecma_is_value_array (original_array))
+  if (ecma_is_value_true (ecma_is_value_array (original_array)))
```
rerobika, line 686: Ditto.

tests/jerry/es2015/array-isarray.js
```diff
@@ -0,0 +1,43 @@
...
+assert(Array.isArray(false) === false);
+assert(Array.isArray(new Uint8Array(32)) === false);
+assert(Array.isArray({ __proto__: Array.prototype }) === false);
+
+var revocable = Proxy.revocable ({}, {});
+var proxy = revocable.proxy;
+revocable.revoke();
+
+try {
+  Array.isArray(proxy);
```
galpeter, line 39: This only tests if the revoke was ok on a revocable Proxy. There should be also test where the `isArray` is tested before the `revoke` is invoked for a revocable Proxy with an array target.

PR #3667, opened 2020-04-06
Add --test262-es2015 option to run-tests.py

ossy-szeged:


clover2123:
AFAIK there is no distinct es2015 release in test262, so we should select one practical commit that includes es2015 related tests but not any es7+ features.
I'm wondering which version of test262 is chosen for es2015.

tools/runners/run-test-suite-test262.py
```diff
@@ -28,37 +29,86 @@ def get_platform_cmd_prefix():
...
+    if return_code:
+        print('Cloning test262 repository failed.')
+        return return_code
+
+    if args.es2015:
+        git_hash = 'fd44cd73dfbce0b515a2474b7cd505d6176a9eb5'
+    else:
+        git_hash = 'es5-tests'
+
+    proc = subprocess.Popen(['git', 'checkout', git_hash], cwd=args.test_dir)
```
galpeter, line 65: Can't we just use `subprocess.call` here?

tools/runners/run-test-suite-test262.py
```diff
@@ -28,37 +29,86 @@ def get_platform_cmd_prefix():
...
+    proc.wait()
+    if proc.returncode:
+        print('Cloning test262 repository failed - invalid git revision.')
+        return proc.returncode
+
+    if args.es2015:
+        shutil.copyfile(os.path.join('tests', 'test262-es6-excludelist.xml'),
+                        os.path.join(args.test_dir, 'excludelist.xml'))
+
+        proc = subprocess.Popen(['git', 'apply', os.path.join('..', 'test262-es6.patch')], cwd=args.test_dir)
```
galpeter, line 75: `subprocess.call` ?

ossy-szeged:
> AFAIK there is no distinct es2015 release in test262, so we should select one practical commit that includes es2015 related tests but not any es7+ features.
> I'm wondering which version of test262 is chosen for es2015.

@clover2123 Yes, you are correct there is no es2015 release in test262. Additionally there is no release since ES5.1, so we have to options. We should pick a good enough revision which covers almost the spec and doesn't contain newer tests. Or we should pick a newer revision with better ES2015 coverage, but we have to skip tests for ES2015+ specs. I think fd44cd7... (which was suggested by you, thanks for it) is good enough to reach our goals. (The next b517ca71... hash adds an ES8 feature)

Additionally my goal was not to modify test262 at all. Unfortunately it wasn't possible, because exclude list handling was buggy at this point, because if you added eg. foo.js to exclude list, you excluded all foo.js in the tree, it wasn't possible to specify which one you want to skip. Apart from this minor bugfix, we can use the runner in the test262 repository. But in the future we will need new runner, because it isn't maintained and was removed from tree in Aug 2016.

tools/runners/run-test-suite-test262.py
ossy-szeged: I wanted to use it, but unfortunately only Popen supports cwd argument.

tools/runners/run-test-suite-test262.py
ossy-szeged: ditto

tools/runners/run-test-suite-test262.py
galpeter: `call` supports the same arguments as the Popen.

ossy-szeged: Ah, good to know. I should have read the manual thoroughly.

PR #3675, opened 2020-04-14
Implement missing Math Inverse Hyperbolic functions

rwalczyna:
Math.acosh, Math.asinh, Math.acosh

C implementation ported from fdlibm

Part of Issue #3568


ossy-szeged:
LGTM (informal)

PR #3676, opened 2020-04-14
Update the Array.prototype.reverse method to support ES6 version

szilagyiadam:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -545,7 +545,6 @@ ecma_builtin_array_prototype_object_reverse (ecma_value_t this_arg, /**< this ar
                                              ecma_object_t *obj_p, /**< array object */
                                              uint32_t len) /**< array object's length */
 {
-  /* 4. */
   uint32_t middle = len / 2;
 
   if (ecma_op_object_is_fast_array (obj_p))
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -610,72 +589,112 @@ ecma_builtin_array_prototype_object_reverse (ecma_value_t this_arg, /**< this ar
...
+      goto str_clean_up;
     }
 #endif /* ENABLED (JERRY_ES2015_BUILTIN_PROXY) */
 
     bool lower_exist = ecma_is_value_true (has_lower);
     bool upper_exist = ecma_is_value_true (has_upper);
 
-    /* 6.h */
+    ecma_value_t lower_value = ECMA_VALUE_EMPTY;
+    ecma_value_t upper_value = ECMA_VALUE_EMPTY;
```
galpeter, line 600: It seems that ES2015 and ES5.1 differs in the order of things to do. Probably we should create two methods to processes the upper/lower values in es6 and in es51.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -610,72 +589,112 @@ ecma_builtin_array_prototype_object_reverse (ecma_value_t this_arg, /**< this ar
...
+
+#if ENABLED (JERRY_ES2015)
+    if (lower_exist)
```
galpeter, line 603: In such cases I think we should only guard the `if` part and leave the opening a closing brackets alone.

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -605,6 +589,18 @@ ecma_builtin_array_prototype_object_reverse (ecma_value_t this_arg, /**< this ar
...
 
+    bool lower_exist = ecma_is_value_true (has_lower);
+
+    if (lower_exist)
+    {
+      lower_value = ecma_op_object_get (obj_p, lower_str_p);
+
+      if (ECMA_IS_VALUE_ERROR (lower_value))
+      {
+        goto clean_up;
```
galpeter, line 600: In these error cases, should we return the error for the user? Do we have test cases for these?

szilagyiadam: These test cases have already been tested in the ES5 tests, this update just changes the order of methods, but the behaviour when one of these error occurs is the same.

PR #3681, opened 2020-04-15
Add value/key sign swap for Set.add and Map.set methods

szilagyiadam:
See also: ECMA-262 v6, 23.2.3.1 step 6 and 23.1.3.9 step 6

### DIFF
--- a/jerry-core/ecma/operations/ecma-container-object.c
+++ b/jerry-core/ecma/operations/ecma-container-object.c
@@ -722,6 +722,31 @@ ecma_op_container_set_weak (ecma_object_t *const key_p, /**< key object */
 } /* ecma_op_container_set_weak */
 
 /**
+ * Helper method for the Map.prototype.set and Set.prototype.add methods to swap the sign of the given value if needed
+ *
+ * See also:
+ *          ECMA-262 v6, 23.2.3.1 step 6
+ *          ECMA-262 v6, 23.1.3.9 step 6
+ *
+ * @return ecma value
+ */
+static ecma_value_t
+ecma_op_container_set_noramlize_zero (ecma_value_t this_arg) /*< this arg */
+{
+  if (ecma_is_value_number (this_arg))
+  {
+    ecma_number_t number_value = ecma_get_number_from_value (this_arg);
+
+    if (JERRY_UNLIKELY (ecma_number_is_zero (number_value) && ecma_number_is_negative (number_value)))
+    {
+      return ecma_make_integer_value (0);
+    }
+  }
+
+  return this_arg;
+} /* ecma_op_container_set_noramlize_zero */
+
+/**
  * The generic Map prototype object's 'set' and Set prototype object's 'add' routine
  *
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -755,6 +755,16 @@ ecma_op_container_set (ecma_value_t this_arg, /**< this argument */
 
   if (entry_p == NULL)
   {
+    if (ecma_is_value_number (key_arg))
```
rerobika, line 758: Please separate it into a helper function to reduce code duplication.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -721,6 +721,31 @@ ecma_op_container_set_weak (ecma_object_t *const key_p, /**< key object */
...
+static ecma_value_t
+ecma_op_container_set_swap_sign_helper (ecma_value_t this_arg) /*< this arg */
+{
+  if (ecma_is_value_number (this_arg))
+  {
+    ecma_number_t number_value = ecma_get_number_from_value (this_arg);
+
+    if (ecma_number_is_zero (number_value) && ecma_number_is_negative (number_value))
+    {
+      return ecma_make_number_value (ECMA_NUMBER_ZERO);
```
rerobika, line 742: This can be `ecma_make_integer_value (0)`.

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -721,6 +721,31 @@ ecma_op_container_set_weak (ecma_object_t *const key_p, /**< key object */
...
+ * Helper method for the Map.prototype.set and Set.prototype.add methods to swap the sign of the given value if needed
+ *
+ * See also:
+ *          ECMA-262 v6, 23.2.3.1 step 6
+ *          ECMA-262 v6, 23.1.3.9 step 6
+ *
+ * @return ecma value
+ */
+static ecma_value_t
+ecma_op_container_set_swap_sign_helper (ecma_value_t this_arg) /*< this arg */
```
galpeter, line 734: Name is a bit awkward as the implementation does not swap the sign, but just nozmalize-s negative zero tto positve zero.
Maybe `ecma_op_container_set_noramlize_zero` ?

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -721,6 +721,31 @@ ecma_op_container_set_weak (ecma_object_t *const key_p, /**< key object */
...
+{
+  if (ecma_is_value_number (this_arg))
+  {
+    ecma_number_t number_value = ecma_get_number_from_value (this_arg);
+
+    if (ecma_number_is_zero (number_value) && ecma_number_is_negative (number_value))
```
galpeter, line 740: We should make these an unlikely path as compared to the ecam_number_t range the negative zero is very unlikely to happen, especially multiple times in a map/set

PR #3684, opened 2020-04-17
Fix is_constructor of bound function

rwalczyna:
Bound function created from built-in function does not have a constructor

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -114,7 +114,17 @@ ecma_object_is_constructor (ecma_object_t *obj_p) /**< ecma object */
 {
   JERRY_ASSERT (!ecma_is_lexical_environment (obj_p));
 
-  const ecma_object_type_t type = ecma_get_object_type (obj_p);
+  ecma_object_type_t type = ecma_get_object_type (obj_p);
+
+  while (type == ECMA_OBJECT_TYPE_BOUND_FUNCTION)
+  {
+    ecma_bound_function_t *bound_func_p = (ecma_bound_function_t *) obj_p;
+
+    obj_p = ECMA_GET_NON_NULL_POINTER_FROM_POINTER_TAG (ecma_object_t,
+                                                        bound_func_p->header.u.bound_function.target_function);
+
+    type = ecma_get_object_type (obj_p);
+  }
 
 #if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
   if (ECMA_OBJECT_TYPE_IS_PROXY (type))
@@ -128,7 +138,7 @@ ecma_object_is_constructor (ecma_object_t *obj_p) /**< ecma object */
     return (!ecma_get_object_is_builtin (obj_p) || !ecma_builtin_function_is_routine (obj_p));
   }
 
-  return (type >= ECMA_OBJECT_TYPE_BOUND_FUNCTION);
+  return (type >= ECMA_OBJECT_TYPE_EXTERNAL_FUNCTION);
 } /* ecma_object_is_constructor */
 
 /**
--- a/tests/jerry/es2015/array-of.js
+++ b/tests/jerry/es2015/array-of.js
@@ -64,6 +64,12 @@ assert(hits === 1);
 assert(array6.length === 2);
 assert(array6[1] === 2);
 
+// Test with bounded builtin function
+var boundedBuiltinFn = Array.of.bind(Array);
+var array7 = Array.of.call(boundedBuiltinFn, boundedBuiltinFn);
+assert(array7.length === 1);
+assert(array7[0] === boundedBuiltinFn);
+
 // Test superficial features
 var desc = Object.getOwnPropertyDescriptor(Array, "of");
 assert(desc.configurable === true);


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -109,7 +109,7 @@ ecma_op_is_callable (ecma_value_t value) /**< ecma value */
  * @return true - if the given object is constructor;
  *         false - otherwise
  */
-inline bool JERRY_ATTR_ALWAYS_INLINE
+inline bool
```
zherczeg, line 112: Why did you removed this?

rwalczyna: **ecma-function-object.c:113:1**: error: inlining failed in call to always_inline ‘ecma_object_is_constructor’: function not considered for inlining

Function has recursive call to itself - i think that's why compilator failed to inline it. But I left first *inline* untouched - it still can be used as suggestion to compilator when building with optimization enabled.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -128,7 +128,17 @@ ecma_object_is_constructor (ecma_object_t *obj_p) /**< ecma object */
     return (!ecma_get_object_is_builtin (obj_p) || !ecma_builtin_function_is_routine (obj_p));
   }
 
-  return (type >= ECMA_OBJECT_TYPE_BOUND_FUNCTION);
+  if (type == ECMA_OBJECT_TYPE_BOUND_FUNCTION)
```
rerobika, line 131: Isn't this change ES2015 specific?

rwalczyna: I think that this code could also be executed on previous version, because there is native function **jerry_value_is_constructor**.

I built jerryscript with **tools/build.py --debug --show-opcodes=ON --clean** on **master** (1b01171a) branch and created simple C code:
```c
#include "jerry-core/include/jerryscript.h"

int
main (void)
{

  const jerry_char_t script1[] = "Math.sin;";
  const jerry_char_t script2[] = "Math.sin.bind(Math);";

  jerry_value_t parsed_code1 =
      jerry_parse (NULL, 0, script1, sizeof (script1) - 1, JERRY_PARSE_NO_OPTS);
  jerry_value_t parsed_code2 =
      jerry_parse (NULL, 0, script2, sizeof (script2) - 1, JERRY_PARSE_NO_OPTS);

  jerry_value_t script_value1 = jerry_run (parsed_code1);
  jerry_value_t script_value2 = jerry_run (parsed_code2);

  bool is_constructor1 = jerry_value_is_constructor (script_value1);
  bool is_constructor2 = jerry_value_is_constructor (script_value2);

  printf ("%s\n", is_constructor1 ? "true" : "false");
  printf ("%s\n", is_constructor2 ? "true" : "false");

  jerry_release_value (script_value1);
  jerry_release_value (script_value2);
  jerry_release_value (parsed_code1);
  jerry_release_value (parsed_code2);
  jerry_cleanup ();
  return 0;
}
```
Execution of this code show:
```
false
true
```

After this commit, it is:
```
false
false
```

rerobika: I see, I've checked the standards and you are right. Good catch!

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -128,7 +128,17 @@ ecma_object_is_constructor (ecma_object_t *obj_p) /**< ecma object */
...
+  {
+    ecma_bound_function_t *bound_func_p = (ecma_bound_function_t *) obj_p;
+
+    ecma_object_t *func_obj_p = ECMA_GET_NON_NULL_POINTER_FROM_POINTER_TAG (
+        ecma_object_t, bound_func_p->header.u.bound_function.target_function);
```
rerobika, line 136: This indentation is a bit strange.

How about this?

```c
ecma_object_t *func_obj_p;
func_obj_p = ECMA_GET_NON_NULL_POINTER_FROM_POINTER_TAG (ecma_object_t,
                                                         bound_func_p->header.u.bound_function.target_function);
```

jerry-core/ecma/operations/ecma-function-object.c
zherczeg: Ok, I see it now. What about using a while loop to skip bound functions? This way the recursion can be removed (saves stack).

jerry-core/ecma/operations/ecma-function-object.c
rwalczyna: It was proposed by clang-format.

Unfortunately the formatting you showed, does not pass vera check:
```
./jerry-core/ecma/operations/ecma-function-object.c:136: error: line is longer than 120 characters
```
It has 132 characters.

jerry-core/ecma/operations/ecma-function-object.c
rwalczyna: Done.

jerry-core/ecma/operations/ecma-function-object.c
rwalczyna: Done.

Issue #3396, opened 2019-12-01
Assertion ecma_find_named_property (frame_ctx_p->lex_env_p, name_p) == NULL in vm_init_loop


renatahodovan:
###### JerryScript revision
51efba40

###### Build platform
Linux-4.15.0-62-generic-x86_64-with-Ubuntu-18.04-bionic

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```

###### Test case
```javascript
function i($, b = (eval("var x"))) {
    function x() {}
    eval()
}
i()
```

###### Output
```text
ICE: Assertion 'ecma_find_named_property (frame_ctx_p->lex_env_p, name_p) == NULL' failed at jerryscript/jerry-core/vm/vm.c(vm_init_loop):996.
Error: ERR_FAILED_INTERNAL_ASSERTION
```

###### Backtrace
```text
bt
#0  0xf7fd5079 in __kernel_vsyscall ()
#1  0xf77fc832 in raise () from /lib/i386-linux-gnu/libc.so.6
#2  0xf77fdcc1 in abort () from /lib/i386-linux-gnu/libc.so.6
#3  0x5657b5c9 in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-port/default/default-fatal.c:30
#4  0x566458b6 in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x566458f7 in jerry_assert_fail (assertion=0x566d2e40 "ecma_find_named_property (frame_ctx_p->lex_env_p, name_p) == NULL", file=0x566d2c80 "jerryscript/jerry-core/vm/vm.c", function=0x566b5900 <__func__.6352.lto_priv.448> "vm_init_loop", line=996) at jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x565d108d in vm_init_loop (frame_ctx_p=0xffffc470) at jerryscript/jerry-core/vm/vm.c:996
[...]

PR #3686, opened 2020-04-20
Implement separate context for arguments.

zherczeg:
#3683 should land first

### DIFF
--- a/jerry-core/parser/js/js-parser.c
+++ b/jerry-core/parser/js/js-parser.c
@@ -918,6 +918,30 @@ parser_post_processing (parser_context_t *context_p) /**< context */
   ecma_value_t *literal_pool_p;
   uint8_t *dst_p;
 
+#if ENABLED (JERRY_ES2015)
+  if ((context_p->status_flags & (PARSER_IS_FUNCTION | PARSER_LEXICAL_BLOCK_NEEDED))
+      == (PARSER_IS_FUNCTION | PARSER_LEXICAL_BLOCK_NEEDED))
+  {
+    PARSER_MINUS_EQUAL_U16 (context_p->stack_depth, PARSER_BLOCK_CONTEXT_STACK_ALLOCATION);
+#ifndef JERRY_NDEBUG
+    PARSER_MINUS_EQUAL_U16 (context_p->context_stack_depth, PARSER_BLOCK_CONTEXT_STACK_ALLOCATION);
+#endif /* !JERRY_NDEBUG */
+
+    context_p->status_flags &= (uint32_t) ~PARSER_LEXICAL_BLOCK_NEEDED;
+
+    parser_emit_cbc (context_p, CBC_CONTEXT_END);
+
+    parser_branch_t branch;
+    parser_stack_pop (context_p, &branch, sizeof (parser_branch_t));
+    parser_set_branch_to_current_position (context_p, &branch);
+  }
+#endif /* ENABLED (JERRY_ES2015) */
+
+  JERRY_ASSERT (context_p->stack_depth == 0);
+#ifndef JERRY_NDEBUG
+  JERRY_ASSERT (context_p->context_stack_depth == 0);
+#endif /* !JERRY_NDEBUG */
+
   if ((size_t) context_p->stack_limit + (size_t) context_p->register_count > PARSER_MAXIMUM_STACK_LIMIT)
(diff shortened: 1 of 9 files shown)


jerry-core/parser/js/js-parser.c
```diff
@@ -1867,17 +1899,39 @@ parser_parse_function_arguments (parser_context_t *context_p, /**< context */
...
+  if (context_p->status_flags & PARSER_LEXICAL_BLOCK_NEEDED)
+  {
+    if ((context_p->next_scanner_info_p->u8_arg & SCANNER_FUNCTION_LEXICAL_ENV_NEEDED)
+        || scanner_is_function_context_needed (context_p))
+    {
+      context_p->status_flags |= PARSER_LEXICAL_ENV_NEEDED;
+
+      parser_branch_t branch;
+      parser_emit_cbc_forward_branch (context_p, CBC_BLOCK_CREATE_CONTEXT, &branch);
+      parser_stack_push (context_p, &branch, sizeof(parser_branch_t));
```
rerobika, line 1920: missing space before `(`;

jerry-core/parser/js/js-parser.c
```diff
@@ -918,6 +918,30 @@ parser_post_processing (parser_context_t *context_p) /**< context */
...
+#ifndef JERRY_NDEBUG
+    PARSER_MINUS_EQUAL_U16 (context_p->context_stack_depth, PARSER_BLOCK_CONTEXT_STACK_ALLOCATION);
+#endif /* !JERRY_NDEBUG */
+
+    context_p->status_flags &= (uint32_t) ~PARSER_LEXICAL_BLOCK_NEEDED;
+
+    parser_emit_cbc (context_p, CBC_CONTEXT_END);
+
+    parser_branch_t branch;
+    parser_stack_pop (context_p, &branch, sizeof(parser_branch_t));
```
rerobika, line 935: ditto.

jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1588,36 +1596,31 @@ scanner_is_context_needed (parser_context_t *context_p) /**< context */
 #else /* !ENABLED (JERRY_ES2015) */
     JERRY_ASSERT (type == SCANNER_STREAM_TYPE_VAR);
 #endif /* ENABLED (JERRY_ES2015) */
+#endif /* !JERRY_NDEBUG */
 
-    size_t length;
-
-    if (!(data_p[0] & SCANNER_STREAM_UINT16_DIFF))
+    if (!(data & SCANNER_STREAM_UINT16_DIFF))
```
galpeter, line 1601: Would it make sense to extract common parts? like:
```c
data_p += 2 + 1;
if (!(data & SCANNER_STREAM_UINT16_DIFF)))
{
  if (data_p[2] == 0)
  {
    data_p += sizeof (...);
  }
}
else
{
  data_p += 1;
}
```

Another question? Why "2" and "1" ? is there any info about these constants somewhere?

galpeter: I see this addition is also present in other parts. Maybe we could extact them into a single function to have them only at one place.

zherczeg: There is a long comment how the stream is compressed:
https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/parser/js/js-scanner.h#L114

I have been thinking about mergeng these functions, but I have doubts. They are similar, but they are also different. I don't know what is the best option here.

galpeter: What if we only create a function to calculate the required bytes to add? eg.:
```c
size_t scanner_stream_next_offset(uint8_t *data_p)
{
  size_t offset = 2 + 1;
  if (!(data_p[0] & SCANNER_STREAM_UINT16_DIFF)))
  {
    if (data_p[2] == 0)
    {
      offset += sizeof (...);
    }
  }
  else
  {
    offset+= 1;
  }
  return offset;
}
```

zherczeg:
Fixes #3396.

jerry-core/parser/js/js-scanner-util.c
zherczeg: I decided to do the merge. I does not seems that bad, and this is not a hot code.

PR #3689, opened 2020-04-21
Fix attributes of length property for builtin functions

clover2123:
fix length property of builtin functions to be configurable (ECMA-262 v6, 19.2.4.1)
conflicting test files are moved under es5.1 directory


zherczeg:
Several tests are moved to es5.1 here. It would be good if only the relevant part is moved to es5.1 and es2015 as  well. Or add an es5.1 tester function and check results based on its return value. An example (you can create something different if you want):

```
function is_es51() {
  return (typeof g == "function");
  { function g() {} }
}
```

clover2123:
@zherczeg Thanks for the review. I have added relevant tests to es5.1 and es2015 both.

zherczeg:
Thank you. Am I see correctly that you duplicated the whole tests? I think we should only duplicate the relevant parts, otherwise it will be harder to maintain the common parts.

clover2123:
@zherczeg 
Each tests/jerry/string-prototype-*.js test checks configurable attribute of 'length' property only at once, so I add es5.1/es2015 tester function to share the common parts.
But for 15.03.04.02-001.js and 15.03.04.02-004.js, I copied and fixed them according to es5.1 and es2015 both because they are composed of simple checks of 'length' property.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -650,17 +650,32 @@ ecma_builtin_try_to_instantiate_property (ecma_object_t *object_p, /**< object *
...
-       * Note:
-       *      We don't need to mark that the property was already lazy instantiated,
-       *      as it is non-configurable and so can't be deleted
        */
 
       ecma_property_t *len_prop_p;
+#if ENABLED (JERRY_ES2015)
+      ecma_extended_object_t *ext_func_p = (ecma_extended_object_t *) object_p;
+      uint32_t *bitset_p = ext_func_p->u.built_in.instantiated_bitset;
+      if (*bitset_p & (1u << 0))
```
zherczeg, line 659: This is kind of confusing for me: how do we know that the 0th bit represents the length?

clover2123: `instantiated_bitset` member is defined but never used for builtin functions.
(It is used to mark each instantiated property of builtin objects such as `Array`, but currently builtin functions also share this structure although never use it)
So, I arbitrarily use the last bit as instantication marker for 'length' property.
I also have a plan to revise `instantiated_bitset` for builtin functions in following patches.

PR #3694, opened 2020-04-23
Handle properties of builtin function separately from builtin objects

clover2123:
separate builtin function part to make code clear and to support 'name' property later

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -807,7 +807,15 @@ typedef struct
   uint8_t length_and_bitset_size; /**< length for built-in functions and
                                    *   bit set size for all built-ins */
   uint16_t routine_id; /**< routine id for built-in functions */
-  uint32_t instantiated_bitset[1]; /**< bit set for instantiated properties */
+  union
+  {
+    uint32_t instantiated_bitset[1]; /**< bit set for instantiated properties */
+    struct
+    {
+      uint16_t name; /**< name of the built-in functions */
+      uint16_t bitset; /**< bit set for instantiated properties of builtin functions */
+    } builtin_routine;
+  } u;
 } ecma_built_in_props_t;
 
 /**
--- a/jerry-core/ecma/builtin-objects/ecma-builtins.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtins.c
@@ -415,13 +415,13 @@ ecma_instantiate_builtin (ecma_builtin_id_t obj_builtin_id) /**< built-in id */
 
   built_in_props_p->id = (uint8_t) obj_builtin_id;
   built_in_props_p->routine_id = (uint16_t) obj_builtin_id;
-  built_in_props_p->instantiated_bitset[0] = 0;
+  built_in_props_p->u.instantiated_bitset[0] = 0;
 
   if (property_count > 32)
   {
     built_in_props_p->length_and_bitset_size = 1 << ECMA_BUILT_IN_BITSET_SHIFT;
 
-    uint32_t *instantiated_bitset_p = built_in_props_p->instantiated_bitset;
+    uint32_t *instantiated_bitset_p = built_in_props_p->u.instantiated_bitset;
     instantiated_bitset_p[1] = 0;
     instantiated_bitset_p[2] = 0;
   }
@@ -557,6 +557,7 @@ static ecma_object_t *
 ecma_builtin_make_function_object_for_routine (ecma_builtin_id_t builtin_id, /**< identifier of built-in object */
                                                uint16_t routine_id, /**< builtin-wide identifier of the built-in
                                                                      *   object's routine property */
+                                               uint16_t name_id, /**< magic string id of 'name' property */
                                                uint8_t length_prop_value) /**< value of 'length' property */
 {
   JERRY_ASSERT (length_prop_value < (1 << ECMA_BUILT_IN_BITSET_SHIFT));
@@ -564,17 +565,6 @@ ecma_builtin_make_function_object_for_routine (ecma_builtin_id_t builtin_id, /**
   ecma_object_t *prototype_obj_p = ecma_builtin_get (ECMA_BUILTIN_ID_FUNCTION_PROTOTYPE);
 
   size_t ext_object_size = sizeof (ecma_extended_object_t);
-
-  size_t property_count = ecma_builtin_get_property_count (builtin_id);
-
-  if (property_count > 32)
-  {
-    /* Only 64 extra properties supported at the moment.
-     * This can be extended to 256 later. */
-    JERRY_ASSERT (property_count <= (32 + 64));
-
-    ext_object_size += sizeof (uint32_t) * 2;
-  }
 
   ecma_object_t *func_obj_p = ecma_create_object (prototype_obj_p,
                                                   ext_object_size,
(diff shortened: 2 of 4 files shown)


zherczeg:
Would it be possible to create a `const uint16_t` array with the names for each built-in function? This way there would be no need to store them in memory.

clover2123:
@zherczeg 
I'm not sure that I correctly understand your comment,
but you mean that allocating one global array for storing names of builtin-functions, right?
That way would use more memory definitely, is it ok?

zherczeg:
It is just a question. If we create a const array, it will goes to the read-only data, so it increase the binary size, but not the read/write memory consumption. Since each function has a unique built-in id, we can get the name of the function from this const array. However, if no built-in routine use more than 16 property, your patch could be good as well.

rerobika:
Hi @clover2123!

I'm planning to start the implementation of the function name property of script functions. To eliminate simultaneous work on this task, please let me know if you have already started it. :)

clover2123:
@zherczeg AFAIK every name of built-in routines could be represented by 16bit lit_magic_string_id_t.

@rerobika I'm going to implement 'name' property one by one starting from builtin function to normal function. But my work process would be of course slower than yours (for name of normal function, parser needs to store function name during the parsing. But I'm not familiar with jerryscript parser, so I have to dig into it now).
There is no problem if you take this item. I'll find another item after this patch.

clover2123:
This patch is still work-in-progress.
I'll update `ecma_built_in_function_props_t` to separate builtin function module from 'ecma_built_in_props_t` completely.

zherczeg:
@clover2123 yes, the name can be represented as a 16 bit integer. My question is, whether the instantiation of the actual properties (length, etc.) can be fit into 16 bit for all of these routines? If yes, then splitting the 32 bit value into two 16 bit values is ok.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -882,8 +881,10 @@ ecma_builtin_try_to_instantiate_property (ecma_object_t *object_p, /**< object *
       is_accessor = true;
       uint16_t getter_id = ECMA_ACCESSOR_READ_WRITE_GET_GETTER_ID (curr_property_p->value);
       uint16_t setter_id = ECMA_ACCESSOR_READ_WRITE_GET_SETTER_ID (curr_property_p->value);
-      getter_p = ecma_builtin_make_function_object_for_getter_accessor (builtin_id, getter_id);
-      setter_p = ecma_builtin_make_function_object_for_setter_accessor (builtin_id, setter_id);
+      getter_p = ecma_builtin_make_function_object_for_getter_accessor (builtin_id,
+                                                                        getter_id, curr_property_p->magic_string_id);
```
zherczeg, line 885: These arguments should go into separate lines.

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -882,8 +881,10 @@ ecma_builtin_try_to_instantiate_property (ecma_object_t *object_p, /**< object *
...
+      setter_p = ecma_builtin_make_function_object_for_setter_accessor (builtin_id,
+                                                                        setter_id, curr_property_p->magic_string_id);
```
zherczeg, line 887: Ditto.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2058,10 +2079,20 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
...
+        if (type == ECMA_OBJECT_TYPE_FUNCTION && ecma_builtin_function_is_routine (obj_p))
+        {
+          ecma_builtin_function_list_lazy_property_names (obj_p,
+                                                          opts,
+                                                          prop_names_p,
+                                                          skipped_non_enumerable_p);
+        }
+        else
+        {
+          ecma_builtin_list_lazy_property_names (obj_p,
```
zherczeg, line 2091: What about keeping this single function, and do this check at the beginnig of the function:
`if (type == ECMA_OBJECT_TYPE_FUNCTION && ecma_builtin_function_is_routine (obj_p))`

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -807,7 +807,15 @@ typedef struct
...
   uint16_t routine_id; /**< routine id for built-in functions */
-  uint32_t instantiated_bitset[1]; /**< bit set for instantiated properties */
+  union
+  {
+    uint32_t instantiated_bitset[1]; /**< bit set for instantiated properties */
+    struct
+    {
+      uint16_t name; /**< name of the built-in functions */
+      uint16_t bitset; /**< bit set for instantiated properties of builtin functions */
+    } builtin_function;
```
zherczeg, line 817: I think the name is misleading. This struct is specialized for routines, not for built-in functions (e.g `Date`).

rerobika:
Hi @clover2123!

I'm interested about the status of the PR, since there was no activity in the last two weeks. The patch itself look good to me after the minor changes requested by @zherczeg , and we would like to land it soon. Do you have capacity to do the requested changes? If you don't, I can do the finals steps as well if you don't mind.

clover2123:
@rerobika 
Sorry for late update.
I'll complete this patch soon.

jerry-core/ecma/operations/ecma-objects.c
clover2123: IMHO builtin routine is a function property of builtin object (e.g. Math.abs), so it should be handled independently from builtin objects.
So, I added `ecma_builtin_routine_list_lazy_property_names` for builtin routines to clearly separate each builtin routine from builtin object.
This separation also follows other use case in which `ecma_op_function_list_lazy_property_names` and `ecma_op_function_try_to_lazy_instantiate_property` exist for function object.

clover2123:
@zherczeg @rerobika 
Thanks for your review. I've updated this patch according to your comments.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -799,17 +799,29 @@ typedef struct
...
+ * Description of built-in routine property of an built-in ECMA-object.
+ */
+typedef struct
+{
+  uint8_t built_in_object_id; /**< built-in object id */
+  uint8_t length; /**< length of built-in routine */
+  uint16_t id; /**< id of built-in routine */
+  uint16_t name; /**< name of built-in routine */
+  uint16_t bitset; /**< bit set for instantiated properties of builtin routine */
+} ecma_built_in_routine_props_t;
```
zherczeg, line 823: I am not sure what was the discussion here, but `ecma_builtin_function_is_routine` silently expects that `id` is mapped to the same position as `routine_id`. I usually prefer unions, since the common fileds are ensured to be in the same position. What do you think?

clover2123: It's my mistake.
Fixed it as you commented. Thanks:)

rerobika:
@clover2123 We are planning to prepare for a release before long and the function name support for builtins/builtin routines are highly important for us. I think this PR was only a preparation for it and according to your previous contributions you are interested about it. Would you like to submit a PR for finalize this topic or let us/me to do the remaining work?

clover2123:
@rerobika 
Yes, I'd like to complete `name` property implementation for builtins. But would you first let me know when a new release is scheduled? If the schedule is too tight for me, it would be better that you do the remaining work.

rerobika:
@clover2123 We are planning in feature freeze in the end of this week please consider your available time for this task according to it and please let me know your decision. :)

clover2123:
@rerobika 
Oh.. sadly I don't have enough time to focus on this issue right now.
Would you please handle it?

PR #3698, opened 2020-04-28
Accessor functions do not have [[Construct]] internal method

szilagyiadam:


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1217,6 +1217,11 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
     return ecma_raise_type_error (ECMA_ERR_MSG ("Arrow functions have no constructor."));
   }
 
+  if (byte_code_p->status_flags & CBC_CODE_FLAGS_ACCESSOR)
```
zherczeg, line 1220: Could we combine these two checks to reduce runtime cost? Pseudocode:
```
if (status_flags & (FLAG1 | FLAG2)) {
  if (status_flags & FLAG1)
   raise_error(A)
  raise_error(B)
}
```

PR #3700, opened 2020-04-29
Fix function declaration issues inside catch blocks.

zherczeg:


jerry-core/parser/js/js-scanner-util.c
```diff
@@ -564,14 +564,23 @@ scanner_pop_literal_pool (parser_context_t *context_p, /**< context */
       {
         no_declarations++;
 
-#if !ENABLED (JERRY_ES2015)
+#if ENABLED (JERRY_ES2015)
+        if ((type & (SCANNER_LITERAL_IS_CONST | SCANNER_LITERAL_IS_ARG)) == SCANNER_LITERAL_IS_CONST)
```
zherczeg, line 568: If a reviewer insists, the two code paths can be merged (using a `condition` boolean), especially because the assert is true in the other case as well. It feels ugly for me, but I will not object.

jerry-core/parser/js/js-scanner-util.c
```diff
@@ -617,12 +626,14 @@ scanner_pop_literal_pool (parser_context_t *context_p, /**< context */
 #if ENABLED (JERRY_ES2015)
       extended_type |= SCANNER_LITERAL_IS_USED;
 
+      const uint8_t mask = (SCANNER_LITERAL_IS_ARG | SCANNER_LITERAL_IS_LOCAL);
+
       if ((type & SCANNER_LITERAL_IS_ARG)
-          || (!(literal_location_p->type & SCANNER_LITERAL_IS_ARG)
-              && (literal_location_p->type & SCANNER_LITERAL_IS_LOCAL)))
+          || (literal_location_p->type & mask) == SCANNER_LITERAL_IS_LET
```
rerobika, line 632: Cannot we use 
` || (literal_location_p->type & mask) == (SCANNER_LITERAL_IS_LET | SCANNER_LITERAL_IS_CONST)` here?

zherczeg: Absolutely no. That is a case we want to exclude here, not include. And the two other cases will be excluded, which we want to include :) But I can change the mask to SCANNER_LITERAL_IS_LOCAL, since `(type & SCANNER_LITERAL_IS_ARG)` handles ensures that SCANNER_LITERAL_IS_ARG is not set.

zherczeg: I was wrong, those types belong to different literals.
The summary is, we have 3 bits, and from the 8 combinations, we reject these two. I don't think we can simplify it.

rerobika: Sure, thanks for the explanation.

PR #3703, opened 2020-04-29
Fix JSON.parse when __proto__ property is used

rwalczyna:
JSON.parse should treat **\_\_proto__** as a normal property

###### Test Code
```js
var str = '{"__proto__": [] }';
var obj = JSON.parse(str);

print(JSON.stringify(obj));
print(Array.isArray(obj.__proto__));
```

###### Current output - before commit
```
{"__proto__":[]}
false
```
###### Expected Output
```
{"__proto__":[]}
true
```

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -98,7 +98,11 @@ vm_op_get_value (ecma_value_t object, /**< base object */
       {
         ecma_object_t *obj_p = ecma_get_object_from_value (object);
 
-        return ecma_builtin_object_object_get_prototype_of (obj_p);
+        ecma_value_t has_property = ecma_op_object_has_property (obj_p, property_name_p);
+        if (ecma_is_value_false (has_property))
+        {
+          return ecma_builtin_object_object_get_prototype_of (obj_p);
+        }
       }
 #endif /* ENABLED (JERRY_ES2015) */
     }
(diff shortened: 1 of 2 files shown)


jerry-core/vm/vm.c
```diff
@@ -98,7 +98,11 @@ vm_op_get_value (ecma_value_t object, /**< base object */
       {
         ecma_object_t *obj_p = ecma_get_object_from_value (object);
 
-        return ecma_builtin_object_object_get_prototype_of (obj_p);
+        ecma_value_t has_property = ecma_op_object_has_property (obj_p, property_name_p);
+        if (ecma_is_value_false (has_property))
+        {
+          return ecma_builtin_object_object_get_prototype_of (obj_p);
+        }
```
zherczeg, line 105: According to ES2015, this thing should be a getter/setter:
http://www.ecma-international.org/ecma-262/6.0/#sec-object.prototype.__proto__

It shouldn't be in the vm at all. If we do this correctly, the issue should go away, since objects will have no such own property.

rwalczyna: I checked before what happen if I remove __proto__ handling from vm and I got some issues that I could not solve. But I checked it again right now and it looks like it would solve the problem as you said. I will recheck one more time and make changes to this PR.

rerobika: @rwalczyna As upon this issue has been revealed I stardted to lookup the reason of this error. I've already found it and I've already have a fix for it (also includes the removal of the __proto__ handling in vm_get). I'd submit a PR with the fix and I'd mark you as a co-author if it is suitable for you.

rerobika: Related PR https://github.com/jerryscript-project/jerryscript/pull/3704.

rwalczyna: @rerobika Sounds good, so I think this PR can be closed.

PR #3705, opened 2020-04-30
Fix function argument handling issues caused by the parser.

zherczeg:
1) Nested function declarations should not overwrite arguments.
2) Functions should be created in the correct scope.

### DIFF
--- a/jerry-core/parser/js/js-scanner-util.c
+++ b/jerry-core/parser/js/js-scanner-util.c
@@ -142,17 +142,19 @@ scanner_get_stream_size (scanner_info_t *info_p, /**< scanner info block */
       case SCANNER_STREAM_TYPE_LET:
       case SCANNER_STREAM_TYPE_CONST:
       case SCANNER_STREAM_TYPE_LOCAL:
-      case SCANNER_STREAM_TYPE_DESTRUCTURED_ARG:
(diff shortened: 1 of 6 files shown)


jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1688,10 +1690,17 @@ scanner_is_context_needed (parser_context_t *context_p, /**< context */
...
     {
-      continue;
+      if (SCANNER_STREAM_TYPE_IS_ARG (type))
+      {
+        continue;
+      }
+
+      if (SCANNER_STREAM_TYPE_IS_ARG_FUNC (type))
+      {
+        return true;
```
zherczeg, line 1702: This case is triggered, if a function has an argument, which is initialized with a function or eval, and there is another argumnt, which has the same name as a function statement in the function scope. While this check could be more sophisticated, I think it is not worth it unless omebody wants it.

rerobika: I'd prefer this comment in the code as well.

zherczeg: Good idea. I have been able to make the check a bit better. Now the corner case (returning with true when unnecessary) is extremely rare.

PR #3716, opened 2020-05-05
Fix prototype of NativeError objects

rwalczyna:
Internal property [[Prototype]] of EvalError, RangeError, ReferenceError,
SyntaxError, TypeError and URIError should be Error.

[ES 5.1 15.11.7.7](http://www.ecma-international.org/ecma-262/5.1/#sec-15.11.7.7)


rwalczyna:
I think that this feature is only for [ES6](http://www.ecma-international.org/ecma-262/6.0/#sec-properties-of-the-nativeerror-constructors), while in [ES5.1](https://www.ecma-international.org/ecma-262/5.1/index.html#sec-15.11.7.5) it should remain Function object. I will fix the commit in a while.

PR #3717, opened 2020-05-05
Add simple stack measure for jerry-test

galpeter:
This change introduces simple stack measuring by coloring the stack before doing the actual JS work.
The dynamic linking should be avoided as it will use the stack during the symbol resolving.

At this moment the `JERRY_STACK_MEASURE` compile time define will enable the measurement for the jerry-test binary.

A usable build configuration:

```
$ ./tools/build.py --jerry-cmdline-test=on --compile-flag=-DJERRY_STACK_MEASURE=1 \
  --compile-flag=-static
```

### DIFF
--- a/jerry-main/CMakeLists.txt
+++ b/jerry-main/CMakeLists.txt
@@ -17,9 +17,11 @@ project (jerry-main C)
 
 # Optional build settings
 set(ENABLE_LINK_MAP OFF CACHE BOOL "Enable generating a link map file?")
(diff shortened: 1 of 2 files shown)


zherczeg:
This looks very low level. Do we want this in trunk?

rerobika:
@zherczeg Thank's an optional feature that mostly fits all *nix architectures. I see no disadvantage of having this feature.

dbatyai:
It is low level, however I don't see that as a concern. The benefit on the other hand would be that it would make it easier to maintain and use in the long run, instead of having to patch it in each time we want to measure stack usage.

Issue #3711, opened 2020-05-04
heap-buffer-overflow in ecma_collection_append


renatahodovan:
###### JerryScript revision
6ff299c8

###### Build platform
Linux-4.15.0-88-generic-x86_64-with-Ubuntu-18.04-bionic

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```

###### Test case
```javascript
function f() {}
if (function() {
        return f.bind()(0, 0, 0, 0, 0, 0, 0)
    }());
```

###### Output
```text
=================================================================
==37810==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xf530077c at pc 0xf798990e bp 0xff919678 sp 0xff919248
WRITE of size 28 at 0xf530077c thread T0
    #0 0xf798990d  (/usr/lib32/libasan.so.4+0x7790d)
    #1 0x567782c2 in ecma_collection_append jerryscript/jerry-core/ecma/base/ecma-helpers-collection.c:182
    #2 0x5671e7fa in ecma_op_function_call_bound jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1006
    #3 0x5671ebd3 in ecma_op_function_call jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1071
    #4 0x566e5e9c in opfunc_call.lto_priv.453 jerryscript/jerry-core/vm/vm.c:773
    #5 0x566a7581 in vm_execute jerryscript/jerry-core/vm/vm.c:4095
    #6 0x566a7af9 in vm_run jerryscript/jerry-core/vm/vm.c:4197
    #7 0x5671e20f in ecma_op_function_call_simple jerryscript/jerry-core/ecma/operations/ecma-function-object.c:864
    #8 0x5671eb9b in ecma_op_function_call jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1063
    #9 0x566e5e9c in opfunc_call.lto_priv.453 jerryscript/jerry-core/vm/vm.c:773
    #10 0x566a7581 in vm_execute jerryscript/jerry-core/vm/vm.c:4095
[...]

PR #3719, opened 2020-05-05
Fix appending elements for ecma collection

galpeter:
During ecma_collection_append the underlying collection
was not increased in the required case. This triggered
a buffer overflow when processing the bound function's arguments
during call or during the Proxy ownKeys method.

Fixes: #3711

### DIFF
--- a/tests/jerry/regression-test-issue-3711.js
+++ b/tests/jerry/regression-test-issue-3711.js
@@ -1,0 +1,42 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tests/jerry/regression-test-issue-3711.js
```diff
@@ -0,0 +1,43 @@
...
+
+function func() {}
+
+var bound = func.bind();
+
+/* original test case from the issue report */
+if (function() {
+  return func.bind()(0, 0, 0, 0, 0, 0, 0)
+}());
+
```
zherczeg, line 23: One newline.

galpeter: @zherczeg I've updated the PR

PR #3723, opened 2020-05-06
Add new target support to RegExp object

szilagyiadam:


jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -75,6 +75,7 @@ ecma_builtin_regexp_dispatch_helper (const ecma_value_t *arguments_list_p, /**<
 
   bool pattern_is_regexp = regexp_value == ECMA_VALUE_TRUE;
   re_compiled_code_t *bc_p = NULL;
+  new_target_p = JERRY_CONTEXT (current_new_target);
```
rerobika, line 78: The declaration also can happen here.

PR #3726, opened 2020-05-07
Update ES6 Regexp accessors descriptors

rwalczyna:
In ES6 accessors of built in object are by default configurable. [ES 6 21.2.6.1](http://www.ecma-international.org/ecma-262/6.0/#sec-lastindex)
Also lastIndex property of RegExp object is created when only used. [ES 6 17](http://www.ecma-international.org/ecma-262/6.0/#sec-ecmascript-standard-built-in-objects)


rwalczyna:
I updated the commit. I changed also descriptors of `flags` accessor.

PR #3730, opened 2020-05-11
Fix apply receiver operation for Proxy objects

rerobika:

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -1194,6 +1194,16 @@ ecma_op_object_put_apply_receiver (ecma_value_t receiver, /**< receiver */
     return result;
   }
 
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1194,6 +1194,16 @@ ecma_op_object_put_apply_receiver (ecma_value_t receiver, /**< receiver */
     return result;
   }
 
+#if ENABLED (JERRY_ES2015_BUILTIN_PROXY)
```
zherczeg, line 1197: The comment says: `[[Set]] ( P, V, Receiver) operation part for ordinary objects` Proxy is not ordinary, isn't it?

rerobika: I think the comment is still valid. The `ordinary object` concern for the argument `O` not the `Receiver`, which means the operation can be called in a normal object, with a Proxy receiver.

PR #3734, opened 2020-05-13
Fix position calculations of String.startsWith

mmatyas:
Fixes two bugs in the index calculations that affect the result of the `String.startsWith` function:

- Fixes the implementation of ECMA-262 v6, 21.1.3.18, step 14: *"If searchLength+start is greater than len, return false"*.
- Fixes the return value of the helper function `ecma_builtin_helper_string_find_index`. If it is called with a starting search position, the returned index should be not less than that.

These changes fix the following test cases in the test262 suite:

- built-ins/String/prototype/startsWith/return-true-if-searchstring-is-empty.js
- built-ins/String/prototype/startsWith/searchstring-found-with-position.js


mmatyas:
Ok, updated the patch with some test cases that fail on the current master but fixed by this PR.

PR #3736, opened 2020-05-13
Add support of escaping backslash and backtick to String.raw

rwalczyna:
Template literals already supports escaping backslash and backtick, this commit enables escaping this in String.raw.

It also fixes invalid behaviour of String.raw when using new line character - escaped new line characters were not included in string.

There was also application crash when escaping backtick in String.raw argument.

### DIFF
--- a/jerry-core/parser/js/js-lexer.c
+++ b/jerry-core/parser/js/js-lexer.c
@@ -974,6 +974,9 @@ lexer_parse_string (parser_context_t *context_p, /**< context */
       else if (*source_p == LEXER_NEWLINE_LS_PS_BYTE_1 && LEXER_NEWLINE_LS_PS_BYTE_23 (source_p))
       {
         source_p += 3;
+#if ENABLED (JERRY_ES2015)
+        length += 3 * raw_length_inc;
+#endif /* ENABLED (JERRY_ES2015) */
         line++;
         column = 1;
         continue;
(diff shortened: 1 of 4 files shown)


jerry-core/parser/js/js-lexer.c
```diff
@@ -974,6 +974,9 @@ lexer_parse_string (parser_context_t *context_p, /**< context */
       else if (*source_p == LEXER_NEWLINE_LS_PS_BYTE_1 && LEXER_NEWLINE_LS_PS_BYTE_23 (source_p))
       {
         source_p += 3;
+#if ENABLED (JERRY_ES2015)
+        length += 3 * raw_length_inc;
```
zherczeg, line 978: Do we need this `raw_length_inc`, or can the raw length be computed as `source_p - string_start_p`. If yes, perhaps we could just remove the variable. This could be more future proof approach.

rerobika: IMO that way is not completely valid, since in raw strings there are character sequences like (\r\n) which going to be transformed (into \n) therefore a single subtraction is not enough.

rwalczyna: Yes, we could remove `raw_length_inc`, but we still would need another variable (which also has to be guarded with __JERRY_ES2015__) to check for __\r\n__ to __\n__ conversions, as __rerobika__ said. Also, `source_p` points to one character after literal ends when there is __$__ in string. So length (for raw string) would be `source_p - string_start_p - raw_length_dec`

@zherczeg I can add it to this commit, there is not so much to change, if you think this would improve performance or code readability :)

PR #3745, opened 2020-05-15
Implement function name support for script functions and classes

rerobika:


rerobika:
The patch is finally done, and ready for the the review.

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -2823,6 +2951,24 @@ parser_pattern_form_assignment (parser_context_t *context_p, /**< context */
     parser_emit_cbc_ext_forward_branch (context_p, CBC_EXT_DEFAULT_INITIALIZER, &skip_init);
 
     parser_parse_expression (context_p, PARSE_EXPR_NO_COMMA);
+
+    if (name_index != UINT16_MAX)
+    {
+      uint16_t function_literal_index = parser_check_anonymous_function_declaration (context_p);
+
+      if (function_literal_index < PARSER_NAMED_FUNCTION)
```
galpeter, line 2959: Would it make sense to reorganize the ifs to something like this?:
```c
if (function_literal_index == PARSER_ANONYMOUS_CLASS)
{
}
else if (function_literal_index < PARSER_NAMED_FUNCTION)
{
}
```

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -571,6 +574,21 @@ parser_parse_var_statement (parser_context_t *context_p) /**< context */
...
+          if (function_literal_index == PARSER_ANONYMOUS_CLASS)
+          {
+            lexer_construct_literal_object (context_p, &name_loc, LEXER_STRING_LITERAL);
+            parser_emit_cbc_ext_literal_from_token (context_p, CBC_EXT_SET_FUNCTION_NAME);
+          }
+          else
+          {
+            parser_set_function_name (context_p, function_literal_index, index, 0);
+          }
+        }
```
galpeter, line 590: This looks identical to the one in the `parser_pattern_form_assignment` method. Would it be possible to create a function for it?

jerry-core/api/jerry-snapshot.c
```diff
@@ -520,6 +550,47 @@ jerry_snapshot_set_offsets (uint32_t *buffer_p, /**< buffer */
...
+
+          literal_start_p[-1] = current_p->literal_offset;
+        }
+
+        literal_start_p--;
+      }
+#endif /* ENABLED (JERRY_ES2015) */
+
+#if ENABLED (JERRY_RESOURCE_NAME)
+      int32_t resource_name_index = (bytecode_p->status_flags & CBC_CODE_FLAG_HAS_TAGGED_LITERALS) ? -2 : -1;
```
galpeter, line 576: I was wondering if we could move this `? :` part to a macro as it is used at a few places, but I'm not sure if it is a good idea.

jerry-core/vm/vm.c
```diff
@@ -1829,6 +1829,64 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+          ecma_object_t *func_obj_p = ecma_get_object_from_value (left_value);
+          ecma_property_value_t *value_p;
+
+          if (ecma_find_named_property (func_obj_p, ecma_get_magic_string (LIT_MAGIC_STRING_NAME)) != NULL)
+          {
+            ecma_free_value (right_value);
+            continue;
+          }
+
+          value_p = ecma_create_named_data_property (ecma_get_object_from_value (left_value),
```
galpeter, line 1876: Can't we use the `func_obj_p` as first arg here?

jerry-core/vm/vm.c
```diff
@@ -1829,6 +1829,64 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+            left_value = stack_top_p[-1];
+          }
+          else
+          {
+            right_value = left_value;
+            left_value = stack_top_p[-1];
+          }
+
+          ecma_object_t *func_obj_p = ecma_get_object_from_value (left_value);
+          ecma_property_value_t *value_p;
```
galpeter, line 1868: This can be moved after the `if`.

jerry-core/vm/vm.c
```diff
@@ -1829,6 +1829,64 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+
+          if (ecma_find_named_property (func_obj_p, ecma_get_magic_string (LIT_MAGIC_STRING_NAME)) != NULL)
```
galpeter, line 1870: How likely is this case? If is likely maybe we could move the whole check at the top of the case block, and user the `stack_top_p[-1]` directly as the first arg?

jerry-core/parser/js/js-parser-expr.c
rerobika: Good idea.

jerry-core/parser/js/js-parser-statm.c
rerobika: Not exactly. At his point the function name for anonymous classes can be constructed from the saved literal location, but in `parser_pattern_form_assignment ` we use `lexer_construct_string_literal_from_ident`.

jerry-core/api/jerry-snapshot.c
rerobika: I'd keep them like this for now, I'm planning to improve the snapshot tool later so maybe this could be improved as well.

jerry-core/vm/vm.c
rerobika: Unfortunately the current order cannot be changed since the ToPropName call may trigger an accessor which sets a name property to the object.

rerobika:
@galpeter Thanks for the review, the PR is updated.

jerry-core/parser/js/js-lexer.c
```diff
@@ -2329,6 +2329,64 @@ lexer_convert_literal_to_chars (parser_context_t *context_p, /**< context */
   return destination_start_p;
 } /* lexer_convert_literal_to_chars */
 
+/**
+ * Construct a literal object from an identifier.
+ */
+void
+lexer_construct_string_literal_from_ident (parser_context_t *context_p, /**< context */
```
zherczeg, line 2336: Instead of constructing a new literal, it should return with an ecma value.

jerry-core/parser/js/js-parser.c
```diff
@@ -2599,6 +2597,124 @@ parser_parse_arrow_function (parser_context_t *context_p, /**< context */
...
+
+  lexer_literal_t *name_lit_p = (lexer_literal_t *) PARSER_GET_LITERAL (name_index);
+
+  uint8_t *name_buffer_p = (uint8_t *) name_lit_p->u.char_p;
+  uint32_t name_length = name_lit_p->prop.length;
+
+  if (status_flags & (PARSER_IS_PROPERTY_GETTER | PARSER_IS_PROPERTY_SETTER))
+  {
+    name_length += 4;
+    name_buffer_p = (uint8_t *) jmem_heap_alloc_block (name_length * sizeof (uint8_t));
```
zherczeg, line 2704: Parser has its own allocator.

tests/unit-core/test-snapshot.c
```diff
@@ -195,79 +195,13 @@ static void test_exec_snapshot (uint32_t *snapshot_p, size_t snapshot_size, uint
 int
 main (void)
 {
-  static uint32_t snapshot_buffer[SNAPSHOT_BUFFER_SIZE];
-
   TEST_INIT ();
 
-  /* Dump / execute snapshot */
```
zherczeg, line 202: What about adding an assertion to the current value of CBC_END, CBC_EXT_END, and a comment before that if this change, the snapshot version should be increased.

rerobika: I resolved this problem via adding static asserts for them.

rerobika:
@zherczeg Thanks for the review the PR is updated.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1417,6 +1464,34 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
...
+  {
+    ecma_extended_object_t *ext_func_p = (ecma_extended_object_t *) object_p;
+    if (!ECMA_GET_SECOND_BIT_FROM_POINTER_TAG (ext_func_p->u.function.scope_cp))
+    {
+      /* Set tag bit to represent initialized 'name' property */
+      ECMA_SET_SECOND_BIT_TO_POINTER_TAG (ext_func_p->u.function.scope_cp);
+      const ecma_compiled_code_t *bytecode_data_p = ecma_op_function_get_compiled_code (ext_func_p);
+
+      ecma_value_t value = *ecma_compiled_code_resolve_function_name (bytecode_data_p);
+      if (value != ECMA_VALUE_EMPTY)
```
zherczeg, line 1478: Isn't there a special constant for this?

jerry-core/parser/js/byte-code.c
```diff
@@ -21,6 +21,15 @@ JERRY_STATIC_ASSERT ((sizeof (cbc_uint8_arguments_t) % sizeof (jmem_cpointer_t))
...
                      sizeof_cbc_uint16_arguments_t_must_be_divisible_by_sizeof_jmem_cpointer_t);
 
+/**
+ * The reason of these two static asserts to notify the developer to increase the JERRY_SNAPSHOT_VERSION
+ * whenever new bytecodes are introduced or existing ones have been deleted.
+ */
+JERRY_STATIC_ASSERT (CBC_END == 238,
+                     number_of_cbc_opcodes_must_be_238);
+JERRY_STATIC_ASSERT (CBC_EXT_END == 115,
+                     number_of_cbc_opcodes_must_be_115);
```
zherczeg, line 31: I would rename this to `number_of_cbc_opcodes_changed` and `number_of_cbc_ext_opcodes_changed` so less number of lines needs to be updated. (Btw why the second is `number_of_cbc_opcodes_must_be_115` and not `number_of_cbc_ext_opcodes_must_be_115`)

jerry-core/ecma/operations/ecma-function-object.c
rerobika: I've changed it to ECMA_VALUE_EMPTY, that helped me to reduce the EMPTY/ANONYMOUS checks in snapshot the snapshot tool.

jerry-core/parser/js/byte-code.c
rerobika: It was a typo, fine!

PR #3746, opened 2020-05-15
Rework RegExp engine and add support for proper unicode matching

dbatyai:
This change includes several bugfixes, general improvements, and support
for additional features.
- Added full support for web compatibility syntax defined in Annex B
- Implemented parsing and matching patterns in unicode mode
- Fixed capture results when iterating with nested capturing groups
- Significantly reduced regexp bytecode size
- Reduced stack usage during regexp execution
- Improved matching performance


dbatyai:
Benchmark | Heap (bytes) | Stack (bytes) | Perf (sec) |
regexp-dna.js | 1307536 -> 1307368 : +0.013% | 1648 -> 1600 : **+2.913%** | 1.379 -> 1.262 : **+8.469%** |
string-tagcloud.js | 1001432 -> 993352 : +0.807% | 2200 -> 2144 : **+2.545%** | 2.265 -> 2.203 : **+2.751%** |
string-unpack-code.js | 645344 -> 645288 : +0.009% | 1944 -> 1896 : **+2.469%** | 0.977 -> 0.919 : **+5.987%** |
string-validate-input.js | 927816 -> 927688 : +0.014% | 1824 -> 1768 : **+3.070%** | 8.243 -> 7.392 : **+10.331%** |
regexp.js | 2140568 -> 2128008 : +0.587% | 3916 -> 3364 : **+14.096%** | 5.770 -> 4.411 : **+23.545%** |

ES5.1 binary size | master | patch | Diff |
.text | 121656 | 121984 | +328 bytes |
.rodata | 14015 | 13999 | -16 bytes |


ES6 binary size | master | patch | Diff |
.text | 191544 | 192424 | +880 bytes |
.rodata | 19289 | 19273 | -16 bytes |

lauriro:
`\.` after non-capturing group and word boundary does not work
```
print("name.lower()".replace(/(?:a|b)\b|\.\w+/g, ""))
// jerry returns: name.lower()
// node returns: name()
```

dbatyai:
Thanks @lauriro, should be fixed now.

lauriro:
one more

```
var re = /([[])\w+(?:=(("|')(?:\\?.)*?\3))?]?/g
console.log("a[href='#a>b']".replace(re, ""))
// jerry returns: a='#a>b']
// node returns: a
```

strange thing is, when `[` is escaped in first character set, then it seems to work.
simpler re `/([[])/` without escape also works, some combination will break.

dbatyai:
Thanks, fixed as well.

jerry-core/ecma/operations/ecma-regexp-object.h
```diff
@@ -82,27 +124,38 @@ typedef struct
...
   const lit_utf8_byte_t *input_start_p;        /**< start of input string */
+  const lit_utf8_byte_t *input_end_p;          /**< end of input string */
   uint32_t captures_count;                     /**< number of capture groups */
-  ecma_regexp_capture_t *captures_p;           /**< capturing groups */
   uint32_t non_captures_count;                 /**< number of non-capture groups */
+  ecma_regexp_capture_t *captures_p;           /**< capturing groups */
   ecma_regexp_non_capture_t *non_captures_p;   /**< non-capturing groups */
-  uint32_t *iterations_p;                      /**< number of iterations */
   uint16_t flags;                              /**< RegExp flags */
+  uint8_t char_size;
```
rerobika, line 147: missing desc.

jerry-core/parser/regexp/re-bytecode.c
```diff
@@ -29,135 +30,103 @@
...
 } /* re_initialize_regexp_bytecode */
 
-/**
- * Realloc the bytecode container
- *
- * @return current position in RegExp bytecode
- */
-static uint8_t *
-re_realloc_regexp_bytecode_block (re_bytecode_ctx_t *bc_ctx_p) /**< RegExp bytecode context */
+uint32_t
```
rerobika, line 41: How about always inline?

jerry-core/parser/regexp/re-bytecode.c
```diff
@@ -167,318 +136,497 @@ re_get_char (const uint8_t **bc_p) /**< pointer to bytecode start */
...
 /**
- * Append a RegExp opcode
+ * Get the encoded size of an uint32_t value.
+ *
+ * @return encoded value size
  */
-void
-re_append_opcode (re_bytecode_ctx_t *bc_ctx_p, /**< RegExp bytecode context */
-                  const re_opcode_t opcode) /**< input opcode */
+static size_t
```
rerobika, line 199: Ditto?

jerry-core/parser/regexp/re-compiler-context.h
```diff
@@ -0,0 +1,60 @@
...
+
+  uint8_t *bytecode_start_p;      /**< start of bytecode block */
+  size_t bytecode_size;           /**< size of bytecode */
+
+  uint32_t captures_count;           /**< number of capture groups */
+  uint32_t non_captures_count;       /**< number of non-capture groups */
+
+  int groups_count;                     /**< number of groups */
+  uint16_t flags;                       /**< RegExp flags */
+  re_token_t token;             /**< current token */
```
rerobika, line 50: please align the descriptions

jerry-core/parser/regexp/re-parser.c
```diff
@@ -35,559 +35,760 @@
...
+  }
+  else
+  {
+    JERRY_ASSERT (idx == capture_start);
+  }
+
+  re_insert_value (re_ctx_p, group_start_offset, idx);
+  re_insert_opcode (re_ctx_p, group_start_offset, re_get_group_start_opcode (is_capturing));
+
+  re_append_opcode (re_ctx_p, re_get_group_end_opcode (re_ctx_p, is_capturing));
```
rerobika, line 115: How about reserving space before these 4 appends?

dbatyai: I don't think it would improve much, this isn't a performance critical code path. Precalculating the required size would also increase complexity due to how values are encoded.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -508,31 +464,159 @@ ecma_regexp_canonicalize_char (lit_code_point_t ch) /**< character */
...
+#endif /* ENABLED (JERRY_ES2015) */
+
+/**
+ * Helper function to revert the string pointer to the previous code point.
+ *
+ * @return pointer to previous code point
+ */
+static JERRY_ATTR_NOINLINE const lit_utf8_byte_t *
+ecma_regexp_step_back (ecma_regexp_ctx_t *re_ctx_p,
+                       const lit_utf8_byte_t *str_p) /**< reference to string pointer */
```
zherczeg, line 589: When this function is exactly used? Word boundary?

If there is a surrogate pair, and match starts from the position between the two characters, `ecma_regexp_step_back` may go back the start of the surrogate pair instead of stopping at the middle.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -541,725 +625,950 @@ ecma_regexp_match (ecma_regexp_ctx_t *re_ctx_p, /**< RegExp matcher context */
...
-#if ENABLED (JERRY_ES2015)
-        if (re_ctx_p->flags & RE_FLAG_UNICODE
-            && lit_is_code_point_utf16_high_surrogate (ch2)
-            && str_curr_p < re_ctx_p->input_end_p)
+        const uint32_t qmin = re_get_value (&bc_p);
+        group_p->end_p = NULL;
+
+        /* If zero iterations are allowed, then execute the end opcode which will handle further iterations,
+         * otherwise run the 1st iteration immediately by executing group bytecode. */
+        if (qmin == 0)
```
zherczeg, line 689: Is this likely? Can it be handled at compile time in some way?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -541,725 +625,950 @@ ecma_regexp_match (ecma_regexp_ctx_t *re_ctx_p, /**< RegExp matcher context */
...
+
+          goto fail;
         }
 
-        if (!(re_ctx_p->flags & RE_FLAG_MULTILINE))
         {
-          JERRY_TRACE_MSG ("fail\n");
-          return NULL; /* fail */
+          /* Save and clear all nested capturing groups, and try to iterate. */
+          JERRY_VLA (const lit_utf8_byte_t *, saved_captures_p, group_p->subcapture_count);
```
zherczeg, line 807: Is this efficient?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -541,725 +625,950 @@ ecma_regexp_match (ecma_regexp_ctx_t *re_ctx_p, /**< RegExp matcher context */
...
+          if (matched_p != NULL)
+          {
+            return matched_p;
+          }
+
+          group_p->iterator--;
+          goto fail;
+        }
+
+        /* Empty matches are not allowed after reaching the minumum number of iterations. */
```
zherczeg, line 872: minimum

jerry-core/ecma/operations/ecma-regexp-object.h
```diff
@@ -44,30 +44,72 @@ typedef enum
...
+ */
+#define RE_CLASS_ESCAPE_COUNT_MASK ((1 << RE_CLASS_ESCAPE_COUNT_MASK_SIZE) - 1u)
 
 /**
- * Structure for storing non-capturing group results
+ * Character class flags
+ */
+typedef enum
+{
+  RE_CLASS_HAS_CHARS = (1 << 5),    /**< contains individual characters */
```
zherczeg, line 76: Why 5? Would be good to have a comment about it.

jerry-core/lit/lit-char-helpers.c
```diff
@@ -458,23 +484,51 @@ lit_read_code_unit_from_hex (const lit_utf8_byte_t *buf_p, /**< buffer with char
...
+  JERRY_ASSERT (lit_char_is_decimal_digit (*current_p));
+
+  uint32_t value = (uint32_t) (*current_p++ - LIT_CHAR_0);
+
+  while (current_p < buffer_end_p && lit_char_is_decimal_digit (*current_p))
+  {
+    const uint32_t digit = (uint32_t) (*current_p++ - LIT_CHAR_0);
+    uint32_t new_value = value * 10 + digit;
+
+    if (JERRY_UNLIKELY (value > UINT32_MAX / 10) || JERRY_UNLIKELY (new_value < value))
```
zherczeg, line 519: Is this comes from the standard or an engine limit?

jerry-core/ecma/operations/ecma-regexp-object.c
dbatyai: It is used in the greedy atom iterator during backtracking.

In unicode mode surrogate pairs are required to be handled as a single code point, matching can never start from the middle of a surrogate pair, and it is required to read the entire surrogate pair when backtracking.

jerry-core/ecma/operations/ecma-regexp-object.c
dbatyai: The likelihood would be roughly 50/50 in my opinion, it could be handled compile time, however that would increase code size, and performance impact would be negligible.

jerry-core/ecma/operations/ecma-regexp-object.c
dbatyai: It's required.

jerry-core/lit/lit-char-helpers.c
dbatyai: Engine limit, the value would overflow otherwise.

dbatyai:
Technically 2 values would be enough since prev_end always equals to current_start, but only in very simple cases. When there are nested capturing groups in the current group, it's no longer enough to save only these values, and we need to save/clear all the nested captures at the start of each iteration.

Take for example this regex: `\((a)|(b))+\.exec("ab")`, the resulting captures would be `0:"ab", 1:"ab", 2: undefined, 3:"b"`, since in the final iteration the `(a)` group did not match. If we do not clear the nested captures, the second capture would still contain `"a"`. Similarly, there are cases where we need to backtrack, and restore the captures of all nested groups after a failed iteration, which is why they also need to be saved.

We currently get away with only saving the starting pointer of the captures by not clobbering the end pointer unnecessarily, so in my opinion this is as efficient as it can get, while still conforming to the standard.

jerry-core/ecma/operations/ecma-regexp-object.c
zherczeg: It seems you are right, The engine overwrites the start offset for surrogate pairs. Is there a test for this?

zherczeg:
It looks like this is working differently from other engines, but this type of regexp has far less features so they can get away with it. If a regexp has 1000 captures inside an iterator, can it consume the stack easily?

By the way, would it be enough to use a capture reset bitset? So instead of storing the previous values (two uint32_t values), just store one bit which says the value is currently undefined (even though the capture vector contains its previous value). This can be a represented by a flag, or a negative number in the capture vector. When the engine enter to a group, just save the flags, and flag all numbers in the ovector. When restore the previous iteration, reset the flags only, not the values. Even using a byte vector is a lot of save compared to 2*4 bytes.

dbatyai:
I don't think a reset bitset would be enough. ``((x)|(.))*def`.exec("abcdef")` In this case the iterator would consume the whole string, the last capture would be 'f', and then we would backtrack three times until the tail 'def' matches, and in each step we have to restore the previous matched capture.

Currently when there are no nested captures we are using less stack than the previous implementation. For every nested capture there is, we have to save one extra value per iteration, which is unfortunately a necessary evil. We could save these on the heap, but that would be horrible performance wise.

If nested captures don't have quantifiers then stack usage is still reasonable, and if they do, that would be bad stack-wise regardless of whether we save captures or not. Also IMO nested captures aren't extremely common, there are usually one or two per regexp at most, so I don't consider them to be major concern.

rerobika:
> It looks like this is working differently from other engines, but this type of regexp has far less features so they can get away with it. If a regexp has 1000 captures inside an iterator, can it consume the stack easily?
> 
> By the way, would it be enough to use a capture reset bitset? So instead of storing the previous values (two uint32_t values), just store one bit which says the value is currently undefined (even though the capture vector contains its previous value). This can be a represented by a flag, or a negative number in the capture vector. When the engine enter to a group, just save the flags, and flag all numbers in the ovector. When restore the previous iteration, reset the flags only, not the values. Even using a byte vector is a lot of save compared to 2*4 bytes.

I think these ideas are **beyond** the PR's scope. The rework currently has significant stack/memory consumption improvement however the main reason was to add the proper unicode matching support for the engine so that's a double benefit. So I'd suggest to keep these ideas as the institution for opening a follow up optimization PR.

jerry-core/ecma/operations/ecma-regexp-object.c
dbatyai: There is one now.

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -508,31 +464,159 @@ ecma_regexp_canonicalize_char (lit_code_point_t ch) /**< character */
...
+} /* ecma_regexp_unicode_advance */
+#endif /* ENABLED (JERRY_ES2015) */
+
+/**
+ * Helper function to revert the string pointer to the previous code point.
+ *
+ * @return pointer to previous code point
+ */
+static JERRY_ATTR_NOINLINE const lit_utf8_byte_t *
+ecma_regexp_step_back (ecma_regexp_ctx_t *re_ctx_p,
```
zherczeg, line 588: Missing description.

rerobika:
Just to mention for the credits:
During the development of the PR the [Fuzzinator ](https://fuzzinator.readthedocs.io/en/latest/)found **10** unique issues with using [grammarinator](https://github.com/renatahodovan/grammarinator).

PR #3747, opened 2020-05-15
Fix rounding in Number.prototype.toFixed

### LINKED ISSUE #3741: Number#toFixed rounds differently from Math.round
```javascript
var n = 123.5
print(n.toFixed(0) + " " + Math.round(n))
```

jerryscript returns: 123 124
nodejs returns: 124 124

> "Arch Linux" (Linux 5.5.9-arch1-2 x86_64)
> commit: 95aa82763501d0bf1592d121fc0e8af5066ee726
> build command: python tools/build.py

dbatyai:
Fixes #3741.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
@@ -673,6 +673,7 @@ ecma_builtin_number_prototype_object_to_fixed (ecma_number_t this_num, /**< this
   if (!ecma_number_is_zero (this_num))
   {
     num_digits = ecma_number_to_binary_floating_point_number (this_num, digits, &exponent);
+    JERRY_ASSERT (exponent >= 0);
   }
   else
   {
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -690,9 +690,11 @@ ecma_builtin_number_prototype_object_to_fixed (ecma_number_t this_num, /**< this
     return ecma_builtin_number_prototype_object_to_string (this_num, NULL, 0);
   }
 
+  const lit_utf8_size_t total_digit_count = (lit_utf8_size_t) ((int32_t) num_digits + exponent);
```
zherczeg, line 693: I have a question: if exponent can be < 0, what ensures it num_digits < exponent. If not, why it is not casted to lit_utf8_size_t and do an unsigned addition? Also, why do we need a local variable for this?

dbatyai: Exponent is always non-negative, patch has been updated.

Issue #3760, opened 2020-05-18
heap-use-after-free  in ecma_regexp_split_helper


renatahodovan:
###### JerryScript revision
e01cfda0

###### Build platform
Linux-4.15.0-88-generic-x86_64-with-Ubuntu-18.04-bionic

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```

###### Test case
```javascript
var split = RegExp.prototype[Symbol.split];
split.call({[Symbol.match]: "g"})
```

###### Output
```text
=================================================================
==45542==ERROR: AddressSanitizer: heap-use-after-free on address 0xf5302230 at pc 0x56799ad5 bp 0xffe5d388 sp 0xffe5d378
READ of size 4 at 0xf5302230 thread T0
    #0 0x56799ad4 in ecma_deref_ecma_string jerryscript/jerry-core/ecma/base/ecma-helpers-string.c:806
    #1 0x5671159e in ecma_regexp_split_helper jerryscript/jerry-core/ecma/operations/ecma-regexp-object.c:1815
    #2 0x566a8afc in ecma_builtin_regexp_prototype_symbol_split jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c:680
    #3 0x566a6fd5 in ecma_builtin_regexp_prototype_dispatch_routine jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.inc.h:60
    #4 0x5675e5bb in ecma_builtin_dispatch_routine jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.c:1095
    #5 0x5675e793 in ecma_builtin_dispatch_call jerryscript/jerry-core/ecma/builtin-objects/ecma-builtins.c:1119
    #6 0x5673723a in ecma_op_function_call_simple jerryscript/jerry-core/ecma/operations/ecma-function-object.c:782
    #7 0x5673810a in ecma_op_function_call jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1085
[...]

PR #3766, opened 2020-05-19
Fix regexp flag handling in case of regexp like object

galpeter:
The regexp flag should be correctly referenced and released
if an existing regexp like object is used for constructing a new one.

Fixes: #3760

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
@@ -136,6 +136,10 @@ ecma_builtin_regexp_dispatch_helper (const ecma_value_t *arguments_list_p, /**< 
         return flags_value;
       }
     }
+    else
+    {
+      flags_value = ecma_copy_value (flags_value);
+    }
 
     free_arguments = true;
   }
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -151,47 +155,40 @@ ecma_builtin_regexp_dispatch_helper (const ecma_value_t *arguments_list_p, /**<
...
+    {
+      ecma_deref_object (new_target_obj_p);
+    }
   }
   else
   {
-#endif /* ENABLED (JERRY_ES2015) */
-    ret_value = ecma_op_create_regexp_from_pattern (new_target_obj_p, pattern_value, flags_value);
-#if  ENABLED (JERRY_ES2015)
+    ret_value = ECMA_VALUE_ERROR;
```
dbatyai, line 181: How about assigning `ECMA_VALUE_ERROR` at the declaration?

jerry-core/ecma/builtin-objects/ecma-builtin-regexp.c
```diff
@@ -151,47 +155,40 @@ ecma_builtin_regexp_dispatch_helper (const ecma_value_t *arguments_list_p, /**<
...
-  if (JERRY_UNLIKELY (new_target_obj_p == NULL))
+  if (JERRY_LIKELY (new_target_obj_p != NULL))
   {
 #if ENABLED (JERRY_ES2015)
-    if (free_arguments)
+    if (create_regexp_from_bc)
     {
-      ecma_free_value (pattern_value);
-      ecma_free_value (flags_value);
+      ret_value = ecma_op_create_regexp_from_bytecode (new_target_obj_p, bc_p);
```
dbatyai, line 166: It seems this can never throw. If that's the case we could move the `ECMA_IS_VALUE_ERROR ` check into the else case, and leave an assert here.

galpeter: True, will do the update

galpeter: @dbatyai I've updated the PR.

Issue #3761, opened 2020-05-18
parseFloat("IE") returns Infinity


lauriro:
```
console.log(parseFloat("IE"), parseFloat("Infinity"))
// jerry returns: Infinity Infinity
// node returns: NaN Infinity
```

setup

> "Arch Linux" (Linux 5.5.9-arch1-2 x86_64)
> commit: 95aa82763501d0bf1592d121fc0e8af5066ee726
> build command: python tools/build.py

PR #3769, opened 2020-05-20
Fix parseFloat for non-infinity values

galpeter:
The parseFloat method performed the "Infinity" string check incorrectly.

Fixes: #3761

### DIFF
--- a/jerry-core/ecma/base/ecma-helpers-number.c
+++ b/jerry-core/ecma/base/ecma-helpers-number.c
@@ -934,19 +934,18 @@ ecma_number_parse_float (const lit_utf8_byte_t *string_buff, /**< routine's firs
     }
   }
 
-  const lit_utf8_byte_t *infinity_str_p = lit_get_magic_string_utf8 (LIT_MAGIC_STRING_INFINITY_UL);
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/base/ecma-helpers-number.c
```diff
@@ -934,19 +934,18 @@ ecma_number_parse_float (const lit_utf8_byte_t *string_buff, /**< routine's firs
...
-  lit_utf8_byte_t *infinity_str_end_p = infinity_str_curr_p + sizeof (*infinity_str_p);
+  const lit_utf8_size_t infinity_length = lit_get_magic_string_size (LIT_MAGIC_STRING_INFINITY_UL);
 
-  /* Check if string is equal to "Infinity". */
-  while (str_curr_p < str_end_p
-         && *str_curr_p++ == *infinity_str_curr_p++)
+  /* The input string should be at least the length of "Infinity" to be correctly processed as
+   * the infinity value.
+   */
+  if (string_buff_size >= infinity_length
```
zherczeg, line 944: Shouldn't be `==` ?

galpeter: No, as "Infinityfoo" should be processed as "Infinity"

PR #3772, opened 2020-05-21
Change raw string length calculation method

rwalczyna:
New method uses length of source to calculate raw string length - as discussed in #3736

Also bug with template literal was fixed. Template object should have
indexed properties enumerable.


jerry-core/parser/js/js-lexer.c
```diff
@@ -1211,6 +1203,13 @@ lexer_parse_string (parser_context_t *context_p, /**< context */
     }
   }
 
+#if ENABLED (JERRY_ES2015)
+  if (opts & LEXER_STRING_RAW)
+  {
+    length = (size_t)(source_p - string_start_p) - raw_length_dec;
```
rerobika, line 1209: One space after `(size_t)`

Issue #3779, opened 2020-05-22
Assertion compressed_pointer != JMEM_CP_NULL in jmem_decompress_pointer


renatahodovan:
###### JerryScript revision
876622a0

###### Build platform
Linux-4.15.0-88-generic-x86_64-with-Ubuntu-18.04-bionic

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --profile=es2015-subset
```

###### Test case
```javascript
var a = [$];
var $ = a.lastIndexOf($, {
    valueOf: function() {
        a.length = 0
    }
})
```

###### Output
```text
ICE: Assertion 'compressed_pointer != JMEM_CP_NULL' failed at jerryscript/jerry-core/jmem/jmem-allocator.c(jmem_decompress_pointer):215.
Error: ERR_FAILED_INTERNAL_ASSERTION
ASAN:DEADLYSIGNAL
=================================================================
==6827==ERROR: AddressSanitizer: ABRT on unknown address 0x00001aab (pc 0xf7fb3079 bp 0xff9cc0bc sp 0xff9cc0a0 T0)
    #0 0xf7fb3078  (linux-gate.so.1+0x1078)
    #1 0xf7fb3078  (linux-gate.so.1+0x1078)
    #2 0xf77da831 in raise (/lib/i386-linux-gnu/libc.so.6+0x2d831)
    #3 0xf77dbcc0 in abort (/lib/i386-linux-gnu/libc.so.6+0x2ecc0)
    #4 0x56603444 in jerry_port_fatal jerryscript/jerry-port/default/default-fatal.c:30
    #5 0x566bab09 in jerry_fatal jerryscript/jerry-core/jrt/jrt-fatals.c:63
    #6 0x566bab4a in jerry_assert_fail jerryscript/jerry-core/jrt/jrt-fatals.c:87
    #7 0x566b9be9 in jmem_decompress_pointer jerryscript/jerry-core/jmem/jmem-allocator.c:215
    #8 0x5666dd30 in ecma_builtin_array_prototype_object_last_index_of jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c:1782
[...]

PR #3789, opened 2020-05-25
Fix array length check in lastIndexOf

galpeter:
When lastIndexOf is executed the array length can change and in
case of fast arrays the length should be re-checked.

Fixes: #3779

### DIFF
--- a/tests/jerry/regression-test-issue-3779.js
+++ b/tests/jerry/regression-test-issue-3779.js
@@ -1,0 +1,32 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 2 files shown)


tests/jerry/regression-test-issue-3779.js
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+var a = [1, 2, 3];
+var idx = a.lastIndexOf(50, {
+    valueOf: function() {
+        // Trigger removing elements from fast array.
+        a.length = 0;
+    }
+})
```
rerobika, line 21: Please add the following testcase as well to ensure that the added element during the callback does not affect the original limit of the lookup.

```js
var a = [1, 2, 3, 4];
var idx = a.lastIndexOf(4, {
    valueOf: function() {
        a.push(4)
    }
})

assert (idx === -1);
```

galpeter: True. Thanks, I've updated the pr.

PR #3792, opened 2020-05-25
Fix exclude list copying when --test262-es2015 flag is used

rwalczyna:
test262-es6-excludelist.xml is not updated in test262 directory after
first initialization.


rwalczyna:
@ossy-szeged, you are right. So I am closing this PR and looking forward for your fix :)

PR #3798, opened 2020-05-26
Fix length check for Array.prototype indexOf, lastIndexOf and fill

szilagyiadam:

### DIFF
--- a/tests/jerry/array-prototype-slice.js
+++ b/tests/jerry/array-prototype-slice.js
@@ -151,3 +151,41 @@ try {
   assert (e.message === "foo");
   assert (e instanceof ReferenceError);
 }
+
+function array_check(result_array, expected_array) {
+  assert(result_array instanceof Array);
+  assert(result_array.length === expected_array.length);
+  for (var idx = 0; idx < expected_array.length; idx++) {
+    assert(result_array[idx] === expected_array[idx]);
+  }
+}
+
+// Remove the buffer
+var array = [1, 2, 3, 4, 5];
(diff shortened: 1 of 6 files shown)


tests/jerry/array-prototype-slice.js
```diff
@@ -151,3 +151,33 @@ try {
...
+
+// Remove the buffer
+var array = [1, 2, 3, 4, 5];
+var value = array.slice(4, {
+    valueOf: function() {
+        array.length = 0;
+    }
+})
+
+print(JSON.stringify(value) === '[]')
```
galpeter, line 163: This should be an assert. And just to make sure, we should avoid the JSON.stringify calls in the future ( we could use an helper method to do the comparison like in: https://github.com/jerryscript-project/jerryscript/commit/908240ba6223fb54e72c2ab85c3765db3492a968#diff-2b73935d70c05987656fb33f1150f9d1R19 ). The same applies to the parts below.

tests/jerry/es2015/array-prototype-fill.js
```diff
@@ -177,3 +177,42 @@ function TestFillFrozenObject () {
...
+
+// Extend the buffer
+var array = [1, 2, 3];
+var value = array.fill(1, {
+    valueOf: function() {
+        array.length = 6;
+    }
+})
+
+array_check(value, [1,1,1,undefined,undefined,undefined]);
```
galpeter, line 208: Please add spaces after the comma in the array. (This should be applied below and the file above)

PR #3799, opened 2020-05-26
Fix Promise thenable.

zherczeg:
A single promise can be resolved any number of times due to thenable functions, but each resolver pair can only be called once.

### DIFF
--- a/jerry-core/ecma/operations/ecma-promise-object.c
+++ b/jerry-core/ecma/operations/ecma-promise-object.c
@@ -162,6 +162,38 @@ ecma_promise_trigger_reactions (ecma_collection_t *reactions, /**< lists of reac
 } /* ecma_promise_trigger_reactions */
 
 /**
+ * Checks whether a resolver is called before.
+ *
+ * @return true if it was called before, false otherwise
+ */
+static bool
+ecma_is_resolver_already_called (ecma_object_t *resolver_p, /**< resolver */
+                                 ecma_object_t *promise_obj_p) /**< promise */
+{
+  ecma_string_t *str_already_resolved_p = ecma_get_magic_string (LIT_INTERNAL_MAGIC_STRING_ALREADY_RESOLVED);
+  ecma_property_t *property_p = ecma_find_named_property (resolver_p, str_already_resolved_p);
+
+  if (property_p == NULL)
+  {
+    return (ecma_promise_get_flags (promise_obj_p) & ECMA_PROMISE_ALREADY_RESOLVED) != 0;
+  }
+
+  JERRY_ASSERT (ECMA_PROPERTY_GET_TYPE (*property_p) == ECMA_PROPERTY_TYPE_NAMEDDATA);
+
+  ecma_value_t already_resolved = ECMA_PROPERTY_VALUE_PTR (property_p)->value;
+  ecma_object_t *object_p = ecma_get_object_from_value (already_resolved);
+  JERRY_ASSERT (ecma_get_object_type (object_p) == ECMA_OBJECT_TYPE_CLASS);
+
+  ecma_extended_object_t *already_resolved_p = (ecma_extended_object_t *) object_p;
+  JERRY_ASSERT (already_resolved_p->u.class_prop.class_id == LIT_MAGIC_STRING_BOOLEAN_UL);
+
+  ecma_value_t current_value = already_resolved_p->u.class_prop.u.value;
+  already_resolved_p->u.class_prop.u.value = ECMA_VALUE_TRUE;
+
+  return current_value == ECMA_VALUE_TRUE;
+} /* ecma_is_resolver_already_called */
+
+/**
  * Reject a Promise with a reason.
  *
  * See also: ES2015 25.4.1.7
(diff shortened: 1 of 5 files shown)


jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -437,6 +467,25 @@ ecma_promise_create_resolving_functions (ecma_object_t *object_p, /**< the promi
...
                                                                   ecma_promise_reject_handler);
+  if (!create_already_resolved)
+  {
+    return;
+  }
+
+  ecma_value_t already_resolved = ecma_op_create_boolean_object (ECMA_VALUE_FALSE);
+  ecma_string_t *str_already_resolved_p = ecma_get_magic_string (LIT_INTERNAL_MAGIC_STRING_ALREADY_RESOLVED);
+
+  ecma_op_object_put (ecma_get_object_from_value (funcs->resolve),
```
rerobika, line 478: I'd rather use create_namedata_property here since, since both `funcs->resolve` and `funcs->reject` created above should not have existing `LIT_INTERNAL_MAGIC_STRING_ALREADY_RESOLVED` property.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -161,6 +161,34 @@ ecma_promise_trigger_reactions (ecma_collection_t *reactions, /**< lists of reac
   }
 } /* ecma_promise_trigger_reactions */
 
+static bool
+ecma_is_resolver_already_called (ecma_object_t *resolver_p, /**< resolver */
+                                 ecma_object_t *promise_obj_p) /**< promise */
+{
+  ecma_value_t already_resolved;
+  already_resolved = ecma_op_object_get_by_magic_id (resolver_p, LIT_INTERNAL_MAGIC_STRING_ALREADY_RESOLVED);
```
rerobika, line 169: `find_named_property` can be used here.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -242,10 +270,9 @@ ecma_promise_reject_handler (const ecma_value_t function, /**< the function itse
   JERRY_ASSERT (ecma_is_promise (promise_obj_p));
 
   /* 3., 4. */
-  if (ecma_promise_get_flags (promise_obj_p) & ECMA_PROMISE_ALREADY_RESOLVED)
+  if (ecma_is_resolver_already_called (function_p, promise_obj_p))
```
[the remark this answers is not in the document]
zherczeg: I wanted it to be consistent with the other similar function. I don't like deep ifs.

jerry-core/ecma/operations/ecma-promise-object.c
zherczeg: I am not sure this is much of a hot path, but I reworked the patch.

jerry-core/ecma/operations/ecma-promise-object.c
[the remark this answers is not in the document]
rerobika: `goto`s in other similar functions are used to skip multiple level nested if/for statements or in some performance cirical code paths. In this case the nesting if level would be **1** which is neither _"deep"_ nor performance critical. So +1 for removing this goto.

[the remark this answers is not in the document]
zherczeg: Shall we keep the goto in the other function?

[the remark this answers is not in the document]
rerobika: Since the other -not exactly specified- function is not corresponds to the fix I'd keep it.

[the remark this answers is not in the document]
galpeter: Yes, for now. Just change this single instance. (The other function is a bigger)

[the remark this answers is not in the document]
zherczeg: I realized this doesn't matter at all, since for `await`, I need to rework the whole thing anyway.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -161,6 +161,37 @@ ecma_promise_trigger_reactions (ecma_collection_t *reactions, /**< lists of reac
   }
 } /* ecma_promise_trigger_reactions */
 
+static bool
```
rerobika, line 164: No description is provided.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -161,6 +161,37 @@ ecma_promise_trigger_reactions (ecma_collection_t *reactions, /**< lists of reac
...
+
+  JERRY_ASSERT (ecma_get_object_type (object_p) == ECMA_OBJECT_TYPE_CLASS);
+
+  ecma_extended_object_t *already_resolved_p = (ecma_extended_object_t *) object_p;
+
+  JERRY_ASSERT (already_resolved_p->u.class_prop.class_id == LIT_MAGIC_STRING_BOOLEAN_UL);
+
+  ecma_value_t current_value = already_resolved_p->u.class_prop.u.value;
+
+  already_resolved_p->u.class_prop.u.value = ECMA_VALUE_TRUE;
```
rerobika, line 190: I do not really prefer the new lines after each instruction.

zherczeg:
Thank you for the review. Patch is updated.

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -437,6 +468,28 @@ ecma_promise_create_resolving_functions (ecma_object_t *object_p, /**< the promi
...
+  ecma_string_t *str_already_resolved_p = ecma_get_magic_string (LIT_INTERNAL_MAGIC_STRING_ALREADY_RESOLVED);
+  ecma_property_value_t *value_p;
+
+  value_p = ecma_create_named_data_property (ecma_get_object_from_value (funcs->resolve),
+                                             str_already_resolved_p,
+                                             ECMA_PROPERTY_FIXED,
+                                             NULL);
+  value_p->value = already_resolved;
+
+  value_p = ecma_create_named_data_property (ecma_get_object_from_value (funcs->reject),
```
galpeter, line 486: Is it possible to have the same kind of test for "reject" handler like the "resolve" one has?

zherczeg: What do you mean?

galpeter: A test case just like the one added in this PR. But probably reject behaves a bit differently.

zherczeg: Ok, I get it now. Reject cannot be "delayed", that is a special feature of "resolve". Probably it would be too easy to start an infinite reject loop :)

PR #3802, opened 2020-05-27
Create extend byte code flags with function types.

zherczeg:
This patch provides the necessary code flag for #3745

### DIFF
--- a/jerry-core/parser/js/byte-code.h
+++ b/jerry-core/parser/js/byte-code.h
@@ -814,23 +814,75 @@ typedef struct
  */
 typedef enum
 {
-  CBC_CODE_FLAGS_FUNCTION = (1u << 0), /**< compiled code is JavaScript function */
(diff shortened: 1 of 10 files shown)


jerry-core/parser/js/byte-code.h
```diff
@@ -821,16 +821,59 @@ typedef enum
...
+
+  /* Bits from bit 13 is reserved for function types (see CBC_CODE_FLAGS_FUNCTION_TYPE_SHIFT). */
+} cbc_code_flags_t;
+
+/**
+ * Compact byte code function types.
+ */
+typedef enum
+{
+  CBC_FUNCTION_NORMAL, /**< function without special properties */
```
rerobika, line 838: Cannot we extend this list with Regexp literals? `CBC_CODE_FLAGS_FUNCTION` is only used to represent that the compiled code is not RegExp, so IMO this should be a new type.

PR #3803, opened 2020-05-27
Fix tools/run-tests.py --test262-es2015

ossy-szeged:
The output of the test262 runner changed between ES5.1 and ES6,
our script should respect it to signal regressions in return code too.

### DIFF
--- a/tools/runners/run-test-suite-test262.py
+++ b/tools/runners/run-test-suite-test262.py
@@ -113,7 +113,7 @@ def main(args):
                             universal_newlines=True,
                             stdout=subprocess.PIPE)
 
-    return_code = 0
+    return_code = 1
     with open(os.path.join(os.path.dirname(args.engine), 'test262.report'), 'w') as output_file:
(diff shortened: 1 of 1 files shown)


tools/runners/run-test-suite-test262.py
```diff
@@ -132,7 +132,7 @@ def main(args):
 
             if summary_found:
                 print(output, end='')
-                if ('Failed tests' in output) or ('Expected to fail but passed' in output):
+                if ('Failed tests' in output) or ('Failed Tests' in output) or ('Expected to fail but' in output):
```
dbatyai, line 135: How about we invert the logic, and check if "All tests succeeded" is in the output instead? That way if the output changes later we don't silently ignore potential test fails.

ossy-szeged: Good catch, it works fine with ES5.1 and ES6 version of test262 runner. I'll update the fix.

PR #3807, opened 2020-05-28
Use stringbuilder in ecma_builtin_helper_object_to_string

szilagyiadam:
Also added steps 4-6 from ECMA-262 v6, 19.1.3.6


jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -117,15 +117,17 @@ ecma_builtin_helper_object_to_string_tag_helper (ecma_value_t tag_value) /**< st
 ecma_value_t
 ecma_builtin_helper_object_to_string (const ecma_value_t this_arg) /**< this argument */
 {
-  lit_magic_string_id_t type_string;
+  ecma_stringbuilder_t builder = ecma_stringbuilder_create_raw ((lit_utf8_byte_t *) "[object ", 8);
```
dbatyai, line 120: This could be delayed right until the end, that way you don't need to destroy it on errors.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -140,58 +142,49 @@ ecma_builtin_helper_object_to_string (const ecma_value_t this_arg) /**< this arg
...
-  {
-    buffer_ptr = lit_copy_magic_string_to_buffer (magic_string_ids[i], buffer_ptr,
-                                                  (lit_utf8_size_t) ((str_buffer + buffer_size) - buffer_ptr));
-    JERRY_ASSERT (buffer_ptr <= str_buffer + buffer_size);
+    ecma_deref_object (obj_p);
   }
 
-  ret_string_p = ecma_new_ecma_string_from_utf8 (str_buffer, (lit_utf8_size_t) (buffer_ptr - str_buffer));
+  ecma_stringbuilder_append_magic (&builder, builtin_tag);
+  ecma_stringbuilder_append_magic (&builder, LIT_MAGIC_STRING_RIGHT_SQUARE_CHAR);
```
dbatyai, line 185: You could do `ecma_stringbuilder_append_byte` instead.

jerry-core/ecma/builtin-objects/ecma-builtin-helpers.c
```diff
@@ -140,58 +140,48 @@ ecma_builtin_helper_object_to_string (const ecma_value_t this_arg) /**< this arg
...
-    LIT_MAGIC_STRING_RIGHT_SQUARE_CHAR
-  };
-
-  for (uint32_t i = 0; i < sizeof (magic_string_ids) / sizeof (lit_magic_string_id_t); ++i)
-  {
-    buffer_ptr = lit_copy_magic_string_to_buffer (magic_string_ids[i], buffer_ptr,
-                                                  (lit_utf8_size_t) ((str_buffer + buffer_size) - buffer_ptr));
-    JERRY_ASSERT (buffer_ptr <= str_buffer + buffer_size);
-  }
+  ecma_stringbuilder_t builder = ecma_stringbuilder_create_raw ((lit_utf8_byte_t *) "[object ", 8);
```
zherczeg, line 179: What about creating a magic string for `[object `? If this string is in the binary file, the JS code should be able to take advantage of it.

dbatyai: I don't think there's a lot to take advantage of. This string practically never happens by itself, so the engine wouldn't be able to use it for anything. It would just clutter the magic strings.

zherczeg: My reasoning is the string is already in the binary, and if a JS will ever use it: it is a gain, if never: no disadvantage. Magic string search is logaritmic, you need to double them to add one new step.

PR #3811, opened 2020-05-29
Implement argument redefinition with var statement.

zherczeg:
#3806 should go in first


zherczeg:
Patch is updated to master. Please review it. Thank you.

jerry-core/parser/js/js-scanner-util.c
```diff
@@ -802,6 +809,14 @@ scanner_pop_literal_pool (parser_context_t *context_p, /**< context */
         {
           type = SCANNER_STREAM_TYPE_DESTRUCTURED_ARG;
         }
+
+        if (literal_p->type & SCANNER_LITERAL_IS_VAR)
+        {
+          type = (uint8_t) (type + 1);
```
rerobika, line 815: `type++`?

zherczeg: That is not working, becasue type is converted to int before ++, and int cannot be assigned to byte without casting when werror is enabled.

PR #3838, opened 2020-06-03
Fix parsing function statements containing invalid tokens

### LINKED ISSUE #3821: Assertion 'context_p->next_scanner_info_p->type == SCANNER_TYPE_FUNCTION' in parser_parse_function_statement
###### JerryScript revision
d06c3a7

###### Build platform
Ubuntu 16.04.6 LTS (Linux 4.15.0-99-generic x86_64)

###### Build steps
```
python tools/build.py --profile=es2015-subset --lto=off --compile-flag=-g \
--error-messages=on --debug --compile-flag=-g --strip=off --logging=on \
--compile-flag=-fsanitize=address --stack-limit=15
```

###### Test case
```
function a") {
```

###### Output
```
ICE: Assertion 'context_p->next_scanner_info_p->type == SCANNER_TYPE_FUNCTION' failed at /home/JerryScript/jerryscript/jerry-core/parser/js/js-parser-statm.c(parser_parse_function_statement):733.
Aborted (core dumped)
```

Credits: This vulnerability is detected by chong from OWL337.

dbatyai:
Fixes #3821.

### DIFF
--- a/jerry-core/parser/js/js-parser-statm.c
+++ b/jerry-core/parser/js/js-parser-statm.c
@@ -710,9 +710,9 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
                 && context_p->token.lit_location.type == LEXER_IDENT_LITERAL);
 
 #if ENABLED (JERRY_ES2015)
-  if (context_p->next_scanner_info_p->source_p == context_p->source_p)
+  if (context_p->next_scanner_info_p->source_p == context_p->source_p
+      && context_p->next_scanner_info_p->type == SCANNER_TYPE_ERR_REDECLARED)
   {
-    JERRY_ASSERT (context_p->next_scanner_info_p->type == SCANNER_TYPE_ERR_REDECLARED);
     parser_raise_error (context_p, PARSER_ERR_VARIABLE_REDECLARED);
   }
 #endif /* ENABLED (JERRY_ES2015) */
@@ -729,8 +729,6 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
   {
     status_flags |= PARSER_HAS_NON_STRICT_ARG;
   }
-
-  JERRY_ASSERT (context_p->next_scanner_info_p->type == SCANNER_TYPE_FUNCTION);
 
 #if ENABLED (JERRY_ES2015)
   if (is_generator_function)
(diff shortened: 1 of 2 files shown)


jerry-core/parser/js/js-parser-statm.c
```diff
@@ -710,9 +710,8 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
                 && context_p->token.lit_location.type == LEXER_IDENT_LITERAL);
 
 #if ENABLED (JERRY_ES2015)
-  if (context_p->next_scanner_info_p->source_p == context_p->source_p)
```
zherczeg, line 713: This is the way it should work. If you check the type, but not the `source_p`, it might refer to something detected later in the source code.

jerry-core/parser/js/js-scanner.c
```diff
@@ -1418,6 +1418,7 @@ scanner_scan_statement (parser_context_t *context_p, /**< context */
 #endif /* ENABLED (JERRY_ES2015) */
 
       scanner_push_literal_pool (context_p, scanner_context_p, status_flags);
+      scanner_context_p->active_literal_pool_p->source_p = context_p->source_p;
```
zherczeg, line 1421: The function arguments should set this.

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -730,7 +730,10 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
     status_flags |= PARSER_HAS_NON_STRICT_ARG;
   }
 
-  JERRY_ASSERT (context_p->next_scanner_info_p->type == SCANNER_TYPE_FUNCTION);
+  if (context_p->next_scanner_info_p->type != SCANNER_TYPE_FUNCTION)
```
zherczeg, line 733: Which test triggers this? At this point the `(` should be parsed where the scanner info is attached to.

zherczeg: Ok this is early. Perhaps simply removing this assertion is enough.

dbatyai: This will be triggered by `function a"b`. The scanner will call `lexer_next_token` to consume the left paren, however the unterminated string literal will throw. Since the scanner info would be created after consuming the opening paren, there will be no scanner info for the function.

zherczeg: Exactly. I think this assert is unnecessary, since no info is created if there will be a parse error before the `(`. I think there is another assert later which should cover this as well.

Issue #3837, opened 2020-06-03
SEGV ecma_deref_object (erry-core/ecma/base/ecma-gc.c:149)


nszetei:
###### JerryScript revision

a56e31f194fe52000ed95605ae3c92390cd60ad5

###### Build platform
Ubuntu 18.04.4 LTS (Linux 4.15.0-91-generic x86_64)

###### Build steps

- first output

```sh
python tools/build.py --profile=es2015-subset --lto=off --compile-flag=-g \
--error-messages=on --debug --strip=off --logging=on \
 --compile-flag=-fsanitize=address --stack-limit=15
```

- second output

```sh
python tools/build.py --profile=es2015-subset --lto=off --compile-flag=-g  \
--strip=off --logging=on  --stack-limit=15
```

###### Test case

```javascript
function main() {
var v1 = [1337,1337,1337,1337,1337];
var v5 = {length:"size",d:343661.18346282793,__proto__:13.37};
var v7 = {ownKeys:gc,defineProperty:gc,getPrototypeOf:gc,has:gc,call:gc,apply:gc,getOwnPropertyDescriptor:gc,isExtensible:gc,deleteProperty:v5,construct:gc};
var v9 = new Proxy(gc,v7);
var v10 = v9;
function v13(v14,v15,v16) {
}
var v17 = {set:Object,apply:Object,call:v13,get:resourceName,getOwnPropertyDescriptor:v13,deleteProperty:resourceName,isExtensible:v13,construct:resourceName,setPrototypeOf:v13,ownKeys:Object,preventExtensions:v13,defineProperty:Object};
var v19 = new Proxy(v10,v17);
delete v19[v1];
}
main();
```

###### Execution steps

```sh
$ ~/jerryscript/build/bin/jerry 01.js
AddressSanitizer:DEADLYSIGNAL
=================================================================
==15982==ERROR: AddressSanitizer: SEGV on unknown address 0x006ddfff806d (pc 0x55b637251552 bp 0x7fffa420e420 sp 0x7fffa420e410 T0)
==15982==The signal is caused by a READ memory access.
    #0 0x55b637251551 in ecma_deref_object /home/jerryscript/jerry-core/ecma/base/ecma-gc.c:149
    #1 0x55b63726dd83 in ecma_free_property_descriptor /home/jerryscript/jerry-core/ecma/base/ecma-helpers.c:1204
[...]

PR #3840, opened 2020-06-03
Fix property descriptor queries

galpeter:
When the getOwnPropertyDescriptor method was invoked the input property
descriptor was not cleared in every case. This could lead to problems when
the property descriptor is not set/modified by the getOwnPropertyDescriptor
call, resulting in a failure at a later state.

Related to this the Proxy getOwnPropertyDescriptor method incorrectly returned
"undefined" value in a single case.

Fixes: #3837

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -1858,6 +1858,8 @@ ecma_op_object_define_own_property (ecma_object_t *obj_p, /**< the object */
  *                                      [Enumerable], [Configurable]
  *                                    }.
  *
+ * The output property descriptor will always be initialized to an empty descriptor.
+ *
  * @return true - if property found
  *         false - otherwise
  */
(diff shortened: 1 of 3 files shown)


zherczeg:
Initializing the descriptior was not a requirement before, so is the `ecma_op_object_define_own_property` function is wrong or the function who calls it?

galpeter:
The original problem came from the incorrect proxy implementation which resulted in check after the `ecma_op_object_define_own_property` call to be false, thus a next instruction incorrectly assumed that the descriptor has some useful data (the input descriptor is a stack allocated variable). The PR fixes the proxy issue and also resets the descriptor in every case to avoid such errors in the future.

In the original `ecma_op_object_define_own_property` the descriptor will only be inited if the property was found. In some places where the `ecma_op_object_define_own_property` is used the descriptor is inited to empy in other places it does not. I can revert the initialization if you want it.

PR #3848, opened 2020-06-04
Fix several function name related issues

rerobika:
- For non-computed name script functions only identifier/string literal should be set as name
 - Implicit class constructor names with non string function name should be ToString-ed


rerobika:
Note: The issues were found by [Fuzzinator](http://fuzzinator.readthedocs.io/) with [grammarinator](https://github.com/renatahodovan/grammarinator).

PR #3852, opened 2020-06-04
Context break should honor private contexts.

zherczeg:
Another boring one-liner.


dbatyai:
This description isn't very appropriate, and adds nothing to the discussion. There's no obligation to provide a description, if you have nothing meaningful to say then it's ok to leave it empty.

akosthekiss:
Please, add the `Fixes #` notation to the commit message.

PR #3854, opened 2020-06-04
Primitive this values of accessors should not be coerced in strict mode

dbatyai:

### DIFF
--- a/jerry-core/vm/vm.c
+++ b/jerry-core/vm/vm.c
@@ -154,84 +154,101 @@ vm_op_get_value (ecma_value_t object, /**< base object */
  *         if the property setting is unsuccessful
  */
 static ecma_value_t
-vm_op_set_value (ecma_value_t object, /**< base object */
+vm_op_set_value (ecma_value_t base, /**< base object */
                  ecma_value_t property, /**< property name */
                  ecma_value_t value, /**< ecma value */
                  bool is_strict) /**< strict mode */
 {
-  ecma_object_t * object_p;
-
-  if (JERRY_UNLIKELY (!ecma_is_value_object (object)))
-  {
-    ecma_value_t to_object = ecma_op_to_object (object);
-    ecma_free_value (object);
-
-    if (ECMA_IS_VALUE_ERROR (to_object))
-    {
-#if ENABLED (JERRY_ERROR_MESSAGES)
-      ecma_free_value (to_object);
-      jcontext_release_exception ();
-
-      ecma_value_t error_value = ecma_raise_standard_error_with_format (ECMA_ERROR_TYPE,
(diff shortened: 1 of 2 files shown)


jerry-core/vm/vm.c
```diff
@@ -154,55 +154,38 @@ vm_op_get_value (ecma_value_t object, /**< base object */
...
                  ecma_value_t value, /**< ecma value */
                  bool is_strict) /**< strict mode */
 {
-  ecma_object_t * object_p;
+  ecma_value_t result = ECMA_VALUE_EMPTY;
+  ecma_object_t *object_p;
+  ecma_string_t *property_p;
 
-  if (JERRY_UNLIKELY (!ecma_is_value_object (object)))
+  if (JERRY_UNLIKELY (!ecma_is_value_object (base)
```
zherczeg, line 166: I think this part is always true if the second part is true.

jerry-core/vm/vm.c
```diff
@@ -211,27 +194,46 @@ vm_op_set_value (ecma_value_t object, /**< base object */
...
   {
-    completion_value = ecma_op_object_put (object_p,
-                                           property_p,
-                                           value,
-                                           is_strict);
+    ecma_value_t object = ecma_op_to_object (base);
+    JERRY_ASSERT (!ECMA_IS_VALUE_ERROR (object));
+
+    object_p = ecma_get_object_from_value (object);
+    ecma_op_ordinary_object_prevent_extensions (object_p);
```
zherczeg, line 203: Is this standard?

jerry-core/vm/vm.c
dbatyai: Since this case is extremely unlikely, I added the first condition to reduce the number of checks when `base` is an object, which is more likely. I can remove it if you want.

jerry-core/vm/vm.c
dbatyai: This has been added as part of #2917, it's seems to me that it's correct.

jerry-core/vm/vm.c
zherczeg: This looks like a very hoth path, and this is checked twice. What about moving the `ecma_is_value_prop_name` part before this, and combine this part with the rest?

dbatyai: Unfortunatley the errors must be thrown in this order, the base object needs to be checked first, followed by the property name. The checks could be reduced by adding a completely separate code path for the non-object case and duplicating/factoring out the property name coercion. What do you think?

zherczeg: This is a key function of vm, so in this case I would not mind the extra binary size.

PR #3863, opened 2020-06-05
Fix three async function issues.

zherczeg:
- Invalid assert
- Add missing async prefix check when an identifier is enclosed in brackets
- Adding a new byte-code


akosthekiss:
In bug fixes, please, mention the related issues in the commit message with the `Fixes #...` notation so that Github can close them automatically.

akosthekiss:
Hmm, I mist have missed something. I cannot see the "Fixes" string in the commit message but Github still links the issues. How's that?...

rerobika:
@akosthekiss There is a new GitHub feature on the right side called linked issues, where you can manually assign  issues to the PR which fixes them.

akosthekiss:
I wonder whether that's a good thing? This way the git history will not contain this information anymore. I.e., only Github will have this metadata. I'd keep the manual linking (and thus the loss of this metadata) to a minimum.

zherczeg:
If github goes away, the issues will go away, so the changelog will only have meaningless numbers. Probably this is the reason of this improvement.

akosthekiss:
@zherczeg You've got a point there. Still, the commit message will contain some hint that this change is a fix. That's not much, I give you that, but still some kind of metadata.

(And if there will _ever_ arise the need for any migration, that hypothetical migration and new platform will more likely support the general basics -- like issue/PR id, title, discussion -- than Github specifics -- like linked issues metadata.)

Adding a link via the commit message is only 11 characters / key presses (space included). I still think that it's not a high price.

ossy-szeged:
I agree with @akosthekiss, I prefer adding #xyz to the commit log too. If the commit log doesn't contain it, you have to open the PR first and then search the linked issues. (Not to mention that github sometimes missed to add the PR number to the commit log too.)

akosthekiss:
> I agree with @akosthekiss, I prefer adding #xyz to the commit log too. If the commit log doesn't contain it, you have to open the PR first and then search the linked issues. 

Ah, indeed. I've almost missed that. The email notification about the PR will contain the issue hyperlinked if "Fixes" is there in the commit message. Which is a nice extra.

> (Not to mention that github sometimes missed to add the PR number to the commit log too.)

Actually, that's on us. Depending on whether we press squash-and-merge or rebase-and-merge when landing a PR, Github behaves differently. We should disable the problematic option at last to avoid mistakes.

zherczeg:
The problem is actually the easiness of adding these numbers without any sanity check. I remember many times when I accidentally added wrong numbers, because many pages were open in my browser, or misread the number, or wrote something different I wanted (i.e. the PR number not the issue number). Often these were not noticed by reviewers as well, and landed my patch before I can change these numbers. But selecting the issue directly from a list reduces the possible errors.

akosthekiss:
As a fix should contain the test reported in the issue as a regression test with the issue number in the file name, I see this as a problem that can be easily cross-checked. (I.e., same number in commit message an in the name of the added test case.)

jerry-core/vm/vm.c
```diff
@@ -2082,6 +2082,13 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
         }
         case VM_OC_RETURN_PROMISE:
         {
+          left_value = ECMA_VALUE_UNDEFINED;
```
rerobika, line 2085: How about adding `VM_OC_GET_STACK` flag for `CBC_EXT_RETURN_PROMISE`? `left_value` is always initialized to `ECMA_VALUE_UNDEFINED` in each loop which is suitable for `CBC_EXT_RETURN_PROMISE_UNDEFINED` .

zherczeg: This is actually a rather clever idea!

PR #3864, opened 2020-06-05
Add Generator prototype custom dispatcher

Yuyupo:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-generator-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-generator-prototype.c
@@ -28,6 +28,22 @@
 #define ECMA_BUILTINS_INTERNAL
 #include "ecma-builtins-internal.h"
 
+/**
+  * This object has a custom dispatch function.
+  */
+ #define BUILTIN_CUSTOM_DISPATCH
+
+/**
+ * List of built-in routine identifiers.
+ */
+enum
+{
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_START = ECMA_BUILTIN_ID__COUNT - 1,
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_RETURN,
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_THROW,
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_NEXT
+};
+
 #define BUILTIN_INC_HEADER_NAME "ecma-builtin-generator-prototype.inc.h"
 #define BUILTIN_UNDERSCORED_ID generator_prototype
 #include "ecma-builtin-internal-routines-template.inc.h"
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-generator-prototype.c
```diff
@@ -28,6 +28,22 @@
 #define ECMA_BUILTINS_INTERNAL
 #include "ecma-builtins-internal.h"
 
+/**
+  * This object has a custom dispatch function.
```
zherczeg, line 32: Wrong indeantation.

jerry-core/ecma/builtin-objects/ecma-builtin-generator-prototype.c
```diff
@@ -28,6 +28,22 @@
...
+
+/**
+ * List of built-in routine identifiers.
+ */
+enum
+{
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_START = ECMA_BUILTIN_ID__COUNT - 1,
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_RETURN,
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_THROW,
+  ECMA_GENERATOR_PROTOTYPE_ROUTINE_NEXT
```
zherczeg, line 44: Missong description comments.

jerry-core/ecma/builtin-objects/ecma-builtin-generator-prototype.c
```diff
@@ -194,52 +178,81 @@ ecma_builtin_generator_prototype_object_do (ecma_value_t this_arg, /**< this arg
...
+    {
+      return ecma_builtin_generator_prototype_object_do (executable_object_p,
+                                                         arguments_list_p[0],
+                                                         ECMA_ITERATOR_THROW);
+    }
+    case ECMA_GENERATOR_PROTOTYPE_ROUTINE_RETURN:
+    {
+      return ecma_builtin_generator_prototype_object_do (executable_object_p,
+                                                         arguments_list_p[0],
+                                                         ECMA_ITERATOR_RETURN);
```
zherczeg, line 246: What about putting the type into a local variable and call `ecma_builtin_generator_prototype_object_do` once after the switch?

jerry-core/ecma/builtin-objects/ecma-builtin-generator-prototype.c
```diff
@@ -194,52 +178,81 @@ ecma_builtin_generator_prototype_object_do (ecma_value_t this_arg, /**< this arg
...
+    }
+    default:
```
zherczeg, line 248: Instead of JERRY_UNREACHABLE, the last case (ECMA_GENERATOR_PROTOTYPE_ROUTINE_RETURN) can be a default case, and put an assert which checks the type.

rerobika: I'd still insist to the `JERRY_UNREACHABLE` solution.

zherczeg: Why? It should be one less check to the compiler.

Yuyupo: The reason why I'm adding custom dispatchers is not just for the binary gain but so the code itself is more readable and uniform.
Imho using the default case with an assert is less "pretty" as simply using the switch cases. I would prefer it the way it is now too.

PR #3885, opened 2020-06-09
Fix ReferenceError for let/const variables in of/in for loop

rwalczyna:
ReferenceError should be thrown when variable is used before assignment.

### DIFF
--- a/jerry-core/parser/js/js-scanner.c
+++ b/jerry-core/parser/js/js-scanner.c
@@ -2108,7 +2108,7 @@ scanner_scan_statement_end (parser_context_t *context_p, /**< context */
         while ((literal_p = (lexer_lit_location_t *) parser_list_iterator_next (&literal_iterator)) != NULL)
         {
           if ((literal_p->type & (SCANNER_LITERAL_IS_LET | SCANNER_LITERAL_IS_CONST))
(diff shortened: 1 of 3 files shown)


zherczeg:
Could you provide a simple example for the issue? I would not mind adding the test case to jerry tests as well, since if we move to another test262 hash, these tests might change.

rwalczyna:
You are right, test262 repository is external, so it might be changed. I have added test to es2015 directory.

jerry-core/parser/js/js-scanner.c
```diff
@@ -2108,7 +2108,7 @@ scanner_scan_statement_end (parser_context_t *context_p, /**< context */
         while ((literal_p = (lexer_lit_location_t *) parser_list_iterator_next (&literal_iterator)) != NULL)
         {
           if ((literal_p->type & (SCANNER_LITERAL_IS_LET | SCANNER_LITERAL_IS_CONST))
-              && literal_p->type & SCANNER_LITERAL_NO_REG)
+              && (literal_p->type & (SCANNER_LITERAL_NO_REG | SCANNER_LITERAL_IS_USED)))
```
zherczeg, line 2111: Based on SCAN_STACK_BINDING_INIT, I think `SCANNER_LITERAL_IS_USED` should be enough alone.

PR #3890, opened 2020-06-10
Refactor ecma_builtin_global_object_encode_uri_helper

szilagyiadam:
The main change is the usage of stringbuilder

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-global.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-global.c
@@ -422,35 +422,30 @@ ecma_builtin_global_object_encode_uri_helper (lit_utf8_byte_t *input_start_p, /*
                                                                            *   string buffer's size */
                                               const uint8_t *unescaped_uri_bitset_p) /**< unescaped bitset */
 {
-  /*
-   * The URI encoding has two major phases: first we validate the input,
-   * and compute the length of the output, then we encode the input.
-   */
-
   lit_utf8_byte_t *input_char_p = input_start_p;
   const lit_utf8_byte_t *input_end_p = input_start_p + input_size;
-  lit_utf8_size_t output_length = 0;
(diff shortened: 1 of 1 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-global.c
```diff
@@ -474,79 +471,25 @@ ecma_builtin_global_object_encode_uri_helper (lit_utf8_byte_t *input_start_p, /*
...
+        ecma_stringbuilder_append_raw (&builder, result_chars, 3);
       }
     }
     else
     {
       for (uint32_t i = 0; i < utf_size; i++)
       {
-        ecma_builtin_global_object_byte_to_hex (output_char_p, octets[i]);
-        output_char_p += URI_ENCODED_BYTE_SIZE;
+        ecma_builtin_global_object_byte_to_hex (result_chars, octets[i]);
```
rerobika, line 486: Move the declaration of `result_chars` here (inside the for loop). Also assert that the `utf_sizte` is not greater that the size of the buffer.

jerry-core/ecma/builtin-objects/ecma-builtin-global.c
```diff
@@ -464,89 +459,38 @@ ecma_builtin_global_object_encode_uri_helper (lit_utf8_byte_t *input_start_p, /*
       }
       else
       {
+        ecma_stringbuilder_destroy (&builder);
         return ecma_raise_uri_error (ECMA_ERR_MSG ("Unicode surrogate pair missing."));
       }
     }
 
     lit_utf8_size_t utf_size = lit_code_point_to_utf8 (cp, octets);
+    static lit_utf8_byte_t result_chars[3];
```
dbatyai, line 468: This shouldn't be static, and the length should be `URI_ENCODED_BYTE_SIZE`.

PR #3897, opened 2020-06-12
Implement async function execution.

zherczeg:
#3802 should go in first

### DIFF
--- a/jerry-core/ecma/operations/ecma-promise-object.c
+++ b/jerry-core/ecma/operations/ecma-promise-object.c
@@ -123,22 +123,28 @@ ecma_promise_trigger_reactions (ecma_collection_t *reactions, /**< lists of reac
 
   while (buffer_p < buffer_end_p)
   {
-    ecma_value_t capability_with_tag = *buffer_p++;
-    ecma_object_t *capability_obj_p = ECMA_GET_NON_NULL_POINTER_FROM_POINTER_TAG (ecma_object_t, capability_with_tag);
-    ecma_value_t capability = ecma_make_object_value (capability_obj_p);
+    ecma_value_t object_with_tag = *buffer_p++;
+    ecma_object_t *object_p = ECMA_GET_NON_NULL_POINTER_FROM_POINTER_TAG (ecma_object_t, object_with_tag);
+    ecma_value_t object = ecma_make_object_value (object_p);
+
+    if (JMEM_CP_GET_THIRD_BIT_FROM_POINTER_TAG (object_with_tag))
+    {
+      ecma_enqueue_promise_async_reaction_job (object, value, is_reject);
+      continue;
+    }
 
     if (!is_reject)
     {
       ecma_value_t handler = ECMA_VALUE_TRUE;
 
-      if (JMEM_CP_GET_FIRST_BIT_FROM_POINTER_TAG (capability_with_tag))
+      if (JMEM_CP_GET_FIRST_BIT_FROM_POINTER_TAG (object_with_tag))
       {
         handler = *buffer_p++;
       }
 
-      ecma_enqueue_promise_reaction_job (capability, handler, value);
+      ecma_enqueue_promise_reaction_job (object, handler, value);
     }
-    else if (JMEM_CP_GET_FIRST_BIT_FROM_POINTER_TAG (capability_with_tag))
+    else if (JMEM_CP_GET_FIRST_BIT_FROM_POINTER_TAG (object_with_tag))
     {
       buffer_p++;
     }
@@ -147,14 +153,14 @@ ecma_promise_trigger_reactions (ecma_collection_t *reactions, /**< lists of reac
     {
       ecma_value_t handler = ECMA_VALUE_FALSE;
 
-      if (JMEM_CP_GET_SECOND_BIT_FROM_POINTER_TAG (capability_with_tag))
+      if (JMEM_CP_GET_SECOND_BIT_FROM_POINTER_TAG (object_with_tag))
       {
         handler = *buffer_p++;
       }
 
-      ecma_enqueue_promise_reaction_job (capability, handler, value);
+      ecma_enqueue_promise_reaction_job (object, handler, value);
     }
-    else if (JMEM_CP_GET_SECOND_BIT_FROM_POINTER_TAG (capability_with_tag))
+    else if (JMEM_CP_GET_SECOND_BIT_FROM_POINTER_TAG (object_with_tag))
     {
       buffer_p++;
     }
@@ -198,7 +204,7 @@ ecma_is_resolver_already_called (ecma_object_t *resolver_p, /**< resolver */
  *
  * See also: ES2015 25.4.1.7
  */
-static void
+void
 ecma_reject_promise (ecma_value_t promise, /**< promise */
                      ecma_value_t reason) /**< reason for reject */
 {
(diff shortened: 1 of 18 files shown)


jerry-core/vm/vm.c
```diff
@@ -2188,7 +2192,52 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+            result = ecma_op_create_promise_object (ECMA_VALUE_EMPTY, ECMA_PROMISE_EXECUTOR_EMPTY);
+
+            JERRY_ASSERT (ecma_is_value_object (result));
+            executable_object_p->frame_ctx.block_result = result;
+
+            JERRY_CONTEXT (current_new_target) = old_new_target_p;
+          }
+          else
+          {
+            uintptr_t diff = (uintptr_t) (&((vm_executable_object_t *) 0x100)->frame_ctx) - 0x100;
```
galpeter, line 2231: What does the `0x100` represents here?

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -963,6 +956,31 @@ ecma_promise_then (ecma_value_t promise, /**< the promise which call 'then' */
   return ret;
 } /* ecma_promise_then */
 
+/**
+ * Resome the execution of an async function after the promise is resolved
```
rerobika, line 960: typo: `Resome`

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -227,14 +233,47 @@ ecma_reject_promise (ecma_value_t promise, /**< promise */
...
   ecma_object_t *obj_p = ecma_get_object_from_value (promise);
 
   JERRY_ASSERT (ecma_promise_get_flags (obj_p) & ECMA_PROMISE_IS_PENDING);
 
+  if (promise == value)
+  {
+    ecma_value_t exception = ecma_raise_type_error (ECMA_ERR_MSG ("A promise cannot be resolved with itself."));
+    JERRY_ASSERT (ECMA_IS_VALUE_ERROR (exception));
+
+    exception = jcontext_take_exception ();
```
rerobika, line 249: How about eliminating this double assignment?

```c
    ecma_raise_type_error (ECMA_ERR_MSG ("A promise cannot be resolved with itself."));
    ecma_value_t exception = jcontext_take_exception ();
```

jerry-core/vm/opcodes.h
```diff
@@ -54,6 +54,19 @@ typedef enum
...
+#if ENABLED (JERRY_ESNEXT)
+
+/**
+ * Types for opfunc_create_executable_object.
+ */
+typedef enum
+{
+  CREATE_EXECUTABLE_OBJECT_GENERATOR, /**< create a generator function */
+  CREATE_EXECUTABLE_OBJECT_ASYNC, /**< create an async function */
+} create_executable_object_type_t;
```
rerobika, line 66: How about `ecma_executable_object_type_t` instead? The `create` prefix is quite strange for me.

jerry-core/vm/opcodes.c
```diff
@@ -575,43 +575,54 @@ opfunc_append_array (ecma_value_t *stack_top_p, /**< current stack top */
...
-  ecma_deref_object (proto_p);
-
   vm_executable_object_t *executable_object_p = (vm_executable_object_t *) object_p;
 
+  if (type == CREATE_EXECUTABLE_OBJECT_GENERATOR)
+  {
+    ecma_deref_object (proto_p);
+  }
+
+  /* Async function objects are not accessible, so their class_id is not relevant. */
```
rerobika, line 621: That's definitely not correct:

```js
async function f () {}
Object.prototype.toString.call(f)
>> "[object AsyncFunction]"
```

jerry-core/vm/opcodes.h
zherczeg: It is the name of the function which uses it. It is not an object type. ecma_create_executable_object_type_t could be, but that is quite long

jerry-core/vm/opcodes.c
zherczeg: That is a different thing. Currently the prototype objects defined by the standard (e.g. AsyncFunction) has not been implemented yet , this patch focuses on the core things.

When 'f' is called, it creates another hidden object with the execution state, and this object is never visible. It does not even need to be an object, but gc likes this. It could be dropped to reduce memory consumption at some point.

rerobika: Thanks for the explanation!

jerry-core/ecma/operations/ecma-promise-object.c
zherczeg: I did it for the assert. If it bothers you, I can remove it.

rerobika: I'd support removing it, otherwise the patch looks good to me.

jerry-core/vm/opcodes.c
```diff
@@ -716,14 +762,15 @@ opfunc_resume_executable_object (vm_executable_object_t *executable_object_p, /*
...
 
   register_p = VM_GET_REGISTERS (&executable_object_p->frame_ctx);
-
-  while (register_p < register_end_p)
-  {
-    ecma_deref_if_object (*register_p++);
-  }
+  stack_top_p = executable_object_p->frame_ctx.stack_top_p;
 
   if (executable_object_p->frame_ctx.context_depth > 0)
```
galpeter, line 767: This `if` and  the `while` after it seems to me used in this form multiple times (with seemingly minimal changes). Would it be possible to extract them into a method?

jerry-core/vm/vm.c
```diff
@@ -2194,7 +2192,52 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+            result = ecma_op_create_promise_object (ECMA_VALUE_EMPTY, ECMA_PROMISE_EXECUTOR_EMPTY);
+
+            JERRY_ASSERT (ecma_is_value_object (result));
+            executable_object_p->frame_ctx.block_result = result;
+
+            JERRY_CONTEXT (current_new_target) = old_new_target_p;
+          }
+          else
+          {
+            uintptr_t diff = (uintptr_t) offsetof (vm_executable_object_t, frame_ctx);
```
galpeter, line 2231: Maybe a better name for the `diff` would be `object_offset` and it can be a const value.

jerry-core/vm/vm.c
```diff
@@ -2194,7 +2192,52 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
-          continue;
+          ecma_value_t promise = ecma_make_object_value (ecma_builtin_get (ECMA_BUILTIN_ID_PROMISE));
+          ecma_value_t argument = *(--stack_top_p);
+
+          result = ecma_promise_reject_or_resolve (promise, argument, true);
+          ecma_free_value (argument);
+
+          if (ECMA_IS_VALUE_ERROR (result))
+          {
+            goto error;
```
galpeter, line 2203: Do we have a test case for this error?

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -227,14 +233,45 @@ ecma_reject_promise (ecma_value_t promise, /**< promise */
...
 ecma_fulfill_promise (ecma_value_t promise, /**< promise */
                       ecma_value_t value) /**< fulfilled value */
 {
   ecma_object_t *obj_p = ecma_get_object_from_value (promise);
 
   JERRY_ASSERT (ecma_promise_get_flags (obj_p) & ECMA_PROMISE_IS_PENDING);
 
+  if (promise == value)
+  {
+    ecma_raise_type_error (ECMA_ERR_MSG ("A promise cannot be resolved with itself."));
```
galpeter, line 246: Do we have a test for this?

jerry-core/vm/vm-stack.c
```diff
@@ -310,21 +324,17 @@ vm_get_context_value_offsets (ecma_value_t *context_item_p) /**< any item of a c
     {
       return PARSER_TRY_CONTEXT_STACK_ALLOCATION;
     }
-#if ENABLED (JERRY_ESNEXT)
```
galpeter, line 313: Why do we need to remove these?

zherczeg: The entire function is guarded by this, there is no need extra guards inside.

jerry-core/vm/vm.c
zherczeg: (note: I changed this to const although the constant folding in the compiler does this automatically)

jerry-core/vm/opcodes.c
zherczeg: This is only used twice, and I would not combine them because they do opposite things.

jerry-core/vm/vm.c
zherczeg: Test added.

jerry-core/ecma/operations/ecma-promise-object.c
zherczeg: Test added.

zherczeg:
Thank you for the reviews. I hope I did all requests, please check the patch again.

jerry-core/vm/opcodes.c
galpeter: ohh I see one is `ref` and the other is `deref`. Sorry for the noise I've missed this.

PR #3899, opened 2020-06-12
Remove duplicate is_object check when calling ecma_op_to_primtive

szilagyiadam:

### DIFF
--- a/jerry-core/ecma/operations/ecma-conversion.c
+++ b/jerry-core/ecma/operations/ecma-conversion.c
@@ -287,39 +287,37 @@ ecma_op_to_number (ecma_value_t value) /**< ecma value */
   }
 #endif /* ENABLED (JERRY_ESNEXT) */
 
-  if (ecma_is_value_object (value))
-  {
-    ecma_value_t primitive_value = ecma_op_to_primitive (value, ECMA_PREFERRED_TYPE_NUMBER);
-
-    if (ECMA_IS_VALUE_ERROR (primitive_value))
-    {
-      return primitive_value;
-    }
-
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -287,39 +287,33 @@ ecma_op_to_number (ecma_value_t value) /**< ecma value */
...
   {
-    num = 0;
+    return ecma_make_integer_value (0);
   }
-  else
+
+  if (ecma_is_value_boolean (value))
   {
-    JERRY_ASSERT (ecma_is_value_boolean (value));
+    return ecma_is_value_true (value) ? ecma_make_integer_value (1) : ecma_make_integer_value (0);
```
rerobika, line 302: `ecma_make_integer_value (ecma_is_value_true (value) ? 1 : 0)`

jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -463,14 +447,29 @@ ecma_op_to_string (ecma_value_t value) /**< ecma value */
...
   }
 #endif /* ENABLED (JERRY_ES2015) */
-  JERRY_ASSERT (ecma_is_value_boolean (value));
-
-  if (ecma_is_value_true (value))
+  else if (ecma_is_value_true (value))
   {
     return ecma_get_magic_string (LIT_MAGIC_STRING_TRUE);
   }
+  else if (ecma_is_value_false (value))
```
rerobika, line 454: Since each condition returns a with a value here and above I'd suggest to use only `if`s instead of `else if`s.

PR #3905, opened 2020-06-16
Extend CI list with Github actions

rerobika:

### DIFF
--- a/.github/workflows/gh-actions.yml
+++ b/.github/workflows/gh-actions.yml
@@ -1,0 +1,109 @@
+name: JerryScript CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+env:
+  RUNNER: tools/run-tests.py
+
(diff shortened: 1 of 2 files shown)


rerobika:
Note: the GitHub actions will only trigger after the PR got merged. The local results for this PR is available [here](https://github.com/rerobika/jerryscript/actions/runs/137367796 ).

ossy-szeged:
The results are promising. I think we should split at least the ARM and OSX tests to have more parallelizable jobs and smaller runtime.

rerobika:
@ossy-szeged The longer runtimes on these jobs are due to the currently separated jerry-test-suite, which is going to be merged into the jerry-tests soon. I'd wait until these merge, after we can rebalance these jobs if needed.

rerobika:
Needs to be updated after #3907 .

rerobika:
Updated results: https://github.com/rerobika/jerryscript/actions/runs/138332968

Issue #3737, opened 2020-05-13
Re-target JerryScript for ES.Next


akosthekiss:
Traditionally, JerryScript has been a fully ES5.1-compatible JavaScript engine. As the standard evolved, the community started adding features from the next edition of EcmaScript (ES6 or ES2015) into a new profile (configuration) of the engine. The original configuration was called the es5.1 profile, while the new one got the name es2015-subset profile. The original idea was to complete the ES2015 profile -- so as to be able to drop the "-subset" suffix -- and then move on to the next edition of the standard, and so on. I.e., the intent was to support all editions of the standard and allow our users to select the edition they need when they configure and build JerryScript.

It turns out that this is not a maintainable approach and does not even meet the needs of the users. Most JavaScript programmers use the latest edition of the standard, whichever that is, and expect engines to support that edition. Or, at least, support something close to that. Even if we finished the exact implementation of the ES6 specification, that would be both outdated and not even the subset of the latest edition of JavaScript. (Yes, there have been backward incompatible changes in the spec since 2015.) And we definitely wont be able to maintain different ES2015, ES2016, ES2017, ES2018, ES2019 (etc.) configurations. (Moreover, we couldn't even properly test conformance, as the test262 conformance suite has ceased releasing test suites for the different editions of the language after ES5.1.)

Therefore, I propose to keep the original ES5.1 profile (because that's a fully spec-compatible and mature configuration, often enough for constrained applications), but otherwise target the always-latest version of the EcmaScript standard (let's call it the ES.Next profile -- disclaimer: I haven't invented the term).

Notes:
- This would cause a big one-time churn in the code base, as all ES2015 macro guards would need a rename (to something like ESNEXT?). Most probably it's not even only the macros.
- This would cause some backward incompatibility as well, at least build system-wise, as profile name and feature guard names would change. Embedders relying on these would break.
- IMO, we still could and should keep the feature guards in the code base to allow users to trim down functionality. Let's remember that JerryScript is for constrained devices.

zherczeg:
I strongly support this idea. It is extremely hard to support multiple variants of the standard, especially at parser side.

bil-ash:
In case jerryscript moves on the path suggested by @akosthekiss please do make a final (github)release with ES2015-subset(including all the improvements since the last release) before moving on to ES.Next .

akosthekiss:
[...]

PR #3906, opened 2020-06-17
Remove arguments and caller properties from ArrowFunction

rwalczyna:
As stated in [ES11 16.1](http://www.ecma-international.org/ecma-262/11.0/#sec-forbidden-extensions), ArrowFunction should not have caller and arguments properties.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -1525,10 +1525,11 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
     bytecode_data_p = ecma_op_function_get_compiled_code ((ecma_extended_object_t *) object_p);
 
 #if ENABLED (JERRY_ESNEXT)
-    if (!(bytecode_data_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE))
+    if (!(bytecode_data_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE)
+        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_ARROW))
     {
       ecma_property_t *value_prop_p;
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1525,7 +1525,8 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
     bytecode_data_p = ecma_op_function_get_compiled_code ((ecma_extended_object_t *) object_p);
 
 #if ENABLED (JERRY_ESNEXT)
-    if (!(bytecode_data_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE))
+    if (!(bytecode_data_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE)
+        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_ARROW))
     {
       ecma_property_t *value_prop_p;
       /* The property_name_p argument contans the name. */
```
ossy-szeged, line 1532: If you already touched this code, could you please fix this typo too? contans -> contains

akosthekiss:
@rwalczyna Please, note that JerryScript has moved to ES.Next (#3737, #3901). I.e., the project is aiming at the always-latest, or living standard version of ECMAScript. What's stated in ES6 _may_ be valid in the latest version, but not necessarily. There have been backward-incompatible changes in corner cases occasionally. As of now, references shall be made to ES11/ES2020 instead of ES6.

rwalczyna:
@akosthekiss, this change is also valid in ES11, should I update the commit message, to reference to ES11 instead of ES6?

akosthekiss:
@rwalczyna Yes, I'd suggest updating the reference in the commit message. IMO we should stress this and keep referring to the latest ES standard version everywhere so that this change sinks in all community members.

PR #3911, opened 2020-06-17
Refactor Number.prototype methods toFixed, toExponential, toPrecision

szilagyiadam:


jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -651,89 +613,106 @@ static ecma_value_t
...
 
-  ecma_value_t ret_value;
-  JMEM_DEFINE_LOCAL_ARRAY (buff, buffer_size, lit_utf8_byte_t);
+  /* Get the parameters of the number */
+  lit_utf8_byte_t digits[ECMA_MAX_CHARS_IN_STRINGIFIED_NUMBER];
+  lit_utf8_size_t num_of_digits;
+  int32_t exponent;
 
-  lit_utf8_byte_t *p = buff;
+  num_of_digits = ecma_number_to_binary_floating_point_number (this_num, digits, &exponent);
```
dbatyai, line 683: I think this function should be removed as well. The other functions all use `ecma_number_to_decimal` to convert the number to digits, we should use that here as well.

The names of some of the variables should also be revised, they seem to store different values than what they imply. For example currently calling `ecma_number_to_binary_floating_point_number ` on the number `5` will result in `num_digits = 0, digits = "5", exponenet = 1`, which makes no sense at all.

After switching to `ecma_number_to_decimal` we should double check the variables and name them appropriately. Maybe even `ecma_number_to_decimal` should be renamed/updated a bit so that it's purpose and the meaning of the returned values are clearer.

szilagyiadam:
Updated the PR!
Also refactored toExponential and toPrecision, now every method uses one common method.
Note: I don't have a solution for the toFixed method where i can use ecma_number_to_decimal yet.

szilagyiadam:
> Updated the PR!
> Also refactored toExponential and toPrecision, now every method uses one common method.
> Note: I don't have a solution for the toFixed method where i can use ecma_number_to_decimal yet.

Did an update so i use ecma_number_to_decimal in all of the three methods

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -574,266 +472,301 @@ typedef enum
...
-  }
-  else
+  /* Handle undefined argument */
+  if (ecma_is_value_undefined (arg))
   {
-    for (int32_t i = 0; i <= frac_digits; i++)
+    if (mode == NUMBER_ROUTINE_TO_FIXED)
     {
-      digits[i] = '0';
+      arg_int = 0;
```
dbatyai, line 575: If the argument was undefined then this should already be 0.

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -574,266 +472,301 @@ typedef enum
...
+  ecma_value_t to_integer = ecma_op_to_integer (arg, &arg_num);
 
-  if (ECMA_IS_VALUE_ERROR (arg_1))
+  if (ECMA_IS_VALUE_ERROR (to_integer))
   {
-    return arg_1;
+    return to_integer;
   }
 
+  /* Argument boundary checks */
```
dbatyai, line 496: These ranges have changed in the recent standard, please update them.

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -574,266 +472,301 @@ typedef enum
...
+                                                              num_of_digits,
+                                                              digits_to_keep,
+                                                              &exponent,
+                                                              false);
 
-  if (is_negative)
+  /* toExponent routine and toPrecision cases where the exponent > precision or exponent < -5 */
+  if (mode == NUMBER_ROUTINE_TO_EXPONENTIAL
+      || (mode == NUMBER_ROUTINE_TO_PRECISION
+          && (exponent < -5 || exponent > arg_int)))
```
dbatyai, line 610: This range seems to be different in the standard. Please add tests that cover the edge cases and make sure they are correct.

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -849,120 +782,9 @@ static ecma_value_t
...
-
-  JERRY_ASSERT (actual_char_p - buff < buffer_size);
-  *actual_char_p = '\0';
-  ecma_string_t *str_p = ecma_new_ecma_string_from_utf8 (buff, (lit_utf8_size_t) (actual_char_p - buff));
-
-  ret_value = ecma_make_string_value (str_p);
-  JMEM_FINALIZE_LOCAL_ARRAY (buff);
-
-  return ret_value;
+  return ecma_builtin_number_prototype_object_to_number_convert (this_num,
```
dbatyai, line 785: We could call these directly from the dispatcher.

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -574,266 +472,301 @@ typedef enum
...
+  num_of_digits = ecma_number_to_decimal (this_num, digits, &exponent);
 
-  if (!ecma_number_is_zero (this_num))
-  {
-    num_digits = ecma_number_to_binary_floating_point_number (this_num, digits, &exponent);
-    JERRY_ASSERT (exponent >= 0);
-  }
-  else
+  /* Handle undefined argument */
+  if (ecma_is_value_undefined (arg))
```
dbatyai, line 571: The required actions when the argument is undefined also changed slightly n the standard, please double check them.

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -574,396 +472,262 @@ typedef enum
...
-    exponent = 1;
+    arg_int = (int32_t) num_of_digits - 1;
   }
 
-  if (ecma_is_value_undefined (fraction_digits))
+  if (mode == NUMBER_ROUTINE_TO_FIXED
+      && exponent > 21)
   {
-    frac_digits = (int32_t) num_digits - 1;
+    ecma_value_t result_num = ecma_builtin_number_prototype_object_to_string (this_num, NULL, 0);
```
dbatyai, line 579: How about destroying the builder first, and returning with the result of the to_string call?

jerry-core/ecma/builtin-objects/ecma-builtin-number-prototype.c
```diff
@@ -1007,15 +771,21 @@ ecma_builtin_number_prototype_dispatch_routine (uint16_t builtin_routine_id, /**
...
+                                                                     NUMBER_ROUTINE_TO_FIXED);
     }
     case ECMA_NUMBER_PROTOTYPE_TO_EXPONENTIAL:
     {
-      return ecma_builtin_number_prototype_object_to_exponential (this_arg_number, routine_arg_1);
+      return ecma_builtin_number_prototype_object_to_number_convert (this_arg_number,
+                                                                     routine_arg_1,
+                                                                     NUMBER_ROUTINE_TO_EXPONENTIAL);
     }
     case ECMA_NUMBER_PROTOTYPE_TO_PRECISION:
```
rerobika, line 784: Please merge these 3 cases. the number routine mode can be calculated like this:

NUMBER_ROUTINE_TO_FIXED + (builtin_id - ECMA_NUMBER_PROTOTYPE_TO_FIXED);

PR #3912, opened 2020-06-17
Run test262 5.1 and es2015 in both release and debug mode

rerobika:
JerryScript-DCO-1.0-Signed-off-by: Robert Fancsik frobert@inf.u-szeged.hu

PR #3913, opened 2020-06-17
Fix releasing 'exec' in regexp replace fast path

rerobika:


rerobika:
Related regression: https://github.com/jerryscript-project/jerryscript/pull/3912/checks?check_run_id=781089438#step:3:1192

PR #3916, opened 2020-06-19
Implement async generators.

zherczeg:
yield* is not supported yet


jerry-core/vm/vm.c
```diff
@@ -2239,6 +2275,30 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+          ecma_free_value (argument);
+
+          if (ECMA_IS_VALUE_ERROR (result))
+          {
+            goto error;
+          }
+
+          frame_ctx_p->call_operation = VM_EXEC_RETURN;
+          frame_ctx_p->byte_code_p = byte_code_p;
+          frame_ctx_p->stack_top_p = stack_top_p;
```
zherczeg, line 2293: Up to this point maybe this could be merged with VM_OC_AWAIT

jerry-core/parser/js/js-parser.c
```diff
@@ -1350,31 +1355,42 @@ parser_post_processing (parser_context_t *context_p) /**< context */
...
-    compiled_code_p->status_flags |= CBC_FUNCTION_TO_TYPE_BITS (CBC_FUNCTION_CONSTRUCTOR);
+    function_type = CBC_FUNCTION_TO_TYPE_BITS (CBC_FUNCTION_CONSTRUCTOR);
   }
   else
   {
-    compiled_code_p->status_flags |= CBC_FUNCTION_TO_TYPE_BITS (CBC_FUNCTION_NORMAL);
+    function_type = CBC_FUNCTION_TO_TYPE_BITS (CBC_FUNCTION_NORMAL);
   }
 
+  compiled_code_p->status_flags |= function_type;
```
rerobika, line 1392: How about using `CBC_FUNCTION_TO_TYPE_BITS` only here?

zherczeg: I have been thinking about this, but that would mean an actual runtime computation (extra CPU instructions for shifting). The current method puts a final constant directly into a variable. The compiler might be able optimize the unnecessary shift, but I am not sure about it.

jerry-core/ecma/builtin-objects/ecma-builtin-async-generator-prototype.c
```diff
@@ -0,0 +1,141 @@
...
+      return ecma_promise_reject_or_resolve (promise, value, false);
+    }
+
+    ecma_value_t iter_result = ecma_create_iter_result_object (ECMA_VALUE_UNDEFINED, ECMA_VALUE_TRUE);
+    ecma_value_t result = ecma_promise_reject_or_resolve (promise, iter_result, true);
+    ecma_free_value (iter_result);
+    return result;
+  }
+
+  ecma_async_generator_command_type_t command = ECMA_ASYNC_GENERATOR_COMMAND_NEXT;
```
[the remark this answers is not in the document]
zherczeg: I will check this, but I am not sure we gain much here. Maybe some binary size.

jerry-core/ecma/builtin-objects/ecma-builtin-async-generator-prototype.c
```diff
@@ -0,0 +1,141 @@
...
+{
+  JERRY_UNUSED (arguments_number);
+
+  vm_executable_object_t *executable_object_p = NULL;
+
+  if (ecma_is_value_object (this_arg))
+  {
+    ecma_object_t *object_p = ecma_get_object_from_value (this_arg);
+
+    if (ecma_get_object_type (object_p) == ECMA_OBJECT_TYPE_CLASS)
```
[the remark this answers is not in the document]
zherczeg: That would duplicate the `ecma_is_value_object` above. This code is similar to the code in normal generators.

jerry-core/ecma/builtin-objects/ecma-builtin-async-generator-prototype.c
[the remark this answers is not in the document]
zherczeg: Done.

jerry-core/parser/js/js-parser.c
rerobika: I think this is not a 'hot' code path and the suggestion results slightly smaller binary size. So I'd still support the suggested changes.

rerobika: I've realizes that these are constants. Just ignore my comments.

zherczeg:
Patch updated, please check it again.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -127,13 +127,19 @@ typedef enum
...
  * Set an internal property value from pointer. Pointer can be NULL.
  */
 #define ECMA_SET_INTERNAL_VALUE_ANY_POINTER(field, pointer) \
-  (field) = ((ecma_value_t) pointer)
+  ((field) = ((ecma_value_t) pointer))
+
+/**
+ * Set an internal property to NULL.
+ */
+#define ECMA_SET_INTERNAL_VALUE_TO_NULL(field) \
```
dbatyai, line 141: Do we really need to add new macros for this? Can't we just use `ECMA_SET_INTERNAL_VALUE_ANY_POINTER`?

zherczeg: Inefficient, and the code does not look good. Eg.:
```
while (IS_NOT_NULL(something)) {
  something = DECODE(something)->next
}
```
Looks like this:
```
while (true) {
  something_p = DECODE_ANY(something)
  // NULL checked twice
  if (something_p == NULL) {
   break;
  }
  something = something_p->next
}
```

dbatyai: The question was regarding `ECMA_SET_INTERNAL_VALUE_TO_NULL`, not `ECMA_IS_INTERNAL_VALUE_NULL`.

dbatyai: As a matter of fact, couldn't we use compressed pointers directly instead of internal values? Do we need tag flags for something?

zherczeg: When compression does not needed pointers can be stored directly, so these macros are preferrable (they were invented for this purpose).

I am sorry I misunderstand your question. `ECMA_SET_INTERNAL_VALUE_ANY_POINTER` could be used, although I think these are just macros, and they should not hurt, but I can remove them if you want.

dbatyai: I would prefer to not add specialized macros when we can use a more general one we already have.

zherczeg:
I did the updates, please check it again.

PR #3918, opened 2020-06-19
Remove "arguments" and "caller" from specific functions

rwalczyna:
As stated in [ES 16.1](http://www.ecma-international.org/ecma-262/11.0/#sec-forbidden-extensions), "arguments" and "caller" must not be created in:
- strict functions created using the Function constructor
- generator functions created using the Generator constructor
- async functions created using the AsyncFunction constructor
- functions created using the bind

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -1531,7 +1531,7 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
 
 #if ENABLED (JERRY_ESNEXT)
     if (!(bytecode_data_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE)
-        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_ARROW))
+        && CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_NORMAL)
     {
       ecma_property_t *value_prop_p;
       /* The property_name_p argument contains the name. */
@@ -1660,6 +1660,7 @@ ecma_op_bound_function_try_to_lazy_instantiate_property (ecma_object_t *object_p
     return len_prop_p;
   }
 
+#if !ENABLED (JERRY_ESNEXT)
   if (ecma_compare_ecma_string_to_magic_id (property_name_p, LIT_MAGIC_STRING_CALLER)
       || ecma_compare_ecma_string_to_magic_id (property_name_p, LIT_MAGIC_STRING_ARGUMENTS))
   {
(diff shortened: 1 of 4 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1531,7 +1531,9 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
 
 #if ENABLED (JERRY_ESNEXT)
     if (!(bytecode_data_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE)
-        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_ARROW))
+        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_ARROW)
+        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_ASYNC)
+        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_GENERATOR))
```
zherczeg, line 1536: I would put `CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags)` into a local variable.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1660,22 +1662,6 @@ ecma_op_bound_function_try_to_lazy_instantiate_property (ecma_object_t *object_p
     return len_prop_p;
   }
 
-  if (ecma_compare_ecma_string_to_magic_id (property_name_p, LIT_MAGIC_STRING_CALLER)
```
rerobika, line 1663: IMO, this should be kept for the ES5.1 profile.

rwalczyna:
Thanks for reviews, I have updated the commit.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1530,8 +1530,12 @@ ecma_op_function_try_to_lazy_instantiate_property (ecma_object_t *object_p, /**<
     bytecode_data_p = ecma_op_function_get_compiled_code ((ecma_extended_object_t *) object_p);
 
 #if ENABLED (JERRY_ESNEXT)
+    uint16_t function_type = CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags);
+
     if (!(bytecode_data_p->status_flags & CBC_CODE_FLAGS_STRICT_MODE)
-        && !(CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_ARROW))
+        && !(function_type == CBC_FUNCTION_ARROW)
+        && !(function_type == CBC_FUNCTION_ASYNC)
+        && !(function_type == CBC_FUNCTION_GENERATOR))
```
zherczeg, line 1538: I checked accessors and constructor functions and it seems they don't have these properties as well. Is there any other function type which has these properties? Because if only normal functions have these properties, we can change the check to:
`CBC_FUNCTION_GET_TYPE (bytecode_data_p->status_flags) == CBC_FUNCTION_NORMAL`

rwalczyna: I think you are right. I checked the newest documentation and it looks like currently only normal function has these properties.

PR #3919, opened 2020-06-19
Fix a few small issues in generators.

zherczeg:


zherczeg:
built-ins/Array/prototype/forEach/15.4.4.18-7-b-10.js this fails, although this test has nothing to do with generators. any idea?

rerobika:
I've restarted the CI since the PR is not related to this testfile, and the error has gone.

PR #3923, opened 2020-06-23
Implement yield* operator

zherczeg:
Missing features:
- caching next() method (this also true for normal generators)
- automatic sync to async generator conversion


jerry-core/ecma/operations/ecma-async-generator-object.c
```diff
@@ -132,8 +293,6 @@ ecma_async_generator_finalize (vm_executable_object_t *async_generator_object_p,
   ecma_value_t next = async_generator_object_p->extended_object.u.class_prop.u.head;
   ecma_async_generator_task_t *task_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_async_generator_task_t, next);
 
-  ECMA_SET_INTERNAL_VALUE_ANY_POINTER (async_generator_object_p->extended_object.u.class_prop.u.head, NULL);
```
zherczeg, line 135: Note: setting head to NULL caused memory errors, since resolving promises may trigger gc, and promises referenced by tasks were not marked.

jerry-core/ecma/operations/ecma-async-generator-object.c
```diff
@@ -169,10 +329,129 @@ ecma_async_generator_finalize (vm_executable_object_t *async_generator_object_p,
...
+      ecma_value_t result = ecma_async_yield_throw (async_generator_object_p, value);
+      ecma_free_value (value);
+      return result;
+    }
+    default:
+    {
+      JERRY_ASSERT (ECMA_ASYNC_YIELD_ITERATOR_GET_STATE (async_generator_object_p)
+                    == ECMA_ASYNC_YIELD_ITERATOR_AWAIT_CLOSE);
+
+      if (!ecma_is_value_object (value))
```
rerobika, line 443: How about saving this into a boolean? Therefore the double free/raise error calls can be reduced into single ones.

jerry-core/vm/vm.c
```diff
@@ -2201,29 +2201,52 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+          frame_ctx_p->call_operation = VM_EXEC_RETURN;
+          frame_ctx_p->byte_code_p = byte_code_p;
+          frame_ctx_p->stack_top_p = --stack_top_p;
+          return ECMA_VALUE_UNDEFINED;
+        }
+        case VM_OC_ASYNC_YIELD_ITERATOR:
+        {
+          const uintptr_t object_offset = (uintptr_t) offsetof (vm_executable_object_t, frame_ctx);
+          ecma_extended_object_t *async_generator_object_p;
+          async_generator_object_p = (ecma_extended_object_t *) (((uintptr_t) frame_ctx_p) - object_offset);
```
rerobika, line 2215: Shouldn't worth to define a macro for these?

zherczeg:
Thnak you for the review, patch is updated.

PR #3928, opened 2020-06-24
Support Unicode supplementary planes

rerobika:


rerobika:
Some micro benchmark results:

Test:

```js
let start = 0x0
let end = 0x10000

let str = "";

for (let i = start; i < end; i++) {
  str += String.fromCodePoint(i);
}

str = str.toUpperCase();
str.toLowerCase();
```

Master:
 - Peak allocated = 584736 bytes
 - real    0m0,622s

PR:
 - Peak allocated = **389768** bytes
 - real    **0m0,584s**

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c
```diff
@@ -1001,96 +1001,41 @@ ecma_builtin_string_prototype_object_conversion_helper (ecma_string_t *input_str
...
-    }
-
-    JERRY_ASSERT (character_length >= 1 && character_length <= LIT_MAXIMUM_OTHER_CASE_LENGTH);
-
-    for (ecma_length_t i = 0; i < character_length; i++)
+#if ENABLED (JERRY_ESNEXT)
+    if (lit_is_code_point_utf16_high_surrogate (cp))
     {
-      output_length += lit_code_unit_to_utf8 (character_buffer[i], utf8_byte_buffer);
+      const ecma_char_t next_ch = lit_cesu8_read_next (&input_curr_p);
```
dbatyai, line 1018: Won't this skip a code unit if `next_ch` is not a low surrogate?

jerry-core/lit/lit-char-helpers.c
```diff
@@ -166,13 +183,30 @@ lit_char_is_line_terminator (ecma_char_t c) /**< code unit */
...
+    return (lit_search_codepoint_in_interval_array (code_point,
+                                                    lit_unicode_letter_interval_sps_sup,
+                                                    lit_unicode_letter_interval_lengths_sup,
+                                                    NUM_OF_ELEMENTS (lit_unicode_letter_interval_sps_sup))
+            || lit_search_codepoint_in_array (code_point,
+                                              lit_unicode_letter_chars_sup,
+                                              NUM_OF_ELEMENTS (lit_unicode_letter_chars_sup)));
+  }
+#else /* !ENABLED (JERRY_ESNEXT) */
+  JERRY_ASSERT (code_point <= LIT_UTF8_4_BYTE_CODE_POINT_MIN);
```
dbatyai, line 200: Shouldn't this be strictly less?

jerry-core/lit/lit-char-helpers.c
```diff
@@ -190,15 +224,33 @@ lit_char_is_unicode_letter (ecma_char_t c) /**< code unit */
...
+    return (lit_search_codepoint_in_interval_array (code_point,
+                                                    lit_unicode_non_letter_ident_part_interval_sps_sup,
+                                                    lit_unicode_non_letter_ident_part_interval_lengths_sup,
+                                                    non_letter_ident_part_length)
+            || lit_search_codepoint_in_array (code_point,
+                                              lit_unicode_non_letter_ident_part_chars_sup,
+                                              NUM_OF_ELEMENTS (lit_unicode_non_letter_ident_part_chars_sup)));
+  }
+#else /* !ENABLED (JERRY_ESNEXT) */
+  JERRY_ASSERT (code_point <= LIT_UTF8_4_BYTE_CODE_POINT_MIN);
```
dbatyai, line 242: Same here.

rerobika:
@dbatyai Thanks for the review, the PR is updated.

tools/gen-unicode.py
```diff
@@ -186,25 +228,57 @@ def split_list(category_list):
...
+                                     "uint32_t",
+                                     "uint32_t",
+                                     "uint16_t",
+                                     "uint32_t")}
+
     c_source.add_table(letter_tables[0],
                        "unicode_letter_interval_sps",
-                       "uint16_t",
+                       suffix,
+                       array_types[plane_type][0],
```
zherczeg, line 281: This code repeated several times. Cannot we build these tables with a for loop with a list of tuples?

tools/gen-unicode.py
```diff
@@ -371,8 +460,12 @@ def read_case_mappings(unicode_data_file, special_casing_file):
             letter_id = int(line[0], 16)
             condition_list = line[4]
 
-            # Skip supplementary planes, ascii chars, and condition_list
-            if letter_id >= 0x10000 or letter_id < 128 or condition_list:
+            # Skip non-supplementary planes
+            if plane_type == 'supplementary':
```
zherczeg, line 464: It is strange for me to see strings for types. Could we use enum constants similar to debugger?

tools/gen-unicode.py
```diff
@@ -21,275 +21,264 @@
...
+                interval_sps.append(element[0])
+                interval_lengths.append(element[1] - element[0])
 
-                # Skip supplementary planes and ascii chars
-                if unicode_id >= 0x10000 or unicode_id < 128:
-                    continue
+        return interval_sps, interval_lengths, chars
 
-                category = line[2]
+    def read_units(self, file_path, category):
```
galpeter, line 162: Maybe we could reorganize this a bit to avoid multiple passes on the same input file. What do you thing about this?: 

```python
def read_units(self, file_path, categories):
    """
    Read the Unicode Derived Core Properties file and extract the ranges
    for the given categories.

    :param file_path: Path to the Unicode "DerivedCoreProperties.txt" file.
    :param categories: A list of category strings to extract from the Unicode file.
    :return: A dictionary each string from the :param categories: is a key and for each
             key list of code points are stored.
    """
    # Create a dictionary in the format: { category[0]: [ ], ..., category[N]: [ ] } 
    units = dict(zip(categories, [[]] * len(categories)))

    with open(file_path, "r") as fp:
        for line in fp:
            # Formats to match:
            #  <HEX>     ; <category> #
            #  <HEX>..<HEX>     ; <category> #
            match = re.match(r"(?P<start>[\dA-F]+)(?:\.\.(?P<end>[\dA-F]+))?\s+; (?P<category>[\w]+) #", line)

            if match and match.group("category") in categories:
                start = int(match.group("start"), 16)
                # if no "end" found use the "start"
                end = int(match.group("end") or match.group("start"), 16)

                matching_code_points = [
                    code_point for code_point in range(start, end + 1) if self.in_range(code_point)
                ]
                units[match.group("category")].extend(matching_code_points)

    return units
```

With this the required categories can be read in a single pass over the input file

tools/gen-unicode.py
```diff
@@ -783,29 +677,33 @@ def generate_conversions(script_args):
 
 
 def main():
-    parser = argparse.ArgumentParser(description='lit-unicode-{conversions,ranges}.inc.h generator',
+    parser = argparse.ArgumentParser(description='lit-unicode-{conversions,ranges}-{sup}.inc.h generator',
                                      epilog='''
                                         The input files (UnicodeData.txt, SpecialCasing.txt)
```
galpeter, line 682: The epilog also needs an update on which files are used (e.g.: "DerivedCoreProperties")

rerobika:
@zherczeg @galpeter Thanks for the reviews, I've updated the PR.

rerobika:
Binary size changes:

```
ES.NEXT
   text    data     bss     dec     hex filename
 369588    7656 1577472 1954716  1dd39c build/bin/jerry
 367284    7656 1577472 1952412  1dca9c build/bin/jerry master

ES5.1
   text    data     bss     dec     hex filename
 225055    5096 1577312 1807463  1b9467 build/bin/jerry
 225351    5088 1577312 1807751  1b9587 build/bin/jerry master
```

PR #3933, opened 2020-06-25
Support symbol property names in Object.defineProperties

rtakacs:


jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -790,7 +795,12 @@ ecma_builtin_object_object_define_properties (ecma_object_t *obj_p, /**< routine
   for (uint32_t i = 0; i < prop_names_p->item_count; i++)
   {
     /* 5.a */
-    ecma_value_t desc_obj = ecma_op_object_get (props_p, ecma_get_string_from_value (buffer_p[i]));
+#if ENABLED (JERRY_ESNEXT)
+    ecma_string_t *prop_name_p = ecma_get_prop_name_from_value (buffer_p[i]);
```
galpeter, line 799: We can use the "get_prop_name_from" variant in every no need to have it guarded. In ES5.1 case it will only work with strings just like the "get_string_from_value".

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -816,13 +826,16 @@ ecma_builtin_object_object_define_properties (ecma_object_t *obj_p, /**< routine
...
   /* 6. */
-  buffer_p = prop_names_p->buffer_p;
-
   for (uint32_t i = 0; i < prop_names_p->item_count; i++)
   {
-    ecma_value_t define_own_prop_ret =  ecma_op_object_define_own_property (obj_p,
-                                                                            ecma_get_string_from_value (buffer_p[i]),
-                                                                            &property_descriptors[i]);
+#if ENABLED (JERRY_ESNEXT)
+    ecma_string_t *prop_name_p = ecma_get_prop_name_from_value (buffer_p[i]);
```
galpeter, line 832: Same as above.

rtakacs: Done.

jerry-core/ecma/builtin-objects/ecma-builtin-object.c
rtakacs: Done.

PR #3939, opened 2020-06-26
Support caching of next method in for-of and built-in methods.

zherczeg:
#3937 should go in first


rerobika:
#3937 Has just landed, please rebase the PR .

PR #3941, opened 2020-06-26
Added fuzzilli integration

nszetei:
Added [Fuzzilli](https://github.com/googleprojectzero/fuzzilli/) integration.

Build: 

```
CC=clang python tools/build.py --fuzzilli=on --lto=off
```

Fuzzing:

```
swift run FuzzilliCli --profile=jerryscript ~/jerryscript/build/bin/jerry-fuzzilli
```

### DIFF
--- a/jerry-main/fuzzilli.c
+++ b/jerry-main/fuzzilli.c
@@ -1,0 +1,376 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript.h"
+#include "jerryscript-ext/handler.h"
+
+/**
+ * Maximum size of source code
+ */
+#define JERRY_BUFFER_SIZE (1048576)
+static uint8_t buffer[JERRY_BUFFER_SIZE];
+
+#include <stdio.h>
+#include <assert.h>
+#include <stdlib.h>
+#include <fcntl.h>
+#include <errno.h>
+#include <unistd.h>
+#include <string.h>
+#include <sys/stat.h>
+#include <sys/wait.h>
+#include <sys/mman.h>
+
+#define REPRL_CRFD 100
+#define REPRL_CWFD 101
+#define REPRL_DRFD 102
+#define REPRL_DWFD 103
+
+#define SHM_SIZE 0x100000
+#define MAX_EDGES ((SHM_SIZE - 4) * 8)
+
+#ifndef DCHECK
+#define DCHECK(condition) \
+  {                       \
+    assert(condition);    \
+    abort();              \
+  }
+#endif
+
+#ifndef CHECK
+#define CHECK DCHECK
+#endif
+
+FILE *fdopen(int fd, const char *mode);
+void __sanitizer_cov_reset_edgeguards(void);
+void __sanitizer_cov_trace_pc_guard_init(uint32_t *start, uint32_t *stop);
+
+struct shmem_data
+{
+  uint32_t num_edges;
+  unsigned char edges[];
+};
+
+struct shmem_data *__shmem;
+uint32_t *__edges_start, *__edges_stop;
+
+void __sanitizer_cov_reset_edgeguards()
+{
+  uint32_t N = 0;
+  for (uint32_t *x = __edges_start; x < __edges_stop && N < MAX_EDGES; x++)
+    *x = ++N;
+}
+
+void __sanitizer_cov_trace_pc_guard_init(uint32_t *start, uint32_t *stop)
+{
+  // Avoid duplicate initialization
+  if (start == stop || *start)
+    return;
+
+  if (__edges_start != NULL || __edges_stop != NULL)
+  {
+    fprintf(stderr, "Coverage instrumentation is only supported for a single module\n");
+    _exit(-1);
+  }
+
+  __edges_start = start;
+  __edges_stop = stop;
+
+  // Map the shared memory region
+  const char *shm_key = getenv("SHM_ID");
+  if (!shm_key)
+  {
+    puts("[COV] no shared memory bitmap available, skipping");
+    __shmem = (struct shmem_data *)malloc(SHM_SIZE);
+  }
+  else
+  {
+    int fd = shm_open(shm_key, O_RDWR, S_IRUSR | S_IWUSR);
+    if (fd <= -1)
+    {
+      fprintf(stderr, "Failed to open shared memory region: %s\n", strerror(errno));
+      _exit(-1);
+    }
+
+    __shmem = (struct shmem_data *)mmap(0, SHM_SIZE, PROT_READ | PROT_WRITE, MAP_SHARED, fd, 0);
+    if (__shmem == MAP_FAILED)
+    {
+      fprintf(stderr, "Failed to mmap shared memory region\n");
+      _exit(-1);
+    }
+  }
+
+  __sanitizer_cov_reset_edgeguards();
+
+  __shmem->num_edges = (uint32_t)(stop - start);
+  printf("[COV] edge counters initialized. Shared memory: %s with %u edges\n", shm_key, __shmem->num_edges);
+}
+
+void __sanitizer_cov_trace_pc_guard(uint32_t *guard)
+{
+  // There's a small race condition here: if this function executes in two threads for the same
+  // edge at the same time, the first thread might disable the edge (by setting the guard to zero)
+  // before the second thread fetches the guard value (and thus the index). However, our
+  // instrumentation ignores the first edge (see libcoverage.c) and so the race is unproblematic.
+  uint32_t index = *guard;
+  __shmem->edges[index / 8] |= 1 << (index % 8);
+  *guard = 0;
+}
+
+// We have to assume that the fuzzer will be able to call this function e.g. by
+// enumerating the properties of the global object and eval'ing them. As such
+// this function is implemented in a way that requires passing some magic value
+// as first argument (with the idea being that the fuzzer won't be able to
+// generate this value) which then also acts as a selector for the operation
+// to perform.
+jerry_value_t
+jerryx_handler_fuzzilli(const jerry_value_t func_obj_val, /**< function object */
+                        const jerry_value_t this_p,       /**< this arg */
+                        const jerry_value_t args_p[],     /**< function arguments */
+                        const jerry_length_t args_cnt)    /**< number of function arguments */
+{
+  (void)func_obj_val; /* unused */
+  (void)this_p;       /* unused */
+
+  jerry_char_t operation[256] = {0};
+  jerry_value_t ret_val = jerry_create_undefined();
+
+  if (args_cnt > 0 && jerry_value_is_string (args_p[0]))
+  {
+    jerry_value_t str_val;
+    str_val = jerry_value_to_string (args_p[0]);
+
+    if (!jerry_value_is_error (str_val))
+    {
+      jerry_length_t length = jerry_get_utf8_string_length (str_val);
+
+      if (length > 0 && length < 256)
+      {
+        jerry_string_to_utf8_char_buffer (str_val, operation, length);
+      }
+    }
+
+    jerry_release_value (str_val);
+  }
+
+  if (strcmp((char *)operation, "FUZZILLI_CRASH") == 0)
+  {
+    if (args_cnt == 2 && jerry_value_is_number(args_p[1]))
+    {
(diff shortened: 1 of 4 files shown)


jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,376 @@
...
+  (void)func_obj_val; /* unused */
+  (void)this_p;       /* unused */
+
+  jerry_char_t operation[256] = {0};
+  jerry_value_t ret_val = jerry_create_undefined();
+
+  if (args_cnt > 0 && jerry_value_is_string (args_p[0]))
+  {
+    jerry_value_t str_val;
+    str_val = jerry_value_to_string (args_p[0]);
```
rerobika, line 153: You checked above that `args_p[0]` is string so calling `toString` for it is unnecessary.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,376 @@
...
+jerry_value_t
+jerryx_handler_fuzzilli(const jerry_value_t func_obj_val, /**< function object */
+                        const jerry_value_t this_p,       /**< this arg */
+                        const jerry_value_t args_p[],     /**< function arguments */
+                        const jerry_length_t args_cnt)    /**< number of function arguments */
+{
+  (void)func_obj_val; /* unused */
+  (void)this_p;       /* unused */
+
+  jerry_char_t operation[256] = {0};
```
rerobika, line 147: Is there any reason of limitation? IMO, you can use `jerry_heap_alloc` at this point then no size check would be needed below.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,376 @@
...
+  jerry_value_t ret_val = jerry_create_undefined();
+
+  if (args_cnt > 0 && jerry_value_is_string (args_p[0]))
+  {
+    jerry_value_t str_val;
+    str_val = jerry_value_to_string (args_p[0]);
+
+    if (!jerry_value_is_error (str_val))
+    {
+      jerry_length_t length = jerry_get_utf8_string_length (str_val);
```
rerobika, line 157: IMO that meant to be `jerry_get_utf8_string_size`. [Here](https://github.com/rerobika/jerryscript/commit/cad0625e14dc6e4ccbb44ca5c107c9504b273e09#diff-ef848aeb6f84fffd11a62d59adf3bf56R532) you can see a short example of dealing with string arguments/copying them into buffers.

nszetei: Thank you for the example, I adjusted the code accordingly.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+  {
+    static FILE *fzliout;
+    fzliout = fdopen(REPRL_DWFD, "w");
+    if (!fzliout)
+    {
+      fprintf(stderr, "Fuzzer output channel not available, printing to stdout instead\n");
+      fzliout = stdout;
+    }
+
+    /* Based on the jerryx_handler_print handler */
```
akosthekiss, line 187: Is this a verbatim copy of `jerryx_handler_print`? (Disclaimer: I haven't checked.) If it is, wouldn't it be better to redirect stdout temporarily to `REPRL_DWFD`, then call `jerryx_handler_print`, and finally, reset stdout file descriptor to its original file? Something using `dup`/`dup2`. (Disclaimer: I haven't tried.) It would look like a bit of magic, but it would prevent copy-paste-induced future maintenance burden.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+ */
+static void
+register_js_function(const char *name_p,                 /**< name of the function */
+                     jerry_external_handler_t handler_p) /**< function callback */
+{
+  jerry_value_t result_val = jerryx_handler_register_global((const jerry_char_t *)name_p, handler_p);
+
+  if (jerry_value_is_error(result_val))
+  {
+    jerry_port_log(JERRY_LOG_LEVEL_WARNING, "Warning: failed to register '%s' method.", name_p);
```
akosthekiss, line 281: It seems that this is the only place where `jerry_port_log` is used. Should this also be `fprintf`? Or should all other places be `jerry_port_log` calls? (Main message is: use one API and use it consistently.)

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+          status_rc = 1;
+        }
+        jerry_release_value(run_value);
+      }
+
+      jerry_release_value(parse_value);
+      jerry_cleanup();
+
+      status_rc <<= 8;
+      if (write(REPRL_CWFD, &status_rc, 4) != 4)
```
akosthekiss, line 363: This is "only" a style issue, but my eyes miss the curly braces. (Doesn't our style checker catch this?...)

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+    assert(condition);    \
+    abort();              \
+  }
+#endif
+
+#ifndef CHECK
+#define CHECK DCHECK
+#endif
+
+FILE *fdopen(int fd, const char *mode);
```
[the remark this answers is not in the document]
nszetei: The function `fdopen` is POSIX-only see more e.g. [here](https://stackoverflow.com/questions/12236124/cannot-interpret-compiler-warning). Currently the `USING_CLANG` variable forces ANSI C99 option and which does not provide `fdopen`. This is a workaround to keep the code simple and uniform between various targets, and at the same time to avoid adding unnecessary new logic. If you are aware of the better solution, please let me know.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+void __sanitizer_cov_reset_edgeguards(void);
```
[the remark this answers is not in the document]
nszetei: I changed the function name with underscores and used the `static` keyword. This was the convention used by the Fuzzilli fuzzer (I wrote only this integration).

For Clang I saw the default option `strict-prototypes` in your `CMakeLists.txt` and that's why I kept these declarations, otherwise my build failed (e.g. invoked with `CC=clang python tools/build.py --fuzzilli=on --lto=off`). What do you recommend to do instead, instead of keeping these declarations?

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+void __sanitizer_cov_trace_pc_guard_init(uint32_t *start, uint32_t *stop);
```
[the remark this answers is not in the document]
nszetei: See the previous point.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+
+struct shmem_data
+{
+  uint32_t num_edges;
+  unsigned char edges[];
+};
+
+struct shmem_data *__shmem;
+uint32_t *__edges_start, *__edges_stop;
```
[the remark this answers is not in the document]
nszetei: `__sanitizer_cov_trace_pc_guard_init` is used internally by Clang and there is a shared memory region used to pass the coverage between the fuzzer and the target.

I added a links to https://github.com/googleprojectzero/fuzzilli/tree/master/Targets and https://github.com/googleprojectzero/fuzzilli/blob/master/Targets/coverage.c which should be up-to-date and it contains the current information about the necessary modifications. Most of this code then be just copy-pasted.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+#define DCHECK(condition) \
+  {                       \
+    assert(condition);    \
+    abort();              \
+  }
+#endif
+
+#ifndef CHECK
+#define CHECK DCHECK
+#endif
```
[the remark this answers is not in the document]
nszetei: Not by this target, I removed it and later used `jerry_port_fatal` instead.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+    {
+      fprintf(stderr, "Failed to mmap shared memory region\n");
+      _exit(-1);
+    }
+  }
+
+  __sanitizer_cov_reset_edgeguards();
+
+  __shmem->num_edges = (uint32_t)(stop - start);
+  printf("[COV] edge counters initialized. Shared memory: %s with %u edges\n", shm_key, __shmem->num_edges);
```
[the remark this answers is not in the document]
nszetei: OK, updated.

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,371 @@
...
+
+  if (strcmp((char *)operation, "FUZZILLI_CRASH") == 0)
+  {
+    if (args_cnt == 2 && jerry_value_is_number(args_p[1]))
+    {
+      int arg = (int)jerry_get_number_value(args_p[1]);
+      switch (arg)
+      {
+      case 0:
+        *((int *)0x41414141) = 0x1337;
```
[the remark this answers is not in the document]
nszetei: This is a part of the initialization sanity check done by the Fuzzilli. It verifies if it's able to intercept the segfault and if not, it refuses to run. The parameters to generate the crash could be adjusted in the target [Profile](https://github.com/googleprojectzero/fuzzilli/blob/e26e7f78f4b364f5431547f68a412d558e56b07a/Sources/FuzzilliCli/Profiles/JerryscriptProfile.swift).

jerry-main/fuzzilli.c
nszetei: I was also thinking about it, but `dup2` wasn't working. JerryScript internally uses `putchar`, which uses `FILE *stdout` stream. Instead of `dup()`-ing the descriptor I made a backup of `stdout`, replaced it by the stream representing `REPRL_DWFD`, called the original `jerryx_handler_print ` handler and restored the `stdout`.

jerry-main/fuzzilli.c
nszetei: I changed the code to use `fprintf(stderr,` everywhere.

jerry-main/fuzzilli.c
nszetei: No. As I am not acquainted with the building process (nor codebase), I found `check-vera.sh` only after some tests failed, so I had to correct about 100+ another styling issues, it should be fine now.

Anyway, thanks for your time and valuable feedback! I fixed most of the things I was able, but if you think this requires too much work to maintain (or adjust to conform the JerryScript convention), I am fine with keeping the integration part as an external patch.

rerobika:
@nszetei Please check the errors reported by the [cppcheck](https://github.com/jerryscript-project/jerryscript/pull/3941/checks?check_run_id=815908770#step:10:10).

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,300 @@
...
+
+  if (strcmp ((char *) operation, "FUZZILLI_CRASH") == 0)
+  {
+    if (args_cnt == 2 && jerry_value_is_number (args_p[1]))
+    {
+      int arg = (int) jerry_get_number_value (args_p[1]);
+      switch (arg)
+      {
+        case 0:
+        *((int *) 0x41414141) = 0x1337;
```
akosthekiss, line 171: Alright, I see that this construct is used in all targets in fuzzilli to trigger a segfault. Leave it then, but please add a comment as it is done in the duktape integration (`/* intentionally segfault */`).

(I still wonder why not `raise (SIGSEGV);` is used everywhere.)

jerry-main/fuzzilli.c
```diff
@@ -0,0 +1,300 @@
...
+#include <sys/wait.h>
+#include <sys/mman.h>
+
+/**
+ * Fuzzilli interface, for more information refer to:
+ * https://github.com/googleprojectzero/fuzzilli/tree/master/Targets
+ * https://github.com/googleprojectzero/fuzzilli/blob/master/Targets/coverage.c
+ */
+
+// BEGIN FUZZING CODE
```
akosthekiss, line 42: Alright, NOW I see what's happening here... So, this is something that is actually in `fuzzilli/Targets/coverage.c`. That code has a quite different coding style. If we applied our coding style and additional changes to that code, it could be a nightmare to sync/maintain it in the future. So, why don't we just add that file to `third-party/fuzzilli/coverge.c` and include(!) "coverage.c" from here. Then maintenance, if ever needed, becomes just copying that file from that project to this one.

Something like the following might be needed in `jerry-main/CMakeLists.txt`:
```cmake
target_include_directories("jerry-fuzzilli" PRIVATE "${CMAKE_SOURCE_DIR}/third-party/fuzzilli")
```

jerry-main/fuzzilli.c
akosthekiss: >  but if you think this requires too much work to maintain (or adjust to conform the JerryScript convention), I am fine with keeping the integration part as an external patch.

Nah, that was not my intent with the review. External patches bit rot. Upstream maintainers don't know about them so don't care about them. Patch maintainers often don't have the time to follow up on upstream. It's better to have this here, JerryScript also benefits from this.

We are getting there, I think we may have the right approach with the third-party subdir. Just bear with me, I'm admittedly nitpicky.

akosthekiss:
One more note: JerryScript has the policy of one commit per PR. So, when all technicalities will get sorted out, be prepared for a request to squash all commits, rebase them on latest master, and force push it to your GH branch.

akosthekiss:
@nszetei Any update on this?

akosthekiss:
@nszetei ping?

akosthekiss:
@nszetei 3rd ping?

nszetei:
Sorry, I don't have a time to work on this anymore, so I am closing this issue.

PR #3942, opened 2020-06-26
Update Symbol.split functions for String and Regexp to match new standard

rwalczyna:
In newest ES standard limit parameter is calculated using
ToUint32 function instead of ToLength.

- there was an assert failure in Regexp split function, when lastIndex were forced to value bigger than size (calculation method was changed in ECMA-262 v11, 21.2.5.11 19.d.ii)
- ToLength function for limit argument was changed to ToUint32
- test built-ins/RegExp/prototype/Symbol.split/coerce-limit.js fails, because it is not valid for new version - ToUint32 works different than ToLength for negative numbers. (Test is changed [here](https://github.com/tc39/test262/commit/bd8c91e250690567240e3aa4ade9a740b94b1aee))

### DIFF
--- a/tests/test262-es6-excludelist.xml
+++ b/tests/test262-es6-excludelist.xml
@@ -86,6 +86,7 @@
   <test id="built-ins/RegExp/prototype/Symbol.replace/y-set-lastindex.js"><reason></reason></test>
   <test id="built-ins/RegExp/prototype/Symbol.search/get-sticky-coerce.js"><reason></reason></test>
   <test id="built-ins/RegExp/prototype/Symbol.search/get-sticky-err.js"><reason></reason></test>
+  <test id="built-ins/RegExp/prototype/Symbol.split/coerce-limit.js"><reason>Test is outdated: ES11, 21.2.5.13 13</reason></test>
   <test id="built-ins/RegExp/prototype/test/get-sticky-err.js"><reason></reason></test>
   <test id="built-ins/String/prototype/normalize/form-is-not-valid-throws.js"><reason></reason></test>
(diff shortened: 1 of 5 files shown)


tests/test262-es6-excludelist.xml
```diff
@@ -86,6 +86,7 @@
   <test id="built-ins/RegExp/prototype/Symbol.replace/y-set-lastindex.js"><reason></reason></test>
   <test id="built-ins/RegExp/prototype/Symbol.search/get-sticky-coerce.js"><reason></reason></test>
   <test id="built-ins/RegExp/prototype/Symbol.search/get-sticky-err.js"><reason></reason></test>
+  <test id="built-ins/RegExp/prototype/Symbol.split/coerce-limit.js"><reason></reason></test>
```
ossy-szeged, line 89: Please add a short comment to reason block, for example: won't fix due to spec change - #3942

tests/test262-es6-excludelist.xml
```diff
@@ -86,6 +86,7 @@
   <test id="built-ins/RegExp/prototype/Symbol.replace/y-set-lastindex.js"><reason></reason></test>
   <test id="built-ins/RegExp/prototype/Symbol.search/get-sticky-coerce.js"><reason></reason></test>
   <test id="built-ins/RegExp/prototype/Symbol.search/get-sticky-err.js"><reason></reason></test>
+  <test id="built-ins/RegExp/prototype/Symbol.split/coerce-limit.js"><reason>Test is wrong because of specification change</reason></test>
```
LaszloLango, line 89: Please refer the ES specification version of the change.

PR #3943, opened 2020-06-26
Improve the JSON C API

galpeter:
* Added more details into documentation.
* Moved the C unit-test into its own file.
* Added extra test cases.
* Extended the API reference documentation with doctests.

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -4332,11 +4332,15 @@ jerry_get_typedarray_buffer (jerry_value_t value, /**< TypedArray to get the arr
 } /* jerry_get_typedarray_buffer */
 
 /**
- * Create an object from JSON
+ * Parse the given JSON string to create a jerry_value_t.
+ *
+ * The behaviour is equivalent with the "JSON.parse(string)" JS call.
  *
  * Note:
- *      The returned value must be freed with jerry_release_value
- * @return jerry_value_t from json formated string or an error massage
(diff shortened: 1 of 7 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-json.c
```diff
@@ -1349,13 +1349,16 @@ static ecma_value_t ecma_builtin_json_str_helper (ecma_json_stringify_context_t
 } /* ecma_builtin_json_str_helper */
 
 /**
- * Function to create a json formated string from an object
+ * Function to create a JSON string from a JS value.
  *
- * @return ecma_value_t containing a json string
- *         Returned value must be freed with ecma_free_value.
+ * Note:
+ *      The returned value must be freed with jerry_release_value,
```
rerobika, line 1355: `jerry_release_value` is not available at ECMA layer.

jerry-core/api/jerry.c
```diff
@@ -4332,11 +4332,15 @@ jerry_get_typedarray_buffer (jerry_value_t value, /**< TypedArray to get the arr
 } /* jerry_get_typedarray_buffer */
 
 /**
- * Create an object from JSON
+ * Parse the given JSON string to create a jerry_value_ŧ.
```
LaszloLango, line 4335: That suffix looks a bit odd, it it a regular `t` character?

galpeter: Should be fixed now

PR #3946, opened 2020-06-28
Implement for-await-of statement.

zherczeg:
AsyncIteratorClose has not supported yet.


zherczeg:
The patch contains renames, which makes it somewhat bigger.

PR #3952, opened 2020-06-29
Correct Object.keys to collect only enumerable property names

rtakacs:
In case of Proxy objects, there were no property descriptor checks when collecting enumerable properties. Therefore, all the non-enumerable Proxy.target properties were listed as well.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -2064,7 +2064,36 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
       }
       else if (!prop_is_symbol && (opts & ECMA_LIST_SYMBOLS_ONLY) == 0)
       {
-        ecma_collection_push_back (return_keys, entry);
+        if ((opts & ECMA_LIST_ENUMERABLE) == 0)
+        {
+          ecma_collection_push_back (return_keys, entry);
+        }
+        else
+        {
+          ecma_property_descriptor_t prop_desc;
+          ecma_value_t status = ecma_proxy_object_get_own_property_descriptor (obj_p, prop_name_p, &prop_desc);
+
+          if (ECMA_IS_VALUE_ERROR (status))
+          {
+            ecma_collection_destroy (proxy_keys);
+
+            return NULL;
+          }
+
+          if ((prop_desc.flags & ECMA_PROP_IS_ENUMERABLE) == 0)
+          {
+            ecma_deref_ecma_string (prop_name_p);
+          }
+          else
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2064,7 +2064,39 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
...
+          ecma_collection_push_back (return_keys, entry);
+        }
+        else
+        {
+          ecma_property_descriptor_t prop_desc;
+          ecma_value_t status = ecma_proxy_object_get_own_property_descriptor (obj_p, prop_name_p, &prop_desc);
+
+          if (ECMA_IS_VALUE_ERROR (status))
+          {
+            ecma_free_value (status);
```
rerobika, line 2078: freeing the error representing value is unnecessary.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2064,7 +2064,39 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
...
+            ecma_collection_destroy (proxy_keys);
+
+            return NULL;
+          }
+
+          if ((prop_desc.flags & ECMA_PROP_IS_ENUMERABLE) == 0)
+          {
+            ecma_free_value (entry);
```
rerobika, line 2086: I'd rather use `ecma_deref_ecma_string` here.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2064,7 +2064,39 @@ ecma_op_object_get_property_names (ecma_object_t *obj_p, /**< object */
...
+          {
+            ecma_collection_push_back (return_keys, entry);
+          }
+
+          if (ecma_is_value_true (status))
+          {
+            ecma_free_property_descriptor (&prop_desc);
+          }
+
+          ecma_free_value (status);
```
rerobika, line 2098: releasing the this simple value is unnecessary.

jerry-core/ecma/operations/ecma-objects.c
rtakacs: Good idea.

jerry-core/ecma/operations/ecma-objects.c
rtakacs: Done.

jerry-core/ecma/operations/ecma-objects.c
rtakacs: Done.

PR #3953, opened 2020-06-29
Fix Object.prototype.toLocaleString() and Array.prototype.toLocaleStr…

ossy-szeged:
…ing()

Since ES6 Object.prototype.toLocaleString() shouldn't call ToObject() and
Array.prototype.toLocaleString() shouldn't call ToObject() for each element.
References:
- https://www.ecma-international.org/ecma-262/11.0/index.html#sec-array.prototype.tolocalestring
- https://www.ecma-international.org/ecma-262/11.0/index.html#sec-object.prototype.tolocalestring


ossy-szeged:
We can do it with less ifdef guard, thanks @rerobika. Updated PR is coming soon.

PR #3962, opened 2020-07-01
Implement missing async function and async iterator prototypes.

zherczeg:
#3955 should land first


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -442,32 +442,42 @@ ecma_op_create_simple_function_object (ecma_object_t *scope_p, /**< function's s
 #if ENABLED (JERRY_ESNEXT)
 
 /**
- * GeneratorFunction object creation operation.
- *
- * See also: ECMA-262 v5, 13.2
+ * Create a fanction object with the appropriate prototype.
```
rerobika, line 445: Typo, fanction.

PR #3964, opened 2020-07-01
Methods shouldn't have prototype property

ossy-szeged:


jerry-core/parser/js/js-parser-internal.h
```diff
@@ -74,11 +74,11 @@ typedef enum
   PARSER_ALLOW_SUPER_CALL = (1u << 22),       /**< allow super constructor call
                                                *   Note: PARSER_CLASS_CONSTRUCTOR must be present */
   PARSER_ALLOW_NEW_TARGET = (1u << 23),       /**< allow new.target parsing in the current context */
-
+  PARSER_IS_METHOD = (1u << 24),              /**< method is parsed */
```
zherczeg, line 77: I think setting both PARSER_IS_PROPERTY_GETTER and PARSER_IS_PROPERTY_SETTER could replace this flag, so we don't need to reduce the number of available flags.

ossy-szeged: > I think setting both PARSER_IS_PROPERTY_GETTER and PARSER_IS_PROPERTY_SETTER could replace this flag, so we don't need to reduce the number of available flags.

I understand that technically we could avoid adding a new parser flag for methods. But I really don't like the idea to reuse PARSER_IS_PROPERTY_GETTER and PARSER_IS_PROPERTY_SETTER flag for it. Methods aren't related to getters and setters at all, and it would make the code confusing, it could cause potential bugs in the future.

rerobika: I think I'd be good for now... If we are going to get out parser flags we can combine these.

PR #3966, opened 2020-07-01
Implement ToPropertyKey operation

galpeter:
From ES 6 the ToPropertyKey operation is used to convert elements to a valid property key.
This "new" method uses the ToPrimitive operation which can call the `@@ToPrimitive`
well-known symbol to convert the given element to a key.

### DIFF
--- a/jerry-core/ecma/operations/ecma-conversion.c
+++ b/jerry-core/ecma/operations/ecma-conversion.c
@@ -490,27 +490,51 @@ ecma_op_to_string (ecma_value_t value) /**< ecma value */
 } /* ecma_op_to_string */
 
 /**
- * ToPropertyName operation.
+ * ToPropertyKey operation.
+ *
+ * See also:
+ *   ECMA 262 v6, 7.1.14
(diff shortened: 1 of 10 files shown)


jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -490,27 +490,42 @@ ecma_op_to_string (ecma_value_t value) /**< ecma value */
...
  */
 ecma_string_t *
-ecma_op_to_prop_name (ecma_value_t value) /**< ecma value */
+ecma_op_to_property_key (ecma_value_t value) /**< ecma value */
 {
-  ecma_check_value_type_is_spec_defined (value);
-
 #if ENABLED (JERRY_ESNEXT)
-  if (ecma_is_value_symbol (value))
+  ecma_value_t key = ecma_op_to_primitive (value, ECMA_PREFERRED_TYPE_STRING);
```
rerobika, line 507: ToPrimitve returns with the passed value, if the value is not object. How about keeping the fast path for non-object values?

PR #3967, opened 2020-07-01
Fix Date.prototype[@@toPrimitive] hint handling

galpeter:
The Date.prototype[@@toPrimitive] only allows the "string", "default" and "number"
hint values. Any other value should throw a TypeError.


ossy-szeged:
LGTM (informal)

PR #3970, opened 2020-07-02
Fix namedaccessor property redefinition

rerobika:
The previously allocated property should not be deleted only be changed to nameddata property.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -428,6 +428,12 @@ typedef enum
 #define ECMA_PROPERTY_NAME_TYPE_SHIFT (ECMA_PROPERTY_FLAG_SHIFT + 4)
 
 /**
+ * Convert data property to accessor property or accessor property to data property
+ */
+#define ECMA_CHANGE_PROPERTY_TYPE(property_p) \
+  *(property_p) ^= ECMA_PROPERTY_TYPE_NAMEDACCESSOR ^ ECMA_PROPERTY_TYPE_NAMEDDATA;
+
+/**
  * Convert data property to internal property.
  */
 #define ECMA_CONVERT_DATA_PROPERTY_TO_INTERNAL_PROPERTY(property_p) \
--- a/tests/jerry/es.next/object-property-redefiniton.js
+++ b/tests/jerry/es.next/object-property-redefiniton.js
@@ -1,0 +1,52 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+function checkOwnProperties(obj, propList)
+{
+  names = Object.getOwnPropertyNames(obj)
+  assert(names.length === propList.length)
+
+  for (var i = 0; i < propList.length; ++i)
+  {
+    assert(names[i] === propList[i])
+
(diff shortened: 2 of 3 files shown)


jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -751,6 +751,34 @@ ecma_get_named_data_property (ecma_object_t *obj_p, /**< object to find property
   return ECMA_PROPERTY_VALUE_PTR (property_p);
 } /* ecma_get_named_data_property */
 
+/**
+ * Release the accessor property value and invalidate it from Lcache
+ */
+void
```
zherczeg, line 757: Can it be always inline?

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -770,13 +798,8 @@ ecma_free_property (ecma_object_t *object_p, /**< object the property belongs to
...
     {
-#if ENABLED (JERRY_CPOINTER_32_BIT)
-      ecma_getter_setter_pointers_t *getter_setter_pair_p;
-      getter_setter_pair_p = ECMA_GET_NON_NULL_POINTER (ecma_getter_setter_pointers_t,
-                                                        ECMA_PROPERTY_VALUE_PTR (property_p)->getter_setter_pair_cp);
-      jmem_pools_free (getter_setter_pair_p, sizeof (ecma_getter_setter_pointers_t));
-#endif /* ENABLED (JERRY_CPOINTER_32_BIT) */
-      break;
+      ecma_release_accessor_property_value (object_p, name_cp, property_p);
+      goto release_prop_name;
```
zherczeg, line 802: Is it a big gain? I don't like goto.

tests/jerry/es.next/object-property-redefiniton.js
```diff
@@ -0,0 +1,41 @@
...
+var o = {
+  get a() {
+    return 5;
+  },
+  b:6,
+  set c(_) {
+  },
+  d:10,
+  a: 11,
+  c: 12
```
zherczeg, line 38: What about the other direction? When the property is a data property first and redefined as accessor.

rerobika: I've updated the PR according to it.

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -427,6 +427,18 @@ typedef enum
...
+ * Convert data property to accessor property.
+ */
+#define ECMA_CONVERT_DATA_PROPERTY_TO_ACCESSOR_PROPERTY(property_p) \
+   *(property_p) = (uint8_t) (*(property_p) + (ECMA_PROPERTY_TYPE_NAMEDACCESSOR - ECMA_PROPERTY_TYPE_NAMEDDATA))
+
+/**
+ * Convert accessor property to data property.
+ */
+#define ECMA_CONVERT_ACCESSOR_PROPERTY_TO_DATA_PROPERTY(property_p) \
+   *(property_p) = (uint8_t) (*(property_p) - (ECMA_PROPERTY_TYPE_NAMEDACCESSOR - ECMA_PROPERTY_TYPE_NAMEDDATA))
```
zherczeg, line 440: I think you can do both direction with a xor operator:
`*(property_p) ^= ECMA_PROPERTY_TYPE_NAMEDACCESSOR ^ ECMA_PROPERTY_TYPE_NAMEDDATA;`

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -751,6 +751,75 @@ ecma_get_named_data_property (ecma_object_t *obj_p, /**< object to find property
...
+    name_cp = (jmem_cpointer_t) ECMA_GET_DIRECT_STRING_VALUE (name_p);
+  }
+  else
+  {
+    ECMA_SET_NON_NULL_POINTER (name_cp, name_p);
+  }
+
+  if (ecma_is_property_lcached (property_p))
+  {
+    ecma_lcache_invalidate (object_p, name_cp, property_p);
```
zherczeg, line 818: Why do we need to invalidate lcache? Isn't hashing only use object_p and name_p? Those will not change.

rerobika: Indeed!

rerobika:
@zherczeg Thanks for the review, I've updated the PR.

PR #3989, opened 2020-07-09
Optimize [[Call]] arguments stack usage

rerobika:
This patch is the first step of incoming changes about function call/construct arguments passing on the C stack.
The patch contains several [[TODO]] comments which are going to be resolved in upcoming PRs.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -832,28 +832,23 @@ ecma_op_get_prototype_from_constructor (ecma_object_t *ctor_obj_p, /**< construc
  * @return the result of the function call.
  */
 static ecma_value_t
-ecma_op_function_call_simple (ecma_object_t *func_obj_p, /**< Function object */
-                              ecma_value_t this_arg_value, /**< 'this' argument's value */
-                              const ecma_value_t *arguments_list_p, /**< arguments list */
-                              ecma_length_t arguments_list_len) /**< length of arguments list */
+ecma_op_function_call_simple (ecma_call_args_t *call_args_p) /**< call arguments */
 {
-  JERRY_ASSERT (ecma_get_object_type (func_obj_p) == ECMA_OBJECT_TYPE_FUNCTION);
+  JERRY_ASSERT (ecma_get_object_type (call_args_p->func_obj_p) == ECMA_OBJECT_TYPE_FUNCTION);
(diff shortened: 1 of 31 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1068,33 +1056,53 @@ ecma_op_bound_function_get_argument_list (ecma_object_t *func_obj_p, /**< bound
...
   return ret_value;
 } /* ecma_op_function_call_bound */
 
+/**
+ * Create a new function call packed info from the given function call arguments
+ *
+ * @return ecma_call_args_t structure
+ */
+ecma_call_args_t
+ecma_op_function_make_args (ecma_object_t *func_obj_p, /**< Function object */
```
dbatyai, line 1091: I think this should be marked always inline just to be sure.

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1068,33 +1056,53 @@ ecma_op_bound_function_get_argument_list (ecma_object_t *func_obj_p, /**< bound
...
+                            ecma_value_t this_arg_value, /**< 'this' argument's value */
+                            const ecma_value_t *arguments_list_p, /**< arguments list */
+                            ecma_length_t arguments_list_len) /**< length of arguments list */
+{
+  ecma_call_args_t call_args =
```
dbatyai, line 1096: This structure doesn't look nice alignment-wise. I think `argc` and `argv` should be swapped.

jerry-core/vm/vm.c
```diff
@@ -336,11 +334,9 @@ vm_run_global (const ecma_compiled_code_t *bytecode_p) /**< pointer to bytecode
   }
 #endif /* ENABLED (JERRY_MODULE_SYSTEM) */
 
-  return vm_run (bytecode_p,
-                 ecma_make_object_value (glob_obj_p),
-                 global_scope_p,
-                 NULL,
-                 0);
+  ecma_call_args_t call_args = ecma_op_function_make_args (glob_obj_p, ecma_make_object_value (glob_obj_p), NULL, 0);
```
dbatyai, line 337: Having NULL as the argument list seems to be quite common, and it seems like a waste of stack. We already know there are no arguments from `argc`, would it be possible to not have to store the NULL pointer in this case?

PR #3993, opened 2020-07-09
Implement Object.{entries, values} built-in methods

rtakacs:
Object.keys also updated to follow the ES11 standard.


jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -694,19 +697,90 @@ ecma_builtin_object_object_is_extensible (ecma_object_t *obj_p) /**< routine's a
...
+  {
+    return ECMA_VALUE_ERROR;
+  }
+
+  ecma_value_t *names_buffer_p = props_p->buffer_p;
+  /* 3. */
+  ecma_value_t array_value = ecma_op_create_array_object (names_buffer_p, props_p->item_count, false);
+  ecma_collection_free (props_p);
+
+  return array_value;
```
rerobika, line 783: I'd create a new helper function for these 3 routine since only the listing option differs.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2016,6 +2016,70 @@ ecma_op_object_is_prototype_of (ecma_object_t *base_p, /**< base object */
...
+    return prop_names_p;
+  }
+
+  ecma_value_t *names_buffer_p = prop_names_p->buffer_p;
+
+  for (uint32_t i = 0; i < prop_names_p->item_count; i++)
+  {
+    JERRY_ASSERT (ecma_is_value_string (names_buffer_p[i]));
+
+    ecma_value_t value = ecma_op_object_get (obj_p, ecma_get_prop_name_from_value (names_buffer_p[i]));
```
rerobika, line 2052: since the element is string, you can use get_string_from_value.

PR #4004, opened 2020-07-15
Update RegExp unicode mode case folding to conform to the standard

dbatyai:


jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -403,30 +403,43 @@ lit_code_point_t
...
   }
+#endif /* !ENABLED (JERRY_ESNEXT) */
 
+  JERRY_UNUSED (unicode);
   lit_code_point_t cu = lit_char_to_upper_case (ch, NULL);
 
-  if (cu == LIT_MULTIPLE_CU)
+  if (ch <= LIT_UTF8_1_BYTE_CODE_POINT_MAX
+      || (cu > LIT_UTF8_1_BYTE_CODE_POINT_MAX
+          && cu != LIT_MULTIPLE_CU))
```
rerobika, line 437: Isn't this equivalent to `(ch <= LIT_UTF8_1_BYTE_CODE_POINT_MAX || cu != LIT_MULTIPLE_CU )` ?

dbatyai: There can be non-ascii codepoints that map to ascii, and these need to be filtered out. Either the original code point (`ch`) is ascii in which case the mapping is implicitly applied, or the original code point is non-ascii, where the mapping is only valid if the mapped code point (`cu`) is also non-ascii and a one-to-one mapping.

rerobika: Thanks for the explanation!

jerry-core/lit/lit-char-helpers.c
```diff
@@ -894,3 +897,51 @@ lit_char_to_upper_case (lit_code_point_t cp, /**< code point */
...
+  return (cp > LIT_UTF16_CODE_UNIT_MAX
+          || (!lit_search_char_in_interval_array ((ecma_char_t) cp,
+                                                  lit_unicode_folding_skip_to_lower_interval_starts,
+                                                  lit_unicode_folding_skip_to_lower_interval_lengths,
+                                                  NUM_OF_ELEMENTS (lit_unicode_folding_skip_to_lower_interval_starts))
+              && !lit_search_char_in_array ((ecma_char_t) cp,
+                                            lit_unicode_folding_skip_to_lower_chars,
+                                            NUM_OF_ELEMENTS (lit_unicode_folding_skip_to_lower_chars))));
+#else /* !ENABLED (JERRY_UNICODE_CASE_CONVERSION) */
+  return true;
```
LaszloLango, line 921: The return value of `lit_char_fold_to_lower` and `lit_char_fold_to_upper` is different when  `JERRY_UNICODE_CASE_CONVERSION` is not enabled. It looks odd to me. Is it intentional?

dbatyai: Yes, 1 byte code points always fold to the lowercase variant when unicode mode is enabled, that's why fold to upper is false.

PR #4005, opened 2020-07-15
Fix sticky flag handling in @@replace

dbatyai:


jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -2499,10 +2499,13 @@ ecma_regexp_split_helper (ecma_value_t this_arg, /**< this value */
  */
 static ecma_value_t
 ecma_regexp_replace_helper_fast (ecma_replace_context_t *ctx_p, /**<replace context */
-                                 const re_compiled_code_t *bc_p, /**< regexp bytecode */
+                                 ecma_extended_object_t *re_obj_p, /**< regexp object */
                                  ecma_string_t *string_p, /**< source string */
                                  ecma_value_t replace_arg) /**< replace argument */
 {
+  const re_compiled_code_t *bc_p = ECMA_GET_INTERNAL_VALUE_ANY_POINTER (re_compiled_code_t,
```
rerobika, line 2506: Since we assert that `bc_p` is not null below, how about using  `ECMA_GET_INTERNAL_VALUE_POINTER` instead?

jerry-core/ecma/operations/ecma-regexp-object.c
```diff
@@ -2519,33 +2522,60 @@ ecma_regexp_replace_helper_fast (ecma_replace_context_t *ctx_p, /**<replace cont
...
+  ecma_regexp_ctx_t re_ctx;
+  ecma_regexp_initialize_context (&re_ctx,
+                                  bc_p,
+                                  ctx_p->string_p,
+                                  string_end_p);
 
 #if ENABLED (JERRY_ESNEXT)
   /* Global matches always start at index 0, but Sticky matches may have a non-zero lastIndex. */
-  if (ctx_p->index > 0)
+  if ((ctx_p->flags & RE_FLAG_GLOBAL) == 0 && (re_ctx.flags & RE_FLAG_STICKY) != 0)
```
rerobika, line 2535: How about `(ctx_p->flags & (RE_FLAG_GLOBAL | RE_FLAG_STICKY)) ==  RE_FLAG_GLOBAL`?

tests/jerry/es.next/symbol-replace.js
```diff
@@ -682,7 +682,6 @@ Object.defineProperty(abruptStickyRegexp, 'sticky', {
 
 try {
   abruptStickyRegexp[Symbol.replace]();
-  assert(false);
 } catch (e) {
-  assert(e === "abrupt sticky");
+  assert(false);
```
rerobika, line 686: Since this should not throw error anymore, how about adding an assert for it's expected result?
`assert(abruptStickyRegexp[Symbol.replace]() === 'undefinedndefined');`

jerry-core/ecma/operations/ecma-regexp-object.c
dbatyai: The two flags come from different contexts. The global and unicode flags come from the replace context, and are coerced from the global and unicode poperties of the 'this' object, the rest of the flags should be used based on the [[OriginalFlags]] internal slot of the regexp instance.

rerobika: Indeed, I've missed that part.

PR #4007, opened 2020-07-16
Update ToLength operation to conform ES6 spec

rerobika:
JerryScript-DCO-1.0-Signed-off-by: Robert Fancsik frobert@inf.u-szeged.hu

PR #4009, opened 2020-07-16
Remove the usage of ecma_length_t

rerobika:
Now the following conventions are applied:
 - passing the number of arguments for a function call is always `uint32_t`
 - string size/length/position related operation should use `lit_utf8_size_t`
 - extended objects internal fields must be `uint32_t`


rerobika:
Related PR: #4007.

Issue #2931, opened 2019-06-28
Discussion: Report unhandled promise rejections


dodgerblue:
Hello,

If my understanding is correct, in the jerryscript implementation, if an error is thrown when resolving a promise, it will only be passed to the rejection handler (if there is one registered). If there is no promise rejection registered, the error is silently ignored. I am looking for a way to make that error return to the calling function, or some other way to signal it.

My approach was to change `ecma_reject_promise()` to take an additional out parameter telling whether the rejection was handled or not. `ecma_promise_trigger_reactions()` sets this parameter to `false` in case the reactions collection was empty from the start. Based on this, `ecma_promise_reject_handler()` will return an error for unhandled rejections.

This works for a snapshot that was loaded, with all the chained promises' reactions registered. In this situation, when a `.then()` handler executes, we can determine whether it has a subsequent `.catch()` rejection handler chained. If it doesn't, `ecma_reject_handler()` returns with an error as expected.

This doesn't work for direct function calls. For instance, a scenario like `foo().then()`, where `foo()` calls into a C implementation that returns a rejected promise immediately via `jerry_resolve_or_reject_promise()`. At this point, I noticed `ecma_promise_do_then()` wasn't called yet to register the handlers and schedule the reactions.

I would like opinions on this approach. Do you think the current jerry design allows to signal unhandled errors thrown during exceptions, or it's always up the the developer to chain a `.catch()` call for thrown errors?

Thank you!

akosthekiss:
Hi @dodgerblue ! Do you have a (preferably small) test case (a js script) that shows what you'd like to achieve?

The goal of JerryScript is to be spec-conforming, in this case ES2015-conforming. I'm not sure that what you outlined is proper use of ES2015 promises. (As a sanity check: what you'd like to achieve, does that work in another engine, e.g., in v8?)

dodgerblue:
Hey @akosthekiss! Here's an example:

	new Promise(function(resolve, reject) {
	  resolve("Check");
	}).then(function(val) {
	  print(val);
	  throw new Error("Boom");
	});

If I run this in jerry (built with `--profile=es2015-subset`), I will only get the `"Check"` print statement on screen, and no info on the error. If I chain a `.catch()` function at the end, the error is correctly passed to it.

If I run above code in node (`node --version v8.10.0`) - and change the `print` to a `console.log` call - I get the error reported as an unhandled promise rejection:

	Check
[...]

PR #4012, opened 2020-07-17
Add promise rejection tracker support

rerobika:

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -619,6 +619,30 @@ typedef jerry_value_t (*jerry_vm_exec_stop_callback_t) (void *user_p);
 
 - [jerry_set_vm_exec_stop_callback](#jerry_set_vm_exec_stop_callback)
 
+## jerry_promise_rejection_tracker_t
+
+**Summary**
+
+The typical implementation of the function might try to notify developers of unhandled rejections.
+The callback is called in two scenarios:
+ - When a promise is rejected without any handlers
+ - When a handler is added to a rejected promise for the first time
+In these cases the callback is invoked with the following arguments:
+ - `promise`: rejected promise
+ - `reason`: reason of the rejection
+
+**Prototype**
+
+```c
+typedef void (*jerry_promise_rejection_tracker_t) (jerry_value_t promise, jerry_value_t reason);
+```
+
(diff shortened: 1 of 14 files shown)


rerobika:
Related feature request: #2931 .

dbatyai:
I feel like it would be better to handle this as a port function instead of a callback that can be set via the API.

tests/unit-core/test-promise.c
```diff
@@ -158,6 +174,34 @@ main (void)
...
+  TEST_ASSERT (!jerry_value_is_error (my_promise4));
+  TEST_ASSERT (tracter_counter == 2);
+  TEST_ASSERT (jerry_get_boolean_value (jerry_binary_operation (JERRY_BIN_OP_STRICT_EQUAL,
+                                                                last_rejected_promise,
+                                                                my_promise4)));
+  TEST_ASSERT (jerry_value_is_object (last_rejected_reason));
+  TEST_ASSERT (jerry_get_error_type (last_rejected_reason) == JERRY_ERROR_TYPE);
+  jerry_release_value (last_rejected_promise);
+  jerry_release_value (last_rejected_reason);
+
```
galpeter, line 204: There should be also a test where the "global" reject handler is not invoked as the reject was handled already.

tests/unit-core/test-promise.c
```diff
@@ -30,8 +30,11 @@ static const jerry_char_t test_source[] = TEST_STRING_LITERAL (
 );
 
 static int count_in_assert = 0;
+static int tracter_counter = 0;
```
galpeter, line 33: What does "tracter" mean?

docs/02.API-REFERENCE.md
```diff
@@ -7692,6 +7716,67 @@ main (void)
 - [jerry_run](#jerry_run)
 - [jerry_vm_exec_stop_callback_t](#jerry_vm_exec_stop_callback_t)
 
+## jerry_set_promise_rejection_callback
+
+**Summary**
+
+Register a promise rejection tracker callback for the engine.
```
galpeter, line 7723: We should provide more details here imho and also mention the `HostPromiseRejectionTracker` from the standard here.

rerobika:
The PR should be reworked as a port function, so closing the PR for now.

PR #4020, opened 2020-07-20
Fix date-construct.js test

ossy-szeged:
Removed an undeterministic subtest.

date-construct.js fails intermittently, because it expects that
two Date() calls return the same date string. But the second one
can be 1 second later. This subtest is undeterministic and already
covered by test262/es2015/test/built-ins/Date/S15.9.2.1_A2.js with
1 second epsilon.

### DIFF
--- a/tests/jerry/date-construct.js
+++ b/tests/jerry/date-construct.js
@@ -75,4 +75,3 @@ catch (e)
 
 assert (typeof Date (2015) == "string");
 assert (typeof Date() != typeof (new Date ()));
-assert (Date (Number.NaN) == Date ());


LaszloLango:
Does this test depends on the speed of the host device?

ossy-szeged:
> Does this test depend on the speed of the host device?

Yes, but two Date() call should happen in 1 second always on all platform. The problem was that
we can get different seconds if there is only 1 milliseconds between Date() calls, for example:

jerry> Date()
Mon Jul 20 2020 14:02:48 GMT+0000
jerry> Date()
Mon Jul 20 2020 14:02:49 GMT+0000

LaszloLango:
> there is only 1 milliseconds between Date() calls

I don't see who or what guarantees this. IMHO, if it cannot be tested in a deterministic way, then simply remove the test.

ossy-szeged:
> > there is only 1 milliseconds between Date() calls
> I don't see who or what guarantees this. IMHO, if it cannot be tested in a deterministic way, then simply remove the test.

Nobody can guarantees that two Date() calls run in 1 millisecond in any JS engine. But if they run in a different millisecond, the second part of the time can be different - bigger with 1 second. It's unlikely in the real world that two Date() calls takes more than a second.

We do need a way to test that Date() and Date(value) calls return "now date string". There is no better way to compare them to each other with a reasonable epsilon after parsing the result. But it would be better to compare both of them to Date.now(). I don't think we should drop this functionality test  completely if it can be fixed.

But after the answer, I found that test262/.../S15.9.2.1_A2.js test do the same check with 1 second epsilon, so we really don't need to duplicate this test, so I'm going to remove this subtest.

PR #4032, opened 2020-07-22
Implement Object.prototype.{__defineGetter__, __defineSetter__}

szilagyiadam:
__defineGetter__ is based on ECMA-262 v11, B.2.2.2
__defineSetter__ is based on ECMA-262 v11, B.2.2.3


jerry-core/ecma/builtin-objects/ecma-builtin-object-prototype.c
```diff
@@ -203,6 +207,64 @@ ecma_builtin_object_prototype_object_property_is_enumerable (ecma_object_t *obj_
...
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Getter is not callable."));
+  }
+
+  ecma_object_t *accessor_obj_p = ecma_get_object_from_value (accessor);
+  ecma_ref_object (accessor_obj_p);
+
+  /* 3. */
+  ecma_property_descriptor_t desc = ecma_make_empty_property_descriptor ();
+  desc.flags |= (ECMA_PROP_IS_ENUMERABLE | ECMA_PROP_IS_CONFIGURABLE
+                 | ECMA_PROP_IS_ENUMERABLE_DEFINED | ECMA_PROP_IS_CONFIGURABLE_DEFINED);
```
dbatyai, line 240: All of these should be on separate lines.

jerry-core/ecma/builtin-objects/ecma-builtin-object-prototype.c
```diff
@@ -203,6 +207,64 @@ ecma_builtin_object_prototype_object_property_is_enumerable (ecma_object_t *obj_
...
+                                                                             false - defineSetter method */
+{
+  /* 2. */
+  if (!ecma_op_is_callable (accessor))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Getter is not callable."));
+  }
+
+  ecma_object_t *accessor_obj_p = ecma_get_object_from_value (accessor);
+  ecma_ref_object (accessor_obj_p);
```
dbatyai, line 235: I don't think we need to add a reference here, the caller will keep the object alive. Then we don't need the free on line 256 either.

PR #4060, opened 2020-07-27
Fix new.target saving in OrdinaryObject.[[Construct]]

### LINKED ISSUE #4059: Assertion 'JERRY_CONTEXT (current_new_target) == NULL' in ecma_finalize
###### JerryScript revision
da5b058

###### Build platform
Ubuntu 16.04.6 LTS (Linux 4.15.0-99-generic x86_64)

###### Build steps
```
 --compile-flag=-fno-common --compile-flag=-g --strip=off \
 --system-allocator=on --logging=on --linker-flag=-fuse-ld=gold \
 --error-messages=on --profile=es2015-subset --lto=off 
```
###### Test case
```
function print(value){}

print(String.fromCharCode(new new Proxy(Function(), { get:String.fromCharCode}), 0, false, 67));

```
###### Output
```
Script Error: TypeError: Cannot convert a Symbol value to a number.
ICE: Assertion 'JERRY_CONTEXT (current_new_target) == NULL' failed at /home/jerryscript/jerry-core/ecma/base/ecma-init-finalize.c(ecma_finalize):76.
Aborted (core dumped)

```

Credits: This vulnerability is detected by chong from OWL337.

rerobika:
This patch fixes #4059.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -1336,10 +1336,6 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
     return ecma_raise_type_error (message_p);
   }
 
-  /* 6. */
-  ecma_object_t *old_new_target_p = JERRY_CONTEXT (current_new_target);
-  JERRY_CONTEXT (current_new_target) = new_target_p;
-
   /* 5. */
   if (!ECMA_GET_THIRD_BIT_FROM_POINTER_TAG (ext_func_obj_p->u.function.scope_cp))
   {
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1350,6 +1350,9 @@ ecma_op_function_construct (ecma_object_t *func_obj_p, /**< Function object */
     /* 5.b */
     if (JERRY_UNLIKELY (proto_p == NULL))
     {
+#if ENABLED (JERRY_ESNEXT)
```
galpeter, line 1353: I was wondering: if we move the `1340-1341` lines from above just before the `.._call_simple` to avoid adding this here ? Would that change anything in the `ecma_op_get_prototype_from_constructor` or `ecma_create_object` internals?

rerobika: That's a good idea.

rerobika:

PR #4062, opened 2020-07-27
Implement BigInt primitve type and some of its operations

zherczeg:
Supported operations:
- parse BigInt (decimal, hexadecimal, binary)
- toString with any radix between 2 and 36
- arithmetic operations: negate, add, subtract, multiply
- left and right shift


zherczeg:
Thank you for the review. Patch is updated.

jerry-core/ecma/operations/ecma-big-uint.c
```diff
@@ -0,0 +1,1252 @@
...
+  if (JERRY_UNLIKELY (size > ECMA_BIGINT_MAX_SIZE))
+  {
+    return NULL;
+  }
+
+  ecma_extended_primitive_t *value_p;
+
+  if (size == 0)
+  {
+    value_p = (ecma_extended_primitive_t *) jmem_pools_alloc (sizeof (ecma_extended_primitive_t));
```
dbatyai, line 48: Are zero sized bigints that common to have a special case for them?

jerry-core/ecma/operations/ecma-bigint.c
```diff
@@ -0,0 +1,508 @@
...
+/**
+ * Raise a not enough memory error
+ *
+ * @return ECMA_VALUE_ERROR
+ *         Returned value must be freed with ecma_free_value.
+ */
+static ecma_value_t
+ecma_bigint_raise_memory_error (void)
+{
+  return ecma_raise_range_error (ECMA_ERR_MSG ("Cannot allocate memory for a BigInt value"));
```
dbatyai, line 33: Generally we used to terminate the engine when we run out of memory, why do we handle bigints differently?

zherczeg: I think this is useful feature, and it does not need too much effort, similar to parser.

jerry-core/ecma/operations/ecma-big-uint.c
zherczeg: This code is copied from external API error handling. On the long term I have been thinking about making bigint zero a special constant.

jerry-core/ecma/builtin-objects/ecma-builtin-bigint-prototype.c
```diff
@@ -0,0 +1,132 @@
...
+
+  uint32_t radix = 10;
+
+  if (arguments_list_len > 0 && !ecma_is_value_undefined (arguments_list_p[0]))
+  {
+    ecma_number_t arg_num;
+
+    if (ECMA_IS_VALUE_ERROR (ecma_op_to_integer (arguments_list_p[0], &arg_num)))
+    {
+      return ECMA_VALUE_ERROR;
```
dbatyai, line 104: `bigint` is leaking here.

jerry-core/ecma/builtin-objects/ecma-builtin-bigint-prototype.c
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    if (arg_num < 2 || arg_num > 36)
+    {
+      return ecma_raise_range_error (ECMA_ERR_MSG ("Radix must be between 2 and 36."));
```
dbatyai, line 109: And here.

jerry-core/ecma/builtin-objects/ecma-builtin-bigint-prototype.c
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    radix = (uint32_t) arg_num;
+  }
+
+  ecma_string_t *string_p = ecma_bigint_to_string (bigint, radix);
+
+  if (string_p == NULL)
+  {
+    return ECMA_VALUE_ERROR;
```
dbatyai, line 119: And here as well.

zherczeg:
Free values added

PR #4080, opened 2020-07-28
Fix block-scoped var/function redeclarations

dbatyai:


jerry-core/parser/js/js-scanner-util.c
```diff
@@ -1382,6 +1382,13 @@ scanner_detect_invalid_var (parser_context_t *context_p, /**< context */
   }
 
   scanner_literal_pool_t *literal_pool_p = scanner_context_p->active_literal_pool_p;
+
+  if (literal_pool_p->status_flags & SCANNER_LITERAL_POOL_BLOCK
+      && (var_literal_p->type & (SCANNER_LITERAL_IS_FUNC)))
```
zherczeg, line 1387: I am not sure this is correct. `SCANNER_LITERAL_IS_FUNC` can be set if it was declared inside.
Is this working: { { function f() {} } var f }`

dbatyai:
Thanks @zherczeg, patch is updated.

PR #4084, opened 2020-07-29
Fix build with disabled ES2015 module system

galpeter:
With a disabled ES2015 module system the build fails as an enum value
is incorrectly guarded and used.

### DIFF
--- a/jerry-core/parser/js/js-parser-statm.c
+++ b/jerry-core/parser/js/js-parser-statm.c
@@ -680,10 +680,11 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
   {
     parser_raise_error (context_p, PARSER_ERR_VARIABLE_REDECLARED);
   }
+
+  uint16_t function_name_index = context_p->lit_object.index;
 #endif /* ENABLED (JERRY_ESNEXT) */
 
 #if ENABLED (JERRY_MODULE_SYSTEM)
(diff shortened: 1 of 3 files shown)


ossy-szeged:
LGTM (informal)

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -804,9 +804,9 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
   literal_p->u.bytecode_p = compiled_code_p;
   literal_p->type = LEXER_FUNCTION_LITERAL;
 
-#if ENABLED (JERRY_ESNEXT)
+#if ENABLED (JERRY_MODULE_SYSTEM)
   parser_compiled_code_set_function_name (context_p, compiled_code_p, function_name_index, 0);
-#endif /* ENABLED (JERRY_ESNEXT) */
+#endif /* ENABLED (JERRY_MODULE_SYSTEM) */
```
dbatyai, line 809: Are you sure this should be module guarded? This seems more related to function names than modules.

galpeter: True, with this the function will not have a name in case of disabled module system. Will update.

galpeter: @dbatyai I've update the PR.

PR #4086, opened 2020-07-30
Fix TypeError name property value

galpeter:
The TypeError.name incorrectly returned the "SyntaxError" value.


ossy-szeged:
LGTM (informal)

PR #4101, opened 2020-08-03
Add Symbol support to integrity testing routine

rwalczyna:
Properties with Symbol caused assertion to fail


jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -596,7 +596,17 @@ ecma_builtin_object_test_integrity_level (ecma_object_t *obj_p, /**< routine's a
 
   for (uint32_t i = 0; i < props_p->item_count; i++)
   {
-    ecma_string_t *property_name_p = ecma_get_string_from_value (buffer_p[i]);
```
rerobika, line 599: You can use `ecma_get_prop_name_from_value` :)

PR #4102, opened 2020-08-03
Implement TrimStart TrimEnd and aliases

Achie72:
Based on: https://tc39.es/ecma262/#sec-string.prototype.trim


jerry-core/lit/lit-magic-strings.h
```diff
@@ -46,6 +46,8 @@ typedef enum
   LIT_INTERNAL_MAGIC_STRING_MAP_PROTOTYPE_ENTRIES, /**< Map.prototype entries and [@@iterator] routines */
   LIT_INTERNAL_MAGIC_THIS_BINDING_VALUE, /**< FunctionEnvironmentRecord [[ThisBindingValue]] internal slot */
   LIT_INTERNAL_MAGIC_PROMISE_CAPABILITY, /**< PromiseCapability record */
+  LIT_INTERNAL_MAGIC_STRING_TRIM_START,
+  LIT_INTERNAL_MAGIC_STRING_TRIM_END,
```
rerobika, line 50: Missing descriptions.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.inc.h
```diff
@@ -76,6 +76,11 @@ ROUTINE (LIT_MAGIC_STRING_CODE_POINT_AT, ECMA_STRING_PROTOTYPE_CODE_POINT_AT, 1,
 ROUTINE (LIT_MAGIC_STRING_PAD_START,  ECMA_STRING_PROTOTYPE_PAD_START, 2, 1)
 ROUTINE (LIT_MAGIC_STRING_PAD_END,  ECMA_STRING_PROTOTYPE_PAD_END, 2, 1)
 ROUTINE (LIT_GLOBAL_SYMBOL_ITERATOR, ECMA_STRING_PROTOTYPE_ITERATOR, 0, 0)
+
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_START, LIT_INTERNAL_MAGIC_STRING_TRIM_START, ECMA_PROPERTY_FIXED)
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_LEFT, LIT_INTERNAL_MAGIC_STRING_TRIM_START, ECMA_PROPERTY_FIXED)
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_END, LIT_INTERNAL_MAGIC_STRING_TRIM_END, ECMA_PROPERTY_FIXED)
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_RIGHT, LIT_INTERNAL_MAGIC_STRING_TRIM_END, ECMA_PROPERTY_FIXED)
```
rerobika, line 83: These routines are not fixed. see `Object.getOwnPropertyDescriptor(String.prototype, "trimLeft")`

jerry-core/ecma/builtin-objects/ecma-builtin-intrinsic.c
```diff
@@ -236,7 +254,6 @@ ecma_builtin_intrinsic_dispatch_routine (uint16_t builtin_routine_id, /**< built
 
       ECMA_FINALIZE_UTF8_STRING (string_buff, string_buff_size);
       ecma_deref_ecma_string (str_p);
-
```
rerobika, line 239: Do not remove this line.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -96,6 +96,16 @@ typedef enum
...
 
+/**
+ * Enumerable for string trimming location.
+ */
+typedef enum
+{
+  ECMA_TRIM_STRING_START, /**< Trim only from the start */
+  ECMA_TRIM_STRING_END, /**<  Trim only from the end */
+  ECMA_TRIM_STRING_STARTEND, /**< Trim from both start and end */
+} ecma_trim_string_where;
```
rerobika, line 107: ecma_trim_string_where -> `ecma_string_trim_flags_t`

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2458,17 +2457,30 @@ ecma_string_trim_helper (const lit_utf8_byte_t **utf8_str_p, /**< [in, out] curr
   while (current_p > nonws_start_p)
   {
     read_size = lit_read_prev_code_unit_from_utf8 (current_p, &ch);
-
     if (!lit_char_is_white_space (ch))
     {
       break;
     }
-
```
rerobika, line 2466: Do not remove these lines

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -96,6 +96,16 @@ typedef enum
   ECMA_STRING_FLAG_MUST_BE_FREED = (1 << 3), /**< The returned buffer must be freed */
 } ecma_string_flag_t;
 
+/**
+ * Enumerable for string trimming location.
+ */
+typedef enum
+{
+  ECMA_TRIM_STRING_START, /**< Trim only from the start */
```
rerobika, line 104: I think these should be flags.

```c
ECMA_TRIM_STRING_START = (1 << 0),
ECMA_TRIM_STRING_END = (1u << 1),
```

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2458,17 +2457,30 @@ ecma_string_trim_helper (const lit_utf8_byte_t **utf8_str_p, /**< [in, out] curr
...
+    {
+      *utf8_str_size = (lit_utf8_size_t) ((*utf8_str_p + *utf8_str_size) - nonws_start_p);
+    }
+    else
+    {
+      *utf8_str_size = (lit_utf8_size_t) (current_p - nonws_start_p);
+    }
+
+    *utf8_str_p = nonws_start_p;
+  }
```
rerobika, line 2483: I think this should be performed in an other way. Currently you trim the string from both directions after you calculate back the required string part which is superfluous.

```c
if (flags & TRIM_START) {
  //  remove leading whitespaces
}

if (flags & TRIM_END) {
  //  remove trailing whitespaces
}

// return as before
*utf8_str_p = nonws_start_p;
*utf8_str_size = (lit_utf8_size_t) (current_p - nonws_start_p);
```

jerry-core/ecma/builtin-objects/ecma-builtin-intrinsic.c
```diff
@@ -206,6 +210,20 @@ ecma_builtin_intrinsic_dispatch_routine (uint16_t builtin_routine_id, /**< built
...
       return ecma_date_value_to_utc_string (*prim_value_p);
     }
+    case ECMA_INTRINSIC_STRING_TRIM_START:
+    {
+      ecma_string_t *to_str_p = ecma_op_to_string (this_arg);
+      ecma_value_t result = ecma_make_string_value (ecma_string_trim (to_str_p, ECMA_TRIM_STRING_START));
+      ecma_deref_ecma_string (to_str_p);
+      return result;
+    }
+    case ECMA_INTRINSIC_STRING_TRIM_END:
```
rerobika, line 220: Please combine these two cases.

jerry-core/ecma/builtin-objects/ecma-builtin-intrinsic.c
```diff
@@ -206,6 +210,20 @@ ecma_builtin_intrinsic_dispatch_routine (uint16_t builtin_routine_id, /**< built
...
+    {
+      ecma_string_t *to_str_p = ecma_op_to_string (this_arg);
```
rerobika, line 222: This operation may throw error.

jerry-core/ecma/builtin-objects/ecma-builtin-intrinsic.c
```diff
@@ -44,7 +46,9 @@ enum
   ECMA_INTRINSIC_ARRAY_TO_STRING,
   ECMA_INTRINSIC_DATE_TO_UTC_STRING,
   ECMA_INTRINSIC_PARSE_FLOAT,
-  ECMA_INTRINSIC_PARSE_INT
+  ECMA_INTRINSIC_PARSE_INT,
+  ECMA_INTRINSIC_STRING_TRIM_START,
+  ECMA_INTRINSIC_STRING_TRIM_END
```
rerobika, line 51: Put a comma for the end.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.inc.h
Achie72: Already working on this, but thanks for the heads up.

jerry-core/ecma/base/ecma-helpers-string.c
Achie72: My problem with this atm, is that if i don't set the `nonws_start_p` in the forward trimming part, than the second part, which checks for `current_p > nonws_start_p` will never pass, as `nonws_start_p` and `current_p` will both be set to `*utf8_str_p + *utf8_str_size`. This will lead to the `utf8_str_p` to be set to the end of the string on trimend functions.

This is the reason why i do this rather ugly code atm. Although i could set `nonws_start_p` to the original `utf8_str_p` when i'm doing trimEnd, which will lead to the if passing, and as i don't really care about what is on the front of the string, this might be a solution for this. Any thoughts?

rerobika: I think this can be resolved with 3 variables. start, current, end

then:

```
current = start;

while (current < end) {
  //trim
}

start = current;
current = end;

while (current < end) {
 // trim
}
*utf8_str_p = start_p;
*utf8_str_size = (lit_utf8_size_t) (end_p - start_p);
```

dbatyai: Why don't we split the trim helper into `ecma_string_trim_front`/`ecma_string_trim_back`?

Achie72: With a nicer solution than mine (such as @rerobika's idea) it would be unnecessary to cut this, even more so if only these functions use the front/end trimming part imo.

dbatyai: The question isn't whether it is necessary or not, it's whether we should. We don't gain anything by keeping the trimming code in a single function, even if separated nothing would be duplicate code. In it's current form all we do is add superfluous conditions, which wouldn't be necessary if they are split properly.

Achie72: Okay, i'll split this into corresponding sub-functions.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -96,6 +96,16 @@ typedef enum
...
 
+/**
+ * Enumerable for string trimming location.
+ */
+typedef enum
+{
+  ECMA_TRIM_STRING_START = (1u << 0), /**< Trim only from the start */
+  ECMA_TRIM_STRING_END = (1u << 1), /**<  Trim only from the end */
+  ECMA_TRIM_STRING_STARTEND = (1u << 2), /**< Trim from both start and end */
+} ecma_string_trim_flags_t;
```
dbatyai, line 107: No need for these, just call `ecma_string_trim_front`/`ecma_string_trim_back` where necessary, and change `ecma_string_trim_helper` to call both, and be always inline.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2453,22 +2447,59 @@ ecma_string_trim_helper (const lit_utf8_byte_t **utf8_str_p, /**< [in, out] curr
...
-  {
-    read_size = lit_read_prev_code_unit_from_utf8 (current_p, &ch);
+/**
+ * Helper function for trimming.
+ *
+ * Used by:
+ *        - ecma_string_trim_helper
+ */
+void ecma_string_trim_back (const lit_utf8_byte_t **utf8_str_p, /**< [in, out] current string position */
+                             lit_utf8_size_t *utf8_str_size)  /**< [in, out] size of the given string */
```
dbatyai, line 2461: I suggest to change the prototype of the new functions to be
```c
const lit_utf8_byte_t *
ecma_string_trim_front (const lit_utf8_byte_t *start_p, const lit_utf8_byte_t *end_p);
const lit_utf8_byte_t *
ecma_string_trim_back (const lit_utf8_byte_t *start_p, const lit_utf8_byte_t *end_p);
```
where the return value is the new start/end position respectively.

jerry-core/ecma/builtin-objects/ecma-builtin-intrinsic.inc.h
```diff
@@ -74,6 +74,8 @@ ROUTINE (LIT_INTERNAL_MAGIC_STRING_ARRAY_PROTOTYPE_VALUES, ECMA_INTRINSIC_ARRAY_
 ROUTINE (LIT_INTERNAL_MAGIC_STRING_TYPEDARRAY_PROTOTYPE_VALUES, ECMA_INTRINSIC_TYPEDARRAY_PROTOTYPE_VALUES, 0, 0)
 ROUTINE (LIT_INTERNAL_MAGIC_STRING_SET_PROTOTYPE_VALUES, ECMA_INTRINSIC_SET_PROTOTYPE_VALUES, 0, 0)
 ROUTINE (LIT_INTERNAL_MAGIC_STRING_MAP_PROTOTYPE_ENTRIES, ECMA_INTRINSIC_MAP_PROTOTYPE_ENTRIES, 0, 0)
+ROUTINE (LIT_INTERNAL_MAGIC_STRING_TRIM_START, ECMA_INTRINSIC_STRING_TRIM_START, 0, 0)
+ROUTINE (LIT_INTERNAL_MAGIC_STRING_TRIM_END, ECMA_INTRINSIC_STRING_TRIM_END, 0, 0)
```
dbatyai, line 78: Routine name (1st argument) should be `LIT_MAGIC_STRING_TRIM_START` and `LIT_MAGIC_STRING_TRIM_END` respectively.

jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.inc.h
```diff
@@ -76,6 +76,15 @@ ROUTINE (LIT_MAGIC_STRING_CODE_POINT_AT, ECMA_STRING_PROTOTYPE_CODE_POINT_AT, 1,
...
 ROUTINE (LIT_GLOBAL_SYMBOL_ITERATOR, ECMA_STRING_PROTOTYPE_ITERATOR, 0, 0)
+
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_START, LIT_INTERNAL_MAGIC_STRING_TRIM_START,
+                    ECMA_PROPERTY_CONFIGURABLE_WRITABLE)
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_LEFT, LIT_INTERNAL_MAGIC_STRING_TRIM_START,
+                    ECMA_PROPERTY_CONFIGURABLE_WRITABLE)
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_END, LIT_INTERNAL_MAGIC_STRING_TRIM_END,
+                    ECMA_PROPERTY_CONFIGURABLE_WRITABLE)
+INTRINSIC_PROPERTY (LIT_MAGIC_STRING_TRIM_RIGHT, LIT_INTERNAL_MAGIC_STRING_TRIM_END,
+                    ECMA_PROPERTY_CONFIGURABLE_WRITABLE)
```
dbatyai, line 87: Second args should be the non-internal magic string equivalents.

jerry-core/lit/lit-magic-strings.h
```diff
@@ -46,6 +46,8 @@ typedef enum
   LIT_INTERNAL_MAGIC_STRING_MAP_PROTOTYPE_ENTRIES, /**< Map.prototype entries and [@@iterator] routines */
   LIT_INTERNAL_MAGIC_THIS_BINDING_VALUE, /**< FunctionEnvironmentRecord [[ThisBindingValue]] internal slot */
   LIT_INTERNAL_MAGIC_PROMISE_CAPABILITY, /**< PromiseCapability record */
+  LIT_INTERNAL_MAGIC_STRING_TRIM_START, /**< String.Prototype.trimStart intrinsic routine */
+  LIT_INTERNAL_MAGIC_STRING_TRIM_END,  /**< String.Prototype.trimEnd intrinsic routine */
```
dbatyai, line 50: These should not be added.

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -397,9 +397,12 @@ lit_magic_string_id_t ecma_get_string_magic (const ecma_string_t *string_p);
 
 lit_string_hash_t ecma_string_hash (const ecma_string_t *string_p);
 ecma_string_t *ecma_string_substr (const ecma_string_t *string_p, lit_utf8_size_t start_pos, lit_utf8_size_t end_pos);
+const lit_utf8_byte_t * ecma_string_trim_front (const lit_utf8_byte_t *start_p, const lit_utf8_byte_t *end_p);
+const lit_utf8_byte_t * ecma_string_trim_back (const lit_utf8_byte_t *start_p, const lit_utf8_byte_t *end_p);
 void ecma_string_trim_helper (const lit_utf8_byte_t **utf8_str_p,
                               lit_utf8_size_t *utf8_str_size);
 ecma_string_t *ecma_string_trim (const ecma_string_t *string_p);
+
```
rerobika, line 405: Please remove this change.

jerry-core/ecma/builtin-objects/ecma-builtin-intrinsic.c
rerobika: This request is still valid.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2426,49 +2426,79 @@ ecma_string_substr (const ecma_string_t *string_p, /**< pointer to an ecma strin
...
+    lit_utf8_size_t read_size = lit_read_code_unit_from_utf8 (start_p, &ch);
 
     if (!lit_char_is_white_space (ch))
     {
-      nonws_start_p = current_p;
       break;
     }
-
-    current_p += read_size;
+    start_p += read_size;
```
dbatyai, line 2447: Please add a newline after the `if` block.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2426,49 +2426,79 @@ ecma_string_substr (const ecma_string_t *string_p, /**< pointer to an ecma strin
...
-    read_size = lit_read_prev_code_unit_from_utf8 (current_p, &ch);
+    lit_utf8_size_t read_size = lit_read_prev_code_unit_from_utf8 (end_p, &ch);
 
     if (!lit_char_is_white_space (ch))
     {
       break;
     }
-
-    current_p -= read_size;
+    end_p -= read_size;
```
dbatyai, line 2475: Newline after the `if` block.

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2426,49 +2426,79 @@ ecma_string_substr (const ecma_string_t *string_p, /**< pointer to an ecma strin
...
+ *        - ecma_utf8_string_to_number
+ *        - ecma_builtin_global_object_parse_int
+ *        - ecma_builtin_global_object_parse_float
+ */
+inline void JERRY_ATTR_ALWAYS_INLINE
+ecma_string_trim_helper (const lit_utf8_byte_t **utf8_str_p, /**< [in, out] current string position */
+                         lit_utf8_size_t *utf8_str_size)  /**< [in, out] size of the given string */
+{
+  const lit_utf8_byte_t *end = *utf8_str_p + *utf8_str_size;
+  const lit_utf8_byte_t *start = *utf8_str_p;
```
dbatyai, line 2495: `end_p`, `start_p`

jerry-core/ecma/base/ecma-helpers.h
```diff
@@ -397,6 +397,8 @@ lit_magic_string_id_t ecma_get_string_magic (const ecma_string_t *string_p);
 
 lit_string_hash_t ecma_string_hash (const ecma_string_t *string_p);
 ecma_string_t *ecma_string_substr (const ecma_string_t *string_p, lit_utf8_size_t start_pos, lit_utf8_size_t end_pos);
+const lit_utf8_byte_t * ecma_string_trim_front (const lit_utf8_byte_t *start_p, const lit_utf8_byte_t *end_p);
+const lit_utf8_byte_t * ecma_string_trim_back (const lit_utf8_byte_t *start_p, const lit_utf8_byte_t *end_p);
```
dbatyai, line 401: There shouldn't be whitespace between the asterix and the function name.

dbatyai (changes requested):

jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2426,49 +2426,79 @@ ecma_string_substr (const ecma_string_t *string_p, /**< pointer to an ecma strin
...
+} /* ecma_string_trim_back */
+
+/**
+ * Helper function for trimming.
+ *
+ * Used by:
+ *        - ecma_string_trim
+ *        - ecma_utf8_string_to_number
+ *        - ecma_builtin_global_object_parse_int
+ *        - ecma_builtin_global_object_parse_float
```
ossy-szeged, line 2488: parseInt and parseFloat should only use trimStart by spec

Achie72: This is a nice catch i did not look into. Although roughly looking at them this should not alter as the code checks for digits when iterating and in the end, the `ecma_utf8_string_to_number` call the whole trim in itself as well. I'm looking into our test suites to check for fails if i modify this.

dbatyai (approved):
LGTM

PR #4119, opened 2020-08-10
Minor fix in ecma_builtin_array_prototype_object_slice

szilagyiadam:
During the property creation the is_throw flag was not used before

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -948,10 +948,13 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
     {
       /* 10.c.ii */
       ecma_value_t put_comp;
+
+      uint32_t opts = (ECMA_PROPERTY_CONFIGURABLE_ENUMERABLE_WRITABLE | ECMA_PROP_IS_THROW);
       put_comp = ecma_builtin_helper_def_prop_by_index (new_array_p,
                                                         n,
                                                         get_value,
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -948,10 +948,13 @@ ecma_builtin_array_prototype_object_slice (ecma_value_t arg1, /**< start */
...
       /* 10.c.ii */
       ecma_value_t put_comp;
+
+      uint32_t opts = (ECMA_PROPERTY_CONFIGURABLE_ENUMERABLE_WRITABLE | ECMA_PROP_IS_THROW);
       put_comp = ecma_builtin_helper_def_prop_by_index (new_array_p,
                                                         n,
                                                         get_value,
-                                                        ECMA_PROPERTY_CONFIGURABLE_ENUMERABLE_WRITABLE);
+                                                        opts);
+
```
ossy-szeged, line 957: It should throw only in ESnext case, but not in ES5.1

tests/jerry/es.next/array-prototype-slice.js
```diff
@@ -53,3 +53,36 @@ var c = new ExactArray (5);
...
 assert (JSON.stringify (sliced3) == '["baz","baz","baz","baz","baz"]');
+
+
+var func = function() {
+  Object.defineProperty(this, "0", {
+    set: function(value) {},
+    configurable: false,
+  });
+};
+
```
ossy-szeged, line 64: Why do we need these new test cases? Aren't they already covered by test262?

PR #4120, opened 2020-08-10
Optimize test262 runner

ossy-szeged:
Changes:
- Add new option to run-tests.py: --test262-test-list to run selected tests only
- Fix exclude list updater accordingly
- Run ESNext tests on GitHub CI in two batches to decrease runtime


tools/runners/run-test-suite-test262.py
```diff
@@ -189,6 +197,10 @@ def main(args):
     if 'excludelist_path' in args and args.mode == 'default':
         test262_command.extend(['--exclude-list', args.excludelist_path])
 
+    if args.test262_test_list:
+        for test in args.test262_test_list.split(','):
```
galpeter, line 201: in python you can do: `test262_command.extend(args.test262_test_list.split(','))` to achieve the same thing.

PR #4124, opened 2020-08-10
Minor fix to ecma_builtin_object_object_define_property

szilagyiadam:
DefineOwnProperty result could be false, and we should throw an error in that case.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-object.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-object.c
@@ -1019,6 +1019,13 @@ ecma_builtin_object_object_define_property (ecma_object_t *obj_p, /**< routine's
     return define_own_prop_ret;
   }
 
+  if (ecma_is_value_false (define_own_prop_ret))
+  {
+    return ecma_raise_type_error (ECMA_ERR_MSG ("The requested property update cannot be performed."));
+  }
+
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-object.c
```diff
@@ -1019,6 +1019,11 @@ ecma_builtin_object_object_define_property (ecma_object_t *obj_p, /**< routine's
     return define_own_prop_ret;
   }
 
+  if (!ecma_is_value_true (define_own_prop_ret))
```
rerobika, line 1022: Please check for false value instead and assert below for true status

tests/jerry/es.next/proxy_define_own_property.js
```diff
@@ -56,33 +56,20 @@ var handler = {
...
-Object.defineProperty(proxy, "name", {
-  get() {
-    return this._name;
-  },
-  set(value) {
-    this._name = value;
-  }
-});
+try {
+  Object.defineProperty(proxy, "bar", desc);
```
galpeter, line 60: After this the "bar" property should exist on the proxy/target, or not? We should add test/check for it.

dbatyai (approved):
LGTM

rerobika (approved):
LGTM

PR #4128, opened 2020-08-11
Make Array.prototype.unshift method conform to ES11

ossy-szeged:
Changes:
- Since ES6 Array.prototype.unshift shouldn't iterate
over the array elements if argument count is 0.
- Add proper guard for a check introduced in ES6.
- Update comments to reference to ES5.1 and ES11 specs too.

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
@@ -1514,7 +1514,8 @@ ecma_builtin_array_prototype_object_splice (const ecma_value_t args[], /**< argu
  * The Array.prototype object's 'unshift' routine
  *
  * See also:
- *          ECMA-262 v5, 15.4.4.13
+ *          ECMA-262  v5, 15.4.4.13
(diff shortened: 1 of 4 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -1562,51 +1563,61 @@ ecma_builtin_array_prototype_object_unshift (const ecma_value_t args[], /**< arg
...
+      ecma_deref_ecma_string (index_str_p);
+
+      if (ECMA_IS_VALUE_ERROR (operation_value))
+      {
+        return operation_value;
+      }
     }
   }
 
   for (uint32_t arg_index = 0; arg_index < args_number; arg_index++)
```
dbatyai, line 1618: Shouldn't this loop also be in the `should_iterate` block?

ossy-szeged: You're right, it should because of the spec. I'll fix it immediately.
Technically it is no-op, because args_number is still 0. 
But let's follow the spec and avoid double checking of args_number.

PR #4136, opened 2020-08-12
Fix memory leak in ecma_op_object_get_enumerable_property_names

ossy-szeged:

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -1934,8 +1934,10 @@ ecma_op_object_get_enumerable_property_names (ecma_object_t *obj_p, /**< routine
         return NULL;
       }
 
+      const bool is_enumerable = (prop_desc.flags & ECMA_PROP_IS_ENUMERABLE) != 0;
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1934,8 +1934,10 @@ ecma_op_object_get_enumerable_property_names (ecma_object_t *obj_p, /**< routine
         return NULL;
       }
 
+      const bool is_enumerable = (prop_desc.flags & ECMA_PROP_IS_ENUMERABLE) != 0;
+      ecma_free_property_descriptor (&prop_desc);
```
rerobika, line 1938: At this point `status` can be false which means that the requested property does not exist. In this case the property descriptor is uninitialized, therefore must not be freed.

galpeter: At the moment the descriptors are empty initialized by default.

rerobika: Indeed, just ignore my comment.

PR #4137, opened 2020-08-13
Fixing test262 built-ins json test-cases

kisbg:


kisbg:
@galpeter  Yes, you are right.

galpeter:
Why? Part of this PR is needed (bigint and stringify fixes)

PR #4140, opened 2020-08-13
Detect assignment pattern for for-in/of

zherczeg:


zherczeg:
This patch fixes nearly 400 test262 tests. Surprising for me.

tests/test262-esnext-excludelist.xml
```diff
@@ -10363,4 +9974,8 @@
   <test id="language/literals/regexp/u-unicode-esc-bounds.js"><reason></reason></test>
   <test id="language/literals/regexp/u-unicode-esc-non-hex.js"><reason></reason></test>
   <test id="language/literals/regexp/unicode-escape-nls-err.js"><reason></reason></test>
+
+  /* [...x = init] should not be allowed */
+  <test id="language/statements/for-in/dstr/array-rest-init.js"><reason></reason></test>
+  <test id="language/statements/for-of/dstr/array-rest-init.js"><reason></reason></test>
```
ossy-szeged, line 9980: Please use valid XML-style comment instead of c-style one.
Why these shouldn't be allowed? Are these tests outdated or maybe a future spec (ES11+) will allow this construction?

zherczeg: What kind of comment?
Its an engine bug, which is revelaed because another bug is fixed.

zherczeg:
Fixed the other issue so no new regressions introduced.

dbatyai (approved):
LGTM

PR #4154, opened 2020-08-18
Fix destructuring initializer bugs

rerobika:
- default value is not allowed after rest initializer
 - property calcuate object base stack index in object initializer


tests/jerry/es.next/array-pattern.js
```diff
@@ -278,6 +279,30 @@ function __createIterableObject (arr, methods) {
   assert (b === 2);
 }) ();
 
+(function () {
+  var value = { y: "42" };
+  var x = {};
+  var assignmentResult, iterationResult, iter;
+
+  iter = (function*() {
+    assignmentResult = { y: x[yield] } = value;
```
zherczeg, line 288: Please add a test for the computed case.

zherczeg (changes requested):
Good patch, please add another test for computed case.

rerobika:
Closing the PR, since these testcases have revealed some serious problems about destructuring patterns. I'll come up with some rework.

PR #4155, opened 2020-08-18
Breaking from for statement should return with empty completion

rerobika:


zherczeg:
This feature is very unclear how it works, and as far as I heard they plan to change it. I would wait a bit with this.

Issue #4148, opened 2020-08-17
Memory leak in TypedArray.prototype.toLocaleString


ossy-szeged:
###### JerryScript revision
086d1cc127f12e31ed315b41ca59ec0c3e88c7f4
(latest master - 2020.08.18.)

###### Build platform
Ubuntu 18.04.5 LTS (Linux 4.15.0-112-generic x86_64)

###### Build steps
tools/build.py --debug --logging=on --error-messages=on --line-info=on

###### Test cases
test262-esnext tests:
- built-ins/TypedArray/prototype/toLocaleString/return-abrupt-from-firstelement-tostring.js
- built-ins/TypedArray/prototype/toLocaleString/return-abrupt-from-firstelement-valueof.js
- built-ins/TypedArray/prototype/toLocaleString/return-abrupt-from-nextelement-tostring.js
- built-ins/TypedArray/prototype/toLocaleString/return-abrupt-from-nextelement-valueof.js

###### Output
```
 ICE: Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed at /home/oszi/jerryscript/jerry-core/jmem/jmem-heap.c(jmem_heap_finalize):107.
Error: ERR_FAILED_INTERNAL_ASSERTION
```

It means a memory leak somewhere in TypedArray.prototype.toLocaleString implementation.

###### Expected behavior
not assert -> not leak

PR #4156, opened 2020-08-18
Correctly release values in TypedArray's toLocaleString

galpeter:
When the `toLocaleString` was called on a TypedArray's value
the resulting object's `toString/valueOf` invocations could
create errors. These error values were not released.

In addition the input element value for the toString operation
was released twice in case of an error.

Fixes: #4148


ossy-szeged:
LGTM (informal)

dbatyai (approved):
LGTM

PR #4163, opened 2020-08-24
Introduce new API function to obtain well-known symbols

rerobika:
- jerry_get_well_known_symbol

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -141,6 +141,25 @@ Container object types:
 
  *New in version 2.3*.
 
+## jerry_well_known_symbol_t
+
+Well-known symbols:
+
+ - JERRY_SYMBOL_HAS_INSTANCE - @@hasInstance well-known symbol
+ - JERRY_SYMBOL_IS_CONCAT_SPREADABLE - @@isConcatSpreadable well-known symbol
+ - JERRY_SYMBOL_ITERATOR - @@iterator well-known symbol
+ - JERRY_SYMBOL_ASYNC_ITERATOR - @@asyncIterator well-known symbol
+ - JERRY_SYMBOL_MATCH - @@match well-known symbol
+ - JERRY_SYMBOL_REPLACE - @@replace well-known symbol
+ - JERRY_SYMBOL_SEARCH - @@search well-known symbol
+ - JERRY_SYMBOL_SPECIES - @@species well-known symbol
+ - JERRY_SYMBOL_SPLIT - @@split well-known symbol
+ - JERRY_SYMBOL_TO_PRIMITIVE - @@toPrimitive well-known symbol
+ - JERRY_SYMBOL_TO_STRING_TAG - @@toStringTag well-known symbol
+ - JERRY_SYMBOL_UNSCOPABLES - @@unscopables well-known symbol
+
+ *New in version [[NEXT_RELEASE]]*.
+
 ## jerry_regexp_flags_t
 
 RegExp object optional flags:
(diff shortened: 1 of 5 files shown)


dbatyai (approved):
LGTM

tests/unit-core/test-symbol.c
```diff
@@ -191,6 +191,130 @@ main (void)
...
+  jerry_release_value (builtin_symbol);
+
+  const jerry_char_t deleter_src[] = "Symbol = {}";
+
+  jerry_value_t deleter = jerry_eval (deleter_src, sizeof (deleter_src) - 1, JERRY_PARSE_NO_OPTS);
+  TEST_ASSERT (!jerry_value_is_error (deleter));
+  jerry_release_value (deleter);
+
+  builtin_symbol = jerry_get_property (global_obj, symbol_str);
+  TEST_ASSERT (jerry_value_is_object (builtin_symbol));
```
galpeter, line 282: Is this correcty? The Symbol is already an object type even without doing the `Symbol = {}` code.

docs/02.API-REFERENCE.md
```diff
@@ -91,6 +91,26 @@ Container object types:
...
+ - JERRY_SYMBOL_ASYNC_ITERATOR - @@asyncIterator well-known symbol
+ - JERRY_SYMBOL_MATCH - @@match well-known symbol
+ - JERRY_SYMBOL_REPLACE - @@replace well-known symbol
+ - JERRY_SYMBOL_SEARCH - @@search well-known symbol
+ - JERRY_SYMBOL_SPECIES - @@species well-known symbol
+ - JERRY_SYMBOL_SPLIT - @@split well-known symbol
+ - JERRY_SYMBOL_TO_PRIMITIVE - @@toPrimitive well-known symbol
+ - JERRY_SYMBOL_TO_STRING_TAG - @@toStringTag well-known symbol
+ - JERRY_SYMBOL_UNSCOPABLES - @@unscopables well-known symbol
+ - JERRY_SYMBOL__NUMBER - number of well-known symbols
```
galpeter, line 110: No other enums have this `__NUMBER`, could we remove this?

jerry-core/api/jerry.c
```diff
@@ -3455,6 +3455,37 @@ jerry_get_promise_state (const jerry_value_t promise) /**< promise object to get
...
+ * @return undefined value - if invalid well-known symbol was requested
+ *         well-known symbol value - otherwise
+ */
+jerry_value_t
+jerry_get_well_known_symbol (jerry_well_known_symbol_t symbol) /**< jerry_well_known_symbol_t enum value */
+{
+  jerry_assert_api_available ();
+
+#if ENABLED (JERRY_ESNEXT)
+  if (symbol >= JERRY_SYMBOL__NUMBER)
```
galpeter, line 3473: Based on this you should also check if it is less then the first argument. Maybe we could directly use the `LIT_IS_GLOBAL_SYMBOL` macro here?

tests/unit-core/test-symbol.c
```diff
@@ -191,6 +191,130 @@ main (void)
...
+                                                     well_known_symbol,
+                                                     current_global_symbol);
+
+    TEST_ASSERT (jerry_value_is_boolean (relation)
+                 && jerry_get_boolean_value (relation));
+
+    jerry_release_value (relation);
+
+    jerry_value_t prop_result_wn = jerry_get_property (obj, well_known_symbol);
+    jerry_value_t prop_result_global = jerry_get_property (obj, well_known_symbol);
```
galpeter, line 257: These two `get_property` calls are doing the same thing (the inputs are the same). Is it intentional?

tests/unit-core/test-symbol.c
```diff
@@ -191,6 +191,130 @@ main (void)
...
+  TEST_ASSERT (!jerry_value_is_error (deleter));
+  jerry_release_value (deleter);
+
+  builtin_symbol = jerry_get_property (global_obj, symbol_str);
+  TEST_ASSERT (jerry_value_is_object (builtin_symbol));
+
+  expected = 1.0;
+  prop_index = 0;
+
+  for (jerry_well_known_symbol_t id = JERRY_SYMBOL_HAS_INSTANCE;
```
galpeter, line 287: This test deserves a comment on why it is here (after the "deletion" of the well-known Symbols) otherwise it almost looks the same as the one `for` loop above.

tests/unit-core/test-symbol.c
rerobika: This is a reassignment. The builtin `Symbol` is substituted with an empty object therefore the well known symbols referenced by the builtin `Symbol` object are no longer accessible.

jerry-core/api/jerry.c
rerobika: we cannot use `LIT_IS_GLOBAL_SYMBOL` directly since there is an offset between the first internal magic string and the enum value.

tests/unit-core/test-symbol.c
rerobika: Good catch, the second should have been `current_global_symbol`

jerry-core/api/jerry.c
galpeter: In the assert below it is used after calculating with the offset. Can do that here also?

tests/unit-core/test-symbol.c
galpeter: That I understand, but the `builtin_symbol` will always be of an object type. This assert does not add any check to ensure that the global "Symbol" property was replaced.

rerobika: Fine, I'll update it.

jerry-core/api/jerry.c
rerobika: Sure!

rerobika:

docs/02.API-REFERENCE.md
```diff
@@ -3691,6 +3710,66 @@ jerry_resolve_or_reject_promise (jerry_value_t promise,
 
 These APIs all depend on the es.next profile (or on build options).
 
+## jerry_get_well_known_symbol
+
+**Summary**
+
+Get the well-known symbol corresponds to the given [well-known symbol id](#jerry_well_known_symbol_t).
```
galpeter, line 3717: Maybe: `... symbol corresponding to the given ... ` ?

rerobika: Thanks, I've fixed this misspelling.

zherczeg (approved):
LGTM

Issue #4159, opened 2020-08-21
Memory corruption in in opfunc_for_in (jerryscript/jerry-core/vm/opcodes.c:350)


Aurorainfinity:
Memory corruption in in opfunc_for_in (jerryscript/jerry-core/vm/opcodes.c:350)

###### JerryScript revision

commit : https://github.com/jerryscript-project/jerryscript/commit/5b3f01af9d3c134b90eaeefe1fb8a05e0a6c1464
version : v2.3.0

###### Build platform

Ubuntu 16.04.1 LTS (4.15.0-45-generic x86_64)

###### Build steps

```
python tools/build.py --profile=es2015-subset --lto=off --compile-flag=-g \
--strip=off --logging=on \
--compile-flag=-fsanitize=address --stack-limit=15
```

###### Test case

Memory_corruption_opfunc_for_in.js
```
function main() {
var v4 = [13.37,13.37,13.37,13.37];
var v6 = ["prototype",1337];
var v7 = [1337,v4,1337,"prototype",13.37,-4294967295,1337,-4294967295,1337,v6];
var v8 = {__proto__:v7,a:"prototype",b:v7,constructor:1337,d:v4,e:13.37,valueOf:v7};
var v9 = {c:Symbol,length:Symbol};
var v10 = v4;
var v13 = typeof BigInt;
var v17 = [13.37];
var v20 = typeof String;
var v21 = async (v22,v23,v24,v25) => {
    var v28 = {apply:isFinite,call:isFinite,construct:isFinite,defineProperty:isFinite,getOwnPropertyDescriptor:isFinite,getPrototypeOf:isFinite,has:isFinite,ownKeys:isFinite,preventExtensions:isFinite,set:isFinite,setPrototypeOf:isFinite};
    var v30 = new Proxy(isFinite,v28);
    for (var v32 in v30) {
        var v33 = {e:BigInt};
    }
};
var v34 = v21(65535,v20,v17,v21);
var v35 = typeof BigInt;
var v36 = 0;
var v37 = 1;
var v38 = 1;
var v39 = /(.*)*Jae5y/m.test("-4096");
// Stderr:
}
main();
```

###### Execution steps

```
$ ls 
Memory_corruption_opfunc_for_in.js
$ ../jerryscript/build/bin/jerry Memory_corruption_opfunc_for_in.js
ASAN:SIGSEGV
=================================================================
[...]

PR #4165, opened 2020-08-24
Correctly report error in for-in start in case of proxies

galpeter:
In case of Proxies the "ownKeys" call can return an incompatible
value for a for-in statement. In such cases the error should be
propagated to the user.

Fixes: #4159


rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4172, opened 2020-08-27
Refactor Array.prototype.splice to conform ES11 standard

szilagyiadam:
The algorithm is based on ECMA-262 v11, 22.1.3.28

### DIFF
--- a/tests/test262-esnext-excludelist.xml
+++ b/tests/test262-esnext-excludelist.xml
@@ -156,14 +156,9 @@
   <test id="built-ins/Array/prototype/slice/create-proto-from-ctor-realm-array.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/slice/create-proto-from-ctor-realm-non-array.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/sort/comparefn-nonfunction-call-throws.js"><reason></reason></test>
-  <test id="built-ins/Array/prototype/splice/create-non-array-invalid-len.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/splice/create-proto-from-ctor-realm-array.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/splice/create-proto-from-ctor-realm-non-array.js"><reason></reason></test>
-  <test id="built-ins/Array/prototype/splice/create-species-length-exceeding-integer-limit.js"><reason></reason></test>
-  <test id="built-ins/Array/prototype/splice/create-species-undef-invalid-len.js"><reason></reason></test>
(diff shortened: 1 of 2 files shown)


tests/test262-esnext-excludelist.xml
```diff
@@ -156,14 +156,9 @@
   <test id="built-ins/Array/prototype/slice/create-proto-from-ctor-realm-array.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/slice/create-proto-from-ctor-realm-non-array.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/sort/comparefn-nonfunction-call-throws.js"><reason></reason></test>
-  <test id="built-ins/Array/prototype/splice/create-non-array-invalid-len.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/splice/create-proto-from-ctor-realm-array.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/splice/create-proto-from-ctor-realm-non-array.js"><reason></reason></test>
-  <test id="built-ins/Array/prototype/splice/create-species-length-exceeding-integer-limit.js"><reason></reason></test>
-  <test id="built-ins/Array/prototype/splice/create-species-undef-invalid-len.js"><reason></reason></test>
-  <test id="built-ins/Array/prototype/splice/create-species.js"><reason></reason></test>
   <test id="built-ins/Array/prototype/splice/property-traps-order-with-species.js"><reason></reason></test>
```
rerobika, line 161: Please check this testcase as well.

ossy-szeged: I already had a WIP fix for splice (functionally similar), but this PR is better, because it completely rewrote splice to follow the spec word by word , including variable naming too. Great work.

I have only a minor request, please keep references for ES5.1 spec too with prefixes to be able to follow both spec.
For example: /* ES5.1: 8-9, ES11: 10-11. */ It would make the code much more readable.

szilagyiadam: This test case is not specifically related to this PR, because it tests the order of the operations on the newly created array, but we use ```ecma_builtin_array_prototype_helper_set_length``` to set the new length instead of the ```Set``` abstract method described in here: https://www.ecma-international.org/ecma-262/11.0/#sec-set-o-p-v-throw
IMHO the fix for this testcase is outside of the scope of this PR.

rerobika: Thanks for the explanation, we can fix it later.

ossy-szeged:
LGTM (informal)

dbatyai (approved):
LGTM

rerobika (approved):
LGTM

PR #4177, opened 2020-09-01
Add missing value release to TypedArray fill method

rwalczyna:
Value was not freed when error occurs. It caused internal assertion to fail
when BigInt was used.


zherczeg (approved):
LGTM

rerobika (approved):
LGTM

PR #4196, opened 2020-09-16
Update object.prototype.toString to conform ES11 standard

kisbg:

### DIFF
--- a/tests/jerry/es.next/object-prototype-tostring.js
+++ b/tests/jerry/es.next/object-prototype-tostring.js
@@ -21,7 +21,7 @@ assert (Symbol.prototype[Symbol.toStringTag] === undefined);
 Symbol.prototype[Symbol.toStringTag] = "myStringTag1";
 assert (Object.prototype.toString.call (Symbol ()) === "[object myStringTag1]");
 Symbol.prototype[Symbol.toStringTag] = {};
-assert (Object.prototype.toString.call (Symbol ()) === "[object Symbol]");
+assert (Object.prototype.toString.call (Symbol ()) === "[object Object]");
 
 /* Math @@toStringTag */
 assert (Math[Symbol.toStringTag] === "Math");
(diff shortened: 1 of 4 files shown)


tests/jerry/es.next/object-prototype-tostring.js
```diff
@@ -21,7 +21,7 @@ assert (Symbol.prototype[Symbol.toStringTag] === undefined);
 Symbol.prototype[Symbol.toStringTag] = "myStringTag1";
 assert (Object.prototype.toString.call (Symbol ()) === "[object myStringTag1]");
 Symbol.prototype[Symbol.toStringTag] = {};
-assert (Object.prototype.toString.call (Symbol ()) === "[object Symbol]");
+assert (Object.prototype.toString.call (Symbol ()) === "[object Object]");
```
rerobika, line 24: This change is incorrect. The testcase is up-to-date with the latest standard.

rerobika (changes requested):

tests/jerry/es.next/object-prototype-tostring.js
rerobika: Worng alert, just ignore my comment.

rerobika (approved):
LGTM

tests/jerry/es.next/object-prototype-tostring.js
```diff
@@ -89,7 +89,7 @@ var typedArrayTypes = ["Int8Array",
                        "Uint32Array",
                        "Float32Array",
                        "Float64Array"];
-
+//
```
dbatyai, line 92: Please remove this comment.

kisbg: Thank you for the review!

tests/jerry/es.next/object-prototype-tostring.js
```diff
@@ -67,7 +67,7 @@ Map.prototype[Symbol.toStringTag] = "myStringTag5";
 assert (Map.prototype.toString.call (new Map ()) === "[object myStringTag5]");
 assert (Object.prototype.toString.call (Map) === "[object Function]");
 Map.prototype[Symbol.toStringTag] = {};
-assert (Map.prototype.toString.call (new Map) === "[object Map]");
+assert (Map.prototype.toString.call (new Map) === "[object Object]");
```
zherczeg, line 70: Botf Chrome and Firefox says `Map.prototype.toString.call (new Map)` -> `[object Map]`. Why do you want to change this?

zherczeg (changes requested):

tests/jerry/es.next/object-prototype-tostring.js
kisbg: This working perfectly, since in line `64` deletes the original `toStringTag` which is not writable and in line `65` create a new toStringTag property which becomes writable and that's why in line `70` result becomes `[object Object]`. 
If we don't delete the prototype result becomes `[object Map]`.

zherczeg (approved):
LGTM

PR #4198, opened 2020-09-16
Convert fast array to normal array when the prototype is changed

galpeter:
If the prototype of a fast array is changed, for example to a Proxy object,
the array should be converted back to a normal array. This is required to
correctly handle any [[Get]]/[[Set]]/etc... calls which should trigger
these method calls on the before mentioned Proxy object.


rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4228, opened 2020-09-23
Fix class name binding creation

dbatyai:


jerry-core/parser/js/js-scanner.c
```diff
@@ -1597,7 +1600,7 @@ scanner_scan_statement (parser_context_t *context_p, /**< context */
       lexer_lit_location_t *literal_p = scanner_add_literal (context_p, scanner_context_p);
 
       scanner_detect_invalid_let (context_p, literal_p);
-      literal_p->type |= SCANNER_LITERAL_IS_LET;
+      literal_p->type |= (SCANNER_LITERAL_IS_LET | SCANNER_LITERAL_EARLY_CREATE);
```
zherczeg, line 1603: This is effectively destroys register stoirage of classes, which I think is an important optimization.

Issue #4213, opened 2020-09-23
heap-use-after-free in bigint/TypedArray related code path


ossy-szeged:
###### JerryScript revision
3c723c9bc1aec08a54bc44bba7987e4324f440e4
(latest master - 2020.09.21.)

###### Build platform
Ubuntu 18.04.5 LTS (Linux 4.15.0-112-generic x86_64)

###### Build steps
./tools/build.py --clean --debug --compile-flag=-fsanitize=address --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --compile-flag=-g --strip=off --system-allocator=on --logging=on --linker-flag=-fuse-ld=gold --error-messages=on --stack-limit=16

###### Test case
found by test262-esnext tests:
- built-ins/TypedArray/prototype/find/BigInt/get-length-ignores-length-prop.js
- built-ins/TypedArray/prototype/find/BigInt/predicate-call-changes-value.js
- built-ins/TypedArray/prototype/find/BigInt/return-found-value-predicate-result-is-true.js

###### Output
```
==19500==ERROR: AddressSanitizer: heap-use-after-free on address 0xf5c00e90 at pc 0x566397e8 bp 0xffde8848 sp 0xffde8838
READ of size 4 at 0xf5c00e90 thread T0
    #0 0x566397e7 in ecma_ref_extended_primitive /home/oszi/jerryscript/jerry-core/ecma/base/ecma-helpers.c:1229
    #1 0x566418a0 in ecma_copy_value /home/oszi/jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:906
    #2 0x5674199a in ecma_fast_copy_value /home/oszi/jerryscript/jerry-core/ecma/base/ecma-helpers-value.c:940
    #3 0x5674199a in vm_init_exec /home/oszi/jerryscript/jerry-core/vm/vm.c:4802
    #4 0x567471e1 in vm_run /home/oszi/jerryscript/jerry-core/vm/vm.c:4938
    #5 0x5669a3ee in ecma_op_function_call_simple /home/oszi/jerryscript/jerry-core/ecma/operations/ecma-function-object.c:980
    #6 0x5669b7bb in ecma_op_function_call /home/oszi/jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1182
    #7 0x5674655f in opfunc_call /home/oszi/jerryscript/jerry-core/vm/vm.c:804
[...]

PR #4230, opened 2020-09-24
Fix invalid free in TypedArray find method.

zherczeg:


ossy-szeged:
It seems it makes these test pass on ESNext:
  built-ins/TypedArray/prototype/find/BigInt/get-length-ignores-length-prop.js
  built-ins/TypedArray/prototype/find/BigInt/predicate-call-changes-value.js
  built-ins/TypedArray/prototype/find/BigInt/return-found-value-predicate-result-is-true.js

Please remove them from excludelist and add "Fixes #4213" to the commit log and issue description.
Otherwise LGTM (informal).

rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4231, opened 2020-09-24
Add new checks to function lazy list property names methods

kisbg:
BugFix: if one of the lazy initialized property if enumerable then,
it must be added to the collection.

updated jerry/es.next/regression-test-issue-3267.js test-case, since it was outdated

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -1824,12 +1824,23 @@ ecma_op_function_list_lazy_property_names (ecma_object_t *object_p, /**< functio
 {
 #if ENABLED (JERRY_ESNEXT)
   ecma_extended_object_t *ext_func_p = (ecma_extended_object_t *) object_p;
-  if (!ECMA_GET_FIRST_BIT_FROM_POINTER_TAG (ext_func_p->u.function.scope_cp))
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -1824,12 +1824,23 @@ ecma_op_function_list_lazy_property_names (ecma_object_t *object_p, /**< functio
 {
 #if ENABLED (JERRY_ESNEXT)
   ecma_extended_object_t *ext_func_p = (ecma_extended_object_t *) object_p;
-  if (!ECMA_GET_FIRST_BIT_FROM_POINTER_TAG (ext_func_p->u.function.scope_cp))
+  ecma_value_t prop_name_val = ecma_make_magic_string_value (LIT_MAGIC_STRING_LENGTH);
+  if (!ECMA_GET_FIRST_BIT_FROM_POINTER_TAG (ext_func_p->u.function.scope_cp)
+      || ecma_op_ordinary_object_has_own_property (object_p, ecma_op_to_string (prop_name_val)))
```
zherczeg, line 1829: This means, if the property is created, and deleted and created again it will appear early. Are you sure this is correct?

kisbg: Good point! You are right, if we delete the `length` property and defined later. Then `length`  should be the end of the list.

PR #4247, opened 2020-09-30
Fix class handling with disabled module system

galpeter:
If the module system is disabled however the es.next is enabled the following
code triggers an assert:

```js
class Demo {}
```

This was due to an incorrect scanner data processing where it was always
assumed that there is a module related information.


zherczeg (approved):
LGTM

ossy-szeged:
LGTM (informal)

rerobika (approved):
LGTM

PR #4248, opened 2020-09-30
Date.parse should reject out of range values

ossy-szeged:
The spec doesn't say explicitly that Date.parse should call TimeClip
abstract operation at the end. But Date object can represent time
values only in the range 1970 Jan 01 +/-8,640,000,000,000,000 ms.


rerobika (approved):
LGTM

dbatyai (approved):
LGTM

dbatyai:
Please rebase.

PR #4249, opened 2020-09-30
Fix Date.parse to handle times without timezone properly

ossy-szeged:
https://www.ecma-international.org/ecma-262/11.0/#sec-date.parse
"When the UTC offset representation is absent, date-only forms are
interpreted as a UTC time and date-time forms are interpreted as a local time."

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date.c
@@ -293,6 +293,7 @@ ecma_builtin_date_parse_ISO_string_format (const lit_utf8_byte_t *date_str_curr_
       day = ecma_date_parse_date_chars (&date_str_curr_p, date_str_end_p, 2, 1, 31);
     }
 
+    bool is_utc = true;
     /* 4. read time if any */
     if (ecma_date_parse_special_char (&date_str_curr_p, date_str_end_p, 'T'))
     {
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-date.c
```diff
@@ -367,13 +368,24 @@ ecma_builtin_date_parse_ISO_string_format (const lit_utf8_byte_t *date_str_curr_
...
       }
     }
 
     if (date_str_curr_p >= date_str_end_p)
     {
       ecma_number_t date = ecma_date_make_day (year, month - 1, day);
-      return ecma_date_make_date (date, time);
+
+      ecma_number_t result_date = ecma_date_make_date (date, time);
+      if (is_time_in_local_time_zone)
```
rerobika, line 383: `ecma_date_make_date` May return NaN, so I'd not pass it to `ecma_date_utc`.

ossy-szeged: ecma_date_utc only subtracts the local time zone offset
and subtracting anything from NaN will be always NaN.

Or did I miss something?

rerobika: Indeed, but there is also a `localtime_r` calculation which can be skipped if we already know that the number is nan.

ossy-szeged: I think parsing invalid dates is very a rare case and we should optimize for common cases.
If we did NaN check here, the rare case could be faster, but the common case would be slower.

rerobika: We are not dealing about heavy optimization. Only about to not perform operations which are unnecessary to do.

ossy-szeged: I'm not trying to do any heavy optimization here. But your suggested NaN check here would obviously slow down the common case (valid date input) in order not to perform some extra operations in a very rare case (invalid date input). I still don't agree with you to optimize the code for rare cases with slowing down the common cases.

Otherwise I haven't seen any NaN check before ecma_date_utc calls in the code base.

rerobika: Convinced.

rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4256, opened 2020-10-01
Error should be thrown later when calling super multiple times

dbatyai:


jerry-core/vm/vm.c
```diff
@@ -577,16 +577,14 @@ vm_super_call (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
-    completion_value = ecma_raise_reference_error (ECMA_ERR_MSG ("Super constructor may only be called once"));
-  }
-  else if (!ecma_is_constructor (func_value))
+  if (!ecma_is_constructor (func_value))
   {
     completion_value = ecma_raise_type_error (ECMA_ERR_MSG ("Class extends value is not a constructor."));
   }
   else
   {
+    const bool called_before = ecma_op_this_binding_is_initialized (prop_p);
```
rerobika, line 586: As I see the this value is only binded at line 624 so it still remains uninitialized after the constructor call so saving this value unnecessary.

rerobika (changes requested):

rerobika (approved):
LGTM

galpeter (approved):
lgtm

PR #4257, opened 2020-10-01
Check for prototype bound name in class static accessors

dbatyai:


jerry-core/parser/js/js-parser-expr.c
```diff
@@ -621,8 +621,16 @@ parser_parse_class_body (parser_context_t *context_p, /**< context */
       {
         is_computed = true;
       }
-      else if (!is_static && parser_is_constructor_literal (context_p))
+      else if (is_static)
       {
+        if (lexer_compare_literal_to_string (context_p, "prototype", 9))
```
rerobika, line 626: Can we ensure that the current literal is ident or string?

Also, as I see we should also throw error in the following case :
```js
class A {
  static get prototyp\u{0065}() {}
}
```

rerobika (changes requested):

jerry-core/parser/js/js-lexer.c
```diff
@@ -3707,12 +3707,12 @@ lexer_compare_literal_to_string (parser_context_t *context_p, /**< context */
                                  const char *string_p, /**< string */
                                  size_t string_length) /**< string length */
 {
-  JERRY_ASSERT (context_p->token.type == LEXER_LITERAL
-                && (context_p->token.lit_location.type == LEXER_IDENT_LITERAL
-                    || context_p->token.lit_location.type == LEXER_STRING_LITERAL));
+  JERRY_ASSERT (context_p->token.type == LEXER_LITERAL);
```
zherczeg, line 3710: I think we should not change this function.

rerobika:
LGTM

rerobika (approved):
LGTM

tests/jerry/es.next/class.js
```diff
@@ -49,6 +49,11 @@ must_throw("class class {}");
 must_throw("class A { constructor() {} this.a = 5 }");
 must_throw("class A { constructor() {} constructor() {} }");
 must_throw("class A { static prototype() {} }");
+must_throw("class A { static get prototype() {} }");
+must_throw("class A { static set prototype() {} }");
```
zherczeg, line 53: I would add some testcases for the string variant as well:
`must_throw("class A { static get 'prototype'() {} }");`

zherczeg (changes requested):

zherczeg (approved):
LGTM

zherczeg:
Please resolve conflicts first, then you can land the patch.

PR #4258, opened 2020-10-01
Fix the value of the caller property of function instances

dbatyai:
We do not support the caller information for functions, and since a
'null' value represents that there has been no caller, the default value
should be changed to 'undefined' to signal that the information is not
available.


rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #4261, opened 2020-10-01
Remove revoked Proxy checks when creating a Proxy

galpeter:
In the newer ecma262 standard (post ES11) the ProxyCreate was
changed and the revoked Proxy handler/target is not checked.

Related standard PR: https://github.com/tc39/ecma262/pull/1814

PR #4264, opened 2020-10-01
Correctly handle the Proxy IsCallable and IsConstructor information

galpeter:
The `IsCallable(target)` and `IsConstructor(target)` info
can't be stored in the target/handler values.
If the input for the ProxyCreate was a revocable Proxy the original target's
callable/constructor information must be retained even after the
Proxy was revoked.

Depends on: #4261

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -2128,6 +2128,10 @@ do \
 #if ENABLED (JERRY_BUILTIN_PROXY)
 /**
  * Description of Proxy objects.
+ *
+ * A Proxy object's property list is used to store extra information:
+ *  * The "header.u1.property_list_cp" 1st tag bit stores the IsCallable information.
+ *  * The "header.u1.property_list_cp" 2nd tag bit stores the IsConstructor information.
  */
 typedef struct
 {
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -786,7 +786,11 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
       case ECMA_OBJECT_TYPE_PROXY:
       {
         ecma_gc_mark_proxy_object (object_p);
-        break;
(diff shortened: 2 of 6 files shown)


rerobika (approved):
lgtm

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -2112,6 +2121,7 @@ typedef struct
   ecma_object_t header; /**< header part */
   ecma_value_t target; /**< [[ProxyTarget]] internal slot */
   ecma_value_t handler; /**< [[ProxyHandler]] internal slot */
+  uint8_t flags; /**< Store ecma_proxy_flags_t bits */
```
zherczeg, line 2124: This increases the struct by 8 bytes. Could we use tags to store these flags?

zherczeg (changes requested):

jerry-core/ecma/base/ecma-globals.h
galpeter: I've tested and the size only increased by 4 bytes. Still we could store the flags as tags, but on which value? on the target? on the handler? on the proxy? If I understand correctly the tags would require a `jmem_cpointer_t` typed value.

galpeter: Well maybe this current solution using the proplist_cp to store the tags could work.

zherczeg: If target or handler is an object, you can use the 3 bottom bits for free. We have a macro system for handling these bits. See byte-code functions.

zherczeg:
We in general use the least used object member for tagging, i.e. the scope pointer of byte code functions. Is the property list the least used member?

jerry-core/ecma/base/ecma-globals.h
galpeter: We need to keep the information even if the proxy is revoked and at that point the target/handler is an `ECMA_VALUE_NULL`.

galpeter:
Based on the tests and investigation the property list is not used in case of proxy objects.

zherczeg (approved):
LGTM

PR #4272, opened 2020-10-05
Fix flag handling when creating arguments objects

dbatyai:
Member variables of the object must be initialized before setting the
required flags, otherwise the uninitalized values can cause problems
during garbage collection.


rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #4273, opened 2020-10-05
Propagate strict mode for Proxy targets during delete

galpeter:
After ES5.1 if the delete returns false a TypeError should be thrown
in strict mode.


rerobika (approved):
lgtm

zherczeg (approved):
LGTM

Issue #3070, opened 2019-09-05
Assertion 'start <= end' in ecma-builtin-string-prototype.c


made0x78:
###### Revision
57f389d

###### Build
./tools/build.py --clean --debug --compile-flag=-fsanitize=address
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer
--compile-flag=-fno-common --compile-flag=-g
--strip=off --system-allocator=on --logging=on
--error-messages=on --profile=es2015-subset

###### OS
Linux 4.15.0-58-generic #64-Ubuntu x86_64 GNU/Linux

###### Test case
```javascript
var r = new RegExp("([X]{6}|.*)", "g");
var s = "a";
s.replace(r, () => r.compile("[PqaCZlWQUT]{0}", "m"))
```

###### Backtrace
Run with `jerry --abort-on-fail poc.js`

```
ICE: Assertion 'start <= end' failed at /jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c(ecma_builtin_string_prototype_object_replace_append_substr):542.
Error: ERR_FAILED_INTERNAL_ASSERTION

Program received signal SIGABRT, Aborted.
0xf7fd5059 in __kernel_vsyscall ()
(gdb) bt
#0  0xf7fd5059 in __kernel_vsyscall ()
#1  0xf7841452 in raise () from /lib32/libc.so.6
#2  0xf7842871 in abort () from /lib32/libc.so.6
#3  0x566bdfef in jerry_port_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at /jerryscript/jerry-port/default/default-fatal.c:71
#4  0x566513cf in jerry_fatal (code=ERR_FAILED_INTERNAL_ASSERTION) at /jerryscript/jerry-core/jrt/jrt-fatals.c:58
#5  0x56651410 in jerry_assert_fail (assertion=0x566da500 "start <= end", file=0x566da100 "/jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-string-prototype.c", 
    function=0x566f94c0 <__func__.13780> "ecma_builtin_string_prototype_object_replace_append_substr", line=542) at /jerryscript/jerry-core/jrt/jrt-fatals.c:82
#6  0x565f17cc in ecma_builtin_string_prototype_object_replace_append_substr (base_string_p=0xf5f00490, appended_string_p=0xf5d00670, start=1, end=0)
[...]

PR #4277, opened 2020-10-06
Replace fast path should keep a reference to the executed bytecode

dbatyai:
Replacer functions might release the original bytecode, keeping an extra reference will make sure that the bytecode stays alive.


dbatyai:
This issue is already covered by the test case from #3070, however it depends on garbage collection.

rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #3980, opened 2020-07-06
Improve function length property calculations

mmatyas:
Starting from ECMAScript 2015, the `length` property of functions
should be the number of parameters without a default value before
the first of such fields or a rest parameter.

See: https://www.ecma-international.org/ecma-262/11.0/index.html#sec-function-definitions-static-semantics-expectedargumentcount

JerryScript-DCO-1.0-Signed-off-by: Mátyás Mustoha mmatyas@inf.u-szeged.hu

PR #4290, opened 2020-10-15
Implement proper function length support

zherczeg:
Comma after last destructuring argument has been fixed as well


rerobika:
Please rebase.

zherczeg:
Done.

jerry-core/parser/js/js-parser.c
```diff
@@ -1726,12 +1715,41 @@ parser_post_processing (parser_context_t *context_p) /**< context */
     *(--base_p) = ECMA_VALUE_EMPTY;
   }
 
+  if (context_p->argument_length != UINT16_MAX)
+  {
+    compiled_code_p->status_flags |= CBC_CODE_FLAGS_HAS_EXTENDED_INFO;
+    *(--base_p) = context_p->argument_length;
```
rerobika, line 1721: If I see correctly the extended info is now an `ecma_value_t`. How about creating a struct like 

```c
typedef struct {
 union {
    uint8_t argument_length;
    ecma_value_t alignment;
  } u;
  
} ecma_compiled_code_extended_info_t;
```

If any extra info should be added later, extending this structure with fields would be more comfortable.

zherczeg: I also consider it as a flagset in the future. I don't know how many bits will be needed for different things, I would let that later.

rerobika: Sure, just we can keep that in mind.

rerobika (approved):
LGTM

ossy-szeged:
It seems after this change we don't need #3980. Please close it if you agree.

dbatyai (approved):
LGTM

PR #4299, opened 2020-10-19
Improve JerryScript debugger

zherczeg:
- Support print (p) as alias to eval (e)
- Fix backtrace processing (incorrect increment when cpointer is 4 byte long)
- Support partial names for pending breakpoints (similar to normal breakpoints)
- Don't print newline after pending breakpoint dialog text
- Add jerryscript-debugger-transport.h to all-in-one build


rerobika (approved):
LGTM

galpeter (approved):
lgtm

PR #4306, opened 2020-10-22
Implement ES11 globalThis

zherczeg:
Also improve symbol built-in instantiation.


jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -851,6 +851,15 @@ ecma_builtin_try_to_instantiate_property (ecma_object_t *object_p, /**< object *
     case ECMA_BUILTIN_PROPERTY_SIMPLE:
     {
       value = curr_property_p->value;
+
+#if ENABLED (JERRY_ESNEXT)
+      if (value == ECMA_VALUE_GLOBAL_THIS)
```
rerobika, line 856: How about introduce `ECMA_BUILTIN_PROPERTY_GLOBAL_THIS` instead of using this extra check in `ECMA_BUILTIN_PROPERTY_SIMPLE`?

zherczeg: I have considered this option, but I don't think it is worth to add another type. Perhaps a `ECMA_BUILTIN_PROPERTY_COMPLEX` could be added if you insist on it.

rerobika: I've just check the codebase and truly not a hot path, keep it like that for now.

rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4320, opened 2020-10-30
Fix the ESNext build with disabled JERRY_BUILTIN_PROXY

kaadam:
With a disabled JERRY_BUILTIN_PROXY option the build fails because
of missing guards.

The buildoption_test is also extended with it.

JerryScript-DCO-1.0-Signed-off-by: Adam Kallai kadam@inf.u-szeged.hu


galpeter (approved):
lgtm

rerobika (approved):
LGTM

PR #4325, opened 2020-11-03
Implement asInteger, asUint32 and asInt32 jerry api methods

kisbg:

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -3768,6 +3768,92 @@ jerry_value_to_bigint (const jerry_value_t value);
 - [jerry_value_is_bigint](#jerry_value_is_bigint)
 - [jerry_get_bigint_digits](#jerry_get_bigint_digits)
 
+## jerry_value_as_integer
+
+**Summary**
+
+Convert any number to integer number. Returns 0 if value is not number.
+Based on ECMA 262 v11 7.1.5
+
+**Prototype**
+
+```c
+double
+jerry_value_as_integer (const jerry_value_t value);
+```
+
+- `value` - api value
+- return
+  - 0, if value is not a number
+  - integer representation of the number, otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t number_val = jerry_create_number (123321);
+  double number = jerry_value_as_integer (number_val);
+  jerry_release_value (number_val);
+}
+```
+
+## jerry_value_as_int32
+
+**Summary**
+
+Convert any number to int32 number. Returns 0 if value is not number.
+Based on ECMA 262 v11 7.1.6
+
+**Prototype**
+
+```c
+int32_t
+jerry_value_as_int32 (const jerry_value_t value);
+```
+
+- `value` - api value
+- return
+  - 0, if value is not a number
+  - int32 representation of the number, otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t number_val = jerry_create_number (123321);
+  int32_t number = jerry_value_as_int32 (number_val);
+  jerry_release_value (number_val);
+}
+```
+
+## jerry_value_as_uint32
+
+**Summary**
+
+Convert any number to uint32 number. Returns 0 if value is not number.
+Based on ECMA 262 v11 7.1.7
+
+**Prototype**
+
+```c
+uint32_t
+jerry_value_as_uint32 (const jerry_value_t value);
+```
+
+- `value` - api value
+- return
+  - 0, if value is not a number
+  - uint32 representation of the number, otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t number_val = jerry_create_number (123321);
+  uint32_t number = jerry_value_as_uint32 (number_val);
+  jerry_release_value (number_val);
+}
+```
 
 # Functions for promise objects
 
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -14,6 +14,7 @@
  */
 
 #include <stdio.h>
+#include <math.h>
 
 #include "debugger.h"
 #include "ecma-alloc.h"
@@ -1718,6 +1719,83 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
   return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_bigint_not_supported_p)));
 #endif /* ENABLED (JERRY_BUILTIN_BIGINT) */
 } /* jerry_value_to_bigint */
+
+/**
+ * Convert any number to integer number.
+ *
+ * Note:
+ *      For non-number values 0 is returned.
+ *
+ * @return integer representation of the number.
+ */
+double
+jerry_value_as_integer (const jerry_value_t value) /**< input value */
+{
+  jerry_assert_api_available ();
+
+  if (!ecma_is_value_number (value))
+  {
+    return 0;
+  }
+
+  double number = ecma_get_number_from_value (value);
+
+  if (ecma_number_is_nan (number))
+  {
+    return ECMA_NUMBER_ZERO;
+  }
+
+  if (ecma_number_is_zero (number) || ecma_number_is_infinity (number))
+  {
+    return number;
+  }
+
+  ecma_number_t floor_fabs = (ecma_number_t) floor (fabs (number));
+
+  return ecma_number_is_negative (number) ? -floor_fabs : floor_fabs;
+} /* jerry_value_as_integer */
+
+/**
+ * Convert any number to int32 number.
+ *
+ * Note:
+ *      For non-number values 0 is returned.
+ *
+ * @return int32 representation of the number.
+ */
+int32_t
+jerry_value_as_int32 (const jerry_value_t value) /**< input value */
+{
+  jerry_assert_api_available ();
+
+  if (!ecma_is_value_number (value))
+  {
+    return 0;
+  }
+
+  return ecma_number_to_int32 (ecma_get_number_from_value (value));
+} /* jerry_value_as_int32 */
+
+/**
+ * Convert any number to uint32 number.
+ *
+ * Note:
+ *      For non-number values 0 is returned.
+ *
+ * @return uint32 representation of the number.
+ */
+uint32_t
+jerry_value_as_uint32 (const jerry_value_t value) /**< input value */
+{
+  jerry_assert_api_available ();
+
+  if (!ecma_is_value_number (value))
+  {
+    return 0;
+  }
+
+  return ecma_number_to_uint32 (ecma_get_number_from_value (value));
+} /* jerry_value_as_uint32 */
 
 /**
  * Acquire specified Jerry API value.
--- a/jerry-core/include/jerryscript-core.h
+++ b/jerry-core/include/jerryscript-core.h
@@ -554,6 +554,9 @@ jerry_value_t jerry_value_to_object (const jerry_value_t value);
 jerry_value_t jerry_value_to_primitive (const jerry_value_t value);
 jerry_value_t jerry_value_to_string (const jerry_value_t value);
 jerry_value_t jerry_value_to_bigint (const jerry_value_t value);
+double jerry_value_as_integer (const jerry_value_t value);
+int32_t jerry_value_as_int32 (const jerry_value_t value);
+uint32_t jerry_value_as_uint32 (const jerry_value_t value);
 
 /**
  * Acquire types with reference counter (increase the references).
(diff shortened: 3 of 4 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,96 @@ jerry_value_to_bigint (const jerry_value_t value);
...
+
+```c
+jerry_value_t
+jerry_value_to_integer (const jerry_value_t value, double *number_p);
+```
+
+- `value` - api value
+- `number_p` - [out] double number
+- return value
+  - empty value, if success
```
zherczeg, line 3787: there is no empty value in api

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,84 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+                       uint32_t *number_p) /**< [out] double number */
+{
+  jerry_assert_api_available ();
+
+  if (ecma_is_value_error_reference (value))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+  }
+
+  ecma_number_t radix_num;
```
zherczeg, line 1759: radix_num? This is not toString

docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,96 @@ jerry_value_to_bigint (const jerry_value_t value);
 - [jerry_value_is_bigint](#jerry_value_is_bigint)
 - [jerry_get_bigint_digits](#jerry_get_bigint_digits)
 
+## jerry_value_to_integer
+
+**Summary**
+
+Call the integer ecma builtin operation on the api value.
```
zherczeg, line 3775: Should be more descriptive.

docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,96 @@ jerry_value_to_bigint (const jerry_value_t value);
...
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_value_to_integer (const jerry_value_t value, double *number_p);
```
zherczeg, line 3781: I think this is not a conversion operation, since the result is undefined on success. It is more like a value casting, so `jerry_value_as_integer` could be a better name. Maybe an `jerry_value_as_number` could be added as well for convenience.

zherczeg (changes requested):

docs/02.API-REFERENCE.md
rerobika: I cannot agree with that. If you can ensure that the value passed to `jerry_value_as_integer` is an number then this is a `cast`.  This method creates integer number from any JS value so it's definitely a conversion.  I'd keep the `jerry_value_to_integer` (and same for the other ones) name for the function. Also better keep it sync with the abstract operation defined in the standard.

docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,96 @@ jerry_value_to_bigint (const jerry_value_t value);
...
+  - thrown error, otherwise
+
+**Example**
+
+```c
+{
+  jerry_value_t number_val = jerry_create_number (123321);
+  double number;
+
+  number_val = jerry_value_as_integer (number_val, &number);
```
rerobika, line 3797: Reassignment for `number_val` is misleading. I suggest : `jerry_value_t conv_result = jerry_value_as_integer (num_val, &number);`.  then release `num_val`

docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,96 @@ jerry_value_to_bigint (const jerry_value_t value);
 - [jerry_value_is_bigint](#jerry_value_is_bigint)
 - [jerry_get_bigint_digits](#jerry_get_bigint_digits)
 
+## jerry_value_as_integer
```
rerobika, line 3771: Keep it as **to** integer.

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,84 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
 #endif /* ENABLED (JERRY_BUILTIN_BIGINT) */
 } /* jerry_value_to_bigint */
 
+/**
+ * Call the integer ecma builtin operation on the api value.
```
rerobika, line 1723: Add the same description as in the documentation.

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,84 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+                        double *number_p)  /**< [out] double number */
+{
+  jerry_assert_api_available ();
+
+  if (ecma_is_value_error_reference (value))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+  }
+
+  return jerry_return (ecma_op_to_integer (value, number_p));
```
rerobika, line 1739: `ecma_op_to_integer` returns `ECMA_VALUE_EMPTY` on success which is not compatible with `jerry_return`. Handle the error case separately and return with undefined otherwise.

tests/unit-core/test-number-converter.c
```diff
@@ -0,0 +1,128 @@
...
+  test_to_interger (-1.1, -1);
+  test_to_interger (4294967295, 4294967295);
+  test_to_interger (-4294967295, -4294967295);
+  test_to_interger (4294967295, 4294967295);
+  test_to_interger (-4294967296, -4294967296);
+  test_to_interger (4294967297, 4294967297);
+  test_to_interger (-4294967297, -4294967297);
+
+  // few test-cases which return with error
+  jerry_value_t error_val =  jerry_create_error (JERRY_ERROR_TYPE, (const jerry_char_t *) "error");
```
rerobika, line 119: Please add tests for an object like this: `{ valueOf() { throw new TypeError("foo")}}`. You can evaluate this string to construct this object more easily.

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,84 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+} /* jerry_value_as_uint32 */
+
+/**
+ * Call the int32 ecma builtin operation on the api value.
+ *
+ * @return undefined value - if success
+ *         thrown error - otherwise
+ */
+jerry_value_t
+jerry_value_as_int32 (jerry_value_t value, /**< input value */
```
rerobika, line 1778: Ditto, same for as described above.

rerobika (changes requested):

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,91 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+                        double *number_p)  /**< [out] double number */
+{
+  jerry_assert_api_available ();
+
+  if (ecma_is_value_error_reference (value))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+  }
+
+  jerry_value_t return_val = ecma_op_to_integer (value, number_p);
```
rerobika, line 1739: This method returns `ecma_value_t`.

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,91 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+
+  if (ecma_is_value_error_reference (return_val))
```
rerobika, line 1741: `if (ECMA_IS_VALUE_ERROR(ret_val))`

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,91 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+  {
+    return jerry_return (return_val);
```
rerobika, line 1743: jerry_return -> jerry_throw

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,91 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+{
+  jerry_assert_api_available ();
+
+  if (ecma_is_value_error_reference (value))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (error_value_msg_p)));
+  }
+
+  ecma_number_t number;
+  value = ecma_op_to_number (value, &number);
```
rerobika, line 1767: use local variable instead.

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,91 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
```
rerobika, line 1796: ditto.

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1719,91 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+} /* jerry_value_to_uint32 */
+
+/**
+ * Call the ToInt32 abstrart operation for the given api value.
+ *
+ * @return undefined value - if success
+ *         thrown error - otherwise
+ */
+jerry_value_t
+jerry_value_to_int32 (jerry_value_t value, /**< input value */
```
rerobika, line 1785: This should be `const jerry_value_t value` by definition (same in all other functions.)

rerobika (changes requested):

kisbg:
@rerobika  Thank you for the review!

docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,102 @@ jerry_value_to_bigint (const jerry_value_t value);
...
+
+**Summary**
+
+Call the ToInteger abstrart operation for the given api value.
```
zherczeg, line 3775: abstrart -> abstract
The comment is incorrect btw. It should only convert numbers and nothing else. Similar cases below.

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1720,85 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+  ecma_op_to_number (value, &number);
+
+  if (ecma_number_is_nan (number))
+  {
+    return ECMA_NUMBER_ZERO;
+  }
+
+  ecma_number_t floor_fabs = (ecma_number_t) floor (fabs (number));
+
+  return ecma_number_is_negative (number) ? -floor_fabs : floor_fabs;;
```
rerobika, line 1752: double semicolons

docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,93 @@ jerry_value_to_bigint (const jerry_value_t value);
...
+```c
+int32_t
+jerry_value_as_int32 (const jerry_value_t value);
+```
+
+- `value` - api value
+- return
+  - 0, if value is not a number
+  - int32 representation of the number, otherwise
+
```
rerobika, line 3818: Only one new line.

jerry-core/include/jerryscript-core.h
```diff
@@ -555,6 +555,13 @@ jerry_value_t jerry_value_to_primitive (const jerry_value_t value);
 jerry_value_t jerry_value_to_string (const jerry_value_t value);
 jerry_value_t jerry_value_to_bigint (const jerry_value_t value);
 
+/**
+ * Get number value from 'jerry_value_t'.
+ */
```
rerobika, line 560: Simply remove this comment and continue the group above.

rerobika (changes requested):

rerobika (approved):
LGTM

docs/02.API-REFERENCE.md
```diff
@@ -3768,6 +3768,92 @@ jerry_value_to_bigint (const jerry_value_t value);
 - [jerry_value_is_bigint](#jerry_value_is_bigint)
 - [jerry_get_bigint_digits](#jerry_get_bigint_digits)
 
+## jerry_value_as_integer
+
+**Summary**
+
+Call the ToInteger abstract operation for the given api value.
```
zherczeg, line 3775: This comment is still incorrect. The `jerry_value_as_integer` does not call toNumber, only works with numbers (and returns with 0 otherwise). This must be described here (similar cases below).

jerry-core/api/jerry.c
```diff
@@ -1719,6 +1720,87 @@ jerry_value_to_bigint (const jerry_value_t value) /**< input value */
...
+jerry_value_as_int32 (const jerry_value_t value) /**< input value */
+{
+  jerry_assert_api_available ();
+
+  if (!ecma_is_value_number (value))
+  {
+    return 0;
+  }
+
+  double number = ecma_get_number_from_value (value);
```
zherczeg, line 1776: I would not create a local variable for this, but this is not a request. Feel free to do as you like it.

zherczeg (changes requested):

zherczeg (approved):
LGTM

PR #4328, opened 2020-11-05
Constructor check should return false for arrow and generator functions

galpeter:
The previous `ecma_is_constructor` implementation did not checked if the
target function was an arrow or generator function. This resulted in
an incorrect execution for these function types.

Depends on: #4264

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -129,24 +129,25 @@ ecma_op_is_callable (ecma_value_t value) /**< ecma value */
 } /* ecma_op_is_callable */
 
 /**
- * Checks whether the given object implements [[Construct]].
+ * Implement IsConstructor abstract operation.
  *
- * @return true - if the given object is constructor;
- *         false - otherwise
+ *
+ * @return ECMA_IS_VALID_CONSTRUCTOR - if object is a valid for constructor call
+ *         any other value - if object is not a valid constructor, the pointer contains the error message.
  */
-inline bool JERRY_ATTR_ALWAYS_INLINE
-ecma_object_is_constructor (ecma_object_t *obj_p) /**< ecma object */
+char *
+ecma_object_check_constructor (ecma_object_t *obj_p) /**< ecma object */
 {
   JERRY_ASSERT (!ecma_is_lexical_environment (obj_p));
 
   ecma_object_type_t type = ecma_get_object_type (obj_p);
 
-  if (type < ECMA_OBJECT_TYPE_PROXY)
+  if (JERRY_UNLIKELY (type < ECMA_OBJECT_TYPE_PROXY))
   {
-    return false;
+    return ECMA_ERR_MSG ("Invalid type for constructor call.");
   }
 
-  while (type == ECMA_OBJECT_TYPE_BOUND_FUNCTION)
+  while (JERRY_UNLIKELY (type == ECMA_OBJECT_TYPE_BOUND_FUNCTION))
   {
     ecma_bound_function_t *bound_func_p = (ecma_bound_function_t *) obj_p;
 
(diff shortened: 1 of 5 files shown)


rerobika (approved):
lgtm

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -786,7 +786,11 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
       case ECMA_OBJECT_TYPE_PROXY:
       {
         ecma_gc_mark_proxy_object (object_p);
-        break;
+        /* No need to free the properties of a proxy (there should be none) */
+        JERRY_ASSERT ((object_p->u1.property_list_cp == JMEM_CP_NULL)
+                      || (ECMA_GET_FIRST_BIT_FROM_POINTER_TAG (object_p->u1.property_list_cp) != 0)
+                      || (ECMA_GET_SECOND_BIT_FROM_POINTER_TAG (object_p->u1.property_list_cp) != 0));
```
zherczeg, line 792: The `(ECMA_GET_FIRST_BIT_FROM_POINTER_TAG (object_p->u1.property_list_cp) != 0` condition invalidates the whole point of `(object_p->u1.property_list_cp == JMEM_CP_NULL)`, doesn't it?

The condition should look like: `(object_p->u1.property_list_cp & ~0x3) == JMEM_CP_NULL` (of course with some macros)

zherczeg (changes requested):

jerry-core/ecma/base/ecma-gc.c
galpeter: The reason I separated into three different case is to clearly indicate what are the allowed cases/situations:
1) The bit 1 and 2 is not set
2) The bit 1 is set
3) The bit 2 is set (this also implies bit 1, maybe the assert order is incorrect)


Should I change it to the `~0x3` format? (Is there a macro for the 0x3 "bits"?)

zherczeg: I think you misunderstand me. It seems to me you want to test here that property_list_cp is null except the last two bits. However, the assert is wrong in that case, becasue the second or third case invalidates the first case.

zherczeg: I.e. 0x101 is accepted by the assert, while it should not be.

rerobika: Since currently only the first and the second pointer tags are used both assertions are correct. However I more like @zherczeg suggestion since it catches any misbehavior upon the third pointer tag is set by accident.

galpeter: Ok Will update, but that is not part of this PR, but the #4264 which this PR depends on.

rerobika:
@galpeter #4264 landed, please rebase.

rerobika (approved):
Still LGTM

jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -129,24 +129,29 @@ ecma_op_is_callable (ecma_value_t value) /**< ecma value */
...
   ecma_object_type_t type = ecma_get_object_type (obj_p);
 
-  if (type < ECMA_OBJECT_TYPE_PROXY)
+  if (JERRY_UNLIKELY (type < ECMA_OBJECT_TYPE_PROXY))
   {
-    return false;
+#if ENABLED (JERRY_ERROR_MESSAGES)
+    return "Invalid type for constructor call.";
+#else /* !ENABLED (JERRY_ERROR_MESSAGES) */
+    return "";
```
zherczeg, line 150: What about using ECMA_ERR_MSG ? In all places. Should simplify the code a lot.

zherczeg (changes requested):

jerry-core/ecma/operations/ecma-function-object.c
galpeter: ECMA_ERR_MSG returns NULL which in our case represents that there was no error.

zherczeg: If it would be once, that would be ok, but there are so many similar cases. We should do something with it.

jerry-core/ecma/base/ecma-gc.c
zherczeg: I may misunderstand something, but this is a logical or operation. The first part check if the pointer is null, the second part checks whether the fisrt bit is set. The problem is, if the second part is true, it doesn't matter the first part. Based on your explanation, 3 values are valid: 0x0, 0x1, 0x3. But the assert accepts 0x101, which is not allowed. So the assert is wrong.

jerry-core/ecma/operations/ecma-function-object.c
galpeter: Updated, now the `ECMA_ERR_MSG` returns an empty string if the error messages are disabled, and the messages passed for the raise_error* methods are checking for NULL and "" strings. Is this version ok? Or should I create another macro beside the `ECMA_ERR_MSG` which can return empty strings instead?

zherczeg: "" may increase binary size. What about returning with `(char*)0x1` on success? We could define a macro for that.

zherczeg (approved):
LGTM

PR #4331, opened 2020-11-10
Drop the minimal variant of the default port implementation

akosthekiss:
The minimal variant became quite meaningless lately. There were two
port APIs originally that had extra functions in the default port
in addition to the core-mandated implementations: the I/O and
Termination port APIs. However, the extra Termination API code was
removed a year ago, leaving some minimal extension in the I/O port
only. As the overhead of the extension is negligible, it is not
worth maintaining two library variants.

Therefore
- this commit removes the minimal variant of the default port lib,
- rewrites uses of the minimal variant to use the variant with the
  I/O extension, and
- updates targets where I/O port code was copy-n-pasted.


rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4335, opened 2020-11-12
Run test262-esnext tests with system allocator

ossy-szeged:
There are many tests need more memory than 512Kb,
system allocator is needed to have larger heap.

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -108,7 +108,7 @@ TEST262_ES2015_TEST_SUITE_OPTIONS = [
 
 # Test options for test262-esnext
 TEST262_ESNEXT_TEST_SUITE_OPTIONS = [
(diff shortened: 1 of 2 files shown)


tools/run-tests.py
```diff
@@ -108,7 +108,7 @@ def skip_if(condition, desc):
 
 # Test options for test262-esnext
 TEST262_ESNEXT_TEST_SUITE_OPTIONS = [
-    Options('test262_tests_esnext', OPTIONS_PROFILE_ESNEXT + ['--line-info=on', '--error-messages=on']),
+    Options('test262_tests_esnext', OPTIONS_PROFILE_ESNEXT + ['--line-info=on', '--error-messages=on', '--compile-flag=-m32', '--system-allocator=on']),
```
zherczeg, line 111: You can use 32 bit cpointers on on 64 bit as well, so you don't need to disable tests.

ossy-szeged: Good idea, I'll check it with cpointer and larger heap limit.

PR #4344, opened 2020-11-19
Add Jerryscript to OSS-fuzz

AdamKorcz:
I have been working on settin up continuous fuzzing of Jerryscript on OSS-fuzz. This will allow Google to run the existing and all future fuzzers continuously with the goal of finding bugs and security vulnerabilities. If and when a bug is found, all maintainers on the mailing list get notified with a detailed bug report.

Fuzzing continuously has many advantages, some of which are:

- It makes sure that the fuzzers are actually run. It happens often that efforts are put into writing the fuzzers, but they are only rarely run if at all.
- Fuzzers can find deeper bugs with more time. There are cases of bugs having been found after several CPU years of continuous fuzzing. 

I have the integration build ready to setup continuous fuzzing of Jerryscript by way of OSS-fuzz. To complete the integration, all I need are the email addresses of maintainers to receive potential bug reports. These will be added to a public list that can be changed at any time. No more changes will need to be made on the Jerryscript repository at this point, but overtime it is recommended to move the build scripts upstream from OSS-fuzz. This usually happens a few months after integration.

Regarding the changes made in this PR: If there is interest from the maintainers to setup continuous fuzzing by way of OSS-fuzz, then the -fsanitize=address flag will need to be removed from the build system. Once integrated, I will be happy to bring this flag back with a seperate flag in the tools/build.py script, but for now I suggest it just gets removed.

### DIFF
--- a/jerry-main/CMakeLists.txt
+++ b/jerry-main/CMakeLists.txt
@@ -59,7 +59,7 @@ endmacro()
 # Jerry with libfuzzer support
 if(JERRY_LIBFUZZER)
   jerry_create_executable("jerry-libfuzzer" "libfuzzer.c")
-  target_link_libraries("jerry-libfuzzer" jerry-port-default -fsanitize=fuzzer)
+  target_link_libraries("jerry-libfuzzer" jerry-port-default)
 endif()
 
 # Jerry standalones


akosthekiss:
Hi @AdamKorcz , thanks for reaching out to us with the idea of setting up OSS-fuzz for JerryScript. The general idea is OK, even appealing, I'd say. But I have some "buts", or questions.

> This will allow Google to run the existing and all future fuzzers

Q: Are you speaking on behalf of Google?

> all I need are the email addresses of maintainers to receive potential bug reports. These will be added to a public list that can be changed at any time.

Q: Is that _you_ who needs the email addresses? Who collects the email addresses / where are the email addresses collected? Where is that mailing list? How does a change happen? Is there a proxy who needs to be contacted to manage the addresses? (GDPR?)

> the -fsanitize=address flag will need to be removed from the build system. Once integrated, I will be happy to bring this flag back with a seperate flag in the tools/build.py script, but for now I suggest it just gets removed.

Q: Put simply: Why? I see no reason to remove a working feature from the build system and degrade its behaviour for a promise of future re-addition (in several months time?). If the tweaking of -fsanitize=address is the only way to get things working, then that change should be proposed now so that we can see what this is all about.

AdamKorcz:
Hi @akosthekiss, thank you for your good questions. 

> Q: Are you speaking on behalf of Google?

No. I am an independent contributor to the OSS-fuzz project. I will be happy to refer you to a member of the OSS-fuzz team.

> Q: Is that you who needs the email addresses? Who collects the email addresses / where are the email addresses collected? Where is that mailing list? How does a change happen? Is there a proxy who needs to be contacted to manage the addresses? (GDPR?)

I need them for the purpose of adding them to the mailing list. You can see an example of a mailing list has [here](https://github.com/google/oss-fuzz/blob/master/projects/brotli/project.yaml). A change to that list happens by making a pull request to the oss-fuzz repository like [this](https://github.com/google/oss-fuzz/pull/4619/files). In terms of a proxy who needs to manage the addresses: No there isn't one. All updates are reported directly on the OSS-fuzz repository.
I am completely fine with integrating Jerryscript into OSS-fuzz and you sharing email addresses over there instead of here. Just let me know beforehand if you are open to sharing these.

> Q: Put simply: Why? I see no reason to remove a working feature from the build system and degrade its behaviour for a promise of future re-addition (in several months time?). If the tweaking of -fsanitize=address is the only way to get things working, then that change should be proposed now so that we can see what this is all about.

The -fsanitize flag needs to be set by OSS-fuzz. In terms of the timeframe for bringing back the address sanitizer in the build script: Once there is an OK from OSS-fuzz and from Jerryscript's side, I will be happy to write a separate rule to manually add the address sanitizer before anyting is merged anywhere.

I hope the above answers your questions, otherwise please feel free to ask. Also, I should mention that the public disclosure policy of OSS-fuzz is 90 days.

AdamKorcz:
I have made a few modifications to allow the `-fsanitize` flag to be added in case the `--oss-fuzz` flag is not passed to `build.py`.

jerry-main/CMakeLists.txt
```diff
@@ -59,7 +59,11 @@ endmacro()
 # Jerry with libfuzzer support
 if(JERRY_LIBFUZZER)
   jerry_create_executable("jerry-libfuzzer" "libfuzzer.c")
-  target_link_libraries("jerry-libfuzzer" jerry-port-default -fsanitize=fuzzer)
+  if(OSS_FUZZ)
```
rerobika, line 62: please keep the `JERRY_` prefix for cmake variables.

tools/build.py
```diff
@@ -188,6 +190,7 @@ def build_options_append(cmakeopt, cliarg):
     build_options_append('JERRY_CMDLINE_SNAPSHOT', arguments.jerry_cmdline_snapshot)
     build_options_append('JERRY_CMDLINE_TEST', arguments.jerry_cmdline_test)
     build_options_append('JERRY_LIBFUZZER', arguments.libfuzzer)
+    build_options_append('OSS_FUZZ', arguments.oss_fuzz)
```
rerobika, line 193: Same as here.

rerobika (changes requested):

rerobika (approved):
LGTM

tools/build.py
```diff
@@ -96,6 +96,8 @@ def devhelp(helpstring):
                          help=devhelp('build test version of the jerry command line tool (%(choices)s)'))
     compgrp.add_argument('--libfuzzer', metavar='X', choices=['ON', 'OFF'], type=str.upper,
                          help=devhelp('build jerry with libfuzzer support (%(choices)s)'))
+    compgrp.add_argument('--oss-fuzz', metavar='X', choices=['ON', 'OFF'], type=str.upper,
+                         help=devhelp('build jerry to run on OSS-fuzz (%(choices)s)'))
```
akosthekiss, line 100: I'm a bit unsure whether this is the right approach to configure the libfuzzer command line tool for oss-fuzz. E.g., `--oss-fuzz=on` makes no sense without `--libfuzzer=on`. I wonder whether the choices of `--libfuzzer` should be extended instead, e.g., to allow ON, OFF, and OSS. OFF and ON would keep their original meaning and OSS would be a "special ON" -- this could cover the three exclusive approaches with the existing cli option. (Plus, it removes the need for a new cmake option as well.)

jerry-main/CMakeLists.txt
```diff
@@ -59,7 +59,11 @@ endmacro()
 # Jerry with libfuzzer support
 if(JERRY_LIBFUZZER)
   jerry_create_executable("jerry-libfuzzer" "libfuzzer.c")
-  target_link_libraries("jerry-libfuzzer" jerry-port-default -fsanitize=fuzzer)
+  if(JERRY_OSS_FUZZ)
+    target_link_libraries("jerry-libfuzzer" jerry-port-default)
+  else()
+    target_link_libraries("jerry-libfuzzer" jerry-port-default -fsanitize=fuzzer)
+  endif()
```
akosthekiss, line 66: I'd suggest reducing duplication. If I'm not mistaken `target_link_libraries` can be called multiple times and that will not overwrite previous items but concatenate them. (From the cmake docs: "Repeated calls for the same `<target>` append items in the order called.") Thus, something like the following might work (already using the on/oss/off approach):

```cmake
target_link_libraries("jerry-libfuzzer" jerry-port-default)
if(NOT (JERRY_LIBFUZZER STREQUAL "OSS"))
  target_link_libraries("jerry-libfuzzer" -fsanitize=fuzzer)
endif()
```

akosthekiss (changes requested):

akosthekiss:
Also, please, squash commits. The project has a one commit per PR policy.

tools/build.py
AdamKorcz: You are correct that fuzzing makes little sense without the sanitizers. In the case of OSS-fuzz, however, the flags will be set by OSS-fuzz itself to fuzz with other sanitizers, and forcing a specific sanitizer on the compilation will be incorrect.

akosthekiss: Nah, that's not what I was trying to point out. In the currently suggested form, the cmake has:
```cmake
if(JERRY_LIBFUZZER)
  jerry_create_executable("jerry-libfuzzer" "libfuzzer.c")
  if(JERRY_OSS_FUZZ)
    target_link_libraries("jerry-libfuzzer" jerry-port-default)
  else()
    target_link_libraries("jerry-libfuzzer" jerry-port-default -fsanitize=fuzzer)
  endif()
endif()
```

That is, if `JERRY_LIBFUZZER` is falsy, then the value of `JERRY_OSS_FUZZ` is indifferent. So, `JERRY_OSS_FUZZ` is not an independent configuration variable. The valid combination are `JERRY_LIBFUZZER==OFF`, `JERRY_LIBFUZZER==ON` and `JERRY_OSS_FUZZ==OFF`, and `JERRY_LIBFUZZER==ON` and `JERRY_OSS_FUZZ==ON`. This means that 3 possible settings are in the system. Therefore, I was suggesting to have ON, OSS, and OFF as valid values for `JERRY_LIBFUZZER` instead.

(The command line options for `tools/build.py` have a 1-to-1 correspondence to cmake options. So, the same goes for the `--libfuzzer` option of the script. It could have `['ON', 'OSS', 'OFF']` as valid choices.)

(Plus, if I'm not mistaken, `ON` and `OSS` are both "truthy" in cmake, that's why
```cmake
if(JERRY_LIBFUZZER)
  target_link_libraries("jerry-libfuzzer" jerry-port-default)
  if(NOT (JERRY_LIBFUZZER STREQUAL "OSS"))
    target_link_libraries("jerry-libfuzzer" -fsanitize=fuzzer)
  endif()
endif()
```
could simply work.)

akosthekiss:
Hm. I've been experimenting with OSS-Fuzz during the past few days and read the relevant docs. It seems that it it possible to set up JerryScript for/in OSS-Fuzz without _any_ modifications to the project (build system included). Let's put this PR on hold a bit.

AdamKorcz:
> It seems that it it possible to set up JerryScript for/in OSS-Fuzz without any modifications to the project (build system included)

That is correct. I will amend the setup to not introduce any intrusive modifications to Jerryscript. Are you interested in receiving the bug reports for Jerryscript @akosthekiss ?

AdamKorcz:
As discussed I have amended the build files on the OSS-fuzz side, and the build now succeeds: https://github.com/google/oss-fuzz/pull/4831

AdamKorcz:
@akosthekiss Kind reminder. Let me know if there is anything you need from my side.

PR #4346, opened 2020-11-25
Do not check call_arguments twice through parsing function call in statement

memnoth:
Checking call_arguments twice happens in a special case that
only if call_arguments <= 1 and opcode variable has CBC_EXT_SUPER_CALL.

`--unittests` and `--jerry-tests` are done.

### DIFF
--- a/jerry-core/parser/js/js-parser-expr.c
+++ b/jerry-core/parser/js/js-parser-expr.c
@@ -2630,8 +2630,7 @@ parser_process_unary_expression (parser_context_t *context_p, /**< context */
             continue;
           }
         }
-
-        if (call_arguments == 2)
+        else if (call_arguments == 2)
         {
           if (opcode == CBC_CALL)
           {


zherczeg:
What is the exact optimization here? if the previous statement ends with break/continue/return/etc. the else should have no effect

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -2619,26 +2619,25 @@ parser_process_unary_expression (parser_context_t *context_p, /**< context */
...
-          if (opcode == CBC_NEW)
+          else if (opcode == CBC_NEW)
           {
             parser_emit_cbc (context_p, (uint16_t) (CBC_NEW0 + call_arguments));
             continue;
           }
         }
-
-        if (call_arguments == 2)
+        else if (call_arguments == 2)
```
memnoth, line 2633: @zherczeg 

I think this is worth only if `call_arguments <= 1` and opcode is `CBC_EXT_SUPER_CALL`. But the others, changes on opcode arms have no effect as you said.

The commit message doesn't make sense with this, so if it is worth doing, I'll change the commit and ask for a review again or close it.

zherczeg: Yes this one is a good change, worth a commit.

memnoth: Thank you for the review!

zherczeg (approved):
LGTM

rerobika (approved):
LGTM

PR #4349, opened 2020-11-30
Handle revoked Proxy during [[Get]] operation

galpeter:
The handling of Proxy.[[Get]] was not fully correctly in the case when the
Proxy was revoked during the execution of the handler.

### DIFF
--- a/jerry-core/ecma/operations/ecma-function-object.c
+++ b/jerry-core/ecma/operations/ecma-function-object.c
@@ -919,6 +919,19 @@ ecma_op_get_prototype_from_constructor (ecma_object_t *ctor_obj_p, /**< construc
   if (!ecma_is_value_object (proto))
   {
     ecma_free_value (proto);
+
+#if ENABLED (JERRY_BUILTIN_PROXY)
+    if (ECMA_OBJECT_IS_PROXY (ctor_obj_p))
+    {
+      ecma_proxy_object_t *proxy_obj_p = (ecma_proxy_object_t *) ctor_obj_p;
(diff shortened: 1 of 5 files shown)


jerry-core/ecma/operations/ecma-function-object.c
```diff
@@ -919,6 +919,19 @@ ecma_op_get_prototype_from_constructor (ecma_object_t *ctor_obj_p, /**< construc
...
   {
     ecma_free_value (proto);
+
+#if ENABLED (JERRY_BUILTIN_PROXY)
+    if (ECMA_OBJECT_IS_PROXY (ctor_obj_p))
+    {
+      ecma_proxy_object_t *proxy_obj_p = (ecma_proxy_object_t *) ctor_obj_p;
+      if (ecma_is_value_null (proxy_obj_p->handler))
+      {
+        ecma_raise_type_error (ECMA_ERR_MSG ("Prototype from revoked Proxy is invalid."));
```
zherczeg, line 929: This message should be improved. E.g: `Getting the prototype of an invalid Proxy is not allowed`

zherczeg (approved):
LGTM

rerobika (approved):
LGTM

jerry-core/ecma/operations/ecma-function-object.c
rerobika: I'd rather suggest: `'prototype' property of revoked proxy is inaccessible`.

zherczeg: Good suggestion.

Issue #4244, opened 2020-09-29
Out-of-bound read in main_print_unhandled_exception


paintedveil5:
On line 180 of the main-utils.c, 
`        do
        {
          ch = source_p[pos++];
          jerry_port_log (JERRY_LOG_LEVEL_ERROR, "%c", ch);
        }
        while (ch != '\n' && char_count++ < SYNTAX_ERROR_MAX_LINE_LENGTH);`
will out-of-bound read at most 0x100 characters when an syntax error happened at the end of the input file.

##### JerryScript revision
0ffe1665bd0ee457dd4c1083811e1592a5e2908d
(latest master - 2020.09.28.)

###### Build platform
Ubuntu 16.04.1 (Linux 4.15.0-117-generic x86_64)

###### Build steps
1. Compile with Address Sanitizer
```sh 
python ./tools/build.py --clean --debug --compile-flag=-fsanitize=address --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --lto=off --error-message=on --system-allocator=on 
```
2.  Compile without Address Sanitizer
```sh
python ./tools/build.py --clean --debug  --compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer --compile-flag=-fno-common --lto=off --error-message=on --system-allocator=on 
```
###### Test case
```sh
 //debugger.js
 try {
 } 
```

###### Execution steps

```sh
./jerry debugger.js
```
###### Output
1. Compile with Address Sanitizer
```sh
==46414==ERROR: AddressSanitizer: heap-buffer-overflow on address 0xf640077b at pc 0x0804ac7c bp 0xffb3af18 sp 0xffb3af08
READ of size 1 at 0xf640077b thread T0
    #0 0x804ac66 in main_print_unhandled_exception /home/lily/Desktop/debugger/jerryscript/jerry-main/main-utils.c:196
    #1 0x80498fc in main /home/lily/Desktop/debugger/jerryscript/jerry-main/main-unix.c:142
    #2 0xf782b646 in __libc_start_main (/lib32/libc.so.6+0x18646)
    #3 0x8049320  (/home/lily/Desktop/debugger/jerryscript/build/bin/jerry+0x8049320)

0xf640077b is located 0 bytes to the right of 27-byte region [0xf6400760,0xf640077b)
allocated by thread T0 here:
    #0 0xf7a5ddee in malloc (/usr/lib32/libasan.so.2+0x96dee)
[...]

PR #4350, opened 2020-11-30
Correct syntax error printing in case of end of file

galpeter:
The old variant of the syntax error printing method could incorrectly over-read the source file contents.

Fixes: #4244


zherczeg (approved):
LGTM

rerobika (approved):
LGTM

PR #4351, opened 2020-11-30
Correct Regexp literal error release

galpeter:
The error object for an incorrect Regexp literal was derefed then the error message
was accessed. This could lead to a state where the message was not available but
it was still accessed.

This could occur in case of heavy memory load or with mem stress mode.


zherczeg (approved):
LGTM

rerobika (approved):
LGTM

PR #4347, opened 2020-11-25
Rework built-in processing of JerryScript

zherczeg:
- All built-ins are native functions now
- Native handlers have a built-in id: ECMA_BUILTIN_ID_HANDLER
- Built-in routine identifiers start from 1
- Built-in routines have an own flag set
- Name property of routines is resolved dynamically
- Style fixes

PR #4352, opened 2020-12-01
Template literal arrays should not be marked.

zherczeg:
The idea is that template literal objects are frozen arrays, so their properties and prototype could not be changed. Their properties are strings (no need marking) and their prototype is a built-in (marked elsewhere). Therfore marking these objects is unnecessary. This is compatible with a future improvement called realms, since the realm of the function and its own template literal is the same.

There are two ways to handle unmarking these objects:
- The current choice: prototype refing is moved to the end of `ecma_gc_mark`. Slight binary increase and each array object has an extra check.
- Other option: check these objects when `root` objects are marked (their refcount is always > 0). Slight binary increase and each root object has an extra check.

I am ok with both, let me know your opinion.

Make array object big endian compatible: I noticed that the current code is wrong on big endian system, since the `ecma_property_t` is one byte long, and mapped to the highest byte instead of the lowest in big endian system.

### DIFF
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -626,13 +626,6 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
   }
   else
   {
-    jmem_cpointer_t proto_cp = object_p->u2.prototype_cp;
-
-    if (proto_cp != JMEM_CP_NULL)
-    {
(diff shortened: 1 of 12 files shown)


zherczeg:
#4347 should land first

rerobika:
Please rebase.

zherczeg:
Rebased.

rerobika (approved):
LGTM

rerobika (approved):
LGTM

rerobika:
I'm fine with the first approach!

LaszloLango (approved):
LGTM

PR #4356, opened 2020-12-08
Continue working on realms

zherczeg:
- Rework symbols to have the same value across realms
- Support realms for native functions
- Support test262
- Use new.target realms for constructing intrinsics

Note: lots of built-ins ignore new.target so realms support cannot be added for them at the moment.


jerry-main/main-utils.c
```diff
@@ -148,6 +148,38 @@ test262_eval_script (const jerry_value_t func_obj_val, /**< function object */
   return ret_value;
 } /* test262_eval_script */
 
+/**
+ * $262.evalScript
```
rerobika, line 152: `$262.createRealm`

jerry-main/main-utils.c
```diff
@@ -148,6 +148,38 @@ test262_eval_script (const jerry_value_t func_obj_val, /**< function object */
...
+ *
+ * A function which accepts a string value as its first argument and executes it
+ *
+ * @return completion of the script parsing and execution.
```
rerobika, line 156: The whole description seems incorrect.

jerry-main/main-utils.c
```diff
@@ -148,6 +148,38 @@ test262_eval_script (const jerry_value_t func_obj_val, /**< function object */
...
+  jerry_value_t name_val = jerry_create_string ((const jerry_char_t *) "global");
+  jerry_value_t realm_val = jerry_create_realm ();
+
+  jerry_value_t result_val = jerry_set_property (result_object, name_val, realm_val);
+
+  jerry_release_value (result_val);
+  jerry_release_value (realm_val);
+  jerry_release_value (name_val);
+
+  return result_object;
```
rerobika, line 180: I think this is incorrect.

According the int `INTERPRETING.md` in test262
```
createRealm - a function which creates a new ECMAScript Realm, defines this API on the new realm's global object, and returns the $262 property of the new realm's global object
```

So what I think we need to do there:
1, Replace `static void register_test262 (void)` to `static jerry_value_t register_test262 (jerry_value_t global_obj)`.
 - which returns the registered `test262` object to the given `global_obj`.
 - Releases the given `global_obj`

2, Replace the current usage of `register_test262()` to `jerry_release_value(register_test262 (jerry_get_global_object ()))`
3, In `test262_create_realm` create a new realm `realm_val` and return with `register_test262 (new_realm)`;

rerobika (changes requested):

jerry-main/main-utils.c
zherczeg: Good idea, it seems it fixed another 4 tests.

zherczeg:
Thank you for the review, patch is ready.

jerry-main/main-utils.c
```diff
@@ -148,17 +148,47 @@ test262_eval_script (const jerry_value_t func_obj_val, /**< function object */
...
 
+static jerry_value_t
+register_test262 (jerry_value_t global_obj);
+
+/**
+ * $262.createRealm
+ *
+ * A function which creates a new realm object
+ *
+ * @return new realm object
```
rerobika, line 159: `@return $test262 property of the new realm object`

rerobika (approved):
LGTM after the description update

dbatyai (approved):
LGTM

PR #4361, opened 2020-12-14
Correctly release wrapper object in JSON stringify

galpeter:
When an error occurs accessing a property during JSON stringify call
the wrapper object is not freed at the correct place.


rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #4362, opened 2020-12-14
Add missing BigInt type for value type query API

galpeter:
Extended the `jerry_type_t` enum with `JERRY_TYPE_BIGINT` and added it to the `jerry_value_get_type`.

### DIFF
--- a/tests/unit-core/test-api-value-type.c
+++ b/tests/unit-core/test-api-value-type.c
@@ -99,6 +99,31 @@ main (void)
     jerry_release_value (symbol_desc_value);
   }
 
+  if (jerry_is_feature_enabled (JERRY_FEATURE_BIGINT))
+  {
+    /* Check simple bigint value type */
+    uint64_t digits_buffer[2] = { 1, 0 };
+    jerry_value_t value_bigint = jerry_create_bigint (digits_buffer, 2, false);
+    jerry_type_t value_type_info = jerry_value_get_type (value_bigint);
+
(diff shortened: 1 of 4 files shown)


tests/unit-core/test-api-value-type.c
```diff
@@ -99,6 +99,31 @@ main (void)
...
+
+    /* Check bigint wrapped in object type */
+    jerry_char_t object_bigint_src[] = "Object(5n)";
+    jerry_value_t object_bigint = jerry_eval (object_bigint_src, sizeof (object_bigint_src) - 1, JERRY_PARSE_NO_OPTS);
+    TEST_ASSERT (!jerry_value_is_error (object_bigint));
+
+    jerry_type_t object_type_info = jerry_value_get_type (object_bigint);
+
+    TEST_ASSERT (object_type_info != JERRY_TYPE_NONE);
+    TEST_ASSERT (object_type_info == JERRY_TYPE_OBJECT);
```
rerobika, line 122: Can you extend the `jerry_object_type_t` enum with bigint wrapped objects as well?

galpeter: That is already there (if I remember correctly).

rerobika: Indeed! Just ignore my comment.

rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #4371, opened 2020-12-21
Fix the return value of jerry_to_property_descriptor function

zherczeg:


rerobika (approved):
LGTM

docs/02.API-REFERENCE.md
```diff
@@ -4084,8 +4085,10 @@ jerry_to_property_descriptor (jerry_value_t obj_value, jerry_property_descriptor
 - `src_prop_desc_p` - the output property descriptor.
 - return
   - [jerry_value_t](#jerry_value_t)
-  - true - if success
-  - false - if error occurs
+  - true - if successful
+  - error - otherwise
+
```
galpeter, line 4090: at other places we use the:
```
 - true, if success
 - thrown error, otherwise
```
syntax. In the future we should reword all places to a common format. This is ok for now.

galpeter: The function prototype above should also be changed.

docs/02.API-REFERENCE.md
```diff
@@ -4104,7 +4107,7 @@ example (void)
 
   jerry_set_property (object, prop_name, prop_name);
```
galpeter, line 4108: afaik `jerry_set_property` should also be released just to keep it consistent.

jerry-core/ecma/operations/ecma-conversion.c
```diff
@@ -723,8 +723,8 @@ ecma_op_from_property_descriptor (const ecma_property_descriptor_t *src_prop_des
 ecma_value_t
 ecma_op_to_property_descriptor (ecma_value_t obj_value, /**< object value */
                                 ecma_property_descriptor_t *out_prop_desc_p) /**< [out] filled property descriptor
-                                                                                  if return value is normal
-                                                                                  empty completion value */
+                                                                              *   if return value is ECMA_VALUE_TRUE,
```
galpeter, line 726: Is the `ECMA_VALUE_TRUE` ok here? in the api above the `ECMA_VALUE_EMPTY` is expected.

tests/unit-core/test-to-property-descriptor.c
```diff
@@ -71,6 +72,13 @@ main (void)
   jerry_release_value (prop_name);
   jerry_release_value (value);
   jerry_free_property_descriptor_fields (&prop_desc);
+
+  object = jerry_create_null ();
+  result = jerry_to_property_descriptor (object, &prop_desc);
```
galpeter, line 77: What will happen if the input "object" is not an object? Ex: a simple number?

jerry-core/api/jerry.c
```diff
@@ -4066,41 +4067,31 @@ jerry_from_property_descriptor (const jerry_property_descriptor_t *src_prop_desc
...
-* ToPropertyDescriptor abstract operation.
-*
-* @return thrown error - if any conversion error happens
-*        ECMA_VALUE_TRUE - otherwise.
-*/
-bool
+ * ToPropertyDescriptor abstract operation.
+ *
+ * @return true - if the conversion is successful
+ *         thrown error - otherwise
```
galpeter, line 4073: This description is not the same as in the docs. We should keep it the same.

galpeter (changes requested):

tests/unit-core/test-to-property-descriptor.c
zherczeg: The called function (`ecma_op_to_property_descriptor`) throws a type error. There is a test for this.

jerry-core/api/jerry.c
zherczeg: This is not the same in other cases as well. The doc is more verbose (intentionally).

galpeter (approved):
lgtm

PR #4378, opened 2020-12-28
timeout are not working on windows.

lygstate:
Fix the test262 timeout on Windows

Use the platform independent python threading.Timer instead of the unix only timeout tool.

The timeout error are like the following things:

```
C:\Users\lygstate>timeout -version
Error: The specified timeout (/T) value is invalid. The valid range is from -1 to 99999 seconds.

C:\Users\lygstate>timeout 0 python
Error: invalid syntax. The default option does not allow more than '1' times.
Type "TIMEOUT /?" to learn how to use it.
```

timeout option are added for future ES5.1 test262 usage
in pull request
https://github.com/jerryscript-project/jerryscript/pull/4391

The timeout setting to 60 to getting maximal tolerance
By doing this also fixes tests for es2015:
built-ins/decodeURIComponent/S15.1.3.2_A2.5_T1.js
built-ins/decodeURI/S15.1.3.1_A2.5_T1.js

### DIFF
--- a/tools/runners/test262-harness.py
+++ b/tools/runners/test262-harness.py
@@ -55,6 +55,7 @@ import xml.dom.minidom
 from collections import Counter
 
 import signal
+import threading
 import multiprocessing
 
 #######################################################################
@@ -63,6 +64,10 @@ import multiprocessing
 
 M_YAML_LIST_PATTERN = re.compile(r"^\[(.*)\]$")
 M_YAML_MULTILINE_LIST = re.compile(r"^ *- (.*)$")
+
+
+# The timeout of each test case
+TEST262_CASE_TIMEOUT = 5
 
 
 def yaml_load(string):
@@ -595,11 +600,14 @@ class TestCase(object):
             logging.info("exec: %s", str(args))
             process = subprocess.Popen(
                 args,
-                shell=is_windows(),
+                shell=False,
                 stdout=stdout.file_desc,
                 stderr=stderr.file_desc
             )
+            timer = threading.Timer(TEST262_CASE_TIMEOUT, process.kill)
+            timer.start()
             code = process.wait()
+            timer.cancel()
             out = stdout.read()
             err = stderr.read()
         finally:
--- a/tools/runners/run-test-suite-test262.py
+++ b/tools/runners/run-test-suite-test262.py
@@ -173,12 +173,6 @@ def main(args):
         util.set_timezone('Pacific Standard Time')
 
     command = (args.runtime + ' ' + args.engine).strip()
-    if args.es2015 or args.esnext:
-        try:
-            subprocess.check_output(["timeout", "--version"])
-            command = "timeout 5 " + command
-        except subprocess.CalledProcessError:
-            pass
 
     kwargs = {}
     if sys.version_info.major >= 3:


akosthekiss (changes requested):
Please, change commit message. The language used throughout the project is English.

Please, fix pylint check errors.

tools/runners/test262-harness.py
```diff
@@ -372,11 +372,6 @@ def validate_options(options):
         report_error("Couldn't find test path '%s'" % options.tests)
 
 
-def is_windows():
-    actual_platform = platform.system()
-    return (actual_platform == 'Windows') or (actual_platform == 'Microsoft')
-
-
```
ossy-szeged, line 379: Please don't remove it.

tools/runners/test262-harness.py
```diff
@@ -585,21 +580,21 @@ def get_parameter(match):
 
     @staticmethod
     def execute(command):
-        if is_windows():
-            args = '%s' % command
-        else:
-            args = command.split(" ")
+        args = command.split(" ")
```
ossy-szeged, line 583: We still need this trick, because run-test-suite-test262.py concatenates args.runtime and args.engine with space.

lygstate: > We still need this trick, because run-test-suite-test262.py concatenates args.runtime and args.engine with space.

Do you means es5.1 need this trick?

lygstate: But according to https://github.com/jerryscript-project/jerryscript/pull/4378/checks?check_run_id=1620216059
 run-test-suite-test262.py are not affected

lygstate: @ossy-szeged Please show me the full command that need to running these tests.

lygstate: As you wish, revert it first, there is no need do this here @ossy-szeged

ossy-szeged: > @ossy-szeged Please show me the full command that need to running these tests.

tools/runners/run-test-suite-test262.py --runtime qemu-***
(or you can use RUNTIME environment variable too instead of command line option)

ossy-szeged:
The change looks good, but I would change the commit log to be more descriptive. Now it says what was wrong, but it should say what have been done in the commit. For example:

Fix the test262 timeout on Windows

Use the platform independent python threading.Timer instead of the unix only timeout tool.

lygstate:
> The change looks good, but I would change the commit log to be more descriptive. Now it says what was wrong, but it should say what have been done in the commit. For example:
> 
> Fix the test262 timeout on Windows
> 
> Use the platform independent python threading.Timer instead of the unix only timeout tool.

done

lygstate:
> Please, change commit message. The language used throughout the project is English.
> 
> Please, fix pylint check errors.

Fixed

tools/runners/test262-harness.py
```diff
@@ -613,7 +620,8 @@ def run_test_in(self, tmp):
         command = TestCase.instantiate_template(self.command_template, {
             'path': tmp.name
         })
-        (code, out, err) = TestCase.execute(command)
+        # Only es2015 or esnext need the timeout option.
+        (code, out, err) = TestCase.execute(command, 60)
```
lygstate, line 624: Add timeout option for the usage in es5.1,
setting to 60 for pass two time consuming tests.

LaszloLango (approved):
LGTM

tools/runners/test262-harness.py
```diff
@@ -595,11 +596,17 @@ def execute(command):
             logging.info("exec: %s", str(args))
             process = subprocess.Popen(
                 args,
-                shell=is_windows(),
+                shell=False,
                 stdout=stdout.file_desc,
                 stderr=stderr.file_desc
             )
-            code = process.wait()
+            if timeout is None:
```
akosthekiss, line 603: `timeout` never seems to be `None`

tools/runners/test262-harness.py
ossy-szeged: I'm not sure if it is a good idea to have 60 seconds timeout unconditionally,
because language/statements/for-of/iterator-next-reference.js in ES2015-test262
is a wrong test (outdated because of spec change) and stucks in an infinite loop.

Increasing timeout means that running ES2015-test262 will take an extra minute because of this wrong test.
8 seconds timeout is enough to remove 2 slow tests from es2015 excludelist. And tests run in 80 seconds
for me with 8 seconds timeout, but in 80+57 seconds with 60 seconds timeout.

In nutshell: I would keep general timeout as low as possible, but we could add extra timeout for ES5.1 if it is necessary.

tools/runners/test262-harness.py
lygstate: for future usage in es5.1

tools/runners/test262-harness.py
lygstate: 60 is needed to pass two new tests

lygstate: > I'm not sure if it is a good idea to have 60 seconds timeout unconditionally,
> because language/statements/for-of/iterator-next-reference.js in ES2015-test262
> is a wrong test (outdated because of spec change) and stucks in an infinite loop.
> 
> Increasing timeout means that running ES2015-test262 will take an extra minute because of this wrong test.
> 8 seconds timeout is enough to remove 2 slow tests from es2015 excludelist. And tests run in 80 seconds
> for me with 8 seconds timeout, but in 80+57 seconds with 60 seconds timeout.
> 
> In nutshell: I would keep general timeout as low as possible, but we could add extra timeout for ES5.1 if it is necessary.

60s needed to pass
  <test id="built-ins/decodeURIComponent/S15.1.3.2_A2.5_T1.js"><reason></reason></test>	
  <test id="built-ins/decodeURI/S15.1.3.1_A2.5_T1.js"><reason></reason></test>

These two tests consume lot of time in jerryscript, maybe it's a performance issue.

ossy-szeged: The bottleneck is the wrong language/statements/for-of/iterator-next-reference.js in test262-es2015. If we increase the timeout, we unnecessarily increase the runtime, because we have to wait to kill this wrong test. Otherwise the 2 unskipped es2015 test is also covered by esnext, because test262 maintainers know these are slow tests and refactored them to be faster. So we don't loose anything if we keep the timeout as is and leave these tests skipped.

lygstate: > The bottleneck is the wrong language/statements/for-of/iterator-next-reference.js in test262-es2015. If we increase the timeout, we unnecessarily increase the runtime, because we have to wait to kill this wrong test. Otherwise the 2 unskipped es2015 test is also covered by esnext, because test262 maintainers know these are slow tests and refactored them to be faster. So we don't loose anything if we keep the timeout as is and leave these tests skipped.

Not that simple, the debug version build jerryscript are slower, we need more time tolerance, and that won't increate the total test262 running time much, because tests are running in parallel, the testes consume lot of time won't be a bottleneck.

lygstate:
@akosthekiss timeout parameter removed, please take a look

tools/runners/test262-harness.py
lygstate: @ossy-szeged Revert to 5 secons.

tools/runners/test262-harness.py
```diff
@@ -595,11 +600,15 @@ def execute(command):
             logging.info("exec: %s", str(args))
             process = subprocess.Popen(
                 args,
-                shell=is_windows(),
+                shell=False,
                 stdout=stdout.file_desc,
                 stderr=stderr.file_desc
             )
+            timeout = TEST262_CASE_TIMEOUT
+            timer = threading.Timer(timeout, process.kill)
```
akosthekiss, line 608: these two lines can go in a single one (no need for the `timeout` variable when `TEST262_CASE_TIMEOUT` is already there)

akosthekiss (changes requested):

tools/runners/test262-harness.py
lygstate: for future usage and clear diff.
If you insist I revert it

akosthekiss: The diff would be just as clear with `timer = threading.Timer(TEST262_CASE_TIMEOUT, process.kill)`. I'm not sure about future use cases. If and when they kick in, it will be trivial to split this into two lines. To be reviewed at that point of time.

lygstate: > The diff would be just as clear with `timer = threading.Timer(TEST262_CASE_TIMEOUT, process.kill)`. I'm not sure about future use cases. If and when they kick in, it will be trivial to split this into two lines. To be reviewed at that point of time.

done

akosthekiss (approved):
LGTM

ossy-szeged (approved):
LGTM

PR #4379, opened 2020-12-28
only tools/packaging/test262.py force using python2.

lygstate:

### DIFF
--- a/tools/runners/run-test-suite-test262.py
+++ b/tools/runners/run-test-suite-test262.py
@@ -24,10 +24,12 @@ import sys
 
 import util
 
-def get_platform_cmd_prefix():
+def get_platform_cmd_prefix(is_python2):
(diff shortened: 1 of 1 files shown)


tools/runners/run-test-suite-test262.py
```diff
@@ -24,10 +24,12 @@
 
 import util
 
-def get_platform_cmd_prefix():
+def get_platform_cmd_prefix(is_python2: False):
```
akosthekiss, line 27: This isn't valid Python syntax.

akosthekiss (changes requested):

tools/runners/run-test-suite-test262.py
lygstate: fixed

ossy-szeged:
It looks good, just a note, it depends on #4381

PR #4380, opened 2020-12-28
Make CI to be running by upgrade to all ubuntu that need  gcc-multilib to 20.04

lygstate:


akosthekiss:
Why is the Ubuntu version bump necessary?

lygstate:
> Why is the Ubuntu version bump necessary?

Because on Ubuntu 18.04
`sudo apt-get install gcc-multilib` are always failing.

ossy-szeged:
I think it is a GH issue. But I prefer #4388 as a workaround instead of migrating to Ubuntu 20.04.

lygstate:
OK with that

PR #4382, opened 2020-12-28
Install pyserial and pylint properly

lygstate:

### DIFF
--- a/.github/workflows/gh-actions.yml
+++ b/.github/workflows/gh-actions.yml
@@ -17,7 +17,13 @@ jobs:
       - name: Updating packages
         run: sudo apt-get update
       - name: Initializtaion
-        run: sudo apt-get install doxygen vera++ cppcheck pylint pylint python-serial -q
+        run: |
+          sudo apt-get install doxygen vera++ cppcheck -q
+          sudo pip install pyserial pylint
+      - name: Display Python version
+        run: |
+          python -c "import sys; print(sys.version)"
+          /usr/bin/env python -c "import sys; print(sys.version)"
       - name: Signed-off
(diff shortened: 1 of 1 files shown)


.github/workflows/gh-actions.yml
```diff
@@ -15,7 +15,13 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - name: Initializtaion
-        run: sudo apt-get install doxygen vera++ cppcheck pylint pylint python-serial -q
+        run: |
+          sudo apt-get install doxygen vera++ cppcheck -q
+          sudo pip install pyserial pylint
```
akosthekiss, line 20: Whether python packages are installed via apt or pip is a matter of taste. I wouldn't call apt-based installation as improper -- as long as it works. And now that #4388 has landed, apt should be working again all right.

.github/workflows/gh-actions.yml
```diff
@@ -15,7 +15,13 @@ jobs:
     steps:
       - uses: actions/checkout@v1
       - name: Initializtaion
-        run: sudo apt-get install doxygen vera++ cppcheck pylint pylint python-serial -q
```
akosthekiss, line 18: it is true, however, that this line may use a minor fix (pylint unnecessarily is listed twice).

.github/workflows/gh-actions.yml
lygstate: Oh, I've done that because on Ubuntu 20.04, apt-get install python-serial would fail, pip install are more stable.

.github/workflows/gh-actions.yml
lygstate: Yeap, I've got that

Issue #4384, opened 2020-12-28
test262 are not match unix testing result on windows


lygstate:
###### JerryScript revision
9676500addbcc5974bf81f7b3726230d26f5c1a5

###### Build platform

```
C:\Users\lygstate>python -c "import platform; print(platform.platform())"
Windows-10-10.0.19041-SP0
```
###### Build steps
For ES 5.1
```
PS C:\work\study\languages\typescript\jerryscript> python tools/run-tests.py --test262
Build command:
cmd \
        /S \
        /C \
        C:\work\study\languages\typescript\jerryscript\tools\build.py \
        --profile=es5.1 \
        --builddir=C:\work\study\languages\typescript\jerryscript\build\tests\test262_tests \
        --install=C:\work\study\languages\typescript\jerryscript\build\tests\test262_tests\local
Test command:
TZ='America/Los_Angeles' \
cmd \
        /S \
        /C \
        C:\work\study\languages\typescript\jerryscript\tools\runners/run-test-suite-test262.py \
        --engine \
        C:\work\study\languages\typescript\jerryscript\build\tests\test262_tests\local\bin\jerry.exe --test262-object \
        --test-dir \
        C:\work\study\languages\typescript\jerryscript\tests/test262 \
        --es51
.................................................. Executed 5000 tests.
.................................................. Executed 10000 tests.
.................
=== Summary ===
 - Ran 11725 tests
 - Passed 11557 tests (98.6%)
 - Failed 168 tests (1.4%)
```

For ES6
```
PS C:\work\study\languages\typescript\jerryscript> python tools/run-tests.py --test262-es2015 update
Build command:
cmd \
        /S \
        /C \
        C:\work\study\languages\typescript\jerryscript\tools\build.py \
        --profile=es.next \
        --line-info=on \
        --error-messages=on \
        --builddir=C:\work\study\languages\typescript\jerryscript\build\tests\test262_tests_es2015 \   
        --install=C:\work\study\languages\typescript\jerryscript\build\tests\test262_tests_es2015\local
Test command:
TZ='America/Los_Angeles' \
cmd \
        /S \
        /C \
        C:\work\study\languages\typescript\jerryscript\tools\runners/run-test-suite-test262.py \
        --engine \
        C:\work\study\languages\typescript\jerryscript\build\tests\test262_tests_es2015\local\bin\jerry.exe --test262-object \
        --test-dir \
        C:\work\study\languages\typescript\jerryscript\tests/test262 \
[...]

PR #4389, opened 2020-12-30
Using a single test262-harness.py for es5.1, es2015 and esnext.

lygstate:
Remove the need of tools/packaging/test262.py from test262 es5.1 branch.
This script are not running the tests in parallel; so we can getting the
es5.1 tests running faster.
And it's not compatible with python3. We can use a single test262-harness.py
to resolve compatibility problem.

This is a superset of https://github.com/jerryscript-project/jerryscript/pull/4379
And fixes https://github.com/jerryscript-project/jerryscript/issues/4384

Depends on:
https://github.com/jerryscript-project/jerryscript/pull/4378


akosthekiss:
It seems that this PR is (ab)used again to replace local/private testing. @lygstate please, explain what you are doing and why. You have been requested not to do this, i.e., not to flood the project's CI (or maintainers' mailbox). Otherwise, I'll suggest maintainers to auto-close these PRs without any review.

lygstate:
> It seems that this PR is (ab)used again to replace local/private testing. @lygstate please, explain what you are doing and why. You have been requested not to do this, i.e., not to flood the project's CI (or maintainers' mailbox). Otherwise, I'll suggest maintainers to auto-close these PRs without any review.

Sorry for the disturb, I'll be carefull next time, this pull request depends two other pull request, and that disturb me.

lygstate:
@akosthekiss ready to review.

tools/runners/test262-harness.py
```diff
@@ -344,6 +391,8 @@ def build_options():
                       help="The command-line to run")
     result.add_option("--tests", default=path.abspath('.'),
                       help="Path to the tests")
+    result.add_option('--es51', default=False, action='store_true',
```
rerobika, line 394: How about `--es5.1` ? Personally I think the `51` is quite deceptive.

lygstate: how to do with self.suite.es51?
self.suite.es5_1?

rerobika: I'm fine with `es5_1` as well.

lygstate: > I'm fine with `es5_1` as well.

I have an extra problem, the es51 option are compiled from `run-test-suite-test262.py`
```
    group.add_argument('--es51', default=False, const='default',
                       nargs='?', choices=['default', 'all', 'update'],
                       help='Run test262 - ES5.1. default: all tests except excludelist, ' +
                       'all: all tests, update: all tests and update excludelist')
```

That's why I use it, maybe we do the rename in separate patch?

lygstate: > I'm fine with `es5_1` as well.

renamed to es5_1:)

tools/runners/test262-harness.py
```diff
@@ -708,9 +785,15 @@ def percent_format(partial, total):
...
     def __init__(self, options):
-        self.test_root = path.join(options.tests, 'test')
-        self.lib_root = path.join(options.tests, 'harness')
+        self.es5_1 = options.es5_1
+        if self.es5_1:
+            self.test_root = path.join(options.tests, 'test', "suite")
+            self.lib_root = path.join(options.tests, 'test', 'harness')
+        else:
+            self.test_root = path.join(options.tests, 'test')
+            self.lib_root = path.join(options.tests, 'harness')
```
rerobika, line 796:
```py
test_folder = '' if self.es5_1 else 'test';
self.test_root = path.join(options.tests, test_folder, 'test')
self.lib_root = path.join(options.tests, test_folder, 'harness')

```

lygstate: > ```python
> test_folder = '' if self.es5_1 else 'test';
> self.test_root = path.join(options.tests, test_folder, 'test')
> self.lib_root = path.join(options.tests, test_folder, 'harness')
> ```

Not what you think, it's 'suite'
            self.test_root = path.join(options.tests, 'test', "suite")

rerobika: Indeed, sorry I've missed that.

lygstate:
@akosthekiss please take a look

lygstate:
@ossy-szeged May have time to review this?

tools/runners/test262-harness.py
```diff
@@ -518,6 +574,8 @@ def get_negative_type(self):
 
     def get_negative_phase(self):
         negative = self.get_negative()
+        if self.suite.es5_1:
+            return negative
```
akosthekiss, line 578: Cannot be the negative record returned by the ES5.1 parser be made more alike to that of ES.next? This would allow avoiding special casing like this.

akosthekiss (changes requested):

tools/runners/test262-harness.py
```diff
@@ -324,6 +324,53 @@ def parse_test_record(src, name, onerror=print):
...
+
+    # Should match anything
+    return re.compile(r"^(" + header_pattern_str +
+                      r")(?:" + capture_comment_pattern_str +
+                      r")?(" + any_pattern_str +
+                      r")$")
+
+
+_AT_ATTRS_ES5_1 = re.compile(r"\s*\n\s*\*\s*@")
+_RECORD_PATTERN_ES5_1 = record_pattern_es5_1_create()
```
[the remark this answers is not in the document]
lygstate: it's directly copy from es5.1 test262.py, not sure change it here is a good idea.

tools/runners/test262-harness.py
```diff
@@ -490,7 +541,12 @@ def __init__(self, suite, name, full_path, strict_mode, command_template, module
         self.strict_mode = strict_mode
         with open(self.full_path) as file_desc:
             self.contents = file_desc.read()
-        test_record = parse_test_record(self.contents, name)
+        if self.suite.es5_1:
+            test_record = parse_test_record_es5_1(self.contents, name)
+            if 'negative' in test_record:
+                test_record['negative'] = None
```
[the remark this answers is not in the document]
lygstate: > Why does `test_record` need a post-processing after `parse_test_record_es5_1`. Why does the function not deal with negative already?

Because in es5_1,  parse_test_record_es5_1 is a direct copy of test_262.py , I dindn't change the sematic,

This pos-processing is used to compatibility with test262-harness.py

tools/runners/test262-harness.py
lygstate: Yeap, looking for suggestion, maybe have better way to solve this issue

tools/runners/test262-harness.py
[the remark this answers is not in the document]
akosthekiss: If you copy it here then it's "our" code from that point on. We cannot and will not make any guarantees to keep this in sync with the original source, that would be a maintenance burden.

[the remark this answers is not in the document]
lygstate: You are right, I am not capable to change the regex, thanks for your suggestion, I've updated with you exp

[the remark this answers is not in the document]
ossy-szeged: > If you copy it here then it's "our" code from that point on. We cannot and will not make any guarantees to keep this in sync with the original source, that would be a maintenance burden.

There is no maintenance burden, ES5.1 version of test262 (and its harness too) is a fixed git hash  from 2015 (tag/branch: es5-tests), nobody maintains it and nobody will change it ever.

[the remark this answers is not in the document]
lygstate: > > If you copy it here then it's "our" code from that point on. We cannot and will not make any guarantees to keep this in sync with the original source, that would be a maintenance burden.
> 
> There is no maintenance burden, ES5.1 version of test262 (and its harness too) is a fixed git hash from 2015 (tag/branch: es5-tests), nobody maintains it and nobody will change it ever.

maybe @akosthekiss are talking about test262-harness.py maintain burden, I've updated with his suggestion

rerobika (approved):

lygstate:
ping for merge

tools/runners/test262-harness.py
lygstate: fixed

lygstate:
can this be merged now?

PR #4390, opened 2020-12-30
Revise tools scripts to be python3 compatible

lygstate:
Revise tools scripts to be python3 compatible
We introduce setup_stdio function to setup stdout/stderr properly.
For python <-> python pipe, we always use 'utf8'/'ignore' encoding for not lost
characters.
For tty <-> python, we using native encoding with xmlcharrefreplace to encode, to
preserve maximal information.
For python <-> native program, we use naive encoding with 'ignore' to not cause error

For reading/writing to file, we use 'utf8' /'ignore' encoding for not lost
characters.

### DIFF
--- a/tools/runners/test262-harness.py
+++ b/tools/runners/test262-harness.py
@@ -42,6 +42,7 @@
 
 from __future__ import print_function
 
+import codecs
 import logging
 import optparse
 import os
@@ -57,6 +58,8 @@ from collections import Counter
 import signal
 import threading
 import multiprocessing
+
+import util
 
 #######################################################################
 # based on _monkeyYaml.py
@@ -402,13 +405,13 @@ class TempFile(object):
             text=self.text)
 
     def write(self, string):
-        os.write(self.file_desc, string)
+        os.write(self.file_desc, string.encode("utf8", "ignore"))
 
     def read(self):
-        file_desc = file(self.name)
+        file_desc = open(self.name, "rb")
         result = file_desc.read()
         file_desc.close()
-        return result
+        return result.decode("utf8", "ignore")
 
     def close(self):
         if not self.is_closed:
@@ -460,12 +463,12 @@ class TestResult(object):
     def write_output(self, target):
         out = self.stdout.strip()
         if out:
-            target.write("--- output --- \n %s" % out)
+            target.write(u"--- output --- \n %s" % out)
         error = self.stderr.strip()
         if error:
-            target.write("--- errors ---  \n %s" % error)
+            target.write(u"--- errors ---  \n %s" % error)
 
-        target.write("\n--- exit code: %d ---\n" % self.exit_code)
+        target.write(u"\n--- exit code: %d ---\n" % self.exit_code)
 
     def has_failed(self):
         return self.exit_code != 0
@@ -495,8 +498,8 @@ class TestCase(object):
         self.name = name
         self.full_path = full_path
         self.strict_mode = strict_mode
-        with open(self.full_path) as file_desc:
-            self.contents = file_desc.read()
+        with open(self.full_path, "rb") as file_desc:
+            self.contents = file_desc.read().decode("utf8", "ignore")
         test_record = parse_test_record(self.contents, name)
         self.test = test_record["test"]
         del test_record["test"]
@@ -645,7 +648,7 @@ class TestCase(object):
         return TestResult(code, out, err, self)
 
     def run(self):
-        tmp = TempFile(suffix=".js", prefix="test262-", text=True)
+        tmp = TempFile(suffix=".js", prefix="test262-")
         try:
             result = self.run_test_in(tmp)
         finally:
@@ -762,8 +765,8 @@ class TestSuite(object):
         if not name in self.include_cache:
             static = path.join(self.lib_root, name)
             if path.exists(static):
-                with open(static) as file_desc:
-                    contents = file_desc.read()
+                with open(static, "rb") as file_desc:
+                    contents = file_desc.read().decode("utf8", "ignore")
                     contents = re.sub(r'\r\n', '\n', contents)
                     self.include_cache[name] = contents + "\n"
             else:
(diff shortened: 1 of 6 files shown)


LaszloLango (dismissed):
LGTM

tools/runners/test262-harness.py
```diff
@@ -392,16 +392,16 @@ def open_file(self):
...
-        os.write(self.file_desc, string)
+        os.write(self.file_desc, string.encode("utf8"))
 
     def read(self):
-        file_desc = file(self.name)
+        file_desc = open(self.name, "rb")
         result = file_desc.read()
         file_desc.close()
-        return result
+        return result.decode("utf8", 'ignore')
```
akosthekiss, line 404: Don't mix single and double quotes. Preferably, use consistent style throughout the file. But at the very least, don't mix them in the same line.

akosthekiss: Some `decode`s use ignore, some don't. Why?

tools/runners/test262-harness.py
```diff
@@ -927,6 +926,7 @@ def main():
 
 if __name__ == '__main__':
     try:
+        sys.stdout = io.TextIOWrapper(sys.stdout.buffer, encoding=sys.stdout.encoding, errors="ignore")
```
akosthekiss, line 929: Why does `stdout` need wrapping?

ossy-szeged:
Do we need it to be able to use our harness for ES5.1 version of test262? If yes, it would be great to mention it in the commit log at least.

tools/runners/test262-harness.py
lygstate: only raised error on my gbk environment use ignore

tools/runners/test262-harness.py
lygstate: because test262 cases have character can not represent in gbk and its print to console

lygstate: > Why does `stdout` need wrapping?

removed, fixed in another pr

tools/runners/test262-harness.py
lygstate: > Some `decode`s use ignore, some don't. Why?

done, all to be consistence

tools/runners/test262-harness.py
```diff
@@ -830,7 +830,7 @@ def run(self, command_template, tests, print_summary, full_summary, logname, job
             report_error("No tests to run")
         progress = ProgressIndicator(len(cases))
         if logname:
-            self.logf = open(logname, "w")
+            self.logf = open(logname, "w", encoding="utf8", errors="ignore")
```
akosthekiss, line 833: I don't think that this will work on py2. The `open` builtin of py2 has neither `encoding` nor `errors`. However, the `codecs` builtin module is awailable both in py2 and py3, and might be used to open files with a given encoding.

See:
- https://docs.python.org/2.7/library/codecs.html#codecs.open
- https://docs.python.org/3.9/library/codecs.html#codecs.open

This approach (i.e., `codecs.open`) might be useful everywhere where files are opened, making `.decode()` calls unnecessary.

akosthekiss (changes requested):

tools/runners/test262-harness.py
lygstate: OK, does

tools/runners/test262-harness.py
```diff
@@ -910,6 +911,13 @@ def list_includes(self, tests):
 
 
 def main():
+    (out_stream, err_stream) = (sys.stdout.buffer, sys.stderr.buffer) \
+                                if sys.version_info.major >= 3 else (sys.stdout, sys.stderr)
+    # For tty using native encoding, otherwise (pipe) use 'utf-8'
```
lygstate, line 916: @akosthekiss how about move these code into util， they can be shared across different tools

akosthekiss:
During the review of the current version of the PR, I've found a potentially useful (and definitely interesting) approach to the problem, which may reduce the necessary code changes considerably. With just two lines of code (per tool), the unicode encode/decode errors can be disabled/hidden/ignored:
```py
import codecs
codecs.register_error("strict", codecs.ignore_errors)
```
(source: https://stackoverflow.com/questions/9537356/can-i-make-decodeerrors-ignore-the-default-for-all-strings-in-a-python-2-7-p)

This would be worth a try. Then, no stdour/err would be necessary, no polluting of the code with encodes/decodes, no `codecs.open` instead of `open`, etc. (Hopefully.)

akosthekiss (changes requested):
Please, edit the commit message. It is unacceptably long. Tracebacks and diagnostics and console logs should not go into a commit message. They belong to the discussion comments of the corresponding PR.

lygstate:
> During the review of the current version of the PR, I've found a potentially useful (and definitely interesting) approach to the problem, which may reduce the necessary code changes considerably. With just two lines of code (per tool), the unicode encode/decode errors can be disabled/hidden/ignored:
> 
> ```python
> import codecs
> codecs.register_error("strict", codecs.ignore_errors)
> ```
> 
> (source: https://stackoverflow.com/questions/9537356/can-i-make-decodeerrors-ignore-the-default-for-all-strings-in-a-python-2-7-p)
> 
> This would be worth a try. Then, no stdour/err would be necessary, no polluting of the code with encodes/decodes, no `codecs.open` instead of `open`, etc. (Hopefully.)

Yeap, I know this method, but the following are still better.
```
def setup_stdio():
    (out_stream, err_stream) = (sys.stdout, sys.stderr)
    if sys.version_info.major >= 3:
        (out_stream, err_stream) = (sys.stdout.buffer, sys.stderr.buffer)
    # For tty using native encoding, otherwise (pipe) use 'utf-8'
    encoding = sys.stdout.encoding if sys.stdout.isatty() else 'utf-8'
    # Always override it to anvoid encode error
    sys.stdout = codecs.getwriter(encoding)(out_stream, 'ignore')
    sys.stderr = codecs.getwriter(encoding)(err_stream, 'ignore')
```
codec ignore is just hidden the error when output isatty, but when we do piping, we always using the 'utf8-8' encoding.
Otherwise the characters will lost in tests script :(
That's why setup_stdio are important, not just simple ignore it

lygstate:
@ossy-szeged  any progress of the PR? after this, I can create PR that support python3 debugger

PR #4391, opened 2020-12-30
Make es5.1 option consistence with es2015 and esnext

lygstate:
When testing jerryscript on new platform, there maybe es5.1 test failure to tracking, so enable es5.1 excludelist option.


ossy-szeged:
It can be useful once we merged #4389. (the old harness doesn't support excludelist at all).
If we have ES5.1 excludelist, we should remove this hack too: 
https://github.com/jerryscript-project/jerryscript/blob/master/tools/runners/run-test-suite-test262.py#L91

lygstate:
Close in favor of https://github.com/jerryscript-project/jerryscript/pull/4389

PR #4392, opened 2021-01-01
Fixes ENABLE_AMALGAM need FORCE set to ON when building with MSVC

lygstate:
Fixes ENABLE_AMALGAM need FORCE set to ON when building with MSVC

Using following regexp to rename JERRY_ATTR_ALWAYS_INLINE to JERRY_ALWAYS_INLINE:

" inline ((const |)[a-zA-Z0-9_]+(\s+\*)?)(\s+JERRY_ATTR_CONST|\s+JERRY_ATTR_PURE|\s+JERRY_ATTR_HOT)?\s*JERRY_ATTR_ALWAYS_INLINE"
" JERRY_ALWAYS_INLINE \1\4"

So that we can implement always inline function on different compiler properly;
especially bring up the gap between MSVC and GNU
The always inline for GNU and MSVC are different:
GNU
```
#define JERRY_ALWAYS_INLINE inline __attribute__((always_inline))
```
MSVC
```
#define JERRY_ALWAYS_INLINE __forceinline
```

### DIFF
--- a/jerry-core/ecma/base/ecma-alloc.c
+++ b/jerry-core/ecma/base/ecma-alloc.c
@@ -71,7 +71,7 @@ ecma_dealloc_number (ecma_number_t *number_p) /**< number to be freed */
  *
  * @return pointer to allocated memory
  */
-inline ecma_object_t * JERRY_ATTR_ALWAYS_INLINE
+extern inline ecma_object_t * JERRY_ATTR_ALWAYS_INLINE
 ecma_alloc_object (void)
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -84,7 +84,7 @@ ecma_alloc_object (void)
 /**
  * Dealloc memory from an ecma-object
  */
-inline void JERRY_ATTR_ALWAYS_INLINE
+extern inline void JERRY_ATTR_ALWAYS_INLINE
 ecma_dealloc_object (ecma_object_t *object_p) /**< object to be freed */
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -99,7 +99,7 @@ ecma_dealloc_object (ecma_object_t *object_p) /**< object to be freed */
  *
  * @return pointer to allocated memory
  */
-inline ecma_extended_object_t * JERRY_ATTR_ALWAYS_INLINE
+extern inline ecma_extended_object_t * JERRY_ATTR_ALWAYS_INLINE
 ecma_alloc_extended_object (size_t size) /**< size of object */
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -112,7 +112,7 @@ ecma_alloc_extended_object (size_t size) /**< size of object */
 /**
  * Dealloc memory of an extended object
  */
-inline void JERRY_ATTR_ALWAYS_INLINE
+extern inline void JERRY_ATTR_ALWAYS_INLINE
 ecma_dealloc_extended_object (ecma_object_t *object_p, /**< extended object */
                               size_t size) /**< size of object */
 {
@@ -128,7 +128,7 @@ ecma_dealloc_extended_object (ecma_object_t *object_p, /**< extended object */
  *
  * @return pointer to allocated memory
  */
-inline ecma_string_t * JERRY_ATTR_ALWAYS_INLINE
+extern inline ecma_string_t * JERRY_ATTR_ALWAYS_INLINE
 ecma_alloc_string (void)
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -141,7 +141,7 @@ ecma_alloc_string (void)
 /**
  * Dealloc memory from ecma-string descriptor
  */
-inline void JERRY_ATTR_ALWAYS_INLINE
+extern inline void JERRY_ATTR_ALWAYS_INLINE
 ecma_dealloc_string (ecma_string_t *string_p) /**< string to be freed */
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -156,7 +156,7 @@ ecma_dealloc_string (ecma_string_t *string_p) /**< string to be freed */
  *
  * @return pointer to allocated memory
  */
-inline ecma_extended_string_t * JERRY_ATTR_ALWAYS_INLINE
+extern inline ecma_extended_string_t * JERRY_ATTR_ALWAYS_INLINE
 ecma_alloc_extended_string (void)
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -169,7 +169,7 @@ ecma_alloc_extended_string (void)
 /**
  * Dealloc memory from extended ecma-string descriptor
  */
-inline void JERRY_ATTR_ALWAYS_INLINE
+extern inline void JERRY_ATTR_ALWAYS_INLINE
 ecma_dealloc_extended_string (ecma_extended_string_t *ext_string_p) /**< extended string to be freed */
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -184,7 +184,7 @@ ecma_dealloc_extended_string (ecma_extended_string_t *ext_string_p) /**< extende
  *
  * @return pointer to allocated memory
  */
-inline ecma_external_string_t * JERRY_ATTR_ALWAYS_INLINE
+extern inline ecma_external_string_t * JERRY_ATTR_ALWAYS_INLINE
 ecma_alloc_external_string (void)
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -197,7 +197,7 @@ ecma_alloc_external_string (void)
 /**
  * Dealloc memory from external ecma-string descriptor
  */
-inline void JERRY_ATTR_ALWAYS_INLINE
+extern inline void JERRY_ATTR_ALWAYS_INLINE
 ecma_dealloc_external_string (ecma_external_string_t *ext_string_p) /**< external string to be freed */
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -212,7 +212,7 @@ ecma_dealloc_external_string (ecma_external_string_t *ext_string_p) /**< externa
  *
  * @return pointer to allocated memory
  */
-inline ecma_string_t * JERRY_ATTR_ALWAYS_INLINE
+extern inline ecma_string_t * JERRY_ATTR_ALWAYS_INLINE
 ecma_alloc_string_buffer (size_t size) /**< size of string */
 {
 #if ENABLED (JERRY_MEM_STATS)
@@ -225,7 +225,7 @@ ecma_alloc_string_buffer (size_t size) /**< size of string */
 /**
  * Dealloc memory of a string with character data
  */
-inline void JERRY_ATTR_ALWAYS_INLINE
+extern inline void JERRY_ATTR_ALWAYS_INLINE
 ecma_dealloc_string_buffer (ecma_string_t *string_p, /**< string with data */
                             size_t size) /**< size of string */
 {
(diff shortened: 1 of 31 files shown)


akosthekiss:
What's happening here and more importantly, why? Why is code moved from sources to headers. Generally, this does not sound like the good direction. The PR comment is longish, but does not explain the idea behind the change.

What's more, it contains the PR template. Again. The template, that has a bold prologue and epilogue highlighting that it should be removed. @lygstate Please, play by the rules.

rerobika:
Honest request. Before one more force push, please describe the intuition of this PR.

lygstate:
@rerobika @akosthekiss  any suggestion about
GNU
```
#define JERRY_ALWAYS_INLINE(x) extern inline x __attribute__((always_inline))
#define JERRY_ALWAYS_INLINE_STATIC(x) static inline x __attribute__((always_inline))
```
MSVC
```
#define JERRY_ALWAYS_INLINE(x) extern __forceinline x
#define JERRY_ALWAYS_INLINE_STATIC(x) __forceinline x
```
Also have a alternative:

GNU
```
#define JERRY_ALWAYS_INLINE(x) inline x __attribute__((always_inline))
```
MSVC
```
#define JERRY_ALWAYS_INLINE(x) __forceinline x
```
And for static and extern use
``
extern JERRY_ALWAYS_INLINE(x)
static JERRY_ALWAYS_INLINE(x)
``
instead.
Which on prefer?

galpeter:
I've tried the following code with gcc and clang:
```c
#include <stdio.h>

static inline void __attribute__((always_inline)) method(void) {
  printf("HI\n");
}

static inline __attribute__((always_inline)) void method2(void) {
  printf("HI\n");
}

static void method3(void) {
  printf("HI\n");
}

int main() {
  method();
  method2();
  method3();
  return 0;
}
```

Note the `__attribute__` placement.

Tried with `-O0` `method` and `method2` were inlined. So the question is that: do we really need the argument for this macro?
What if the macro is defined as:
```c
#ifdef __GNUC__
#define JERRY_ALWAYS_INLINE inline __attribute__((always_inline))
#endif

#ifdef _MSC_VER
#define JERRY_ALWAYS_INLINE __forceinline
#endif
```

and used as:
```c
extern JERRY_ALWAYS_INLINE bool
jcontext_has_pending_abort (void)
```
That is moving the macro (`JERRY_ALWAYS_INLINE`) before the return type (which we obviously must document somewhere that this is the preferred style).

What do you think? Would this work on Windows also?

rerobika:
If it works I'd second @galpeter's suggestion.

lygstate:
> If it works I'd second @galpeter's suggestion. @lygstate Please check it.

Let's me have an try

lygstate:
@galpeter Good suggestion, it's worked, @rerobika Please take a look

jerry-core/ecma/base/ecma-alloc.c
```diff
@@ -128,7 +128,7 @@ ecma_dealloc_extended_object (ecma_object_t *object_p, /**< extended object */
  *
  * @return pointer to allocated memory
  */
-inline ecma_string_t * JERRY_ATTR_ALWAYS_INLINE
+extern JERRY_ALWAYS_INLINE ecma_string_t *
```
rerobika, line 131: Could you explain me why the `extern` keyword is always required in C files as well?

lygstate: > Could you explain me why the `extern` keyword is always required in C files as well?

Because in msvc, inline without extern would  cause the function don't generate symbol and then cause linkage error.

jerry-core/include/jerryscript-compiler.h
```diff
@@ -41,6 +40,7 @@ extern "C"
 #define JERRY_ATTR_PURE __attribute__((pure))
 #define JERRY_ATTR_WARN_UNUSED_RESULT __attribute__((warn_unused_result))
 
+#define JERRY_ALWAYS_INLINE inline __attribute__((always_inline))
```
rerobika, line 43: Please add a note about the usage of this macro: 
 - before it's usage `static` or `extern` modifier must be used.

rerobika (commented):
Look good to me after this note.

jerry-core/include/jerryscript-compiler.h
lygstate: done in latter line

jerry-core/include/jerryscript-compiler.h
```diff
@@ -144,6 +139,16 @@ void * __cdecl _alloca (size_t _Size);
 #define JERRY_ATTR_WARN_UNUSED_RESULT
 #endif /* !JERRY_ATTR_WARN_UNUSED_RESULT */
 
+/**
+ * Macro to setup function attributes to force inline function to all call sites.
+ * Before it's usage `static` or `extern` modifier must be used.
+ * Because in msvc, inline without `extern` would cause the function don't generate
+ * symbol and then cause linkage error.
```
lygstate, line 146: @rerobika The comment are here

rerobika (approved):
LGTM

lygstate:
@akosthekiss @LaszloLango Please take a look at this pull request.

LaszloLango:
@lygstate I think no need for me here. This PR has two reviewers already. One of them approved it and there is an ongoing review from @galpeter. BTW I can take a look on it if Peter cannot, but I think he will do it soon. :)

lygstate:
@galpeter Ping for review

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -166,7 +166,7 @@ ecma_create_object_lex_env (ecma_object_t *outer_lexical_environment_p, /**< out
  * @return true  - if object is a lexical environment
  *         false - otherwise
  */
-inline bool JERRY_ATTR_PURE
+extern JERRY_ALWAYS_INLINE bool JERRY_ATTR_PURE
```
akosthekiss, line 169: This is not an identical change. This function is simply `inline`, not "always inline". (Whether that's right, that can be discussed ofc.)

(This holds for several functions in this module.)

jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -72,7 +72,7 @@ enum
  * @return ecma value of the new promise.
  *         Returned value must be freed with ecma_free_value.
  */
-inline static ecma_value_t
+static JERRY_ALWAYS_INLINE ecma_value_t
```
akosthekiss, line 75: same in this module

jerry-core/ecma/operations/ecma-container-object.c
```diff
@@ -44,7 +44,7 @@
  *
  * @return pointer to the internal buffer
  */
-static inline ecma_collection_t *
+static JERRY_ALWAYS_INLINE ecma_collection_t *
```
akosthekiss, line 47: ditto (I stop repeating this comment)

jerry-core/include/jerryscript-compiler.h
```diff
@@ -144,6 +139,16 @@ void * __cdecl _alloca (size_t _Size);
...
+ */
+#ifndef JERRY_ALWAYS_INLINE
+#define JERRY_ALWAYS_INLINE(x) inline
```
akosthekiss, line 149: This is not a function-like macro anymore.

akosthekiss (changes requested):

jerry-core/ecma/base/ecma-helpers.c
lygstate: inline have same effect with __forceinline in MSVC, so the JERRY_ALWAYS_INLINE macro are needed to make sure use always prefix JERRY_ALWAYS_INLINE with static or extern.
And if the function have no need inline, we can always not use?
What's the choose between inline and always_inline?

jerry-core/include/jerryscript-compiler.h
lygstate: OK

jerry-core/ecma/base/ecma-helpers.c
akosthekiss: I'm happy to call in other maintainers of the project to discuss why some functions are simply inline while others are always inline. However, if we cannot dig to the roots of that design decision now - which can easily happen - then we should not introduce non-identical changes with potential unknown side-effects just because of MSVC.

If I get it correctly, we could
- simply keep the `inline` keyword everywhere,
- add `extern` keyword to non-static functions where that was missing (I'm OK with that addition to make MSVC happy),
- keep `JERRY_ATTR_ALWAYS_INLINE` macro name, but change its placement in the function signature, and
- define `JERRY_ATTR_ALWAYS_INLINE` for MSVC as `__forceinline`.

And this would be the least disruptive approach then. Would that work?

lygstate: make sense, done with your suggestion

jerry-core/api/jerry-snapshot.c
```diff
@@ -34,7 +34,7 @@
  *
  * @return configuration flags
  */
-static inline uint32_t JERRY_ATTR_ALWAYS_INLINE
+static JERRY_ATTR_ALWAYS_INLINE uint32_t
```
akosthekiss, line 37: That's not exactly as suggested. I've suggested `static inline JERRY_ATTR_ALWAYS_INLINE uint32_t`.

jerry-core/include/jerryscript-compiler.h
```diff
@@ -31,7 +31,7 @@ extern "C"
  * Compiler-specific macros relevant for GCC.
  */
 #define JERRY_ATTR_ALIGNED(ALIGNMENT) __attribute__((aligned(ALIGNMENT)))
-#define JERRY_ATTR_ALWAYS_INLINE __attribute__((always_inline))
+#define JERRY_ATTR_ALWAYS_INLINE inline __attribute__((always_inline))
```
akosthekiss, line 34: That's not exactly as suggested. I've suggested to keep the GCC version of `JERRY_ATTR_ALWAYS_INLINE` unchanged.

lygstate: then read the code carefully, you are suggesting invalid.things

jerry-core/api/jerry-snapshot.c
lygstate: invalid

lygstate:
@galpeter Ping for review

jerry-core/include/jerryscript-compiler.h
lygstate: > That's not exactly as suggested. I've suggested to keep the GCC version of `JERRY_ATTR_ALWAYS_INLINE` unchanged.

inline and __forceinline can not co-exist.  read the title first.

akosthekiss: charming

akosthekiss: Frankly, this is a C99 project primarily targeting resource-constrained embedded systems. Windows is not a resource-constrained platform and more importantly, MSVC is not C99 conforming. I'm not really convinced to replace even the standard C `inline` keyword for the sake of MSVC.

This PR aims at removing the need for amalgamated builds in case of MSVC. Then it should do that, and only that. According to what has been revealed so far, the only thing necessary to achieve that is to add `extern` in front of `inline` in same cases. No new macros, no macro re-purposing,  no macro placement reordering.

If `__forceinline` is not a modifier, like `__attribute__(())` or `__declspec()`, but a replacement for the `inline` keyword then we should not consider adding it to the mix, at least for now. Then `JERRY_ATTR_ALWAYS_INLINE` will remain empty for MSVC. Should not be a big deal, the project will compile even if some functions are "simple" inlines, not forced ones.

lygstate: @akosthekiss done

akosthekiss (approved):
LGTM

PR #4410, opened 2021-01-05
Rename jerry-libm to jerry-math

akosthekiss:
That "libm" in the name of the library resulted in awkward naming
on *nix systems (`libjerry-libm.*`, "lib" occurring twice). And the
name of the corresponding header is `math.h` anyway.

Note that this is a breaking change in some sense. The commit
contains no API change, but the build system does change for users
of the math library.


rerobika (commented):
I found 2 remaining `libm` references.
https://github.com/akosthekiss/jerryscript/blob/libjerry-math/CMakeLists.txt#L259
https://github.com/akosthekiss/jerryscript/blob/libjerry-math/jerry-math/CMakeLists.txt#L33

Do you want to keep them consciously or are these leftovers of the renaming process?

akosthekiss:
@rerobika Yes, they are intentional. I thought it would be OK to signal somehow that jerry-math is a libm drop-in replacement.

rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #4414, opened 2021-01-05
Fix the include directory in the pkgconfig of the math lib

akosthekiss:
The pkgconfig specified `include/jerry-math` as the include
directory, but cmake installs header to `include/jerryscript-math`.


rerobika (approved):
LGTM

dbatyai (approved):
LGTM

Issue #4399, opened 2021-01-02
Assertion 'uint_ptr % JMEM_ALIGNMENT == 0' in jmem_decompress_pointer


owl337:
###### JerryScript revision
2faafa4

###### Build platform
Ubuntu 18.04.5 LTS(Linux 4.15.0-119-generic x86_64)

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g --strip=off \
--system-allocator=on --logging=on --linker-flag=-fuse-ld=gold \
--error-messages=on --profile=es2015-subset  --builddir=$PWD/build
```

###### Test case
```

var obj0 = {};
var obj1 = {};
var obj2 = [];
var obj3 = new Proxy(Function(), { get: function(o, k) { get.push(k); return o[k]; }});
var obj4 = new Error();
var obj5 = JSON;

var values = [
    [ obj0, 0 ],
    [ obj1, 1 ],
    [ obj2, 2 ],
    [ obj3, 3 ],
    [ obj4, 4 ],
    [ obj5, 5 ],
    [ obj4, 4 ],
    [ obj3, 3 ],
    [ obj2, 2 ],
    [ obj1, 1 ],
    [ obj0, 0 ],
];
var map = new WeakMap(values);
```

###### Output
```
ICE: Assertion 'uint_ptr % JMEM_ALIGNMENT == 0' failed at /root/jerryscript/jerry-core/jmem/jmem-allocator.c(jmem_decompress_pointer):222.
Error: ERR_FAILED_INTERNAL_ASSERTION
Aborted
```

Credits: Found by chong from OWL337.

Issue #4396, opened 2021-01-02
Assertion '(object_p->u1.property_list_cp & ~JMEM_TAG_MASK) == 0' in ecma_gc_free_object


owl337:
###### JerryScript revision
2faafa4

###### Build platform
Ubuntu 18.04.5 LTS(Linux 4.15.0-119-generic x86_64)

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g --strip=off \
--system-allocator=on --logging=on --linker-flag=-fuse-ld=gold \
--error-messages=on --profile=es2015-subset  --builddir=$PWD/build

```

###### Test case
```
var obj1 = new Proxy({length: 2, 0: '', 1: ''}, { get: function(o, k) { get.push(k); return o[k]; }});
var weakset = new WeakSet();
weakset.add(obj1);

```

###### Output
```
ICE: Assertion '(object_p->u1.property_list_cp & ~JMEM_TAG_MASK) == 0' failed at /root/jerryscript/jerry-core/ecma/base/ecma-gc.c(ecma_gc_free_object):1655.
Error: ERR_FAILED_INTERNAL_ASSERTION
Aborted

```

Credits: Found by chong from OWL337.

PR #4415, opened 2021-01-05
Don't use property list pointer to store Proxy flags

galpeter:
Proxy flags (IsCallable, IsConstructor) can't be stored on the
property list compressed pointer. As adding a Proxy to a WeakSet
would add a property to the Proxy object causing failures down the line.

The prototype internal "slot" can be used to store there flags as
it is not used in case of Proxies (as per standard).

Fixes: #4396
Fixes: #4399

### DIFF
--- a/jerry-core/ecma/base/ecma-gc.c
+++ b/jerry-core/ecma/base/ecma-gc.c
@@ -685,6 +685,12 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
     }
 #endif /* ENABLED (JERRY_BUILTIN_REALMS) */
 
+    /**
+     * Have the object's prototype here so the object could set it to JMEM_CP_NULL
+     * if the prototype should be ignored (like in case of PROXY).
+     */
+    jmem_cpointer_t proto_cp = object_p->u2.prototype_cp;
+
     switch (object_type)
     {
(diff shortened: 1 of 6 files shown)


jerry-core/ecma/base/ecma-gc.c
```diff
@@ -852,11 +858,11 @@ ecma_gc_mark (ecma_object_t *object_p) /**< object to mark from */
             }
           }
 
-          jmem_cpointer_t proto_cp = object_p->u2.prototype_cp;
+          jmem_cpointer_t array_proto_cp = object_p->u2.prototype_cp;
 
-          if (proto_cp != JMEM_CP_NULL)
+          if (array_proto_cp != JMEM_CP_NULL)
```
rerobika, line 863: I think we dont need that local variable. `array_proto_cp` is eqivalent to `proto_cp ` above.

rerobika (changes requested):

jerry-core/ecma/base/ecma-gc.c
galpeter: True. PR updated

rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #4417, opened 2021-01-05
Enable generating compile_commands.json

rerobika:
This feature is useful for intellisense users.

Signed-off-by: Robert Fancsik <frobert@inf.u-szeged.hu>

### DIFF
--- a/jerry-main/CMakeLists.txt
+++ b/jerry-main/CMakeLists.txt
@@ -16,11 +16,13 @@ cmake_minimum_required (VERSION 2.8.12)
 project (jerry-main C)
 
 # Optional build settings
-set(ENABLE_LINK_MAP OFF CACHE BOOL "Enable generating a link map file?")
-set(JERRY_TEST_STACK_MEASURE OFF CACHE BOOL "Enable stack measurement for the jerry-test binary?")
+set(ENABLE_LINK_MAP           OFF CACHE BOOL "Enable generating a link map file?")
+set(ENABLE_COMPILE_COMMANDS   ON  CACHE BOOL "Enable generating compile_commands.json?")
+set(JERRY_TEST_STACK_MEASURE  OFF CACHE BOOL "Enable stack measurement for the jerry-test binary?")
 
 # Status messages
 message(STATUS "ENABLE_LINK_MAP                " ${ENABLE_LINK_MAP})
+message(STATUS "ENABLE_COMPILE_COMMANDS        " ${ENABLE_COMPILE_COMMANDS})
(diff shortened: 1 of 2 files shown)


akosthekiss (approved):
LGTM

CMakeLists.txt
```diff
@@ -148,6 +148,9 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/bin/")
 set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib/")
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/lib/")
 
+# Enable generating compile_commands.json
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
```
lygstate, line 152: Prefer option instead set.
For building procedure, there is no need generate it

rerobika: There is no disadvantage of generating it, and it's useful for some IDEs and intellisense.

rerobika: This reasoning would more convince me if the recommended build platform would be the low-end device. However, most of the time the (cross)-build is performed on the host device where this extra json doesn't add to much for the "pollution" what cmake generates.

akosthekiss: @lygstate What exactly do you mean by pollution? The file is generated into the build tree which is already full of cmake-generated auxiliary files.

lygstate: > @lygstate What exactly do you mean by pollution? The file is generated into the build tree which is already full of cmake-generated auxiliary files.

The parent project also generating compile_commands.json

akosthekiss: @lygstate Is that a real issue anywhere?

@rerobika How about finding a middle ground and adding a cli option to build.py that controls this cmake variable?

rerobika: I can deal with that if it is enabled by default.

lygstate: option(CMAKE_EXPORT_COMPILE_COMMANDS "Enable export compile commands for jerryscript" ON)

akosthekiss: @lygstate Thanks, we _do_ know how to add an option to cmake. The suggestion was to add an option to `build.py` (we know how to add that, too). What we do not know whether the "pollution" is a real issue in any embedding project?

lygstate:
LGTM

dbatyai (approved):
LGTM

PR #4424, opened 2021-01-05
Fix unavailable source file handling in main_print_unhandled_exception

### LINKED ISSUE #4403: SEGV in main_print_unhandled_exception
###### JerryScript revision
2faafa4
###### Build platform
Ubuntu 18.04.5 LTS(Linux 4.15.0-119-generic x86_64)
###### Build steps
```
--compile-flag=-fno-common --compile-flag=-g --strip=off \
--system-allocator=on --logging=on --linker-flag=-fuse-ld=gold \
--error-messages=on --profile=es2015-subset  --builddir=$PWD/build
```

###### Test case
```
throw new SyntaxError("[0: 98, 1: 54, 8: false, 10: 21, foo: bar]");
```
###### Output
```
Error: Failed to open file: 0
ASAN:DEADLYSIGNAL
=================================================================
==13994==ERROR: AddressSanitizer: SEGV on unknown address 0x00000000 (pc 0x5678dd4b bp 0xffca4f88 sp 0xffca4d10 T0)
==13994==The signal is caused by a READ memory access.
==13994==Hint: address points to the zero page.
    #0 0x5678dd4a in main_print_unhandled_exception /root/jerryscript/jerry-main/main-utils.c:341
    #1 0x5678baae in main /root/jerryscript/jerry-main/main-unix.c:140
    #2 0xf7760f20 in __libc_start_main (/lib32/libc.so.6+0x18f20)
    #3 0x5662c3d0  (/root/jerryscript/build/bin/jerry+0x1d3d0)

AddressSanitizer can not provide additional info.
SUMMARY: AddressSanitizer: SEGV /root/jerryscript/jerry-main/main-utils.c:341 in main_print_unhandled_exception
==13994==ABORTING
```
Credits: Found by chong from OWL337.

rerobika:
This patch fixes #4403.

### DIFF
--- a/jerry-main/main-utils.c
+++ b/jerry-main/main-utils.c
@@ -323,9 +323,6 @@ main_print_unhandled_exception (jerry_value_t error_value) /**< error value */
 
       if (err_line != 0 && err_col > 0 && err_col < SYNTAX_ERROR_MAX_LINE_LENGTH)
       {
-        uint32_t curr_line = 1;
-        uint32_t pos = 0;
-
         /* Temporarily modify the error message, so we can use the path. */
         *path_str_end_p = '\0';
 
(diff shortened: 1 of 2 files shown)


jerry-main/main-utils.c
```diff
@@ -332,6 +332,13 @@ main_print_unhandled_exception (jerry_value_t error_value) /**< error value */
         size_t source_size;
         uint8_t *source_p = jerry_port_read_source (path_str_p, &source_size);
 
+        if (source_p == NULL)
+        {
+          jerry_release_value (err_str_val);
+          jerry_release_value (error_value);
+          return;
```
akosthekiss, line 339: Isn't this a way too early return? It seems to me that it is only this block (roughly up until `jerry_port_log (JERRY_LOG_LEVEL_ERROR, "^\n\n");`) that's depending on `source_p`.

`jerry_port_log (JERRY_LOG_LEVEL_ERROR, "%s\n", err_str_buf);` and whatever is following should happen whether `source_p` is NULL or not, should it not?

If so, then "2. seek and print" and the rest should be put in an `if (source_p != NULL)` instead.

BTW, the check should happen _after_ `*path_str_end_p = ':';` IMO.

rerobika: I wouldn't say that's too early. I think if we cannot open the file we want to read from executing all the upcoming operations are unnecessary. The `jerry_port_read_source ` logs the unsuccesful operation for the user, I think that's enough.

akosthekiss: According to the docs of the API of `jerry_port_read_source`:

> This port function is called by jerry-core when JERRY_MODULE_SYSTEM
> is enabled. The path is specified in the import statement's 'from "..."'
> section.

So, it's a bit "shaky" whether we can rely on `jerry_port_read_source` if `JERRY_MODULE_SYSTEM` is disabled.

rerobika: Not exactly. The existence and the functionality of `jerry_port_read_source` is independent from `JERRY_MODULE_SYSTEM`. The doc says only a note about that if the module system enabled this port function is used for processing `import from...`, but regardless it `Opens a source file and reads its contents into a buffer.`.

akosthekiss: Well, most of the port API functions have some comment at the end of such notes stating whether a function is unused in a given configuration, or whether they are used by application code as well. This function has no such note, unfortunately.

But that bragging does not move this PR forward, perhaps.

However, then, why would it hurt to be on the safe side, and write

```c
if (err_line != 0 && err_col > 0 && err_col < SYNTAX_ERROR_MAX_LINE_LENGTH)
{
  uint32_t curr_line = 1;
  uint32_t pos = 0;

  /* Temporarily modify the error message, so we can use the path. */
  *path_str_end_p = '\0';

  size_t source_size;
  uint8_t *source_p = jerry_port_read_source (path_str_p, &source_size);

  /* Revert the error message. */
  *path_str_end_p = ':';

  if (source_p != NULL)
  {
    /* 2. seek and print */
    while (pos < source_size && curr_line < err_line)
    {
      if (source_p[pos] == '\n')
      {
        curr_line++;
      }

      pos++;
    }

    /* Print character if:
     * - The max line length is not reached.
     * - The current position is valid (it is not the end of the source).
     * - The current character is not a newline.
     **/
    for (uint32_t char_count = 0;
         (char_count < SYNTAX_ERROR_MAX_LINE_LENGTH) && (pos < source_size) && (source_p[pos] != '\n');
         char_count++, pos++)
    {
      jerry_port_log (JERRY_LOG_LEVEL_ERROR, "%c", source_p[pos]);
    }
    jerry_port_log (JERRY_LOG_LEVEL_ERROR, "\n");

    jerry_port_release_source (source_p);

    while (--err_col)
    {
      jerry_port_log (JERRY_LOG_LEVEL_ERROR, "~");
    }

    jerry_port_log (JERRY_LOG_LEVEL_ERROR, "^\n\n");
  }
}
```

rerobika: I think no one is bragging here, only your and mine sights about the ideal solution differs. I'm not convinced but let's move forward: I can deal with the code suggested above.

akosthekiss: Please, note: "bragging" was a self-reference.

akosthekiss (approved):
LGTM

dbatyai (approved):
LGTM

PR #4426, opened 2021-01-05
Fix VM stack size calculation for class initializer assignments

### LINKED ISSUE #4241: Stack corruption on Windows related to static class fields
###### JerryScript revision
f58f15077f4ca975285025e484cbff58fc061ab4
(latest master - 2020.09.25.)

update: still valid on 68b47f5f300453b0d22a0c28a4e8f30e462d424b (latest master - 2020.12.30.)

###### Build platform
Windows 10 (x86 32 bit jerry.exe)

###### Build steps
./tools/build.py --clean --debug --logging=on --error-messages=on

###### Test cases
- tests\jerry\es.next\class-fields1.js
- tests\jerry\es.next\class-fields3.js
- tests\jerry\es.next\class-fields4.js
- tests\jerry\es.next\class-fields5.js

##### Output
```
Run-Time Check Failure #4 - Stack area around _alloca memory reserved by this function is corrupted
Address: 0x006FF0B0
Size: 48
Allocation number within this function: 1
Data: <T�o> 54 F1 6F 00 1C 45 0C 01 1C 45 0C 01 DC F0 6F 00 
```

backtrace:
```
jerry.exe!vm_run(vm_frame_ctx_shared_t * shared_p, unsigned int this_binding_value, ecma_object_t * lex_env_p) Line 4984	C
jerry.exe!opfunc_init_static_class_fields(unsigned int function_object, unsigned int this_val) Line 1054	C
jerry.exe!vm_loop(vm_frame_ctx_t * frame_ctx_p) Line 2102	C
jerry.exe!vm_execute(vm_frame_ctx_t * frame_ctx_p) Line 4876	C
jerry.exe!vm_run(vm_frame_ctx_shared_t * shared_p, unsigned int this_binding_value, ecma_object_t * lex_env_p) Line 4983	C
jerry.exe!vm_run_global(const ecma_compiled_code_t * bytecode_p) Line 339	C
jerry.exe!jerry_run(const unsigned int func_val) Line 593	C
jerry.exe!main(int argc, char * * argv) Line 123	C
[External Code]	
[Frames below may be incorrect and/or missing, no symbols loaded for kernel32.dll]	
```

###### Expected behavior
no stack corruption

### LINKED ISSUE #4413: Dynamic stack buffer overflow in class initializers
###### JerryScript revision
5962562e2dd1528472e6bea38e354b6ea73f559d

###### Build platform
Ubuntu 20.04.1 LTS (Linux 5.4.0-42-generic x86_64)
gcc (Ubuntu 9.3.0-17ubuntu1~20.04) 9.3.0

###### Build steps
```
--compile-flag=-fno-common --compile-flag=-O2 --debug \
--system-allocator=on --linker-flag=-fuse-ld=gold
```

###### Test case
```javascript
new (class C2 { x = 5 })
```

###### Output
```sh
==9633==ERROR: AddressSanitizer: dynamic-stack-buffer-overflow on address 0xffd311ec at pc 0x566f201e bp 0xffd30ee8 sp 0xffd30ed8
WRITE of size 4 at 0xffd311ec thread T0
    #0 0x566f201d in vm_loop /home/jerryscript/jerry-core/vm/vm.c:3034
    #1 0x566d8988 in vm_execute /home/jerryscript/jerry-core/vm/vm.c:4953
    #2 0x566da514 in vm_run /home/jerryscript/jerry-core/vm/vm.c:5060
    #3 0x566fbae2 in opfunc_init_class_fields /home/jerryscript/jerry-core/vm/opcodes.c:1021
    #4 0x5673cb00 in ecma_op_function_call_native /home/jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1244
    #5 0x5674517d in ecma_op_function_construct_native /home/jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1474
    #6 0x5674517d in ecma_op_function_construct /home/jerryscript/jerry-core/ecma/operations/ecma-function-object.c:1552
    #7 0x566d8f35 in opfunc_construct /home/jerryscript/jerry-core/vm/vm.c:907
    #8 0x566d8f35 in vm_execute /home/jerryscript/jerry-core/vm/vm.c:4980
    #9 0x566da514 in vm_run /home/jerryscript/jerry-core/vm/vm.c:5060
    #10 0x56781451 in vm_run_global /home/jerryscript/jerry-core/vm/vm.c:350
    #11 0x56781451 in jerry_run /home/jerryscript/jerry-core/api/jerry.c:608
    #12 0x56781451 in jerry_run /home/jerryscript/jerry-core/api/jerry.c:582
    #13 0x5665bef7 in main /home/jerryscript/jerry-main/main-unix.c:123
    #14 0xf774dee4 in __libc_start_main (/lib32/libc.so.6+0x1eee4)
    #15 0x5665dd94 in _start (/home/jerryscript/build/bin/jerry+0xcd94)

Address 0xffd311ec is located in stack of thread T0
SUMMARY: AddressSanitizer: dynamic-stack-buffer-overflow /home/jerryscript/jerry-core/vm/vm.c:3034 in vm_loop
Shadow bytes around the buggy address:
  0x3ffa61e0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa61f0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa6200: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa6210: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa6220: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x3ffa6230: 00 00 00 00 ca ca ca ca 00 00 00 00 00[04]cb cb
  0x3ffa6240: cb cb cb cb 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa6250: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa6260: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa6270: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
  0x3ffa6280: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
  Addressable:           00
  Partially addressable: 01 02 03 04 05 06 07
  Heap left redzone:       fa
  Freed heap region:       fd
  Stack left redzone:      f1
  Stack mid redzone:       f2
  Stack right redzone:     f3
  Stack after return:      f5
  Stack use after scope:   f8
  Global redzone:          f9
  Global init order:       f6
  Poisoned by user:        f7
  Container overflow:      fc
  Array cookie:            ac
  Intra object redzone:    bb
  ASan internal:           fe
  Left alloca redzone:     ca
  Right alloca redzone:    cb
  Shadow gap:              cc
==9633==ABORTING
```
<sup>Found by [Fuzzinator](http://fuzzinator.readthedocs.io/) with [grammarinator](https://github.com/renatahodovan/grammarinator).</sup>

rerobika:
This patch fixes #4413 and fixes #4241.

### DIFF
--- a/jerry-core/parser/js/js-parser.c
+++ b/jerry-core/parser/js/js-parser.c
@@ -2533,6 +2533,13 @@ parser_parse_class_fields (parser_context_t *context_p) /**< context */
     if (class_field_type & PARSER_CLASS_FIELD_NORMAL)
     {
       parser_emit_cbc_literal (context_p, CBC_ASSIGN_PROP_THIS_LITERAL, literal_index);
+
+      /* Prepare stack slot for assignment property reference base. Needed by vm.c */
+      if (context_p->stack_limit == context_p->stack_depth)
+      {
+        context_p->stack_limit++;
+        JERRY_ASSERT (context_p->stack_limit <= PARSER_MAXIMUM_STACK_LIMIT);
+      }
     }
     else
     {
--- a/tests/jerry/es.next/regression-test-issue-4413.js
+++ b/tests/jerry/es.next/regression-test-issue-4413.js
@@ -1,0 +1,18 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+assert((new (class C2 { x = 5 })).x === 5);
+assert((new (class C2 { x = {} + {} })).x === "[object Object][object Object]");
+assert((new (class C2 { ["foo"] = 5; x = {} + {} })).foo === 5);
+assert((new (class C2 { ["foo"] = 5; x = 6 })).foo === 5);


zherczeg (approved):
LGTM

jerry-core/parser/js/js-parser.c
```diff
@@ -2533,6 +2533,12 @@ parser_parse_class_fields (parser_context_t *context_p) /**< context */
     if (class_field_type & PARSER_CLASS_FIELD_NORMAL)
     {
       parser_emit_cbc_literal (context_p, CBC_ASSIGN_PROP_THIS_LITERAL, literal_index);
+
+      /* Prepare stack slot for assignment property reference base */
+      if (context_p->stack_limit == context_p->stack_depth)
+      {
+        context_p->stack_limit++;
```
akosthekiss, line 2540: isn't something like
```c
        if (context_p->stack_limit > PARSER_MAXIMUM_STACK_LIMIT)
        {
          parser_raise_error (context_p, PARSER_ERR_STACK_LIMIT_REACHED);
        }
```
needed here?

zherczeg: Maybe we need the assert after all :D
stack_depth can be 1 or 2, so reaching PARSER_MAXIMUM_STACK_LIMIT is impossible

rerobika: Indeed, I think the assert would be enough.

rerobika:
@ossy-szeged validated the patch on windows and it seems it fixes #4241 as well.

jerry-core/parser/js/js-parser.c
```diff
@@ -2533,6 +2533,13 @@ parser_parse_class_fields (parser_context_t *context_p) /**< context */
     if (class_field_type & PARSER_CLASS_FIELD_NORMAL)
     {
       parser_emit_cbc_literal (context_p, CBC_ASSIGN_PROP_THIS_LITERAL, literal_index);
+      JERRY_ASSERT (context_p->stack_depth <= 2);
```
akosthekiss, line 2536: Is this the correct assertion? It must be only me, but I cannot trivially see why this assert ensures `context_p->stack_limit <= PARSER_MAXIMUM_STACK_LIMIT` (which is the invariant that needs to hold, if I'm right). Wouldn't `JERRY_ASSERT (context_p->stack_limit <= PARSER_MAXIMUM_STACK_LIMIT);` be the right thing to add _after_ `context_p->stack_limit++;` ?

rerobika: The `stack limit` is only increased by one if the actual `stack_depth` is equal to the `stack_limit`. This assertion is correct since after parsing the rhs of the assignment and emitting `CBC_ASSIGN_PROP_THIS_LITERAL` the maximum actual stack depth is 1 or 2, depending on whether the class has computed initializer field(s). However we do not exact stack limit currently, but if the execution reaches this point it means the limit is less than `PARSER_MAXIMUM_STACK_LIMIT`.

So this could only cause problem if (in a really weird scenario) the stack limit would be `2` or `1` since increasing it to`3` or `2` would reach the limit. However in this case this line is not executed since the engine would already thrown error during the rhs part parsing of the assignment.

akosthekiss: I'm a bit hesitant. Let me explain. What would be the "real" problem? Is it stack_depth being larger than 2, or stack_limit growing above the max allowed value? After thorough analysis (your explanation above) we "know" (suspect, assume, expect) that stack_limit cannot grow too large, because rhs parsing would already throw error, etc. But that's kind of a "transitive" knowledge. If any (hypothetical) change happens in that part of the code, we have to revisit the assumptions here whether they still hold. Whether we can still rely on them here. Whether `stack_depth <= 2` still ensures that `stack_limit <= PARSER_MAXIMUM_STACK_LIMIT`. However, asserting directly on `stack_limit <= PARSER_MAXIMUM_STACK_LIMIT` does not have this problem. (I think)

rerobika: Fine, to reduce the transitive dependency, I've added the requested assertion.

jerry-core/parser/js/js-parser.c
```diff
@@ -2533,6 +2533,14 @@ parser_parse_class_fields (parser_context_t *context_p) /**< context */
...
+
+      /* Prepare stack slot for assignment property reference base. Needed by vm.c */
+      if (context_p->stack_limit == context_p->stack_depth)
+      {
+        JERRY_ASSERT (context_p->stack_limit <= PARSER_MAXIMUM_STACK_LIMIT);
```
akosthekiss, line 2541: The assertion should come _after_ the change of stack_limit.

(Plus, is the other assert still necessary then?)

akosthekiss (changes requested):

jerry-core/parser/js/js-parser.c
rerobika: Indeed, it can be removed.

akosthekiss (approved):
LGTM

PR #4427, opened 2021-01-06
Remove file(GLOB ...) usage from CMakeLists.txt

rerobika:


LaszloLango:
Why is this needed? What is the problem with GLOB?

rerobika:
From Cmake 3.19.2 docs:

> Note: We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate.

jerry-core/CMakeLists.txt
```diff
@@ -122,20 +122,191 @@ set(INCLUDE_CORE_PRIVATE ${INCLUDE_CORE_PRIVATE} PARENT_SCOPE) # for tests/unit-
...
+    parser/regexp/re-compiler.c
+    parser/regexp/re-parser.c
+    vm/opcodes-ecma-arithmetics.c
+    vm/opcodes-ecma-bitwise.c
+    vm/opcodes-ecma-relational-equality.c
+    vm/opcodes.c
+    vm/vm-stack.c
+    vm/vm-utils.c
+    vm/vm.c
+    )
```
akosthekiss, line 309: (style) This looks weird. But more importantly, it is used inconsistently. Sometimes it is indented by 4 spaces (like here). Sometimes it is indented by 3 spaces (so that the closing parentheses aligns with the opening one). And sometimes it is on the same line as the last element in the list.

Personally, I don't like any of them.
```cmake
set(LIST_VAR
  elem1
  elem2
  elem3
)
```
This style above is more appealing to me. (Indentation is 2 spaces, like anywhere else in the cmake files. Closing parenthesis is on a separate line to allow for clean list edits. Etc.)

However, I do realize that the "closing parenthesis on the line of the last item" is the most commonly used style throughout the files.

jerry-math/CMakeLists.txt
```diff
@@ -28,14 +28,46 @@ set(COMPILE_FLAGS_MATH "${COMPILE_FLAGS_MATH} -Wno-strict-aliasing")
...
+    )
 
 # Amalgamated JerryScript source/header build.
 #  The process will create the following files:
 #   * jerryscript-math.c
 #   * math.h
 if(ENABLE_AMALGAM)
-  file(GLOB HEADER_MATH *.h)
+  set(HEADER_MATH
+      include/math.h
```
akosthekiss, line 69: Is this right? jerry-math also has an internal header `jerry-math-internal.h`.

jerry-core/CMakeLists.txt
rerobika: I'm ok with the suggested style.

jerry-math/CMakeLists.txt
rerobika: Thank, that was a rebase error.

rerobika:
@akosthekiss Thanks for the review, I've updated the PR.

akosthekiss (approved):
LGTM

ossy-szeged:
LGTM (informal)

LaszloLango (approved):
LGTM

PR #4428, opened 2021-01-06
Disable libjerry-math by default

akosthekiss:
Normally, it is more usual and safe to use a toolchain's native
math library. Especially, if multiple components of a project use
math functions, in which case all components should be linked
against the same libm.

The libjerry-math can be used, of course, but as it needs extra
care and consideration, it should be opt-in.


tests/jerry/es.next/math-cbrt.js
```diff
@@ -29,5 +29,5 @@ assert(Math.cbrt(Number.NEGATIVE_INFINITY) === Number.NEGATIVE_INFINITY);
 assert(Math.cbrt(1.0) === 1.0);
 assert(Math.cbrt(-1.0) === -1.0);
 
-assert(Math.cbrt(27.0) === 3.0);
+// assert(Math.cbrt(27.0) === 3.0); // FIXME: unstable, depending on compiler and libm
```
lygstate, line 32: How much unsable it is? do we need specialize cbrt?

akosthekiss: This is not a test262 failure, but a failure in jerry's own test suite.
As far as I could track it down, this a floating point numeric instability issue, perhaps rounding on the least significant bit (3.000...x != 3.000...0). Observed with gcc + libm + -fno-builtins. Does not reproduce with clang. (thanks to @galpeter)

rerobika (approved):
LGTM

zherczeg (approved):
LGTM

PR #4430, opened 2021-01-06
Move ARM and target checks to Github Actions from Travis

rerobika:
Note: For easier porting RIOT-STM32F4 target is updated to their latest release.

### DIFF
--- a/.github/workflows/gh-actions.yml
+++ b/.github/workflows/gh-actions.yml
@@ -125,3 +125,85 @@ jobs:
           $RUNNER -q --jerry-tests
           --buildoptions=--compile-flag=-fsanitize=undefined,--compile-flag=-m32,--compile-flag=-fno-omit-frame-pointer,--compile-flag=-fno-common,--debug,--system-allocator=on,--linker-flag=-fuse-ld=gold
           --skip-list=parser-oom.js,parser-oom2.js
+
+  Linux_ARMv7l_Tests:
+    runs-on: ubuntu-latest
+    env:
+      RUNTIME: qemu-arm-static
+      TIMEOUT: 300
+    steps:
+      - uses: actions/checkout@v2
+      - run: sudo apt update
+      - run: sudo apt install gcc-arm-linux-gnueabihf libc6-dev-armhf-cross qemu-user-static
+      - run: >-
+          $RUNNER -q --jerry-tests
+          --buildoptions=--toolchain=cmake/toolchain_linux_armv7l.cmake,--linker-flag=-static
+
+  Linux_AArch64_Tests:
+    runs-on: ubuntu-latest
+    env:
+      RUNTIME: qemu-aarch64-static
+      TIMEOUT: 300
+    steps:
+      - uses: actions/checkout@v2
+      - run: sudo apt update
+      - run: sudo apt install gcc-aarch64-linux-gnu libc6-dev-armhf-cross qemu-user-static
+      - run: >-
+          $RUNNER -q --jerry-tests
+          --buildoptions=--toolchain=cmake/toolchain_linux_aarch64.cmake,--linker-flag=-static
+
+  MbedOS5_K64F_Build_Test:
+    runs-on: ubuntu-18.04 # needed due to ppa:team-gcc-arm-embedded/ppa
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.8' # needed due to 'intelhex' module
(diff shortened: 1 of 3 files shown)


.github/workflows/gh-actions.yml
```diff
@@ -150,3 +150,91 @@ jobs:
...
+  Linux_ARMv7l_Tests:
+    runs-on: ubuntu-20.04
+    env:
+      RUNTIME: qemu-arm-static
+      TIMEOUT: 300
+    steps:
+      - uses: actions/checkout@v2
+      - run: |
+          sudo apt update
+          sudo apt install gcc-arm-linux-gnueabihf libc6-dev-armhf-cross qemu-user-static
```
akosthekiss, line 163: Is the yaml block syntax needed? Could this be
```yaml
- run: sudo apt update
- run: sudo apt install gcc-arm-linux-gnueabihf libc6-dev-armhf-cross qemu-user-static
```
?
(It's shorter and the two commands get green/red status separately, which may be helpful in case of an error.)

.github/workflows/gh-actions.yml
```diff
@@ -150,3 +150,91 @@ jobs:
         run: sudo apt-get install gcc-multilib
       - name: UBSAN - jerry-tests
         run: $RUNNER --jerry-tests $SKIP_LIST $BUILD_OPTIONS -q
+
+  Linux_ARMv7l_Tests:
+    runs-on: ubuntu-20.04
```
akosthekiss, line 155: Could be ubuntu-latest (BTW, ubuntu-latest and ubuntu-20.04 are now the same.)

.github/workflows/gh-actions.yml
```diff
@@ -150,3 +150,91 @@ jobs:
...
+          $RUNNER --jerry-tests -q
+          --buildoptions=--toolchain=cmake/toolchain_linux_aarch64.cmake,--linker-flag=-static
+
+  MbedOS5_K64F_Build_Test:
+    runs-on: ubuntu-18.04 # needed due to ppa:team-gcc-arm-embedded/ppa
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-python@v2
+        with:
+          python-version: '3.6'
```
akosthekiss, line 188: This could by '3.x', if 3.6 is not an exact requirement.

rerobika: According to https://github.com/jerryscript-project/jerryscript/blob/master/.travis.yml#L57 3.6 is a requirement.  I've tried it with '3.x' but MBED build requires it see: https://github.com/jerryscript-project/jerryscript/runs/1662357072?check_suite_focus=true#step:8:32.

However we can use '3.x' in `Zephyr_Arduino_101_Build_Test`.

akosthekiss: BTW, from the logs and backtraces it seems that python's `array.array.tostring` is expected by an old version of the intelhex package. That function had been deprecated since py3.2 and was removed in py3.9. So, actually, 3.8 should be the highest python version to work (i.e., not necessarily 3.6).

Note that the latest version of intelhex is already compatible with py3.9. And mbed-cli installs that version. However, when installing `targets/mbedos5/mbed-os/requirements.txt`, it downgrades intelhex to an old version. Most probably the mbed-os revision of the target is too old.

Perhaps the python-version deserves a comment just like runs-on has one.

akosthekiss:
Please, rebase.

.github/workflows/gh-actions.yml
```diff
@@ -125,3 +125,85 @@ jobs:
...
+    runs-on: ubuntu-latest
+    env:
+      RUNTIME: qemu-arm-static
+      TIMEOUT: 300
+    steps:
+      - uses: actions/checkout@v2
+      - run: sudo apt update
+      - run: sudo apt install gcc-arm-linux-gnueabihf libc6-dev-armhf-cross qemu-user-static
+      - run: >-
+          $RUNNER --jerry-tests -q
```
akosthekiss, line 139: I personally prefer to have -q / --quiet as the first argument (as it was in the travis configuration). But that is somewhat subjective.

rerobika: Personally I prefer it as the last argument, but I can deal with both options. The choice is on you.

rerobika: Ignore it, I checked the usage elsewhere and `-q` is always the first argument.

akosthekiss (approved):
LGTM

lygstate:
LGTM

dbatyai (approved):
LGTM

PR #4436, opened 2021-01-10
Farewell to Travis CI

akosthekiss:
Travis CI became very slow lately. Most of the jobs have already
been moved to GH Actions. This commit finishes the migration.

- IRC notification is adapted (result of jobs is not reported).
- Sonar integration is dropped without replacement, as its GH
  Action does not support C projects.
- Coverity Scan integration is dropped without replacement as it
  has no GH Action at all.


akosthekiss:
Let me know if this step is too aggressive and Coverity Scan and/or SonarCloud integration should be kept until their GH Action becomes available.

README.md
```diff
@@ -1,11 +1,8 @@
 ![](https://github.com/jerryscript-project/jerryscript/blob/master/LOGO.png)
 # JerryScript: JavaScript engine for the Internet of Things
 [![License](https://img.shields.io/badge/licence-Apache%202.0-brightgreen.svg?style=flat)](LICENSE)
-[![Travis CI Build Status](https://travis-ci.org/jerryscript-project/jerryscript.svg?branch=master)](https://travis-ci.org/jerryscript-project/jerryscript)
```
rerobika, line 4: According to the [docs](https://docs.github.com/en/free-pro-team@latest/actions/managing-workflow-runs/adding-a-workflow-status-badge) we can create a badge for GH actions as well. I think it would be useful.

rerobika (approved):
LGTM

README.md
akosthekiss: Indeed, added.

zherczeg (approved):
LGTM

PR #4437, opened 2021-01-11
Fix HAVE_M_LIB check for older CMake

ossy-szeged:
With older CMake (for example 2.8.12.2) the feature check of HAVE_M_LIB
reports false positive fail due to -Werror=strict-prototypes. It is already
fixed in CMake https://gitlab.kitware.com/cmake/cmake/-/commit/4bc17345

The feature is present and works, so we should only ignore this build warning.

Additionally the necessary GCC/Clang guard added to the similar
CMake feature check in jerry-port/default/CMakeLists.txt


LaszloLango (approved):
Do we really need to support such old build environments? It looks unnecessary extra maintenance work for me. I'd expect more things like this in the future. Otherwise the patch LGTM.

akosthekiss (approved):
LGTM

PR #4443, opened 2021-01-11
Make the whole GH Actions Notification job conditional

akosthekiss:
Don't even start the job if the workflow event is not a push to the
master repo.


rerobika (approved):
LGTM

PR #4444, opened 2021-01-11
Improve the path normalization of the default port

akosthekiss:
Determine the length of strings as early as possible and use that
to prevent unnecessary memory allocations, and in memory copies
instead of string concatenations.


rerobika (approved):
LGTM

LaszloLango (approved):
LGTM

PR #4449, opened 2021-01-11
Fix the build and timezone offset with MinGW

### LINKED ISSUE #4240: Date interface returns date without timezone offset on windows platform
Date interface works fine on ubuntu, but not on windows platform
# on ubuntu
```
jerry> var a = new Date()
undefined
jerry> a
Sun Sep 27 2020 11:08:36 GMT+0800
```

# windows platform
```
jerry> var a = new Date()
undefined
jerry> a
Sun Sep 27 2020 03:10:05 GMT+00:00
jerry>
```

ossy-szeged:
Changes:
- Use _WIN32 macro instead of _WINDOWS
- Fix build error - "no previous declaration for 'UnixTimeToFileTime'"
- Fix conversion build errors
- Remove useless _tzset() call
- Use GetSystemTimeAsFileTime() on Windows to get current time
- Unify and simplify time conversion Windows helper functions

Fixes #4240

### DIFF
--- a/jerry-port/default/default-date.c
+++ b/jerry-port/default/default-date.c
@@ -17,12 +17,12 @@
 #include <time.h>
 #endif /* HAVE_TM_GMTOFF */
 
-#ifdef _WINDOWS
+#ifdef _WIN32
 #include <windows.h>
 #include <winbase.h>
 #include <winnt.h>
 #include <time.h>
-#endif /* _WINDOWS */
+#endif /* _WIN32 */
 
 #ifdef __GNUC__
 #include <sys/time.h>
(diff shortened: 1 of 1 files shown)


jerry-port/default/default-date.c
```diff
@@ -101,17 +100,20 @@ double jerry_port_get_current_time (void)
...
   }
 #endif /* __GNUC__ */
 
-#ifdef _WINDOWS
+#ifdef _WIN32
   time_t ltime;
   time (&ltime);
   return ltime * 1000;
-#endif /* _WINDOWS */
+#endif /* _WIN32 */
```
akosthekiss, line 116: if the `#ifdef _WIN32` check came first, followed by an `#elif defined(__GNUC__)`, would that not make the pragmas unnecessary? (I.e., even if there was a GCC on Windows -- MinGW --, the Windows-specific code would be taken.)

LaszloLango: Is this comment resolved? I see `#endif /* __GNUC__ */` now instead of `#elif`

LaszloLango (approved):
LGTM

jerry-port/default/default-date.c
akosthekiss: This is still the original code, not updated according to review.

ossy-szeged: The idea is good. But I found a better solution, we can use time() on all platform, we don't need platform specific implementation at all here.

jerry-port/default/default-date.c
```diff
@@ -98,20 +97,7 @@ double jerry_port_get_local_time_zone_adjustment (double unix_ms,  /**< ms since
...
-#endif /* __GNUC__ */
-
-#ifdef _WINDOWS
   time_t ltime;
   time (&ltime);
-  return ltime * 1000;
-#endif /* _WINDOWS */
-
-  return 0.0;
+  return (double) ltime * 1000.0;
```
akosthekiss, line 102: But `time()` has seconds granularity only! `jerry_port_get_current_time` is used by `Date.now()`, which is returns the number of _milliseconds_ elapsed since January 1, 1970 00:00:00 UTC. I think the Windows-specific code path used `time()` only as a fallback, as (non-MinGW) Windows has no `gettimeofday()`. But I don't think that we should lower precision generally.

akosthekiss (changes requested):

jerry-port/default/default-date.c
ossy-szeged: Ah, I missed it. In this case, I think we should keep my original proposal with pragmas not to lower precision for MinGW build.

ossy-szeged: Or maybe we can get epoch time on Windows with ms granularity. Let me try, what I have found.

ossy-szeged: I found the best solution, we should use GetSystemTimeAsFileTime() on Windows. It's a little bit tricky,
because we have to convert Windows Epoch (1601) to Unix Epoch (1970) and the granularity is different too.

jerry-port/default/default-date.c
lygstate: time is seconds， not ms

jerry-port/default/default-date.c
```diff
@@ -65,12 +75,11 @@ double jerry_port_get_local_time_zone_adjustment (double unix_ms,  /**< ms since
...
   (void) unix_ms;
   (void) is_utc;
-#ifdef _WINDOWS
+#ifdef _WIN32
   FILETIME fileTime, localFileTime;
   SYSTEMTIME systemTime, localSystemTime;
   ULARGE_INTEGER time, localTime;
 
-  _tzset ();
   UnixTimeToFileTime ((LONGLONG) (unix_ms / 1000), &fileTime);
```
akosthekiss, line 83: To unify computations, shouldn't this function be also updated a bit? I.e., to have it as `UnixTimeMsToFileTime`, so that division is unnecessary here, and the body of the function becomes actually a mathematical inverse of `FileTimeToUnixTimeMs`.

ossy-szeged: Good point. I already unified and simplified them (casts, magic numbers, ...)

akosthekiss (approved):
LGTM

rerobika (approved):
LGTM

PR #4451, opened 2021-01-11
Introduce jerry_port_track_promise_rejection

### LINKED ISSUE #2931: Discussion: Report unhandled promise rejections

rerobika:
This patch resolves #2931.

### DIFF
--- a/docs/05.PORT-API.md
+++ b/docs/05.PORT-API.md
@@ -151,6 +151,32 @@ jerry_port_get_native_module (jerry_value_t name) /**< module specifier */
 }
 ```
 
+## Promise
+
+```c
+/**
+ * HostPromiseRejectionTracker operations
+ */
+typedef enum
+{
+  JERRY_PROMISE_REJECTION_OPERATION_REJECT, /**< promise is rejected without any handlers */
+  JERRY_PROMISE_REJECTION_OPERATION_HANDLE, /**< handler is added to a rejected promise for the first time */
+} jerry_promise_rejection_operation_t;
+
+/**
+ * Track unhandled promise rejections.
+ *
+ * Note:
+ *      This port function is called by jerry-core when JERRY_BUILTIN_PROMISE
+ *      is enabled.
+ *
+ * @param promise rejected promise
+ * @param operation HostPromiseRejectionTracker operation
+ */
+void jerry_port_track_promise_rejection (const jerry_value_t promise,
+                                         const jerry_promise_rejection_operation_t operation);
+```
+
 ## Date
 
 ```c
@@ -391,4 +417,33 @@ void jerry_port_sleep (uint32_t sleep_time)
   (void) sleep_time;
 } /* jerry_port_sleep */
 #endif /* defined (JERRY_DEBUGGER) && (JERRY_DEBUGGER == 1) */
+```
+
+## Promise
+
+```c
+#include "jerryscript-port.h"
+
+/**
+ * Default implementation of jerry_port_track_promise_rejection.
+ * Prints the reason of the unhandled rejections.
+ */
+void
+jerry_port_track_promise_rejection (const jerry_value_t promise, /**< rejected promise */
+                                    const jerry_promise_rejection_operation_t operation) /**< operation */
+{
+  (void) operation; /* unused */
+
+  jerry_value_t reason = jerry_get_promise_result (promise);
+  jerry_value_t reason_to_string = jerry_value_to_string (reason);
+  jerry_size_t req_sz = jerry_get_utf8_string_size (reason_to_string);
+  jerry_char_t str_buf_p[req_sz + 1];
+  jerry_string_to_utf8_char_buffer (reason_to_string, str_buf_p, req_sz);
+  str_buf_p[req_sz] = '\0';
+
+  jerry_release_value (reason_to_string);
+  jerry_release_value (reason);
+
+  printf ("Uncaught (in promise) %s\n", str_buf_p);
+} /* jerry_port_track_promise_rejection */
 ```
--- a/jerry-core/ecma/operations/ecma-promise-object.c
+++ b/jerry-core/ecma/operations/ecma-promise-object.c
@@ -182,6 +182,23 @@ ecma_is_resolver_already_called (ecma_object_t *promise_obj_p) /**< promise */
 } /* ecma_is_resolver_already_called */
 
 /**
+ * HostPromiseRejectionTracker
+ *
+ * See also: ES11 25.6.1.9
+ */
+static void
+ecma_track_promise_rejection (ecma_object_t *obj_p, /**< rejected promise */
+                              jerry_promise_rejection_operation_t operation) /**< operation */
+{
+  JERRY_ASSERT (ecma_is_promise (obj_p));
+
+  if (!(ecma_promise_get_flags (obj_p) & ECMA_PROMISE_HANDLED))
+  {
+    jerry_port_track_promise_rejection (ecma_make_object_value (obj_p), operation);
+  }
+} /* ecma_track_promise_rejection */
+
+/**
  * Reject a Promise with a reason.
  *
  * See also: ES2015 25.4.1.7
@@ -208,6 +225,7 @@ ecma_reject_promise (ecma_value_t promise, /**< promise */
   promise_p->reactions = ecma_new_collection ();
 
   ecma_collection_destroy (reactions);
+  ecma_track_promise_rejection (obj_p, JERRY_PROMISE_REJECTION_OPERATION_REJECT);
 } /* ecma_reject_promise */
 
 /**
@@ -818,9 +836,13 @@ ecma_promise_do_then (ecma_value_t promise, /**< the promise which call 'then' *
   {
     /* 9. */
     ecma_value_t reason = ecma_promise_get_result (promise_obj_p);
+    ecma_track_promise_rejection (promise_obj_p, JERRY_PROMISE_REJECTION_OPERATION_HANDLE);
     ecma_enqueue_promise_reaction_job (ecma_make_object_value (result_capability_obj_p), on_rejected, reason);
     ecma_free_value (reason);
   }
+
+  /* ES11: 11. */
+  promise_p->header.u.class_prop.extra_info |= ECMA_PROMISE_HANDLED;
 
   /* 10. */
   return ecma_copy_value (capability_p->header.u.class_prop.u.promise);
--- a/jerry-port/default/default-promise.c
+++ b/jerry-port/default/default-promise.c
@@ -1,0 +1,39 @@
+/* Copyright JS Foundation and other contributors, http://js.foundation
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript-port.h"
+
+/**
+ * Default implementation of jerry_port_track_promise_rejection.
+ * Prints the reason of the unhandled rejections.
+ */
+void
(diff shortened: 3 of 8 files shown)


jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -181,6 +181,23 @@ ecma_is_resolver_already_called (ecma_object_t *promise_obj_p) /**< promise */
...
+ */
+static void
+ecma_promise_track_rejection (ecma_object_t *obj_p, /**< promise */
+                              ecma_value_t reason) /**< reason for reject */
+{
+  JERRY_ASSERT (ecma_is_promise (obj_p));
+
+  if (!(ecma_promise_get_flags (obj_p) & ECMA_PROMISE_HANDLED))
+  {
+    jerry_port_report_unhandled_promise_rejection (ecma_make_object_value (obj_p), reason);
```
akosthekiss, line 197: The name seems a bit counterintuitive to me. If the abstract operation of the spec is called HostPromiseRejectionTracker and the ecma core function is called `ecma_promise_track_rejection`, then why is the port function called `jerry_port_report_unhandled_promise_rejection`? Why report? Why unhandled?

Wouldn't be `jerry_port_track_promise_rejection` more natural?

jerry-core/ecma/operations/ecma-promise-object.c
```diff
@@ -181,6 +181,23 @@ ecma_is_resolver_already_called (ecma_object_t *promise_obj_p) /**< promise */
...
 } /* ecma_is_resolver_already_called */
 
+/**
+ * HostPromiseRejectionTracker
+ *
+ * See also: ES11 25.6.1.9
+ */
+static void
+ecma_promise_track_rejection (ecma_object_t *obj_p, /**< promise */
+                              ecma_value_t reason) /**< reason for reject */
```
akosthekiss, line 191: Is this correct? According to the referenced section in the spec, the second argument of HostPromiseRejectionTracker is "operation", which can only take two values: "reject" and "handle". (Also see 25.6.1.7 (RejectPromise) and 25.6.5.4.1 (PerformPromiseThen)).

jerry-port/default/default-promise.c
```diff
@@ -0,0 +1,42 @@
...
+
+  jerry_value_t reason_to_string = jerry_value_to_string (reason);
+  jerry_size_t req_sz = jerry_get_utf8_string_size (reason_to_string);
+  JERRY_VLA (jerry_char_t, str_buf_p, req_sz + 1);
+  jerry_string_to_utf8_char_buffer (reason_to_string, str_buf_p, req_sz);
+  str_buf_p[req_sz] = '\0';
+
+  jerry_release_value (reason_to_string);
+
+  jerry_port_log (jerry_port_default_get_log_level (), "Uncaught (in promise) %s\n", str_buf_p);
```
akosthekiss, line 41: Why `jerry_port_default_get_log_level ()`? Looks like a hack to make the log always pass through. Why not a fixed WARNING level?

jerry-port/default/default-promise.c
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * This function called when:
+ *  - A promise is rejected without any handlers
+ *  - A handler is added to a rejected promise for the first time
+ */
+void
+jerry_port_report_unhandled_promise_rejection (const jerry_value_t promise, /**< rejected promise */
+                                               const jerry_value_t reason) /**< reason of the rejection */
+{
+  (void) promise;
```
akosthekiss, line 31: (style) Add a comment `/* unused */`?

jerry-core/ecma/operations/ecma-promise-object.c
rerobika: > Why report? Why unhandled?

  - A typical implementation of [HostPromiseRejectionTracker ](https://www.ecma-international.org/ecma-262/11.0/#sec-host-promise-rejection-tracker)might try to notify developers of unhandled rejections,
 
 I wasn't sure with the name, but I think `jerry_port_track_promise_rejection` is fine

jerry-core/ecma/operations/ecma-promise-object.c
rerobika: These two values are the kind of the rejection.

"reject":
```js
Promise.reject(42)
> Uncaught (in promise) 42
```
in this case the port function is triggered with `(<new promise>, <42>)`


"handle":
```js
Promise.race()
> Uncaught (in promise) TypeError: undefined is not iterable...
```
in this case the port function is triggered with `(<new promise>, <TypeError>)`

IMO the point is the reason of the rejection not the type of the rejection. However, we can pass an enum value as a 3rd argument to indicate these two modes but i think it's superfluous.

jerry-port/default/default-promise.c
rerobika: Warning will be fine.

jerry-core/ecma/operations/ecma-promise-object.c
akosthekiss: But I've thought that by the time control reaches the HostPromiseRejectionTracker, the promise object already "contains" the reason for rejection, so the tracker can extract it from the promise if needed.

In `ecma_promise_do_then`, step 9, that's already the case:
```c
    ecma_value_t reason = ecma_promise_get_result (promise_obj_p);
    ecma_promise_track_rejection (promise_obj_p, reason);
```

And ES11 25.6.1.7 (RejectPromise), step 3 ("Set promise.[[PromiseResult]] to reason.") also deals with this. It seems, however, that the current code (`ecma_reject_promise`) does not follow this abstract operation.

On the other hand, the "operation" argument is something that is not / cannot be deduced or extracted from either the promise object or the reason for rejection. Thus, if we want the engine to be conforming to the spec, I think we need to have that argument. (We cannot know whether a user of the engine wants to make use of it or not. So, we should err on the safe side.)

I see two options now:
1) Have the port function take two arguments, promise and operation, as per the spec. But I don't know how big changes that may implicate.
2) Have the port function take three arguments, promise, reason, and operation, and document that `ecma_promise_get_result` (`jerry_get_promise_result`?) may or may not contain the reason and therefore port implementers must not rely on that but have to use the reason argument instead.

akosthekiss (commented):
The commit message got mixed up a bit during the update, please fix (superfluous "report" remained.) Plus, the PR title could also be updated.

jerry-core/ecma/operations/ecma-promise-object.c
rerobika: > And ES11 25.6.1.7 (RejectPromise), step 3 ("Set promise.[[PromiseResult]] to reason.") also deals with this. It seems, however, that the current code (ecma_reject_promise) does not follow this abstract operation.

It is set by :https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/ecma/operations/ecma-promise-object.c#L198

So probably the 2 arguments will be enough, just need to update the port implementation to obtain the reason from the rejected promise.

akosthekiss: Ah, OK. I've missed that.

docs/05.PORT-API.md
```diff
@@ -151,6 +151,32 @@ jerry_port_get_native_module (jerry_value_t name) /**< module specifier */
...
+
+```c
+/**
+ * HostPromiseRejectionTracker operations
+ */
+typedef enum
+{
+  JERRY_PROMISE_REJECTION_TRACKER_REJECT, /**< promise is rejected without any handlers */
+  JERRY_PROMISE_REJECTION_TRACKER_HANDLE, /**< handler is added to a rejected promise for the first time */
+} jerry_promise_rejection_tracker_t;
```
akosthekiss, line 164: I'm not sure about the names proposed here. This is definitely not the type of a promise rejection tracker (what would be intuitive from this name), but the type of operations. (The type of a tracker would be a function, I guess.)

How about
```c
typedef enum
 {
   JERRY_PROMISE_REJECTION_OPERATION_REJECT, /**< promise is rejected without any handlers */
   JERRY_PROMISE_REJECTION_OPERATION_HANDLE, /**< handler is added to a rejected promise for the first time */
 } jerry_promise_rejection_operation_t;
```
?

(Note: I've also considered suggesting `jerry_promise_rejection_tracker_operation`, as that would follow the pattern of other typedefs in the port API. But that would be too long for a name, I guess.)

docs/05.PORT-API.md
```diff
@@ -151,6 +151,32 @@ jerry_port_get_native_module (jerry_value_t name) /**< module specifier */
...
+ * Note:
+ *      This port function is called by jerry-core when JERRY_BUILTIN_PROMISE
+ *      is enabled.
+ *
+ * @param promise rejected promise
+ * @param operation HostPromiseRejectionTracker operation
+ */
+void jerry_port_track_promise_rejection (const jerry_value_t promise,
+                                         const jerry_promise_rejection_tracker_t operation);
+```
```
akosthekiss, line 178: BTW, shouldn't the "How to port JerryScript" section be extended?

docs/05.PORT-API.md
```diff
@@ -151,6 +151,32 @@ jerry_port_get_native_module (jerry_value_t name) /**< module specifier */
...
+ * Track unhandled promise rejections.
+ *
+ * Note:
+ *      This port function is called by jerry-core when JERRY_BUILTIN_PROMISE
+ *      is enabled.
+ *
+ * @param promise rejected promise
+ * @param operation HostPromiseRejectionTracker operation
+ */
+void jerry_port_track_promise_rejection (const jerry_value_t promise,
```
akosthekiss, line 176: I was also considering suggesting yet another name change, to have the new function named `jerry_port_promise_rejection_tracker`. That would make the new typedef and the new function names aligned, as well as the name used by the spec. However, existing port API function names usually start with a verb (get this, set that, sleep, print, log, etc. -- the only exception being fatal). So, I'm not sure about this. But we should agree upon a good name so that it doesn't need to be changed, at least in the near future.

jerry-port/default/default-promise.c
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "jerryscript-port.h"
+#include "jerryscript-port-default.h"
```
akosthekiss, line 17: Is the default port header needed? Does this file use anything default port-specific?

docs/05.PORT-API.md
rerobika: `jerry_promise_rejection_operation_t` will be fine.

docs/05.PORT-API.md
rerobika: I'd keep this `jerry_port_track_promise_rejection` for the due to the starting verb convention with the other API functions. 

`jerry_port_promise_rejection_tracker` would be suitable if we would rename `jerry_port_log` to `jerry_port_logger` and so on. So I'd leave as it for now.

docs/05.PORT-API.md
rerobika: I've added it.

rerobika:
@akosthekiss Thanks for the review, I've updated the PR.

akosthekiss (approved):
LGTM

dbatyai (approved):
LGTM

PR #4454, opened 2021-01-12
Implement jerry_get_backtrace_from API function

zherczeg:


rerobika (approved):
LGTM

jerry-core/vm/vm-utils.c
```diff
@@ -59,15 +59,42 @@ vm_is_direct_eval_form_call (void)
  * @return array ecma value
  */
 ecma_value_t
-vm_get_backtrace (uint32_t max_depth) /**< maximum backtrace depth, 0 = unlimited */
+vm_get_backtrace (uint32_t max_depth, /**< maximum backtrace depth, 0 = unlimited */
+                  ecma_object_t *ignored_function_p) /**< ignore functions up to this function */
 {
 #if ENABLED (JERRY_LINE_INFO)
+  vm_frame_ctx_t *context_p = JERRY_CONTEXT (vm_top_context_p);
+
```
LaszloLango, line 67: I'd put here an assert, just in case if someone tries to call it from internal code rather than the new API function in the future. Something like: 
```
ecma_object_type_t type = ecma_get_object_type (ignored_function_p);
JERRY_ASSERT (type == ECMA_OBJECT_TYPE_FUNCTION || type == ECMA_OBJECT_TYPE_NATIVE_FUNCTION || type == ECMA_OBJECT_TYPE_BOUND_FUNCTION);
```

LaszloLango (changes requested):
With this new API function the old `jerry_get_backtrace` becomes obsolete. Unfortunately (as far as I know) we do not have any declared process for function deprecation. Not sure this should be discussed under this PR. I'd suggest to mark the `jerry_get_backtrace` as deprecate and suggest to the users to use `jerry_get_backtrace_from` with an undefined jerry_value_t to get the full backtrace instead. What is your opinions?

zherczeg:
I would not call the other obsolote. You don't always want to pass a parameter.

jerry-core/vm/vm-utils.c
zherczeg: The `ignored_function_p` is not mandatory, it can be NULL.

zherczeg: I have added an assertion below.

zherczeg:
Thnak you for the review, the patch is updated.

LaszloLango (approved):
Thanks for the update, LGTM

PR #4457, opened 2021-01-12
Drop `-fno-...` compiler flags from CMakeLists

akosthekiss:
- `-fno-builtin` should not be necessary: a) by default, we assume
  that a libc exists on the system the project is compiled for, but
  b) even if the assumption does not hold, `-fno-builtin` can be
  passed as an external CFLAG to the build system, plus c) the use
  of builtins is expected to help both code size and performance.
  (Except when jerry-math is enabled.)
- `-fno-stack-protector` should not be necessary: it is the default
  not to instrument code with stack protection.


akosthekiss:
Some additional notes:
- Code size does improve a bit indeed (by cca. 300 bytes on ARM32). The effect on performance is statistically insignificant. Memory and stack consumption is not affected.
- The detection of libm had to be amended (in jerry-core's cmake file), because both gcc and clang issued warnings when compiling the test code (gcc: `-Werror=builtin-declaration-mismatch`, clang: `-Werror,-Wincompatible-library-redeclaration`; the solution to work with both compilers was simply `-Wno-error`).

rerobika (approved):
LGTM

LaszloLango (approved):
LGTM

Issue #4445, opened 2021-01-11
heap-use-after-free in ecma_is_lexical_environment


owl337:
###### JerryScript revision
fdaacde6

###### Build platform
Ubuntu 16.04.6 LTS (Linux 4.4.0-179-generic x86_64)

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g --strip=off \
--system-allocator=on --logging=on --linker-flag=-fuse-ld=gold \
--error-messages=on --profile=es2015-subset  
```

###### Test case
```
var a= ["", "\0", "\t", "\n", "\v", "\f", "\r", " ", "\u00a0", "\u2028", "\u2029", "\ufeff"]
Array.prototype[4] = 10;
function Test()
{
    a.sort(function() {
        var A = function() { };
        A.prototype.x = 42;
        var o = new Proxy({
        "3": {
            writable:false,
            value:20
        }
    }, {
    getPrototypeOf: function (val, size, ch) {
    var result = new String(val);
    if (ch == null) {
        ch = " ";
    }
    while (result.length < size) {
        result = ch + result;
    }
    return result;
}
});
        o.x = 43;
        var result = "";
        for (var p in o)
            result += o[p];
        return a | 0;
    });
    WScript.Echo(a);
}

Test();

```
###### Execution steps
```
./build/bin/jerry ./build/bin/poc.js
```

###### Output
```
=================================================================
==179417==ERROR: AddressSanitizer: heap-use-after-free on address 0xf5304720 at pc 0x566b5311 bp 0xff9aa468 sp 0xff9aa458
READ of size 2 at 0xf5304720 thread T0
    #0 0x566b5310 in ecma_is_lexical_environment /root/jerryscript/jerry-core/ecma/base/ecma-helpers.c:177
    #1 0x566b5447 in ecma_get_object_type /root/jerryscript/jerry-core/ecma/base/ecma-helpers.c:203
    #2 0x56670759 in ecma_op_object_get_own_property_descriptor /root/jerryscript/jerry-core/ecma/operations/ecma-objects.c:1862
    #3 0x56672f01 in ecma_op_object_enumerate /root/jerryscript/jerry-core/ecma/operations/ecma-objects.c:2513
[...]

PR #4458, opened 2021-01-13
Fix prototype chain traversing

galpeter:
After the introduction of the Proxy builtin object there was
a possibility to traverse the prototype chain with an invalid object.
The prototype was freed before it's data/properties were queried resulting
in accessing invalid information.

By forcing the allocator to always do a gc (`--mem-stres-test=on` build option)
it was possible to trigger the issue without complicated tests.

New internal method:
* `ecma_op_object_get_prototype_of` which always returns the prototype
  of an object and the return value must be freed (if it is valid).

Updated prototype chain traversing in:
* `jerry_object_get_property_names`
* `ecma_builtin_object_prototype_lookup_getter_setter`
* `ecma_op_function_has_instance`
* `ecma_op_function_get_super_constructor`
* `ecma_op_object_is_prototype_of`
* `ecma_op_object_enumerate`

Removed method `ecma_proxy_object_prototype_to_cp`

Fixes: #4445

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -1990,6 +1990,55 @@ ecma_op_object_has_instance (ecma_object_t *obj_p, /**< the object */
 } /* ecma_op_object_has_instance */
 
 /**
+ * General [[GetPrototypeOf]] abstract operation
+ *
+ * Note: returned valid object must be freed.
+ *
+ * @return ecma_object_t * - prototype of the input object.
+ *         ECMA_OBJECT_POINTER_ERROR - error reported during Proxy resolve.
+ *         NULL - the input object does not have a prototype.
+ */
+ecma_object_t *
+ecma_op_object_get_prototype_of (ecma_object_t *obj_p) /**< input object */
+{
+  JERRY_ASSERT (obj_p != NULL);
+
+#if ENABLED (JERRY_BUILTIN_PROXY)
+  if (ECMA_OBJECT_IS_PROXY (obj_p))
+  {
+    ecma_value_t proto = ecma_proxy_object_get_prototype_of (obj_p);
+
+    if (ECMA_IS_VALUE_ERROR (proto))
+    {
+      return ECMA_OBJECT_POINTER_ERROR;
+    }
+    if (ecma_is_value_null (proto))
+    {
+      return NULL;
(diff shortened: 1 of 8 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1977,6 +1977,51 @@ ecma_op_object_has_instance (ecma_object_t *obj_p, /**< the object */
...
+#if ENABLED (JERRY_BUILTIN_PROXY)
+  if (ECMA_OBJECT_IS_PROXY (obj_p))
+  {
+    ecma_value_t proto = ecma_proxy_object_get_prototype_of (obj_p);
+
+    if (ECMA_IS_VALUE_ERROR (proto))
+    {
+      return ECMA_OBJECT_POINTER_ERROR;
+    }
+    else if (ecma_is_value_null (proto))
```
rerobika, line 2003: I think a simple `if` is enough.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -1977,6 +1977,51 @@ ecma_op_object_has_instance (ecma_object_t *obj_p, /**< the object */
...
+    }
+
+    JERRY_ASSERT (ecma_is_value_object (proto));
+
+    return ecma_get_object_from_value (proto);
+  }
+  else
+#endif /* ENABLED (JERRY_BUILTIN_PROXY) */
+  {
+    ecma_object_t *proto = ECMA_GET_POINTER (ecma_object_t, ecma_op_ordinary_object_get_prototype_of (obj_p));
```
rerobika, line 2015: I see a double null check here, since `ECMA_GET_POINTER ` also checks for JMEM_CP_NULL. I suggest:

```c
jmem_cpointer_t proto_cp = ecma_op_ordinary_object_get_prototype_of (func_obj_p);

if (proto_cp == JMEM_CP_NULL)
{   
  return NULL;
}

ecma_object_t *proto_p = ECMA_GET_NON_NULL_POINTER (ecma_object_t, proto_cp);
ecma_ref_object (proto_p );
return proto_p;
```

rerobika (changes requested):

rerobika (approved):
LGTM

dbatyai (approved):
LGTM

Issue #4440, opened 2021-01-11
Assertion 'uint_ptr % JMEM_ALIGNMENT == 0' failed in jmem_decompress_pointer


owl337:
###### JerryScript revision
fdaacde

###### Build platform
Ubuntu 18.04.5 LTS(Linux 4.15.0-119-generic x86_64)

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g --strip=off \
--system-allocator=on --logging=on --linker-flag=-fuse-ld=gold \
--error-messages=on --profile=es2015-subset  
```
###### Test case
```
throw new Proxy(Function(), { get:function(rnd, array, weights) {
    var p = deRandom.deRandom_getFloat(rnd, [0.0, weightSum]);
    return array[0];

}});
```
###### Output
```
ICE: Assertion 'uint_ptr % JMEM_ALIGNMENT == 0' failed at /root/jerryscript/jerry-core/jmem/jmem-allocator.c(jmem_decompress_pointer):222.
Error: ERR_FAILED_INTERNAL_ASSERTION
Aborted

```

Credits: Found by chong from OWL337.

PR #4459, opened 2021-01-13
Fix error instance query if the error is from a Proxy

galpeter:
When accessing an error's type there is no need to check the Proxy
prototype (as there is none).

Fixes: #4440


rerobika (approved):
LGTM

robertsipka (approved):
LGTM

PR #4460, opened 2021-01-13
Fix objects foreach unittest in case of mem-stress-test

galpeter:
When the mem-stress-test is enabled for the objects foreach unittest
the object counting is off by one. This was due to incorrect assumption
on when the gc is triggered.


rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4461, opened 2021-01-13
Fix assert in ecma_op_object_own_property_keys

zherczeg:


LaszloLango (approved):
The fix LGTM, but more information about the issue in commit message would be nice.

rerobika (approved):
LGTM

PR #4462, opened 2021-01-13
Rework module parsing and execution

dbatyai:
This patch disables automatic detection of module code, and instead
requires the user to explicitly specifier whether to parse a source
as a module or as a script.

To achieve this the jerry_parse API function now takes a new option
which signals that the source should be parsed as a module.


jerry-core/api/jerry.c
```diff
@@ -465,16 +470,55 @@ jerry_parse (const jerry_char_t *resource_name_p, /**< resource name (usually a
...
+
+    if (ECMA_IS_VALUE_ERROR (ecma_module_parse_referenced_modules ()))
+    {
+      ecma_bytecode_deref (bytecode_data_p);
+      ecma_module_cleanup_context ();
+
+      return ecma_create_error_reference_from_context ();
+    }
+
+    ecma_object_t *obj_p = ecma_create_object (NULL, sizeof (ecma_extended_object_t), ECMA_OBJECT_TYPE_CLASS);
```
zherczeg, line 496: What happens if you execute this with postfix `()` in js?

tools/runners/run-test-suite.py
```diff
@@ -136,7 +138,13 @@ def run_normal_tests(args, tests):
...
         test_path = os.path.relpath(test)
         is_expected_to_fail = os.path.join(os.path.sep, 'fail', '') in test
-        (returncode, stdout) = execute_test_command(test_cmd + [test])
+
+        test_argument = [];
+        if test.endswith ('.mjs'):
+            test_argument.extend(['-m'])
+        test_argument.extend([test])
+
+        (returncode, stdout) = execute_test_command(test_cmd + test_argument)
```
zherczeg, line 147: `execute_test_command(test_cmd + test_argument + [test])` ?

jerry-core/api/jerry.c
dbatyai: You would get a TypeError, these objects are not callable. The only way to execute these is to pass them to jerry_run.

docs/02.API-REFERENCE.md
```diff
@@ -212,6 +212,7 @@ Option bits for [jerry_parse](#jerry_parse) and
 
  - JERRY_PARSE_NO_OPTS - no options passed
  - JERRY_PARSE_STRICT_MODE - enable strict mode
+ - JERRY_PARSE_MODULE - parse source as an ECMAScript module
```
galpeter, line 215: A note is required for the new option below, like :`*Changed in version[[NEXT_RELEASE]]* : ...`

jerry-core/api/jerry.c
```diff
@@ -465,16 +470,55 @@ jerry_parse (const jerry_char_t *resource_name_p, /**< resource name (usually a
...
+    }
+#endif /* ENABLED (JERRY_MODULE_SYSTEM) */
+
     return ecma_create_error_reference_from_context ();
   }
 
+#if ENABLED (JERRY_MODULE_SYSTEM)
+  if ((parse_opts & JERRY_PARSE_MODULE) != 0)
+  {
+    ecma_module_t *root_module_p = JERRY_CONTEXT (module_current_p);
```
galpeter, line 486: If there is an error below: shouldn't we do a cleanup on the `root_module_p`? or the cleanup_context below takes care of this?

jerry-core/api/jerry.c
```diff
@@ -588,15 +632,34 @@ jerry_run (const jerry_value_t func_val) /**< function to run */
     return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (wrong_args_msg_p)));
   }
 
-  ecma_object_t *func_obj_p = ecma_get_object_from_value (func_val);
+  ecma_object_t *obj_p = ecma_get_object_from_value (func_val);
+
+#if ENABLED (JERRY_MODULE_SYSTEM)
+  if (ecma_object_class_is (obj_p, LIT_MAGIC_STRING_RUNNABLE_UL))
```
galpeter, line 638: At the moment there the only runnable is a module object. However if the script parsing is adapted then we should also check the `.extra_info` here to see if it has the ECMA_RUNNABLE_FLAGS_MODULE flag. This check we can already put into this code. Maybe at least as an assert?

jerry-core/ecma/base/ecma-module.c
```diff
@@ -823,6 +885,46 @@ ecma_module_connect_imports (void)
...
+      ecma_value_t result = ecma_module_resolve_export (indirect_export_p->module_request_p,
+                                                        name_p->local_name_p,
+                                                        &record);
+
+      if (ECMA_IS_VALUE_ERROR (result))
+      {
+        return result;
+      }
+
+      if (record.module_p == NULL)
```
galpeter, line 914: At this point the `result` is an empty value? If so we could add an assert on it.

jerry-core/ecma/base/ecma-module.c
```diff
@@ -831,17 +933,42 @@ ecma_module_connect_imports (void)
...
   {
-    ret_value = ecma_module_check_indirect_exports ();
+    ret_value = ecma_module_check_indirect_exports (module_p);
   }
 
   return ret_value;
-} /* ecma_module_initialize_current */
+} /* ecma_module_initialize */
+
+static jerry_value_t ecma_module_parse (ecma_module_t *module_p);
```
galpeter, line 948: Hm... should the `ecma_module_parse` return `jerry_value_t`? Wouldn't `ecma_value_t` be better?

jerry-core/api/jerry.c
zherczeg: Can they be executed twice? If not, would it be better to execute them here?
When the referenced modules are executed?

dbatyai: The main module can be executed multiple times. The referenced modules are executed only once before the main module is first evaluated, and are cached for subsequent runs.
This could be changed so that the whole module tree is evaluated each time the main module is executed, but in that case we would need to keep the bytecode for the referenced modules, which we currently don't do to conserve memory.

dbatyai: Looking at it a bit, subsequent runs of the main module will (almost) always result in an error, since the imported variables will already exist, so it doesn't make a lot of sense to be able to run them multiple times. 

However, I would like to keep parsing and execution separate.

docs/02.API-REFERENCE.md
```diff
@@ -212,8 +212,10 @@ Option bits for [jerry_parse](#jerry_parse) and
...
 
 *New in version 2.0*.
+*Changed in version 2.4: Added JERRY_PARSE_MODULE.*
```
galpeter, line 218: Should be: 
```
*Changed in version 2.4*: Added `JERRY_PARSE_MODULE`.
```

galpeter (approved):
lgtm

rerobika (approved):
LGTM

PR #4465, opened 2021-01-15
Add a callback which is called when Error objects are created

zherczeg:


jerry-core/ecma/builtin-objects/ecma-builtin-helpers-error.c
```diff
@@ -53,14 +53,14 @@ ecma_builtin_helper_error_dispatch_call (ecma_standard_error_t error_type, /**<
       return ECMA_VALUE_ERROR;
     }
 
-    ecma_object_t *new_error_object_p = ecma_new_standard_error_with_message (error_type,
-                                                                              message_string_p);
+    ecma_object_t *new_error_object_p = ecma_new_standard_error (error_type,
```
rerobika, line 56: place the arguments into one line.

docs/02.API-REFERENCE.md
```diff
@@ -2994,6 +3020,47 @@ jerry_get_value_from_error (jerry_value_t value, bool release)
...
+- `callback` - callback function, the previously set value is overwritten, and setting NULL
+               disables the operation
+
+*New in version [[NEXT_RELEASE]]*.
+
+**Example**
+
+```c
+static void
+error_object_update_callback (const jerry_value_t error_object_t) /**< new error object */
```
rerobika, line 3045: Please drop the `_t` suffix from `error_object_t`

zherczeg:
Thanks for the review, patch is updated.

docs/02.API-REFERENCE.md
```diff
@@ -584,6 +584,32 @@ typedef void (*jerry_object_native_free_callback_t) (void *native_p);
...
+*Note*:
+ - The callback function cannot be called recursively, so the Error objects created
+   when the callback is running are not updated.
+
+*New in version [[NEXT_RELEASE]]*.
+
+**Prototype**
+
+```c
+typedef void (*jerry_error_object_update_callback_t) (const jerry_value_t error_object);
```
rerobika, line 604: To keep it consistent with vm_stop callback, shouldn't a `void* user_p` need to be passed to this callback?

docs/02.API-REFERENCE.md
```diff
@@ -2994,6 +3020,48 @@ jerry_get_value_from_error (jerry_value_t value, bool release)
...
+  (void) error_object;
+  printf ("Notification: a new error is created\n");
+}
+
+void main(void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_set_error_object_update_callback (error_object_update_callback);
+
```
rerobika, line 3056: I'd manually create a new error after the callback is set.

jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -140,23 +144,44 @@ ecma_new_standard_error (ecma_standard_error_t error_type) /**< native error typ
...
+    prop_value_p->value = ecma_make_string_value (message_string_p);
+  }
 
-  ecma_string_t *stack_str_p = ecma_new_ecma_string_from_utf8 ((const lit_utf8_byte_t *) stack_id_p, 5);
+  /* Avoid calling the decorator function recursively. */
+  if (JERRY_CONTEXT (error_object_update_callback_p) != NULL
+      && !(JERRY_CONTEXT (status_flags) & ECMA_STATUS_ERROR_UPDATE))
+  {
+    JERRY_CONTEXT (status_flags) |= ECMA_STATUS_ERROR_UPDATE;
+    JERRY_CONTEXT (error_object_update_callback_p) (ecma_make_object_value (new_error_obj_p));
```
rerobika, line 164: Are you sure that whenever the decorator callback is active, no `stack` property need to be assigned to the error object?

docs/02.API-REFERENCE.md
zherczeg: But that means extra data. Do we need it?

docs/02.API-REFERENCE.md
zherczeg: Manuallly?

jerry-core/ecma/operations/ecma-exceptions.c
zherczeg: It is up to the application which is using JerryScript. It can create such property in any way it wants.

docs/02.API-REFERENCE.md
rerobika: I meant to simply create a new error value.

rerobika (approved):
LGTM

jerry-core/api/jerry.c
```diff
@@ -1538,6 +1538,20 @@ jerry_get_value_from_error (jerry_value_t value, /**< api value */
   return ret_val;
 } /* jerry_get_value_from_error */
 
+/**
+ * Set new decorator callback for Error objects. The decorator can
+ * create or update any properties of the newly created Error object.
+ */
+void
+jerry_set_error_object_update_callback (jerry_error_object_update_callback_t callback, /**< new callback */
```
dbatyai, line 1546: One thing I don't really like is the name. Why do we call this an "update" callback, when it's called upon object creation?

zherczeg: The point is you can update (decorate) the created error object any way you want (e.g. add/remove/change properties). What name would you prefer?

dbatyai: I think the name should reference the event that triggers the callback, instead of what you can do with it. (I mean, technically you could do anything that is permissibly by the API, not just update the object.)
How about calling it `error_object_created_callback` or something similar?

zherczeg: Ok. Your suggestion is good. Patch is updated.

dbatyai (approved):
LGTM

PR #4472, opened 2021-01-15
logging the crash properly when running test262-harness.py

lygstate:
When running tests262 cases, the jerry program may exit with error code other than 1,
that's means a crash or JERRY_ASSERT triggered and should be resolved, so dump the test
name and stderr/stdout for tracking those crash in CI.

Depends on https://github.com/jerryscript-project/jerryscript/pull/4492

### DIFF
--- a/tools/runners/test262-harness.py
+++ b/tools/runners/test262-harness.py
@@ -434,6 +434,12 @@ class TestResult(object):
     def report_outcome(self, long_format):
         name = self.case.get_name()
         mode = self.case.get_mode()
+
+        if self.exit_code != 0 and self.exit_code != 1:
+            sys.stderr.write(u"===%s failed in %s with negative:%s===\n"
+                             % (name, mode, self.case.get_negative_type()))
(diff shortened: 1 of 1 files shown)


ossy-szeged:
I like the idea to track crashes, assertions, timeouts, etc. But I would be much stricter.
In my opinion crashes and assertions isn't acceptable at all, an engine shouldn't do it on any input.
There is only one exception: OOM is the expected results for TCO tests, since jerry isn't support tail call optimization.

But unfortunately there are crashes and assertions now, so first we should fix them, then we can make
the harness to interpret any assertion and crash as failure  (in case of negative test too).

I'm not against to have more verbose errors for these tests in test262.report file. But we should filter
them to have compact output of run-test --test262. Maybe we can extend the summary to count
assertions and crashes, and the user can check the test262.report to investigate the details.

dbatyai (approved):
LGTM

rerobika (approved):

Issue #4463, opened 2021-01-15
Fatal error detected by debug version with MSVC


lygstate:
###### JerryScript revision
df6d430289692ffa0922b108a7229ccf4502b79b

###### Build platform
```
PS C:\work\study\languages\typescript\jerryscript> python -c "import platform; print(platform.platform())"
Windows-10-10.0.19041-SP0
```

###### Build steps
```bat
python  tools/run-tests.py  --test262-esnext=update --build-debug 2>err.txt
```

###### Build log
exit_code:3221225477 means C0000005

https://docs.microsoft.com/en-us/previous-versions/troubleshoot/visualstudio/foxpro/c0000005-error-troubleshoot

```

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors', 'no-args', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors', 'no-args', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors', 'object-arg', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors', 'object-arg', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors', 'typedarray-arg', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors', 'typedarray-arg', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors-bigint', 'buffer-arg', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors-bigint', 'buffer-arg', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

WARNING:root:['built-ins', 'TypedArrayConstructors', 'ctors-bigint', 'length-arg', 'custom-proto-access-throws'] negative:False exit_code:3221225477 stdout: stderr:

[...]

PR #4475, opened 2021-01-15
Disable the debug popup on win32/msvc.

lygstate:
For not popup dialog when crash happend on MSVC/win32
related issue: https://github.com/jerryscript-project/jerryscript/issues/4463

### DIFF
--- a/jerry-main/main-unix.c
+++ b/jerry-main/main-unix.c
@@ -17,6 +17,10 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+#if defined (_WIN32)
+#include <crtdbg.h>
+#include <Windows.h>
+#endif
 
 #include "jerryscript.h"
 #include "jerryscript-ext/debugger.h"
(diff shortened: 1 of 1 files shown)


akosthekiss:
I'm not completely sure that console apps should be tweaked just because Windows doesn't let them crash without displaying graphical UI elements. Isn't it easier to simply disable that service on Windows? https://docs.microsoft.com/en-us/powershell/module/windowserrorreporting/disable-windowserrorreporting?view=win10-ps

rerobika:
I'd second @akosthekiss's suggestion and I can agree that it must not be implemented in the console app.

ossy-szeged:
Is it possible to disable these popup by runner script somehow? If we would like to run Windows testing ever on the CI, we will need to solve this issue somehow.

akosthekiss:
@ossy-szeged I've suggested (or rather requested feedback on) an approach but haven't heard back since.

lygstate:
> @ossy-szeged I've suggested (or rather requested feedback on) an approach but haven't heard back since.

disable service will affect other applicationsuch as UI

akosthekiss:
@lygstate You are free to re-enable the service after the execution of the automated tests.

lygstate:
> @lygstate You are free to re-enable the service after the execution of the automated tests.

Yeap, but it's a budren,

ossy-szeged:
> > @lygstate You are free to re-enable the service after the execution of the automated tests.
> 
> Yeap, but it's a budren,

+1 for @akosthekiss's suggestion. It's not a burden to modify the test runner script to disable the 
service before testing and enable the service after testing. It's obviously a testing feature/issue that's
why we don't want to modify jerry console application, but modify the testing script to reach your goal.

akosthekiss:
> > @lygstate You are free to re-enable the service after the execution of the automated tests.
> 
> Yeap, but it's a budren,

Well, it can be scripted, I guess. Then it is less of a burden. And I definitely prefer the "scripting burden" over cluttering code with Windows-specific debug-popup-blocking magic only because Windows effectively blocks automated testing.

lygstate:
> > > @lygstate You are free to re-enable the service after the execution of the automated tests.
> > 
> > 
> > Yeap, but it's a budren,
> 
> Well, it can be scripted, I guess. Then it is less of a burden. And I definitely prefer the "scripting burden" over cluttering code with Windows-specific debug-popup-blocking magic only because Windows effectively blocks automated testing.

No sure if have such a script, there code is also references to other implementation. I'l stick this locally as it works for me

PR #4476, opened 2021-01-15
Rename 'this' to 'this_arg' in function arguments

rerobika:
'this' is a restricted keyword in C++ so it's a good practice to avoid it's usage as an identifier.


akosthekiss (approved):
LGTM

galpeter (approved):
lgtm

Issue #4467, opened 2021-01-15
Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed in jmem_heap_finalize


owl337:
###### JerryScript revision
df6d430

###### Build platform
Ubuntu 18.04.5 LTS(Linux 4.15.0-119-generic x86_64)
###### Test case
```
var arr = new Array(10);
Date.parse((arr, "length", exception = new Proxy(((arr, "length", exception = new Proxy(Function(), (arr, "length", exception = new Proxy(Function(), { get:function(){ return function(){ return (Number.MAX_VALUE); }; }}))))), (arr, "length", exception = new Proxy(Function(), { get:function(){ return function(){ return (Number.MAX_VALUE); }; }})))));

```
###### Output
```
ICE: Assertion 'JERRY_CONTEXT (jmem_heap_allocated_size) == 0' failed at /root/jerryscript/jerry-core/jmem/jmem-heap.c(jmem_heap_finalize):107.
Error: ERR_FAILED_INTERNAL_ASSERTION
Aborted

```
Credits: Found by chong from OWL337.

Issue #4466, opened 2021-01-15
Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed in ecma_free_string_list


owl337:
###### JerryScript revision
df6d430

###### Build platform
Ubuntu 18.04.5 LTS(Linux 4.15.0-119-generic x86_64)

###### Build steps
```
./tools/build.py --clean --debug --compile-flag=-fsanitize=address \
--compile-flag=-m32 --compile-flag=-fno-omit-frame-pointer \
--compile-flag=-fno-common --compile-flag=-g --strip=off \
--system-allocator=on --logging=on --linker-flag=-fuse-ld=gold \
--error-messages=on --profile=es2015-subset  
```
###### Test case
```
var c = new Proxy(Function(), new Proxy(new Proxy(Function(), new Proxy(Function(), { get:function(){ return Proxy; }})), { get:function(){ return Proxy; }}))
var d = new Date();
c.__proto__ = d;

```
###### Output
```
ICE: Assertion 'ECMA_STRING_IS_REF_EQUALS_TO_ONE (string_p)' failed at /root/jerryscript/jerry-core/ecma/base/ecma-literal-storage.c(ecma_free_string_list):77.
Error: ERR_FAILED_INTERNAL_ASSERTION
Aborted

```
Credits: Found by chong from OWL337.

PR #4477, opened 2021-01-15
Correctly release values in Proxy.[[Get]]

galpeter:
In Proxy.[[Get]] if the target.[[GetOwnPropertyDescriptor]] fails the
trap result should be freed.

Fixes: #4466
Fixes: #4467


rerobika (approved):
LGTM

PR #4481, opened 2021-01-16
Trying to fixes module-resource-name.js win32

lygstate:
Try to fix the following testcases on win32:

```
[ 237/1146] FAIL (3221226505): tests\jerry\es.next\module-resource-name.js
================================================
Script Error: assertion failed
Script backtrace (top 5):
 0: C:\work\study\languages     ypescript\jerryscript   ests/jerry\es.next\module-resource-name.js:17

================================================

[summary] build\tests\jerry_tests-es.next\local\bin\jerry.exe tests\jerry

TOTAL: 1146
PASS: 1145
FAIL: 1

Success: 99%
```

See also https://github.com/lygstate/jerryscript/runs/1713796124?check_suite_focus=true
After revise default-module.c, I found the debugger'result path currently all to be relative path, not absolute path
That's not easy to locate the exact file place when debugging.

### DIFF
--- a/tests/jerry/es.next/module-resource-name.js
+++ b/tests/jerry/es.next/module-resource-name.js
@@ -14,14 +14,14 @@
 
 import { getName, getNamePromise } from "./module-resource-name-export.mjs"
 
-assert(getName().endsWith("/module-resource-name-export.mjs") === true);
+assert(getName().endsWith("module-resource-name-export.mjs"));
 
 var collector = {};
 getNamePromise(collector).then(() => { collector["end"] = resourceName(); });
 
 function __checkAsync() {
-  assert(collector["start"].endsWith("/module-resource-name-export.mjs") === true);
-  assert(collector["middle"].endsWith("/module-resource-name-export.mjs") === true);
-  assert(collector["end"].endsWith("/module-resource-name.js") === true);
+  assert(collector["start"].endsWith("module-resource-name-export.mjs"));
+  assert(collector["middle"].endsWith("module-resource-name-export.mjs"));
+  assert(collector["end"].endsWith("module-resource-name.js"));
   assert(Object.keys(collector).length === 3);
 }


jerry-port/default/default-module.c
```diff
@@ -150,7 +150,15 @@ jerry_port_normalize_path (const char *in_path_p,   /**< input file path */
...
   {
+    size_t i;
     ret = strnlen (norm_p, out_buf_size);
+    for (i = 0; i < ret; i += 1)
+    {
+      if (norm_p[i] == '\\')
+      {
+        norm_p[i] = '/';
+      }
+    }
```
akosthekiss, line 161: Why is this necessary? Back slash is still the primary directory separator on Windows. Forward slashes (the alternative direcory separator) are canonicalized to back slashes. So why do double work?

jerry-port/default/default-module.c
```diff
@@ -116,7 +116,7 @@ jerry_port_normalize_path (const char *in_path_p,   /**< input file path */
 {
   size_t ret = 0;
 
-#if defined (WIN32)
+#if defined (_WIN32)
```
akosthekiss, line 119: I suggest to have this change in a separate PR. Unfortunately, the code base does not use Windows-specific macros consistently. Preferrably a single commit should rewrite all `WIN32` occurrences to `_WIN32`.

akosthekiss (changes requested):

jerry-port/default/default-module.c
lygstate: For fixes tests on win32. Hard to fix

akosthekiss: Sorry, but this is no explanation. Please, provide example where this is necessary.

lygstate:
```tests\jerry\es.next\module-resource-name.js
import { getName, getNamePromise } from "./module-resource-name-export.mjs"

assert(getName().endsWith("/module-resource-name-export.mjs") === true);

var collector = {};
getNamePromise(collector).then(() => { collector["end"] = resourceName(); });

function __checkAsync() {
  assert(collector["start"].endsWith("/module-resource-name-export.mjs") === true);
  assert(collector["middle"].endsWith("/module-resource-name-export.mjs") === true);
  assert(collector["end"].endsWith("/module-resource-name.js") === true);
  assert(Object.keys(collector).length === 3);
}
```
This testcase want endes with "/.
That's unix style.

akosthekiss: But this is more of an error (invalid assumption) in the test case, not in the port module implementation.

lygstate: > But this is more of an error (invalid assumption) in the test case, not in the port module implementation.

OK, I'll modify the test instead

tests/jerry/es.next/module-resource-name.js
```diff
@@ -14,14 +14,19 @@
 
 import { getName, getNamePromise } from "./module-resource-name-export.mjs"
 
-assert(getName().endsWith("/module-resource-name-export.mjs") === true);
+function modulePathEndsWith(path, name)
+{
+  return path.endsWith('/' + name) || path.endsWith('\\' + name)
+}
+
+assert(modulePathEndsWith(getName(), "module-resource-name-export.mjs") === true);
```
akosthekiss, line 22: I think that this is an overcomplication. Wouldn't it be simply enough to remove the slashes from the paths? E.g.
`assert(getName().endsWith("module-resource-name-export.mjs") === true);`

BTW, `String.prototype.endsWith` returns a boolean. Is it necessary to compary it to `true` in an assert? I think that the `=== true` part could be dropped, that would be more natural.

rerobika (approved):
LGTM

akosthekiss (approved):
LGTM

PR #4482, opened 2021-01-16
Add FEATURE_INIT_FINI support on Windows.

lygstate:

### DIFF
--- a/docs/12.EXT-REFERENCE-MODULE.md
+++ b/docs/12.EXT-REFERENCE-MODULE.md
@@ -37,8 +37,8 @@ resolved using the native JerryScript module resolver `jerryx_module_native_reso
 `jerryx_module_resolve()`. Native modules are registered during application startup and by calling `dlopen()` by means
 of library constructors, support for which can be turned on using the `FEATURE_INIT_FINI` build flag. In the absence of
 such a flag, the module registration and unregistration functions are exposed as global symbols which can be called
-explicitly. Note: `FEATURE_INIT_FINI` build flag isn't supported on Windows, because Microsoft Visual C/C++ Compiler
-doesn't support library constructors and destructors.
+explicitly. Note: On windows, `FEATURE_INIT_FINI` build flag only supported with GNU toolchain or Microsoft Visual C/C++ Compiler
+2008 and upper.
 
 ## jerryx_module_resolve
 
--- a/jerry-ext/include/jerryscript-ext/module.h
+++ b/jerry-ext/include/jerryscript-ext/module.h
@@ -42,16 +42,56 @@ typedef struct jerryx_native_module_t
  */
 #ifdef ENABLE_INIT_FINI
 #ifdef _MSC_VER
-#error "`FEATURE_INIT_FINI` build flag isn't supported on Windows, because Microsoft Visual C/C++ Compiler \
-doesn't support library constructors and destructors."
+/**
+ * Only Visual Studio 2008 and upper version support for __pragma keyword
+ * refer to https://msdn.microsoft.com/en-us/library/d9x1s805(v=vs.90).aspx
+ */
+#if _MSC_VER >= 1500
+#ifdef _WIN64
+#define JERRYX_MSVC_INCLUDE_SYM(s) comment (linker, "/include:" #s)
+#else /* !_WIN64 */
+#define JERRYX_MSVC_INCLUDE_SYM(s) comment (linker, "/include:_" #s)
+#endif /* _WIN64 */
+
+#ifdef __cplusplus
+#define JERRYX_MSCV_EXTERN_C extern "C"
+#else /* !__cplusplus */
+#define JERRYX_MSCV_EXTERN_C
+#endif /* __cplusplus */
+
+#pragma section(".CRT$XCU", read)
+#pragma section(".CRT$XTU", read)
+
+#define JERRYX_MSVC_FUNCTION_ON_SECTION(sec_name, f)                               \
+  static void f (void);                                                            \
+  __pragma (JERRYX_MSVC_INCLUDE_SYM (f##_section)) __declspec(allocate (sec_name)) \
+    JERRYX_MSCV_EXTERN_C void (*f##_section) (void) = f;                           \
+  static void f (void)
+
+#define JERRYX_MODULE_CONSTRUCTOR(f) JERRYX_MSVC_FUNCTION_ON_SECTION (".CRT$XCU", f)
+#define JERRYX_MODULE_DESTRUCTOR(f)  JERRYX_MSVC_FUNCTION_ON_SECTION (".CRT$XTU", f)
+#else /* !(_MSC_VER >= 1500) */
+#error "Only Visual Studio 2008 and upper version are supported."
+#endif /* _MSC_VER >= 1500 */
+#elif defined(__GNUC__)
+#define JERRYX_MODULE_CONSTRUCTOR(f)                  \
+  static void f (void) __attribute__ ((constructor)); \
+  static void f (void)
+
+#define JERRYX_MODULE_DESTRUCTOR(f)                  \
+  static void f (void) __attribute__ ((destructor)); \
+  static void f (void)
+#else /* __GNUC__ */
+#error "`FEATURE_INIT_FINI` build flag isn't supported on this compiler"
 #endif /* _MSC_VER */
-#define JERRYX_MODULE_CONSTRUCTOR_ATTRIBUTE  __attribute__ ((constructor))
-#define JERRYX_MODULE_DESTRUCTOR_ATTRIBUTE   __attribute__ ((destructor))
-#define JERRYX_MODULE_REGISTRATION_QUALIFIER static
 #else /* !ENABLE_INIT_FINI */
-#define JERRYX_MODULE_CONSTRUCTOR_ATTRIBUTE
-#define JERRYX_MODULE_DESTRUCTOR_ATTRIBUTE
-#define JERRYX_MODULE_REGISTRATION_QUALIFIER
+#define JERRYX_MODULE_CONSTRUCTOR(f) \
+  void f (void);                     \
+  void f (void)
+
+#define JERRYX_MODULE_DESTRUCTOR(f) \
+  void f (void);                    \
+  void f (void)
 #endif /* ENABLE_INIT_FINI */
 
 /**
@@ -59,21 +99,19 @@ doesn't support library constructors and destructors."
  */
 #define JERRYX_NATIVE_MODULE(module_name, on_resolve_cb) JERRYX_NATIVE_MODULE_IMPLEM (module_name, on_resolve_cb)
 
-#define JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb)                                                 \
-  static jerryx_native_module_t _##module_name##_definition = { .name_p = (jerry_char_t *) #module_name,        \
-                                                                .on_resolve_p = (on_resolve_cb),                \
-                                                                .next_p = NULL };                               \
-                                                                                                                \
-  JERRYX_MODULE_REGISTRATION_QUALIFIER void module_name##_register (void) JERRYX_MODULE_CONSTRUCTOR_ATTRIBUTE;  \
-  JERRYX_MODULE_REGISTRATION_QUALIFIER void module_name##_register (void)                                       \
-  {                                                                                                             \
-    jerryx_native_module_register (&_##module_name##_definition);                                               \
-  }                                                                                                             \
-                                                                                                                \
-  JERRYX_MODULE_REGISTRATION_QUALIFIER void module_name##_unregister (void) JERRYX_MODULE_DESTRUCTOR_ATTRIBUTE; \
-  JERRYX_MODULE_REGISTRATION_QUALIFIER void module_name##_unregister (void)                                     \
-  {                                                                                                             \
-    jerryx_native_module_unregister (&_##module_name##_definition);                                             \
+#define JERRYX_NATIVE_MODULE_IMPLEM(module_name, on_resolve_cb)                                          \
+  static jerryx_native_module_t _##module_name##_definition = { .name_p = (jerry_char_t *) #module_name, \
+                                                                .on_resolve_p = (on_resolve_cb),         \
+                                                                .next_p = NULL };                        \
+                                                                                                         \
+  JERRYX_MODULE_CONSTRUCTOR (module_name##_register)                                                     \
+  {                                                                                                      \
+    jerryx_native_module_register (&_##module_name##_definition);                                        \
+  }                                                                                                      \
+                                                                                                         \
+  JERRYX_MODULE_DESTRUCTOR (module_name##_unregister)                                                    \
+  {                                                                                                      \
+    jerryx_native_module_unregister (&_##module_name##_definition);                                      \
   }
 
 /**
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -68,10 +68,7 @@ JERRY_UNITTESTS_OPTIONS = [
             OPTIONS_COMMON + OPTIONS_DOCTESTS + OPTIONS_PROFILE_ES51),
     Options('unittests-es5.1-init-fini',
             OPTIONS_COMMON + OPTIONS_UNITTESTS + OPTIONS_PROFILE_ES51
-            + ['--cmake-param=-DFEATURE_INIT_FINI=ON'],
-            skip=skip_if((sys.platform == 'win32'), 'FEATURE_INIT_FINI build flag isn\'t supported on Windows,' +
-                         ' because Microsoft Visual C/C++ Compiler doesn\'t support' +
-                         ' library constructors and destructors.')),
+            + ['--cmake-param=-DFEATURE_INIT_FINI=ON']),
     Options('unittests-es5.1-math',
             OPTIONS_COMMON + OPTIONS_UNITTESTS + OPTIONS_PROFILE_ES51
             + ['--jerry-math=on']),


tools/run-tests.py
```diff
@@ -68,10 +68,7 @@ def skip_if(condition, desc):
             OPTIONS_COMMON + OPTIONS_DOCTESTS + OPTIONS_PROFILE_ES51 + OPTIONS_DEBUG),
     Options('unittests-es5.1-debug-init-fini',
             OPTIONS_COMMON + OPTIONS_UNITTESTS + OPTIONS_PROFILE_ES51 + OPTIONS_DEBUG
-            + ['--cmake-param=-DFEATURE_INIT_FINI=ON'],
-            skip=skip_if((sys.platform == 'win32'), 'FEATURE_INIT_FINI build flag isn\'t supported on Windows,' +
-                         ' because Microsoft Visual C/C++ Compiler doesn\'t support' +
-                         ' library constructors and destructors.')),
+            + ['--cmake-param=-DFEATURE_INIT_FINI=ON']),
```
rerobika, line 71: Can you clarify somehow that the build and the related unit test is passing?

docs/12.EXT-REFERENCE-MODULE.md
```diff
@@ -37,8 +37,7 @@ resolved using the native JerryScript module resolver `jerryx_module_native_reso
 `jerryx_module_resolve()`. Native modules are registered during application startup and by calling `dlopen()` by means
 of library constructors, support for which can be turned on using the `FEATURE_INIT_FINI` build flag. In the absence of
 such a flag, the module registration and unregistration functions are exposed as global symbols which can be called
-explicitly. Note: `FEATURE_INIT_FINI` build flag isn't supported on Windows, because Microsoft Visual C/C++ Compiler
-doesn't support library constructors and destructors.
+explicitly.
```
akosthekiss, line 40: This is not correct, according to this change. FEATURE_INIT_FINI is still not supported on MSVC 2005 and older.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -18,6 +18,12 @@
 
 #include "jerryscript.h"
 
+#ifdef __cplusplus
+#define JERRYX_EXTERN_C extern "C"
+#else
+#define JERRYX_EXTERN_C
+#endif
```
akosthekiss, line 25: This does not look right here. No.1: it is confusing to have this here right in front of another `#ifdef __cplusplus`. No.2: This is only used by MSVC, so it should be added to the MSVC-specific block, if it is indeed necessary.

jerry-ext/include/jerryscript-ext/module.h
```diff
@@ -45,15 +51,59 @@ typedef struct jerryx_native_module_t
...
+  }                                                \
+  JERRYX_MSVC_DECL_SECTION_FUNCTION_POINTER(".CRT$XCU", construct##_constructor)
+
+#define JERRYX_MODULE_DESTRUCTOR(destruct)       \
+  static int __cdecl destruct##_destructor(void) \
+  {                                              \
+    destruct();                                  \
+    return 0;                                    \
+  }                                              \
+  JERRYX_MSVC_DECL_SECTION_FUNCTION_POINTER(".CRT$XTU", destruct##_destructor)
```
akosthekiss, line 88: I'm not sure about these trampoline functions. Quite recently I had to work on code that gets auto-executed on load, and managed to get it work on Windows/MSVC, too, with a similar but simpler approach:

```c
#ifdef _MSC_VER
  #pragma section(".CRT$XCU", read)
  #define JERRYX_MODULE_CONSTRUCTOR2(f, prefix) \
    static void f (void); \
    __declspec(allocate(".CRT$XCU")) void (*f##_) (void) = f; \
    __pragma(comment(linker, "/include:" prefix #f "_")) \
    static void f (void)
  #ifdef _WIN64
    #define JERRYX_MODULE_CONSTRUCTOR(f) JERRYX_MODULE_CONSTRUCTOR2 (f, "")
  #else
    #define JERRYX_MODULE_CONSTRUCTOR(f) JERRYX_MODULE_CONSTRUCTOR2 (f, "_")
  #endif
#else
  #define JERRYX_MODULE_CONSTRUCTOR(f) \
    static void f (void) __attribute__((constructor)); \
    static void f (void)
#endif
```

Notes: Macro names have been adapted to the current scenario, but the pasted code is not feature-complete. E.g., it does not deal with destructors (but should be easy to adapt further). Static is also hard-coded. Etc.

akosthekiss (changes requested):

tools/run-tests.py
lygstate: It's currently just the build flags are supported, didn't related to the unit testing yet.

jerry-ext/include/jerryscript-ext/module.h
lygstate: Yes, it's necessary, as without exntern "C", C++ code include this module.h would take no effect

jerry-ext/include/jerryscript-ext/module.h
lygstate: > I'm not sure about these trampoline functions. Quite recently I had to work on code that gets auto-executed on load, and managed to get it work on Windows/MSVC, too, with a similar but simpler approach:
> 
> ```c
> #ifdef _MSC_VER
>   #pragma section(".CRT$XCU", read)
>   #define JERRYX_MODULE_CONSTRUCTOR2(f, prefix) \
>     static void f (void); \
>     __declspec(allocate(".CRT$XCU")) void (*f##_) (void) = f; \
>     __pragma(comment(linker, "/include:" prefix #f "_")) \
>     static void f (void)
>   #ifdef _WIN64
>     #define JERRYX_MODULE_CONSTRUCTOR(f) JERRYX_MODULE_CONSTRUCTOR2 (f, "")
>   #else
>     #define JERRYX_MODULE_CONSTRUCTOR(f) JERRYX_MODULE_CONSTRUCTOR2 (f, "_")
>   #endif
> #else
>   #define JERRYX_MODULE_CONSTRUCTOR(f) \
>     static void f (void) __attribute__((constructor)); \
>     static void f (void)
> #endif
> ```
> 
> Notes: Macro names have been adapted to the current scenario, but the pasted code is not feature-complete. E.g., it does not deal with destructors (but should be easy to adapt further). Static is also hard-coded. Etc.

I've seen you code, but have mistake in this code, The #pragma section(".CRT$XCU", read) should not used unconditionally, should also convert to use __param(section(".CRT$XCU", read))
and also it's didn't consider the amd64 and x86 differencies in
```
    __pragma(comment(linker, "/include:" prefix #f "_")) \

```

akosthekiss: The #pragma simply defines a section. What problem does it cause if it is left empty? But it can be converted to __pragma().

As for the amd64/x86 differences: if you mean the underscore prefix necessary on x86, it is there. If you mean something else, please, specify.

And, as mentioned below the code snippet, this is not a fully working suggestion but an example only that may need further adaptation. Feel free to adapt. The key point of the comment was to get rid of the trampolines.

lygstate: I am talking about  #pragma section(".CRT$XCU", read), without   __pragma (data_seg (push))  
 __pragma (data_seg (pop))

It's will affect other section definition.

The  underscore prefix necessary on x86, but should not on x64 that's why macro JERRYX_MSVC_INCLUDE_SYM defined

lygstate: As Igor says, #ifdef is a compile-time check.  But if you really do want a compile-time check, you don’t want to use _WIN64 because that is not not automatically defined when setting up the project configuration.  Instead, use _M_AMD64.  http://msdn.microsoft.com/en-us/library/b0084kay.aspx

akosthekiss: There is only a single variable declaration/initialization between the segment push/pop pairs, and that variable has an explicit allocation declspec. Please, explain in detail why the push/pop is needed then.

I'm fine with _M_AMD64, but as I recall from the recent jerry-math PR, we have also used _M_X64 as well.

(And who is Igor? I'm losing context)

lygstate: Ignore Igor, copied from blog, I am not sure if we need use _WIN64 or _M_AMD64, I don't know what's the C ABI of ARM for msvc and didn't find the document for it.
But the push/pop are necessary.

#ifdef _MSC_VER
  #pragma section(".CRT$XCU", read)
Means the following code are setting ".CRT$XCU" to read

but after this, anyone can modify the ".CRT$XCU" attribute.
that's why `__pragma (section (sec_name, read))` needed.
As between the module.h and the code you define the constructor can have any kinds of code.

the pop and push is a standard way to guard the __pragma (section) not pollute other code. because in C++
You won't know where the macro are called.
And what's kinds of code after the macro

akosthekiss: > Ignore Igor, copied from blog

How about not copying it then...

> I am not sure if we need use _WIN64 or _M_AMD64, I don't know what's the C ABI of ARM for msvc and didn't find the document for it.

I was not referring to ARM at all. I was suggesting to use `#if defined (_M_AMD64) || defined (_M_X64)` (just like it is done at https://github.com/jerryscript-project/jerryscript/blob/master/jerry-math/jerry-math-internal.h#L41 )

> But the push/pop are necessary.
>
> #ifdef _MSC_VER
> #pragma section(".CRT$XCU", read)
> Means the following code are setting ".CRT$XCU" to read
>
> but after this, anyone can modify the ".CRT$XCU" attribute.
> that's why __pragma (section (sec_name, read)) needed.
> As between the module.h and the code you define the constructor can have any kinds of code.
> 
> the pop and push is a standard way to guard the __pragma (section) not pollute other code. 

Unfortunately, I'm not sure I can follow your argument.

The `section` pragma does _not_ change where following code or data is placed. See: "Creates a section in an OBJ file. [...] Once a section is defined, it remains valid for the rest of the compilation. However, you must use __declspec(allocate), or nothing is placed in the section." (https://docs.microsoft.com/en-us/cpp/preprocessor/section?view=msvc-160)

This means that in the code snippet below, between the push and pop, there is a section definition that does not affect how following data or code is placed, and a variable declaration that is explicitly allocated in a given section.

```c
  __pragma (data_seg (push))                                              \
      __pragma (section (sec_name, read)) __declspec(allocate (sec_name)) \
          JERRYX_MSCV_EXTERN_C jerryx_msvc_section_function_t             \
              func_name##_section = func_name;                            \
  __pragma (data_seg (pop))                                               \
```

 According to my understanding this makes the push/pop useless. Please, explain where I'm wrong.

> because in C You won't know where the macro are called.

I think I have some basic knowledge of C, but thanks for the heads up.

lygstate: Ah, sorry for that, you mean  __pragma (section (sec_name, read)) only take effect on `func_name##_section`, that's make sense
I forgot that may code was compatiable to MSVC 2005 and below but now it's unnecessary anymore. I'll update the patch.

Thanks for the review

lygstate: This is the MSVC demo:
```
#define CATCH_ALL_DUAL \
CATCH(COleException, e) \
{ \
_hr = e->m_sc; \
} \
AND_CATCH_ALL(e) \
{ \
__pragma(warning(push)) \
__pragma(warning(disable:6246)) /*disable _ctlState prefast warning*/ \
AFX_MANAGE_STATE(pThis->m_pModuleState); \
__pragma(warning(pop)) \
_hr = DualHandleException(_riidSource, e); \
} \
END_CATCH_ALL \
return _hr; \
```
Please taka a look at https://docs.microsoft.com/en-us/cpp/preprocessor/pragma-directives-and-the-pragma-keyword?view=msvc-160

akosthekiss: I'm sorry, but I don't understand why you are copying & linking this here. There was no discussion about warnings here, nor about exceptions.

lygstate: > I'm sorry, but I don't understand why you are copying & linking this here. There was no discussion about warnings here, nor about exceptions.

I am talking about __pragma(warning(push)) , __pragma(warning(pop)) 
For section it's the same.
Please note, it's msvc extention __pragma, not C99 _Pragma

akosthekiss: I totally understand how a push/pop mechanism works: for warnings, for data section, or -- say -- for machine stack. What I don't understand is its relevance here. I've given a detailed explanation why I think that section pragma does not have to be guarded by a data_seg push/pops (because section pragma does not change where following data is placed, it only creates a section for future reference by allocate declspec).

For warnings, it is pretty much different, because a warning(disable:N) indeed has effect on what follows. That's why push/pop guards are needed, to save/restore state.

It seemed that you agreed because there was a comment about that -- which is now gone, most probably deleted. I've recovered it from my mail archive:

> Ah, sorry for that, you mean __pragma (section (sec_name, read)) only take effect on func_name##_section, that's make sense
> I forgot that may code was compatiable to MSVC 2005 and below but now it's unnecessary anymore. I'll update the patch.
>
> Thanks for the review

I don't think that erasing / rewriting history is a good approach and does not help trust.

Plus, you are communicating in half sentences and/or without context:

> I am talking about __pragma(warning(push)) , __pragma(warning(pop))

No, you were not talking about that, you just copy-pasted a code snippet and a link without any explanation why you though it would be relevant. I'm not good at reading minds -- and I'm not eager to learn it either.

I think I'll suspend communication and review until attitude changes for the better.

lygstate:
Too hard to following you altitude, if you altitude is you are always right, then I have no idea to follow

akosthekiss:
Well, well. You still haven't answered some open questions, have you?
1. Why did you delete your comment?
2. Based on what do you state that section pragma changes where all following variables are placed?

I'm happy to stand corrected if you have explanations to these questions.

(My altitude is cca. 76 m above Baltic see level.)

lygstate:
> Well, well. You still haven't answered some open questions, have you?
> 
> 1. Why did you delete your comment?
That comment is that I am tring to convice me I was wrong( and I am conviced), but after carefully reading the meaning of __pragma, It's turn out that my usage of __pragma is correct.
> 2. Based on what do you state that section pragma changes where all following variables are placed?
I  don't know what'ts the meaning of  this question, please forgive my poor English.
But at least, my code is valid, maybe it's redudant but valid, about your code, I am trying to digging Microsoft's document, didn't found anything to state it's valid and invalid.

> 
> I'm happy to stand corrected if you have explanations to these questions.
> 
> (My altitude is cca. 76 m above Baltic ~see~ Sea level.)

akosthekiss:
> > 1. Why did you delete your comment?
> 
> That comment is that I am tring to convice me I was wrong( and I am conviced), but after carefully reading the meaning of __pragma, It's turn out that my usage of __pragma is correct.

That does not justify deleting an earlier comment. A previously discussed topic can be reiterated later and previous statements can be reviewed and corrected. But hiding/removing them sends a bad message.

> > 2. Based on what do you state that section pragma changes where all following variables are placed?
> 
> I don't know what'ts the meaning of this question, please forgive my poor English.
> But at least, my code is valid, maybe it's redudant but valid, about your code, I am trying to digging Microsoft's document, didn't found anything to state it's valid and invalid.

Let me rephrase: Why do you state that `#pragma section()` (or `__pragma(section())`, or `_Pragma(section())`) changes where a variable is placed? Why do you state that after such a pragma, all variables are put in the new section?

The documentation of MSVC is quite explicit, and even has an example. Copied below:

```c
#pragma section("mysec",read,write)
int j = 0;

__declspec(allocate("mysec"))
int i = 0;

int main(){}
```

The discussion of the example goes like: "The integer j isn't put into mysec because it wasn't declared using __declspec(allocate). Instead, j goes into the data section. The integer i does go into mysec because of its __declspec(allocate) storage-class attribute."

(Source: https://docs.microsoft.com/en-us/cpp/preprocessor/section?view=msvc-160)

I have never stated that the code in the PR was invalid, but I stated that I've been able to work with a _simpler_ approach. (Actually, it was you who stated that there is a mistake in my code.) A PR review is about quality control. It does not only mean preventing bugs but also preventing superfluous code leaking in. If there is redundancy in the code, it should be pruned away. Superfluous / redundant / useless code is bad for the maintainers.

lygstate:
> > > 1. Why did you delete your comment?
> > 
> > 
> > That comment is that I am tring to convice me I was wrong( and I am conviced), but after carefully reading the meaning of __pragma, It's turn out that my usage of __pragma is correct.
> 
> That does not justify deleting an earlier comment. A previously discussed topic can be reiterated later and previous statements can be reviewed and corrected. But hiding/removing them sends a bad message.
> 
> > > 1. Based on what do you state that section pragma changes where all following variables are placed?
> > 
> > 
> > I don't know what'ts the meaning of this question, please forgive my poor English.
> > But at least, my code is valid, maybe it's redudant but valid, about your code, I am trying to digging Microsoft's document, didn't found anything to state it's valid and invalid.
> 
> Let me rephrase: Why do you state that `#pragma section()` (or `__pragma(section())`, or `_Pragma(section())`) changes where a variable is placed? Why do you state that after such a pragma, all variables are put in the new section?
> 
> The documentation of MSVC is quite explicit, and even has an example. Copied below:
> 
> ```c
> #pragma section("mysec",read,write)
> int j = 0;
> 
> __declspec(allocate("mysec"))
> int i = 0;
> 
> int main(){}
> ```

Yes, I've reading the code, I am worried about this demo
```
#pragma section("mysec",read)

#pragma data_seg (push)
#pragma section("mysec",read,write)
int j = 0;

__declspec(allocate("mysec"))
int i = 0;


#pragma data_seg(pop)

__declspec(allocate("mysec"))
int k = 0;

int main(){}

```

And now I tried compile it, the following error happens:
```
C:\work\study\languages\typescript\iotjs>cl test.c
Microsoft (R) C/C++ Optimizing Compiler Version 19.28.29335 for x86
Copyright (C) Microsoft Corporation.  All rights reserved.

test.c
test.c(4): error C3415: multiple 'mysec' sections found with different attributes ('0xc0000040')
test.c(1): note: see declaration of 'mysec'
```

So you were right, the section can only appear in way one. and can appear multiple times.




> 
> The discussion of the example goes like: "The integer j isn't put into mysec because it wasn't declared using __declspec(allocate). Instead, j goes into the data section. The integer i does go into mysec because of its __declspec(allocate) storage-class attribute."
> 
> (Source: https://docs.microsoft.com/en-us/cpp/preprocessor/section?view=msvc-160)
> 
> I have never stated that the code in the PR was invalid, but I stated that I've been able to work with a _simpler_ approach. (Actually, it was you who stated that there is a mistake in my code.) A PR review is about quality control. It does not only mean preventing bugs but also preventing superfluous code leaking in. If there is redundancy in the code, it should be pruned away. Superfluous / redundant / useless code is bad for the maintainers.

#

PR #4483, opened 2021-01-16
Use approxEq to judge double compare in math-cbrt.js

lygstate:

### DIFF
--- a/tests/jerry/es.next/math-cbrt.js
+++ b/tests/jerry/es.next/math-cbrt.js
@@ -20,6 +20,11 @@ function isSameZero (x, y)
   return x === 0 && (1 / x) === (1 / y);
 }
 
+function approxEq (x, y)
+{
+  return Math.abs(x - y) <= Number.EPSILON * 2
+}
+
 assert(isNaN(Math.cbrt(NaN)));
 assert(isSameZero(Math.cbrt(p_zero), p_zero));
 assert(isSameZero(Math.cbrt(n_zero), n_zero));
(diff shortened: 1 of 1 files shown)


tests/jerry/es.next/math-cbrt.js
```diff
@@ -20,6 +20,11 @@ function isSameZero (x, y)
   return x === 0 && (1 / x) === (1 / y);
 }
 
+function isSameDouble (x, y)
```
akosthekiss, line 23: I suggest to use a different name for this function. Perhaps `approxEq` is the shortest name that could be used. It uses "accepted" abbreviations, describes what comparison it performs, and does not conflict with English grammar. (`isSameZero` isn't really a good pattern to follow, unfortunately.)

akosthekiss (changes requested):
Will be OK technically after a terminology fix.

tests/jerry/es.next/math-cbrt.js
lygstate: To follow function

function isSameZero (x, y)
{
  return x === 0 && (1 / x) === (1 / y);
}

akosthekiss (approved):
LGTM

akosthekiss (changes requested):
Ah, late hit. Missed the commit message. Please update that.

akosthekiss (approved):
LGTM

tests/jerry/es.next/math-cbrt.js
```diff
@@ -20,6 +20,11 @@ function isSameZero (x, y)
   return x === 0 && (1 / x) === (1 / y);
 }
 
+function approxEq (x, y)
+{
+  return Math.abs(x - y) < 1e-15
```
rerobika, line 25: How about using `Number.EPSILON` instead of a handcrafted one?

rerobika (changes requested):

tests/jerry/es.next/math-cbrt.js
lygstate: Not sure, 1e-16 failed before.
Let's me have a try.
The EPSILON property has a value of approximately 2.2204460492503130808472633361816E-16

lygstate: [ 212/1145] FAIL (-6): tests/jerry/es.next/math-cbrt.js
================================================
Script Error: assertion failed

================================================

[summary] build/tests/jerry_tests-es.next-debug/local/bin/jerry tests/jerry

TOTAL: 1145
PASS: 1144
FAIL: 1

Success: 99%

Not working, how about Number.EPSILON * 10? or other number

rerobika (approved):
LGTM

PR #4488, opened 2021-01-16
Unify asserts in jerry-ext

akosthekiss:
Some components of the jerry-ext library re-declared their own
assert and/or static assert macros. No need to re-invent the wheel
all the time. This patch makes all components use the assert macros
defined in jext-common.h. This also makes maintenance easier.

### DIFF
--- a/jerry-ext/handle-scope/handle-scope.c
+++ b/jerry-ext/handle-scope/handle-scope.c
@@ -15,6 +15,9 @@
 
 #include <stdlib.h>
 #include "handle-scope-internal.h"
+#include "jext-common.h"
+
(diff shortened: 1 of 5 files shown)


rerobika (approved):
LGTM

jerry-ext/handle-scope/handle-scope.c
```diff
@@ -15,6 +15,9 @@
 
 #include <stdlib.h>
 #include "handle-scope-internal.h"
+#include "jext-common.h"
+
+JERRYX_STATIC_ASSERT (JERRYX_SCOPE_PRELIST_SIZE < 32, JERRYX_SCOPE_PRELIST_SIZE_must_be_less_than_size_of_uint8_t);
```
zherczeg, line 20: These should be all capital letters.

akosthekiss: This was copy-pasted here from existing code. But updated now.

zherczeg (approved):
LGTM

PR #4495, opened 2021-01-17
Move get_platform_cmd_prefix into util.py

lygstate:
get_platform_cmd_prefix implement multiple times in multiple function.
Also add function get_python_cmd_prefix, call to python script always
add python command so that on windows, the python script doesn't have
executable permission also can be called.

Now it's can be reviewed independently

### DIFF
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -25,6 +25,12 @@ import subprocess
 import sys
 import settings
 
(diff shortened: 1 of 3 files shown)


dbatyai (approved):
LGTM

rerobika (approved):

PR #4497, opened 2021-01-17
Get isfinite, isinf, and isnan right in jerry-math

akosthekiss:
- `finite` is not C99 but a BSD fp classification function, so it
  is removed.
- The standard specifies that `isnan` is a macro (just like
  `isfinite` and `isinf`), so the `isnan` function implementation
  is removed.
- `isfinite` incorrectly classified NAN as finite, which is fixed.
- `isinf` returned 1 for negative infinity. This is not a bug
  according to the standard, but is not aligned with recent glibc,
  which returns -1. This is changed to simplify testing.
- Added test cases for `isfinite` and `isinf` to the unit test of
  jerry-math.
- Added a new pass to unittests to ensure that jerry-math is
  tested.


rerobika (approved):
LGTM

dbatyai (approved):
LGTM

PR #4501, opened 2021-01-18
Waiting the remaining tcp data are sent

lygstate:
So that the tcp socket are closed by debugger client first and
the debugger client won't receive socket are closed by remote error
```
The error message:
Test command:
cmd \
	/S \
	/C \
	C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe \
	D:\a\jerryscript\jerryscript\tools\runners/run-debugger-test.py \
	D:\a\jerryscript\jerryscript\build\tests\jerry_debugger_tests\local\bin\jerry.exe \
	websocket \
	D:\a\jerryscript\jerryscript\jerry-debugger/jerry_client.py \
	tests\debugger\do_abort
run debug server: D:\a\jerryscript\jerryscript\build\tests\jerry_debugger_tests\local\bin\jerry.exe tests\debugger\do_abort.js --start-debug-server --debug-channel websocket
run debug client: cmd /S /C C:\hostedtoolcache\windows\Python\3.7.9\x64\python.exe D:\a\jerryscript\jerryscript\jerry-debugger/jerry_client.py --channel websocket --non-interactive
input debug cmd: tests\debugger\do_abort.cmd
['git', '--no-pager', 'diff', '--ignore-space-at-eol', '--no-index', 'tests\\debugger\\do_abort.expected', 'C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\do_abort5k9rpkyiout']
FAIL: tests\debugger\do_abort
warning: LF will be replaced by CRLF in C:\Users\RUNNER~1\AppData\Local\Temp\do_abort5k9rpkyiout.
The file will have its original line endings in your working directory
diff --git "a/tests\\debugger\\do_abort.expected" "b/C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\do_abort5k9rpkyiout"
index 7656e41..8d31910 100644
--- "a/tests\\debugger\\do_abort.expected"
+++ "b/C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\do_abort5k9rpkyiout"
@@ -10,3 +10,5 @@ Stopped at tests/debugger/do_abort.js:20 (in g() at line:19, col:1)
 Stopped at tests/debugger/do_abort.js:16 (in f() at line:15, col:1)
 (jerry-debugger) abort new Error('Fatal error :)')
 err: Error: Fatal error :)
+Failed to connect to the JerryScript debugger.
+Error: [WinError 10054] An existing connection was forcibly closed by the remote host
```

This is the random failure I am talking about:
https://github.com/jerryscript-project/jerryscript/runs/1721915835


jerry-ext/debugger/debugger-tcp.c
```diff
@@ -131,6 +131,16 @@ jerryx_debugger_tcp_close (jerry_debugger_transport_header_t *header_p) /**< tcp
 
   jerryx_debugger_transport_tcp_t *tcp_p = (jerryx_debugger_transport_tcp_t *) header_p;
 
+  /**
+   * Waiting the remaining tcp data are sent, so that the tcp socket are
+   * closed by debugger client first and the debugger client won't receive
+   * socket are closed by remote error
```
zherczeg, line 137: I don't understand this sentence.

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -131,6 +131,16 @@ jerryx_debugger_tcp_close (jerry_debugger_transport_header_t *header_p) /**< tcp
...
+   */
+  for (int i = 0; i < 5; i += 1)
+  {
+    jerry_debugger_transport_sleep ();
+  }
```
akosthekiss, line 142: My two cents:
- Five seems like a very arbitrary magic number as the loop count to me. May or may not work on different systems. 
- `i += 1` is better as `i++` or `++i`.
- Is this really something that has to be solved here, or could the debugger client be better prepared for premature closing of the socket?

lygstate: > My two cents:
> 
> * Five seems like a very arbitrary magic number as the loop count to me. May or may not work on different systems.
> * `i += 1` is better as `i++` or `++i`.
> * Is this really something that has to be solved here, or could the debugger client be better prepared for premature closing of the socket?

Didn't figure out the best solution yet, looking for help.
I've seeing the CI randomly failure with the debuger tests, The source is here, we need waiting the debugger client close first at normal case.

lygstate: > My two cents:
> 
> * Five seems like a very arbitrary magic number as the loop count to me. May or may not work on different systems.
> * `i += 1` is better as `i++` or `++i`.
> * Is this really something that has to be solved here, or could the debugger client be better prepared for premature closing of the socket?

Here is the things that I solved:
https://github.com/jerryscript-project/jerryscript/runs/1721915835

jerry-ext/debugger/debugger-tcp.c
lygstate: Updated, waiting the debug client close the connection first.

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -144,6 +144,26 @@ jerryx_debugger_tcp_close (jerry_debugger_transport_header_t *header_p) /**< tcp
 
   jerryx_debugger_transport_tcp_t *tcp_p = (jerryx_debugger_transport_tcp_t *) header_p;
 
+  /* Waiting the debug client close the tcp connection first */
+  for (;;)
```
zherczeg, line 148: use `while(true)`

The problem is this loop might never ends.

zherczeg (changes requested):

jerry-ext/debugger/debugger-tcp.c
lygstate: Yeap, any suggestion? post here looking for suggestion

lygstate: > use `while(true)`
> 
> The problem is this loop might never ends.

maybe we doesn't need consider that, as when debugger stopping, means the debug client always closing the socket?

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -144,6 +144,26 @@ jerryx_debugger_tcp_close (jerry_debugger_transport_header_t *header_p) /**< tcp
 
   jerryx_debugger_transport_tcp_t *tcp_p = (jerryx_debugger_transport_tcp_t *) header_p;
 
+  /* Waiting the debug client close the tcp connection first */
```
zherczeg, line 147: Comment: Waiting for the debug client close the tcp connection first.

jerry-ext/debugger/debugger-tcp.c
```diff
@@ -144,6 +144,26 @@ jerryx_debugger_tcp_close (jerry_debugger_transport_header_t *header_p) /**< tcp
...
+  while (true)
+  {
+    char buf[8];
+    jerryx_socket_ssize_t result = recv (tcp_p->tcp_socket, buf, sizeof (buf), 0);
+    if (result == 0)
+    {
+      /**
+       * If result >= 0, this means that there is either data available on the socket, or the socket
+       * has been closed. So waiting the socket closed by remote client by result == 0
```
zherczeg, line 156: This part breaks the loop. Where is the 'waiting' part?

zherczeg (changes requested):
What about simply ignoring error 10054 on windows?

jerry-ext/debugger/debugger-tcp.c
lygstate: == 0 means remote closed the tcp socket, > 1 means still have data, < 0 means have error, JERRYX_EWOULDBLOCK error should be ignored(means have no data but the tcp socket still there)

rerobika:
@zherczeg Just to clarify that's not only windows related issue. https://github.com/jerryscript-project/jerryscript/runs/1721915835#step:5:708

zherczeg:
I am still curious: [WinError 10054] An existing connection was forcibly closed by the remote host - this is a hint, not an error, why can't we simply ignore it?

lygstate:
> I am still curious: [WinError 10054] An existing connection was forcibly closed by the remote host - this is a hint, not an error, why can't we simply ignore it?

Are you talking about ignore the error in python side?
There is a testcase that restart the jerry engine in debugger, so ignore this error is not acceptable, as you don't know it's a normal connection disconnect or caused by the user.

zherczeg:
In the restart debug case you are expecting a restart, it doesn't matter if it fails why it fails. A successful close does not guarantee that the restart will successful. So I still don't get why we cannot ignore this.

lygstate:
> In the restart debug case you are expecting a restart, it doesn't matter if it fails why it fails. A successful close does not guarantee that the restart will successful. So I still don't get why we cannot ignore this.

As I said, we can not disinguish normal disconnection(restart instruction) or it's the jerry executable crash(cause connection reset).
So it's better to waiting normal close when restarting, and when crash,  [WinError 10054]  happens

PR #4506, opened 2021-01-18
Invalid regexp literals should throw syntax error in ES11

rerobika:


tests/test262-esnext-excludelist.xml
```diff
@@ -668,229 +667,6 @@
   <!-- END - Uncategorized failing tests -->
 
   <!-- regression caused by: Invalid regexp patterns should not throw syntax error during parsing (#4038) -->
```
galpeter, line 669: This start comment can be removed and also the end comment for this block.

rerobika: Thanks!

tests/jerry/fail/regression-test-issue-2058.js
```diff
@@ -12,9 +12,4 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-try {
-  /?:/
-  assert(false);
-} catch (e) {
-  assert(e instanceof SyntaxError);
-}
+/?:/
```
galpeter, line 15: Would it make sense to have these tests in an eval call and have correct error type check?

rerobika: Sure!

zherczeg (approved):
LGTM

galpeter (approved):
lgtm

PR #4508, opened 2021-01-18
Add expm1 test case to cover the k==1024 case

lygstate:
Refer to https://github.com/jerryscript-project/jerryscript/blob/8e83638daa24b9f9ffffc83e61d55063e2e9d490/jerry-math/expm1.c#L268

### DIFF
--- a/tools/unit-tests/gen-test-math.c
+++ b/tools/unit-tests/gen-test-math.c
@@ -405,6 +405,7 @@ main (int argc, char **args)
   GEN_DBL_TEST (expm1 (-INFINITY));
   GEN_DBL_TEST (expm1 (NAN));
   GEN_DBL_TEST (expm1 (7.08e+02));
+  GEN_DBL_TEST (expm1 (7.095e+02));
   GEN_DBL_TEST (expm1 (7.10e+02));
   GEN_DBL_TEST (expm1 (-7.40e+02));
   GEN_DBL_TEST (expm1 (-7.50e+02));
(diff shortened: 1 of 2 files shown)


akosthekiss:
What is 0x1p1023 and why does it need extra coverage? Furthermore, why is the negated value not tested?

lygstate:
> What is 0x1p1023 and why does it need extra coverage? Furthermore, why is the negated value not tested?

Refer to https://github.com/jerryscript-project/jerryscript/blob/8e83638daa24b9f9ffffc83e61d55063e2e9d490/jerry-math/expm1.c#L268

tools/unit-tests/gen-test-math.c
```diff
@@ -404,6 +404,7 @@ main (int argc, char **args)
   GEN_DBL_TEST (expm1 (INFINITY));
   GEN_DBL_TEST (expm1 (-INFINITY));
   GEN_DBL_TEST (expm1 (NAN));
+  GEN_DBL_TEST (expm1 (7.095e+02));
```
akosthekiss, line 407: move this one line down, between 708 and 710 (and re-generate the .inc.h)

akosthekiss (changes requested):
In addition to the change requested in the code, please, update commit title and message as well.

Current commit title is misleading, as it is hard to cover a literal/number, unless it is the input value to the function. But it is not the case here. The test case aims at covering the k==1024 case. So, the title should read "Add expm1 test case to cover the k==1024 case".

As for the commit message, don't include the discussion in there. Use comments to answer questions. The commit message should describe what has been done, and why. But rarely something like "refer to ..."

akosthekiss:
BTW, kind of answer-to-self: 0x1p1023 is a hexadecimal floating point literal with the value of `0x1 * 2**1023`. I haven't used them, like ever.

akosthekiss (approved):
LGTM

rerobika (approved):

PR #4509, opened 2021-01-18
test-math.c check signbit

lygstate:
By this PR, problem in jerry math library detected.

### DIFF
--- a/tests/unit-math/test-math.c
+++ b/tests/unit-math/test-math.c
@@ -42,6 +42,12 @@ typedef union
   uint32_t bits32[2];
 } double_bits_t;
 
+static uint8_t
+check_signbit (double_bits_t val)
+{
+  return (uint8_t) (val.bits64 >> 63);
+} /* check_signbit */
+
 static void
 check_double (const char *expr, double computed, double expected)
 {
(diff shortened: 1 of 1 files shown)


akosthekiss (commented):
Why is it important what sign a NaN has? NaN is not a number.

Even if it was important, the extra code does not deserve a new function. That could simply be
`computed_bits.bits64 & 1 << 63 == expected_bits.bits64 & 1 << 63`.

But tests on CI clearly show that this change breaks the math unit tests. So, the first question remains.

lygstate:
> Why is it important what sign a NaN has? NaN is not a number.
> 
> Even if it was important, the extra code does not deserve a new function. That could simply be
> `computed_bits.bits64 & 1 << 63 == expected_bits.bits64 & 1 << 63`.
> 
> But tests on CI clearly show that this change breaks the math unit tests. So, the first question remains.

It's about confirming https://en.wikipedia.org/wiki/IEEE_754#2019
IEEE_754 specified how to output signed NAN and unsigned NAN about different math operators

lygstate:
misunderstand IEEE754

PR #4511, opened 2021-01-19
Fixes math library compiling on msvc

lygstate:

### DIFF
--- a/jerry-math/expm1.c
+++ b/jerry-math/expm1.c
@@ -265,7 +265,12 @@ expm1 (double x)
       y = one - (e - x);
       if (k == 1024)
       {
-        y = y * 2.0 * 0x1p1023;
+        const double twop1023 = ((double_accessor)
+          {
+            .as_int = { .hi = 0x7fe00000, .lo = 0 }
+          }
+        ).dbl; /* 0x1p1023 */
+        y = y * 2.0 * twop1023;
       }
       else
       {
--- a/jerry-math/include/math.h
+++ b/jerry-math/include/math.h
@@ -22,9 +22,14 @@ extern "C"
 #endif /* __cplusplus */
 
 /* General Constants. */
-#define INFINITY    (1.0/0.0)
-#define NAN         (0.0/0.0)
-#define HUGE_VAL    INFINITY
+#ifdef _MSC_VER
+#define INFINITY    ((float) (1e+300 * 1e+300)) /* 1e+300*1e+300 must overflow */
+#define NAN         ((float) (INFINITY * 0.0f))
+#else /* !_MSC_VER */
+#define INFINITY    ((float) (1.0 / 0.0))
+#define NAN         ((float) (0.0 / 0.0))
+#endif /* _MSC_VER */
+#define HUGE_VAL    ((double) INFINITY)
 
 #define isnan(x)    ((x) != (x))
 #define isinf(x)    ((x) == INFINITY ? 1 : (x) == -INFINITY ? -1 : 0)
--- a/jerry-math/jerry-math-internal.h
+++ b/jerry-math/jerry-math-internal.h
@@ -38,8 +38,10 @@
 #if (defined (i386) || defined (__i386) || defined (__i386__) || \
      defined (i486) || defined (__i486) || defined (__i486__) || \
      defined (intel) || defined (x86) || defined (i86pc) || \
+     defined (_M_IX86) || defined (_M_AMD64) || defined (_M_X64) || \
      defined (__alpha) || defined (__osf__) || \
      defined (__x86_64__) || defined (__arm__) || defined (__aarch64__) || \
+     defined (_M_ARM) || defined (_M_ARM64) || \
      defined (__xtensa__) || defined (__MIPSEL)) || \
 (defined (__BYTE_ORDER__) && (__BYTE_ORDER__ == __ORDER_LITTLE_ENDIAN__))
 #define __LITTLE_ENDIAN
@@ -74,8 +76,14 @@ typedef union
 #endif /* __LITTLE_ENDIAN */
 
 #ifndef NAN
-#define NAN (0.0/0.0)
-#endif
+#ifdef _MSC_VER
+#define INFINITY    ((float) (1e+300 * 1e+300)) /* 1e+300*1e+300 must overflow */
+#define NAN         ((float) (INFINITY * 0.0f))
+#else /* !_MSC_VER */
+#define INFINITY    ((float) (1.0 / 0.0))
+#define NAN         ((float) (0.0 / 0.0))
+#endif /* _MSC_VER */
+#endif /* !NAN */
 
 /*
  * ANSI/POSIX
--- a/jerry-math/log.c
+++ b/jerry-math/log.c
@@ -105,11 +105,11 @@ log (double x)
   {
     if (((hx & 0x7fffffff) | lx) == 0) /* log(+-0) = -inf */
     {
-      return -two54 / zero;
+      return -INFINITY;
     }
     if (hx < 0) /* log(-#) = NaN */
     {
-      return (x - x) / zero;
+      return NAN;
     }
     k -= 54;
     x *= two54; /* subnormal number, scale up x */
--- a/jerry-math/log10.c
+++ b/jerry-math/log10.c
@@ -84,12 +84,12 @@ log10 (double x)
     if (((hx & 0x7fffffff) | lx) == 0)
     {
       /* log(+-0)=-inf */
-      return -two54 / zero;
+      return -INFINITY;
     }
     if (hx < 0)
     {
       /* log(-#) = NaN */
-      return (x - x) / zero;
+      return NAN;
     }
     k -= 54;
     x *= two54;    /* subnormal number, scale up x */
--- a/jerry-math/log1p.c
+++ b/jerry-math/log1p.c
@@ -122,7 +122,7 @@ log1p (double x)
       if (x == -1.0)
       {
         /* log1p(-1) = -inf */
-        return -two54 / zero;
+        return -INFINITY;
       }
       else
       {
--- a/jerry-math/log2.c
+++ b/jerry-math/log2.c
@@ -65,11 +65,11 @@ log2 (double x)
   { /* x < 2**-1022  */
     if (((hx & 0x7fffffff) | lx) == 0)
     {
-      return -two54 / zero; /* log(+-0)=-inf */
+      return -INFINITY; /* log(+-0)=-inf */
     }
     if (hx < 0)
     {
-      return (x - x) / zero; /* log(-#) = NaN */
+      return NAN; /* log(-#) = NaN */
     }
     k -= 54;
     x *= two54;    /* subnormal number, scale up x */
--- a/jerry-math/CMakeLists.txt
+++ b/jerry-math/CMakeLists.txt
@@ -18,11 +18,11 @@ project (${JERRY_MATH_NAME} C)
 
 # Compiler / linker flags
 # TODO: Reduce the below list of warning/error disablings as much as possible
-set(COMPILE_FLAGS_MATH "${COMPILE_FLAGS_MATH} -Wno-error=sign-compare")
-set(COMPILE_FLAGS_MATH "${COMPILE_FLAGS_MATH} -Wno-error=sign-conversion")
-set(COMPILE_FLAGS_MATH "${COMPILE_FLAGS_MATH} -Wno-sign-conversion")
-set(COMPILE_FLAGS_MATH "${COMPILE_FLAGS_MATH} -Wno-sign-compare")
-set(COMPILE_FLAGS_MATH "${COMPILE_FLAGS_MATH} -Wno-strict-aliasing")
+set(COMPILE_FLAGS_MATH_GCC_CLANG "${COMPILE_FLAGS_MATH_GCC_CLANG} -Wno-error=sign-compare")
+set(COMPILE_FLAGS_MATH_GCC_CLANG "${COMPILE_FLAGS_MATH_GCC_CLANG} -Wno-error=sign-conversion")
+set(COMPILE_FLAGS_MATH_GCC_CLANG "${COMPILE_FLAGS_MATH_GCC_CLANG} -Wno-sign-conversion")
+set(COMPILE_FLAGS_MATH_GCC_CLANG "${COMPILE_FLAGS_MATH_GCC_CLANG} -Wno-sign-compare")
+set(COMPILE_FLAGS_MATH_GCC_CLANG "${COMPILE_FLAGS_MATH_GCC_CLANG} -Wno-strict-aliasing")
 
 # Include directories
 set(INCLUDE_MATH "${CMAKE_CURRENT_SOURCE_DIR}/include")
@@ -85,9 +85,10 @@ if(ENABLE_AMALGAM)
 endif()
 
 add_library(${JERRY_MATH_NAME} ${SOURCE_MATH})
-set_property(TARGET ${JERRY_MATH_NAME}
-             PROPERTY COMPILE_FLAGS "${COMPILE_FLAGS_MATH}")
-
+if(USING_GCC OR USING_CLANG)
+  set_property(TARGET ${JERRY_MATH_NAME}
+               PROPERTY COMPILE_FLAGS "${COMPILE_FLAGS_MATH_GCC_CLANG}")
+endif()
 target_include_directories(${JERRY_MATH_NAME} PUBLIC ${INCLUDE_MATH})
 
 configure_file(libjerry-math.pc.in libjerry-math.pc @ONLY)


jerry-math/jerry-math-internal.h
```diff
@@ -35,7 +35,7 @@
  * If byte order is supported and set to little or target is among common
  * cases checked define __LITTLE_ENDIAN.
  */
-#if (defined (i386) || defined (__i386) || defined (__i386__) || \
+#if (defined (_MSC_VER) || defined (i386) || defined (__i386) || defined (__i386__) || \
```
akosthekiss, line 38: This is not correct. MSVC can compile for multiple architectures, e.g., for ARM (bot 32 and 64 bit). The macros to be tested for are `_M_IX86` and `_M_X64` (see: https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros?view=msvc-160 )

Moreover, add these tests on a new line after `defined (i86pc)`.

jerry-math/include/math.h
```diff
@@ -22,9 +22,9 @@ extern "C"
 #endif /* __cplusplus */
 
 /* General Constants. */
-#define INFINITY    (1.0/0.0)
-#define NAN         (0.0/0.0)
-#define HUGE_VAL    INFINITY
+#define INFINITY    ((float) (1e+300 * 1e+300)) /* 1e+300*1e+300 must overflow */
+#define NAN         (-(float) (INFINITY * 0.0F))
```
akosthekiss, line 26: This is confusing. Why are INFINITY and NAN "downgraded" to float? And even if they are, what's wrong with the existing computations?

jerry-math/expm1.c
```diff
@@ -265,7 +265,10 @@ expm1 (double x)
       y = one - (e - x);
       if (k == 1024)
       {
-        y = y * 2.0 * 0x1p1023;
+        double_accessor const_0x1p1023;
+        const_0x1p1023.as_int.hi = 0x7fe00000;
+        const_0x1p1023.as_int.lo = 0;
+        y = y * 2.0 * const_0x1p1023.dbl;
```
akosthekiss, line 271: I'd prefer a slightly different approach:
```c
        const double twop1023 = ((double_accessor) { .as_int = { .hi = 0x7fe00000, .lo = 0 } }).dbl; // 0x1p1023
        y = y * 2.0 * twop1023;
```

akosthekiss (changes requested):

jerry-math/expm1.c
lygstate: not works on old msvc，

jerry-math/include/math.h
lygstate: > This is confusing. Why are INFINITY and NAN "downgraded" to float? And even if they are, what's wrong with the existing computations?

msvc raise divde zero error

jerry-math/expm1.c
akosthekiss: That will get problematic at some point. Please, specify what old means. JerryScript is a C99 project. We will have to draw the line somewhere, what deviations from the standard do we still tolerate. MSVC starts to build up quite a list.

jerry-math/include/math.h
akosthekiss: I guess that this is an answer to the computation question. But why the float?

akosthekiss:
BTW, why is it a must to compile jerry-math for Windows / with MSVC? MSVC has its own math implementation, why not just disable jerry-math on Windows? What's the gain with jerry-math?

Note that jerry-math, previously known as jerry-libm, was originally written for embedded systems where size constraints were so strict that even the math lib had to be optimized. I don't think that desktop Windows faces such constraints.

lygstate:
> BTW, why is it a must to compile jerry-math for Windows / with MSVC? MSVC has its own math implementation, why not just disable jerry-math on Windows? What's the gain with jerry-math?
> 
> Note that jerry-math, previously known as jerry-libm, was originally written for embedded systems where size constraints were so strict that even the math lib had to be optimized. I don't think that desktop Windows faces such constraints.

1. jerry-math already support on osx and linux, we can do it on msvc/win32 to match with these two os.
2. by using jerry-math, we can remove the need of msvcr(I want it), msvcrt also are a big things. so we can create minimal size javascript engine on win32, because small size is the advantage of jerryscript, we can take the advantage of that
3. we can creating cross-platform(win32,osx,linux,bare-metal) javascript engine and testing it with same code path(jerry-math also contained in the code path)

jerry-math/jerry-math-internal.h
lygstate: I've guard _MSC_VER with separate condition, this line are too long and also gcc specfic

jerry-math/expm1.c
lygstate: > That will get problematic at some point. Please, specify what old means. JerryScript is a C99 project. We will have to draw the line somewhere, what deviations from the standard do we still tolerate. MSVC starts to build up quite a list.

you are right, works:)

jerry-math/include/math.h
lygstate: > I guess that this is an answer to the computation question. But why the float?

Yeap, float unnecessary, removed that

jerry-math/log2.c
```diff
@@ -52,6 +51,7 @@
 double
 log2 (double x)
 {
+  const double zero = 0.0;
```
lygstate, line 54: By doing this, can remove dive by zero warning/error on msvc

jerry-math/expm1.c
lygstate: > I'd prefer a slightly different approach:
> 
> ```c
>         const double twop1023 = ((double_accessor) { .as_int = { .hi = 0x7fe00000, .lo = 0 } }).dbl; // 0x1p1023
>         y = y * 2.0 * twop1023;
> ```

./jerry-math/expm1.c:268: error: brace should be placed on a separate line
Error: Process completed with exit code 1.

single line are invalid style

akosthekiss:
> by using jerry-math, we can remove the need of msvcr(I want it), msvcrt also are a big things. so we can create minimal size javascript engine on win32, because small size is the advantage of jerryscript, we can take the advantage of that

Please, note that jerry-math is not a fully fledged libm. It does not contain all mathematical functions mandated by the C standard. Only those functions were imported/adapted that are directly used by the engine itself. If the engine is embedded into an application, as it is intended, it is highly probable that other parts of the application will use mathematical functions that are not provided by jerry-math. And in that case the toolchain's own libm implementation (msvcrt, in the case of MSVC) will be the only way to move forward.

lygstate:
> > by using jerry-math, we can remove the need of msvcr(I want it), msvcrt also are a big things. so we can create minimal size javascript engine on win32, because small size is the advantage of jerryscript, we can take the advantage of that
> 
> Please, note that jerry-math is not a fully fledged libm. It does not contain all mathematical functions mandated by the C standard. Only those functions were imported/adapted that are directly used by the engine itself. If the engine is embedded into an application, as it is intended, it is highly probable that other parts of the application will use mathematical functions that are not provided by jerry-math. And in that case the toolchain's own libm implementation (msvcrt, in the case of MSVC) will be the only way to move forward.

Yes, just jerry-math not seeking for fully fledged libm

jerry-math/jerry-math-internal.h
```diff
@@ -35,7 +35,11 @@
  * If byte order is supported and set to little or target is among common
  * cases checked define __LITTLE_ENDIAN.
  */
-#if (defined (i386) || defined (__i386) || defined (__i386__) || \
+#if defined (_MSC_VER)
+#if defined (_M_IX86) || defined (_M_AMD64) || defined (_M_ARM) || defined (_M_ARM64)
+#define __LITTLE_ENDIAN
+#endif
```
akosthekiss, line 41: Don't do this. Just add the `defined (_M_IX86) || defined (_M_AMD64) || defined (_M_ARM) || defined (_M_ARM64)` line to the existing list. BTW, you've omitted `_M_X64` from the original suggestion.

It doesn't matter if some macros are only defined by gcc or clang, others only by msvc. They can be tested in the same list. They are just macros. It makes no harm if a compiler doesn't define / use / recognize some of them (as long as they are tested with `defined()`.

jerry-math/log2.c
akosthekiss: Please describe what kind of diagnostics are reported by MSVC. Division by a zero double literal should behave just like a division by a constant zero double variable. Usually. In a decent compiler where constants are propagated.

akosthekiss (changes requested):

jerry-math/log2.c
lygstate: error messages：

```
build] [18/385   0% :: 0.065] Building C object jerry-math\CMakeFiles\jerry-math.dir\log10.c.obj
[build] FAILED: jerry-math/CMakeFiles/jerry-math.dir/log10.c.obj 
[build] C:\PROGRA~2\MICROS~1.0\VC\bin\amd64\cl.exe  /nologo  -I..\..\jerry-math\include /DWIN32 /D_WINDOWS /W3 /wd4996 /MD /O1 /Ob1 /DNDEBUG /showIncludes /Fojerry-math\CMakeFiles\jerry-math.dir\log10.c.obj /Fdjerry-math\CMakeFiles\jerry-math.dir\jerry-math.pdb /FS -c ..\..\jerry-math\log10.c
[build] ..\..\jerry-math\log10.c(87): error C2124: divide or mod by zero
[build] [18/385   0% :: 0.065] Building C object jerry-math\CMakeFiles\jerry-math.dir\log1p.c.obj
[build] FAILED: jerry-math/CMakeFiles/jerry-math.dir/log1p.c.obj 
[build] C:\PROGRA~2\MICROS~1.0\VC\bin\amd64\cl.exe  /nologo  -I..\..\jerry-math\include /DWIN32 /D_WINDOWS /W3 /wd4996 /MD /O1 /Ob1 /DNDEBUG /showIncludes /Fojerry-math\CMakeFiles\jerry-math.dir\log1p.c.obj /Fdjerry-math\CMakeFiles\jerry-math.dir\jerry-math.pdb /FS -c ..\..\jerry-math\log1p.c
[build] ..\..\jerry-math\log1p.c(125): error C2124: divide or mod by zero
[build] [18/385   0% :: 0.092] Building C object jerry-math\CMakeFiles\jerry-math.dir\asinh.c.obj
[build] [18/385   1% :: 0.093] Building C object jerry-math\CMakeFiles\jerry-math.dir\atan2.c.obj
[build] ..\..\jerry-math\atan2.c(77): warning C4146: unary minus operator applied to unsigned type, result still unsigned
[build] [18/385   1% :: 0.102] Building C object jerry-math\CMakeFiles\jerry-math.dir\asin.c.obj
[build] [18/385   1% :: 0.103] Building C object jerry-math\CMakeFiles\jerry-math.dir\acosh.c.obj
[build] [18/385   1% :: 0.106] Building C object jerry-math\CMakeFiles\jerry-math.dir\cbrt.c.obj
[build] [18/385   2% :: 0.110] Building C object jerry-math\CMakeFiles\jerry-math.dir\copysign.c.obj
[build] [18/385   2% :: 0.113] Building C object jerry-math\CMakeFiles\jerry-math.dir\ceil.c.obj
[build] ..\..\jerry-math\ceil.c(116): warning C4018: '<': signed/unsigned mismatch
[build] [18/385   2% :: 0.115] Building C object jerry-math\CMakeFiles\jerry-math.dir\floor.c.obj
[build] ..\..\jerry-math\floor.c(115): warning C4018: '<': signed/unsigned mismatch
[build] [18/385   2% :: 0.116] Building C object jerry-math\CMakeFiles\jerry-math.dir\atanh.c.obj
[build] c:\work\study\languages\typescript\jerryscript\jerry-math\atanh.c(69) : warning C4723: potential divide by 0
[build] [18/385   3% :: 0.116] Building C object jerry-math\CMakeFiles\jerry-math.dir\fabs.c.obj
[build] [18/385   3% :: 0.116] Building C object jerry-math\CMakeFiles\jerry-math.dir\cosh.c.obj
[build] [18/385   3% :: 0.118] Building C object jerry-math\CMakeFiles\jerry-math.dir\acos.c.obj
[build] [18/385   3% :: 0.119] Building C object jerry-math\CMakeFiles\jerry-math.dir\expm1.c.obj
[build] [18/385   4% :: 0.120] Building C object jerry-math\CMakeFiles\jerry-math.dir\atan.c.obj
[build] [18/385   4% :: 0.126] Building C object jerry-math\CMakeFiles\jerry-math.dir\fmod.c.obj
[build] ..\..\jerry-math\fmod.c(54): warning C4146: unary minus operator applied to unsigned type, result still unsigned
[build] [18/385   4% :: 0.126] Building C object jerry-math\CMakeFiles\jerry-math.dir\exp.c.obj
[build] ninja: build stopped: subcommand failed.
[build] Build finished with exit code 1

```

lygstate: on msvc, they becave differently, warning vs error

jerry-math/include/math.h
```diff
@@ -22,8 +22,8 @@ extern "C"
 #endif /* __cplusplus */
 
 /* General Constants. */
-#define INFINITY    (1.0/0.0)
-#define NAN         (0.0/0.0)
+#define INFINITY    (1e+300 * 1e+300) /* 1e+300*1e+300 must overflow */
+#define NAN         (INFINITY * 0.0)
```
akosthekiss, line 26: So, I did my homework. And it just turned out that (a) I did not know precisely how the standard specifies infinity and NaN, and (b) jerry-math's math.h was incorrect all along.

According to the specs:
- `INFINITY` and `NAN` are _float_ constant expressions, and
- `HUGE_VAL` is a _double_ constant expression (that evaluates to infinity).

Therefore, I suggest to have these macros as suggested _for the Windows/MSVC configuration_, but also keep a slightly updated version of the existing macros for other configs. That is:
```c
#ifdef _MSC_VER
#define INFINITY    ((float) (1e+300 * 1e+300))
#define NAN         ((float) (INFINITY * 0.0f))
#else
#define INFINITY    ((float) (1.0/0.0))
#define NAN         ((float) (0.0/0.0))
#endif
#define HUGE_VAL    ((double) INFINITY)
```

jerry-math/jerry-math-internal.h
```diff
@@ -74,7 +76,8 @@ typedef union
 #endif /* __LITTLE_ENDIAN */
 
 #ifndef NAN
-#define NAN (0.0/0.0)
+/* 1e+300*1e+300 must overflow */
+#define NAN         ((1e+300 * 1e+300) * 0.0)
```
akosthekiss, line 80: This NAN definition should be changed likewise.

jerry-math/log.c
```diff
@@ -93,6 +92,7 @@
 double
 log (double x)
 {
+  const double zero = 0.0;
```
akosthekiss, line 95: And in these files, instead of using a zero variable to make division by zero possible on MSVC, the div0 expressions should be changed to +/-INFINITY or NAN. (This is how the patch looked like at some point of time, but then it has changed.)

akosthekiss (changes requested):

jerry-math/log.c
```diff
@@ -105,7 +105,7 @@ log (double x)
   {
     if (((hx & 0x7fffffff) | lx) == 0) /* log(+-0) = -inf */
     {
-      return -two54 / zero;
+      return -INFINITY;
```
akosthekiss, line 108: How come that line 112 does not have to be edited? That's a div0 as well, only for NaN.

jerry-math/log10.c
```diff
@@ -84,7 +84,7 @@ log10 (double x)
     if (((hx & 0x7fffffff) | lx) == 0)
     {
       /* log(+-0)=-inf */
-      return -two54 / zero;
+      return -INFINITY;
```
akosthekiss, line 87: same here

jerry-math/log1p.c
```diff
@@ -122,7 +122,7 @@ log1p (double x)
       if (x == -1.0)
       {
         /* log1p(-1) = -inf */
-        return -two54 / zero;
+        return -INFINITY;
```
akosthekiss, line 125: and here

jerry-math/jerry-math-internal.h
```diff
@@ -74,7 +76,13 @@ typedef union
 #endif /* __LITTLE_ENDIAN */
 
 #ifndef NAN
-#define NAN (0.0/0.0)
+#ifdef _MSC_VER
+#define INFINITY    ((float) (1e+300 * 1e+300)) /* 1e+300*1e+300 must overflow */
+#define NAN         ((float) (INFINITY * 0.0f))
+#else
```
akosthekiss, line 82: (minor) If the PR gets updated, it would be nice to add a comment here: `/* !_MSC_VER */`

jerry-math/jerry-math-internal.h
```diff
@@ -74,7 +76,13 @@ typedef union
...
+#define INFINITY    ((float) (1.0/0.0))
+#define NAN         ((float) (0.0/0.0))
```
akosthekiss, line 84: (minor) for some legacy reason, it seems that there were no spaces around the division operator. Now would be a chance to add then.

jerry-math/log.c
lygstate: weird， the compiler didn't complain that， anyway， I will update there

jerry-math/jerry-math-internal.h
akosthekiss: this suggestion did not get applied

akosthekiss (approved):
LGTM (except for the missing preproc directive comment)

lygstate:
> LGTM (except for the missing preproc directive comment)

forgot, let's do it

dbatyai (approved):
LGTM

PR #4513, opened 2021-01-19
Refactoring time-related default port implementations

akosthekiss:
In the non-Windows code paths:
- New approach to compute TZA without the need for GNU-specific
  `struct tm.tm_gmtoff`.
- Always using `usleep` to sleep. (No real need for `nanosleep` as
  port API has sleep granularity of milliseconds.)
- Not checking for "time.h" at build configuration time as that
  header is mandated by the C standard.
- Not checking for "unistd.h" at build configuration time as that
  header is mandated by the POSIX standard (the default port is
  targeting POSIX systems -- and Windows).
- Fixing some macro guards.

### DIFF
--- a/jerry-port/default/default-date.c
+++ b/jerry-port/default/default-date.c
@@ -13,20 +13,17 @@
  * limitations under the License.
  */
 
-#ifdef HAVE_TM_GMTOFF
 #include <time.h>
-#endif /* HAVE_TM_GMTOFF */
 
 #ifdef _WIN32
 #include <windows.h>
 #include <winbase.h>
 #include <winnt.h>
-#include <time.h>
 #endif /* _WIN32 */
 
-#ifdef __GNUC__
+#if defined (__GNUC__) || defined (__clang__)
 #include <sys/time.h>
-#endif /* __GNUC__ */
+#endif /* __GNUC__ || __clang__ */
 
 #include "jerryscript-port.h"
 #include "jerryscript-port-default.h"
@@ -54,9 +51,7 @@ static double FileTimeToUnixTimeMs (FILETIME ft)
 #endif /* _WIN32 */
 
 /**
- * Default implementation of jerry_port_get_local_time_zone_adjustment. Uses the 'tm_gmtoff' field
- * of 'struct tm' (a GNU extension) filled by 'localtime_r' if available on the
- * system, does nothing otherwise.
+ * Default implementation of jerry_port_get_local_time_zone_adjustment.
  *
  * @return offset between UTC and local time at the given unix timestamp, if
  *         available. Otherwise, returns 0, assuming UTC time.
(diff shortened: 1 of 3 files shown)


akosthekiss:
I have to admit that I'm not sure whether existing test cases cover these routines properly. I've partially validated the TZA computation with the C code below:

```c
#include <stdio.h>
#include <time.h>

double tza_old(time_t time_utc) {
  tzset(); // should be executed only once?

  // time in broken-down structure, expressed in local time zone
  struct tm tm_local;
  localtime_r(&time_utc, &tm_local);

  // seconds east of UTC (from GNU-specific structure field)
  return (double)tm_local.tm_gmtoff;
}

double tza_new(time_t time_utc) {
  // time in broken-down structure, expressed as UTC time
  struct tm tm_utc;
  gmtime_r(&time_utc, &tm_utc);

  // time converted back to seconds since Epoch UTC
  tm_utc.tm_isdst = -1; // if not overridden, DST will not be taken into account
  time_t time_local = mktime(&tm_utc);

  // seconds east of UTC (from simple time differences)
  return difftime(time_utc, time_local);
}

int main() {
  time_t now_utc = time(NULL); // seconds since Epoch UTC
  time_t past_utc = now_utc - 182L * 24 * 60 * 60; // cca. half a year ago

  printf("tza-old now: %lf\n", tza_old(now_utc));
  printf("tza-old past: %lf\n", tza_old(past_utc));

  printf("tza-new now: %lf\n", tza_new(now_utc));
  printf("tza-new past: %lf\n", tza_new(past_utc));

  return 0;
}
```

Built and ran on Ubuntu 18.04.5 LTS (Linux 4.15.0-130-generic x86_64) as:
```
$ gcc tza.c -o tza
$ ./tza 
tza-old now: 3600.000000
tza-old past: 7200.000000
tza-new now: 3600.000000
tza-new past: 7200.000000
```

lygstate (approved):
LGTM

lygstate:
@akosthekiss Does this what you want?
```
lygstate@DESKTOP-94PU0GB:/mnt/c/work/study/languages/typescript/jerryscript$ ./a.out
tza-old now: -28800.000000
tza-old past: -25200.000000
tza-new now: -28800.000000
tza-new past: -25200.000000
```

akosthekiss:
@lygstate Thanks. Depends on your actual time zone. Looks like GMT-8: PST? Validates part of the code. But the `is_utc=false` logic of the TZA port function is still hard to check whether it works OK. (The above C code does not validate that at all.)

lygstate:
Yeap, it's GMT-8

lygstate:
> @lygstate Thanks. Depends on your actual time zone. Looks like GMT-8: PST? Validates part of the code. But the `is_utc=false` logic of the TZA port function is still hard to check whether it works OK. (The above C code does not validate that at all.)

If I read the code correctly, the is_utc option didn't considerated on _WIN32

lygstate:
I have  a question, If the OS provide what's the current time zone is, can we use the timezone info directly implement jerry_port_get_local_time_zone_adjustment ?

jerry-port/default/default-date.c
```diff
@@ -94,9 +76,39 @@ double jerry_port_get_local_time_zone_adjustment (double unix_ms,  /**< ms since
...
+    is_utc = true;
+  }
+
+  return tza_s * 1000;
+#else
+  (void) unix_ms; /* unused */
+  (void) is_utc; /* unused */
   return 0.0;
-#endif /* HAVE_TM_GMTOFF */
+#endif
```
rerobika, line 111: Please add the missing comments to preprocessor directives. (here and below)

rerobika (approved):
LGTM after the style fix

akosthekiss:
Added the pre-processor directive comments. PR updated. Thanks for the review.

zherczeg (approved):
LGTM

PR #4514, opened 2021-01-19
Fixes compile error for rand() can't be overriden on Windows (benchmarking.c)

lygstate:
Fixes compile error for rand() can't be overriden on Windows (benchmarking.c)


akosthekiss:
You don't want to do this. There is no point in guarding out the benchmarking.c. This file is used to create a stable version of the jerry executable (called jerry-test) that ensures that all test cases execute the same, even if they use time or rng-related functions. If you guard out this file, that means basically that you don't want to have a jerry variant useful for benchmarking. Then, simply don't build that executable (i.e., configure the build with `JERRY_CMDLINE_TEST=OFF`).

PR #4517, opened 2021-01-21
Make install_jerry work with MinGW

lygstate:
On windows, the generator can be Visual Studio or Ninja, when targeting mingw,
the generator can be Ninja or Makefile, they both use 'install' instead MSVC 'INSTALL',
so check the solution file for MSVC

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -269,7 +269,7 @@ def make_jerry(arguments):
     return proc.returncode
 
 def install_jerry(arguments):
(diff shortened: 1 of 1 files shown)


akosthekiss (approved):
LGTM

Unfortunately, I couldn't find a better solution to differentiate between MSVC and non-MSVC builds on Windows either.

The commit message should be edited for language.

dbatyai (approved):
LGTM

PR #4520, opened 2021-01-21
Minor Fix in date object prototype

kisbg:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
@@ -630,7 +630,7 @@ ecma_builtin_date_prototype_dispatch_routine (uint8_t builtin_routine_id, /**< b
 
     *prim_value_p = ecma_date_time_clip (time_num);
 
-    return ecma_make_number_value (time_num);
+    return ecma_make_number_value (*prim_value_p);
   }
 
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
```diff
@@ -630,7 +630,7 @@ ecma_builtin_date_prototype_dispatch_routine (uint8_t builtin_routine_id, /**< b
 
     *prim_value_p = ecma_date_time_clip (time_num);
 
-    return ecma_make_number_value (time_num);
+    return ecma_make_number_value (*prim_value_p);
```
galpeter, line 633: Do we need to set the prim_value_p above at all? Can't we just call the date_time_clip directly here?

rerobika: I think we need to set, since `prim_value_p ` points to the date object's `u.class_prop.u.value` which need to be updated as well.

galpeter: ah... true, this is the set time operation. Sorry about that.

galpeter (approved):
lgtm

kisbg:
@galpeter  Thank you for the review!

rerobika (approved):

PR #4524, opened 2021-01-22
Remove full stop after error messages

zherczeg:


lygstate:
Can we turning all ECMA_ERR_MSG  to interger by using some auto tools?
For smaller binary size

jerry-core/ecma/builtin-objects/ecma-builtin-array-prototype.c
```diff
@@ -2851,7 +2851,7 @@ ecma_builtin_array_prototype_object_flat_map (ecma_value_t callback, /**< callba
 {
   if (!ecma_op_is_callable (callback))
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Callback function is not callable."));
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Callback function is not callable"));
```
akosthekiss, line 2854: Lots of occurrences.

jerry-core/ecma/builtin-objects/ecma-builtin-dataview-prototype.c
```diff
@@ -132,7 +132,7 @@ ecma_builtin_dataview_prototype_object_getters (ecma_value_t this_arg, /**< this
 
       if (ecma_arraybuffer_is_detached (obj_p->buffer_p))
       {
-        return ecma_raise_type_error (ECMA_ERR_MSG ("ArrayBuffer has been detached."));
+        return ecma_raise_type_error (ECMA_ERR_MSG ("ArrayBuffer has been detached"));
```
akosthekiss, line 135: 2nd occurrence

jerry-core/ecma/builtin-objects/ecma-builtin-global.c
```diff
@@ -380,7 +380,7 @@ ecma_builtin_global_object_encode_uri_helper (lit_utf8_byte_t *input_start_p, /*
       else
       {
         ecma_stringbuilder_destroy (&builder);
-        return ecma_raise_uri_error (ECMA_ERR_MSG ("Unicode surrogate pair missing."));
+        return ecma_raise_uri_error (ECMA_ERR_MSG ("Unicode surrogate pair missing"));
```
akosthekiss, line 383: 2nd

jerry-core/ecma/builtin-objects/ecma-builtins.c
```diff
@@ -1545,7 +1545,7 @@ ecma_builtin_dispatch_construct (ecma_object_t *obj_p, /**< built-in object */
 
   if (ecma_builtin_function_is_routine (obj_p))
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Built-in routines have no constructor."));
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Built-in routines have no constructors"));
```
akosthekiss, line 1548: I think singular was correct here

akosthekiss (commented):
Looks good in general
- Added some comments for improvement
- Found that in some files, repeated error messages are moved into a variable and that variable is reused multiple times. However, several files do contain identical error messages. Commented on some of these, but gave up ultimately (there were just too many). Not sure whether it's worth adding al those variables now.

jerry-core/ecma/builtin-objects/ecma-builtin-arraybuffer-prototype.c
```diff
@@ -92,23 +92,23 @@ ecma_builtin_arraybuffer_prototype_object_slice (ecma_value_t this_arg, /**< thi
 {
   if (!ecma_is_value_object (this_arg))
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Argument 'this' is not object."));
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Argument 'this' is not object"));
```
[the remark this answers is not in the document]
zherczeg: This varies in the code base, and would be good to decide. Put a/an or put nothing. I like the latter, but engines like v8 prefer the former, so the best choice is debatable. What should we do?

jerry-core/ecma/builtin-objects/ecma-builtin-arraybuffer-prototype.c
```diff
@@ -92,23 +92,23 @@ ecma_builtin_arraybuffer_prototype_object_slice (ecma_value_t this_arg, /**< thi
...
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Argument 'this' is not an ArrayBuffer object"));
   }
 
   /* TODO: step 3. if SharedArrayBuffer will be implemented */
 
   /* 4. */
   if (ecma_arraybuffer_is_detached (object_p))
   {
-    return ecma_raise_type_error (ECMA_ERR_MSG ("ArrayBuffer has been detached."));
+    return ecma_raise_type_error (ECMA_ERR_MSG ("ArrayBuffer has been detached"));
```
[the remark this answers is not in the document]
zherczeg: This also another problem I don't know what to do with them. Especially if the message is spread into multiple files. Perhaps we should introduce another file in core, and put many error messages there. Another question: shall we do that here, or in another patch.

jerry-core/ecma/builtin-objects/ecma-builtin-arraybuffer-prototype.c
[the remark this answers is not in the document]
akosthekiss: I prefer "an object", because that's the grammatically correct phrase. These messages are to inform humans, so they should (try to) follow language rules.

jerry-core/ecma/builtin-objects/ecma-builtin-arraybuffer-prototype.c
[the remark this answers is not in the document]
akosthekiss: I'm totally OK with the refactoring of the "error message approach" in a later PR. But. Don't start it partially here then. I'd suggest then to revert the introduction of the  new variables in this PR (I guess that happened in jerry.c and in another file, too.)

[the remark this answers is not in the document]
zherczeg: jerry.c has these messages, I only added some extra.
The message in vm.c could be removed, that is ok.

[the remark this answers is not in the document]
zherczeg: https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/api/jerry.c#L109
This was the inspiration

ossy-szeged:
> Can we turning all ECMA_ERR_MSG to interger by using some auto tools?
> For smaller binary size

Error messages are disabled by default, they don't increase the binary size. If you enable it for debugging purposes, code size shouldn't matter.

jerry-core/ecma/operations/ecma-array-object.c
```diff
@@ -173,7 +173,7 @@ ecma_op_new_array_object_from_length (ecma_length_t length) /**< length of the n
 #if ENABLED (JERRY_ESNEXT)
   if (length > UINT32_MAX)
   {
-    ecma_raise_range_error (ECMA_ERR_MSG ("Invalid array length"));
+    ecma_raise_range_error (ECMA_ERR_MSG ("Invalid Array length"));
```
akosthekiss, line 176: Object is often rewritten to object, but array is almost always rewritten to Array? Why this duality?

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -685,7 +685,7 @@ ecma_proxy_object_get_own_property_descriptor (ecma_object_t *obj_p, /**< proxy
   if (!ecma_is_value_object (trap_result) && !ecma_is_value_undefined (trap_result))
   {
     ecma_free_value (trap_result);
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Trap is not object nor undefined."));
+    return ecma_raise_type_error (ECMA_ERR_MSG ("Trap is not object nor undefined"));
```
akosthekiss, line 688: neither an object nor undefined?

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -789,24 +789,21 @@ ecma_proxy_object_get_own_property_descriptor (ecma_object_t *obj_p, /**< proxy
...
-    if ((prop_desc_p->flags & (ECMA_PROP_IS_WRITABLE_DEFINED | ECMA_PROP_IS_WRITABLE)) == ECMA_PROP_IS_WRITABLE_DEFINED
-        && target_is_writable)
+    if (!target_has_desc
+        || target_is_configurable
+        || ((prop_desc_p->flags & mask) == ECMA_PROP_IS_WRITABLE_DEFINED
+            && target_is_writable))
     {
       ecma_free_property_descriptor (prop_desc_p);
-      return ecma_raise_type_error (ECMA_ERR_MSG ("Not compatible."));
+      return ecma_raise_type_error (ECMA_ERR_MSG ("The two descriptors are not compatible"));
```
akosthekiss, line 806: incompatible?

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -1147,15 +1144,14 @@ ecma_proxy_object_get (ecma_object_t *obj_p, /**< proxy object */
...
+      ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("Incorrect value is returned by a Proxy 'get' trap"));
     }
     else if (!(target_desc.flags & ECMA_PROP_IS_CONFIGURABLE)
             && (target_desc.flags & (ECMA_PROP_IS_GET_DEFINED | ECMA_PROP_IS_SET_DEFINED))
             && target_desc.get_p == NULL
             && !ecma_is_value_undefined (trap_result))
     {
-      ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("given property is a non-configurable property and"
+      ret_value = ecma_raise_type_error (ECMA_ERR_MSG ("Property of a Proxy is non-configurable and"
                                                        " does not have a getter function"));
```
akosthekiss, line 1155: move the leading space from the string literal on this line to the end of the previous line (to make it consistent with other multi-line literals)

jerry-core/ecma/operations/ecma-typedarray-object.c
```diff
@@ -1710,7 +1707,7 @@ ecma_typedarray_species_create (ecma_value_t this_arg, /**< this argument */
   if (ECMA_TYPEDARRAY_IS_BIGINT_TYPE (info.id) ^ ECMA_TYPEDARRAY_IS_BIGINT_TYPE (result_info.id))
   {
     ecma_free_value (result);
-    return ecma_raise_type_error (ECMA_ERR_MSG ("Source and result array does not match in [[ContentType]]"));
+    return ecma_raise_type_error (ECMA_ERR_MSG ("TypedArray returned by [[ContentType]] does not match to source"));
```
akosthekiss, line 1710: does not match source?

akosthekiss (commented):
Looks good in general. However, I still have some suggestions/remarks.

jerry-core/ecma/operations/ecma-array-object.c
zherczeg: There should be some logic behind naming. My current preference is that primitive types (can be the result of a `typeof` operator) are lowercase, and specialized objects are uppercase. So `number` (value) and `Number` (object) are two different things. The exception is `BigInt`, which looks ugly in lowercase and `Symbol` because of v8. I don't like the latter, but I have no choice. There is no `array` in ECMAScript only a specialized `Array` object.

zherczeg:

akosthekiss (approved):
LGTM

rerobika (approved):

PR #4527, opened 2021-01-22
Improve destructuring patterns

rerobika:
- Fix evaluation order of the operands
- Implement proper iterator closing of array destructuring
- Support next method caching


jerry-core/vm/vm.c
```diff
@@ -2613,16 +2668,33 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
         {
-          result = left_value;
-          left_value = ECMA_VALUE_UNDEFINED;
-
           ecma_value_t *stack_bottom_p = VM_GET_REGISTERS (frame_ctx_p) + register_end + frame_ctx_p->context_depth;
 
           while (stack_top_p > stack_bottom_p)
           {
-            ecma_fast_free_value (*(--stack_top_p));
+            ecma_value_t stack_item = *(--stack_top_p);
```
zherczeg, line 2675: I would do this differently. I would push a normal context to the stack, and store the previous value of `context_depth` as part of this context. Then set `context_depth` to the top of the stack. When `vm_stack_context_abort` runs, it does not only remove the context, but also free the JS values stored before this context is pushed.

zherczeg (changes requested):

jerry-core/parser/js/js-parser-expr.c
```diff
@@ -3294,6 +3292,7 @@ parser_pattern_emit_rhs (parser_context_t *context_p, /**< context */
                          uint16_t rhs_opcode, /**< opcode to process the rhs value */
                          uint16_t literal_index) /**< literal index for object pattern */
 {
+
```
zherczeg, line 3295: Remove this newline.

jerry-core/vm/vm-stack.c
```diff
@@ -41,6 +41,42 @@ JERRY_STATIC_ASSERT (PARSER_WITH_CONTEXT_STACK_ALLOCATION == PARSER_TRY_CONTEXT_
...
+                                ecma_value_t *vm_stack_top_p, /**< current stack top */
+                                uint32_t context_stack_allocation) /**< 0 - if all context element should be released
+                                                                    *   stack allocation of the context - otherwise */
+{
+  JERRY_ASSERT (VM_GET_CONTEXT_TYPE (vm_stack_top_p[-1]) == VM_CONTEXT_ITERATOR
+                || VM_GET_CONTEXT_TYPE (vm_stack_top_p[-1]) == VM_CONTEXT_OBJ_INIT
+                || VM_GET_CONTEXT_TYPE (vm_stack_top_p[-1]) == VM_CONTEXT_OBJ_INIT_REST);
+
+  uint32_t context_size = VM_GET_CONTEXT_END (vm_stack_top_p[-1]);
+  VM_MINUS_EQUAL_U16 (frame_ctx_p->context_depth, context_size);
```
zherczeg, line 61: What about storing absolute values? Would it make the code simpler?

jerry-core/vm/vm.c
```diff
@@ -1001,6 +1001,12 @@ opfunc_construct (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
     opcode_data &= (uint32_t) ~VM_OC_PUT_BLOCK; \
   }
 
+/**
+ * Get the current stack bottom above the existing contexts
+ */
+#define VM_STACK_BOTTOM() \
```
zherczeg, line 1007: The name is misleading. Maybe `VM_LAST_CONTEXT_END` ?

zherczeg (changes requested):
Good in general. I like it.

Is this the generic layout of all of these "dynamic" frames:

[previous JS values stored by the main_loop][JS values belong to the context][context descriptor]

Since no `catch` can be in these contexts, they store the previous stack depth in the context descriptor rather than the end?

Maybe some descriptive comment cound be added somewhere.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -562,6 +562,23 @@ ecma_gc_mark_executable_object (ecma_object_t *object_p) /**< object */
 
     do
     {
+      if (VM_CONTEXT_HAS_VARIABLE_LENGTH (VM_GET_CONTEXT_TYPE (context_top_p[-1])))
```
zherczeg, line 565: VM_CONTEXT_IS_VARIABLE_LENGTH

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -1245,6 +1262,23 @@ ecma_gc_free_executable_object (ecma_object_t *object_p) /**< object */
...
 
+      if (VM_CONTEXT_HAS_VARIABLE_LENGTH (VM_GET_CONTEXT_TYPE (context_top_p[-1])))
+      {
+        int32_t context_size = (int32_t) VM_GET_CONTEXT_END (context_top_p[-1]);
+
+        for (int32_t offset = 2; offset <= context_size; offset++)
+        {
+          if (ecma_is_value_object (context_top_p[-offset]))
+          {
+            context_top_p[offset] = ECMA_VALUE_UNDEFINED;
```
zherczeg, line 1273: I think this is incorrect. The code below calls `vm_stack_context_abort` after the object values are cleared. Maybe freeing the non-objects could work.

jerry-core/vm/vm-stack.c
```diff
@@ -189,16 +231,15 @@ vm_stack_find_finally (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
 
   while (frame_ctx_p->context_depth > 0)
   {
-    vm_stack_context_type_t context_type;
+    vm_stack_context_type_t context_type = VM_GET_CONTEXT_TYPE (stack_top_p[-1]);
     uint32_t context_end = VM_GET_CONTEXT_END (stack_top_p[-1]);
 
-    if (search_limit < context_end)
+    if (!VM_CONTEXT_HAS_VARIABLE_LENGTH (context_type) && search_limit < context_end)
```
zherczeg, line 237: Since `break` and `continue` are statements, it is impossible to encounter a variable length context when they are running. This could be checked by an assert and expolited in some way (e.g. when the search_limit is UINT32_MAX, this check is never true).

jerry-core/vm/vm-stack.c
```diff
@@ -323,62 +366,71 @@ vm_stack_find_finally (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
...
-            if (!ECMA_IS_VALUE_ERROR (result))
+            if (context_type == VM_CONTEXT_FOR_AWAIT_OF && !ECMA_IS_VALUE_ERROR (result))
             {
-              uint16_t extra_flags = (ECMA_EXECUTABLE_OBJECT_DO_AWAIT_OR_YIELD
-                                      | (ECMA_AWAIT_FOR_CLOSE << ECMA_AWAIT_STATE_SHIFT));
-              async_generator_object_p->u.class_prop.extra_info |= extra_flags;
+              ecma_extended_object_t *async_generator_object_p = VM_GET_EXECUTABLE_OBJECT (frame_ctx_p);
 
-              stack_top_p = vm_stack_context_abort (frame_ctx_p, stack_top_p);
+              result = ecma_promise_async_await (async_generator_object_p, result);
```
zherczeg, line 397: Is this part is coming from somwhere else? Is it a code duplication?

zherczeg (changes requested):

jerry-core/ecma/base/ecma-gc.c
rerobika: Good catch. Moreover only a `vm_stack_context_abort` is enough here.

jerry-core/vm/vm-stack.c
rerobika: Nope, that's an already existing code, just indented by one. That's why it seems so big change.

jerry-core/ecma/base/ecma-gc.c
```diff
@@ -1245,6 +1262,20 @@ ecma_gc_free_executable_object (ecma_object_t *object_p) /**< object */
     {
       context_top_p[-1] &= (uint32_t) ~VM_CONTEXT_HAS_LEX_ENV;
 
+      if (VM_CONTEXT_IS_VARIABLE_LENGTH (VM_GET_CONTEXT_TYPE (context_top_p[-1])))
+      {
+        int32_t context_size = (int32_t) VM_GET_CONTEXT_END (context_top_p[-1]);
+
+        for (int32_t offset = 2; offset <= context_size; offset++)
```
zherczeg, line 1269: What about:
```
last_item_p = context_top_p - VM_GET_CONTEXT_END (context_top_p[-1])
ASSERT(last_item_p >= context_end_p);
context_top_p--;
do {
  ecma_free_value_if_not_object (*(--context_top_p));
} while (context_top_p > last_item_p);
continue;
```

jerry-core/vm/vm.c
```diff
@@ -2406,16 +2445,52 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+          stack_top_p[-2] = left_value;
+
+          if (context_type == VM_CONTEXT_OBJ_INIT_REST)
+          {
+            stack_top_p[-3] = ecma_make_object_value (ecma_op_new_array_object (0));
+          }
           continue;
         }
-        case VM_OC_INITIALIZER_PUSH_REST:
+
```
zherczeg, line 2473: I think this newline is not needed

jerry-core/vm/vm.c
```diff
@@ -2304,24 +2310,38 @@ vm_loop (vm_frame_ctx_t *frame_ctx_p) /**< frame context */
...
+          result = ecma_op_get_iterator (stack_top_p[-1], ECMA_VALUE_SYNC_ITERATOR, &left_value);
 
           if (ECMA_IS_VALUE_ERROR (result))
           {
             goto error;
           }
 
-          *stack_top_p++ = result;
+          uint32_t context_size = (uint32_t) (stack_top_p + PARSER_ITERATOR_CONTEXT_STACK_ALLOCATION
+                                              - VM_LAST_CONTEXT_END ());
```
zherczeg, line 2323: There are a + and - operators, so this is a style error. It does not look too nice.

zherczeg (changes requested):
Some more changes

zherczeg (approved):
LGTM

jerry-core/ecma/operations/ecma-iterator-object.c
```diff
@@ -242,83 +239,27 @@ ecma_op_get_iterator (ecma_value_t value, /**< value to get iterator from */
...
+    return next_method;
   }
 
-  ecma_object_t *next_obj_p = ecma_get_object_from_value (func_next);
-
-  bool has_value = !ecma_is_value_empty (value);
-
-  ecma_value_t result;
-  if (has_value)
+  if (ecma_is_value_object (next_method) && ecma_op_is_callable (next_method))
```
galpeter, line 251: AFAIK, we don't need to check if next method is an object, the op_is_callable already does this internally.

jerry-core/vm/vm-stack.c
```diff
@@ -323,62 +368,71 @@ vm_stack_find_finally (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
...
+      }
+      else
+      {
+        ecma_value_t iterator = stack_top_p[-3];
+        result = ecma_op_get_method_by_magic_id (iterator, LIT_MAGIC_STRING_RETURN);
 
-          if (context_type == VM_CONTEXT_FOR_AWAIT_OF && !ECMA_IS_VALUE_ERROR (result))
+        if (!ECMA_IS_VALUE_ERROR (result) && !ecma_is_value_undefined (result))
+        {
+          if (!ecma_is_value_object (result) || !ecma_op_is_callable (result))
```
galpeter, line 384: The `!ecma_op_is_callable (result)` check should be enough. However if I read this correctly this is just indentation change so maybe we'll change this in a different PR.

tests/jerry/es.next/array-pattern.js
```diff
@@ -329,33 +329,6 @@ function __createIterableObject (arr, methods) {
   assert (x.prop === "42");
 }) ();
 
-mustThrow ("var iter = __createIterableObject([], "
```
galpeter, line 332: So these these should not throw error now? Would it make sense to adapt the tests to expect "no error"?

jerry-core/vm/vm-stack.c
```diff
@@ -189,16 +231,17 @@ vm_stack_find_finally (vm_frame_ctx_t *frame_ctx_p, /**< frame context */
 
   while (frame_ctx_p->context_depth > 0)
   {
-    vm_stack_context_type_t context_type;
+
```
galpeter, line 234: AFAIK, no need for this empty line.

jerry-core/ecma/operations/ecma-iterator-object.c
rerobika: Good catch. I've also removed it elsewhere.

rerobika:
@galpeter Thanks for the review, I've updated the patch.

galpeter (approved):
lgtm

Issue #3785, opened 2020-05-23
Stack Exhaustion (ecma_proxy_object_get, ecma_proxy_object_set)


nszetei:
###### JerryScript revision
https://github.com/jerryscript-project/jerryscript/commit/6cd309b5252d15f3e3e6481c72a4d8fa7f9557a9

###### Build platform
Ubuntu 20.04 LTS (Linux 5.4.0-31-generic x86_64)

###### Build steps

```sh
python tools/build.py --profile=es2015-subset --lto=off --compile-flag=-g \
--error-messages=on --debug --compile-flag=-g --strip=off --logging=on \
 --compile-flag=-fsanitize=address
```

###### Test cases

```
var v2 = {};
var v4 = new Proxy(Uint8Array,v2);
v4.__proto__ = v4;
v4[1] = 2;
```

```
var v1 = {};
var v3 = new Proxy(parseFloat,v1);
v3.__proto__ = v3;
var v6 = "aa".constructor;
var v7 = parseFloat & v6;
```

akosthekiss:
Could you please try whether you get SO even if you set a limit for the stack? (Note: You can use `--stack-limit` option when building the engine to limit the maximum amount of stack that the engine can use.)

nszetei:
> Could you please try whether you get SO even if you set a limit for the stack? (Note: You can use `--stack-limit` option when building the engine to limit the maximum amount of stack that the engine can use.)

Yes. Unlike https://github.com/jerryscript-project/jerryscript/issues/3783 here I got SO in both cases (e.g. with `--stack-limit=10`).

nszetei:
Just to have it grouped together, a PoC for `ecma_proxy_object_has`:

```
function main() {
var v1 = [13.37,13.37,13.37,13.37];
var v4 = {isExtensible:Infinity};
var v6 = new Proxy(WeakMap,v4);
v6.__proto__ = v1;
v1.__proto__ = v6;
with (v1) {
    valueOf = 0;
}
}
main();
```

PR #3796, opened 2020-05-25
Add stack limit check to proxy operations

dbatyai:
Fixes #3785.

PR #4530, opened 2021-01-23
Fixes the Clang compiler caused deadloop issue

lygstate:
Clang mis-optimization tail calling to ecma_op_object_get_with_receiver:
When ecma_op_object_get_with_receiver are called, the stack size will not increase.

That's cause testcase tests/jerry/es.next/regression-test-issue-3785.js to be deadloop
instead of stack overflow exception.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -213,6 +213,11 @@ if(USING_GCC OR USING_CLANG)
   endif()
   jerry_add_compile_warnings(all extra format-nonliteral init-self conversion sign-conversion format-security missing-declarations shadow strict-prototypes undef old-style-definition)
   jerry_add_compile_flags(-Wno-stack-protector -Wno-attributes -Werror)
+
+  # When clang do tail optimization calling to ecma_op_object_get_with_receiver,
+  # the stack size will not increase. Use -fno-optimize-sibling-calls
+  # to disable the tail optimization.
+  jerry_add_compile_flags(-fno-optimize-sibling-calls)
 endif()
 
 if(USING_GCC)
--- a/tools/run-tests.py
+++ b/tools/run-tests.py
@@ -79,6 +79,9 @@ JERRY_UNITTESTS_OPTIONS = [
 
 # Test options for jerry-tests
 JERRY_TESTS_OPTIONS = [
+    Options('jerry_tests-es.next',
+            OPTIONS_COMMON + OPTIONS_PROFILE_ESNEXT + OPTIONS_STACK_LIMIT + OPTIONS_GC_MARK_LIMIT
+            + OPTIONS_MEM_STRESS),
     Options('jerry_tests-es.next-debug',
             OPTIONS_COMMON + OPTIONS_PROFILE_ESNEXT + OPTIONS_DEBUG + OPTIONS_STACK_LIMIT + OPTIONS_GC_MARK_LIMIT
             + OPTIONS_MEM_STRESS),


lygstate:
Here is the full description of the bug:
https://lists.llvm.org/pipermail/llvm-dev/2021-January/148017.html

ossy-szeged:
I think it isn't a bug in clang. Maybe a bug in jerryscript or a bug in the testcase.
Maybe @dbatyai could help to clarify if the spec says anything about this situation.
( He added tests/jerry/es.next/regression-test-issue-3785.js to fix #3785 in #3796 )

lygstate:
> Maybe @dbatyai could help to clarify if the spec says anything about this situation.

Yeap, it's not a bug of clang, I've verified that.

lygstate:
Maybe we need use something like this to disable tail-optimization:

https://chromium.googlesource.com/external/github.com/abseil/abseil-cpp/+/HEAD/absl/base/optimization.h#29
```



// ABSL_BLOCK_TAIL_CALL_OPTIMIZATION
  | //
  | // Instructs the compiler to avoid optimizing tail-call recursion. This macro is
  | // useful when you wish to preserve the existing function order within a stack
  | // trace for logging, debugging, or profiling purposes.
  | //
  | // Example:
  | //
  | //   int f() {
  | //     int result = g();
  | //     ABSL_BLOCK_TAIL_CALL_OPTIMIZATION();
  | //     return result;
  | //   }
  | #if defined(__pnacl__)
  | #define ABSL_BLOCK_TAIL_CALL_OPTIMIZATION() if (volatile int x = 0) { (void)x; }
  | #elif defined(__clang__)
  | // Clang will not tail call given inline volatile assembly.
  | #define ABSL_BLOCK_TAIL_CALL_OPTIMIZATION() __asm__ __volatile__("")
  | #elif defined(__GNUC__)
  | // GCC will not tail call given inline volatile assembly.
  | #define ABSL_BLOCK_TAIL_CALL_OPTIMIZATION() __asm__ __volatile__("")
  | #elif defined(_MSC_VER)
  | #include <intrin.h>
  | // The __nop() intrinsic blocks the optimisation.
  | #define ABSL_BLOCK_TAIL_CALL_OPTIMIZATION() __nop()
  | #else
  | #define ABSL_BLOCK_TAIL_CALL_OPTIMIZATION() if (volatile int x = 0) { (void)x; }
  | #endif
 


```

zherczeg:
Please provide a backtrace so we can understand what is happening here.

lygstate:
Here is the backtrace
```
 jerry-core/ecma/operations/ecma-proxy-object.c | 4 +++-
 1 file changed, 3 insertions(+), 1 deletion(-)

diff --git a/jerry-core/ecma/operations/ecma-proxy-object.c
b/jerry-core/ecma/operations/ecma-proxy-object.c
index c5d299c6..77aed25c 100644
--- a/jerry-core/ecma/operations/ecma-proxy-object.c
+++ b/jerry-core/ecma/operations/ecma-proxy-object.c
@@ -1108,7 +1108,9 @@ ecma_proxy_object_get (ecma_object_t *obj_p, /**<
proxy object */
   if (ecma_is_value_undefined (trap))
   {
     ecma_object_t *target_obj_p = ecma_get_object_from_value
(proxy_obj_p->target);
-    return ecma_op_object_get_with_receiver (target_obj_p, prop_name_p,
receiver);
+    ecma_value_t value = ecma_op_object_get_with_receiver (target_obj_p,
prop_name_p, receiver);
+    fflush (stdout);
+    return value;
   }

   ecma_object_t *func_obj_p = ecma_get_object_from_value (trap);
```
In which condition,  ecma_op_object_get_with_receiver tail calling will not
increase the stack-size?
I have a test-case that cause infinite loop, because that in normal
conition, it's expect to raise an stack-exceed error but
now with llvm-compiled release build binary, it's turn out to be infinite
loop because  ecma_op_object_get_with_receiver
will not increase the stack size anymore, once
 ecma_op_object_get_with_receiver getting called, it's returned to the
previous stack position。
Reproduced on OSX/Ubuntu 20.04

Stack-trace-base
```
jerry!ecma_proxy_object_get
(\Users\lygstate\work\typescript\jerryscript\jerry-core\ecma\operations\ecma-proxy-object.c:1111)
jerry!ecma_op_object_get_with_receiver
(\Users\lygstate\work\typescript\jerryscript\jerry-core\ecma\operations\ecma-objects.c:790)
jerry!ecma_op_object_get
(\Users\lygstate\work\typescript\jerryscript\jerry-core\ecma\operations\ecma-objects.c:763)
jerry!vm_op_get_value
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:114)
jerry!vm_loop
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:2799)
jerry!vm_execute
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:4939)
jerry!vm_run
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:5046)
jerry!vm_run_global
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:328)
jerry!jerry_run
(\Users\lygstate\work\typescript\jerryscript\jerry-core\api\jerry.c:671)
jerry!main
(\Users\lygstate\work\typescript\jerryscript\jerry-main\main-unix.c:156)
libdyld.dylib!start (Unknown Source:0)
libdyld.dylib!start (Unknown Source:0)
```

stack-trace-new, it's hope  ecma_op_object_get_with_receiver to deeper, but
turn out to step-back.
```
jerry!ecma_op_object_get_with_receiver
(\Users\lygstate\work\typescript\jerryscript\jerry-core\ecma\operations\ecma-objects.c:784)
jerry!ecma_op_object_get
(\Users\lygstate\work\typescript\jerryscript\jerry-core\ecma\operations\ecma-objects.c:763)
jerry!vm_op_get_value
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:114)
jerry!vm_loop
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:2799)
jerry!vm_execute
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:4939)
jerry!vm_run
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:5046)
jerry!vm_run_global
(\Users\lygstate\work\typescript\jerryscript\jerry-core\vm\vm.c:328)
jerry!jerry_run
(\Users\lygstate\work\typescript\jerryscript\jerry-core\api\jerry.c:671)
jerry!main
(\Users\lygstate\work\typescript\jerryscript\jerry-main\main-unix.c:156)
libdyld.dylib!start (Unknown Source:0)
libdyld.dylib!start (Unknown Source:0)
```

Same clang , build it with debug flag, the issue gone. I am wonder if there
is a misoptimization here

The c code
https://gist.github.com/lygstate/b554e74a22353d50a24240128f875474

The full source tree
https://github.com/lygstate/jerryscript/tree/osx-clang-bug

rerobika:
@galpeter could you please share your progress about this issue? AFAIK you also tried some other approach.

galpeter:
yeah I have a way to resolve this. The macros @lygstate mentioned were used to add the tail call disable at given points and also added extra stack checks in the Proxy internal methods. Will upload soon. Note: this is not a clang specific thing and is a very edge case to construct a proxy which recurses (infinite times).

lygstate:
Close as proper fix exist.

PR #4539, opened 2021-01-25
Fix literal index decoding for CBC_MOV_IDENT

### LINKED ISSUE #4532: How to generate bytecode with jerryscript?
Hi! 
As far as I know, when a Js engine executes Javascript code. It will firstly parse the Javascript code into bytecode, then use a intepreter to execute the bytecode. Is this the same for jerryscript? If so, how can I use jerry to get the parsed bytecode?

rerobika:
This patch fixes #4532.


rerobika:
Note: We need that large testcase to trigger this issue, so I kept the reported tc.

zherczeg (approved):
LGTM

dbatyai (approved):
LGTM

PR #4456, opened 2021-01-12
Add --build-debug options for run-tests.py script.

lygstate:
This makes have option to running test262-es2015 and test262-esnext with debug version
of jerryscript.

Depends
https://github.com/jerryscript-project/jerryscript/pull/4495
https://github.com/jerryscript-project/jerryscript/pull/4486
https://github.com/jerryscript-project/jerryscript/pull/4390

JerryScript-DCO-1.0-Signed-off-by: Yonggang Luo luoyonggang@gmail.com

PR #4542, opened 2021-01-26
Remove a wrong test not to slow down test262-es2015

ossy-szeged:
Since ES2018 iterator's next method is called once during the prologue of iteration,
rather than during each step. The test is incorrect and stuck in an infinite loop.
https://github.com/tc39/test262/pull/1248 fixed the test and it passes on test262-esnext.

Removing test/language/statements/for-of/iterator-next-reference.js from test262-es2015
is necessary, because it won't pass ever and only slow down testing.

### DIFF
--- a/tools/runners/run-test-suite-test262.py
+++ b/tools/runners/run-test-suite-test262.py
@@ -97,6 +97,15 @@ def prepare_test262_test_suite(args):
         if os.path.isdir(path_to_remove):
             shutil.rmtree(path_to_remove)
 
+    # Since ES2018 iterator's next method is called once during the prologue of iteration,
+    # rather than during each step. The test is incorrect and stuck in an infinite loop.
+    # https://github.com/tc39/test262/pull/1248 fixed the test and it passes on test262-esnext.
+    if args.es2015:
+        test_to_remove = 'test/language/statements/for-of/iterator-next-reference.js'
+        path_to_remove = os.path.join(args.test_dir, os.path.normpath(test_to_remove))
+        if os.path.isfile(path_to_remove):
+            os.remove(path_to_remove)
+
     return 0
 
 


ossy-szeged:
Note: It is needed to be able to increase timeout for debug tests. (#4456)
Telling the truth, this workaround isn't nice, but unfortunately there is no other way to ignore this test.

zherczeg:
If we land this, and move to another test262 version, will there be a notification to remove this workaround?

ossy-szeged:
> If we land this, and move to another test262 version, will there be a notification to remove this workaround?

It affects only the es2015 version of test262 which is a fixed git revision.
The esnext version of test262 already contains the fixed test and it passes.

We can drop this workaround when we drop es2015 version of test262.

zherczeg:
Do we still need it? We don't have es2015 support anymore.

ossy-szeged:
> Do we still need it? We don't have es2015 support anymore.

I don't think if we should drop test262-es2015 until we make all supported feature compatible with ES11 (ES2020),
there are still ~1000 uncategorized test failures in test262-esnext exclude list. Now we know that we are at least
compatible with the older ES2015 spec. If we dropped it, we wouldn't know if these features are compatible with any spec.

zherczeg (approved):
LGTM

lygstate:
> > Do we still need it? We don't have es2015 support anymore.
> 
> I don't think if we should drop test262-es2015 until we make all supported feature compatible with ES11 (ES2020),
> there are still ~1000 uncategorized test failures in test262-esnext exclude list. Now we know that we are at least
> compatible with the older ES2015 spec. If we dropped it, we wouldn't know if these features are compatible with any spec.

Hi, maybe using a exclude-list and pass into test262-harness instead?

ossy-szeged:
> Hi, maybe using a exclude-list and pass into test262-harness instead?

On the CI we always run tests with update option (to see not only regressions, but progressions too), which doesn't use exclude list at all.

lygstate:
> > Hi, maybe using a exclude-list and pass into test262-harness instead?
> 
> On the CI we always run tests with update option (to see not only regressions, but progressions too), which doesn't use exclude list at all.

Make sense, after this pull request, we can increase the timeout value to a very large value such as 5minutes,

LGTM

lygstate (approved):
LGTM

rerobika (approved):

PR #4543, opened 2021-01-26
Stop after the first matching property in the prototype chain

zherczeg:
If multiple properties with the same name is in the prototype chain, only the first one should be inspected


rerobika (approved):

ossy-szeged (approved):
LGTM

PR #4548, opened 2021-01-29
Add unittest to check realm object and realm this types

galpeter:
Make sure that realm object and realm this types are tested.

### DIFF
--- a/tests/unit-core/test-api-value-type.c
+++ b/tests/unit-core/test-api-value-type.c
@@ -124,6 +124,33 @@ main (void)
     jerry_release_value (object_bigint);
   }
 
+  if (jerry_is_feature_enabled (JERRY_FEATURE_REALM))
+  {
+    jerry_value_t new_realm = jerry_create_realm ();
+    jerry_value_t old_realm = jerry_set_realm (new_realm);
+
+    jerry_type_t new_realm_type = jerry_value_get_type (new_realm);
+    TEST_ASSERT (new_realm_type == JERRY_TYPE_OBJECT);
+
+    jerry_value_t new_realm_this = jerry_realm_get_this (new_realm);
+    jerry_type_t new_realm_this_type = jerry_value_get_type (new_realm_this);
+    TEST_ASSERT (new_realm_this_type == JERRY_TYPE_OBJECT);
+    jerry_release_value (new_realm_this);
+
+    jerry_type_t old_realm_type = jerry_value_get_type (old_realm);
+    TEST_ASSERT (old_realm_type == JERRY_TYPE_OBJECT);
+
+    jerry_release_value (new_realm);
+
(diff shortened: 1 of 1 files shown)


tests/unit-core/test-api-value-type.c
```diff
@@ -124,6 +124,33 @@ main (void)
...
+  if (jerry_is_feature_enabled (JERRY_FEATURE_REALM))
+  {
+    jerry_value_t new_realm = jerry_create_realm ();
+    jerry_value_t old_realm = jerry_set_realm (new_realm);
+
+    jerry_type_t new_realm_type = jerry_value_get_type (new_realm);
+    TEST_ASSERT (new_realm_type == JERRY_TYPE_OBJECT);
+
+    jerry_value_t new_realm_this = jerry_realm_get_this (new_realm);
+    jerry_type_t new_realm_this_type = jerry_value_get_type (new_realm_this);
```
zherczeg, line 136: By default the `this` is the same as `realm`.

tests/unit-core/test-api-value-type.c
```diff
@@ -124,6 +124,33 @@ main (void)
...
+
+    jerry_type_t old_realm_type = jerry_value_get_type (old_realm);
+    TEST_ASSERT (old_realm_type == JERRY_TYPE_OBJECT);
+
+    jerry_release_value (new_realm);
+
+    /* Restore the old realm as per docs */
+    jerry_set_realm (old_realm);
+
+    jerry_value_t old_realm_this = jerry_realm_get_this (old_realm);
```
zherczeg, line 148: You should not use old_realm after it is restoered.

zherczeg (changes requested):
I would put this into:
https://github.com/jerryscript-project/jerryscript/blob/master/tests/unit-core/test-realm.c

What is the case that is tested here, but not in the test file?

galpeter:
> What is the case that is tested here, but not in the test file?

There is an API method `jerry_value_get_type` which returns the type of a value. There were no tests where a realm's type is queried. Previously there a problem where an value was returned but the API method was unable to query it's type. To avoid such cases it would be good to always add type test for various (at least create) methods.

And all value type tests are in this file.

galpeter:
In addition I've just realized, the `jerry_object_get_type` for realms should also be tested.

tests/unit-core/test-api-value-type.c
galpeter: Okay and? I've explicitly testing it's type.

tests/unit-core/test-api-value-type.c
galpeter: yeah I've wanted to add this before the set_realm call above. Will fix it.

galpeter:
@zherczeg I've updated the PR

zherczeg (approved):
LGTM

rerobika (approved):

PR #4555, opened 2021-01-30
Introduce generic backtrace capturing

zherczeg:
Remove jerry_get_backtrace_from function

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -245,6 +245,15 @@ memory blocks but the performance may drop after the garbage collection.
 
 *New in version 2.0*.
 
+## jerry_backtrace_frame_types_t
+
+List of backtrace frame types returned by
+[jerry_backtrace_get_frame_type](#jerry_backtrace_get_frame_type).
+
+ - JERRY_BACKTRACE_FRAME_JS - indicates that the frame is created for a JavaScript function/method
+
+*New in version [[NEXT_RELEASE]]*.
+
 ## jerry_generate_snapshot_opts_t
 
 Flags for [jerry_generate_snapshot](#jerry_generate_snapshot) and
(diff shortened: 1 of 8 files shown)


rerobika (approved):

jerry-core/include/jerryscript-core.h
```diff
@@ -754,12 +787,20 @@ void jerry_heap_free (void *mem_p, size_t size);
  */
 jerry_context_t *jerry_create_context (uint32_t heap_size, jerry_context_alloc_t alloc, void *cb_data_p);
 
+/**
+ * Backtrace functions.
+ */
+jerry_value_t jerry_get_backtrace (uint32_t max_depth);
+void jerry_capture_backtrace (jerry_backtrace_callback_t callback, void *user_p);
```
galpeter, line 794: Most of the backtrace functions start with the `jerry_backtrace` prefix. Could we rename this to `jerry_backtrace_capture`? Sidenote: "capture" seems a bit off for me because the engine already knows the backtrace we just access/read it here, but at the moment I don't know any good alternatives for the name.

docs/02.API-REFERENCE.md
```diff
@@ -9560,8 +9838,8 @@ jerry_realm_set_this (jerry_value_t realm_value, jerry_value_t this_value)
 - `realm_value` - realm value
 - `this_value` - new this value
 - return
-  - type error - if realm_value is not a realm or this_value is not object
-  - true - otherwise
+  - type error, if realm_value is not a realm or this_value is not object
+  - true, otherwise
```
galpeter, line 9842: AFAIK, this change ins not related to the PR.

docs/02.API-REFERENCE.md
```diff
@@ -234,6 +234,15 @@ memory blocks but the performance may drop after the garbage collection.
 
 *New in version 2.0*.
 
+## jerry_backtrace_frame_types_t
+
+List of backtrace frame types returned by
+[jerry_backtrace_get_frame_type](#jerry_backtrace_get_frame_type).
+
+ - JERRY_BACKTRACE_JS_CODE - frame is created for running ECMAScript code
```
galpeter, line 242: * I was wondering if having a `JERRY_BACKTRACE_UNKNOWN` like entry would make sense? Which would indicate in the future that there was some kind of error/or the frame's type is not known (eg.: native/external, incorrect input).
* Would the `JERRY_BACKTRACE_FRAME_JS` be a better name? This way the "frame" as information is in the name. The "_CODE" does not provide extra info in the current variant.
* The comment could be: `indicates that the frame is created for a  function/method implemented in JavaScript`.

docs/02.API-REFERENCE.md
zherczeg: I would not create another pr for this small change.

jerry-core/include/jerryscript-core.h
zherczeg: I think capture term is used in connection with backtrace frequently

docs/02.API-REFERENCE.md
zherczeg: There should never been an unknown frame, if we add more frame types for the engine, this api should be extended as well.

zherczeg:
Thnak you for the review, the patch is updated.

docs/02.API-REFERENCE.md
galpeter: AFIK: All of the return descriptions should be updated not just this one.

galpeter (approved):
lgtm

docs/02.API-REFERENCE.md
zherczeg: Ok, I restored it.

PR #4558, opened 2021-02-03
Ensure that [[Set]] operation does not reject for TypedArray objects

rerobika:

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -531,12 +531,10 @@ ecma_op_object_find_own (ecma_value_t base_value, /**< base value */
       /* ES2015 9.4.5.4 */
       if (ecma_object_is_typedarray (object_p))
       {
-#if ENABLED (JERRY_ESNEXT)
         if (ecma_prop_name_is_symbol (property_name_p))
         {
           break;
         }
-#endif /* ENABLED (JERRY_ESNEXT) */
 
         uint32_t array_index = ecma_string_get_array_index (property_name_p);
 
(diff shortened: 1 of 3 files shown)


jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -531,12 +531,10 @@ ecma_op_object_find_own (ecma_value_t base_value, /**< base value */
       /* ES2015 9.4.5.4 */
       if (ecma_object_is_typedarray (object_p))
       {
-#if ENABLED (JERRY_ESNEXT)
```
zherczeg, line 534: I think typedarray can be enabled without esnext.

zherczeg (changes requested):

jerry-core/ecma/operations/ecma-objects.c
rerobika: If it is, that's a huge problem.

rerobika: No you cannot. `config.h:588` requires JERRY_ESNEXT for JERRY_BUILTIN_TYPEDARRAY

ossy-szeged: As far as I remember we don't support any ES5.1+ feature without ESNEXT enabled officially. Maybe it builds and works fine, but I think it isn't tested at all. If we would like to support eg TypedArray with ES5.1 profile, this configuration should have at least a buildoption test.

rerobika: No, the build does not work either.

zherczeg: Ok.

zherczeg (approved):
LGTM

dbatyai (approved):
LGTM

PR #4561, opened 2021-02-04
Fix failing Date tests in test262-esnext

pmarkee:

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
@@ -445,7 +445,13 @@ ecma_builtin_date_prototype_dispatch_set (uint16_t builtin_routine_id, /**< buil
 #if ENABLED (JERRY_BUILTIN_ANNEXB)
       case ECMA_DATE_PROTOTYPE_SET_YEAR:
       {
-        year = converted_number[0];
+        if (ecma_number_is_nan (converted_number[0]))
+        {
+          *ECMA_GET_INTERNAL_VALUE_POINTER (ecma_number_t, ext_object_p->u.class_prop.u.value) = converted_number[0];
+          return ecma_make_number_value (converted_number[0]);
+        }
+
+        year = ecma_number_trunc (converted_number[0]);
         if (year >= 0 && year <= 99)
         {
           year += 1900;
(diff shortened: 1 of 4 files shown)


pmarkee:
Note: `tests/jerry/date-tostring.js` needs to be changed because the es11 standard specifies the expected format of `Date.prototype.toDateString()` and `Date.prototype.toTimeString()`, while es6 did not.

tests/jerry/date-tostring.js
```diff
@@ -58,8 +58,8 @@ catch (e)
 
 assert (new Date (NaN).toTimeString () == "Invalid Date");
 assert (new Date (Number.POSITIVE_INFINITY).toString () === "Invalid Date");
-assert (new Date ("2015-02-13").toTimeString () == "00:00:00.000");
-assert (new Date ("2015-07-08T11:29:05.023").toTimeString () == "11:29:05.023");
```
rerobika, line 62: The change is good here, but could you please keep the originals and move these 4 asserts into `tests/jerry/es5.1/date-tostring.js`?

rerobika (changes requested):

tests/jerry/date-tostring.js
pmarkee: The original asserts will fail with every profile because of the changed format, unless we keep the old implementation for the es5.1 and es6 builds. Only esnext specifies the correct format so I think it's unnecessary to have two different implementations. But of course I can do that if you prefer.

rerobika (approved):

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
```diff
@@ -445,7 +445,12 @@ ecma_builtin_date_prototype_dispatch_set (uint16_t builtin_routine_id, /**< buil
 #if ENABLED (JERRY_BUILTIN_ANNEXB)
       case ECMA_DATE_PROTOTYPE_SET_YEAR:
       {
-        year = converted_number[0];
+        if (ecma_number_is_nan (converted_number[0]))
+        {
+          *ECMA_GET_INTERNAL_VALUE_POINTER (ecma_number_t, ext_object_p->u.class_prop.u.value) = converted_number[0];
+          return ecma_make_number_value (converted_number[0]);
+        }
+        year = ecma_number_trunc (converted_number[0]);
```
zherczeg, line 453: Put a newline above this line.

zherczeg (changes requested):

jerry-core/ecma/builtin-objects/ecma-builtin-date-prototype.c
pmarkee: Done.

zherczeg (approved):
LGTM

PR #4562, opened 2021-02-04
Add non-standard behaviour support for Proxies

zherczeg:


jerry-core/ecma/base/ecma-globals.h
```diff
@@ -2176,6 +2176,18 @@ do \
...
+/**
+ * Proxy object flags.
+ */
+typedef enum
+{
+  ECMA_PROXY_SKIP_GET_CHECKS = (1u << 0), /**< skip [[Get]] result checks */
+  ECMA_PROXY_SKIP_GET_OWN_PROPERTY_CHECKS = (1u << 1), /**< skip [[GetOwnProperty]] result checks */
+  ECMA_PROXY_IS_CALLABLE = (1u << 2), /**< proxy is callable */
+  ECMA_PROXY_IS_CONSTRUCTABLE = (1u << 3), /**< proxy is constructable */
+} ecma_proxy_flag_types_t;
```
galpeter, line 2189: * The comment below is now invalid for the callable/constructable elements.
* The two always "used" flags should be the first ones (ECMA_PROXY_IS_CALLABLE, ECMA_PROXY_IS_CONSTRUCTABLE) these will never change, however later on if we add other skip it would look strange having the two is flags between them,

rerobika (approved):

tests/unit-core/test-realm.c
```diff
@@ -13,8 +13,7 @@
  * limitations under the License.
  */
 
-#include "ecma-globals.h"
-#include "ecma-helpers.h"
+#include "jerryscript.h"
```
galpeter, line 16: Why was this changed in this PR?

jerry-core/include/jerryscript-core.h
```diff
@@ -471,6 +471,15 @@ typedef enum
   JERRY_ITERATOR_TYPE_SET,          /**< Set iterator */
 } jerry_iterator_type_t;
 
+/**
+ * JerryScript special Proxy object options.
+ */
+typedef enum
+{
+  JERRY_PROXY_SKIP_GET_CHECKS = (1u << 0), /**< skip [[Get]] result checks */
```
galpeter, line 479: It would be good to have a NULL value at the start. or something like JERRY_PROXY_ALL_CHECKS. The main reason would be to not have the "0" as a value which can disable things.

dbatyai (approved):
LGTM

docs/02.API-REFERENCE.md
```diff
@@ -5392,6 +5405,70 @@ main (void)
 **See also**
 
 - [jerry_value_is_proxy](#jerry_value_is_proxy)
+- [jerry_create_special_proxy](#jerry_create_special_proxy)
+- [jerry_release_value](#jerry_release_value)
+
+
+## jerry_create_special_proxy
```
galpeter, line 5412: Maybe: `jerry_create_proxy_special`?

jerry-core/ecma/base/ecma-globals.h
zherczeg: These are internal flags, the `JERRY_*` flags are the external. The internal flags can be reorganized in any way any time.

tests/unit-core/test-realm.c
zherczeg: I noticed a small mistake.

PR #4565, opened 2021-02-04
Improve ecma_reject to give more detailed error messages

rerobika:
Note: TypedArray.[[DefineOwnProperty]] has been slightly reworked (without semantical changes) for better error messages.

### DIFF
--- a/jerry-core/ecma/operations/ecma-objects.h
+++ b/jerry-core/ecma/operations/ecma-objects.h
@@ -27,6 +27,35 @@
  * @{
  */
 
+#if JERRY_ERROR_MESSAGES
+/**
+ * Reject with TypeError depending on 'is_throw' with the given format
+ */
+#define ECMA_REJECT_WITH_FORMAT(is_throw, msg, ...) \
+  ((is_throw) ? ecma_raise_standard_error_with_format (ECMA_ERROR_TYPE, (msg), __VA_ARGS__) : ECMA_VALUE_FALSE)
+
+/**
+ * Reject with TypeError depending on 'is_throw' with the given message
+ */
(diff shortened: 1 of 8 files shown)


dbatyai (approved):

jerry-core/ecma/operations/ecma-objects.h
```diff
@@ -27,6 +27,35 @@
...
+
+/**
+ * Reject with TypeError depending on is_throw flags wit the given message
+ */
+#define ECMA_REJECT(is_throw, msg) \
+  ((is_throw) ? ecma_raise_type_error (NULL) : ECMA_VALUE_FALSE)
+#endif /* JERRY_ERROR_MESSAGES */
+
+ecma_value_t ecma_raise_property_redefinition (ecma_string_t *property_name_p, uint16_t flags);
+ecma_value_t ecma_raise_readonly_assignment (ecma_string_t *property_name_p, bool is_throw);
```
galpeter, line 57: To make it consistent should we also pass `flags` here? (just like in the `ecma_raise_property_redefinition` method)

rerobika: Currently `ecma_raise_readonly_assignment ` is only used by `[[Put]]` method where the `is_throw` is passed as a boolean. we should use a ternary to convert it to flags, so I'd leave as it for now.

rerobika:
The patch is rebased.

jerry-core/ecma/operations/ecma-objects.h
galpeter: True, ok. We'll take a look at it later if we really want to change it.

galpeter (approved):
lgtm

PR #4567, opened 2021-02-05
Add cross-realm support to RegExp.prototype

kisbg:
JerryScript-DCO-1.0-Signed-off-by: Bence Gabor Kis kisbg@inf.u-szeged.hu

### DIFF
--- a/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
+++ b/jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
@@ -744,7 +744,7 @@ ecma_builtin_regexp_prototype_dispatch_routine (uint8_t builtin_routine_id, /**<
     {
       if (!ecma_object_class_is (obj_p, LIT_MAGIC_STRING_REGEXP_UL))
       {
-        if (ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_REGEXP_PROTOTYPE))
+        if (obj_p == ecma_builtin_get (ECMA_BUILTIN_ID_REGEXP_PROTOTYPE))
         {
           return ecma_make_magic_string_value (LIT_MAGIC_STRING_EMPTY_NON_CAPTURE_GROUP);
         }
@@ -764,7 +764,7 @@ ecma_builtin_regexp_prototype_dispatch_routine (uint8_t builtin_routine_id, /**<
     {
       if (!ecma_object_class_is (obj_p, LIT_MAGIC_STRING_REGEXP_UL))
       {
-        if (ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_REGEXP_PROTOTYPE))
+        if (obj_p == ecma_builtin_get (ECMA_BUILTIN_ID_REGEXP_PROTOTYPE))
         {
           return ECMA_VALUE_UNDEFINED;
         }
--- a/jerry-core/ecma/operations/ecma-objects.c
+++ b/jerry-core/ecma/operations/ecma-objects.c
@@ -2605,6 +2605,42 @@ ecma_op_object_enumerate (ecma_object_t *obj_p) /**< object */
   return return_names_p;
 } /* ecma_op_object_enumerate */
 
+#ifndef JERRY_NDEBUG
+
+/**
+ * Check if passed object is the instance of specified built-in.
+ *
+ * @return true  - if the object is instance of the specified built-in
+ *         false - otherwise
+ */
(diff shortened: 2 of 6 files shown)


jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -744,7 +744,7 @@ ecma_builtin_regexp_prototype_dispatch_routine (uint8_t builtin_routine_id, /**<
     {
       if (!ecma_object_class_is (obj_p, LIT_MAGIC_STRING_REGEXP_UL))
       {
-        if (ecma_builtin_is (obj_p, ECMA_BUILTIN_ID_REGEXP_PROTOTYPE))
```
rerobika, line 747: `ecma_builtin_is ` is a widely used function, I think we should update it to properly support built-in objects from different realms. Maybe it would fix other testcases as well.

rerobika (changes requested):

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
zherczeg: `builtin_get` checks type regardless of realm. Is this the problem? A regexp coming from a different realm is not a regexp?

rerobika: @zherczeg The key from the standard is `If SameValue(R, %RegExpPrototype%)`

Where `%RegExpPrototype%` comes from the current realm's intrinsics.  

```js
/*---
esid: sec-get-regexp.prototype.dotall
description: Invoked on a cross-realm object
info: |
    get RegExp.prototype.dotAll

    1. Let R be the this value.
    2. If Type(R) is not Object, throw a TypeError exception.
    3. If R does not have an [[OriginalFlags]] internal slot, then
      a. If SameValue(R, %RegExpPrototype%) is true, return undefined.
      b. Otherwise, throw a TypeError exception.
features: [regexp-dotall, cross-realm]
---*/

var dotAll = Object.getOwnPropertyDescriptor(RegExp.prototype, 'dotAll').get;
var other = $262.createRealm().global;
var otherRegExpProto = other.RegExp.prototype;
var otherRegExpGetter = Object.getOwnPropertyDescriptor(otherRegExpProto, 'dotAll').get;

assert.throws(TypeError, function() {
  dotAll.call(otherRegExpProto);
}, 'cross-realm RegExp.prototype');

assert.throws(other.TypeError, function() {
  otherRegExpGetter.call(RegExp.prototype);
}, 'cross-realm RegExp.prototype getter method against primary realm RegExp.prototype');
```

zherczeg: In short, a regex from another realm is regex, but the regex prototype from another realm is not a regexp prototype?

jerry-core/ecma/builtin-objects/ecma-builtin-regexp-prototype.c
```diff
@@ -764,11 +764,10 @@ ecma_builtin_regexp_prototype_dispatch_routine (uint8_t builtin_routine_id, /**<
...
+        if (obj_p == ecma_builtin_get (ECMA_BUILTIN_ID_REGEXP_PROTOTYPE))
         {
           return ECMA_VALUE_UNDEFINED;
         }
-
```
zherczeg, line 771: Don't delete this line.

zherczeg (changes requested):

zherczeg (approved):
LGTM

rerobika (changes requested):
After this, the `ecma_builtin_is` is only used by `ecma_object_check_class_name_is_object`. So please remove it from `ecma-builtins.h` and make it static above `ecma_object_check_class_name_is_object`.

jerry-core/ecma/operations/ecma-objects.c
```diff
@@ -2605,6 +2605,42 @@ ecma_op_object_enumerate (ecma_object_t *obj_p) /**< object */
   return return_names_p;
 } /* ecma_op_object_enumerate */
 
+#if JERRY_NDEBUG
```
rerobika, line 2608: This should be `#ifndef JERRY_NDEBUG`

kisbg: Yes, you are right!

rerobika (approved):

PR #4568, opened 2021-02-05
Improve Proxy recrusion check in their internal methods.

galpeter:
In some parts of Proxy internal methods the compiler can do a bit of
tail call optimalization which would be ok, but the current stack limit
check framework in Jerry uses the stack pointer to calculate the stack depth.
This way of stack depth calculation is affected by the tail call optimalization
(as the stack does not increase).

By disabling the tail call optimalization at given points the stack limit calculation
will work as expected. This causes a bit of stack overhead, but the Proxy in it self
is a fairly big chunk of code and this stack limit would only be relevant if the Proxy
already does recursion which already very edge case.

The stack limit (--stack-limit=..) should be enabled to correctly report such stack
depth errors.

### DIFF
--- a/jerry-core/ecma/base/ecma-globals.h
+++ b/jerry-core/ecma/base/ecma-globals.h
@@ -2268,16 +2268,29 @@ typedef struct
 #if (JERRY_STACK_LIMIT != 0)
 /**
  * Check the current stack usage. If the limit is reached a RangeError is raised.
+ * The macro argument specifies the return value which is usally ECMA_VALUE_ERROR or NULL.
  */
-#define ECMA_CHECK_STACK_USAGE() \
+#define ECMA_CHECK_STACK_USAGE_RETURN(RETURN_VALUE) \
 do \
 { \
   if (ecma_get_current_stack_usage () > CONFIG_MEM_STACK_LIMIT) \
   { \
-    return ecma_raise_range_error (ECMA_ERR_MSG ("Maximum call stack size exceeded")); \
+    ecma_raise_range_error (ECMA_ERR_MSG ("Maximum call stack size exceeded")); \
+    return RETURN_VALUE; \
   } \
 } while (0)
+
+/**
+ * Specialized version of ECMA_CHECK_STACK_USAGE_RETURN which returns ECMA_VALUE_ERROR.
+ * This version should be used in most cases.
+ */
+#define ECMA_CHECK_STACK_USAGE() ECMA_CHECK_STACK_USAGE_RETURN(ECMA_VALUE_ERROR)
 #else /* JERRY_STACK_LIMIT == 0) */
+/**
+ * If the stack limit is unlimited, this check is an empty macro.
+ */
+#define ECMA_CHECK_STACK_USAGE_RETURN(RETURN_VALUE)
+
 /**
  * If the stack limit is unlimited, this check is an empty macro.
  */
--- a/jerry-core/ecma/operations/ecma-proxy-object.c
+++ b/jerry-core/ecma/operations/ecma-proxy-object.c
@@ -263,6 +263,7 @@ ecma_value_t
 ecma_proxy_object_get_prototype_of (ecma_object_t *obj_p) /**< proxy object */
 {
   JERRY_ASSERT (ECMA_OBJECT_IS_PROXY (obj_p));
+  ECMA_CHECK_STACK_USAGE ();
 
   ecma_proxy_object_t *proxy_obj_p = (ecma_proxy_object_t *) obj_p;
 
@@ -284,7 +285,9 @@ ecma_proxy_object_get_prototype_of (ecma_object_t *obj_p) /**< proxy object */
   /* 7. */
   if (ecma_is_value_undefined (trap))
   {
-    return ecma_builtin_object_object_get_prototype_of (target_obj_p);
+    ecma_value_t result = ecma_builtin_object_object_get_prototype_of (target_obj_p);
+    JERRY_BLOCK_TAIL_CALL_OPTIMIZATION ();
+    return result;
   }
 
   ecma_object_t *func_obj_p = ecma_get_object_from_value (trap);
@@ -373,6 +376,7 @@ ecma_proxy_object_set_prototype_of (ecma_object_t *obj_p, /**< proxy object */
                                     ecma_value_t proto) /**< new prototype object */
 {
   JERRY_ASSERT (ECMA_OBJECT_IS_PROXY (obj_p));
+  ECMA_CHECK_STACK_USAGE ();
 
   /* 1. */
   JERRY_ASSERT (ecma_is_value_object (proto) || ecma_is_value_null (proto));
@@ -399,10 +403,14 @@ ecma_proxy_object_set_prototype_of (ecma_object_t *obj_p, /**< proxy object */
   {
     if (ECMA_OBJECT_IS_PROXY (target_obj_p))
     {
-      return ecma_proxy_object_set_prototype_of (target_obj_p, proto);
+      ecma_value_t result = ecma_proxy_object_set_prototype_of (target_obj_p, proto);
+      JERRY_BLOCK_TAIL_CALL_OPTIMIZATION ();
+      return result;
     }
 
-    return ecma_op_ordinary_object_set_prototype_of (target_obj_p, proto);
+    ecma_value_t result = ecma_op_ordinary_object_set_prototype_of (target_obj_p, proto);
+    JERRY_BLOCK_TAIL_CALL_OPTIMIZATION ();
+    return result;
   }
 
   ecma_object_t *func_obj_p = ecma_get_object_from_value (trap);
@@ -488,6 +496,7 @@ ecma_value_t
 ecma_proxy_object_is_extensible (ecma_object_t *obj_p) /**< proxy object */
 {
   JERRY_ASSERT (ECMA_OBJECT_IS_PROXY (obj_p));
+  ECMA_CHECK_STACK_USAGE ();
 
   ecma_proxy_object_t *proxy_obj_p = (ecma_proxy_object_t *) obj_p;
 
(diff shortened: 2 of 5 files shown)


jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -1127,7 +1148,10 @@ ecma_proxy_object_get (ecma_object_t *obj_p, /**< proxy object */
   if (ecma_is_value_undefined (trap))
   {
     ecma_object_t *target_obj_p = ecma_get_object_from_value (proxy_obj_p->target);
-    return ecma_op_object_get_with_receiver (target_obj_p, prop_name_p, receiver);
+    ecma_value_t result = ecma_op_object_get_with_receiver (target_obj_p, prop_name_p, receiver);
+
```
akosthekiss, line 1152: (style) Why the extra empty line? Not used elsewhere.

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -1543,7 +1572,12 @@ ecma_collection_t *
 ecma_proxy_object_own_property_keys (ecma_object_t *obj_p) /**< proxy object */
 {
   JERRY_ASSERT (ECMA_OBJECT_IS_PROXY (obj_p));
-
+/*  if (++JERRY_CONTEXT (proxy_depth) > ECMA_CONFIG_PROXY_RECURSION_DEPTH)
+  {
+    ecma_raise_range_error (ECMA_ERR_MSG ("Maximum call stack size exceeded."));
+    return NULL;
+  }
+*/
```
akosthekiss, line 1580: Is this a leftover?

akosthekiss (commented):
Looks good in general

jerry-core/ecma/operations/ecma-proxy-object.c
galpeter: ohh yeah was shuffling around the commits (during update) and got here accidentally. Will fix it. I don't know how I've missed this. (and this also applies to the style issue above)

rerobika (approved):

lygstate:
Maybe it's worth to contribute this testcase to test262

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -2149,16 +2149,29 @@ typedef struct
...
+ * The macro argument specifies the return value which is usally ECMA_VALUE_ERROR or NULL.
  */
-#define ECMA_CHECK_STACK_USAGE() \
+#define ECMA_CHECK_STACK_USAGE_RETURN(RETURN_VALUE) \
 do \
 { \
   if (ecma_get_current_stack_usage () > CONFIG_MEM_STACK_LIMIT) \
   { \
-    return ecma_raise_range_error (ECMA_ERR_MSG ("Maximum call stack size exceeded")); \
+    ecma_raise_range_error (ECMA_ERR_MSG ("Maximum call stack size exceeded")); \
```
lygstate, line 2159: Just a suggestion, maybe we can disable tail optimization here?

galpeter: I'm not sure what you mean here: At this point the stack usage is over the limit and we will not have any recursion here as we only return with a value. No tail call will happen here (afaik).

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -2149,16 +2149,29 @@ typedef struct
...
+    return RETURN_VALUE; \
   } \
 } while (0)
+
+/**
+ * Specialized version of ECMA_CHECK_STACK_USAGE_RETURN which returns ECMA_VALUE_ERROR.
+ * This version should be used in most cases.
+ */
+#define ECMA_CHECK_STACK_USAGE() ECMA_CHECK_STACK_USAGE_RETURN(ECMA_VALUE_ERROR)
```
lygstate, line 2168: Oh, sorry comment at the wrong place. it's should be here.

jerry-core/ecma/base/ecma-globals.h
lygstate: Sorry , should be in macro
#define ECMA_CHECK_STACK_USAGE() ECMA_CHECK_STACK_USAGE_RETURN(ECMA_VALUE_ERROR)

lygstate:
Also please update
https://github.com/jerryscript-project/jerryscript/blob/fe2b5aae5ff190d456089f761bee1d4a7ff7b1cf/.github/workflows/gh-actions.yml#L50-L52

jerry-core/ecma/base/ecma-globals.h
galpeter: That will do nothing, as the tail call does not happen here but at the point where the return is. This would not turn off the tail call optimization for the whole method.

lygstate: > That will do nothing, as the tail call does not happen here but at the point where the return is. This would not turn off the tail call optimization for the whole method.

OK, see that, thanks

galpeter:
> 
> 
> Also please update
> 
> https://github.com/jerryscript-project/jerryscript/blob/fe2b5aae5ff190d456089f761bee1d4a7ff7b1cf/.github/workflows/gh-actions.yml#L50-L52

true, thanks

lygstate:
Oh type in title `Improve Proxy recrusion check in their internal methods`
Should be `Improve Proxy recursion check in their internal methods`

lygstate:
For mark, fixes https://github.com/jerryscript-project/jerryscript/pull/4530

jerry-core/jrt/jrt.h
```diff
@@ -153,4 +153,36 @@ void JERRY_ATTR_NORETURN jerry_fatal (jerry_fatal_code_t code);
...
+ *     JERRY_BLOCK_TAIL_CALL_OPTIMIZATION();
+ *     return result;
+ *   }
+ *
+ * This macro is intentionally here as jerryscript-compiler.h is a public header and
+ * it does not make sense to expose this macro to the public.
+ */
+#if defined (__clang__) || defined (__GNUC__)
+/* Clang/GCC will not tail call given inline volatile assembly. */
+#define JERRY_BLOCK_TAIL_CALL_OPTIMIZATION() __asm__ __volatile__ ("")
```
zherczeg, line 178: These things seems random, which is hard to maintain. What about doing something before return, e.g clearing a status bit in the jerry context?

lygstate: That's would not stop tail optimization

lygstate: > These things seems random, which is hard to maintain. What about doing something before return, e.g clearing a status bit in the jerry context?

That's would not stop tail optimization

tests/jerry/es.next/proxy-evil-recursion.js
```diff
@@ -0,0 +1,166 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Copyright 2015 the V8 project authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
akosthekiss, line 17: I think that this should be handled. Copy-pasting these 3 lines here has no real meaning as there is no LICENSE file that is referred. Most probably the whole contents of the original LICENSE file should be copied here, similarly to jerry-core/ecma/base/ecma-helpers-errol.c or jerry-math/acos.c .

akosthekiss (changes requested):

tests/jerry/es.next/proxy-evil-recursion.js
galpeter: Ah sorry about this. The test was not copied from v8. I've created it. I just copied the comment block from another test file :(.

akosthekiss: Well and wow. We may need to check that test file (those test files??) then.

jerry-core/jrt/jrt.h
galpeter: Setting status bit seems to be working.

lygstate: > Setting status bit seems to be working.

Then would be good

tests/jerry/es.next/proxy-evil-recursion.js
rerobika: I can agree that we can visit these tests, but in a separate PR.

jerry-core/jrt/jrt.h
```diff
@@ -153,4 +153,33 @@ void JERRY_ATTR_NORETURN jerry_fatal (jerry_fatal_code_t code);
...
+#else /* !defined(__clang__) && !defined(__GNUC__) */
+/* On GCC 10.x this version also works. */
+#define JERRY_BLOCK_TAIL_CALL_OPTIMIZATION() \
+  do { JERRY_CONTEXT (status_flags) |= ECMA_STATUS_API_AVAILABLE; } while (0)
```
zherczeg, line 182: I am not sure this syntax is allowed even in macros

zherczeg (approved):
LGTM with minor change

jerry-core/jrt/jrt.h
galpeter: Could you please elaborate? What do you mean?

zherczeg: newlines
```
do \
{ \
  ... \
} \
while (0)
```

galpeter: Ahh... true. I've updated the PR. thx!

lygstate (approved):

lygstate (approved):

akosthekiss (approved):
LGTM

PR #4569, opened 2021-02-05
Fix incorrect assert expression in lexer_check_property_modifier

galpeter:
In the lexer_check_property_modifier the assert incorrectly does
an assignment instead of a simple equals check.


rerobika (approved):

dbatyai (approved):
LGTM

zherczeg (commented):
LGTM

PR #4570, opened 2021-02-05
Fix deleting native pointers

zherczeg:


rerobika (approved):

tests/unit-core/test-native-pointer.c
```diff
@@ -0,0 +1,116 @@
...
+  jerry_set_object_native_pointer (object_value, NULL, &native_info_2);
+
+  check_native_info (object_value, &native_info_1, global_p);
+  check_native_info (object_value, &native_info_2, NULL);
+
+  jerry_release_value (object_value);
+
+  jerry_gc (JERRY_GC_PRESSURE_HIGH);
+  TEST_ASSERT (global_counter == 1);
+  global_counter = 0;
```
galpeter, line 69: Is there any reason to set this to zero? We don't use this value after this line. Maybe we could remove this and add a check at the end to see how many times the native free is called in total.

jerry-core/ecma/base/ecma-helpers-external-pointers.c
```diff
@@ -176,49 +188,40 @@ ecma_delete_native_pointer_property (ecma_object_t *obj_p, /**< object to delete
...
+  do
   {
     if (native_pointer_p->info_p == info_p)
     {
       if (prev_p == NULL)
       {
-        if (native_pointer_p->next_p == NULL)
-        {
-          /* Only one native pointer property exists, so the property can be deleted as well. */
-          ecma_op_general_object_delete (obj_p, name_p, false);
```
galpeter, line 194: Is it ok to not remove the property storing native ptrs?

zherczeg: That is intentional. Instead of creating deleted entires (holes), we can reuse the first entry if later a new native data is added.

tests/unit-core/test-native-pointer.c
zherczeg: Forgot to add a check at the end.

zherczeg:
Thank you for the reiview, patch is updated.

galpeter (approved):
lgtm

PR #4572, opened 2021-02-07
Fix template memory leak when debugger is running

zherczeg:
The problem is, that the test does not fail without the fix, because `run-debugger-test.sh` ignores the crash of the background process, and returns with success.


rerobika (approved):

dbatyai (approved):
LGTM

PR #4575, opened 2021-02-08
Split globalThis to be configurable.

lygstate:
plit globalThis to be configurable.

JERRY_BUILTIN_GLOBAL_THIS can be enabled without JERRY_ESNEXT


ossy-szeged:
ES5.1 is a 10 years old legacy specification, globalThis is a new ES2020 feature,
Personally, I don't agree enabling any ES5.1+ feature for ES5.1 profile by default.

I think the goal of the feature guards is to be able to disable unnecessary features,
not to enable new features for legacy configuration.

rerobika:
I can agree with that, ES5.1 profile **must not** be burdened with es.next features. Just swap the logic, and instead of extending the featureset of ES5.1, you can reduce the featureset of ES.next profile until it's almost equivalent to ES5.1, but global this is enabled.

lygstate:
Updated

jerry-core/config.h
```diff
@@ -87,6 +87,10 @@
 # define JERRY_BUILTIN_DATAVIEW JERRY_ESNEXT
 #endif /* !defined (JERRY_BUILTIN_DATAVIEW) */
 
+#ifndef JERRY_BUILTIN_GLOBAL_THIS
```
rerobika, line 90: There must be a check below, where you validate the `JERRY_BUILTIN_GLOBAL_THIS` must be zero or one.

rerobika (changes requested):

jerry-core/config.h
lygstate: Please clarify more cleanly, 
I references example like:
```
#ifndef JERRY_BUILTIN_MAP
# define JERRY_BUILTIN_MAP JERRY_ESNEXT	# define JERRY_BUILTIN_MAP JERRY_ESNEXT
#endif /* !defined (JERRY_BUILTIN_MAP) */	#endif /* !defined (JERRY_BUILTIN_MAP) */
```
Anything special here?
If they comes for same reason, I suggestion fix it in separate pull request

rerobika: https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/config.h#L530. No extra PR is needed.

lygstate: done

jerry-core/api/jerry.c
```diff
@@ -1446,6 +1446,9 @@ jerry_is_feature_enabled (const jerry_feature_t feature) /**< feature to check *
 #if JERRY_LOGGING
           || feature == JERRY_FEATURE_LOGGING
 #endif /* JERRY_LOGGING */
+#if JERRY_BUILTIN_GLOBAL_THIS
+          || feature == JERRY_FEATURE_GLOBAL_THIS
+#endif
```
rerobika, line 1451: missing endif comment

jerry-core/include/jerryscript-core.h
```diff
@@ -100,6 +100,7 @@ typedef enum
   JERRY_FEATURE_SYMBOL, /**< symbol support */
   JERRY_FEATURE_DATAVIEW, /**< DataView support */
   JERRY_FEATURE_PROXY, /**< Proxy support */
+  JERRY_FEATURE_GLOBAL_THIS, /**<GlobalThisValue support */
```
rerobika, line 103: This must be after `JERRY_FEATURE_REALM`. Also add a pace after `/**<`

rerobika (changes requested):

jerry-core/include/jerryscript-core.h
lygstate: OK, done

rerobika (approved):

dbatyai (approved):

PR #4576, opened 2021-02-08
Add implementation specific date parsing

zherczeg:


ossy-szeged:
I wouldn't say non-standard, but implementation specific date format, because the specification explicitly allows any implementation specific heuristics. https://www.ecma-international.org/ecma-262/11.0/#sec-date.parse

rerobika (approved):

dbatyai (approved):

PR #4582, opened 2021-02-09
Move static before inline.

lygstate:
Make this consistence with other static inline definition.


rerobika (approved):

zherczeg (approved):
LGTM

PR #4584, opened 2021-02-09
Fixes comment in date-parse.js

lygstate:


rerobika:
I see no problem with this comment and it's more futureproof if any other engine implements it.

PR #4585, opened 2021-02-09
--engine parameter should be a file

lygstate:
The --test262-object option should be passed independently,
otherwise os.path.dirname(args.engine) in tools\runners\run-test-suite-test262.py
makes no sense


rerobika (approved):

ossy-szeged (approved):
LGTM

PR #4586, opened 2021-02-09
Improve TEST_ASSERT output so that vscode IDE can recognize it, didn't know other IDE

lygstate:

### DIFF
--- a/tests/unit-core/test-common.h
+++ b/tests/unit-core/test-common.h
@@ -33,10 +33,10 @@
     if (JERRY_UNLIKELY (!(x))) \
     { \
       jerry_port_log (JERRY_LOG_LEVEL_ERROR, \
-                      "TEST: Assertion '%s' failed at %s(%s):%lu.\n", \
+                      "TEST: Assertion '%s' failed at (%s)%s:%lu.\n", \
                       #x, \
-                      __FILE__, \
                       __func__, \
+                      __FILE__, \
                       (unsigned long) __LINE__); \
       jerry_port_fatal (ERR_FAILED_INTERNAL_ASSERTION); \
     } \


akosthekiss (changes requested):
I have quite some comments here:
- If the assertion message format is to be changed, it should not be changed in the unit tests only. jrt-fatals.c in jerry-core also prints assertion messages (in jerry_assert_fail and in jerry_unreachable), which have a similar format. These should be kept in sync.
- I'm not convinced that the current format is so bad. I've checked the format of three different libc implementations. One of them uses the currently used EXPRESSION-FILE-FUNCTION-LINE order (musl libc), while another doesn't print FUNCTION information at all (msvc):
    - glibc's assert format: "FILE:LINE: FUNCTION: Assertion `EXPRESSION' failed." (https://www.gnu.org/software/libc/manual/html_mono/libc.html#Consistency-Checking)
    - musl libc's assert format: "Assertion failed: EXPRESSION (FILE: FUNCTION: LINE)" (https://git.musl-libc.org/cgit/musl/tree/src/exit/assert.c)
    - MS CRT assert format: "Assertion failed: EXPRESSION, file FILE, line LINE" (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/assert-macro-assert-wassert?view=msvc-160)
- It is not mentioned which IDE would benefit from the proposed assert format change. Would other IDEs also benefit from this? Are other IDEs already able to extract file:line info from the error message?

lygstate:
> * ert format: "FILE:LINE: FUNCTION: Assertion `EXPRESSION' failed." ([https://www.gnu.org/software/libc/manual/html_mon](https://www.gnu.org/software/libc/manual/html_mono/libc.html#Consistency-Checking)



> I have quite some comments here:
> 
> * If the assertion message format is to be changed, it should not be changed in the unit tests only. jrt-fatals.c in jerry-core also prints assertion messages (in jerry_assert_fail and in jerry_unreachable), which have a similar format. These should be kept in sync.
> * I'm not convinced that the current format is so bad. I've checked the format of three different libc implementations. One of them uses the currently used EXPRESSION-FILE-FUNCTION-LINE order (musl libc), while another doesn't print FUNCTION information at all (msvc):
>   
>   * glibc's assert format: "FILE:LINE: FUNCTION: Assertion `EXPRESSION' failed." (https://www.gnu.org/software/libc/manual/html_mono/libc.html#Consistency-Checking)
>   * musl libc's assert format: "Assertion failed: EXPRESSION (FILE: FUNCTION: LINE)" (https://git.musl-libc.org/cgit/musl/tree/src/exit/assert.c)
>   * MS CRT assert format: "Assertion failed: EXPRESSION, file FILE, line LINE" (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/assert-macro-assert-wassert?view=msvc-160)
> * It is not mentioned which IDE would benefit from the proposed assert format change. Would other IDEs also benefit from this? Are other IDEs already able to extract file:line info from the error message?



> I have quite some comments here:
> 
> * If the assertion message format is to be changed, it should not be changed in the unit tests only. jrt-fatals.c in jerry-core also prints assertion messages (in jerry_assert_fail and in jerry_unreachable), which have a similar format. These should be kept in sync.
> * I'm not convinced that the current format is so bad. I've checked the format of three different libc implementations. One of them uses the currently used EXPRESSION-FILE-FUNCTION-LINE order (musl libc), while another doesn't print FUNCTION information at all (msvc):

For the note, jerryscript are using  `%s(%s):%lu.`
Musl using: `fprintf(stderr, "Assertion failed: %s (%s: %s: %d)\n", expr, file, func, line);`

For people, it's looks like the same, but for IDE and other auto tools, they are totally different things.
Also I suggest the glibc format as they are broadly used and can handled by more tools.

>   
>   * glibc's assert format: "FILE:LINE: FUNCTION: Assertion `EXPRESSION' failed." (https://www.gnu.org/software/libc/manual/html_mono/libc.html#Consistency-Checking)
>   * musl libc's assert format: "Assertion failed: EXPRESSION (FILE: FUNCTION: LINE)" (https://git.musl-libc.org/cgit/musl/tree/src/exit/assert.c)
>   * MS CRT assert format: "Assertion failed: EXPRESSION, file FILE, line LINE" (https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/assert-macro-assert-wassert?view=msvc-160)
> * It is not mentioned which IDE would benefit from the proposed assert format change. Would other IDEs also benefit from this? Are other IDEs already able to extract file:line info from the error message?

akosthekiss:
My problem is that there is no single universally accepted format. E.g., see here: https://artalus.github.io/cpptooling/2019/01/13/ide-custom-errors.html . The author of the linked post was struggling with different IDEs expecting differently formatted diagnostic messages (Qt Creator vs Visual Studio). It may be impossible to meet the needs of all IDEs.

The good thing, however, is that the IDE you are targeting (VS Code) has so-called "problem matchers" (https://code.visualstudio.com/Docs/editor/tasks#_defining-a-problem-matcher) that use regexes to find file, line, message, etc. information in diagnostic messages. I.e., there may be no need to change the code base, only a problem matcher to be defined for VS Code.

lygstate:
> My problem is that there is no single universally accepted format. E.g., see here: https://artalus.github.io/cpptooling/2019/01/13/ide-custom-errors.html . The author of the linked post was struggling with different IDEs expecting differently formatted diagnostic messages (Qt Creator vs Visual Studio). It may be impossible to meet the needs of all IDEs.

Yeap, but we have no need creating new custom errors, we can use glibc's assert format: `"FILE:LINE: FUNCTION: Assertion `EXPRESSION' failed."`, What's the downside doing that?


> 
> The good thing, however, is that the IDE you are targeting (VS Code) has so-called "problem matchers" (https://code.visualstudio.com/Docs/editor/tasks#_defining-a-problem-matcher) that use regexes to find file, line, message, etc. information in diagnostic messages. I.e., there may be no need to change the code base, only a problem matcher to be defined for VS Code.

akosthekiss:
> Yeap, but we have no need creating new custom errors, we can use glibc's assert format: `"FILE:LINE: FUNCTION: Assertion 'EXPRESSION' failed."`, What's the downside doing that?

"Breaking tradition", perhaps. That's how JerryScript has formatted its errors for a long time (more than 5 years). E.g., try and convince Chrome developers that the output format of their CHECK/DCHECK macros is not IDE-friendly, and they should be changed to follow glibc's pattern. Or pylint, eslint, etc. project about the same. (These projects don't even print diagnostics on a single line.)

A bit more technical answer is that there are already some tools that look for the existing format of error messages (like fuzzing tools). Should the format change, the tools need an update.

But let's see what other maintainers think.

P.S.: I did some further digging. Even GCC does not follow the glibc pattern in case of an ICE (actually it both does and does not follow the pattern). An ICE message looks like: `main.c:128:1: internal compiler error: in extract_constrain_insn, at recog.c:2211`. I.e., the first `FILE:LINE:COLUMN:` pattern refers to the input file that is being compiled, which triggered the error, and the `in FUNCTION, at FILE:LINE` pattern at the end of the message refers to the location of the error inside the compiler. I'm bringing this up because JerryScript is kind of a compiler (interpreter), so I see similarities. However, I'm not suggesting to use this format either, because we cannot always point at the input file location in case of an ICE.

lygstate:
> > Yeap, but we have no need creating new custom errors, we can use glibc's assert format: `"FILE:LINE: FUNCTION: Assertion 'EXPRESSION' failed."`, What's the downside doing that?
> 
> "Breaking tradition", perhaps. That's how JerryScript has formatted its errors for a long time (more than 5 years). E.g., try and convince Chrome developers that the output format of their CHECK/DCHECK macros is not IDE-friendly, and they should be changed to follow glibc's pattern. Or pylint, eslint, etc. project about the same. (These projects don't even print diagnostics on a single line.)
> 
> A bit more technical answer is that there are already some tools that look for the existing format of error messages (like fuzzing tools). Should the format change, the tools need an update.
> 
> But let's see what other maintainers think.
> 
> P.S.: I did some further digging. Even GCC does not follow the glibc pattern in case of an ICE (actually it both does and does not follow the pattern). An ICE message looks like: `main.c:128:1: internal compiler error: in extract_constrain_insn, at recog.c:2211`. I.e., the first `FILE:LINE:COLUMN:` pattern refers to the input file that is being compiled, which triggered the error, and the `in FUNCTION, at FILE:LINE` pattern at the end of the message refers to the location of the error inside the compiler. I'm bringing this up because JerryScript is kind of a compiler (interpreter), so I see similarities. However, I'm not suggesting to use this format either, because we cannot always point at the input file location in case of an ICE.

OK, then I'll pending this PR, if this is not proper, close it

rerobika:
Based on the conversation above I've still not convinced that the new format is better than the currently existing one.

Moreover, humble request:
> Improve TEST_ASSERT output so that vscode IDE can recognize it, didn't know other IDE

Next time please use short and descriptive PR titles and elaborate the intuition of the patch in the commit description.

PR #4592, opened 2021-02-11
Use C99 flexible array member notation in parser memory struct

galpeter:
The usage of a single byte sized array as a flexible array member can lead to
compiler/analyzer errors or warnings. Not specifying the size in the array is
correct as per C99 standard.

Flexible array members are defined in the C Standard, 6.7.2.1 (ISO/IEC 9899:1999).

### DIFF
--- a/jerry-core/parser/js/js-parser-internal.h
+++ b/jerry-core/parser/js/js-parser-internal.h
@@ -330,7 +330,7 @@ typedef struct
 typedef struct parser_mem_page_t
 {
   struct parser_mem_page_t *next_p;           /**< next page */
-  uint8_t bytes[1];                           /**< memory bytes */
+  uint8_t bytes[];                            /**< memory bytes, C99 flexible array member */
 } parser_mem_page_t;
 
 /**


galpeter:
Without this the following error was triggered with gcc 10.2.0:
```
$ ./tools/build.py --clean --stack-limit=96 --cmake-param=-DCMAKE_BUILD_TYPE=Release

In function ‘parser_stack_iterator_write’,
    inlined from ‘parser_parse_if_statement_end’ at jerry-core/parser/js/js-parser-statm.c:871:3,
    inlined from ‘parser_parse_statements’ at jerry-core/parser/js/js-parser-statm.c:3354:15:
jerry-core/parser/js/js-parser-mem.c:733:5: error: writing 16 bytes into a region of size 1 [-Werror=stringop-overflow=]
  733 |     memcpy (iterator->current_p->bytes + iterator->current_position - length,
      |     ^
jerry-core/parser/js/js-parser-statm.c: In function ‘parser_parse_statements’:
jerry-core/parser/js/js-parser-internal.h:333:11: note: at offset 0 to object ‘bytes’ with size 1 declared here
  333 |   uint8_t bytes[1];                           /**< memory bytes */
      |           ^
```

rerobika (approved):

ossy-szeged (approved):
LGTM

PR #4594, opened 2021-02-15
Refactor jerry_property_descriptor_t to use status flags

pmarkee:

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -473,6 +473,28 @@ Enum that contains the supported binary operation types
 
 - [jerry_binary_operation](#jerry_binary_operation)
 
+## jerry_property_descriptor_flags_t
+
+Enum that contains the flags of property descriptors.
+
+ - JERRY_PROP_NO_OPTS - empty property descriptor
+ - JERRY_PROP_IS_GET_DEFINED - Is [[Get]] defined?
+ - JERRY_PROP_IS_SET_DEFINED - Is [[Set]] defined?
+ - JERRY_PROP_IS_CONFIGURABLE - [[Configurable]]
+ - JERRY_PROP_IS_ENUMERABLE - [[Enumerable]]
+ - JERRY_PROP_IS_WRITABLE - [[Writable]]
+ - JERRY_PROP_IS_THROW - Flag that controls failure handling
+ - JERRY_PROP_IS_VALUE_DEFINED - Is [[Value]] defined?
+ - JERRY_PROP_IS_CONFIGURABLE_DEFINED - Is [[Configurable]] defined?
+ - JERRY_PROP_IS_ENUMERABLE_DEFINED - Is [[Enumerable]] defined?
+ - JERRY_PROP_IS_WRITABLE_DEFINED - Is [[Writable]] defined?
+
+*New in version [[NEXT_RELEASE]]*.
+
+**See also**
+
+- [jerry_property_descriptor_t](#jerry_property_descriptor_t)
+
 ## jerry_property_descriptor_t
 
 **Summary**
@@ -491,32 +513,8 @@ the `value` field should contain the value for the property.
 ```c
 typedef struct
 {
-  /** Is [[Value]] defined? */
-  bool is_value_defined;
-
-  /** Is [[Get]] defined? */
-  bool is_get_defined;
-
-  /** Is [[Set]] defined? */
-  bool is_set_defined;
-
-  /** Is [[Writable]] defined? */
-  bool is_writable_defined;
-
-  /** [[Writable]] */
-  bool is_writable;
-
-  /** Is [[Enumerable]] defined? */
-  bool is_enumerable_defined;
-
-  /** [[Enumerable]] */
-  bool is_enumerable;
-
-  /** Is [[Configurable]] defined? */
-  bool is_configurable_defined;
-
-  /** [[Configurable]] */
-  bool is_configurable;
+  /** any combination of jerry_property_descriptor_flags_t bits */
+  uint16_t flags;
 
   /** [[Value]] */
   jerry_value_t value;
@@ -531,6 +529,7 @@ typedef struct
 
 **See also**
 
+- [jerry_property_descriptor_flags_t](#jerry_property_descriptor_flags_t)
 - [jerry_define_own_property](#jerry_define_own_property)
 
 ## jerry_backtrace_location_t
@@ -4230,17 +4229,16 @@ example (void)
 {
   jerry_value_t prop_name = jerry_create_string_from_utf8 ((jerry_char_t *) "value");
 
-  jerry_property_descriptor_t prop_desc;
-  jerry_init_property_descriptor_fields (&prop_desc);
+  jerry_property_descriptor_t prop_desc = jerry_property_descriptor_create ();
 
   prop_desc.value = prop_name;
-  prop_desc.is_value_defined = true;
+  prop_desc.flags |= JERRY_PROP_IS_VALUE_DEFINED;
 
   jerry_value_t from_object = jerry_from_property_descriptor (&prop_desc);
 
   jerry_release_value (prop_name);
   jerry_release_value (from_object);
-  jerry_free_property_descriptor_fields (&prop_desc);
+  jerry_property_descriptor_free (&prop_desc);
 }
 ```
 
@@ -4293,7 +4291,7 @@ example (void)
   jerry_release_value (object);
   jerry_release_value (prop_name);
   jerry_release_value (value);
-  jerry_free_property_descriptor_fields (&prop_desc);
+  jerry_property_descriptor_free (&prop_desc);
 }
 ```
 
@@ -7007,32 +7005,31 @@ main (void)
 - [jerry_get_internal_property](#jerry_get_internal_property)
 
 
-## jerry_init_property_descriptor_fields
+## jerry_property_descriptor_create
 
 **Summary**
 
-Initialize property descriptor. This means that all fields in the `jerry_property_descriptor_t`
-struct will be set to zero or false depending on the field's type.
+Create a property descriptor and initialize it with default values. This means that all fields in
+the `jerry_property_descriptor_t` struct will be set to zero or false depending on the field's type.
 
 **Prototype**
 
 ```c
-void
-jerry_init_property_descriptor_fields (jerry_property_descriptor_t *prop_desc_p);
+jerry_property_descriptor_t
+jerry_property_descriptor_create (void);
 ```
 
-- `prop_desc_p` - pointer to property descriptor
+*New in version [[NEXT_RELEASE]]*: Replaces `jerry_init_property_descriptor_fields`.
 
 **Example**
 
 ```c
 {
-  jerry_property_descriptor_t prop_desc;
-  jerry_init_property_descriptor_fields (&prop_desc);
+  jerry_property_descriptor_t prop_desc = jerry_property_descriptor_create ();
 
   ... // usage of prop_desc
 
-  jerry_free_property_descriptor_fields (&prop_desc);
+  jerry_property_descriptor_free (&prop_desc);
 }
 ```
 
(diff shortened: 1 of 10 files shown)


jerry-core/include/jerryscript-core.h
```diff
@@ -165,36 +165,31 @@ typedef uint32_t jerry_length_t;
...
 
-  /** Is [[Enumerable]] defined? */
-  bool is_enumerable_defined;
+  JERRY_PROP_IS_CONFIGURABLE = (1 << 2), /** [[Configurable]] */
+  JERRY_PROP_IS_ENUMERABLE = (1 << 3), /** [[Enumerable]] */
+  JERRY_PROP_IS_WRITABLE = (1 << 4), /** [[Writable]] */
 
-  /** [[Enumerable]] */
-  bool is_enumerable;
+  JERRY_PROP_IS_VALUE_DEFINED = (1 << 6), /** Is [[Value]] defined? */
```
zherczeg, line 180: (1 << 5) is missing

pmarkee: It's missing so that we keep alignment with `ecma_property_descriptor_status_flags_t`, where (1 << 5) is the `ECMA_PROP_IS_THROW` flag, which we don't need to publish in `jerry_property_descriptor_status_flags_t`.

zherczeg: I was worried about that. You cannot just "copy" the flags, since invalid flag combinations can be set by users.

pmarkee: That's true. Maybe we could check that the (1 << 5) flag is not set when converting `jerry_property_descriptor_t` to `ecma_property_descriptor_t` and vice versa. The other solution is to not skip the (1 << 5) value in `jerry_property_descriptor_t`, but then we would need to copy the flags in two parts, the part before (1 << 5) and after (1 << 5), instead of a single assignment like we currently can. Which would you prefer?

rerobika: I'd prefer moving `ECMA_PROP_IS_THROW` to the end of the enum internally. For the first check I see no any special usecase why it cannot be moved, so please try it.

zherczeg: I just run into this a few weeks ago. I think it must not clash with the writable/enumerable/configurable bits of the 1 byte property descriptor. So it cannot be moved freely.

rerobika: Okay, seems reasonable. Or how about publishing this flags though? TBH I dont see any disavantage of having this flags and the standard also defines `defineProperty` and `definePropetyOrThrow` operations as well. So by that we can extend the functionality of the API function.

zherczeg: Maybe we could move the writable/enumerable/configurable to the first 3 bits of the prop descriptor, and the other bits in any order. Then the order of the bits woudn't be so random.

docs/02.API-REFERENCE.md
```diff
@@ -491,32 +491,8 @@ the `value` field should contain the value for the property.
...
-
-  /** [[Enumerable]] */
-  bool is_enumerable;
-
-  /** Is [[Configurable]] defined? */
-  bool is_configurable_defined;
-
-  /** [[Configurable]] */
-  bool is_configurable;
+  /** any combination of jerry_property_descriptor_status_flags_t bits */
```
galpeter, line 494: Missing documentation for: `jerry_property_descriptor_status_flags_t`

docs/02.API-REFERENCE.md
```diff
@@ -7007,28 +6982,25 @@ main (void)
 - [jerry_get_internal_property](#jerry_get_internal_property)
 
 
-## jerry_init_property_descriptor_fields
+## jerry_make_empty_property_descriptor
```
galpeter, line 6985: This is a new function and should be documented with `New in .... release` (please see other "new" api methods for the format of this). Also I was wondering if we want to mention if the API replaces another method? That is adding an entry like `Replaces .... api method`.

jerry-core/api/jerry.c
```diff
@@ -3311,24 +3289,10 @@ jerry_property_descriptor_to_ecma (const jerry_property_descriptor_t *prop_desc_
...
-    flags |= (uint32_t) (ECMA_PROP_IS_ENUMERABLE_DEFINED | (prop_desc_p->is_enumerable ? ECMA_PROP_IS_ENUMERABLE
-                                                                                       : ECMA_PROP_NO_OPTS));
-  }
-
-  if (prop_desc_p->is_configurable_defined)
-  {
-    flags |= (uint32_t) (ECMA_PROP_IS_CONFIGURABLE_DEFINED | (prop_desc_p->is_configurable ? ECMA_PROP_IS_CONFIGURABLE
-                                                                                           : ECMA_PROP_NO_OPTS));
-  }
+  prop_desc.flags = prop_desc_p->flags;
```
galpeter, line 3292: If we are expecting that the internal prop flags are the same as the public api ones, we need to have a static assert to validated they have the same value.

jerry-core/include/jerryscript-core.h
```diff
@@ -165,36 +165,31 @@ typedef uint32_t jerry_length_t;
...
+  JERRY_PROP_IS_CONFIGURABLE_DEFINED = (1 << 7), /** Is [[Configurable]] defined? */
+  JERRY_PROP_IS_ENUMERABLE_DEFINED = (1 << 8), /** Is [[Enumerable]] defined? */
+  JERRY_PROP_IS_WRITABLE_DEFINED = (1 << 9), /** Is [[Writable]] defined? */
+} jerry_property_descriptor_status_flags_t;
```
galpeter, line 184: What if we use the  `jerry_property_descriptor_flags_t`? The `status` does not provide extra info IMHO.

pmarkee: I named it like this in order to follow `ecma_property_descriptor_status_flags_t`, but of course we can name it like you suggest.

jerry-core/include/jerryscript-core.h
pmarkee: A little update on this: I have applied your suggestions, but currently it is causing some tests to fail, I will update the PR once I fix them.

docs/02.API-REFERENCE.md
```diff
@@ -473,6 +473,35 @@ Enum that contains the supported binary operation types
...
+  ECMA_PROP_IS_CONFIGURABLE = (1 << 2), /** [[Configurable]] */
+  ECMA_PROP_IS_ENUMERABLE = (1 << 3), /** [[Enumerable]] */
+  ECMA_PROP_IS_WRITABLE = (1 << 4), /** [[Writable]] */
+  ECMA_PROP_IS_THROW = (1 << 5), /** Flag that controls failure handling */
+
+  ECMA_PROP_IS_VALUE_DEFINED = (1 << 6), /** Is [[Value]] defined? */
+  ECMA_PROP_IS_CONFIGURABLE_DEFINED = (1 << 7), /** Is [[Configurable]] defined? */
+  ECMA_PROP_IS_ENUMERABLE_DEFINED = (1 << 8), /** Is [[Enumerable]] defined? */
+  ECMA_PROP_IS_WRITABLE_DEFINED = (1 << 9), /** Is [[Writable]] defined? */
+} ecma_property_descriptor_status_flags_t;
```
galpeter, line 500: AFAIK this is not correct:
* This has internal names
* All enum members should be listed and described (check other enum descriptions in the doc).

docs/02.API-REFERENCE.md
```diff
@@ -7088,19 +7092,23 @@ main (void)
...
+  jerry_value_t value_to_set = jerry_create_number (33);
 
   // set the property descriptor fields:
-  // set the "is_value_defined" field to "true" to indicate the "value"
+  // set the "JERRY_PROP_IS_VALUE_DEFINED" flag to indicate the "value"
   //  field should be used during the property registration.
-  prop_desc.is_value_defined = true;
+  // set the writable, configurable, enumerable flags to true
+  prop_desc.flags |= JERRY_PROP_IS_WRITABLE_DEFINED
+    | JERRY_PROP_IS_WRITABLE
```
zherczeg, line 7099: Incorrect alignment. Use brackets, similar to how an `if ()` is organized.

zherczeg (changes requested):

tests/unit-core/test-api-object-property-names.c
```diff
@@ -104,16 +104,14 @@ main (void)
...
-  jerry_property_descriptor_t prop_desc;
-  jerry_init_property_descriptor_fields (&prop_desc);
-  prop_desc.is_configurable_defined = true;
-  prop_desc.is_configurable = true;
-  prop_desc.is_writable_defined = true;
-  prop_desc.is_writable = true;
-  prop_desc.is_enumerable_defined = true;
+  jerry_property_descriptor_t prop_desc = jerry_make_empty_property_descriptor ();
+  prop_desc.flags |= JERRY_PROP_IS_CONFIGURABLE_DEFINED
+  | JERRY_PROP_IS_CONFIGURABLE
```
zherczeg, line 109: Incorrect alignment.

jerry-core/api/jerry.c
```diff
@@ -3472,28 +3457,22 @@ jerry_get_own_property_descriptor (const jerry_value_t  obj_val, /**< object val
...
-  prop_desc_p->is_enumerable = (prop_desc.flags & ECMA_PROP_IS_ENUMERABLE) != 0;
-
-  prop_desc_p->is_writable_defined = (prop_desc.flags & ECMA_PROP_IS_WRITABLE_DEFINED) != 0;
-  prop_desc_p->is_writable = prop_desc_p->is_writable_defined ? (prop_desc.flags & ECMA_PROP_IS_WRITABLE) != 0 : false;
-
-  prop_desc_p->is_value_defined = (prop_desc.flags & ECMA_PROP_IS_VALUE_DEFINED) != 0;
-  prop_desc_p->is_get_defined = (prop_desc.flags & ECMA_PROP_IS_GET_DEFINED) != 0;
-  prop_desc_p->is_set_defined = (prop_desc.flags & ECMA_PROP_IS_SET_DEFINED) != 0;
+  prop_desc_p->flags = prop_desc.flags;
+  if ((prop_desc_p->flags & JERRY_PROP_IS_WRITABLE_DEFINED) == 0 || (prop_desc.flags & ECMA_PROP_IS_WRITABLE) == 0)
```
zherczeg, line 3461: A similar check might be needed for configurable, enumerable. I am not sure their values are ignored when the DEFINED is not set.

zherczeg (changes requested):

jerry-core/api/jerry.c
```diff
@@ -3472,28 +3457,31 @@ jerry_get_own_property_descriptor (const jerry_value_t  obj_val, /**< object val
...
+  {
+    prop_desc_p->flags &= (uint16_t) ~JERRY_PROP_IS_WRITABLE;
+  }
+  if ((prop_desc_p->flags & JERRY_PROP_IS_CONFIGURABLE_DEFINED) == 0
+      || (prop_desc.flags & ECMA_PROP_IS_CONFIGURABLE) == 0)
```
zherczeg, line 3466: I think the second part of the check is unnecessary. Also true for the other three checks.

tests/unit-core/test-to-property-descriptor.c
```diff
@@ -139,14 +151,18 @@ main (void)
...
-  TEST_ASSERT (prop_desc.is_enumerable_defined && prop_desc.is_enumerable);
-  TEST_ASSERT (prop_desc.is_configurable_defined && !prop_desc.is_configurable);
+  TEST_ASSERT ((prop_desc.flags & JERRY_PROP_IS_VALUE_DEFINED) == 0);
+  TEST_ASSERT ((prop_desc.flags & JERRY_PROP_IS_WRITABLE_DEFINED) == 0);
+  TEST_ASSERT ((prop_desc.flags & JERRY_PROP_IS_GET_DEFINED) == 0);
+  TEST_ASSERT (prop_desc.flags & JERRY_PROP_IS_SET_DEFINED);
+  TEST_ASSERT (prop_desc.flags & JERRY_PROP_IS_ENUMERABLE_DEFINED);
+  TEST_ASSERT (prop_desc.flags & JERRY_PROP_IS_ENUMERABLE);
+  TEST_ASSERT (prop_desc.flags & JERRY_PROP_IS_CONFIGURABLE_DEFINED);
+  TEST_ASSERT ((prop_desc.flags & JERRY_PROP_IS_CONFIGURABLE) == 0);
```
zherczeg, line 165: I would use `!(prop_desc.flags & JERRY_PROP_IS_CONFIGURABLE)` since in the treu cases you don't use `!= 0` checks. Same for all checks above.

zherczeg (approved):
LGTM with some minor fixes

zherczeg (approved):
still LGTM

galpeter (approved):
lgtm

PR #4599, opened 2021-02-16
Rework external function handlers

zherczeg:
Instead of a fixed number of arguments, a call info structure is passed to the handlers, which can be extended in the future without breaknig the API. This structure holds new.target value, so its getter function is removed.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -603,6 +603,29 @@ typedef struct
 
 - [jerry_get_memory_stats](#jerry_get_memory_stats)
 
+## jerry_call_info_t
+
+**Summary**
+
+Call related information passed to [jerry_external_handler_t](#jerry_external_handler_t).
+
+**Prototype**
+
+```c
+typedef struct jerry_call_info_t
+{
+  jerry_value_t function; /**< invoked function object */
+  jerry_value_t this_value; /**< this value passed to the function  */
+  jerry_value_t new_target; /**< current new target value, undefined for non-constructor calls */
+} jerry_call_info_t;
+```
+
+*New in version [[NEXT_RELEASE]]*.
+
+**See also**
+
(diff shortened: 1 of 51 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -603,6 +603,29 @@ typedef struct
...
+## jerry_external_call_info_t
+
+**Summary**
+
+Call related information passed to [jerry_external_handler_t](#jerry_external_handler_t).
+
+**Prototype**
+
+```c
+typedef struct jerry_external_call_info_t
```
zherczeg, line 615: What do you think about this name? Shall I shorten it to `jerry_call_info_t`?

rerobika: I second the `jerry_call_info_t`.

zherczeg: If we are at it, shall I rename jerry_external_handler_t to jerry_native_handler_t ?

galpeter:
As this is a very big API change I would suggest to roll the API version forward to 3.0.

jerry-core/ecma/operations/ecma-proxy-object.h
```diff
@@ -37,10 +37,8 @@ ecma_proxy_create_revocable (ecma_value_t target,
                              ecma_value_t handler);
 
 ecma_value_t
-ecma_proxy_revoke_cb (const ecma_value_t function_obj,
-                      const ecma_value_t this_val,
-                      const ecma_value_t args_p[],
-                      const uint32_t args_count);
+ecma_proxy_revoke_cb (ecma_object_t *function_obj_p,
+                      const ecma_value_t args_p[], const uint32_t args_count);
```
rerobika, line 41: This indentation is strange. To create the pattern in the file please move all arguments into single lines.

rerobika (approved):
Looks good to me after this minor change

dbatyai (approved):

PR #4601, opened 2021-02-18
Add size checks before using JERRY_VLA in unittests

galpeter:


tests/unit-core/test-arraybuffer.c
```diff
@@ -55,7 +55,8 @@ assert_handler (const jerry_call_info_t *call_info_p, /**< call information */
       && jerry_value_is_string (args_p[1]))
   {
     jerry_length_t utf8_sz = jerry_get_string_size (args_p[1]);
-    JERRY_VLA (char, string_from_utf8, utf8_sz);
+    TEST_ASSERT (utf8_sz <= 127);
```
rerobika, line 58: Could you please specify that, where this 127 comes from?

galpeter: I can add a comment "max assert message size" if it is ok for you.

rerobika: Fine by me, significantly better than a magic constant.

rerobika (approved):

zherczeg (approved):
LGTM

PR #4603, opened 2021-02-18
Limit the string size in jerry_port_track_promise_rejection

galpeter:
The maximum Promise reject reason string is now limited to a 5 kib string.


jerry-port/default/default-promise.c
```diff
@@ -24,16 +24,25 @@ jerry_port_track_promise_rejection (const jerry_value_t promise, /**< rejected p
                                     const jerry_promise_rejection_operation_t operation) /**< operation */
 {
   (void) operation; /* unused */
+  const jerry_size_t max_allowed_size = 5 * 1024;
```
zherczeg, line 27: I would choose `5 * 1024 - 1` because of the 0 terminator.

zherczeg (approved):
LGTM with minor suggestion

rerobika (approved):

PR #4605, opened 2021-02-22
Add filters and more events to Promise callback.

zherczeg:


docs/02.API-REFERENCE.md
```diff
@@ -933,6 +946,43 @@ Possible values:
 - [jerry_promise_set_callback](#jerry_promise_set_callback)
 
 
+## jerry_promise_filter_type_t
```
rerobika, line 949: How about `jerry_promise_event_filter_t` ? (like jerry_property_filter_t;)

jerry-core/api/jerry.c
```diff
@@ -4362,15 +4362,25 @@ jerry_get_promise_state (const jerry_value_t promise) /**< promise object to get
...
                                  void *user_p) /**< user pointer passed to the callback */
 {
   jerry_assert_api_available ();
 
 #if JERRY_BUILTIN_PROMISE && JERRY_PROMISE_CALLBACK
-  JERRY_CONTEXT (promise_callback) = callback;
-  JERRY_CONTEXT (promise_callback_user_p) = user_p;
+  if (filters == JERRY_PROMISE_FILTER_DISABLE || callback == 0)
+  {
+    JERRY_CONTEXT (promise_callback_filters) = 0;
```
rerobika, line 4374: How about `JERRY_PROMISE_FILTER_DISABLE ` instead of `0`?

rerobika (commented):
The feature looks good to me otherwise.

zherczeg:

jerry-core/include/jerryscript-core.h
```diff
@@ -786,14 +799,43 @@ typedef enum
                                            *   value: value for rejecting */
 } jerry_promise_event_type_t;
 
+/**
+ * Filter types for jerry_promise_set_callback callback function.
+ * The callback is only called for those events which are enabled by the filters.
+ */
+typedef enum
+{
+  JERRY_PROMISE_FILTER_DISABLE = 0, /**< disable reporting of all events */
```
rerobika, line 808: Since the enum is called promise_event_filter_t, I'd rather use jerry_promise_event_filter_disable. I think that better represents that the filter is applied on the events not on the promises.

zherczeg: Then all enums should be renamed to event filter, but I think that is too long.

rerobika: Indeed, that's exactly what I meant. I think the extra `_EVENT` 6 characters does not add to much, but makes it more consistent. The longest option is currently `JERRY_PROMISE_FILTER_ASYNC_REACTION_JOB` which is 39 characters which would be 45, which still less then the half of the current column limit. So I'd still support adding the `_EVENT` into the enum values.

rerobika (approved):
LGTM

jerry-core/api/jerry.c
```diff
@@ -4362,15 +4362,24 @@ jerry_get_promise_state (const jerry_value_t promise) /**< promise object to get
...
  */
-void jerry_promise_set_callback (jerry_promise_callback_t callback,  /**< notification callback */
+void jerry_promise_set_callback (jerry_promise_event_filter_t filters, /**< combination of event filters */
+                                 jerry_promise_callback_t callback, /**< notification callback */
                                  void *user_p) /**< user pointer passed to the callback */
 {
   jerry_assert_api_available ();
 
 #if JERRY_BUILTIN_PROMISE && JERRY_PROMISE_CALLBACK
+  if (filters == JERRY_PROMISE_EVENT_FILTER_DISABLE || callback == 0)
```
galpeter, line 4372: minor thing: Should we use `callback == NULL` instead of a simple zero, to indicate that the callback is a pointer?

zherczeg:
Thank you for the reivew, patch is rebased and updated.

galpeter (approved):
lgtm

PR #4606, opened 2021-02-22
Add new.target argument to Promise create

zherczeg:


jerry-core/ecma/builtin-objects/ecma-builtin-promise.c
```diff
@@ -455,7 +455,9 @@ ecma_builtin_promise_dispatch_construct (const ecma_value_t *arguments_list_p, /
     return ecma_raise_type_error (ECMA_ERR_MSG ("First parameter must be callable"));
   }
 
-  return ecma_op_create_promise_object (arguments_list_p[0], ECMA_VALUE_UNDEFINED);
+  return ecma_op_create_promise_object (arguments_list_p[0],
```
rerobika, line 458: I was wondering whether we can use the same logic as in array objects.

As I see this is the only case when the current new target affects the constructed object's prototype, so how about something like this?

https://github.com/jerryscript-project/jerryscript/blob/master/jerry-core/ecma/builtin-objects/ecma-builtin-array.c#L517

By that we can remove the whole getPrototypeFromCtor logic from promise create, and it would always return with an object with %Promise% prototype.

And here at the dispatch construct we can query the result of getPrototypeFromCtor, construct the object, and set the internal prototype manually.

zherczeg: Is this safe?

rerobika: According the the latest [standard ](https://262.ecma-international.org/11.0/#sec-promise-constructor) I see no reason why I cannot be performed or wouldn't be safe.

galpeter (approved):
lgtm

PR #4607, opened 2021-02-22
Reduce memory consumption of Promise objects

zherczeg:
No need to keep a reference to resolver functions. Unused resolvers are cleaned up sooner by GC.


rerobika (approved):

galpeter (approved):
lgtm

PR #4612, opened 2021-02-26
Implement jerry_get_own_property API function

zherczeg:


zherczeg:
Note: this is not a mandatory function, its functionaility can be combined from many api functions, but I think it is very useful.

docs/02.API-REFERENCE.md
```diff
@@ -6992,6 +6992,75 @@ jerry_get_property_by_index (const jerry_value_t obj_val,
...
+jerry_get_own_property (const jerry_value_t obj_val,
+                        const jerry_value_t prop_name_val,
+                        const jerry_value_t receiver_val,
+                        bool *found_p);
+```
+
+- `obj_val` - object value
+- `prop_name_val` - property name
+- `receiver_val` - receiver object
+- `found_p` - [out] true, if the proeprty is found, false otherwise
```
rerobika, line 7020: tipo: `proeprty `

jerry-core/api/jerry.c
```diff
@@ -3025,6 +3025,63 @@ jerry_get_property_by_index (const jerry_value_t obj_val, /**< object value */
...
+    }
+
+    return jerry_return (ret_value);
+  }
+
+  if (found_p != NULL)
+  {
+    *found_p = false;
+  }
+  return ECMA_VALUE_UNDEFINED;
```
rerobika, line 3082: I'd put a new line before

tests/unit-core/test-get-own-property.c
```diff
@@ -0,0 +1,153 @@
...
+
+int
+main (void)
+{
+  TEST_INIT ();
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_value_t pp_string = jerry_create_string ((const jerry_char_t *) "pp");
+  jerry_value_t qq_string = jerry_create_string ((const jerry_char_t *) "qq");
+  jerry_value_t rr_string = jerry_create_string ((const jerry_char_t *) "rr");
```
rerobika, line 52: Please add one testcase for symbol as well.

rerobika (changes requested):
I support the concept, only minor changes.

zherczeg:

rerobika (approved):

jerry-core/api/jerry.c
```diff
@@ -3025,6 +3025,64 @@ jerry_get_property_by_index (const jerry_value_t obj_val, /**< object value */
...
+
+  ecma_object_t *object_p = ecma_get_object_from_value (obj_val);
+  ecma_string_t *property_name_p = ecma_get_prop_name_from_value (prop_name_val);
+
+#if JERRY_BUILTIN_PROXY
+  if (ECMA_OBJECT_IS_PROXY (object_p))
+  {
+    if (found_p != NULL)
+    {
+      *found_p = true;
```
galpeter, line 3060: AFAIK in case of proxy it is not guaranteed if the property is found. Eg.:
```
var data = new Proxy({}, {});
``` 
for this `data` this method will return true for all property. Do we want this?

tests/unit-core/test-get-own-property.c
```diff
@@ -0,0 +1,174 @@
...
+compare_string (jerry_value_t value, /**< value to compare */
+                const char *string_p) /**< expected value */
+{
+  jerry_char_t string_buffer[64];
+
+  TEST_ASSERT (jerry_value_is_string (value));
+
+  size_t size = strlen (string_p);
+  TEST_ASSERT (size <= sizeof (string_buffer));
+  TEST_ASSERT (size == jerry_get_string_size (value));
```
galpeter, line 38: As we are accessing the string as an utf8 string, should we also check the utf8 size here (`jerry_get_utf8_string_size`)?

tests/unit-core/test-get-own-property.c
```diff
@@ -0,0 +1,174 @@
...
+
+  jerry_string_to_utf8_char_buffer (value, string_buffer, (jerry_size_t) size);
+  TEST_ASSERT (memcmp (string_p, string_buffer, size) == 0);
+} /* compare_string */
+
+int
+main (void)
```
galpeter, line 45: Afaik we'll need a tests for:
* `obj_val` is not an object
* `receiver_val` is not an object
* `prop_name_val` is not a valid prop name
* Proxy which triggers an error

jerry-core/api/jerry.c
zherczeg: Yes. If special handling is needed, the user can do it in the API. Another option is never returning anything with Proxy, but I think this is better.

tests/unit-core/test-get-own-property.c
zherczeg: using the utf8 was a mistake

zherczeg:

galpeter (approved):
lgtm

PR #4613, opened 2021-02-26
Remove jerry_port_track_promise_rejection

zherczeg:


rerobika (commented):
TBH I'm not sure that reverting the whole patch is the best way to change the existing port function's functionality to the universal promise callback. I think the current status has several reusable code-paths so beside that removing the port function, I'd simultaneously introduce the unhandled promise tracking in the universal promise callback.

zherczeg:
It is there.

zherczeg:
I get it now, it is a jerry-main logging thing. Patch is updated.

jerry-main/main-utils.c
```diff
@@ -201,6 +201,41 @@ create_test262 (jerry_value_t global_obj) /**< global object */
...
+{
+  (void) value; /* unused */
+  (void) user_p; /* unused */
+
+  if (event_type != JERRY_PROMISE_EVENT_REJECT_WITHOUT_HANDLER)
+  {
+    return;
+  }
+
+  jerry_value_t reason = jerry_get_promise_result (object);
```
rerobika, line 218: Please keep the original port implementation: https://github.com/jerryscript-project/jerryscript/pull/4613/files#diff-25a64be3160f6b2ab557a9c399b8f8a0476d69057f7e7bfaa6fa5786624c3ba4L23 since it has a string length limit for the reason of the rejection.

rerobika (changes requested):

rerobika (approved):

jerry-core/ecma/operations/ecma-promise-object.h
```diff
@@ -34,9 +34,8 @@ typedef enum
   ECMA_PROMISE_IS_PENDING = (1 << 0), /**< pending state */
   ECMA_PROMISE_IS_FULFILLED = (1 << 1), /**< fulfilled state */
   ECMA_PROMISE_ALREADY_RESOLVED = (1 << 2), /**< already resolved */
-  ECMA_PROMISE_HANDLED = (1 << 3), /**< ES11: 25.6.6 [[PromiseIsHandled]] internal slot */
 #if JERRY_PROMISE_CALLBACK
-  ECMA_PROMISE_UNHANDLED_REJECT = (1 << 4), /**< a Promise is rejected without a catch handler */
+  ECMA_PROMISE_UNHANDLED_REJECT = (1 << 3), /**< a Promise is rejected without a catch handler */
```
galpeter, line 38: So this bit is used for the `[[PromiseIsHandled]]` ? Could we add the ES reference into the comment? So we can grep for if someone looks for it.

galpeter (approved):
otherwise lgtm

PR #4614, opened 2021-02-26
Add more skip check options for Proxy objects

zherczeg:
Reorganize the flags to follow the list in ES2020 section 9.5

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -94,8 +94,9 @@ These option bits allow specializing Proxies with non-standard behaviour.
 These flags are recommended only for those trusted Proxies, whose handlers
 produce correct results.
 
-- JERRY_PROXY_SKIP_GET_CHECKS - skip [[Get]] result checks
-- JERRY_PROXY_SKIP_GET_OWN_PROPERTY_CHECKS - skip [[GetOwnProperty]] result checks
+- JERRY_PROXY_SKIP_RESULT_VALIDATION - skip result validation for [[GetPrototypeOf]], [[SetPrototypeOf]], [[IsExtensible]],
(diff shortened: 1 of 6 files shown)


zherczeg:
I am thinking whether it is a good idea to create a flag for each handler or creating groups would be better. Or create only one flag.

zherczeg:
I have been thinking about this, and I am not sure there are use cases where you want to disable checks for `get` but not for `delete`. The point is, the user declartes that the Proxy is trusted, and no need sanity checks. Or the `target` is more like the prototype of the Proxy, so the checks are invalid.

rerobika (approved):

docs/02.API-REFERENCE.md
```diff
@@ -94,8 +94,8 @@ These option bits allow specializing Proxies with non-standard behaviour.
 These flags are recommended only for those trusted Proxies, whose handlers
 produce correct results.
 
-- JERRY_PROXY_SKIP_GET_CHECKS - skip [[Get]] result checks
-- JERRY_PROXY_SKIP_GET_OWN_PROPERTY_CHECKS - skip [[GetOwnProperty]] result checks
+- JERRY_PROXY_SKIP_RESULT_VALIDATION - skip result validation for [[IsExtensible]], [[PreventExtensions]], [[GetOwnProperty]],
+                                       [[DefineOwnProperty]], [[HasProperty]], [[Get]], [[Set]], [[Delete]] and [[OwnPropertyKeys]]
```
galpeter, line 98: What about the Get/SetPrototype other validations? Shouldn't we add those also?

zherczeg: I have added the prototype handlers, but that should be enough.

galpeter (approved):
lgtm

PR #4615, opened 2021-03-01
Add reference support for native pointers.

zherczeg:


zherczeg:
It is an option to allow null/object references (and call these references as weak references) since storing non-objects was never a problem. In this case there is no need a call a finalization function.

docs/02.API-REFERENCE.md
```diff
@@ -8065,6 +8084,164 @@ best-practice example.
...
+  /* References must be initialized before a buffer is attached. */
+  jerry_native_pointer_init_references ((void *) buffer_p, 2);
+
+  jerry_set_object_native_pointer (object_value, (void *) buffer_p, &native_info);
+
+  /* References can be modified after the buffer is attached.
+   * This example sets a self reference. */
+  jerry_native_pointer_set_reference (&buffer_p->a, object_value);
+
+  jerry_release_value (object_value);
```
galpeter, line 8159: What if we use the `native_info` to access the number of references for the init/finalize/set_ref methods? That way the number_of_references is only specified at a single place. The rationale is that the buffer's format is tied to the native_info struct's number_of_references member and by providing the native_info for these methods, it would be harder to miss type the number_of_refs for the init/finalize method.

For example:
```c
jerry_native_pointer_init_references (&native_info, (void *) buffer_p);
   ...
jerry_native_pointer_finalize_references (&native_info, native_p);
```

Related to this the idea is to also provide the native info, buffer start pointer, index for the ref to change and the new value for the set_ref method:
```c
jerry_native_pointer_set_reference (&native_info, (void*)buffer_p, 0, object_value);
```

Or another idea would be:
```c
jerry_native_pointer_set_reference (target_object, &native_info, 0, new_ref_value);
```
Where internally the target_object's native_info will be accessed and based on the provided index a new_ref_value will be set. However with this there would be too many native pointer access calls internally and would potentially be slower.

What do you think?

jerry-core/ecma/base/ecma-helpers-external-pointers.c
```diff
@@ -27,61 +27,70 @@
  * @{
  */
 
+/**
+ * Get the number of weak references of the native pointer
+ */
+#define ECMA_NATIVE_POINTER_GET_REFERENCES(info_p) \
+  ((info_p) != NULL ? ((const jerry_object_native_info_t *) (info_p))->number_of_references \
```
galpeter, line 34: AFAIK this type here is a layering violation. Why do we use API type here, can't we keep the ecma typedef?

docs/02.API-REFERENCE.md
```diff
@@ -741,24 +741,40 @@ typedef bool (*jerry_backtrace_callback_t) (jerry_backtrace_frame_t *frame_p, vo
...
+finalize function also initializes all values to undefined.
 
 **Prototype**
 
 ```c
 typedef struct
 {
-  jerry_object_native_free_callback_t free_cb;
+  jerry_object_native_free_callback_t free_cb; /**< the free callback of the native pointer */
+  uint32_t number_of_references; /**< the number of value references stored in the beginning
```
galpeter, line 776: Was wondering if the "reference" is a good term for this. Maybe "slot" would be better? However I'm not sure. For me the "references" would indicate something related to refcounting and this functionality is slightly different.

docs/02.API-REFERENCE.md
zherczeg: Unfortunately you don't get the native info in the finalization callback, so passing it might be hard.

In the other case, you need to manually count the variables, and that is difficult. If you have a struct like:

typedef struct {
  jerry_value_t something;
  jerry_value_t my_val;
  jerry_value_t object_ref;
}
```

For me `&buffer_p->object_ref` is more descriptive than `(buffer_p, 2)`. You can also reorganize the values freely in the former case. The latter can only work if you use an array and constants for indexing. But that is a lot of definition sperad accross the code.

File: docs/02.API-REFERENCE.md (hunk 3 above)
Comment (zherczeg) [reply docs/02.API-REFERENCE.md:776]: This is a mixed list of strong and weak references, hence I give the name. I don't like the slots. It could be weak references, if we would restrinct these to null.object values.

File: jerry-core/ecma/base/ecma-helpers-external-pointers.c (hunk 2 above)
Comment (zherczeg) [reply jerry-core/ecma/base/ecma-helpers-external-pointers.c:34]: I deleted the ecma typedef, this is not a good way to handle these situations. Too easy to make mistakes when you have to maintain two copies of the same thing. The gc already uses `jerryscript.h`, and I think any .c file should be allowed to use it. `jcontext.h` also includes it

**Comment by zherczeg** [comment]
Thnak you for the review, patch is updated.

**Comment by zherczeg** [comment]
I have reworked the patch, please check it again.

File: jerry-core/api/jerry.c [hunk 4]
```diff
@@ -4345,6 +4345,72 @@ jerry_delete_object_native_pointer (const jerry_value_t obj_val, /**< object to
...
+ * The area must be part of a buffer which is currently assigned to an object.
+ *
+ * Note:
+ *      Error references are not supported, they are replaced by undefined values.
+ */
+void
+jerry_native_pointer_set_reference (jerry_value_t *reference_p, /**< pointer to a reference in a buffer
+                                                                 *   pointed by a native pointer */
+                                    jerry_value_t value) /**< new value of the reference */
+{
```
galpeter, line 4402: Should we add a null check for the reference_p ?

docs/02.API-REFERENCE.md
```diff
@@ -876,33 +876,82 @@ typedef bool (*jerry_backtrace_callback_t) (jerry_backtrace_frame_t *frame_p, vo
...
 
 **Prototype**
 
 ```c
 typedef struct
 {
-  jerry_object_native_free_callback_t free_cb;
+  jerry_object_native_free_callback_t free_cb; /**< the free callback of the native pointer */
+  uint16_t number_of_references; /**< the number of value references which are marked by the garbage collector */
+  uint16_t offset_of_references; /**< byte offset of value references which are marked by the garbage collector */
```
galpeter, line 912: maybe: "Byte offset indicating the start offset of value references in the user allocated buffer." ?

docs/02.API-REFERENCE.md
```diff
@@ -876,33 +876,82 @@ typedef bool (*jerry_backtrace_callback_t) (jerry_backtrace_frame_t *frame_p, vo
...
 **See also**
 
 - [jerry_set_object_native_pointer](#jerry_set_object_native_pointer)
 - [jerry_get_object_native_pointer](#jerry_get_object_native_pointer)
+- [jerry_delete_object_native_pointer](#jerry_delete_object_native_pointer)
+- [jerry_native_pointer_init_references](#jerry_native_pointer_init_references)
+- [jerry_native_pointer_finalize_references](#jerry_native_pointer_finalize_references)
+- [jerry_native_pointer_set_reference](#jerry_native_pointer_set_reference)
+
+## JERRY_OBJECT_NATIVE_INFO_GET_NUMBER_OF_REFERENCES
```
galpeter, line 929: We don't really have macros in the API, do we really want to provide this?

docs/02.API-REFERENCE.md
```diff
@@ -8916,6 +8965,171 @@ best-practice example.
...
+}
+```
+
+**See also**
+
+- [jerry_set_object_native_pointer](#jerry_set_object_native_pointer)
+- [jerry_native_pointer_finalize_references](#jerry_native_pointer_finalize_references)
+- [jerry_native_pointer_set_reference](#jerry_native_pointer_set_reference)
+
+## jerry_native_pointer_release_references
```
galpeter, line 9057: I thinks this is a typo as this should be `jerry_native_pointer_finalize_references` or the other places should use this name.

docs/02.API-REFERENCE.md
```diff
@@ -8916,6 +8965,171 @@ best-practice example.
...
+
+**Summary**
+
+Finalize the references stored in a buffer pointed by a native pointer.
```
galpeter, line 9061: Here finalize means free/release, am I correct?

jerry-core/api/jerry.c
```diff
@@ -4345,6 +4345,72 @@ jerry_delete_object_native_pointer (const jerry_value_t obj_val, /**< object to
...
 
+/**
+ * Initialize the references stored in a buffer pointed by a native pointer.
+ * The references are initialized to undefined.
+ */
+void
+jerry_native_pointer_init_references (void *native_pointer_p, /**< native pointer to a buffer */
+                                      const jerry_object_native_info_t *native_info_p) /**< the type info of
+                                                                                        *   the native pointer */
+{
```
galpeter, line 4356: Should we have nullptr checks?

jerry-core/api/jerry.c
```diff
@@ -4345,6 +4345,72 @@ jerry_delete_object_native_pointer (const jerry_value_t obj_val, /**< object to
...
+/**
+ * Finalize the references after a buffer pointed by a native pointer is not
+ * attached to an object anymore. All references are set to undefined
+ * similar to jerry_native_pointer_init_references.
+ */
+void
+jerry_native_pointer_finalize_references (void *native_pointer_p, /**< native pointer */
+                                          const jerry_object_native_info_t *native_info_p) /**< the type info of
+                                                                                            *   the native pointer */
+{
```
galpeter, line 4377: Should we have nullptr checks?

docs/02.API-REFERENCE.md
zherczeg: Shall I delete it? I am ok with it.

zherczeg:
Patch is updated, please review.

jerry-core/api/jerry.c
```diff
@@ -4362,6 +4362,87 @@ jerry_delete_object_native_pointer (const jerry_value_t obj_val, /**< object to
...
+ *      Error references are not supported, they are replaced by undefined values.
+ */
+void
+jerry_native_pointer_set_reference (jerry_value_t *reference_p, /**< pointer to a reference in a buffer
+                                                                 *   pointed by a native pointer */
+                                    jerry_value_t value) /**< new value of the reference */
+{
+  jerry_assert_api_available ();
+
+  if (reference_p == NULL)
```
zherczeg, line 4432: I have added a null check here, but I am not sure this is useful since the function gets a value reference inside a buffer.

rerobika (approved):
still LGTM

galpeter (approved):
lgtm

PR #4626, opened 2021-03-11
Return early when checking logical operators

zherczeg:
Story: `(1 || 1 || 1)` reported uninitialized data read in valgrind.

The issue was: when logical binary operators are pushed onto the stack, a space for branch target is also pushed:
`parser_stack_push (context_p, &branch, sizeof (parser_branch_t));`

This structure is not initialized, and parser_check_invalid_logical_op checks these bytes, and may throw SyntaxErrors.

When I was thinking how to fix this, I realized I can return early if a logical operator is encountered. E.g. if the parser searches `||` or `&&` and it encounters a `??`, that means there was no SyntaxError when the `??` was pushed, so there was no `||` or `&&` before `??`. Hence we can simply return. This is much simpler than skipping the bytes and continue the search.


rerobika (approved):

galpeter (approved):
lgtm

PR #4631, opened 2021-03-16
Add check Proxy object [[SetPrototypeOf]]

kisbg:

### DIFF
--- a/jerry-core/ecma/operations/ecma-proxy-object.c
+++ b/jerry-core/ecma/operations/ecma-proxy-object.c
@@ -361,6 +361,7 @@ ecma_proxy_object_get_prototype_of (ecma_object_t *obj_p) /**< proxy object */
  *
  * See also:
  *          ECMAScript v6, 9.5.2
+ *          ECMAScript v11: 9.5.2
  *
  * Note: Returned value is always a simple value so freeing it is unnecessary.
  *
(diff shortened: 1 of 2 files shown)


jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -422,6 +422,11 @@ ecma_proxy_object_set_prototype_of (ecma_object_t *obj_p, /**< proxy object */
 
   ecma_free_value (trap_result);
 
+  if (!boolean_trap_result)
```
galpeter, line 425: Is this from a newer standard? If so it would be great if we could reference the standard, maybe even update all of the other references for the standards (for example see the proxy_create method above).

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -404,56 +404,60 @@ ecma_proxy_object_set_prototype_of (ecma_object_t *obj_p, /**< proxy object */
...
   {
     return trap_result;
   }
 
+  /* 9. */
   bool boolean_trap_result = ecma_op_to_boolean (trap_result);
 
   ecma_free_value (trap_result);
 
+  /* 9. */
```
galpeter, line 425: The step "9" is mentioned twice. Do we need it twice?

jerry-core/ecma/operations/ecma-proxy-object.c
```diff
@@ -360,7 +360,7 @@ ecma_proxy_object_get_prototype_of (ecma_object_t *obj_p) /**< proxy object */
  * The Proxy object [[SetPrototypeOf]] internal routine
  *
  * See also:
- *          ECMAScript v6, 9.5.2
+ *          ES11+: 9.5.2
```
galpeter, line 363: Could we use the ES11 (or ECMAString v11) name?  Adding the "+" could mean anything and the newer standards could change. Having a fix version reference would be better.

jerry-core/ecma/operations/ecma-proxy-object.c
kisbg: Yes, since the 9th step is different in the ES6 and ES11 standard.

zherczeg (approved):
LGTM

galpeter (approved):
lgtm

PR #4632, opened 2021-03-19
Rework module linking

zherczeg:
The module linking process from jerry_parse is moved out into a new jerry_module_link function, and jerry_parse is limited to create unlinked modules.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -747,6 +747,37 @@ typedef void (*jerry_error_object_created_callback_t) (const jerry_value_t error
 
 - [jerry_set_error_object_created_callback](#jerry_set_error_object_created_callback)
 
+## jerry_module_resolve_callback_t
+
+**Summary**
+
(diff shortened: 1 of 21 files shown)


rerobika (approved):

zherczeg:
Actually the ES2020 standard does not want to define the module resolving logic, it delegates everything to the user:

15.2.1.18 HostResolveImportedModule
15.2.1.16.1 Link ( ) Concrete Method

This is how v8 works (probably not a coincidence). Node.js for example heavily rely on this method, and maintain its own data set about the resolved modules and their dependencies.

zherczeg:
An alternative could be to move the code from main to jerry-ext. That requires to design an api, so could be in a future patch.

zherczeg:
The patch is reworked to use port for resolving. Please check the patch.

rerobika (approved):
still lgtm

dbatyai (approved):
LGTM

PR #4634, opened 2021-03-22
Extract type defintions from jerryscript-core.h into jerryscript-types.h

zherczeg:
The function declarations in jerryscript-core.h should not be visible internally, only the type declarations in jerryscript-types.h


rerobika (approved):

galpeter (approved):
lgtm

PR #4642, opened 2021-03-30
Add info to external pointer free callback

zherczeg:
Furthermore reduce memory consumption when only one external pointer is assigned to an object.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -696,11 +696,11 @@ typedef jerry_value_t (*jerry_external_handler_t) (const jerry_call_info_t *call
 
 - [jerry_create_external_function](#jerry_create_external_function)
 
-## jerry_object_native_free_callback_t
+## jerry_value_free_callback_t
 
 **Summary**
 
-Native free callback of an object. It is used in `jerry_object_native_info_t` and for external Array buffers.
+Native free callback of generic value types.
 
 *Note*:
  - Referred values by this method must have at least 1 reference. (Correct API usage satisfies this condition)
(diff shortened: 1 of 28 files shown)


rerobika (approved):

docs/02.API-REFERENCE.md
```diff
@@ -696,29 +696,51 @@ typedef jerry_value_t (*jerry_external_handler_t) (const jerry_call_info_t *call
...
 
 ```c
-typedef void (*jerry_object_native_free_callback_t) (void *native_p);
+typedef void (*jerry_object_native_free_callback_t) (void *native_p, struct jerry_object_native_info_t *info_p);
 ```
 
 *New in version 2.0*: Renamed from `jerry_object_free_callback_t`.
-
 *Changed in version 2.2*: API calls are once again allowed. (See note)
+*Changed in version [[NEXT_RELEASE]]*: `info_p` argument is added
```
galpeter, line 739: There should be a newline before the new "changed" entry so the generated html from the md file looks good. (This applies to all such changes below.

docs/02.API-REFERENCE.md
```diff
@@ -696,29 +696,51 @@ typedef jerry_value_t (*jerry_external_handler_t) (const jerry_call_info_t *call
...
+
+Native free callback of generic value types.
+
+*Note*:
+ - Referred values by this method must have at least 1 reference. (Correct API usage satisfies this condition)
+
+**Prototype**
+
+```c
+typedef void (*jerry_value_free_callback_t) (void *native_p);
```
galpeter, line 711: Was wondering if it would make sense to have a "type" information as second argument for this callback. What do you think? Could this be a useful feature? This "type" could/would indicate what kind of data should be freed.

docs/02.API-REFERENCE.md
```diff
@@ -696,29 +696,51 @@ typedef jerry_value_t (*jerry_external_handler_t) (const jerry_call_info_t *call
...
+- [jerry_create_external_string](#jerry_create_external_string)
+- [jerry_create_external_string_sz](#jerry_create_external_string_sz)
+- [jerry_create_arraybuffer_external](#jerry_create_arraybuffer_external)
+
 ## jerry_object_native_free_callback_t
 
 **Summary**
 
-Native free callback of an object. It is used in `jerry_object_native_info_t` and for external Array buffers.
+Native free callback of an object. It is part of the type information provided by `jerry_object_native_info_t`.
```
galpeter, line 726: I'm not sure I understand what do you mean with the `It is part of the type information provided by jerry_object_native_info_t.` sentence.

docs/02.API-REFERENCE.md
```diff
@@ -696,29 +696,51 @@ typedef jerry_value_t (*jerry_external_handler_t) (const jerry_call_info_t *call
...
 
 *Note*:
  - Referred values by this method must have at least 1 reference. (Correct API usage satisfies this condition)
 
 **Prototype**
 
 ```c
-typedef void (*jerry_object_native_free_callback_t) (void *native_p);
+typedef void (*jerry_object_native_free_callback_t) (void *native_p, struct jerry_object_native_info_t *info_p);
```
galpeter, line 734: We should document the arguments just like it is done for the `jerry_external_handler_t` type.

galpeter (commented):
Minor doc change requests.

docs/02.API-REFERENCE.md
zherczeg: I am neutral. If others think this is a good idea we could do it in another patch (before release).

docs/02.API-REFERENCE.md
zherczeg: jerry_object_native_info_t structure represents the internal type information of an object, and jerry_object_native_free_callback_t is a mandatory part of this type info. Let me know if you would prefer another sentence.

galpeter: For me the new sentence does not provide extra info, maybe?: `Native free callback of an object. In addition to the memory pointer the callback receives the native info pointer which was provided by the jerry_set_native_ptr... method.`

docs/02.API-REFERENCE.md
galpeter: That's true. We could even do a different WIP PR to see how it would look.

zherczeg:
Patch is updated, please check.

galpeter (approved):
lgtm

PR #4645, opened 2021-04-07
Remove class id storing for class objects

zherczeg:
The two bytes which stored the id is free to use. Currently the only exception is containers.


rerobika (approved):

dbatyai (approved):
LGTM

PR #4650, opened 2021-04-13
Fix uninitialized data in create native property

zherczeg:
The extra tests does not necessary trigger the issue, since the member is often initialized to 0.


rerobika (approved):

dbatyai (approved):
LGTM

PR #4658, opened 2021-04-20
Fix external pointer allocation issues.

zherczeg:
The allocations during the update of external pointers may trigger the GC. Due to a new feature, the GC uses the external pointer data to mark objects, which caused crashes in some cases.


rerobika (approved):

galpeter (approved):
lgtm

PR #4660, opened 2021-04-20
Export assignment expression must not track variables

zherczeg:
Only export the result of the assignment expression. This value cannot be changed later.


galpeter (approved):
lgtm

rerobika (approved):

PR #4663, opened 2021-04-26
Do not mark the unused items of a WeakMap

zherczeg:
The new method is slower, but correctly release unused objects.


rerobika (approved):

galpeter (approved):
lgtm

PR #4666, opened 2021-05-05
Add container related API function

Achie72:
The new function returns with an array, containing the
elements of the given Container or Container iterator.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -12011,3 +12011,72 @@ main (void)
   return 0;
 }
 ```
+
+# Container Functions
+
+## jerry_get_array_from_container
+
+**Summary**
+
+Return a new array containing elements from a Container, or a Container Iterator. Sets the is_key_value_p to true.
+if the container object contains key-value structure and false if not.
+
+*Notes*
+- The return value will be an empty array if the Map/Set or Iterator object was empty or finished.
+- This API function depends on a build option (`JERRY_BUILTIN_CONTAINER`) and can be checked
+  runtime with the `JERRY_FEATURE_MAP, JERRY_FEATURE_SET, JERRY_FEATURE_WEAKMAP, JERRY_FEATURE_WEAKSET`
+  feature enum values.
+  see: [jerry_is_feature_enabled](#jerry_is_feature_enabled).
+- The es.next profile enables this by default.
+
+*New in version [[NEXT_RELEASE]]*.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_get_array_from_container(jerry_value_t value,
+                               bool *is_key_value_p);
+```
+
+- `value` - Map/Set or iterator object
+- `is_key_value` - Will be set to `true` if the given container has key-value pairs, `false` otherwise.
+- return value
+  - jerry_value_t containing an array of values from the Map/Set or iterator object
+  - Error if the `value` is nor a Container or a Container Iterator.
+  - `undefined` if the `value` is undefined/null.
+**Example**
+
+[doctest]: # ()
+
+```c
+#include "jerryscript.h"
+int
+main (void)
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_char_t src[] = "var map = new Map(); map.set(1,2); map.entries()";
+  jerry_value_t iterable = jerry_eval (src, sizeof (src) - 1, JERRY_PARSE_NO_OPTS);
+
+  bool is_key_value_container = false;
+  jerry_value_t buffer_from_map = jerry_get_array_from_container (iterable, &is_key_value_container);
+
+  /*
+    The buffer_from_map contains two elements: 1 and 2, which is the key/value pair of the only item in the set.
+    is_key_value set to true, as the original is a key-value structure (a Map Iterator)
+  */
+
+  jerry_release_value (iterable);
+  jerry_release_value (buffer_from_map);
+
+  jerry_cleanup ();
+
+  return 0;
+}
+```
+
+**See also**
+
+- [jerry_create_container](#jerry_create_container)
+- [jerry_container_type_t](#jerry_container_type_t)
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -32,6 +32,7 @@
 #include "ecma-gc.h"
 #include "ecma-helpers.h"
 #include "ecma-init-finalize.h"
+#include "ecma-iterator-object.h"
 #include "ecma-lex-env.h"
 #include "lit-char-helpers.h"
 #include "ecma-literal-storage.h"
@@ -6446,7 +6447,7 @@ jerry_create_container (jerry_container_type_t container_type, /**< Type of the 
   JERRY_UNUSED (arguments_list_p);
   JERRY_UNUSED (arguments_list_len);
   JERRY_UNUSED (container_type);
-  return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG ("Containers are disabled")));
+  return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (ecma_error_container_not_supported_p)));
 #endif /* JERRY_BUILTIN_CONTAINER */
 } /* jerry_create_container */
 
@@ -6506,6 +6507,119 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
 #endif /* JERRY_BUILTIN_CONTAINER */
   return JERRY_CONTAINER_TYPE_INVALID;
 } /* jerry_get_container_type */
+
+/**
+ * Return a new array containing elements from a Container or a Container Iterator.
+ * Sets the boolean input value to `true` if the container object has key/value pairs.
+ *
+ * Note:
+ *     the returned value must be freed with a jerry_release_value call
+ *
+ * @return an array of items for maps/sets or their iterators, error otherwise
+ */
+jerry_value_t
+jerry_get_array_from_container (jerry_value_t value, /**< the container or iterator object */
+                                bool *is_key_value_p) /**< [out] is key-value structure */
+{
+  jerry_assert_api_available ();
+
+#if JERRY_BUILTIN_CONTAINER
+  const char *container_needed = ECMA_ERR_MSG ("Value is not a Container or Iterator");
+
+  if (!ecma_is_value_object (value))
+  {
+    return jerry_throw (ecma_raise_type_error (container_needed));
+  }
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (value);
+
+  if (ecma_get_object_type (obj_p) != ECMA_OBJECT_TYPE_CLASS)
+  {
+    return jerry_throw (ecma_raise_type_error (container_needed));
+  }
+
+  ecma_extended_object_t *ext_obj_p = (ecma_extended_object_t *) obj_p;
+
+  uint32_t entry_count;
+  uint8_t entry_size;
+
+  uint32_t index = 0;
+  uint8_t iterator_kind = ECMA_ITERATOR__COUNT;
+  ecma_value_t *start_p;
+
+  *is_key_value_p = false;
+
+  if (ext_obj_p->u.cls.type == ECMA_OBJECT_CLASS_MAP_ITERATOR
+      || ext_obj_p->u.cls.type == ECMA_OBJECT_CLASS_SET_ITERATOR)
+  {
+    ecma_value_t iterated_value = ext_obj_p->u.cls.u3.iterated_value;
+
+    if (ecma_is_value_empty (iterated_value))
+    {
+      return ecma_op_new_array_object_from_collection (ecma_new_collection (), false);
+    }
+
+    ecma_extended_object_t *map_object_p = (ecma_extended_object_t *) (ecma_get_object_from_value (iterated_value));
+
+    ecma_collection_t *container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t, map_object_p->u.cls.u3.value);
+    entry_count = ECMA_CONTAINER_ENTRY_COUNT (container_p);
+    index = ext_obj_p->u.cls.u2.iterator_index;
+
+    entry_size = ecma_op_container_entry_size (map_object_p->u.cls.u2.container_id);
+    start_p = ECMA_CONTAINER_START (container_p);
+
+    iterator_kind = ext_obj_p->u.cls.u1.iterator_kind;
+  }
+  else if (jerry_get_container_type (value) != JERRY_CONTAINER_TYPE_INVALID)
+  {
+    ecma_collection_t *container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t, ext_obj_p->u.cls.u3.value);
+    entry_count = ECMA_CONTAINER_ENTRY_COUNT (container_p);
+    entry_size = ecma_op_container_entry_size (ext_obj_p->u.cls.u2.container_id);
+
+    index = 0;
+    iterator_kind = ECMA_ITERATOR_KEYS;
+    start_p = ECMA_CONTAINER_START (container_p);
+
+    if (ext_obj_p->u.cls.u2.container_id == LIT_MAGIC_STRING_MAP_UL
+        || ext_obj_p->u.cls.u2.container_id == LIT_MAGIC_STRING_WEAKMAP_UL)
+    {
+      iterator_kind = ECMA_ITERATOR_ENTRIES;
+    }
+  }
+  else
+  {
+    return jerry_throw (ecma_raise_type_error (container_needed));
+  }
+
+  *is_key_value_p = (iterator_kind == ECMA_ITERATOR_ENTRIES);
+  ecma_collection_t *collection_buffer = ecma_new_collection ();
+
+  for (uint32_t i = index; i < entry_count; i += entry_size)
+  {
+    ecma_value_t *entry_p = start_p + i;
+
+    if (ecma_is_value_empty (*entry_p))
+    {
+      continue;
+    }
+
+    if (iterator_kind != ECMA_ITERATOR_VALUES)
+    {
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (entry_p[0]));
+    }
+
+    if (iterator_kind != ECMA_ITERATOR_KEYS)
+    {
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (entry_p[1]));
+    }
+  }
+  return ecma_op_new_array_object_from_collection (collection_buffer, false);
+#else /* JERRY_BUILTIN_CONTAINER */
+  JERRY_UNUSED (value);
+  JERRY_UNUSED (is_key_value_p);
+  return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG (ecma_error_container_not_supported_p)));
+#endif
+} /* jerry_get_array_from_container */
 
 /**
  * @}
--- a/tests/unit-core/test-container.c
+++ b/tests/unit-core/test-container.c
@@ -34,6 +34,41 @@ static const jerry_object_native_info_t native_info =
   .offset_of_references = 0,
 };
 
+static jerry_value_t
+create_array_from_container_handler (const jerry_call_info_t *call_info_p,
+                                     const jerry_value_t args_p[],
+                                     const jerry_length_t args_count)
+{
+  JERRY_UNUSED (call_info_p);
+
+  if (args_count < 2)
+  {
+    return jerry_create_undefined ();
+  }
+
+  bool is_key_value_pairs = false;
+  jerry_value_t result = jerry_get_array_from_container (args_p[0], &is_key_value_pairs);
+
+  TEST_ASSERT (is_key_value_pairs == jerry_get_boolean_value (args_p[1]));
+  return result;
+} /* create_array_from_container_handler */
+
+static void
+run_eval (const char *source_p)
+{
+  jerry_value_t result = jerry_eval ((const jerry_char_t *) source_p, strlen (source_p), 0);
+
+  TEST_ASSERT (!jerry_value_is_error (result));
+  jerry_release_value (result);
+} /* run_eval */
+
+static void
+run_eval_error (const char *source_p)
+{
+  jerry_value_t result = jerry_eval ((const jerry_char_t *) source_p, strlen (source_p), 0);
+  jerry_release_value (result);
+} /* run_eval_error */
+
 int
 main (void)
 {
@@ -59,6 +94,15 @@ main (void)
   jerry_value_t global_set = jerry_get_property (global, set_str);
   jerry_value_t global_weakmap = jerry_get_property (global, weakmap_str);
   jerry_value_t global_weakset = jerry_get_property (global, weakset_str);
+
+  jerry_value_t function = jerry_create_external_function (create_array_from_container_handler);
+  jerry_value_t name = jerry_create_string ((const jerry_char_t *) "create_array_from_container");
+  jerry_value_t res = jerry_set_property (global, name, function);
+  TEST_ASSERT (!jerry_value_is_error (res));
+
+  jerry_release_value (res);
+  jerry_release_value (name);
+  jerry_release_value (function);
 
   jerry_release_value (global);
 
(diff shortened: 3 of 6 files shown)


jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (*entry_p));
+    }
+
+  }
+
+  return ecma_op_new_array_object_from_collection (collection_buffer, false);
+#endif /* JERRY_BUILTIN_CONTAINER */
+  JERRY_UNUSED (value);
+  JERRY_UNUSED (is_key_value_p);
+  return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG ("Containers are disabled")));
```
zherczeg, line 6563: Introduce an error message in errors.h

docs/02.API-REFERENCE.md
```diff
@@ -2839,6 +2839,74 @@ main (void)
 - [jerry_container_type_t](#jerry_container_type_t)
 
 
+## jerry_get_buffer_from_container
+
+**Summary**
+
+Returns the underlying buffer from a Container, or a Container Iterator. Sets the boolean input value to `true`
+if the container object is a key/value structure and false if not.
```
zherczeg, line 2847: container object *contains* a key/value

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+                                bool *is_key_value_p) /**< [out] is key-value structure */
+{
+  jerry_assert_api_available ();
+
+#if JERRY_BUILTIN_CONTAINER
+  ecma_collection_t *collection_buffer = ecma_new_collection ();
+
+  if (!ecma_is_value_object (value))
+  {
+    return false;
```
zherczeg, line 6454: Throw an error

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+  }
+
+  ecma_object_t *obj_p = ecma_get_object_from_value (value);
+  ecma_extended_object_t *ext_obj_p = (ecma_extended_object_t *) obj_p;
+  ecma_object_type_t type = ecma_get_object_type (obj_p);
+
+  if (jerry_get_container_type (value) == JERRY_CONTAINER_TYPE_INVALID &&
+        !(type == ECMA_OBJECT_TYPE_CLASS &&
```
zherczeg, line 6462: Invalid style

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+    }
+    else if (iterator_kind == ECMA_ITERATOR_VALUES)
+    {
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (entry_p[1]));
+    }
+    else
+    {
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (*entry_p));
+    }
+
```
zherczeg, line 6556: Remove this newline.

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+    if (is_value_iterator)
+    {
+      ext_obj_p->u.cls.u2.iterator_index = (uint16_t) (i + entry_size);
+
+      if (iterator_kind != ECMA_ITERATOR_ENTRIES)
+      {
+        ext_obj_p->u.cls.u2.iterator_index = (uint16_t) (i + entry_size);
+      }
+    }
```
zherczeg, line 6541: Don't overwrite the iterator.

tests/unit-core/test-container.c
```diff
@@ -34,6 +34,70 @@ static const jerry_object_native_info_t native_info =
   .offset_of_references = 0,
 };
 
+static void container_value_test (jerry_value_t check_values, jerry_value_t created_values,
+                                 uint8_t length, bool is_key_value_pairs, bool checking_keys)
+{
+  for (uint8_t i = 0; i < length; i++)
+  {
+
```
zherczeg, line 42: Remove this newline.

zherczeg (changes requested):

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+          (ext_obj_p->u.cls.type == ECMA_OBJECT_CLASS_MAP_ITERATOR ||
+           ext_obj_p->u.cls.type == ECMA_OBJECT_CLASS_SET_ITERATOR)))
+  {
+    ecma_collection_free (collection_buffer);
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG ("Value is not a Container or Iterator")));
+  }
+
+  ecma_collection_t *container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t, ext_obj_p->u.cls.u3.value);
+  lit_magic_string_id_t lit_id = ext_obj_p->u.cls.u2.container_id;
+  uint32_t entry_count =  ECMA_CONTAINER_ENTRY_COUNT (container_p);
```
galpeter, line 6472: Extra space after the `=`.

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+
+    ecma_value_t iterated_value = ext_obj_p->u.cls.u3.iterated_value;
+
+    if (ecma_is_value_empty (iterated_value))
+    {
+      return ecma_op_new_array_object_from_collection (collection_buffer, false);
+    }
+
+    ecma_extended_object_t *map_object_p = (ecma_extended_object_t *) (ecma_get_object_from_value (iterated_value));
+    lit_id = map_object_p->u.cls.u2.container_id;
```
galpeter, line 6496: AFAIK, this can be removed and the `map_object_p->u.cls.u2.container_id` can be used directly in line 6507.

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+    {
+      *is_key_value_p = true;
+    }
+  }
+  else if (lit_id == LIT_MAGIC_STRING_MAP_UL || lit_id == LIT_MAGIC_STRING_WEAKMAP_UL)
+  {
+    *is_key_value_p = true;
+  }
+
+  bool is_key_value_container = *is_key_value_p;
```
galpeter, line 6522: Why do we need this extra variable if it is the same as the `*is_key_value_p`?

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (*entry_p));
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (entry_p[1]));
+    }
+    else if (iterator_kind == ECMA_ITERATOR_VALUES)
+    {
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (entry_p[1]));
+    }
+    else
+    {
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (*entry_p));
```
galpeter, line 6554: Should we add an assert for the `iterator_kind` for expected kinds?

jerry-core/api/jerry.c
```diff
@@ -6428,6 +6429,140 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+  uint32_t index = 0;
+  uint8_t iterator_kind = ECMA_ITERATOR__COUNT;
+  ecma_value_t *start_p = ECMA_CONTAINER_START (container_p);
+  bool is_value_iterator = false;
+
+  *is_key_value_p = false;
+
+  if (type == ECMA_OBJECT_TYPE_CLASS &&
+      (ext_obj_p->u.cls.type == ECMA_OBJECT_CLASS_MAP_ITERATOR ||
+        ext_obj_p->u.cls.type == ECMA_OBJECT_CLASS_SET_ITERATOR))
```
galpeter, line 6484: This is already "calculated" above we could move the bool result of this into a variable before the `if` and just use the value where required.

galpeter (changes requested):

jerry-core/api/jerry.c
Achie72: I don't think there is a need for that. 

- An `ECMA_ITERATOR_ENTRIES` iterator will set the `is_key_value_container` as well as a Map or WeakMap (hence the `ECMA_ITERATOR__COUNT` doesn't matter there).
- The `ECMA_ITERATOR_VALUES` has it's own branch a`
- The `ECMA_ITERATOR_KEYS` will fall into the last one, as well as a Set or Weakset which is the only option where `ECMA_ITERATOR__COUNT` default should matter, but that container type is the same on keys-values in the jerry architecture if i understood it correctly.

Feel free to correct me if i'm wrong and/or missing the point in your comment.

docs/02.API-REFERENCE.md
```diff
@@ -2839,6 +2839,74 @@ main (void)
...
+  jerry_char_t src[] = "var map = new Map(); map.set(1,2); map.entries()";
+  jerry_value_t iterable = jerry_eval (src, sizeof (src) - 1, JERRY_PARSE_NO_OPTS);
+
+  bool is_key_value_container = false;
+  jerry_value_t buffer_from_map = jerry_get_buffer_from_container (iterable, &is_key_value_container);
+  
+  if (jerry_get_array_length (buffer_from_map) == 2)
+  {
+    /* 
+      buffer_from_map contains [1,2]
```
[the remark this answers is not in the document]
Achie72: This function is present in other engines as well, where it works following this logic (basically flattening maps and entry iterators)

docs/02.API-REFERENCE.md
```diff
@@ -2839,6 +2839,74 @@ main (void)
 - [jerry_container_type_t](#jerry_container_type_t)
 
 
+## jerry_get_buffer_from_container
```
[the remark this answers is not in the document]
Achie72: Will add this new section to the end of the README then if it's fine with you.

jerry-core/api/jerry.c
```diff
@@ -6365,7 +6366,7 @@ jerry_create_container (jerry_container_type_t container_type, /**< Type of the
   JERRY_UNUSED (arguments_list_p);
   JERRY_UNUSED (arguments_list_len);
   JERRY_UNUSED (container_type);
-  return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG ("Containers are disabled")));
+  return jerry_throw (ecma_raise_syntax_error (ECMA_ERR_MSG (ecma_error_container_not_supported_p)));
```
zherczeg, line 6369: This is a type error, please keep the original code.

jerry-core/api/jerry.c
```diff
@@ -6426,6 +6427,135 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+ * Sets the boolean input value to `true` if the container object is a
+ * key-value structure and `false` if not.
+ *
+ * Note:
+ *     the returned value must be freed with a jerry_release_value call
+ *
+ * @return jerry_value_t representing an array containing the elemets of the map.
+ *         jerry_value_t representing an empty array, if the Iterator is finished/empty.
+ *         Error if the value is not a Container or Container Iterator.
+ *         `undefined` if the `value` is undefined or null.
```
zherczeg, line 6441: I would siplify this:
`@ return an array of items for maps/sets or their iterators, error otherwise`
Passing an undefined as value is an error.

jerry-core/api/jerry.c
```diff
@@ -6426,6 +6427,135 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+ */
+jerry_value_t
+jerry_get_array_from_container (jerry_value_t value, /**< the container or iterator object */
+                                bool *is_key_value_p) /**< [out] is key-value structure */
+{
+  jerry_assert_api_available ();
+
+  if (jerry_value_is_undefined (value) || jerry_value_is_null (value))
+  {
+    return jerry_create_undefined ();
```
zherczeg, line 6451: No need special case for this.

jerry-core/api/jerry.c
```diff
@@ -6426,6 +6427,135 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+  uint32_t entry_count;
+  uint8_t entry_size;
+
+  uint32_t index = 0;
+  uint8_t iterator_kind = ECMA_ITERATOR__COUNT;
+  ecma_value_t *start_p;
+
+  *is_key_value_p = false;
+
+  if (type == ECMA_OBJECT_TYPE_CLASS
```
zherczeg, line 6476: maps/sets also has this type, we could return early if this check fails. There is no need a `type` variable then in the code.

jerry-core/api/jerry.c
```diff
@@ -6426,6 +6427,135 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+  for (uint32_t i = index; i < entry_count; i+= entry_size)
+  {
+    ecma_value_t *entry_p = start_p + i;
+
+    if (ecma_is_value_empty (*entry_p))
+    {
+      continue;
+    }
+
+    if (*is_key_value_p)
```
zherczeg, line 6536: I would keep this as a local variable as well, no need for performing a `*` all the time.

tests/unit-core/test-container.c
```diff
@@ -122,6 +167,73 @@ main (void)
   jerry_gc (JERRY_GC_PRESSURE_LOW);
   TEST_ASSERT (global_counter == 1);
 
+  run_eval ("function assert (v) {\n"
+            "  if (v !== true)\n"
+            "    throw 'Assertion failed!'\n"
+            "}");
+
+  run_eval ("function test_values (arr1, arr2) {\n"
+            "  assert (arr1.length == arr2.length);\n"
```
zherczeg, line 176: Use `===`. Furthermore, since this code requires ESNEXT, you can also use `Array.isArray()` to check that arr1 is an Array.

tests/unit-core/test-container.c
```diff
@@ -122,6 +167,73 @@ main (void)
...
+            "  for (let i = 0; i < arr1.length; i++) {\n"
+            "    assert (arr1[i] == arr2[i]);\n"
```
zherczeg, line 178: Ditto.

tests/unit-core/test-container.c
```diff
@@ -122,6 +167,73 @@ main (void)
...
+
+  run_eval ("var entries_check_array = [1, 3.14, 2, true, 3, 'foo'];\n"
+            "var values_check_array = [3.14, true, 'foo'];\n"
+            "var keys_check_array = [1, 2, 3,];\n"
+            "var obj = { x:3, y:'foo'};\n"
+            "var b_int = 1n;\n"
+            "var object_check_array = [1, obj];\n");
+
+  run_eval ("var map = new Map();\n"
+            "map.set(1,3.14);\n"
```
zherczeg, line 190: Style: add space after comma. Similar cases below.

tests/unit-core/test-container.c
```diff
@@ -122,6 +167,73 @@ main (void)
...
+            "map.set(2,true);\n"
+            "map.set(3,'foo');\n"
+            "var set = new Set();\n"
+            "set.add(3.14);\n"
+            "set.add(true);\n"
+            "set.add('foo');\n"
+            "var obj_map = new Map();\n"
+            "obj_map.set(1, obj);\n");
+
+  run_eval ("var result = create_array_from_container (map, true);\n"
```
zherczeg, line 200: You are put a space before `(` but not in the above cases. Please be consistent.

tests/unit-core/test-container.c
```diff
@@ -122,6 +167,73 @@ main (void)
...
+            "test_values (result, entries_check_array);");
+
+  run_eval ("var result = create_array_from_container (set, false);\n"
+            "test_values (result, values_check_array);");
```
zherczeg, line 204: Even if there are repeats, I would put the arrays here:
`test_values (result, [3.14, true, 'foo']);`
More readable.

tests/unit-core/test-container.c
```diff
@@ -122,6 +167,73 @@ main (void)
...
+
+  run_eval ("var result = create_array_from_container (obj_map, true)\n"
+            "test_values (result, object_check_array);");
+
+  run_eval ("var map = new Map();\n"
+            "map.set(1,1);\n"
+            "var iter = map.entries();\n"
+            "iter.next();\n"
+            "var result = create_array_from_container (iter, false);\n"
+            "assert (result.length == 0);");
```
zherczeg, line 223: Check that result is an array.

zherczeg (changes requested):

jerry-core/api/jerry.c
zherczeg: Becasue the compiler can optimize this case. Otherwise it needs to expect that is can be changed randomly externally.

docs/02.API-REFERENCE.md
```diff
@@ -11924,3 +11924,76 @@ main (void)
...
 }
 ```
+
+# Container Functions
+
+## jerry_get_buffer_from_container
+
+**Summary**
+
+Return a new array containing elements from a Container, or a Container Iterator. Sets the boolean input value to `true`
```
zherczeg, line 11934: Sets the `is_key_value_p` to `true`

docs/02.API-REFERENCE.md
```diff
@@ -11924,3 +11924,76 @@ main (void)
...
+- The es.next profile enables this by default.
+
+*New in version [[NEXT_RELEASE]]*.
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_get_buffer_from_container(jerry_value_t value,
+                                bool* is_key_value_p)
```
zherczeg, line 11951: Missing semicolon at the end.

docs/02.API-REFERENCE.md
```diff
@@ -11924,3 +11924,76 @@ main (void)
...
+```
+
+- `value` - Container object
+- `is_key_value` - Will be set to `true` if the given container has key-value pairs, `false` otherwise.
+- return value
+  - jerry_value_t containing an array of values of the underlying buffer, if the orignal is not empty
```
zherczeg, line 11957: What is the original? I think the description is too long, this could go to the notes, and `- jerry_value_t containing an array of values` could be enough here

docs/02.API-REFERENCE.md
```diff
@@ -11924,3 +11924,76 @@ main (void)
...
+
+**Prototype**
+
+```c
+jerry_value_t
+jerry_get_buffer_from_container(jerry_value_t value,
+                                bool* is_key_value_p)
+```
+
+- `value` - Container object
```
zherczeg, line 11954: Map/Set or iterator object

docs/02.API-REFERENCE.md
```diff
@@ -11924,3 +11924,76 @@ main (void)
...
+{
+  jerry_init (JERRY_INIT_EMPTY);
+
+  jerry_char_t src[] = "var map = new Map(); map.set(1,2); map.entries()";
+  jerry_value_t iterable = jerry_eval (src, sizeof (src) - 1, JERRY_PARSE_NO_OPTS);
+
+  bool is_key_value_container = false;
+  jerry_value_t buffer_from_map = jerry_get_array_from_container (iterable, &is_key_value_container);
+  
+  if (jerry_get_array_length (buffer_from_map) == 2)
```
zherczeg, line 11978: This `if` can be removed here, the comment is enough.

docs/02.API-REFERENCE.md
```diff
@@ -11924,3 +11924,76 @@ main (void)
...
+  {
+    /* 
+      buffer_from_map contains will contain elements [1,2] as 1 is the key and 2 is the value from the
```
zherczeg, line 11981: The buffer_from_map contains two elements: 1 and 2, which is the key/value pair of the only item in the set.

jerry-core/api/jerry.c
```diff
@@ -6480,6 +6481,127 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
   return JERRY_CONTAINER_TYPE_INVALID;
 } /* jerry_get_container_type */
 
+/**
+ * Return a new array containing elements from a Container or a Container Iterator.
+ * Sets the boolean input value to `true` if the container object is a
```
zherczeg, line 6486: if the container object has key/value pairs

jerry-core/api/jerry.c
```diff
@@ -6480,6 +6481,127 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+jerry_get_array_from_container (jerry_value_t value, /**< the container or iterator object */
+                                bool *is_key_value_p) /**< [out] is key-value structure */
+{
+  jerry_assert_api_available ();
+
+#if JERRY_BUILTIN_CONTAINER
+
+  if (!ecma_is_value_object (value))
+  {
+    return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG ("Value is not a Container or Iterator")));
```
zherczeg, line 6504: You could put this error message to a local value and use it here and below:
`const char *container_needed = ECMA_ERR_MSG ("Value is not a Container or Iterator"));`

jerry-core/api/jerry.c
```diff
@@ -6480,6 +6481,127 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+
+    if (ext_obj_p->u.cls.u2.container_id == LIT_MAGIC_STRING_MAP_UL
+        || ext_obj_p->u.cls.u2.container_id == LIT_MAGIC_STRING_WEAKMAP_UL)
+    {
+      is_key_value_pair = true;
+    }
+  }
+  else
+  {
+    ecma_collection_free (collection_buffer);
```
zherczeg, line 6573: The construction of `collection_buffer` could be moved before the `for`, and it would not needed to be freed here.

jerry-core/api/jerry.c
```diff
@@ -6480,6 +6481,127 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+    ecma_value_t *entry_p = start_p + i;
+
+    if (ecma_is_value_empty (*entry_p))
+    {
+      continue;
+    }
+
+    if (iterator_kind != ECMA_ITERATOR_VALUES)
+    {
+      ecma_collection_push_back (collection_buffer, ecma_copy_value_if_not_object (entry_p[0])); //
```
zherczeg, line 6588: Remove the `//` at the end.

jerry-core/api/jerry.c
```diff
@@ -6480,6 +6481,127 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+    }
+
+    if (is_key_value_pair || iterator_kind == ECMA_ITERATOR_VALUES)
```
zherczeg, line 6591: iterator_kind != ECMA_ITERATOR_KEYS or something?
This enum should have 3 values: keys, values, entries. The latter two only valid for maps, sets only have "keys". 
The first `if` is correct, the second should be simplified.

jerry-core/api/jerry.c
```diff
@@ -6480,6 +6481,127 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+  ecma_collection_t *collection_buffer = ecma_new_collection ();
+  ecma_collection_t *container_p;
+  uint32_t entry_count;
+  uint8_t entry_size;
+
+  uint32_t index = 0;
+  uint8_t iterator_kind = ECMA_ITERATOR__COUNT;
+  ecma_value_t *start_p;
+
+  bool is_key_value_pair = false;
```
zherczeg, line 6525: This local variable should be unnecessary, `iterator_kind` should tell its value.

zherczeg (changes requested):

jerry-core/api/jerry.c
Achie72: My problem here, and hence the `if` statement, is the following:

- You need to push the `key` on all, except `ECMA_ITERATOR_VALUES`
- You need the `value` on all, except `Set` and `ECMA_ITERATOR_KEY`

And this latter one is the culprit, as it's not a tri-value but a quad, the code - as of now - uses `ECMA_ITERATOR__COUNT` for Sets to distinguish them. Because of this Sets will slip through the `iterator_kind != ECMA_ITERATOR_KEYS` statement and have their "values" pushed as well. The `is_key_value_pair` is there to let only Maps through from the two non-iterator collections. 


Now that i think about it some more, this could be "simplified" down to:
`iterator_kind != ECMA_ITERATOR_KEYS && iterator_kind != ECMA_ITERATOR__COUNT`, which allow the cutting out of the `is_key_value_pair`, if that solution is fine with you.

jerry-core/api/jerry.c
```diff
@@ -6507,6 +6508,127 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+      return ecma_op_new_array_object_from_collection (ecma_new_collection (), false);
+    }
+
+    ecma_extended_object_t *map_object_p = (ecma_extended_object_t *) (ecma_get_object_from_value (iterated_value));
+
+    container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t, map_object_p->u.cls.u3.value);
+    entry_count = ECMA_CONTAINER_ENTRY_COUNT (container_p);
+    index = ext_obj_p->u.cls.u2.iterator_index;
+
+    if (index == entry_count)
```
zherczeg, line 6569: Is this `if` needed?

zherczeg (approved):
Please fix the `if (index == entry_count)` check and I am ok with this patch.

docs/02.API-REFERENCE.md
```diff
@@ -12006,3 +12006,72 @@ main (void)
   return 0;
 }
 ```
+
+# Container Functions
+
+## jerry_get_buffer_from_container
```
galpeter, line 12012: The real API method does not match the documentation "jerry_get_array_from_container " vs "jerry_get_buffer_from_container". The docs should be updated.

tests/unit-core/test-container.c
```diff
@@ -122,6 +186,155 @@ main (void)
...
+    jerry_value_t iterable = jerry_eval (src, sizeof (src) - 1, JERRY_PARSE_NO_OPTS);
+
+    bool is_key_value_container = false;
+    jerry_value_t buffer_from_map = jerry_get_buffer_from_container (iterable, &is_key_value_container);
+    TEST_ASSERT (jerry_value_is_error (buffer_from_map));
+
+    jerry_release_value (iterable);
+    jerry_release_value (buffer_from_map);
+  }
+
```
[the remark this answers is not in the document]
galpeter: What about these tests?

jerry-core/api/jerry.c
```diff
@@ -6507,6 +6508,122 @@ jerry_get_container_type (const jerry_value_t value) /**< the container object *
...
+    container_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_collection_t, map_object_p->u.cls.u3.value);
+    entry_count = ECMA_CONTAINER_ENTRY_COUNT (container_p);
+    index = ext_obj_p->u.cls.u2.iterator_index;
+
+    entry_size = ecma_op_container_entry_size (map_object_p->u.cls.u2.container_id);
+    start_p = ECMA_CONTAINER_START (container_p);
+
+    iterator_kind = ext_obj_p->u.cls.u1.iterator_kind;
+
+    *is_key_value_p = (iterator_kind == ECMA_ITERATOR_ENTRIES);
```
galpeter, line 6574: If I understand correctly then this can be moved after this whole `if-elseif` block and no need to check it in both case and then the line 6589 can be removed

galpeter (approved):
lgtm

PR #4674, opened 2021-05-20
Fix native error check

zherczeg:


rerobika (approved):

jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -312,26 +314,14 @@ ecma_new_aggregate_error (ecma_value_t error_list_val, /**< errors list */
...
-
-  for (uint8_t idx = 0; idx < sizeof (ecma_error_mappings) / sizeof (ecma_error_mappings[0]); idx++)
-  {
-    if (ecma_error_mappings[idx].error_prototype_id == builtin_id)
-    {
-      return ecma_error_mappings[idx].error_type;
-    }
-  }
-
-  return JERRY_ERROR_NONE;
```
galpeter, line 334: AFAIK: After this then we would not need the `ecma_error_mappings` array at all (and related type).

jerry-core/ecma/base/ecma-globals.h
```diff
@@ -1032,6 +1032,7 @@ typedef struct
       union
       {
         uint8_t arguments_flags; /**< arguments object flags */
+        uint8_t error_type; /**< type of native error objects */
```
galpeter, line 1035: Should we mention here that this basically stores a "jerry_error_t" type value? I mean mention the "jerry_error_t" type here.

galpeter (changes requested):

galpeter (approved):
lgtm

PR #4678, opened 2021-05-23
Python debugger support for Python 3 (in addition to Python 2)

zmedico:
- Added safe_ord compatibility to pass through int arguments
- Fixed JerryDebugger to decode bytes as utf8 strings when necessary
- Fixed WebSocket send_message method to use packed_data[0:1] bytes slice

JerryScript-DCO-1.0-Signed-off-by: Zac Medico <zmedico@gmail.com>

### DIFF
--- a/jerry-debugger/jerry_client_main.py
+++ b/jerry-debugger/jerry_client_main.py
@@ -149,6 +149,15 @@ def arguments_parse():
     return args
 
 
+if sys.version_info.major >= 3:
+    def safe_ord(c):
+        if isinstance(c, int):
+            return c
+        return ord(c)
+else:
+	safe_ord = ord
+
+
 class JerryBreakpoint(object):
 
     def __init__(self, line, offset, function):
@@ -307,12 +316,12 @@ class JerryDebugger(object):
         # cpointer_size [1]
         result = self.channel.connect(config_size)
 
-        if len(result) != config_size or ord(result[0]) != JERRY_DEBUGGER_CONFIGURATION:
+        if len(result) != config_size or safe_ord(result[0]) != JERRY_DEBUGGER_CONFIGURATION:
             raise Exception("Unexpected configuration")
 
-        self.little_endian = ord(result[1]) & JERRY_DEBUGGER_LITTLE_ENDIAN
-        self.max_message_size = ord(result[6])
-        self.cp_size = ord(result[7])
+        self.little_endian = safe_ord(result[1]) & JERRY_DEBUGGER_LITTLE_ENDIAN
+        self.max_message_size = safe_ord(result[6])
+        self.cp_size = safe_ord(result[7])
 
         if self.little_endian:
             self.byte_order = "<"
@@ -400,7 +409,7 @@ class JerryDebugger(object):
                    "to clear all the given breakpoints\n "
         elif args in ['all', 'pending', 'active']:
             if args != "pending":
-                for i in self.active_breakpoint_list.values():
+                for i in list(self.active_breakpoint_list.values()):
                     breakpoint = self.active_breakpoint_list[i.active_index]
                     del self.active_breakpoint_list[i.active_index]
                     breakpoint.active_index = -1
@@ -561,6 +570,7 @@ class JerryDebugger(object):
         self._exec_command(JERRY_DEBUGGER_MEMSTATS)
 
     def _send_string(self, args, message_type, index=0):
+        args = args.encode("utf8")
 
         # 1: length of type byte
         # 4: length of an uint32 value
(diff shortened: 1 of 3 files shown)


jerry-debugger/jerry_client_main.py
```diff
@@ -1151,33 +1163,34 @@ def _process_incoming_text(self, buffer_type, data):
...
 
                 message = self.current_log + message
-                lines = message.split("\n")
-                self.current_log = lines.pop()
+                lines = message.decode("utf8").split("\n")
+                self.current_log = lines.pop().encode("utf8")
 
                 return "".join(["%s%s\n" % (log_type, line) for line in lines])
 
             if not message.endswith("\n"):
```
galpeter, line 1180: This check will fail if the message is a byte string. So we'll need to decode the string just before this.

jerry-debugger/jerry_client_main.py
```diff
@@ -1151,33 +1163,34 @@ def _process_incoming_text(self, buffer_type, data):
...
-                return "%serr: %s%s" % (self.red, self.nocolor, message)
+                return "%serr: %s%s" % (self.red, self.nocolor, message.decode("utf8"))
             elif subtype == JERRY_DEBUGGER_OUTPUT_TRACE:
-                return "%strace: %s%s" % (self.blue, self.nocolor, message)
+                return "%strace: %s%s" % (self.blue, self.nocolor, message.decode("utf8"))
 
         # Subtypes of eval
         self.prompt = True
 
+        message = message.decode("utf8")
```
galpeter, line 1193: This decode is only required if the `buffer_type == JERRY_DEBUGGER_OUTPUT_RESULT_END` check above is not true.

jerry-debugger/jerry_client_main.py
```diff
@@ -149,6 +149,17 @@ def arguments_parse():
...
 
+if sys.version_info.major >= 3:
+    # pylint: disable=invalid-name
+    _ord_orig = ord
+    def _ord_compat(c):
+        if isinstance(c, int):
+            return c
+        return _ord_orig(c)
+    # pylint: disable=redefined-builtin
+    ord = _ord_compat
```
galpeter, line 160: Was wondering which one would be better:
A) Having this ord compatibility paching or
B) Having a new method (`safe_ord`?) which replaces all occurrences of the `ord` calls in the python code and does the same as this compatibility method.

In case of B we could also remove these newly added pylint disable comments.

Also, the `jerry_client_rawpacket.py` also needs the same change.

galpeter (changes requested):
Have you tried running the debugger tests? ` ./tools/run-tests.py --jerry-debugger` for me this returned various errors which should be also fixed for Python 3.

(I've did a quick hackish changes to resolve some of them: https://gist.github.com/galpeter/a6ac851b6ab113aaad62d60001c5df23 )

zmedico:
> Have you tried running the debugger tests? ` ./tools/run-tests.py --jerry-debugger` for me this returned various errors which should be also fixed for Python 3.
> 
> (I've did a quick hackish changes to resolve some of them: https://gist.github.com/galpeter/a6ac851b6ab113aaad62d60001c5df23 )

I've included your changes, and the debugger tests do pass with them. Thank you! I will look into making you suggested `safe_ord` change now.

jerry-debugger/jerry_client_main.py
```diff
@@ -903,11 +915,11 @@ def _parse_source(self, data):
...
-                              "source_name": source_code_name,
+                stack.append({"source": source_code.decode("utf8"),
+                              "source_name": source_code_name.decode("utf8"),
                               "line": position[0],
                               "column": position[1],
-                              "name": function_name,
+                              "name": function_name.decode("utf8"),
                               "lines": [],
                               "offsets": []})
                 function_name = ""
```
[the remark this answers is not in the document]
zmedico: Thanks, I've applied your change from https://gist.github.com/galpeter/a6ac851b6ab113aaad62d60001c5df23.

jerry-debugger/jerry_client_main.py
zmedico: Thanks, I've applied you change from https://gist.github.com/galpeter/a6ac851b6ab113aaad62d60001c5df23.

jerry-debugger/jerry_client_main.py
zmedico: Thanks, I've applied your change from https://gist.github.com/galpeter/a6ac851b6ab113aaad62d60001c5df23.

jerry-debugger/jerry_client_main.py
zmedico: I've changed it to use a new `safe_ord` function as you've suggested.

lygstate:
this can be closed, python3 is already supported?

zmedico:
> this can be closed, python3 is already supported?

Thank you! Closing now.

PR #4690, opened 2021-06-15
Rework external string free operation

zherczeg:


tests/unit-core/test-external-string.c
```diff
@@ -53,38 +65,45 @@ main (void)
...
   TEST_ASSERT (free_count == 1);
   jerry_release_value (external_string);
   TEST_ASSERT (free_count == 1);
 
-  external_string = jerry_create_external_string ((jerry_char_t *) external_2, free_external2);
+  jerry_string_set_external_string_free_callback (external_string_free_callback_2);
+  external_string = jerry_create_external_string ((jerry_char_t *) external_2, (void *) &free_count);
   TEST_ASSERT (free_count == 2);
   jerry_release_value (external_string);
   TEST_ASSERT (free_count == 2);
```
galpeter, line 84: Shouldn't the `free_count` change after the string release? Or am I missing something? (Maybe we need a force gc() call?)

zherczeg: Nope. external_2 is `"Object"`, and a magic string is used, so the callback is called by `jerry_create_external_string`.

zherczeg: This is always worked this way. See `ecma_new_ecma_external_string_from_cesu8`. I will add some words about it in the documentation, that is a good idea.

zherczeg:
Patch is updated, please check it.

rerobika (approved):

galpeter (approved):
lgtm

PR #4707, opened 2021-07-05
Rework storing the line/column/bytecode info

zherczeg:
This information is stored in a separate memory block instead of being part of the byte code. Snapshot does not supported.

Note: the patch focuses on line information encoding (compressing) and decoding. It does not aim to provide better line information (still statement granularity, although multiple statements in a single line are now supported). This should be done later. We should also think about convenient line info support in gdb, since `vm_context->current_line` is not available anymore.


zherczeg:
Patch is ready for review.

rerobika (approved):
LGTM

zherczeg:
Patch is updated, please review.

galpeter (approved):
lgtm

PR #4710, opened 2021-07-14
Support shared user data for scripts

zherczeg:
The same data is returned for the script and all of its functions, including those which are created by an eval call.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -243,6 +243,7 @@ Option bits for [jerry_parse_options_t](#jerry_parse_options_t).
  - JERRY_PARSE_MODULE - Parse source as an ECMAScript module
  - JERRY_PARSE_HAS_RESOURCE - `resource_name_p` and `resource_name_length` fields are valid
  - JERRY_PARSE_HAS_START - `start_line` and `start_column` fields are valid
+ - JERRY_PARSE_HAS_USER_VALUE - `user_value` field is valid
 
 *New in version [[NEXT_RELEASE]]*.
 
@@ -310,6 +311,10 @@ Flags for [jerry_exec_snapshot](#jerry_exec_snapshot) functions:
  - JERRY_SNAPSHOT_EXEC_COPY_DATA - copy snapshot data into memory (see below)
  - JERRY_SNAPSHOT_EXEC_ALLOW_STATIC - allow executing static snapshots
  - JERRY_SNAPSHOT_EXEC_LOAD_AS_FUNCTION - load snapshot as function instead of executing it
+ - JERRY_SNAPSHOT_EXEC_HAS_RESOURCE - resource_name_p and resource_name_length fields are valid
+                                      in [jerry_exec_snapshot_option_values_t](#jerry_exec_snapshot_option_values_t)
+ - JERRY_SNAPSHOT_EXEC_HAS_USER_VALUE - user_value field is valid
+                                        in [jerry_exec_snapshot_option_values_t](#jerry_exec_snapshot_option_values_t)
 
 *Changed in version [[NEXT_RELEASE]]*: The `JERRY_SNAPSHOT_EXEC_LOAD_AS_FUNCTION` value is added,
                                        which replaces the `jerry_load_function_snapshot` function.
@@ -536,6 +541,8 @@ typedef struct
                                 *   if JERRY_PARSE_HAS_RESOURCE is set in options */
   uint32_t start_line; /**< start line of the source code if JERRY_PARSE_HAS_START is set in options */
   uint32_t start_column; /**< start column of the source code if JERRY_PARSE_HAS_START is set in options */
+  jerry_value_t user_value; /**< user value assigned to all functions created by this script including eval
+                             *   calls executed by the script if JERRY_PARSE_HAS_USER_VALUE is set in options */
 } jerry_parse_options_t;
 ```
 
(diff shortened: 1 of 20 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -10219,19 +10224,24 @@ jerry_value_t
 jerry_exec_snapshot (const uint32_t *snapshot_p,
                      size_t snapshot_size,
                      size_t func_index,
-                     uint32_t exec_snapshot_opts);
+                     uint32_t exec_snapshot_opts,
+                     const jerry_parse_options_t *options_p);
```
galpeter, line 10228: This seems weird to me. The method "executes" something and we would need to pass "parse" options to it. In addition the comment below mentions "loading options".  So which one is this then? execute/parse/load?

docs/02.API-REFERENCE.md
```diff
@@ -10306,19 +10317,24 @@ jerry_value_t
 jerry_load_function_snapshot (const uint32_t *snapshot_p,
                               size_t snapshot_size,
                               size_t func_index,
-                              uint32_t exec_snapshot_opts);
+                              uint32_t exec_snapshot_opts,
+                              const jerry_parse_options_t *options_p);
```
galpeter, line 10321: See comments for "jerry_exec_snapshot"

docs/02.API-REFERENCE.md
```diff
@@ -11178,6 +11195,71 @@ main (void)
 
 - [jerry_create_external_function](#jerry_create_external_function)
 
+## jerry_get_user_value
```
galpeter, line 11198: Is it required to release the return value? Or we can leave it forever alone?

rerobika (approved):

jerry-core/api/jerry-snapshot.c
```diff
@@ -997,16 +1001,68 @@ jerry_snapshot_result (const uint32_t *snapshot_p, /**< snapshot */
...
+
+    if (user_value != ECMA_VALUE_EMPTY)
+    {
+      script_p->refs_and_type = (CBC_SCRIPT_REF_ONE | CBC_SCRIPT_USER_VALUE);
+
+      if (ecma_is_value_object (user_value))
+      {
+        script_p->refs_and_type = (CBC_SCRIPT_REF_ONE | CBC_SCRIPT_USER_OBJECT);
+      }
+    }
```
galpeter, line 1026: This code part seems very similar to the changes in the parser (see: https://github.com/jerryscript-project/jerryscript/pull/4710/files#diff-baabef101cbc924f0aff83bdc8e60e17890ed01390e7fcc9fd519c8b07bb4171R1846 ). Maybe we could have a common method for this?

docs/02.API-REFERENCE.md
zherczeg: You can pass the resource name and user value to this function. These are parts of the `jerry_parse_options_t`, I don't want to introduce another structure for it. Later the resource name could be stored in the snapshot, and only needs to be overwritten if needed. Maybe simple user values can be stored as well.

jerry-core/api/jerry-snapshot.c
zherczeg: Similar but not the same. I think it is not worth for combining it. If you insist, where should we put the function? Into the parser or into the snapshot?

jerry-core/ecma/base/ecma-helpers.c
```diff
@@ -1576,6 +1603,72 @@ ecma_bytecode_deref (ecma_compiled_code_t *bytecode_p) /**< byte code pointer */
...
+        }
+        return NULL;
+      }
+      case ECMA_OBJECT_TYPE_BOUND_FUNCTION:
+      {
+        ecma_extended_object_t *ext_object_p = (ecma_extended_object_t *) object_p;
+
+        object_p = ECMA_GET_NON_NULL_POINTER_FROM_POINTER_TAG (ecma_object_t,
+                                                               ext_object_p->u.bound_function.target_function);
+        break;
```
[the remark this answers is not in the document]
zherczeg: This is a `while(true)` loop which continues with the next iteration.

jerry-core/parser/js/js-parser.c
```diff
@@ -1841,24 +1825,81 @@ parser_parse_source (const uint8_t *arg_list_p, /**< function argument list */
...
+
+  if (context.user_value != ECMA_VALUE_EMPTY)
+  {
+    context.script_p->refs_and_type = (CBC_SCRIPT_REF_ONE | CBC_SCRIPT_USER_VALUE);
+
+    if (ecma_is_value_object (context.user_value))
+    {
+      context.script_p->refs_and_type = (CBC_SCRIPT_REF_ONE | CBC_SCRIPT_USER_OBJECT);
+    }
+  }
```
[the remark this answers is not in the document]
zherczeg: I don't know if the compiler is able to detect that the `or` opertation is unnecessary. But propbably not that performance critical anyway.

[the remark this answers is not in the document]
zherczeg: I invented a macro for this.

jerry-core/api/jerry-snapshot.c
zherczeg: I have added a macro for it.

jerry-core/ecma/base/ecma-helpers.c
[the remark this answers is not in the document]
galpeter: Ahh true, missed that.

jerry-core/api/jerry-snapshot.c
```diff
@@ -998,16 +998,58 @@ jerry_exec_snapshot (const uint32_t *snapshot_p, /**< snapshot */
...
+
+    if ((exec_snapshot_opts & JERRY_SNAPSHOT_EXEC_HAS_USER_VALUE)
+        && option_values_p != NULL)
+    {
+      user_value = option_values_p->user_value;
+    }
+
+    uint32_t script_size = (user_value != ECMA_VALUE_EMPTY ? sizeof (cbc_script_user_t)
+                                                           : sizeof (cbc_script_t));
+    cbc_script_t *script_p = jmem_heap_alloc_block (script_size);
```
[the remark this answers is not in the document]
zherczeg: That is a different function. That can return with null, this one is not.

galpeter (approved):
lgtm

PR #4719, opened 2021-07-16
Replace jerry_load_function_snapshot function with a flag

zherczeg:


docs/02.API-REFERENCE.md
```diff
@@ -305,11 +305,14 @@ when the snapshot is generated and executed. Furthermore the
...
-Flags for [jerry_exec_snapshot](#jerry_exec_snapshot) and
-[jerry_load_function_snapshot](#jerry_load_function_snapshot) functions:
+Flags for [jerry_exec_snapshot](#jerry_exec_snapshot) functions:
 
  - JERRY_SNAPSHOT_EXEC_COPY_DATA - copy snapshot data into memory (see below)
  - JERRY_SNAPSHOT_EXEC_ALLOW_STATIC - allow executing static snapshots
+ - JERRY_SNAPSHOT_EXEC_LOAD_AS_FUNCTION - load snapshot as function instead of executing it
+
+*Changed in version [[NEXT_RELEASE]]*: The `JERRY_SNAPSHOT_EXEC_LOAD_AS_FUNCTION` value is added,
+                                       which replaces the `jerry_load_function_snapshot_at` function.
```
galpeter, line 315: * There is no "jerry_load_function_snapshot_at" function.
* This new option should be clearly mentioned in the "jerry_exec_snapshot" docs as with this we are merging two functions (also we need a relevant example).

jerry-core/api/jerry-snapshot.c
```diff
@@ -917,26 +917,27 @@ jerry_generate_snapshot (const jerry_char_t *source_p, /**< script source */
 #endif /* JERRY_SNAPSHOT_SAVE */
 } /* jerry_generate_snapshot */
 
-#if JERRY_SNAPSHOT_EXEC
 /**
- * Execute/load snapshot from specified buffer
+ * Execute snapshot from specified buffer
```
galpeter, line 921: With the new flag this will load the function, and will return a function object (if I understand correctly). So removing the load seems weird. Maybe the whole method description should be reworded. Maybe: "Executes a snapshot from as specified buffer or if <flag> is used this will load/return it as a function object.".

docs/02.API-REFERENCE.md
```diff
@@ -10284,106 +10286,6 @@ main (void)
...
-- `exec_snapshot_opts` - any combination of [jerry_exec_snapshot_opts_t](#jerry_exec_snapshot_opts_t) flags.
-- return value
-  - function object built from the snapshot.
-  - thrown error, otherwise.
-
-*New in version 2.0*.
-
-**Example**
-
-[doctest]: # ()
```
galpeter, line 10324: AFAIK, the example should be moved and adapted for the "jerry_exec_snapshot" method's doc.

galpeter (approved):
lgtm

ossy-szeged (approved):
LGTM

PR #4724, opened 2021-07-27
Change resource name to a string

zherczeg:
This patch has various small advantages:
- resource name can be utf8 string
- resource name is not kept forever in literal storage
- plays better with user vale

The API is not released, so this is not a breaking change.


rerobika (approved):

galpeter (approved):
lgtm

PR #4726, opened 2021-07-29
Implement vm_throw callback

zherczeg:
Slightly improve the description of vm_exec_stop callback.


docs/02.API-REFERENCE.md
```diff
@@ -11320,8 +11347,7 @@ backtrace_callback (jerry_backtrace_frame_t *frame_p,
 
 **Summary**
 
-When JERRY_FEATURE_VM_EXEC_STOP is enabled a callback function can be
-specified by this function. This callback is periodically called when
+The callback passed to this function is periodically called when
```
galpeter, line 11350: How is this change related to the new vm throw? AFAIK these changes should not be in this PR, but in a separate one.

docs/02.API-REFERENCE.md
```diff
@@ -11393,12 +11424,77 @@ main (void)
...
+```c
+void
+jerry_set_vm_throw_callback (jerry_vm_throw_callback_t throw_cb,
+                             void *user_p);
+```
+
+- `throw_cb` - callback which is called on throws (passing NULL disables this feature)
+- `user_p` - user pointer passed to the `throw_cb` function
+
+*New in version 2.0*.
```
galpeter, line 11453: This version info is incorrect, as it should be `[[NEXT_RELEASE]]`

docs/02.API-REFERENCE.md
```diff
@@ -11393,12 +11424,77 @@ main (void)
...
+  const jerry_char_t script[] = "try { throw new Error('1') } catch (e) { throw new Error('2') }";
+
+  jerry_value_t parsed_code = jerry_parse (script, sizeof (script) - 1, NULL);
+  jerry_release_value (jerry_run (parsed_code));
+  jerry_release_value (parsed_code);
+
+  /* The counter contains 2. */
+
+  jerry_cleanup ();
+}
```
galpeter, line 11491: There should be a `return 0` as the `main` expects an `int` return value.

docs/02.API-REFERENCE.md
```diff
@@ -11393,12 +11424,77 @@ main (void)
...
+
+```c
+#include "jerryscript.h"
+
+static void
+vm_throw_callback (const jerry_value_t error_value, /**< captured error */
+                   void *user_p) /**< user pointer */
+{
+  (void) error_value;
+  (void) user_p;
```
galpeter, line 11467: No need for this line.

docs/02.API-REFERENCE.md
zherczeg: I think it is not worth it. Small QOL changes.

zherczeg:

galpeter (approved):
lgtm

jerry-core/jcontext/jcontext.c
```diff
@@ -113,8 +113,14 @@ jcontext_take_exception (void)
...
-  jcontext_set_abort_flag (false);
-  jcontext_set_exception_flag (false);
+#if JERRY_VM_THROW
+  JERRY_CONTEXT (status_flags) &= (uint32_t) ~(ECMA_STATUS_EXCEPTION
+                                               | ECMA_STATUS_ABORT
+                                               | ECMA_STATUS_ERROR_THROWN);
+#else /* !JERRY_VM_THROW */
+  JERRY_CONTEXT (status_flags) &= (uint32_t) ~(ECMA_STATUS_EXCEPTION
+                                               | ECMA_STATUS_ABORT);
+#endif /* JERRY_VM_THROW */
```
rerobika, line 123: How about this?

```c
  JERRY_CONTEXT (status_flags) &= (uint32_t) ~(ECMA_STATUS_EXCEPTION
  #if JERRY_VM_THROW
                                               | ECMA_STATUS_ERROR_THROWN
  #endif /* JERRY_VM_THROW */
                                               | ECMA_STATUS_ABORT);
```

rerobika (approved):

PR #4728, opened 2021-08-02
Support parsing of scripts / functions stored in string values

zherczeg:
Function arguments must be passed as string values.

This patch is mostly a preparation for a future `toString` support. Passing scripts as strings is optional, but it can reduce memory consumption if used well.

Function arguments must be string values after this patch is landed. I think these are used in special cases, and most applications use only a few argument list variations. This way we encourage to store them as global values. The worst case is some extra memory consumption during parsing, since function arguments are rarely large strings. External strings is also a way to control them.

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -243,6 +243,7 @@ Option bits for [jerry_parse_options_t](#jerry_parse_options_t).
  - JERRY_PARSE_NO_OPTS - No options passed
  - JERRY_PARSE_STRICT_MODE - Enable strict mode
  - JERRY_PARSE_MODULE - Parse source as an ECMAScript module
+ - JERRY_PARSE_HAS_ARGUMENT_LIST - `argument_list` field is valid
  - JERRY_PARSE_HAS_RESOURCE - `resource_name` field is valid
  - JERRY_PARSE_HAS_START - `start_line` and `start_column` fields are valid
  - JERRY_PARSE_HAS_USER_VALUE - `user_value` field is valid
@@ -278,8 +279,7 @@ List of backtrace frame types returned by
 
 ## jerry_generate_snapshot_opts_t
 
-Flags for [jerry_generate_snapshot](#jerry_generate_snapshot) and
-[jerry_generate_function_snapshot](#jerry_generate_function_snapshot) functions:
+Flags for [jerry_generate_snapshot](#jerry_generate_snapshot):
 
  - JERRY_SNAPSHOT_SAVE_STATIC - generate static snapshot (see below)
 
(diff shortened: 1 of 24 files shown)


zherczeg:
Patch is heavily reworked.

docs/02.API-REFERENCE.md
```diff
@@ -1669,9 +1668,7 @@ main (void)
 
 **Summary**
 
-Parse script and construct an EcmaScript function. The lexical environment is
-set to the global lexical environment. The resource name can be used by
-debugging systems to provide line / backtrace info.
+Parse a script, module, or function and create a compiled code using a character string.
```
galpeter, line 1671: We should add an example on how to parse function as this replaces `jerry_parse_function` method. Also the fact that this replaces the function parse method should be mentioned.

docs/02.API-REFERENCE.md
```diff
@@ -10423,7 +10368,8 @@ jerry_generate_snapshot (const jerry_char_t *source_p,
   - thrown error, otherwise.
 
 *New in version 2.0*.
-*Changed in version [[NEXT_RELEASE]]*: The `resource_name_p`, and `resource_name_length` arguments are replaced by `options_p`.
+*Changed in version [[NEXT_RELEASE]]*: The `source_p`, `source_size`, `resource_name_p`,
```
galpeter, line 10371: AFAIK there should be a newline before this. In addition the fact that a compiled/parsed JS code should be passed as first argument should also be mentioned here (as this is what changed since the previous versions).

docs/02.API-REFERENCE.md
```diff
@@ -10440,108 +10386,16 @@ main (void)
...
-**See also**
-
-- [jerry_init](#jerry_init)
-- [jerry_cleanup](#jerry_cleanup)
-- [jerry_generate_function_snapshot](#jerry_generate_function_snapshot)
-- [jerry_exec_snapshot](#jerry_exec_snapshot)
-- [jerry_parse_options_t](#jerry_parse_options_t)
-
-
-## jerry_generate_function_snapshot
```
galpeter, line 10472: We should mention that using the `jerry_generate_snapshot` method the same functionality (which was provided by this removed function) can be achieved/reproduced. Ideally an example should also be there.

zherczeg: I am not sure we need to overdo the examples.

jerry-core/parser/js/js-parser.c
```diff
@@ -1761,9 +1761,7 @@ JERRY_STATIC_ASSERT (PARSER_SCANNING_SUCCESSFUL == PARSER_HAS_LATE_LIT_INIT,
  * @return compiled code
  */
 static ecma_compiled_code_t *
-parser_parse_source (const uint8_t *arg_list_p, /**< function argument list */
-                     size_t arg_list_size, /**< size of function argument list */
-                     const uint8_t *source_p, /**< valid UTF-8 source code */
+parser_parse_source (const uint8_t *source_p, /**< valid UTF-8 source code */
```
rerobika, line 1764: I'd suggest to use void* here and in parser_parser_script as well.

rerobika (commented):
The remaining looks good to me.

rerobika (approved):

galpeter (approved):
lgtm

PR #4730, opened 2021-08-09
Simplify jerry_get_resource_name

gergocs:
JerryScript-DCO-1.0-Signed-off-by: Gergo Csizi csgergo92@gmail.com

### DIFF
--- a/jerry-core/api/jerry.c
+++ b/jerry-core/api/jerry.c
@@ -5388,26 +5388,24 @@ jerry_value_t
 jerry_get_resource_name (const jerry_value_t value) /**< jerry api value */
 {
 #if JERRY_RESOURCE_NAME
-  if (ecma_is_value_undefined (value))
+  if (ecma_is_value_undefined (value) && JERRY_CONTEXT (vm_top_context_p) != NULL)
   {
-    if (JERRY_CONTEXT (vm_top_context_p) != NULL)
-    {
-      return ecma_copy_value (ecma_get_resource_name (JERRY_CONTEXT (vm_top_context_p)->shared_p->bytecode_header_p));
-    }
+    return ecma_copy_value (ecma_get_resource_name (JERRY_CONTEXT (vm_top_context_p)->shared_p->bytecode_header_p));
   }
-  else if (ecma_is_value_object (value))
+
+  const ecma_compiled_code_t *bytecode_p = ecma_bytecode_get_from_value (value);
+
+  if (bytecode_p == NULL)
   {
-    ecma_object_t *obj_p = ecma_get_object_from_value (value);
+    return ecma_make_magic_string_value (LIT_MAGIC_STRING_RESOURCE_ANON);
+  }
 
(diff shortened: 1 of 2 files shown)


jerry-core/api/jerry.c
```diff
@@ -5383,12 +5383,9 @@ jerry_value_t
 jerry_get_resource_name (const jerry_value_t value) /**< jerry api value */
 {
 #if JERRY_RESOURCE_NAME
-  if (ecma_is_value_undefined (value))
+  if (ecma_is_value_undefined (value) && JERRY_CONTEXT (vm_top_context_p) != NULL)
```
zherczeg, line 5386: I would like to be this function as the `jerry_get_user_value` function below. It starts with:
```
  const ecma_compiled_code_t *bytecode_p = ecma_bytecode_get_from_value (value);
  if (bytecode_p == NULL)
  {
    return ECMA_VALUE_UNDEFINED;
  }
```

Then extract the string value. This will support more forms (e.g bound object, module) than the current one, you can even add some tests for the new forms.

This particular case can be kept though. Maybe this could be added to `jerry_get_user_value` for consistency (with tests)

zherczeg (changes requested):

gergocs:
Patch is ready for review

jerry-core/api/jerry.c
```diff
@@ -5385,24 +5385,22 @@ jerry_get_resource_name (const jerry_value_t value) /**< jerry api value */
 #if JERRY_RESOURCE_NAME
   if (ecma_is_value_undefined (value))
```
zherczeg, line 5386: Your `JERRY_CONTEXT (vm_top_context_p) != NULL` check was correct here.

jerry-core/api/jerry.c
```diff
@@ -5385,24 +5385,22 @@ jerry_get_resource_name (const jerry_value_t value) /**< jerry api value */
...
-    ecma_object_t *obj_p = ecma_get_object_from_value (value);
 
-    if (ecma_get_object_type (obj_p) == ECMA_OBJECT_TYPE_FUNCTION
-        && !ecma_get_object_is_builtin (obj_p))
-    {
-      ecma_extended_object_t *ext_func_p = (ecma_extended_object_t *) obj_p;
+  const ecma_compiled_code_t *bytecode_p = ecma_bytecode_get_from_value (value);
 
-      const ecma_compiled_code_t *bytecode_data_p = ecma_op_function_get_compiled_code (ext_func_p);
+  if (bytecode_p == NULL && JERRY_CONTEXT (vm_top_context_p) != NULL)
```
zherczeg, line 5393: The right side of the && does not make sense here.

zherczeg (changes requested):
Please also add test cases for newly supported types.

gergocs:
Patch is ready for review

zherczeg (changes requested):
Good patch, only the second test needs improvement.

zherczeg (approved):
LGTM

rerobika (approved):

PR #4732, opened 2021-08-13
Typo fix in 02.API-REFERENCE.md

batizdaniel:
JerryScript-DCO-1.0-Signed-off-by: Daniel Batiz batizjob@gmail.com

### DIFF
--- a/docs/02.API-REFERENCE.md
+++ b/docs/02.API-REFERENCE.md
@@ -738,8 +738,8 @@ typedef jerry_value_t (*jerry_external_handler_t) (const jerry_call_info_t *call
 
 Native free callback of generic value types.
 
-*Note*:
- - Referred values by this method must have at least 1 reference. (Correct API usage satisfies this condition)
+*Additional info*:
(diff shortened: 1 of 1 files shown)


docs/02.API-REFERENCE.md
```diff
@@ -1495,7 +1497,7 @@ someplace_in_the_code (void)
 
 Registers an external magic string array.
 
-*Notes*:
+*Note*:
```
akosthekiss, line 1500: My two cents: plural (i.e., "Notes") is quite OK when there are multiple comments.

(Of course, it depends whether "Note" is a verb or a noun here. If it stands for "Note that...", then the trailing "s" is incorrect. If it stands for the noun "Note", then my above comment holds.)

zherczeg: I think it is very easy to forget updateing note / notes after a comment is added / removed. I would prefer to use a single form.
Alternative: `Additional info:` It has no plural problem.

zherczeg (approved):
LGTM, may need an update

LaszloLango:
Cannot be applied. The master diverged a lot.

PR #4738, opened 2021-08-17
Add custom dispatcher to every builtin object for consistency

orkvi:
JerryScript-DCO-1.0-Signed-off-by: Orkenyi Virag orkvi@inf.u-szeged.hu


orkvi:
> What is the binary size effect of this change? Or runtime? In other words, what is the benefit?
The only reason is consistency. (We would like to use custom dispatcher everywhere)

jerry-core/ecma/builtin-objects/ecma-builtin-aggregateerror.c
```diff
@@ -33,6 +33,11 @@
 #define ECMA_BUILTINS_INTERNAL
 #include "ecma-builtins-internal.h"
 
+/**
+ * This object has a custom dispatch function.
+ */
+#define BUILTIN_CUSTOM_DISPATCH
```
zherczeg, line 39: If this is enabled everywhere, this should be removed, and the corresponding macros as well in a follow-up patch.

zherczeg (approved):
LGTM

PR #4742, opened 2021-08-23
Implement getter function for external string user pointer

zherczeg:


docs/02.API-REFERENCE.md
```diff
@@ -4283,6 +4283,77 @@ main (void)
...
+  - In some cases (e.g. when the string is also a magic string registered by
+    [jerry_register_magic_strings](#jerry_register_magic_strings)), the
+    string is a normal string without a user pointer even if it is created
+    by [jerry_create_external_string](#jerry_create_external_string).
+
+**Prototype**
+
+```c
+void *jerry_string_get_external_user_pointer (const jerry_value_t value,
+                                              bool *is_external);
```
galpeter, line 4307: Do we even need the `is_external` at all? As far as I understand this is only `true` if the return value is not-null.

zherczeg: NULL is a valid value for `user_p`, so that is not enough.

galpeter: And if it is a NULL value do we even care?

rerobika (approved):

galpeter (approved):
lgtm

PR #4753, opened 2021-08-27
Remove IRC notification action to have a green master

galpeter:
IRC notification GitHub action returns an error at the moment
and makes the master red (even if all builds passed).

Removing this action till we fix the notification.

### DIFF
--- a/.github/workflows/gh-actions.yml
+++ b/.github/workflows/gh-actions.yml
@@ -293,7 +293,7 @@ jobs:
 
   Notification:
     runs-on: ubuntu-latest
-    if: github.event_name == 'push' && github.repository == 'jerryscript-project/jerryscript'
+    if: false && github.event_name == 'push' && github.repository == 'jerryscript-project/jerryscript'
     steps:
       - uses: rectalogic/notify-irc@v1
         with:


.github/workflows/gh-actions.yml
```diff
@@ -291,15 +291,3 @@ jobs:
       - run: make -f ./targets/esp8266/Makefile.travis install-noapt
       - run: make -f ./targets/esp8266/Makefile.travis script
 
-  Notification:
-    runs-on: ubuntu-latest
-    if: github.event_name == 'push' && github.repository == 'jerryscript-project/jerryscript'
```
rerobika, line 296: How about?
```
if: false && github.event_name == 'push' && github.repository == 'jerryscript-project/jerryscript' # temporary disable this job
```

rerobika (changes requested):

.github/workflows/gh-actions.yml
galpeter: Yeah, maybe this will be the best. Updated to use this format.

rerobika (approved):

PR #4757, opened 2021-09-01
Fix invalid typed array construction

zherczeg:


rerobika (approved):

tests/jerry/es.next/typedArray-constructor.js
```diff
@@ -0,0 +1,35 @@
...
+// Invalid cases
+
+check_range_error("new Uint8Array(1e15)");
+check_range_error("new Uint8Array(4294967296)");
+check_range_error("new Uint8Array(4294967296 - 1)");
+check_range_error("new Uint8Array(-1)");
+
+// Valid cases
+new Uint8Array(3.5)
+new Uint8Array(-0.5)
```
galpeter, line 35: and what are the expected lengths for these? Maybe we should also have an assert for that.

zherczeg: Good idea. Patch updated.

galpeter (approved):

PR #4759, opened 2021-09-02
add missing include "jerryscript-core.h"

ChrisPei:
add missing include "jerryscript-core.h"

### DIFF
--- a/targets/esp-idf/module.c
+++ b/targets/esp-idf/module.c
@@ -21,6 +21,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include "jerryscript-core.h"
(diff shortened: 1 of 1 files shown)


targets/esp-idf/module.c
```diff
@@ -21,6 +21,7 @@
 #include <sys/stat.h>
 #include <unistd.h>
 
+#include "jerryscript-core.h"
```
zherczeg, line 24: jerryscript-core.h is internal, jerryscript.h should be used instead. Please also fix the commit message.

ChrisPei (commented):
add missing include "jerryscript.h"

rerobika:
Looks good, please squash the commits.

PR #4773, opened 2021-09-20
Rework own property enumeration

zherczeg:
Properry name filtering is done during the enumeration.

Should use less memory, since no "sorting" phase, and should be much faster in some corner cases, e.g. when the indexed properties are not needed for a large typed array / large strings.


rerobika:
Please rebase.

LaszloLango:
@zherczeg, please rebase if still needed or close if it is not.

PR #4780, opened 2021-09-27
Implement source info retrieval API function

zherczeg:


jerry-core/api/jerry.c
```diff
@@ -5513,6 +5514,185 @@ jerry_get_user_value (const jerry_value_t value) /**< jerry api value */
...
+#if JERRY_FUNCTION_TO_STRING
+  jerry_source_info_t source_info;
+
+  source_info.enabled_fields = 0;
+  source_info.source_code = ECMA_VALUE_UNDEFINED;
+  source_info.function_arguments = ECMA_VALUE_UNDEFINED;
+  source_info.source_range_start = 0;
+  source_info.source_range_length = 0;
+
+  if (!ecma_is_value_object (value))
```
galpeter, line 5536: If we do a return here, do we even need to initialize the source_info above? Maybe we could do it after this check?

jerry-core/api/jerry.c
```diff
@@ -5513,6 +5514,185 @@ jerry_get_user_value (const jerry_value_t value) /**< jerry api value */
...
+        const ecma_compiled_code_t *bytecode_p = NULL;
+
+        if (ext_object_p->u.cls.type == ECMA_OBJECT_CLASS_SCRIPT)
+        {
+          bytecode_p = ECMA_GET_INTERNAL_VALUE_POINTER (ecma_compiled_code_t,
+                                                        ext_object_p->u.cls.u3.value);
+        }
+
+#if JERRY_MODULE_SYSTEM
+        if (ext_object_p->u.cls.type == ECMA_OBJECT_CLASS_MODULE)
```
galpeter, line 5560: I think this should be an else-if.

jerry-core/api/jerry.c
```diff
@@ -5513,6 +5514,185 @@ jerry_get_user_value (const jerry_value_t value) /**< jerry api value */
...
+        {
+          ecma_module_t *module_p = (ecma_module_t *) object_p;
+
+          if (!(module_p->header.u.cls.u2.module_flags & ECMA_MODULE_IS_NATIVE))
+          {
+            bytecode_p = module_p->u.compiled_code_p;
+
```
galpeter, line 5567: No need for the empty newline.

jerry-core/api/jerry.c
```diff
@@ -5513,6 +5514,185 @@ jerry_get_user_value (const jerry_value_t value) /**< jerry api value */
...
+          }
+        }
+#endif /* JERRY_MODULE_SYSTEM */
+
+        if (bytecode_p == NULL)
+        {
+          return NULL;
+        }
+
+        ecma_value_t script_value = ((cbc_uint8_arguments_t *) bytecode_p)->script_value;
```
galpeter, line 5577: This cast relies on the fact that the script_value is in the same place (same offset) in both cbc_{uint8,uint16}_arguments_t cases. Do we have an assert for this?

docs/02.API-REFERENCE.md
```diff
@@ -11872,6 +11917,90 @@ main (void)
...
+## jerry_get_source_info
+
+**Summary**
+
+Returns a newly created source info structure corresponding to the passed script/module/function.
+The function is lower level than `toString()` operation, but provides more contextual information.
+
+*Notes*:
+- Returned value must be freed with [jerry_free_source_info](#jerry_free_source_info) when it
+is no longer needed.
```
galpeter, line 11929: It should be mentioned that this method requires the `JERRY_FUNCTION_TO_STRING` build option to be enabled otherwise it will only return NULL values.

jerry-core/api/jerry.c
```diff
@@ -5513,6 +5514,185 @@ jerry_get_user_value (const jerry_value_t value) /**< jerry api value */
...
+jerry_free_source_info (jerry_source_info_t *source_info_p) /**< source info block */
+{
+  jerry_assert_api_available ();
+
+  if (source_info_p != NULL)
+  {
+    ecma_free_value (source_info_p->source_code);
+    ecma_free_value (source_info_p->function_arguments);
+    jmem_heap_free_block (source_info_p, sizeof (jerry_source_info_t));
+  }
```
galpeter, line 5693: Should we guard this with the same macro as the `jerry_get_source_info` ?

tests/unit-core/test-source-info.c
```diff
@@ -0,0 +1,162 @@
...
+    jerry_release_value (value);
+  }
+
+  source_p = TEST_STRING_LITERAL ("( function f() {} )");
+
+  value = jerry_eval ((const jerry_char_t *) source_p, strlen (source_p), 0);
+  source_info_p = jerry_get_source_info (value);
+  TEST_ASSERT (source_info_p != NULL);
+  TEST_ASSERT (source_info_p->enabled_fields == (JERRY_SOURCE_INFO_HAS_SOURCE_CODE
+                                                 | JERRY_SOURCE_INFO_HAS_SOURCE_RANGE));
```
galpeter, line 110: Is it possible to have a case where all three "options"  are true? (has source, has source range, has arguments)

jerry-core/api/jerry.c
zherczeg: This is a common technique these dfays, and afaik there is a check for it.

tests/unit-core/test-source-info.c
zherczeg: Currently no.

zherczeg:

jerry-core/api/jerry.c
galpeter: I know it's common, and it is ok for me. I was just wandering if there is a check for it.

galpeter (approved):
lgtm

rerobika (approved):

PR #4790, opened 2021-10-05
Add notes to jerry_create_undefined

matedabis:


docs/02.API-REFERENCE.md
```diff
@@ -8176,6 +8176,11 @@ main (void)
 
 Creates a `jerry_value_t` representing an undefined value.
 
+*Notes*:
+- Calling [jerry_release_value](#jerry_release_value) is optional.
+- Calling [jerry_acquire_value](#jerry_acquire_value) is optional.
```
zherczeg, line 8181: I would write something like this:
The undefined value is a fixed constant. Its value can be copied any number of times without calling `jerry_acquire_value`, and freeing it with `jerry_release_value` is optional. 

This is also true for true/false/null values, so add this comment to them.

zherczeg (changes requested):

docs/02.API-REFERENCE.md
```diff
@@ -6870,6 +6870,9 @@ jerry_create_shared_arraybuffer_external (const jerry_length_t size
 
 Create a jerry_value_t representing a boolean value from the given boolean parameter.
 
+*Notes*:
+- The undefined value is a fixed constant. Its value can be copied any number of times without calling [jerry_acquire_value](#jerry_acquire_value), and freeing it with [jerry_release_value](#jerry_release_value) is optional.
```
zherczeg, line 6874: undefined value -> boolean values (true/false)

docs/02.API-REFERENCE.md
```diff
@@ -7238,6 +7241,9 @@ jerry_create_number_nan (void);
 
 Creates and returns a `jerry_value_t` with type null object.
 
+*Notes*:
+- The undefined value is a fixed constant. Its value can be copied any number of times without calling [jerry_acquire_value](#jerry_acquire_value), and freeing it with [jerry_release_value](#jerry_release_value) is optional.
```
zherczeg, line 7245: undefined ->null

zherczeg (changes requested):
Break these comments into two lines, they are too long.
Please also rewrite the commit message as well.

PR #4815, opened 2021-11-03
Generator declaration is not allowed in single statement position

robertsipka:


rerobika (changes requested):

rerobika (approved):

jerry-core/parser/js/js-parser-statm.c
```diff
@@ -609,7 +609,9 @@ parser_parse_function_statement (parser_context_t *context_p) /**< context */
   JERRY_ASSERT (context_p->token.type == LEXER_KEYW_FUNCTION);
 
 #if JERRY_ESNEXT
-  if (JERRY_UNLIKELY (parser_statement_flags[context_p->stack_top_uint8] & PARSER_STATM_SINGLE_STATM))
+  bool is_single_statement = (parser_statement_flags[context_p->stack_top_uint8] & PARSER_STATM_SINGLE_STATM) != 0;
+
+  if (JERRY_UNLIKELY (is_single_statement))
   {
     if (context_p->status_flags & PARSER_IS_STRICT)
```
zherczeg, line 616: Wghat about adding the `lexer_consume_generator (context_p)` check after this if? Then there is no need to create the `is_single_statement ` variable.

zherczeg (changes requested):

jerry-core/parser/js/js-parser-statm.c
robertsipka: As I noticed, we need to avoid calling the lexer_skip_spaces (inside lexer_consume_generator) function before creating the block context below (for the if statement) to avoid unexpected behavior. 
For example, if the function is not a generator function, calling lexer_consume_generator here will result an unexpected behavior:
```
if (true)
     function g () {} 
```
So to make sure we don’t call this function twice unnecessarily, I would still suggest the original solution.

zherczeg: Could you give me a bit more info about what cause the error exactly? I only see flag tests and stack checks below, nothing space related.

You can do the generator check before the end of this `if` skipping the possible issues.

jerry-core/parser/js/js-parser-util.c
```diff
@@ -1243,6 +1243,10 @@ parser_error_to_string (parser_error_t error) /**< error code */
     {
       return "Lexical declaration cannot appear in a single-statement context";
     }
+    case PARSER_ERR_GENERATOR_IN_SINGLE_STATEMENT_POS:
+    {
+      return "Generator function is not allowed in single statement position";
```
zherczeg, line 1248: position -> context
(similar error message above)

zherczeg: "Lexical declaration cannot appear in a single-statement context" this message is the same in v8, but the other looks like this: "Generators can only be declared at the top level or inside a block." (with a full stop at the end). Maybe there are other restrictions for generators.

jerry-core/parser/js/js-parser-statm.c
robertsipka: I could put it at the end of the if condition, but the function call is practically at the end right now, right after the condition. If I put it in the end of the if condition as well, it will lead to unnecessary duplication in some cases (as in the example above).

zherczeg (approved):
LGTM

PR #4816, opened 2021-11-03
Improve location info for syntax errors.

zherczeg:


zherczeg:
Patch is ready for review.

rerobika:
@zherczeg Please rebase after #4818 .

jerry-core/api/jerryscript.c
```diff
@@ -5645,6 +5645,64 @@ jerry_free_source_info (jerry_source_info_t *source_info_p) /**< source info blo
...
+
+        error_location_p->column_end = error_location_p->column_start + difference;
+      }
+
+      ecma_ref_ecma_string (ecma_get_string_from_value (result));
+      return result;
+    }
+  }
+
+  return jerry_throw (ecma_raise_type_error (ECMA_ERR_MSG ("Location is not available")));
```
dbatyai, line 5697: I'm wondering wether it would be better to simply return with undefined if there is no location, or with a pre-defined magic string. And similarly when error messages are disabled.

It feels to me like the location info is an added bonus, it can be used when it's available, but if it's not, then that's not really a problem. Or at least not something that would warrant an error.

jerry-core/api/jerryscript.c
```diff
@@ -5645,6 +5645,64 @@ jerry_free_source_info (jerry_source_info_t *source_info_p) /**< source info blo
...
 } /* jerry_free_source_info */
 
+/**
+ * Gets the resource name and location info assigned to a SyntaxError object generated by the parser.
+ *
+ * @return resource name, if a location info is available
+ *         error, otherwise
+ */
+jerry_value_t
+jerry_get_syntax_error_location (jerry_value_t value, /**< SyntaxError object */
```
dbatyai, line 5655: How about not making this syntax error specific, and changing the name to be generic for errors. Functionally it already seems to be generic. Also in the future we could possibly include location info for other error types as well.

Same for the `jerry_syntax_error_location_t` type, I would just call it `jerry_error_location_t`.

LaszloLango:
@zherczeg please rebase

PR #4822, opened 2021-11-12
Update NuttX target to releases/10.2

rtakacs:
* Updated NuttX from releases/9.0 to releases/10.2
  * Modified Makefile according to NuttX changes
* Updated README.md file
  * Added ROM File System and memory card usage descriptions as possible ways to share JavaScript input files with the device

### DIFF
--- a/targets/nuttx-stm32f4/Kconfig
+++ b/targets/nuttx-stm32f4/Kconfig
@@ -3,28 +3,18 @@
 # see the file kconfig-language.txt in the NuttX tools repository.
 #
 
-config JERRYSCRIPT
-	bool "Jerryscript"
+config INTERPRETERS_JERRYSCRIPT
+	bool "Jerryscript JavaScript engine"
 	default n
-	---help---
-		Enable Jerryscript ECMAScript 5.1 interpreter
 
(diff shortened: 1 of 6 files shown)


robertsipka (approved):

robertsipka (approved):
LGTM, I've tried it, works like a charm.
Minor suggestions: More dependencies are already needed for the toolchain and the kconfig generator, so I would move up the following description: 
`https://github.com/apache/incubator-nuttx/blob/releases/10.2/Documentation/quickstart/install.rst`
stlink also has dependencies that are worth mentioning to install:
`gtk+-3.0 libusb-1.0-0-dev minicom`

rtakacs:
@robertsipka I've added an `install-deps.sh` file that contains all the packages that are required by projects (NuttX, STLink) and serial communication.

robertsipka:
even better, still looks good to me.

rtakacs:
@robertsipka FYI, I've extended the `Makefile` with JerryScript compilation. In this case, `NuttX` automatically builds JerryScript as its dependency. (And that is not a different step for the user.)

dbatyai (approved):

PR #4825, opened 2021-11-15
Upgrade jerry-debugger client to use python3

repasics:
Also upgrade pylint to use python3.

### DIFF
--- a/jerry-debugger/jerry_client.py
+++ b/jerry-debugger/jerry_client.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright JS Foundation and other contributors, http://js.foundation
 #
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import print_function
 from cmd import Cmd
 from pprint import pprint
 import math
@@ -104,7 +103,7 @@ class DebuggerPrompt(Cmd):
                     if res_type == result.END:
                         self.quit = True
                         return
-                    elif res_type == result.TEXT:
+                    if res_type == result.TEXT:
                         write(result.get_text())
                     elif res_type == result.PROMPT:
                         break
@@ -325,7 +324,7 @@ def main():
 
         if res_type == result.END:
             break
-        elif res_type == result.PROMPT:
+        if res_type == result.PROMPT:
             prompt.cmdloop()
         elif res_type == result.TEXT:
             write(result.get_text())
@@ -339,7 +338,7 @@ if __name__ == "__main__":
         MSG = str(error_msg)
         if ERRNO == 111:
             sys.exit("Failed to connect to the JerryScript debugger.")
-        elif ERRNO == 32 or ERRNO == 104:
+        elif ERRNO in (32, 104):
             sys.exit("Connection closed.")
         else:
             sys.exit("Failed to connect to the JerryScript debugger.\nError: %s" % (MSG))
--- a/jerry-debugger/jerry_client_main.py
+++ b/jerry-debugger/jerry_client_main.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 
 # Copyright JS Foundation and other contributors, http://js.foundation
 #
@@ -14,7 +14,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from __future__ import print_function
 import argparse
 import logging
 import re
@@ -149,7 +148,12 @@ def arguments_parse():
     return args
 
 
-class JerryBreakpoint(object):
+def safe_ord(data):
+    if isinstance(data, int):
+        return data
+    return ord(data)
+
+class JerryBreakpoint():
 
     def __init__(self, line, offset, function):
         self.line = line
@@ -171,7 +175,7 @@ class JerryBreakpoint(object):
         return ("Breakpoint(line:%d, offset:%d, active_index:%d)"
                 % (self.line, self.offset, self.active_index))
 
-class JerryPendingBreakpoint(object):
+class JerryPendingBreakpoint():
     def __init__(self, line=None, source_name=None, function=None):
         self.function = function
         self.line = line
@@ -188,14 +192,14 @@ class JerryPendingBreakpoint(object):
         return result
 
 
-class JerryFunction(object):
+class JerryFunction():
     # pylint: disable=too-many-instance-attributes,too-many-arguments
     def __init__(self, is_func, byte_code_cp, source, source_name, line, column, name, lines, offsets):
         self.is_func = bool(is_func)
         self.byte_code_cp = byte_code_cp
-        self.source = re.split("\r\n|[\r\n]", source)
-        self.source_name = source_name
-        self.name = name
+        self.source = re.split("\r\n|[\r\n]", source.decode())
+        self.source_name = source_name.decode()
+        self.name = name.decode()
         self.lines = {}
         self.offsets = {}
         self.line = line
@@ -208,20 +212,20 @@ class JerryFunction(object):
 
         for i, _line in enumerate(lines):
             offset = offsets[i]
-            breakpoint = JerryBreakpoint(_line, offset, self)
-            self.lines[_line] = breakpoint
-            self.offsets[offset] = breakpoint
+            br_point = JerryBreakpoint(_line, offset, self)
+            self.lines[_line] = br_point
+            self.offsets[offset] = br_point
 
     def __repr__(self):
         result = ("Function(byte_code_cp:0x%x, source_name:%r, name:%r, line:%d, column:%d { "
                   % (self.byte_code_cp, self.source_name, self.name, self.line, self.column))
 
-        result += ','.join([str(breakpoint) for breakpoint in self.lines.values()])
+        result += ','.join([str(br_point) for br_point in self.lines.values()])
 
         return result + " })"
 
 
-class Multimap(object):
+class Multimap():
 
     def __init__(self):
         self.map = {}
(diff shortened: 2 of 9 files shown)


jerry-debugger/jerry_client.py
```diff
@@ -104,7 +104,7 @@ def do_next(self, args):
                     if res_type == result.END:
                         self.quit = True
                         return
-                    elif res_type == result.TEXT:
+                    if res_type == result.TEXT:
```
galpeter, line 107: Why do we need to change `elif` to  `if`? This applies to most of the cases.

jerry-debugger/jerry_client.py
```diff
@@ -279,7 +279,6 @@ def main():
 
         protocol = Socket(address)
     elif args.protocol == "serial":
-        from jerry_client_serial import Serial
```
galpeter, line 282: Do we really need to move this out of here? If so, we always expect that a user will install the serial package even if they will only use the socket based communication.

jerry-debugger/jerry_client_tcp.py
```diff
@@ -29,7 +29,7 @@ def connect(self):
         Connect to a remote socket at address (host, port).
         The format of address depends on the address family.
         """
-        print("Connecting to: %s:%s" % (self.address[0], self.address[1]))
+        print(("Connecting to: %s:%s" % (self.address[0], self.address[1])))
```
galpeter, line 32: Why do we need the extra parentheses?

tools/runners/test262-harness.py
```diff
@@ -405,6 +405,7 @@ def write(self, string):
         os.write(self.file_desc, string)
 
     def read(self):
+        # pylint: disable=undefined-variable
```
galpeter, line 408: Which variable is undefined here?

jerry-debugger/jerry_client_main.py
```diff
@@ -208,20 +212,20 @@ def __init__(self, is_func, byte_code_cp, source, source_name, line, column, nam
 
         for i, _line in enumerate(lines):
             offset = offsets[i]
-            breakpoint = JerryBreakpoint(_line, offset, self)
-            self.lines[_line] = breakpoint
-            self.offsets[offset] = breakpoint
+            br_point = JerryBreakpoint(_line, offset, self)
```
galpeter, line 215: Why do we need to rename the "breakpoint" variable? This also applies to other places.

jerry-debugger/jerry_client_main.py
```diff
@@ -208,20 +212,20 @@ def __init__(self, is_func, byte_code_cp, source, source_name, line, column, nam
...
+            self.lines[_line] = br_point
+            self.offsets[offset] = br_point
 
     def __repr__(self):
         result = ("Function(byte_code_cp:0x%x, source_name:%r, name:%r, line:%d, column:%d { "
                   % (self.byte_code_cp, self.source_name, self.name, self.line, self.column))
 
-        result += ','.join([str(breakpoint) for breakpoint in self.lines.values()])
+        result += ','.join([str(br_point) for br_point in list(self.lines.values())])
```
galpeter, line 223: Why do we need the `list` call here? the `.values()` should already be an iterable object.

jerry-debugger/jerry_client_main.py
```diff
@@ -381,12 +385,12 @@ def breakpoint_list(self):
         result = ''
         if self.active_breakpoint_list:
             result += "=== %sActive breakpoints %s ===\n" % (self.green_bg, self.nocolor)
-            for breakpoint in self.active_breakpoint_list.values():
-                result += " %d: %s\n" % (breakpoint.active_index, breakpoint)
+            for br_point in list(self.active_breakpoint_list.values()):
```
galpeter, line 388: Same as above:
* Why the variable rename
* Why the `list` call?

jerry-debugger/jerry_client_main.py
```diff
@@ -381,12 +385,12 @@ def breakpoint_list(self):
...
+                result += " %d: %s\n" % (br_point.active_index, br_point)
         if self.pending_breakpoint_list:
             result += "=== %sPending breakpoints%s ===\n" % (self.yellow_bg, self.nocolor)
-            for breakpoint in self.pending_breakpoint_list.values():
-                result += " %d: %s (pending)\n" % (breakpoint.index, breakpoint)
+            for br_point in list(self.pending_breakpoint_list.values()):
```
galpeter, line 392: Same as above.

jerry-debugger/jerry_client_main.py
```diff
@@ -398,13 +402,13 @@ def delete(self, args):
             return "Error: Breakpoint index expected\n" \
                    "Delete the given breakpoint, use 'delete all|active|pending' " \
                    "to clear all the given breakpoints\n "
-        elif args in ['all', 'pending', 'active']:
```
galpeter, line 401: Why change the "elif" to "if" ?

jerry-debugger/jerry_client_main.py
```diff
@@ -418,20 +422,19 @@ def delete(self, args):
...
             del self.pending_breakpoint_list[breakpoint_index]
             if not self.pending_breakpoint_list:
                 self._send_parser_config(0)
             return "Pending breakpoint %d deleted\n" % (breakpoint_index)
-        else:
-            return "Error: Breakpoint %d not found\n" % (breakpoint_index)
+        return "Error: Breakpoint %d not found\n" % (breakpoint_index)
 
-    def next(self):
+    def __next__(self):
```
galpeter, line 437: This should not be changed. The `__next__` would mean that this is an iterator/iterable, but this is not an iterator but a command sent via the debugger connection.

jerry-debugger/jerry_client_main.py
```diff
@@ -888,7 +891,7 @@ def _parse_source(self, data):
                 logging.error("Syntax error found")
                 return ""
 
-            elif buffer_type in [JERRY_DEBUGGER_SOURCE_CODE, JERRY_DEBUGGER_SOURCE_CODE_END]:
+            if buffer_type in [JERRY_DEBUGGER_SOURCE_CODE, JERRY_DEBUGGER_SOURCE_CODE_END]:
```
galpeter, line 894: Why change `elif` to `if`?

jerry-debugger/jerry_client_main.py
```diff
@@ -980,34 +983,34 @@ def _parse_source(self, data):
         # Copy the ready list to the global storage.
         self.function_list.update(new_function_list)
 
-        for function in new_function_list.values():
-            for line, breakpoint in function.lines.items():
-                self.line_list.insert(line, breakpoint)
+        for function in list(new_function_list.values()):
+            for line, br_point in list(function.lines.items()):
```
galpeter, line 987: Why do we need the list calls?

jerry-debugger/jerry_client_main.py
```diff
@@ -980,34 +983,34 @@ def _parse_source(self, data):
...
-                for src in new_function_list.values():
-                    if (src.source_name == breakpoint.source_name or
-                            src.source_name.endswith("/" + breakpoint.source_name) or
-                            src.source_name.endswith("\\" + breakpoint.source_name)):
+                for src in list(new_function_list.values()):
+                    if (src.source_name == br_point.source_name or
+                            src.source_name.endswith("/" + br_point.source_name) or
+                            src.source_name.endswith("\\" + br_point.source_name)):
                         source_lines = len(src.source)
                         break
```
galpeter, line 1002: Why do we need the `list` call here?

galpeter (changes requested):
General question: do we want to remove full python2 support?

tools/runners/test262-harness.py
ossy-szeged: pylint with python3 thinks that file is undefined variable. But test262-harness.py is still python2 script. (will be changed soon)

galpeter: Isn't this code just reading the file contents and returning? If yes the solution is simple: use "open" as a context manager read the data and return the contents.

ossy-szeged: Yes. but it is completely unrelated to this PR.

jerry-debugger/jerry_client.py
repasics: Pylint does not accepts elif here becasue of return in previous if statement:

************* Module jerry_client
jerry-debugger/jerry_client.py:104:20: R1705: Unnecessary "elif" after "return" (no-else-return)

jerry-debugger/jerry_client.py
repasics: Moved import to top level because of pylint.

************* Module jerry_client
jerry-debugger/jerry_client.py:281:8: C0415: Import outside toplevel (jerry_client_serial) (import-outside-toplevel)

repasics: I can move back the import there and disable pylint on that row.

jerry-debugger/jerry_client_main.py
repasics: Rename is needed because breakopoint is a built-in.

************* Module jerry_client_main
jerry-debugger/jerry_client_main.py:215:12: W0622: Redefining built-in 'breakpoint' (redefined-builtin)

repasics:
Updated the PR. 2to3 app changed the parenthesis, next function and list call.
I reverted the parenthesis- and 'next' function change, but had to keep 2 of list calls to avoid runtime errors.

jerry-debugger/jerry_client.py
repasics: This was only error locally so I reverted this.

jerry-debugger/jerry_client_main.py
rerobika: this question is still open.

jerry-debugger/jerry_client_main.py
rerobika: There is a return above.

jerry-debugger/jerry_client_main.py
rerobika: ditto.

lygstate:
this can be closed, python3 is already supported?

robertsipka:
I've just closed this PR due to python3 is already supported.

PR #4828, opened 2021-11-17
Restore esp-idf port

jacquesg:
Restore ESP-IDF port. This also cleans up all the files, moving them into a single file for the port itself and module support respectively.


robertsipka (approved):
LGTM

dbatyai (approved):

rerobika:
@jacquesg If you have any intentiton to make a CI build job for this target we would really appreciate it.

jacquesg:
> @jacquesg If you have any intentiton to make a CI build job for this target we would really appreciate it.

I may very well do that in the not too distant future, I only realised after the fact that one exists only for the ESP8266

PR #4858, opened 2021-12-07
Fixes the code before update test262 git hash, and update exclude list

lygstate:
Added minor fixes to be able to update the test262 revision without failing on the CI
- Fixed TypedArray related functions where the ECMA_ARRAYBUFFER_CHECK_BUFFER_ERROR macro was used, because it was incorrect.
  According to the ECMAScript standard it is not always a TypeError is the underlying arraybuffer is detached
  (see also ECMAScript 2022: 10.4.5.6 [[Delete]] 1.b.i IsValidIntegerIndex)

### DIFF
--- a/jerry-core/api/jerryscript.c
+++ b/jerry-core/api/jerryscript.c
@@ -35,6 +35,7 @@
 #include "ecma-extended-info.h"
 #include "ecma-function-object.h"
 #include "ecma-gc.h"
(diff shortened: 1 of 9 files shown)


rerobika:
Please, let the community decide whether this commit should be exlcuded or not from #4858. Moreover if the exlcude is necessary, let the author do this job and earn the credits for it.

PR #4829, opened 2021-11-19
Rework the public API

dbatyai:
Related to #4186.

Some notable changes:
  - The term 'Error' now strictly refers to native Error objects defined in
    the ECMA standard, which are ordinary objects. All other uses of
    'error' or 'error reference' where the term refers to a thrown value is
    now called 'exception'.

  - Simplified the naming scheme of many String API functions. These functions
    will now also take an 'encoding' argument to specify the desired
    encoding in which to operate.

  - Removed the substring-copy-to-buffer functions. These functions
    behaved awkwardly, as they use character index to specify the
    start/end positions, and were mostly used incorrectly with byte
    offsets instead. The functionality can still be replicated with
    other functions if necessary.

  - String-to-buffer functions will no longer fail if the buffer is not
    sufficiently large, the string will instead be cropped.

  - Fixed the usage of the '_sz' prefix in many API functions. The term
    'sz' means zero-terminated string in hungarian notation, this was
    used incorrectly in many cases.

  - Renamed most of the public API functions to have shorter, more on-point
    names, rather than the often too long descriptive names. Functions are now
    also grouped by the type of value they operate on, where this makes
    sense.

PR #4860, opened 2021-12-07
Rename jerry_port_print_char to jerry_port_string_print for printing UTF8 directly

lygstate:
Use name jerry_port_string_print  to consistence with jerry_string_print.
Add API jerry_string_iterate_code_point and jerry_code_point_print to be called by jerry_port_string_print

Followup of https://github.com/jerryscript-project/jerryscript/pull/4829

### DIFF
--- a/docs/05.PORT-API.md
+++ b/docs/05.PORT-API.md
@@ -72,15 +72,18 @@ typedef enum
 void jerry_port_log (jerry_log_level_t level, const char *fmt, ...);
 ```
 
-The `jerry_port_print_char` is currently not used by the jerry-core directly.
+The `jerry_port_string_print` is currently not used by the jerry-core directly.
 However, it provides a port specific way for `jerry-ext` components to print
 information.
 
 ```c
 /**
- * Print a character to stdout.
+ * Print a zero-terminated UTF-8 string with length to standard output.
+ *
+ * @param s The zero-terminated UTF-8 string to print
+ * @param len The length of the UTF-8 string.
  */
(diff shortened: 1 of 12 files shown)


rerobika (changes requested):
If i get it right there was no agreement in #4829 about dropping `jerry_port_print_char`, rather than it should be kept but we can introduce `jerry_port_string_print` for convenience.
However @dbatyai is currently working on this topic, so be patient.

lygstate:
> If i get it right there was no agreement in #4829 about dropping `jerry_port_print_char`, rather than it should be kept but we can introduce `jerry_port_string_print` for convenience. However @dbatyai is currently working on this topic, so be patient.

hi we have jerry_code_point_print do the same thing,  anyway, I am looking on it

dbatyai:
As I've said, and as @rerobika pointed out, this is not the right approach, jerry_port_print_char should not be removed, and jerry_code_point_print is not the same. I also don't see any reason to introduce the concept of code points to the api. 

Also, as I've said before, I'm making a lot of changes to the entire port infrastructure, including how i/o is handled in jerry-ext and jerry-main, among many other things, and the changes present here are already incompatible. I do not wan't to merge anything that I would have to rewrite entirely in the next api revision patch, there's no reason for it.

Also, the debugger and port api interaction needs more tought to it than just slamming in new api functions.

I will introduce a port function for printing strings in my next patch soon, please be patient.

lygstate:
@dbatyai 
That's sounds very good, I am looking for it.

PR #4849, opened 2021-12-07
Prevent arguments object creation if 'arguments' function argument is present

rerobika:
This patch fixes #4847.

JerryScript-DCO-1.0-Signed-off-by: Robert Fancsik robert.fancsik@h-lab.eu

PR #4867, opened 2021-12-08
Fix arguments object detection in non-complex param list

rerobika:
This patch is the followup of #4849.

### DIFF
--- a/tests/jerry/es.next/arguments.js
+++ b/tests/jerry/es.next/arguments.js
@@ -217,3 +217,8 @@ function f22 (arguments, [a = arguments]) {
   assert(arguments === 3.1);
 }
 f22(3.1, []);
+
+function f23(arguments, eval = () => eval()) {
+  assert(arguments === undefined);
+}
+f23(undefined);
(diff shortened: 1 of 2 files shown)


tests/jerry/es.next/arguments.js
```diff
@@ -217,3 +217,8 @@ function f22 (arguments, [a = arguments]) {
   assert(arguments === 3.1);
 }
 f22(3.1, []);
+
+function f23(arguments, eval = () => eval()) {
+  assert(arguments === 3.1);
```
zherczeg, line 222: I would prefer an undefined comparison to prove that the arguments object has never been created (and not overwritten later by the number argument)

zherczeg (changes requested):
This fix is better.

tests/jerry/es.next/arguments.js
rerobika: updated.

zherczeg (approved):
LGTM

ossy-szeged (approved):

PR #4868, opened 2021-12-08
Adjust fast array growth

rerobika:
Since we allow lagre buffer allocations during array constructor we could extend this behavior to [[Set]] operation as well.

### DIFF
--- a/jerry-core/ecma/operations/ecma-array-object.c
+++ b/jerry-core/ecma/operations/ecma-array-object.c
@@ -35,22 +35,6 @@
  * @{
  */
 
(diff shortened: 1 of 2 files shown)


rerobika:
Test:

```js
for (let i = 0; i < 1000; i++)
{
  var a = [];
  a.length = 10000;
  a.fill(7, 0);
}
```

Before: 0.96s
After: 0.1s

dbatyai (approved):

szilagyiadam (approved):
LGTM (informal)

ossy-szeged (approved):

PR #4881, opened 2021-12-09
Use external string when raising error

repasics:

### DIFF
--- a/jerry-core/ecma/operations/ecma-exceptions.c
+++ b/jerry-core/ecma/operations/ecma-exceptions.c
@@ -300,14 +300,14 @@ ecma_get_error_type (ecma_object_t *error_object_p) /**< possible error object *
  */
 ecma_value_t
 ecma_raise_standard_error (jerry_error_t error_type, /**< error type */
-                           const char *msg_p) /**< error message */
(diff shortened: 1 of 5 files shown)


rerobika (changes requested):

jerry-core/ecma/operations/ecma-exceptions.c
```diff
@@ -295,6 +295,8 @@ ecma_get_error_type (ecma_object_t *error_object_p) /**< possible error object *
 /**
  * Raise a standard ecma-error with the given type and message.
  *
+ * Note: The given buffer must be CESU8 encoded and the buffer must exist until the error message is not freed.
+ *
```
dbatyai, line 299: The `char *` message argument should be replaced with `ecma_error_msg_t`, and this function should call `ecma_get_error_msg`. This guarantees that the string is a magic string and it exists statically, and then there's no reason for the note. Also the string size can then be retrieved, without the need for calling `lit_zt_utf8_string_size`.

dbatyai (changes requested):

rerobika (approved):
still LGTM

dbatyai (approved):

PR #4903, opened 2021-12-13
Implement String.prototype.normalize

rerobika:


jerry-core/ecma/base/ecma-helpers-string.c
```diff
@@ -2805,6 +2805,58 @@ ecma_op_advance_string_index (ecma_string_t *str_p, /**< input string */
 } /* ecma_op_advance_string_index */
 #endif /* JERRY_ESNEXT */
 
+#if JERRY_ICU
```
dbatyai, line 2808: Should have a default value in `jerry-core/config.h`

dbatyai (changes requested):

lygstate:
I suggest move icu out of core by providing a port api.

PR #4912, opened 2021-12-22
Fix runtime error: left shift

### LINKED ISSUE #4703: ubsan: jerryscript/jerry-core/vm/vm.c:3731:77: runtime error: left shift of 33554431 by 22 places cannot be represented in type 'int'
###### JerryScript revision

ccc1525
###### Build platform

###### Build steps

```sh
readonly FLAGS=(
)
make
make install
```

###### Test case

```javascript
33554431 << 22
```

###### Output

```
jerryscript/jerry-core/vm/vm.c:3731:77: runtime error: left shift of 33554431 by 22 places cannot be represented in type 'int'
```

###### Expected behavior

According to C99 standard (6.5.7/4):
> The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated bits are filled with
zeros. If E1 has an unsigned type, the value of the result is E1 × 2 E2 , reduced modulo
one more than the maximum value representable in the result type. If E1 has a signed
type and nonnegative value, and E1 × 2 E2 is representable in the result type, then that is
the resulting value; otherwise, the behavior is undefined.

batizdaniel:
This patch fixes #4703


dbatyai:
I'm not a big fan of doing a clz call for every left shift operation, even if its handled by a builtin. It can have a significant performance overhead.

According to the C99 standard when the left value has an unsigned type the result will always be defined.

> If E1 has an unsigned type, the value of the result is E1 × 2^E2, reduced modulo
one more than the maximum value representable in the result type

If the goal is to avoid undefined behavior, we could just cast the left operand to unsigned, and the result would be the same for non-negative values.

For negative values it's not defined what the result should be, neither the C99, not the ecma standard provide anything specific. Looking at implementations however it seems that the same `E1 × 2^E2` rule is used for negative values as well.

So if the left value is negative, we could mask the low 31 bits, do an unsigned left shift to avoid undefined behavior, and if the result is non-zero, set the sign bit on the result.

IMO this approach would yield the same results as previously, avoids undefined behavior, and should also be much faster.

batizdaniel:
> 
I was not sure, what is the correct way to fix this. Your approach is much better. Fixed.

jerry-core/ecma/base/ecma-helpers-number.c
```diff
@@ -414,7 +414,8 @@ ecma_integer_multiply (ecma_integer_value_t left_integer, /**< left operand */
   if (JERRY_UNLIKELY ((left_integer & (left_integer - 1)) == 0))
   {
     /* Right shift right_integer with log2 (left_integer) */
-    return ecma_make_integer_value (right_integer << (__builtin_ctz ((unsigned int) left_integer)));
+    return ecma_make_integer_value (
+      (int32_t) ((uint32_t) right_integer << (__builtin_ctz ((unsigned int) left_integer))));
```
dbatyai, line 418: The shift in the condition after this will need a cast as well.

dbatyai (changes requested):

dbatyai (approved):

rerobika (approved):

PR #4914, opened 2021-12-23
Fix runtime error Date object

### LINKED ISSUE #4704: ubsan: jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c:82:27: runtime error: signed integer overflow: 19998030 * 365 cannot be represented in type 'int'
###### JerryScript revision

ccc1525
###### Build platform

###### Build steps

```sh
readonly FLAGS=(
)
make
make install
```

###### Test case

```javascript
new Date(20000000, 0)
```

###### Output

```
jerryscript/jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c:82:27: runtime error: signed integer overflow: 19998030 * 365 cannot be represented in type 'int'
```

batizdaniel:
This patch fixes #4704


rerobika (approved):

dbatyai (commented):
This doesn't really fix the issue, `new Date(-20000000, 0)` will trigger the same behavior on the other branch.
On top of that, even though it no longer triggers an undefined behavior, the value can still overflow, for example `new Date(200000000, 0)` (one extra zero), which actually results in a valid date (incorrectly).

I think the best approach would be to clamp the input year so that it's not abnormally large. The valid year range is [-271821, 275760], so clamping to a slightly larger range than that should be enough, making sure that invalid year values remain invalid.

ossy-szeged:
> I think the best approach would be to clamp the input year so that it's not abnormally large. The valid year range is [-271821, 275760], so clamping to a slightly larger range than that should be enough, making sure that invalid year values remain invalid.

It's not as easy. We can't clamp the input year, because there is no restriction for the arguments of the Date constructor, only the output Date object should be in the allowed interval. for example new Date(-271822, 16) is valid Date, but new Date(-271822, 15) is invalid.

dbatyai:
That's why I said clamping it to a slightly larger range. That will filter out abnormally large input values, that should be invalid regardless, then the rest will be handled by the conversion logic correctly without overflowing.

ossy-szeged:
https://262.ecma-international.org/12.0/#eqn-DaysFromYear

The latest spec clarifies that we should handle year as mathematical value (Arbitrary real numbers, used as the default numeric type) and the result should be Number (double):
DayFromYear(y) = 𝔽(365 × (ℝ(y) - 1970) + floor((ℝ(y) - 1969) / 4) - floor((ℝ(y) - 1901) / 100) + floor((ℝ(y) - 1601) / 400))

jerry-core/ecma/builtin-objects/ecma-builtin-helpers-date.c
```diff
@@ -420,19 +422,23 @@ ecma_date_make_day (ecma_number_t year, /**< year value */
...
 
   /* 2., 3., 4. */
-  int32_t y = (int32_t) (year);
   ecma_number_t m = ecma_number_trunc (month);
   ecma_number_t dt = ecma_number_trunc (date);
 
   /* 5. */
-  int32_t ym = y + (int32_t) (floor (m / 12));
+  year += (floor (m / 12));
+  if (fabs (year) > 275800)
```
rerobika, line 436: Where does this magic number come from? Please add a macro with a description for it.

rerobika (dismissed):

LaszloLango (approved):

ossy-szeged (approved):

PR #4957, opened 2022-01-06
Fix runtime error: left shift of negative value

### LINKED ISSUE #4701: ubsan: jerryscript/jerry-core/vm/opcodes-ecma-bitwise.c:96:70: runtime error: left shift of negative value -1
###### JerryScript revision

ccc1525
###### Build platform

###### Build steps

```sh
readonly FLAGS=(
)
make
make install
```

###### Test case

```javascript
function f() {
    var x = (("-1.234" << 0));
    var y = 3.14;
    assertEq(true && x < y, true);
}
f();
```

###### Output

```
jerryscript/jerry-core/vm/opcodes-ecma-bitwise.c:96:70: runtime error: left shift of negative value -1
```

###### Expected behavior

According to C99 standard (6.5.7/4):
> The result of E1 << E2 is E1 left-shifted E2 bit positions; vacated bits are filled with
zeros. If E1 has an unsigned type, the value of the result is E1 × 2 E2 , reduced modulo
one more than the maximum value representable in the result type. If E1 has a signed
type and nonnegative value, and E1 × 2 E2 is representable in the result type, then that is
the resulting value; otherwise, the behavior is undefined.

batizdaniel:
This patch fixes #4701


ossy-szeged (approved):
https://262.ecma-international.org/12.0/#sec-numeric-types-number-leftShift
The latest spec defines this operation: "Return the result of left shifting lnum by shiftCount bits. The mathematical value of the result is exactly representable as a 32-bit two's complement bit string."

With two's complement, the signed and unsigned left shift is exactly the same operation. To avoid the undefined behavior I think it is the right fix is doing unsigned left shift here and then interpreting the results as signed int. The PR exactly does it, so LGTM.

dbatyai (approved):

PR #4977, opened 2022-02-09
Improve the jerry port api documents.

lygstate:
Notable changes:
  - Remove the comments in port impl, that's easily getting to in-consistence
  - Sync the jerryscript-port.h and 05.PORT-API.md
  - Fixes the invalid comment in port codes

### DIFF
--- a/jerry-port/common/jerry-port-context.c
+++ b/jerry-port/common/jerry-port-context.c
@@ -27,13 +27,6 @@
  */
 static jerry_context_t *current_context_p = NULL;
 
-/**
- * Allocate a new external context.
- *
- * @param context_size: requested context size
- *
- * @return total allcoated size
- */
 size_t JERRY_ATTR_WEAK
 jerry_port_context_alloc (size_t context_size)
 {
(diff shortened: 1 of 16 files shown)


lygstate:
@LaszloLango Any suggestion about this?

LaszloLango (approved):
LGTM

jerry-port/common/jerry-port-context.c
```diff
@@ -27,13 +27,6 @@
  */
 static jerry_context_t *current_context_p = NULL;
 
-/**
```
zherczeg, line 30: Why these are removed?

zherczeg (commented):
A lot of comments are removed. Why?

jerry-port/common/jerry-port-context.c
lygstate: Remove the comments in port impl, that's easily getting to in-consistence

Those already documented in header, document twice create a lot of inconsistence, and not updated. So just leave a single copy of the comment

LaszloLango: I agree with @lygstate, these comments should be place in the header file only.

robertsipka (approved):
LGTM

PR #5032, opened 2023-01-03
Fix failing CI jobs

ossy-szeged:

PR #5033, opened 2023-01-03
Ensure command line prompt is displayed

rtakacs:
As #5030 pointed out, `js > ` prompt hasn't been shown in the terminal. This patch solves that by flushing `stdout` manually.


rtakacs:
Note: failing jobs are fixed by #5032

ossy-szeged (approved):
LGTM

PR #5047, opened 2023-03-09
fix building without BIGINT

aksdfauytv:
someone forgot ifdef for bigint

JerryScript-DCO-1.0-Signed-off-by: Maciej Musiał xt1@o2.pl

### DIFF
--- a/jerry-core/ecma/operations/ecma-typedarray-object.c
+++ b/jerry-core/ecma/operations/ecma-typedarray-object.c
@@ -668,6 +668,7 @@ ecma_set_typedarray_element (ecma_typedarray_info_t *info_p, /**< typedarray inf
                              uint32_t index) /**< element index */
 {
   ecma_value_t to_num;
+#if JERRY_BUILTIN_BIGINT
   if (ECMA_TYPEDARRAY_IS_BIGINT_TYPE (info_p->id))
   {
     to_num = ecma_bigint_to_bigint (value, false);
@@ -678,6 +679,7 @@ ecma_set_typedarray_element (ecma_typedarray_info_t *info_p, /**< typedarray inf
     }
   }
   else
+#endif /* JERRY_BUILTIN_BIGINT */
   {
     ecma_number_t result_num;
     to_num = ecma_op_to_numeric (value, &result_num, ECMA_TO_NUMERIC_NO_OPTS);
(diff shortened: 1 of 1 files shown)


robertsipka (changes requested):
Thanks, the fix is ​​necessary but currently it is just a half-solution. Please guard the bigint related usages  in `ecma_set_typedarray_element` method as well.

aksdfauytv:
> Thanks, the fix is ​​necessary but currently it is just a half-solution. Please guard the bigint related usages in `ecma_set_typedarray_element` method as well.

Can you elaborate?

robertsipka:
> > Thanks, the fix is ​​necessary but currently it is just a half-solution. Please guard the bigint related usages in `ecma_set_typedarray_element` method as well.
> 
> Can you elaborate?

As I saw, the `ecma_bigint_to_bigint` call is not guarded inside the mentioned function neither, but it should be. You can receive a more detailed error message if you run the build without the bigint builtin.

aksdfauytv:
Obviously, that's why I created this pull request, it adds the guard. But you flagged the pull request as "Changes requested", so that's what I'm asking you about, what additional changes do you need?

ossy-szeged:
There is at least one more build fail related to ecma_bigint_to_bigint:

jerryscript/jerry-core/ecma/operations/ecma-atomics-object.c:171:11: error: implicit declaration of function ‘ecma_bigint_to_bigint’; did you mean ‘jerry_bigint_to_digits’? [-Werror=implicit-function-declaration]
  171 |     val = ecma_bigint_to_bigint (value, true);
      |           ^~~~~~~~~~~~~~~~~~~~~

robertsipka:
Although the additional bug is still present in your branch, within that line that @ossy-szeged marked, please rebase it onto master, and you will see which function contains the issue and what I mentioned to you and which where you also needs to add guards if the mentioned builtin is not enabled. You just need to follow the instructions you wrote in your PR, meaning build the project with the bigint builtin disabled, you will see what is required and why i requested changes...

aksdfauytv:
I see, I didn't notice the issue in ecma-atomics-object.c, because I was building with atomics disabled. Thank you for pointing it out.

ossy-szeged (approved):

robertsipka (approved):

LaszloLango:
After #5166 this is fixes the build, no need for changes.

LaszloLango (approved):

ossy-szeged:
Let's merge it as is, atomics part is fixed in #5166

PR #5048, opened 2023-03-09
fix building with LINE_INFO

aksdfauytv:
config.h has to be included first

JerryScript-DCO-1.0-Signed-off-by: Maciej Musiał xt1@o2.pl

### DIFF
--- a/jerry-core/ecma/base/ecma-line-info.h
+++ b/jerry-core/ecma/base/ecma-line-info.h
@@ -23,9 +23,10 @@
  * @{
  */
 
+#include "ecma-globals.h"
+
 #if JERRY_LINE_INFO
 
-#include "ecma-globals.h"
 
 /**
  * Increase the current value of line or column.


akosthekiss:
Please do explain why it would be necessary to hoist the include out of the `JERRY_LINE_INFO` guard. In which configuration does it cause any problems?

aksdfauytv:
When you modify config.h so that JERRY_LINE_INFO is defined as 1 by default, the project no longer compiles.

akosthekiss:
I am pretty sure that the root cause for that is not in this file. It may be true that the `ecma-globals.h` header is needed somewhere where it is not included, but moving the include in `jerry-core/ecma/base/ecma-line-info.h` out of the guard is certainly a quick'n'dirty fix only.

If there was an issue report for the problem (with properly filled in details) then it were easier to reproduce the build fail and better see which files do need to be fixed actually.

aksdfauytv:
It is generally considered good practice to make your includes independent of who is including them. It is not the case with ecma-line-info.h:
ecma-line-info.h references JERRY_LINE_INFO before including "ecma-globals.h", which defines it. This makes whoever includes ecma-line-info.h responsible for including "ecma-globals.h" first. However, "emac-line-info.c", "js-parser-line-info-create.c" and others don't do that, and therefore they don't compile (all definitions from ecma-line-info.h are missing).

The project only compiles as is, if you define JERRY_LINE_INFO as a global compiler switch with command line argument -D.

PR #5095, opened 2023-09-13
Update tests and CI to python3

matetokodi:
This patch updates the debugger and test262 tests to python3.

However some CI jobs also need to be updated to ubuntu-22.04, because ubuntu-18.04 is EOL: these jobs have been disabled for the time being.

### DIFF
--- a/jerry-debugger/jerry_client_main.py
+++ b/jerry-debugger/jerry_client_main.py
@@ -113,6 +113,7 @@ JERRY_DEBUGGER_VALUE_FUNCTION = 7
 JERRY_DEBUGGER_VALUE_ARRAY = 8
 JERRY_DEBUGGER_VALUE_OBJECT = 9
 
+
 def arguments_parse():
     parser = argparse.ArgumentParser(description="JerryScript debugger client")
 
@@ -290,8 +291,8 @@ class JerryDebugger(object):
         self.src_offset = 0
         self.src_offset_diff = 0
         self.non_interactive = False
-        self.current_out = b""
-        self.current_log = b""
+        self.current_out = ""
+        self.current_log = ""
         self.channel = channel
 
         config_size = 8
@@ -303,12 +304,12 @@ class JerryDebugger(object):
         # cpointer_size [1]
         result = self.channel.connect(config_size)
 
-        if len(result) != config_size or ord(result[0]) != JERRY_DEBUGGER_CONFIGURATION:
+        if len(result) != config_size or result[0] != JERRY_DEBUGGER_CONFIGURATION:
             raise Exception("Unexpected configuration")
 
-        self.little_endian = ord(result[1]) & JERRY_DEBUGGER_LITTLE_ENDIAN
-        self.max_message_size = ord(result[6])
-        self.cp_size = ord(result[7])
+        self.little_endian = result[1] & JERRY_DEBUGGER_LITTLE_ENDIAN
+        self.max_message_size = result[6]
+        self.cp_size = result[7]
 
         if self.little_endian:
             self.byte_order = "<"
@@ -396,7 +397,7 @@ class JerryDebugger(object):
                    "to clear all the given breakpoints\n "
         elif args in ['all', 'pending', 'active']:
             if args != "pending":
-                for i in self.active_breakpoint_list.values():
+                for i in list(self.active_breakpoint_list.values()):
                     breakpoint = self.active_breakpoint_list[i.active_index]
                     del self.active_breakpoint_list[i.active_index]
                     breakpoint.active_index = -1
@@ -557,6 +558,7 @@ class JerryDebugger(object):
         self._exec_command(JERRY_DEBUGGER_MEMSTATS)
 
     def _send_string(self, args, message_type, index=0):
+        args = args.encode('utf8')
 
         # 1: length of type byte
         # 4: length of an uint32 value
@@ -680,7 +682,7 @@ class JerryDebugger(object):
             if not data:  # Break the while loop if there is no more data.
                 return DebuggerAction(DebuggerAction.END, "")
 
-            buffer_type = ord(data[0])
+            buffer_type = data[0]
             buffer_size = len(data) -1
 
             logging.debug("Main buffer type: %d, message size: %d", buffer_type, buffer_size)
@@ -733,11 +735,8 @@ class JerryDebugger(object):
                 self.prompt = True
                 return DebuggerAction(DebuggerAction.TEXT, result)
 
-            elif buffer_type == JERRY_DEBUGGER_EXCEPTION_STR:
-                self.exception_string += data[1:]
-
-            elif buffer_type == JERRY_DEBUGGER_EXCEPTION_STR_END:
-                self.exception_string += data[1:]
+            elif buffer_type in [JERRY_DEBUGGER_EXCEPTION_STR, JERRY_DEBUGGER_EXCEPTION_STR_END]:
+                self.exception_string += data[1:].decode('utf8')
 
             elif buffer_type == JERRY_DEBUGGER_BACKTRACE_TOTAL:
                 total = struct.unpack(self.byte_order + self.idx_format, data[1:])[0]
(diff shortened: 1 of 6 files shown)


zherczeg (approved):
LGTM. Of course we need to re-enable the disabled checks as next steps.

jerry-debugger/jerry_client_main.py
```diff
@@ -734,10 +736,10 @@ def process_messages(self):
                 return DebuggerAction(DebuggerAction.TEXT, result)
 
             elif buffer_type == JERRY_DEBUGGER_EXCEPTION_STR:
-                self.exception_string += data[1:]
+                self.exception_string += data[1:].decode('latin1')
 
             elif buffer_type == JERRY_DEBUGGER_EXCEPTION_STR_END:
-                self.exception_string += data[1:]
+                self.exception_string += data[1:].decode('latin1')
```
akosthekiss, line 742: Is Latin-1 encoding OK for exception strings? Cannot they contain arbitrary Unicode characters?

Btw, this could be simplified to `elif buffer_type in [JERRY_DEBUGGER_EXCEPTION_STR, JERRY_DEBUGGER_EXCEPTION_STR_END]:`, just like elsewhere in the code.

jerry-debugger/jerry_client_main.py
```diff
@@ -885,13 +887,13 @@ def _parse_source(self, data):
...
-                source_code += data[1:]
+                source_code += data[1:].decode('latin1')
 
             elif buffer_type in [JERRY_DEBUGGER_SOURCE_CODE_NAME, JERRY_DEBUGGER_SOURCE_CODE_NAME_END]:
-                source_code_name += data[1:]
+                source_code_name += data[1:].decode('latin1')
 
             elif buffer_type in [JERRY_DEBUGGER_FUNCTION_NAME, JERRY_DEBUGGER_FUNCTION_NAME_END]:
-                function_name += data[1:]
+                function_name += data[1:].decode('latin1')
```
akosthekiss, line 896: Same question here for function names (cannot they contain arbitrary Unicode chars?)

jerry-debugger/jerry_client_main.py
```diff
@@ -885,13 +887,13 @@ def _parse_source(self, data):
                 return ""
 
             elif buffer_type in [JERRY_DEBUGGER_SOURCE_CODE, JERRY_DEBUGGER_SOURCE_CODE_END]:
-                source_code += data[1:]
+                source_code += data[1:].decode('latin1')
 
             elif buffer_type in [JERRY_DEBUGGER_SOURCE_CODE_NAME, JERRY_DEBUGGER_SOURCE_CODE_NAME_END]:
-                source_code_name += data[1:]
+                source_code_name += data[1:].decode('latin1')
```
akosthekiss, line 893: Same question here for source code names.

jerry-debugger/jerry_client_main.py
```diff
@@ -1176,17 +1178,17 @@ def _process_scope_variables(self):
 
         while buff_pos != buff_size:
             # Process name
-            name_length = ord(self.scope_vars[buff_pos:buff_pos + 1])
+            name_length = int.from_bytes(self.scope_vars[buff_pos:buff_pos + 1].encode('latin1'))
```
akosthekiss, line 1181: Why is this rewrite necessary? `self.scope_vars` is a string. `self.scope_vars[buff_pos:buff_pos + 1]` is a single-character string. Encoding it to latin1 gives a single-byte byte array containing the codepoint of the character. Then, the single byte is converted to an integer. How is this different from calling `ord()` on the character?

Notes:
- `self.scope_vars[buff_pos:buff_pos + 1]` can be simplified to `self.scope_vars[buff_pos]`.
- `int.from_bytes`, as used here, is not generic python3-compatible, as it needed a second argument up until python 3.10.

jerry-debugger/jerry_client_main.py
```diff
@@ -1176,17 +1178,17 @@ def _process_scope_variables(self):
...
             buff_pos += 1
             name = self.scope_vars[buff_pos:buff_pos + name_length]
             buff_pos += name_length
 
             # Process type
-            value_type = ord(self.scope_vars[buff_pos:buff_pos + 1])
+            value_type = int.from_bytes(self.scope_vars[buff_pos:buff_pos + 1].encode('latin1'))
```
akosthekiss, line 1187: same here

jerry-debugger/jerry_client_main.py
```diff
@@ -1176,17 +1178,17 @@ def _process_scope_variables(self):
...
 
             buff_pos += 1
 
-            value_length = ord(self.scope_vars[buff_pos:buff_pos + 1])
+            value_length = int.from_bytes(self.scope_vars[buff_pos:buff_pos + 1].encode('latin1'))
```
akosthekiss, line 1191: same here

akosthekiss (changes requested):

jerry-debugger/jerry_client_main.py
```diff
@@ -804,7 +806,7 @@ def process_messages(self):
                 return DebuggerAction(DebuggerAction.TEXT, result)
 
             elif buffer_type in [JERRY_DEBUGGER_SCOPE_VARIABLES, JERRY_DEBUGGER_SCOPE_VARIABLES_END]:
-                self.scope_vars += "".join(data[1:])
+                self.scope_vars += data[1:].decode()
```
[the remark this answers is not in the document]
matetokodi: Yes, it should be.

jerry-debugger/jerry_client_main.py
```diff
@@ -396,7 +397,7 @@ def delete(self, args):
                    "to clear all the given breakpoints\n "
         elif args in ['all', 'pending', 'active']:
             if args != "pending":
-                for i in self.active_breakpoint_list.values():
+                for i in list(self.active_breakpoint_list.values()):
```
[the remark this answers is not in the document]
matetokodi: It is iterable, however the `del ...` a few lines below causes a `RuntimeError: dictionary changed size during iteration`, and using `list()` creates a copy of the list so we can remove items from the original, without changing the size of the iterable object.

jerry-debugger/jerry_client_main.py
```diff
@@ -885,13 +887,13 @@ def _parse_source(self, data):
                 return ""
 
             elif buffer_type in [JERRY_DEBUGGER_SOURCE_CODE, JERRY_DEBUGGER_SOURCE_CODE_END]:
-                source_code += data[1:]
+                source_code += data[1:].decode('latin1')
```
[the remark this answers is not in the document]
matetokodi: Yes utf-8 should be used.

Originally I used `latin1` in some places because in some cases due to how the data was in the buffer, a `utf-8` byte sequence could be split apart, and the the first part of it could not be decoded, causing an error; but after further investigation decoding the bytes elsewhere, where their values are used eliminates this issue.

jerry-debugger/jerry_client_main.py
matetokodi: Yes utf-8 should be used.

same reason as here: https://github.com/jerryscript-project/jerryscript/pull/5095#discussion_r1330545099

jerry-debugger/jerry_client_main.py
matetokodi: Yes utf-8 should be used.

same reason as here: https://github.com/jerryscript-project/jerryscript/pull/5095#discussion_r1330545099

jerry-debugger/jerry_client_main.py
matetokodi: Yes utf-8 should be used.

same reason as here: https://github.com/jerryscript-project/jerryscript/pull/5095#discussion_r1330545099

jerry-debugger/jerry_client_main.py
matetokodi: It is not necessary.

Originally I removed all `ord` calls, which in hindsight was a mistake, then I reinvented the wheel without realizing that `ord` already did what was needed

jerry-debugger/jerry_client_main.py
matetokodi: same as here: https://github.com/jerryscript-project/jerryscript/pull/5095#discussion_r1330582748

jerry-debugger/jerry_client_main.py
matetokodi: same as here: https://github.com/jerryscript-project/jerryscript/pull/5095#discussion_r1330582748

jerry-debugger/jerry_client_main.py
```diff
@@ -985,7 +987,7 @@ def _parse_source(self, data):
             logging.debug("Pending breakpoints available")
             bp_list = self.pending_breakpoint_list
 
-            for breakpoint_index, breakpoint in bp_list.items():
+            for breakpoint_index, breakpoint in list(bp_list.items()):
```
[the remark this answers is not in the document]
akosthekiss: Indeed.

akosthekiss (approved):
lgtm

PR #5096, opened 2023-09-21
Re-enable pylint CI checker

matetokodi:
This patch re-enables the pylint CI checker

Adjust pylintrc to account for new checkers introduced since updating from ubuntu-18.04 to ubuntu-22-04: ignore imports, disable some style-based checks: e.g: consider-using-f-string, no-else-continue, among others.

Update `test262-harness.py` to use `argparse` instead of the now deprecated `optparse`.

Rename variables in `jerry_client_main.py` that redefined python builtins or shadowed variables from an outer scope


.github/workflows/gh-actions.yml
```diff
@@ -18,6 +18,7 @@ jobs:
       - run: sudo apt update
 # TODO: update checkers to current versions available in ubuntu 22.04
 #      - run: sudo apt install doxygen clang-format-10 cppcheck pylint python-serial
+      - run: sudo apt install pylint
```
zherczeg, line 21: I would keep the rest in the comment. I suspect they will be needed.

tools/pylint/pylintrc
```diff
@@ -59,7 +59,7 @@ confidence=
...
+disable=import-star-module-level,old-octal-literal,oct-method,unpacking-in-except,parameter-unpacking,backtick,old-raise-syntax,old-ne-operator,long-suffix,dict-view-method,dict-iter-method,metaclass-assignment,next-method-called,raising-string,indexing-exception,raw_input-builtin,long-builtin,file-builtin,execfile-builtin,coerce-builtin,cmp-builtin,buffer-builtin,basestring-builtin,apply-builtin,filter-builtin-not-iterating,using-cmp-argument,useless-suppression,range-builtin-not-iterating,suppressed-message,no-absolute-import,old-division,cmp-method,reload-builtin,zip-builtin-not-iterating,intern-builtin,unichr-builtin,reduce-builtin,standarderror-builtin,unicode-builtin,xrange-builtin,coerce-method,delslice-method,getslice-method,setslice-method,input-builtin,round-builtin,hex-method,nonzero-method,map-builtin-not-iterating,missing-docstring,locally-disabled,consider-using-f-string,unspecified-encoding,consider-using-from-import,consider-using-with,consider-using-in,no-else-return,no-else-continue,no-else-break,consider-using-sys-exit,chained-comparison,useless-object-inheritance,no-else-raise,singleton-comparison,redundant-u-string-prefix,import-outside-toplevel
```
zherczeg, line 62: Is there new additions here?

tools/pylint/pylintrc
```diff
@@ -297,7 +297,7 @@ ignore-comments=yes
 ignore-docstrings=yes
 
 # Ignore imports when computing similarities.
-ignore-imports=no
+ignore-imports=yes
```
zherczeg, line 300: What is the effect of this change?

zherczeg (changes requested):

.github/workflows/gh-actions.yml
matetokodi: I was thinking they would the added back as needed, also some of their names are different in ubuntu-22.04.

zherczeg: The comment shows the missing tools, regardless of the new name. If we delete it, we don't know which tools are still not added back. Instead, we should delete them one by one when they are added back.

tools/pylint/pylintrc
zherczeg: How much these affect our code? Linting looses its purpose if we disable everything. If there are not much changes, we could enable them.

tools/pylint/pylintrc
zherczeg: Maybe we could rewrite the code to avoid duplicates, if it is not too ugly.

tools/pylint/pylintrc
matetokodi: the big one is `consider-using-f-string`, that affects most places strings are formatted, the rest only affect a couple lines each

tools/pylint/pylintrc
matetokodi: Renaming the `get_arguments` function in one of the files is enough.

tools/pylint/pylintrc
akosthekiss: Do we have an exact minimum requirement for python version? f-strings have "only" been supported since py36. It may not be a big deal if we consider that officially even py37 is EOL, but some developers my be stuck on older systems where old python versions may only be available. (As I recall, that used to be the reason for using py2 for a very long time.)

tools/runners/test262-harness.py
```diff
@@ -417,8 +418,8 @@ def dispose(self):
         try:
             self.close()
             os.unlink(self.name)
-        except OSError as exception:
-            logging.error("Error disposing temp file: %s", str(exception))
+        except OSError as os_error:
+            logging.error("Error disposing temp file: %s", str(os_error))
```
akosthekiss, line 422: why the str() around os_error? the %s in the format string will force conversion to string, exactly via str().

tools/pylint/pylintrc
akosthekiss, line 62: having all checkers listed on a single line is quite unreadable and makes spotting changes in the list quite hard. an alternative approach is to list every checker on a single line, ending in a comma. then, every change (addition, removal) can be easily recognized in a diff.

e.g.:
```
disable=
    import-star-module-level,
    old-octal-literal,
    oct-method,
    unpacking-in-except,
    parameter-unpacking,
    ...
```

(an even better approach is to have the list in alphabetical order)

matetokodi: I do not think we have a set minimum, and for now I left `consider-using-f-string` disabled, and fixed the rest.

tools/runners/test262-harness.py
```diff
@@ -861,15 +862,15 @@ def run(self, command_template, tests, print_summary, full_summary, logname, job
...
-                pool.join()
+            with multiprocessing.Pool(processes=job_count, initializer=pool_init) as pool:
+                try:
+                    for result in pool.imap(test_case_run_process, cases):
+                        if logname:
+                            self.write_log(result)
+                        progress.has_run(result)
+                except KeyboardInterrupt:
+                    pool.terminate()
+                    pool.join()
```
akosthekiss, line 873: is the try-except construct still needed? when multiprocessing Pool is used as a context manager, it automatically calls terminate on exit (even for other types of exceptions, not only for KeyboardInterrupts)

tools/runners/test262-harness.py
```diff
@@ -609,18 +610,18 @@ def execute(command):
...
-            timer.cancel()
-            out = stdout.read()
-            err = stderr.read()
+            ) as process:
+                timer = threading.Timer(TEST262_CASE_TIMEOUT, process.kill)
+                timer.start()
+                code = process.wait()
+                timer.cancel()
+                out = stdout.read()
+                err = stderr.read()
```
akosthekiss, line 624: I am a bit confused about this code here. process.wait() has a timeout since py3.3, so this seems to be overly complex now. the code below _might_ achieve the same effect but simpler. (disclaimer: the snippet has not been tested)

```py
with subprocess.Popen(...) as process:
    try:
        process.wait(timeout=TEST262_CASE_TIMEOUT)
    except TimeoutExpired:
        process.kill()
out = stdout.read()
err = stderr.read()
```

akosthekiss (commented):
I realize that some of my comments are not really related to pylint-induced changes but more to py2-vs-py3 differences. Perhaps they could be separated into another patch that deals with py3 transition.

tools/runners/test262-harness.py
matetokodi: no, I do not think its still needed

matetokodi:
Over time this became an overall Python modernization patch, with the following additions:
- Use f-strings
- Remove `from __future__` imports
- Remove shebangs from non-executable files
- Re-enable all pylint checkers except `missing-docstring`
- Set minimum python version to 3.6

zherczeg (approved):
LGTM

tools/build.py
```diff
@@ -271,11 +269,10 @@ def make_jerry(arguments):
     make_cmd = ['cmake', '--build', arguments.builddir, '--config', arguments.build_type]
     env = dict(os.environ)
     env['CMAKE_BUILD_PARALLEL_LEVEL'] = str(arguments.jobs)
-    env['MAKEFLAGS'] = '-j%d' % (arguments.jobs) # Workaround for CMake < 3.12
-    proc = subprocess.Popen(make_cmd, env=env)
-    proc.wait()
-
-    return proc.returncode
+    env['MAKEFLAGS'] = f'-j{arguments.jobs}'
```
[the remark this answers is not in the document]
matetokodi: It was not intentional, but looking into cmake versions, most likely no one today uses cmake<3.12, as the current latest version is 3.27.7, and the latest in ubuntu 22.04 is 3.22.1, so I have removed this workaround.

tools/runners/run-test-suite-test262.py
```diff
@@ -167,49 +166,46 @@ def main(args):
...
+                    if (counter % 5000) == 0:
+                        print(f" Executed {counter} tests.")
+
+        proc.wait()
+
+        if sys.platform == 'win32':
+            util.set_timezone(original_timezone)
+
+        if args.mode == 'update':
+            return_code = update_exclude_list(args)
```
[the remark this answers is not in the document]
matetokodi: Yes they shouldn't be, my bad

When I initially was going over the code I failed to see that `return_code` was defined at the beginning of the function, and not just before `with open...output_file:`, so I put the rest of the function inside the context manager to make sure return_code is defined at the end as well.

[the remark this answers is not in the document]
akosthekiss: Note: the scoping rules of Python are not C-like. A variable does not need to be defined at the same (or higher) "indentation level" as its usage. Even if `return_code` was not defined at the beginning of the function, but within the context only, the code would work.

E.g.:
```py
>>> def func(x):
...   if (x):
...     r = 'YES'
...   else:
...     r = 'NO'
...   return r
... 
>>> func(True)
'YES'
>>> func(False)
'NO'
```

akosthekiss (approved):
lgtm

PR #5108, opened 2023-11-14
Update cppcheck

matetokodi:
This patch depends on https://github.com/jerryscript-project/jerryscript/pull/5106
Once that is merged, this patch will be rebased on only include the cppcheck commit.

In this patch:
- Re-enable cppcheck CI job
- Update cppcheck suppression list:
        The new version of cppcheck raises warnings for many potential
        issues that are guarded against, so those warnings have been
        supressed.
- Handle realloc failures:
    - `jerry-ext/util/sources.c`
    - `jerry-port/common/jerry-port-io.c`


jerry-ext/util/sources.c
```diff
@@ -145,7 +145,11 @@ jerryx_source_exec_stdin (void)
     }
 
     jerry_size_t new_size = source_size + line_size;
-    source_p = realloc (source_p, new_size);
+    jerry_char_t *tmp = realloc (source_p, new_size);
+    if (tmp == NULL) {
+      return jerry_throw_sz(JERRY_ERROR_COMMON, "Out of memory.");
+    }
+    source_p = tmp;
```
akosthekiss, line 152: is a new temporary needed? why not simply
```c
source_p = realloc (source_p, new_size);
if (source_p == NULL) {
  return jerry_throw_sz(JERRY_ERROR_COMMON, "Out of memory.");
}
```
?

jerry-port/common/jerry-port-io.c
```diff
@@ -70,7 +70,12 @@ jerry_port_line_read (jerry_size_t *out_size_p)
   while (true)
   {
     allocated += 64;
-    line_p = realloc (line_p, allocated);
+    char *tmp = realloc (line_p, allocated);
+    if (tmp == NULL)
+    {
+      jerry_port_fatal (JERRY_FATAL_OUT_OF_MEMORY);
+    }
+    line_p = tmp;
```
akosthekiss, line 78: same here.
```c
line_p = realloc (line_p, allocated);
if (line_p == NULL)
{
  jerry_port_fatal (JERRY_FATAL_OUT_OF_MEMORY);
}
```
?

tests/unit-core/test-api.c
```diff
@@ -709,12 +709,12 @@ main (void)
   if (jerry_feature_enabled (JERRY_FEATURE_PROXY))
   {
     jerry_value_t target = jerry_object ();
-    jerry_value_t handler = jerry_object ();
-    jerry_value_t proxy = jerry_proxy (target, handler);
+    jerry_value_t handler_ = jerry_object ();
+    jerry_value_t proxy = jerry_proxy (target, handler_);
```
akosthekiss, line 713: this isn't nice. what is the problem of cppcheck with handler?

tests/unit-core/test-snapshot.c
```diff
@@ -284,11 +284,8 @@ main (void)
     jerry_cleanup ();
 
     test_exec_snapshot (snapshot_buffer, snapshot_size, JERRY_SNAPSHOT_EXEC_ALLOW_STATIC);
-  }
 
-  /* Merge snapshot */
-  if (jerry_feature_enabled (JERRY_FEATURE_SNAPSHOT_SAVE) && jerry_feature_enabled (JERRY_FEATURE_SNAPSHOT_EXEC))
-  {
```
akosthekiss, line 291: I think that instead of merging these two blocks, which are actually two separate test cases that happen to be guarded by the same condition, all test cases of `main` should be organized into separate test functions. 

interestingly, at the end of `main`, there are already calls to three test cases organized into functions (`test_function_snapshot`, `test_function_arguments_snapshot`, `test_snapshot_with_user`). but for some reason, the first three test cases are still big blocks with main (prefixed with comments `/* Static snapshot */`, `/* Merge snapshot */`, and `/* Save literals */`). let's refactor them.

akosthekiss (changes requested):

tests/unit-core/test-api.c
matetokodi: The problem was that is shadowed the `handler` function defined at line 64.
I agree its not a good name, it was just a placeholder.

zherczeg (approved):
LGTM

akosthekiss (approved):
lgtm

PR #5142, opened 2024-05-17
Fix CI for macos-latest

matetokodi:
Add a non-inlineable function for negating a number on macos, bacause due to a compiler bug `r = -r` was being optimized out, resulting in some modulo operations failing such as `-1 % -1` or `-1 % 1`.

Disable `-Wliteral-range` for `test-math.c` on macos, because it was falsely raised for `isnan`, `isinf`, and `isfinite` macros.

### DIFF
--- a/.github/workflows/gh-actions.yml
+++ b/.github/workflows/gh-actions.yml
@@ -57,7 +57,7 @@ jobs:
       - run: $RUNNER -q --jerry-tests --buildoptions=--compile-flag=-m32,--cpointer-32bit=on --build-debug
 
   OSX_x86-64_Build_Correctness_Unit_Tests:
-    runs-on: macos-latest
+    runs-on: macos-13
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-python@v4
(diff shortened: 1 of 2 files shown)


zherczeg (approved):
LGTM

jerry-core/ecma/base/ecma-helpers-number.c
```diff
@@ -370,7 +373,11 @@ ecma_number_remainder (ecma_number_t left_num, /**< left operand */
 
   if (ecma_number_is_zero (r) && ecma_number_is_negative (left_num))
   {
+#ifdef __APPLE__
+    r = ecma_number_negate (r);
+#else /* !__APPLE__ */
```
akosthekiss, line 378: Is this the right macro guard to use? If this is a compiler bug then is it OK to use a guard on the platform? Which versions of which compiler are affected? Is there an issue report for this anywhere that could be linked for reference? Is there no cli option for the compiler to ensure that signed/negative zeros are properly handled, to work around the issue without code base change?

matetokodi: This bug only occurs on `macos-14`, which by default uses `Clang`, Version: `15.0.0.15000040`
this seems to be describing the same problem: https://github.com/llvm/llvm-project/issues/82744 (for the first issue, I was unable to find a report for the second)

These issues first popped up when `macos-latest` in github actions switched from `macos-13` to `macos-14`.
The easiest fix for now without a codebase change would be to go back to using `macos-13` for the time being, and switch back to `macos-latest` at a leter date when these issues have hopefully been resolved.

zherczeg (approved):
LGTM

akosthekiss (approved):
lgtm

Issue #5141, opened 2024-05-16
Assertion in parser_check_duplicated_private_field


Qbtly:
###### Commit ID
47bd5d4b910a41fdc517e39edcf9220d8030a61f

###### Build platform
Ubuntu 22.04.3

###### Build steps
```sh
python3 ./tools/build.py --builddir=xxx --debug --clean --compile-flag=-fsanitize=address --compile-flag=-g --strip=off --lto=off --logging=on --line-info=on --error-message=on --stack-limit=20
```

###### Test case
```sh
class C { #get [Symbol]; }
```

###### Execution steps
```sh
./jerry poc.js
```

###### Output

```sh
Debug:
ICE: Assertion 'context_p->token.type == LEXER_LITERAL' failed at /jerryscript/jerry-core/parser/js/js-parser-expr.c(parser_check_duplicated_private_field):408.
Error: JERRY_FATAL_FAILED_ASSERTION

pwndbg> bt
#0  __pthread_kill_implementation (no_tid=0, signo=6, threadid=140737350398144) at ./nptl/pthread_kill.c:44
#1  __pthread_kill_internal (signo=6, threadid=140737350398144) at ./nptl/pthread_kill.c:78
#2  __GI___pthread_kill (threadid=140737350398144, signo=signo@entry=6) at ./nptl/pthread_kill.c:89
#3  0x00007ffff7cb2476 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#4  0x00007ffff7c987f3 in __GI_abort () at ./stdlib/abort.c:79
#5  0x0000555555874565 in jerry_port_fatal (code=code@entry=JERRY_FATAL_FAILED_ASSERTION) at /jerryscript/jerry-port/common/jerry-port-process.c:29
#6  0x00005555557624d8 in jerry_fatal (code=code@entry=JERRY_FATAL_FAILED_ASSERTION) at /jerryscript/jerry-core/jrt/jrt-fatals.c:63
#7  0x000055555576238c in jerry_assert_fail (assertion=<optimized out>, file=<optimized out>, function=<optimized out>, line=line@entry=408) at /jerryscript/jerry-core/jrt/jrt-fatals.c:83
#8  0x0000555555848162 in parser_check_duplicated_private_field (context_p=context_p@entry=0x7fffffffde60, opts=<optimized out>) at /jerryscript/jerry-core/parser/js/js-parser-expr.c:408
[...]

PR #5144, opened 2024-06-12
Fix parsing of invalid private class properties

matetokodi:
Raise syntax error instead of failing with an assert.

Fixes https://github.com/jerryscript-project/jerryscript/issues/5141


zherczeg (approved):
LGTM

akosthekiss (approved):
lgtm

Issue #5134, opened 2024-03-26
Heap-Buffer-Overflow  in lexer_convert_ident_to_cesu8 /jerryscript/jerry-core/parser/js/js-lexer.c:2083:9


Qbtly:
###### JerryScript revision
cefd391772529c8a9531d7b3c244d78d38be47c6

###### Build platform
Ubuntu 22.04.3

###### Build steps
```sh
python ./tools/build.py --builddir=xxx --clean --compile-flag=-fsanitize=address --compile-flag=-g --strip=off --lto=off --logging=on --line-info=on --error-message=on --stack-limit=20
```

###### Test case
```sh
import{a as "\{{91406,456}"
```

###### Execution steps
```sh
./xxx/bin/jerry poc.js
```

###### Output
```sh
Release:
=================================================================
==2144424==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x60300000005c at pc 0x559a32bf48a3 bp 0x7ffedf4f4450 sp 0x7ffedf4f4448
READ of size 1 at 0x60300000005c thread T0
    #0 0x559a32bf48a2 in lexer_convert_ident_to_cesu8 /jerryscript/jerry-core/parser/js/js-lexer.c:2083:9
    #1 0x559a32bf4e08 in lexer_convert_literal_to_chars /jerryscript/jerry-core/parser/js/js-lexer.c:2133:5
    #2 0x559a32bf5d19 in lexer_construct_literal_object /jerryscript/jerry-core/parser/js/js-lexer.c:2367:5
    #3 0x559a32b707db in scanner_check_variables /jerryscript/jerry-core/parser/js/js-scanner-util.c:2279:5
    #4 0x559a32b67485 in parser_parse_source /jerryscript/jerry-core/parser/js/js-parser.c:2274:9
    #5 0x559a32b65924 in parser_parse_script /jerryscript/jerry-core/parser/js/js-parser.c:3332:38
    #6 0x559a32ac2f38 in jerry_parse_common /jerryscript/jerry-core/api/jerryscript.c:418:21
    #7 0x559a32ac2d34 in jerry_parse /jerryscript/jerry-core/api/jerryscript.c:486:10
    #8 0x559a32c2876f in jerryx_source_parse_script /jerryscript/jerry-ext/util/sources.c:52:26
    #9 0x559a32c2892f in jerryx_source_exec_script /jerryscript/jerry-ext/util/sources.c:63:26
[...]

PR #5146, opened 2024-06-24
Raise syntax error on invalid unicode escape sequences

matetokodi:
Raise syntax error on invalid unicode escape sequences instead of failing with an assertion

The parser context was required for raising the error, which had to be added to all functions that can call `lexer_unchecked_hex_to_character`.

This fixes https://github.com/jerryscript-project/jerryscript/issues/5134


zherczeg:
This patch adds a lot of extra operations. I am thinking why this was not needed before, how we handled this in other cases.

PR #5147, opened 2024-07-26
Add build argument: --arch

winsoft666:
Add `--arch` build argument to special CMake `-A` option.

For example:
On Windows, we can use this command to build 32 bit library:
```bash
python tools/build.py --arch Win32
```


zherczeg (approved):
LGTM

zherczeg (approved):
LGTM

PR #5148, opened 2024-07-26
Add build argument: --static-crt

winsoft666:
On Windows, We can specify whether the program/library uses static CRT (/MT or /MTd) or dynamic CRT (/MD or /MDd).

I add `--static-crt` argument to specify library to use static CRT.

```bash
python tools/build.py --static-crt=ON
```


LaszloLango (approved):

zherczeg (approved):
LGTM

PR #4898, opened 2021-12-09
Fix Typedarray.slice fastpath when the content type is matching

rerobika:
This patch fixes #4888.

JerryScript-DCO-1.0-Signed-off-by: Robert Fancsik robert.fancsik@h-lab.eu

PR #5177, opened 2024-11-25
Fix Typedarray.slice fastpath when the content type is matching

### LINKED ISSUE #4888: Assertion 'new_typedarray_info.offset == 0' failed at jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c(ecma_builtin_typedarray_prototype_slice):1631.
## JerryScript commit hash
55acdf2048b390d0f56f12e64dbfb2559f0e70ad

## Build platform
Ubuntu 20.04 LTS

## Build steps
```
--compile-flag=-fno-common --compile-flag=-g \
--strip=off --system-allocator=on --logging=on \
--linker-flag=-fuse-ld=gold --error-messages=on --line-info=ON \
--stack-limit=10
```

## poc
```
var ab = new Int8Array(20).map((v, i) => i).buffer;
var ta = new Int8Array(ab, 0, 10);
var seen_length = -1;
ta.constructor = {
    [Symbol.species]: function (len) {
        seen_length = len;
        return new Int8Array(ab, 1, len);
    }
};
print(-1, seen_length);
print([
    0,
    1,
    2,
    3,
    4,
    5,
    6,
    7,
    8,
    9
], ta);
var tb = ta.slice();
print(10, seen_length);
print([
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0
], ta);
print([
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0,
    0
], tb);
```

## assert log
```
ICE: Assertion 'new_typedarray_info.offset == 0' failed at /home/sakura/jerryscript/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c(ecma_builtin_typedarray_prototype_slice):1631.

```

gergocs:
This patch fixes #4888.

The implementation is based on PR #4898, only resolved the conflicts and
applied requested changes.

Co-authored-by: Robert Fancsik robert.fancsik@h-lab.eu
JerryScript-DCO-1.0-Signed-off-by: Gergo Csizi gergocs@inf.u-szeged.hu

### DIFF
--- a/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
+++ b/jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
@@ -1711,14 +1711,35 @@ ecma_builtin_typedarray_prototype_slice (ecma_value_t this_arg, /**< this argume
 
   uint8_t *dst_buffer_p = ecma_typedarray_get_buffer (&new_typedarray_info);
 
-  JERRY_ASSERT (new_typedarray_info.offset == 0);
-
   src_buffer_p += relative_start << info_p->shift;
 
   if (info_p->id == new_typedarray_info.id)
   {
     // 22.2.3.23. Step 22. h-i.
-    memcpy (dst_buffer_p, src_buffer_p, count << info_p->shift);
+
+    if (JERRY_LIKELY (new_typedarray_info.offset == 0))
+    {
+      if (info_p->array_buffer_p == new_typedarray_info.array_buffer_p)
+      {
+        return new_typedarray;
+      }
+
+      memcpy (dst_buffer_p, src_buffer_p, count << info_p->shift);
+    }
+    else
+    {
+      uint32_t byte_shift = (uint32_t) (1 << info_p->shift);
+
+      while (count)
+      {
+        memmove (dst_buffer_p, src_buffer_p, byte_shift);
+
+        dst_buffer_p += byte_shift;
+        src_buffer_p += byte_shift;
+
+        count--;
+      }
+    }
   }
   else
   {
--- a/tests/jerry/array-slice.js
+++ b/tests/jerry/array-slice.js
@@ -1,0 +1,33 @@
+// Copyright JS Foundation and other contributors, http://js.foundation
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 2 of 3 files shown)


jerry-core/ecma/builtin-objects/typedarray/ecma-builtin-typedarray-prototype.c
```diff
@@ -1711,14 +1711,35 @@ ecma_builtin_typedarray_prototype_slice (ecma_value_t this_arg, /**< this argume
...
+    if (JERRY_LIKELY (new_typedarray_info.offset == 0))
+    {
+      if (info_p->array_buffer_p == new_typedarray_info.array_buffer_p)
+      {
+        return new_typedarray;
+      }
+
+      memcpy (dst_buffer_p, src_buffer_p, count << info_p->shift);
+    }
+    else if (count >= new_typedarray_info.offset)
```
LaszloLango, line 1729: Just to be fair, there was a comment on the original PR which has not been answered yet. I paste it here.

> What if count is less than the target offset? I don't think this condition is correct.
> 
> ```
> var buf = new ArrayBuffer(10);
> var a1 = new Int8Array(buf, 0, 5);
> a1.fill(1);
> a1.constructor = {
>     [Symbol.species]: function (len) {
>         return new Int8Array(buf, 5, 5);
>     }
> };
> var a2 = a1.slice(2,4);
> res = new Int8Array(buf, 0, 10);
> ```
> 
> Also, I don't think checking the offset is the right way, we need to check whether the two typedarrays use the same arraybuffer instead, and only then fall back to copying element-by-element.

_Originally posted by @dbatyai in https://github.com/jerryscript-project/jerryscript/pull/4898#discussion_r766609489_

gergocs: I checked the example and it worked just fine.

LaszloLango: Great, thanks.

LaszloLango (approved):
LGTM

robertsipka (changes requested):
I have concerns about the results. The test case linked by Dániel run without an error message, but the result is correct? Sadly, no, please check it, use the mentioned test case in the PR as well, and make a results validation for `res` in the last line. Dániel's review seems appropriate to me.

robertsipka (approved):
LGTM

LaszloLango (approved):
still LGTM

PR #5187, opened 2024-12-06
CMake update cmake_minimum_required (VERSION 3.10)

lygstate:
```
CMake Deprecation Warning at jerry-port/CMakeLists.txt:15 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required (VERSION 2.8.12)
+cmake_minimum_required (VERSION 3.10)
 project (Jerry C)
 
 if(NOT DEFINED PYTHON)
--- a/jerry-core/CMakeLists.txt
+++ b/jerry-core/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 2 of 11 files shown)


LaszloLango:
Why? Do we know that the project does not build with older cmake versions than 3.10?

lygstate:
> Why? Do we know that the project does not build with older cmake versions than 3.10?

Because newest cmake are deprecating cmake before 3.10

https://github.com/jerryscript-project/jerryscript/actions/runs/12069004842/job/33655344964#step:6:101

LaszloLango:
Fair enough, good to me. Thanks.

LaszloLango (approved):

zherczeg:
3.10 release date: 20 November 2017. Not that long ago.

ossy-szeged:
just a note: Ubuntu 18.04 - which already reached its EOL on 2023.05.31 - shipped cmake 3.10
I don't think it is a big thing to require at least 3.10 version. 

Not to mention that github CI has only cmake 3.31 on each supported Ubuntu runner (20.04, 22.04 and 24.04) and we don't test if the minimum cmake version works or not. Maybe we could add one more CI job to ensure that out minimum
requirement cmake version works - https://github.com/marketplace/actions/get-cmake

PR #5207, opened 2025-02-20
Remove deprecated enum conversions

kasperisager:
Implicit enum conversion was deprecated in C++20 with the result that JerryScript doesn't compile in newer versions of Clang, in my case LLVM 20. This patch removes implicit enum conversions and instead performs explicit casts.


kasperisager:
Is there anything I can do to help land this?

zherczeg (approved):
LGTM

kasperisager:
I'm hitting this on quite a few systems now, including macOS and Linux. A second review so this can land would be much appreciated!

PR #5235, opened 2025-06-26
Remove `CMAKE_C_ARCHIVE_*` variables

kasperisager:
It assumes that `libtool` is used when compiling for macOS, which isn't the case when using LLVM and `llvm-ranlib`.


kasperisager:
Solved by #5258.

PR #5241, opened 2025-07-22
 The skipping of null values during array processing has been removed

### LINKED ISSUE #5026: Array concatenation exception
# Version

*Version: 3.0.0* ( 6fe763f )

# Execution steps

```bash
/root/.jsvu/jerry Testcase.js
```

# Test case 1

```javascript
var foo = function(it) {
    var a = [...it, , ];
    print(a);
    print(a.length);
};
var Parameter0 = [1, 2];
foo(Parameter0);
```

## Output 1

```
1,2
2
```

## Expected behavior

```
1,2,
3
```

# Test case 2

```javascript
var foo = function(it) {
    var arr = [ , ];
    var a = it.concat(arr);
    print(a);
    print(a.length);
};
var Parameter0 = [1, 2];
foo(Parameter0);
```

## Output 2

```
1,2,
3
```

# Description

We tested on multiple JS engines, Testcase 1 and Testcase 2 should have the same consequent. But jerryscript outputs abnormal results when executing Testcase 1.

Maybe the problem is about array literal handling which looks like it deleted all the vacancies. The same bug has reported to Hermes(https://github.com/facebook/hermes/issues/729) and has been fixed. 

Looking forward to your reply :)

happy-boy-maker:
This patch fixes https://github.com/jerryscript-project/jerryscript/issues/5026
Signevior of skipping null values during array processing, which was inconsistent with the mainstream js engine behavior, has been removed.

JerryScript-DCO-1.0-Signed-off-by: tangbin 2387440390@qq.com

### DIFF
--- a/jerry-core/vm/opcodes.c
+++ b/jerry-core/vm/opcodes.c
@@ -326,11 +326,6 @@ opfunc_append_to_spread_array (ecma_value_t *stack_top_p, /**< current stack top
 
   for (uint32_t i = 0, idx = old_length; i < values_length; i++, idx++)
   {
-    if (ecma_is_value_array_hole (stack_top_p[i]))
-    {
-      continue;
-    }
-
     if (stack_top_p[i] == ECMA_VALUE_SPREAD_ELEMENT)
     {
       i++;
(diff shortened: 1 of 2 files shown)


zherczeg:
Was the old code an optimization, which is not working anymore?

happy-boy-maker:
Apart from code format errors, no new issues have been introduced in ci. The deleted code seems to skip whenever an empty value is encountered during the array processing, without any further purpose.

PR #5258, opened 2025-10-08
cmake: avoid Apple ranlib flags when not using AppleClang

renatahodovan:
On macOS, `-no_warning_for_no_symbols -c` options are specific to the Apple `ranlib` tool. When using other toolchains (e.g. Homebrew clang), these flags cause errors since `llvm-ranlib` does not recognize them.

The patch ensures these flags are only passed when using AppleClang.


akosthekiss (approved):
lgtm

zherczeg (approved):

PR #5262, opened 2025-10-18
chore: allow CMake though to 3.10

zmedico:
This is allows the build with cmake-4.0.0 without deprecation warnings.

use min...max syntax to allow build with newer cmake. ref: https://cmake.org/cmake/help/latest/command/cmake_minimum_required.html

Fixes:
CMake Error at CMakeLists.txt:1 (cmake_minimum_required):
  Compatibility with CMake < 3.5 has been removed from CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.

  Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

### DIFF
--- a/targets/baremetal-sdk/espressif/CMakeLists.txt
+++ b/targets/baremetal-sdk/espressif/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-cmake_minimum_required(VERSION 3.5)
+cmake_minimum_required(VERSION 3.5...3.10)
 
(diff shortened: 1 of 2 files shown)


LaszloLango:
I think we can simply bump the `cmake_minimum_required` to 3.10. I don't think we should support older cmake versions anymore.

zherczeg:
Yes, probably we could just move to a newer version.

